PR #1, opened 2018-06-07
Setup controller-platform repo from the template

pwittrock:
- https://github.com/kubernetes/kubernetes-template-project


grodrigues3:
LGTM.

Bots being enabled in test-infra

PR #6, opened 2018-06-08
Implement the ChannelSource

lichuqiang:
Address: https://github.com/kubernetes-sigs/kubebuilder/issues/254

### DIFF
--- a/pkg/runtime/inject/inject.go
+++ b/pkg/runtime/inject/inject.go
@@ -83,6 +83,21 @@ func SchemeInto(scheme *runtime.Scheme, i interface{}) (bool, error) {
 	return false, nil
 }
 
+// Stoppable is used by the ControllerManager to inject stop channel into Sources,
+// EventHandlers, Predicates, and Reconciles.
+type Stoppable interface {
+	InjectStopChannel(<-chan struct{}) error
+}
+
+// StopChannelInto will set stop channel on i and return the result if it implements Stoppable.
+// Returns false if i does not implement Stoppable.
+func StopChannelInto(stop <-chan struct{}, i interface{}) (bool, error) {
+	if s, ok := i.(Stoppable); ok {
+		return true, s.InjectStopChannel(stop)
+	}
+	return false, nil
+}
+
 // Func injects dependencies into i.
 type Func func(i interface{}) error
 
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -96,6 +96,9 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	if _, err := inject.InjectorInto(cm.SetFields, i); err != nil {
 		return err
 	}
+	if _, err := inject.StopChannelInto(cm.stop, i); err != nil {
+		return err
+	}
 	return nil
 }
 
--- a/pkg/source/example_test.go
+++ b/pkg/source/example_test.go
@@ -36,5 +36,9 @@ func ExampleKind() {
 // provided by the event.
 func ExampleChannel() {
 	events := make(chan event.GenericEvent)
-	ctrl.Watch(source.Channel(events), &handler.Enqueue{})
+
+	ctrl.Watch(
+		&source.Channel{Source: events},
+		&handler.Enqueue{},
+	)
 }
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -18,6 +18,7 @@ package source
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/kubernetes-sigs/controller-runtime/pkg/event"
 	"github.com/kubernetes-sigs/controller-runtime/pkg/handler"
@@ -28,6 +29,11 @@ import (
 
 	"github.com/kubernetes-sigs/controller-runtime/pkg/cache"
 	"github.com/kubernetes-sigs/controller-runtime/pkg/predicate"
+)
+
+const (
+	// defaultBufferSize is the default number of event notifications that can be buffered.
+	defaultBufferSize = 1024
 )
 
 // Source is a source of events (eh.g. Create, Update, Delete operations on Kubernetes Objects, Webhook callbacks, etc)
@@ -45,19 +51,128 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
+var _ Source = &Channel{}
+
 // Channel is used to provide a source of events originating outside the cluster
 // (eh.g. GitHub Webhook callback).  Channel requires the user to wire the external
 // source (eh.g. http handler) to write GenericEvents to the underlying channel.
-type Channel chan event.GenericEvent
+type Channel struct {
+	// once ensures the event distribution goroutine will be performed only once
+	once sync.Once
 
-var _ Source = Channel(make(chan event.GenericEvent))
+	// Source is the source channel to fetch GenericEvents
+	Source <-chan event.GenericEvent
+
+	// stop is to end ongoing goroutine, and close the channels
+	stop <-chan struct{}
+
+	// dest is the destination channels of the added event handlers
+	dest []chan event.GenericEvent
+
+	// DestBufferSize is the specified buffer size of dest channels.
+	// Default to 1024 if not specified.
+	DestBufferSize int
+
+	// destLock is to ensure the destination channels are safely added/removed
+	destLock sync.Mutex
+}
+
+var _ inject.Stoppable = &Channel{}
+
+// InjectStopChannel is internal should be called only by the Controller.
+// It is used to inject the stop channel initialized by the ControllerManager.
+func (cs *Channel) InjectStopChannel(stop <-chan struct{}) error {
+	if cs.stop == nil {
+		cs.stop = stop
+	}
+
+	return nil
+}
 
 // Start implements Source and should only be called by the Controller.
-func (ks Channel) Start(
+func (cs *Channel) Start(
 	handler handler.EventHandler,
 	queue workqueue.RateLimitingInterface,
 	prct ...predicate.Predicate) error {
+	// Source should have been specified by the user.
+	if cs.Source == nil {
+		return fmt.Errorf("must specify Channel.Source")
+	}
+
+	// stop should have been injected before Start was called
+	if cs.stop == nil {
+		return fmt.Errorf("must call InjectStop on Channel before calling Start")
+	}
+
+	// use default value if DestBufferSize not specified
+	if cs.DestBufferSize == 0 {
+		cs.DestBufferSize = defaultBufferSize
+	}
+
+	cs.once.Do(func() {
+		// Distribute GenericEvents to all EventHandler / Queue pairs Watching this source
+		go cs.syncLoop()
+	})
+
+	dst := make(chan event.GenericEvent, cs.DestBufferSize)
+	go func() {
+		for evt := range dst {
+			shouldHandle := true
+			for _, p := range prct {
+				if !p.Generic(evt) {
+					shouldHandle = false
+					break
+				}
+			}
+
+			if shouldHandle {
+				handler.Generic(queue, evt)
+			}
+		}
+	}()
+
+	cs.destLock.Lock()
+	defer cs.destLock.Unlock()
+
+	cs.dest = append(cs.dest, dst)
+
 	return nil
+}
+
+func (cs *Channel) doStop() {
+	cs.destLock.Lock()
+	defer cs.destLock.Unlock()
+
+	for _, dst := range cs.dest {
+		close(dst)
+	}
+}
+
+func (cs *Channel) distribute(evt event.GenericEvent) {
+	cs.destLock.Lock()
+	defer cs.destLock.Unlock()
+
+	for _, dst := range cs.dest {
+		// We cannot make it under goroutine here, or we'll meet the
+		// race condition of writing message to closed channels.
+		// To avoid blocking, the dest channels are expected to be of
+		// proper buffer size. If we still see it blocked, then
+		// the controller is thought to be in an abnormal state.
+		dst <- evt
+	}
+}
+
+func (cs *Channel) syncLoop() {
+	for {
+		select {
+		case <-cs.stop:
+			// Close destination channels
+			cs.doStop()
+			return
+		case evt := <-cs.Source:
+			cs.distribute(evt)
+		}
+	}
 }
 
 // Kind is used to provide a source of events originating inside the cluster from Watches (eh.g. Pod Create)
--- a/pkg/source/source_test.go
+++ b/pkg/source/source_test.go
@@ -26,10 +26,11 @@ import (
 	"github.com/kubernetes-sigs/controller-runtime/pkg/source"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"k8s.io/client-go/util/workqueue"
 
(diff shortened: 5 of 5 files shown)


pkg/controller/controller_manager.go
```diff
@@ -166,6 +175,9 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 	cm.stop = stop
 	cm.informers.Start(stop)
 
+	// Start the outer informer
+	cm.outerInformer.Run(stop)
```
pwittrock: We should be consistent about Start vs Run in our naming

pkg/controller/source/channel_source.go
```diff
@@ -0,0 +1,293 @@
...
+	initialBufferSize = 1024
+)
+
+var csLog = logf.KBLog.WithName("source").WithName("ChannelSource")
+
+// ChannelSource is used to provide a source of events originating outside the cluster
+// (eh.g. GitHub Webhook callback).  ChannelSource requires the user to wire the external
+// source (eh.g. http handler) to write GenericEvents to the underlying channel.
+type ChannelSource struct {
+	informer *ChannelSourceInformer
```
pwittrock: Why do we need an informer for this?  I envisioned reading events from the Channel and passing them directly to the Event handler

pwittrock:
Hey can we keep this in another PR.  I think it would be useful for a SlackBotSource or a GitHubSource of events.

pkg/controller/source/source.go
```diff
@@ -46,18 +47,115 @@ type Source interface {
...
+	}
+
+	cs.once.Do(func() {
+		// Distribute GenericEvents to all EventHandler / Queue pairs Watching this source
+		go func() {
+			for {
+				select {
+				case <-cs.stop:
+					// Close source channel.
+					// TODO: should we close the source channel here?
```
lichuqiang: Not sure if we need to close the source channel here.
Maybe users meant to use it somewhere out of the controller, and handle its lifecycle themselves?

pkg/controller/manager.go
```diff
@@ -139,6 +143,12 @@ func (cm *controllerManager) injectInto(i interface{}) error {
 	if _, err := inject.DoInformers(cm.informers, i); err != nil {
```
lichuqiang: I guess we need a fix here.
We expect to have it inject whichever params that `i` implement, rather than return directly once an error occur

pkg/controller/source/source.go
```diff
@@ -46,18 +52,113 @@ type Source interface {
...
 	handler eventhandler.EventHandler,
 	queue workqueue.RateLimitingInterface,
 	prct ...predicate.Predicate) error {
+	// source should have been injected before Start was called.
+	if cs.source == nil {
+		return fmt.Errorf("must call DoSourceChannel on ChannelSource before calling Start")
+	}
+
+	// stop should have been injected before Start was called
+	if cs.stop == nil {
```
pwittrock: This is a bug in the error you copied :P This should be InjectSourceChannel instead of DoStop

pkg/controller/source/source.go
```diff
@@ -46,18 +52,113 @@ type Source interface {
...
+	dest []chan event.GenericEvent
+	// destLock is to ensure the destination channels are safely added/removed
+	destLock sync.Mutex
+}
+
+var _ inject.SourceChannel = &ChannelSource{}
+
+// InjectSourceChannel is internal should be called only by the Controller.
+// It is used to inject the source event channel initialized by the ControllerManager.
+func (cs *ChannelSource) InjectSourceChannel(c chan event.GenericEvent) error {
```
pwittrock: What do you think of having the user provide the channel when creating the object (like Type on KindSource)?

Private Injected fields are initialized by the ControllerManager.  Public fields should be supplied by the user.

pkg/controller/source/source.go
pwittrock: I think the lifecycle of the source channel should be completely managed by the user.  We shouldn't create it or close it for them.

pkg/controller/source/source.go
```diff
@@ -46,18 +52,113 @@ type Source interface {
...
+
 // ChannelSource is used to provide a source of events originating outside the cluster
 // (eh.g. GitHub Webhook callback).  ChannelSource requires the user to wire the external
 // source (eh.g. http handler) to write GenericEvents to the underlying channel.
-type ChannelSource chan event.GenericEvent
+type ChannelSource struct {
+	// once ensures the event distribution goroutine will be performed only once
+	once sync.Once
+	// source is the source channel to fetch GenericEvents
+	source chan event.GenericEvent
```
pwittrock: Wondering if this should be a read channel only.  What do you think?

pkg/controller/source/source.go
```diff
@@ -46,18 +52,113 @@ type Source interface {
...
+		go func() {
+			for {
+				select {
+				case <-cs.stop:
+					// Close source channel.
+					// TODO: should we close the source channel here?
+					close(cs.source)
+
+					// Close destination channels
+					cs.destLock.Lock()
```
pwittrock: not sure if this is a common pattern, but if we wrap this in an in-line function we can call defer unlock.

```go
func() {
cs.destLock.Lock()
defer cs.destLock.Unlock()
for _, dst := range cs.dest {
  close(dst)
}
}()


```

pkg/controller/source/source.go
```diff
@@ -46,18 +52,113 @@ type Source interface {
...
+					for _, dst := range cs.dest {
+						close(dst)
+					}
+					cs.destLock.Unlock()
+					return
+				case evt := <-cs.source:
+					cs.destLock.Lock()
+					for _, dst := range cs.dest {
+						dst <- evt
```
pwittrock: Should this be synchronous or asynchronous?  With synchronous if one of the EventHandlers blocks, it will block all of them.  Is there any advantage to this?

pkg/controller/source/source.go
lichuqiang: good catch

pkg/controller/source/source.go
lichuqiang: I'd I haven't got  enough knowledge on how users will use it.
I think you have more say over this point, so let's make it the way you suggested :)

pkg/controller/source/source.go
lichuqiang: Graceful vs Easy to understand
I'm fine with both

pkg/controller/source/source.go
pwittrock: I guess this is still asynchronous because of the buffer.  Is the buffer a more common / better pattern?  Please leave a comment with the information for future readers.

pkg/controller/source/source.go
pwittrock: or 

```go
case <-cs.stop:
  cs.doStop()
case evt := <-cs.source:
  cs.DoEvent(evt)
```

pkg/runtime/inject/inject.go
```diff
@@ -82,3 +83,33 @@ func DoScheme(scheme *runtime.Scheme, i interface{}) (bool, error) {
...
+
+// SourceChannel is used by the ControllerManager to inject channel into Sources, EventHandlers, Predicates, and
+// Reconciles
+type SourceChannel interface {
+	InjectSourceChannel(chan event.GenericEvent) error
+}
+
+// DoSourceChannel will set source channel on i and return the result if it implements SourceChannel.
+// Returns false if i does not implement SourceChannel.
+func DoSourceChannel(ch chan event.GenericEvent, i interface{}) (bool, error) {
```
pwittrock: Lets no inject this.

pkg/runtime/inject/inject.go
```diff
@@ -82,3 +83,33 @@ func DoScheme(scheme *runtime.Scheme, i interface{}) (bool, error) {
...
+
+// Stop is used by the ControllerManager to inject stop channel into Sources, EventHandlers, Predicates, and
+// Reconciles
+type Stop interface {
+	InjectStop(<-chan struct{}) error
+}
+
+// DoStop will set stop channel on i and return the result if it implements Stop.
+// Returns false if i does not implement Stop.
+func DoStop(stop <-chan struct{}, i interface{}) (bool, error) {
```
pwittrock: Call this InjectStop.  (functions should have been renamed.)

pkg/controller/source/source.go
```diff
@@ -31,6 +32,11 @@ import (
 	logf "github.com/kubernetes-sigs/controller-runtime/pkg/runtime/log"
 )
 
+const (
+	// initialBufferSize is the initial number of event notifications that can be buffered.
+	initialBufferSize = 1024
```
pwittrock: lets make this a variable on the ChannelSource and write a test for what happens when we run out of buffer (e.g. we can set the buffer size to 1)

pkg/controller/source/source_test.go
```diff
@@ -203,4 +204,57 @@ var _ = Describe("Source", func() {
 			})
 		})
 	})
+	Describe("ChannelSource", func() {
```
pwittrock: Add a test for when the ChannelSource is reused by multiple watches with different EventHandlers.

pkg/controller/source/source.go
lichuqiang: that makes sense

pkg/controller/source/source_test.go
```diff
@@ -203,4 +204,57 @@ var _ = Describe("Source", func() {
...
+		var c chan struct{}
+		var ch chan event.GenericEvent
+
+		BeforeEach(func() {
```
pwittrock: Add tests for error conditions where the stop or channel are not specified.

pkg/controller/source/source_test.go
```diff
@@ -26,9 +26,10 @@ import (
 	"github.com/kubernetes-sigs/controller-runtime/pkg/runtime/inject"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"k8s.io/client-go/util/workqueue"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
pwittrock: Lets also add a controller_integration_test.go case that exercises the end-to-end flow.

pwittrock:
This looks good.  A few minor comments.

pkg/controller/source/source.go
lichuqiang: yeah, I think the buffer should have been big enough normally, and if we still see it blocked, the controller is thought to be in an abnormal state.
Will add comment for clarification.

pkg/controller/source/source.go
lichuqiang: So when do you think it should be specified?
Or we'll in fact always use the default value, and the field will only be changed in test?

pkg/controller/source/source.go
```diff
@@ -45,21 +51,129 @@ type Source interface {
...
+	Source <-chan event.GenericEvent
+
+	// stop is to end ongoing goroutine, and close the channels
+	stop <-chan struct{}
+
+	// dest is the destination channels of the added event handlers
+	dest []chan event.GenericEvent
+
+	// DestBufferSize is the specified buffer size of dest channels
+	DestBufferSize int
```
pwittrock: followup: Include the default value in the comment.

pkg/controller/source/source_test.go
```diff
@@ -251,11 +253,194 @@ var _ = Describe("Source", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				ch <- evt
+				<-c
+				close(done)
+			})
+			It("should block if exceed buffer size", func(done Done) {
+				ch := make(chan event.GenericEvent)
+				evt := event.GenericEvent{}
+				interval := 5 * time.Second
```
pwittrock: This will make the test run a really long time.  Instead of sleeping, consider waiting on a signal / channel.

pkg/controller/source/source_test.go
```diff
@@ -251,11 +253,194 @@ var _ = Describe("Source", func() {
...
+						defer GinkgoRecover()
+						Fail("Unexpected UpdateEvent")
+					},
+					DeleteFunc: func(workqueue.RateLimitingInterface, event.DeleteEvent) {
+						defer GinkgoRecover()
+						Fail("Unexpected DeleteEvent")
+					},
+					GenericFunc: func(q2 workqueue.RateLimitingInterface, evt event.GenericEvent) {
+						defer GinkgoRecover()
+						time.Sleep(interval)
```
pwittrock: Don't sleep, wait on a channel or something

pkg/controller/source/source_test.go
```diff
@@ -251,11 +253,194 @@ var _ = Describe("Source", func() {
...
+				ch <- evt
+				ch <- evt
+
+				beforeEvent := time.Now()
+				ch <- evt
+				// validate event distribution get blocked.
+				distributeInterval := time.Now().Sub(beforeEvent)
+				Expect(distributeInterval >= interval).To(BeTrue())
+				close(done)
+			}, 15)
```
pwittrock: keep the timeout at 1 second

pkg/controller/source/source_test.go
```diff
@@ -251,11 +253,194 @@ var _ = Describe("Source", func() {
...
+						close(c)
+					},
+				}, q)
+				Expect(err).NotTo(HaveOccurred())
+
+				ch <- evt
+				<-c
+				close(done)
+			})
+			It("should block if exceed buffer size", func(done Done) {
```
pwittrock: I think we care less that it blocks, and more that it doesn't blow up.  I think the main thing we want to test is that we can write 3 events to the source channel while the eventhandler is blocked, then unblock the eventhandler, and all the 3 events will get processed.

pwittrock:
lgtm once the test run time is reduce back down to sub second

pkg/controller/source/source.go
```diff
@@ -45,21 +51,129 @@ type Source interface {
...
+
+	// destLock is to ensure the destination channels are safely added/removed
+	destLock sync.Mutex
+}
+
+var _ inject.Stop = &ChannelSource{}
+
+// InjectStop is internal should be called only by the Controller.
+// It is used to inject the stop channel initialized by the ControllerManager.
+func (cs *ChannelSource) InjectStop(stop <-chan struct{}) error {
```
droot: Better name it explicit `InjectStopChannel` ?

pkg/runtime/inject/inject.go
```diff
@@ -82,3 +82,18 @@ func SchemeInto(scheme *runtime.Scheme, i interface{}) (bool, error) {
 	}
 	return false, nil
 }
+
+// Stop is used by the ControllerManager to inject stop channel into Sources, EventHandlers, Predicates, and
+// Reconciles
+type Stop interface {
```
droot: `Stoppable` ?

pkg/controller/source/source.go
```diff
@@ -45,21 +51,129 @@ type Source interface {
 	Start(eventhandler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
+var _ Source = &ChannelSource{}
+
 // ChannelSource is used to provide a source of events originating outside the cluster
```
droot: I am trying to understand the correct usage pattern for the channel-source so that we can advise users accordingly.

For external triggers which maps to changes in a `type` in K8s, controller will automatically pick them up because it is watching that type.
Now left are the ones which don't result in any data change in the type directly but impact the type, channel source is the right fit for those.

We have to come up with a better example demonstrating this source.

PR #23, opened 2018-06-12
Implement InjectRecorder

lichuqiang:
Address: https://github.com/kubernetes-sigs/kubebuilder/issues/255

more tests are to be added

### DIFF
--- a/pkg/internal/recorder/recorder_integration_test.go
+++ b/pkg/internal/recorder/recorder_integration_test.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package recorder_test
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/scheme"
+	ref "k8s.io/client-go/tools/reference"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("recorder", func() {
+	var stop chan struct{}
+
+	BeforeEach(func() {
+		stop = make(chan struct{})
+		Expect(cfg).NotTo(BeNil())
+	})
+
+	AfterEach(func() {
+		close(stop)
+	})
+
+	Describe("recorder", func() {
+		It("should publish events", func(done Done) {
+			By("Creating the Manager")
+			cm, err := manager.New(cfg, manager.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Creating the Controller")
+			recorder := cm.GetRecorder("test-recorder")
+			instance, err := controller.New("foo-controller", cm, controller.Options{
+				Reconcile: reconcile.Func(
+					func(request reconcile.Request) (reconcile.Result, error) {
+						dp, err := clientset.AppsV1().Deployments(request.Namespace).Get(request.Name, metav1.GetOptions{})
+						Expect(err).NotTo(HaveOccurred())
+						recorder.Event(dp, corev1.EventTypeNormal, "test-reason", "test-msg")
+						return reconcile.Result{}, nil
+					}),
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Watching Resources")
+			err = instance.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.Enqueue{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Starting the Manager")
+			go func() {
+				defer GinkgoRecover()
+				Expect(cm.Start(stop)).NotTo(HaveOccurred())
+			}()
+
+			deployment := &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "deployment-name"},
+				Spec: appsv1.DeploymentSpec{
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{"foo": "bar"},
+					},
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"foo": "bar"}},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{
+								{
+									Name:  "nginx",
+									Image: "nginx",
+								},
+							},
+						},
+					},
+				},
+			}
+
+			By("Invoking Reconciling")
+			deployment, err = clientset.AppsV1().Deployments("default").Create(deployment)
+			Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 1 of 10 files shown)


pkg/recorder/recorder.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/golang/glog"
+	"github.com/kubernetes-sigs/controller-runtime/pkg/client/config"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+type RecorderGetFuncType func(string, *runtime.Scheme) record.EventRecorder
```
pwittrock: I don't think the interface should take the Scheme.  This should be set by the manager.Manager so users don't need to do the wiring themselves.

pwittrock: If this is an interface instead of a function, we can provide the interface.  Maybe `RecoderProvider`?

pkg/recorder/recorder.go
```diff
@@ -0,0 +1,41 @@
...
+
+var eventBroadcaster record.EventBroadcaster
+
+func GetEventRecorderFor(name string, scheme *runtime.Scheme) record.EventRecorder {
+	if eventBroadcaster == nil {
+		cs := config.GetKubernetesClientSetOrDie()
```
pwittrock: We should get the clientset from the manager.Manager, or have users set it on a struct.  All components should use the same client.

pkg/manager/manager.go
```diff
@@ -131,5 +133,9 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	cm.fieldIndexes = cm.cache
 	cm.client = client.DelegatingClient{ReadInterface: cm.cache, WriteInterface: writeObj}
+
+	cm.recorderProvider = recorder.NewProvider()
+	cm.recorderProvider.SetScheme(cm.scheme)
+	cm.recorderProvider.SetClientSet(clientconfig.GetKubernetesClientSetOrDieWithConfig(config))
```
pwittrock: We shouldn't be dying.  This function returns an error.

pkg/recorder/recorder.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type provider struct {
+	clientSet        *kubernetes.Clientset
+	scheme           *runtime.Scheme
+	eventBroadcaster record.EventBroadcaster
+}
+
+// NewProvider create a new Provider instance.
+func NewProvider() Provider {
```
pwittrock: The `New` pattern is useful when step requires synchronous steps, or setup can fail.  Since we don't have either in this case, I am not sure how much value this provides.

pkg/recorder/recorder.go
```diff
@@ -0,0 +1,78 @@
...
+		p.scheme = scheme
+	}
+}
+
+func (p *provider) GetEventRecorderFor(name string) (record.EventRecorder, error) {
+	if p.scheme == nil {
+		return nil, fmt.Errorf("must call SetScheme to specify Scheme before getting recorder")
+	}
+
+	if p.eventBroadcaster == nil {
```
pwittrock: This need to be threadsafe.  Either wrap it in a `sync.Once.Do` call, or put it in "New"

pkg/manager/manager.go
```diff
@@ -131,5 +133,9 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	cm.fieldIndexes = cm.cache
 	cm.client = client.DelegatingClient{ReadInterface: cm.cache, WriteInterface: writeObj}
+
+	cm.recorderProvider = recorder.NewProvider()
```
pwittrock: Add a function to manager.go for calling the provider - e.g. add this to the Manager interface and implement it on manager.

```go
GetRecorder(name string) (record.EventRecorder, error)
```

pkg/recorder/recorder.go
pwittrock: We should either take the deps in the New function *or* inject them with Inject functions.  provider doesn't implement either inject.Scheme or inject.Config so it can't automatically get its deps from manager.SetFields(provider).  If we are making the implementation internal, it doesn't really matter which we choose since we can change it without breaking consumers.  `New` is probably the simpler approach for now, so I'd go with that.

e.g. either

```go
// setup clientset and stuff here
New(config *rest.Config, s scheme *runtime.Scheme) { ... }
```
or

```go
// Setup clientset and stuff here
(p *provider) InjectClient() {...}
(p *provider) InjectScheme() {...}
```

pwittrock: Lets move  `New` under `pkg/internal` until we are sure we know users will need to get informers directly.  They should be able to get them through the manager.Manager.

pkg/runtime/inject/inject.go
```diff
@@ -98,6 +99,21 @@ func StopChannelInto(stop <-chan struct{}, i interface{}) (bool, error) {
 	return false, nil
 }
 
+// Recorder is used by the ControllerManager to inject recorder into Sources, EventHandlers, Predicates, and
+// Reconciles
+type Recorder interface {
+	InjectRecorder(provider recorder.Provider) error
```
pwittrock: Document this.

pkg/recorder/recorder_test.go
```diff
@@ -0,0 +1,57 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package recorder_test
+
+import (
+	"fmt"
```
pwittrock: Also update the pkg/controller/controller_integration_test.go with a case that uses an EventRecorder and creates and event from the Reconcile loop, then the test checks that the event was recorded.

pwittrock:
Headed in the right direction.  Left a few comments on how we want to structure things.  Please run `test.sh` and address the lint errors.

pkg/recorder/recorder.go
lichuqiang: Yah, I think we do inject for things that cross components, and here we just make it once, so a New function should be enough.

pkg/recorder/recorder.go
lichuqiang: good catch

pkg/manager/manager.go
lichuqiang: what the function for? you mean we should not take the provider as a field of manager, and pass it into the recorder injection?

pkg/recorder/recorder_test.go
lichuqiang: sure thing, just have not got time to make it today

lichuqiang:
have not inject recorder into the components, and want to make clear its boundary:

1. Whether we really need recorder in all of the components: Reconcilers, Sources, EventHandlers, and Predicates...
2. Whether we need recorder in all the instances of certain component, or simply that of certain kind.
eg: whether we should implement `InjectRecorder` in the source interface layer, or just implement it for Channel/Kind source.

pkg/internal/recorder/recorder_test.go
```diff
@@ -0,0 +1,51 @@
...
+			_, err := recorder.NewProvider(&cfg1, scheme.Scheme)
+			Expect(err.Error()).To(ContainSubstring("failed to init clientSet"))
+		})
+	})
+	Describe("GetEventRecorder", func() {
+		It("should return a recorder instance.", func() {
+			provider, err := recorder.NewProvider(cfg, scheme.Scheme)
+			Expect(err).NotTo(HaveOccurred())
+
+			recorder := provider.GetEventRecorderFor("test")
```
pwittrock: Can you also test that the event recorder will publish events?

lichuqiang: done

pkg/internal/recorder/recorder_integration_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			By("Invoking Reconciling")
+			deployment, err = clientset.AppsV1().Deployments("default").Create(deployment)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Validate event is published as expected")
+			evtWatcher, err := clientset.CoreV1().Events("default").Watch(metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			resultEvent := <-evtWatcher.ResultChan()
```
pwittrock: This is an awesome way to both test the watching of these events and make the test run faster without polling!

pkg/manager/internal.go
```diff
@@ -122,6 +128,10 @@ func (cm *controllerManager) GetCache() cache.Cache {
 	return cm.cache
 }
 
+func (cm *controllerManager) GetRecorder(name string) record.EventRecorder {
```
pwittrock: Please add a comment for this function

PR #26, opened 2018-06-12
Add simple application pattern package

pwittrock:
Looking to get feedback on simplifying this pattern.

### DIFF
--- a/pkg/patterns/application/application.go
+++ b/pkg/patterns/application/application.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package application
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// application is a simple Controller for a single API type.  It will create a Manager for itself
+// if one is not provided.
+type application struct {
+	mgr  manager.Manager
+	ctrl controller.Controller
+}
+
+// Supporting mocking out functions for testing
+var getConfig = config.GetConfig
+var newController = controller.New
+var newManager = manager.New
+var getGvk = apiutil.GVKForObject
+
+// Builder builds an Application Controller (e.g. Operator) and returns a manager.Manager to start it.
+type Builder struct {
+	apiType        runtime.Object
+	mgr            manager.Manager
+	predicates     []predicate.Predicate
+	managedObjects []runtime.Object
+	config         *rest.Config
+	reconcile      reconcile.Reconciler
+	ctrl           controller.Controller
+}
+
+// BuilderFor returns a new Builder for and BuilderFor
+func BuilderFor(apiType runtime.Object) *Builder {
+	return &Builder{apiType: apiType}
+}
+
+// Owns configures the Application Controller to respond to create / delete / update events for objects it managedObjects
+// - e.g. creates.  apiType is an empty instance of an object matching the managed object type.
+func (b *Builder) Owns(apiType runtime.Object) *Builder {
+	b.managedObjects = append(b.managedObjects, apiType)
+	return b
+}
+
+// WithConfig sets the Config to use for configuring clients.  Defaults to the in-cluster config or to ~/.kube/config.
+func (b *Builder) WithConfig(config *rest.Config) *Builder {
+	b.config = config
+	return b
+}
+
+// WithManager sets the Manager to use for registering the Controller.  Defaults to a new manager.Manager.
+func (b *Builder) WithManager(m manager.Manager) *Builder {
+	b.mgr = m
+	return b
+}
+
+// WithEventFilter sets the event filters, to filter which create/update/delete/generic events eventually
+// trigger reconciliations.  For example, filtering on whether the resource version has changed.
+// Defaults to the empty list.
+func (b *Builder) WithEventFilter(p predicate.Predicate) *Builder {
+	b.predicates = append(b.predicates, p)
+	return b
+}
+
+// WithReconciler sets the Reconcile called in response to create / update / delete events for the
+// BuilderFor type or Created object types.
+func (b *Builder) WithReconciler(r reconcile.Reconciler) *Builder {
+	b.reconcile = r
+	return b
+}
+
+// Build builds the Application Controller and returns the Manager used to start it.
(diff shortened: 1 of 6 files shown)


pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	internalcontroller "sigs.k8s.io/controller-runtime/pkg/internal/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// application is a very simple Controller that embeds a Manager.
+type application struct {
+	mrg    manager.Manager
```
DirectXMan12: `mgr`?

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	stopFn func() <-chan struct{}
+	ctrl   *internalcontroller.Controller
+}
+
+var noop = reconcile.Func(func(reconcile.Request) (reconcile.Result, error) {
```
DirectXMan12: docs

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	internalcontroller "sigs.k8s.io/controller-runtime/pkg/internal/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// application is a very simple Controller that embeds a Manager.
```
DirectXMan12: write why in the docs.

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	mrg    manager.Manager
+	stopFn func() <-chan struct{}
+	ctrl   *internalcontroller.Controller
+}
+
+var noop = reconcile.Func(func(reconcile.Request) (reconcile.Result, error) {
+	return reconcile.Result{}, nil
+})
+
+var getConfig = config.GetConfig
```
DirectXMan12: these for testing purposes?  Write that in a comment.

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	return b
+}
+
+// WithManager sets the Manager to use for registering the Controller.  Defaults to a new manager.Manager.
+func (b *Builder) WithManager(m manager.Manager) *Builder {
+	b.mrg = m
+	return b
+}
+
+// WithEventFilter sets the event filters.  Defaults to the empty list.
```
DirectXMan12: `... the event filters, to filter which create/update/delete/generic events eventually trigger reconciliations`

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+func (b *Builder) WithEventFilter(p predicate.Predicate) *Builder {
+	b.pred = append(b.pred, p)
+	return b
+}
+
+// WithHandler sets th Reconcile called in response to create / update / delete events for the
```
DirectXMan12: `sets the`

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+// BuilderFor type or Created object types.
+func (b *Builder) WithHandler(r reconcile.Reconciler) *Builder {
```
DirectXMan12: maybe `WithReconciler` so that we're using the same terminology?

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+	b.r = r
```
DirectXMan12: give this a proper name please.  You have autocomplete, so you have no excuse :-P

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, err
+	}
+
+	// Create the controller
+	name := fmt.Sprintf("%s-application", strings.ToLower(gvk.Kind))
+	ctrl, err := newController(name, a.mrg, controller.Options{Reconciler: b.r})
+	if err != nil {
+		return nil, err
+	}
+	a.ctrl = ctrl.(*internalcontroller.Controller)
```
DirectXMan12: eeewww.  why is this cast necessary?

pkg/patterns/application/application.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Reconcile type
+	err = a.ctrl.Watch(&source.Kind{Type: b.apiType}, &handler.EnqueueRequestForObject{}, b.pred...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Watch the managed types
+	for _, t := range b.manages {
+		err = a.ctrl.Watch(&source.Kind{Type: t}, &handler.EnqueueRequestForOwner{
```
DirectXMan12: is there/should there be a way to add non-owns relationships with this builder?

pkg/patterns/application/application.go
DirectXMan12: `for example, filtering by a particular label selector or annotation`

pkg/patterns/application/doc.go
```diff
@@ -0,0 +1,29 @@
...
+// Note: Application is an alpha library and make have backwards compatibility breaking changes.
+//
+// Projects built with an application can trivially be rebased on top of the underlying Controller and Manager
+// packages if the project requires more customized behavior in the future.
+//
+// Application
+//
+// An application is a Controller that implements the operational logic for an application.  It is often used
+// to take off-the-shelf OSS applications, and make them Kubernetes native.
+package application
```
DirectXMan12: if it's for operators, maybe just call it `operator`?

pkg/patterns/application/example_test.go
```diff
@@ -0,0 +1,96 @@
...
+	rs := &appsv1.ReplicaSet{}
+	err := a.Get(context.TODO(), req.NamespacedName, rs)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// List the Pods matching the PodTemplate Labels
+	pods := &corev1.PodList{}
+	err = a.List(context.TODO(),
+		&client.ListOptions{
```
DirectXMan12: I believe we have a builder for this (if we don't have one for the namespace, we should), so you should be able to write something like

`client.InNamespace(req.Namespace).WithLabels(rs.Spec.Template.Labels)`, which is a bit easier to read and doesn't require pulling out `labels.SelectorFromSet`.

pkg/patterns/application/application.go
pwittrock: Deleted

pkg/patterns/application/application.go
pwittrock: done

pkg/patterns/application/application.go
pwittrock: Done

pkg/patterns/application/application.go
pwittrock: Done

pkg/patterns/application/application.go
pwittrock: Its not

pkg/patterns/application/application.go
pwittrock: Not right now.  The point of this package is to offer a simplified view for the most common way to build application controllers.  If arbitrary mappings become common for operators, then we can add them.  Otherwise users can break out and use the underlying libraries directly.

pkg/patterns/application/doc.go
pwittrock: Calling this Application because that is the common response when folks as "what is an operator".  I created an operator package to explain this.

pkg/patterns/application/example_test.go
pwittrock: Done

pkg/patterns/application/application.go
pwittrock: Done

pkg/patterns/application/application.go
pwittrock: Dione

pkg/patterns/application/application.go
pwittrock: Done

pwittrock:
@DirectXMan12 LMK if there is anything else.

PR #29, opened 2018-06-13
Refactor cache package.

pwittrock:
Follow up with tests for the cache and cache/internal packages.


pwittrock:
@seans3 This is the other half of the client PR.  Once the structure looks good, we should add thorough testing.

pkg/cache/internal/factory.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Get will create a new Informer and added it to the map of InformersMap if none exists.  Returns
+// the Informer from the map.
+func (ip *InformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Object) (*MapEntry, error) {
+	// Do the mutex part in its own function so we can use defer without blocking pieces that don't
+	// to be locked
+	var sync bool
+	i, err := func() (*MapEntry, error) {
+		ip.mu.Lock()
```
DirectXMan12: This is going to cause issues, because it means that only one read operation can happen across any index at once, which is suboptimal. That's one of the reasons why the separation between creation and client lookup was useful.  I think what we might want to do is

```
// assume the happy path first
with read-only-lock {
  if i, ok := informers[gvk]; ok {
    return i
  }
}

// sad path, first time only
with write-lock {
    // check again in case we raced
    if i, ok := informers[gvk]; ok {
    return i
  }

  // only construct if we actually need to do so
  i := construct informer
  informers[gvk] = i
  return i
}
```

pwittrock: Done

pkg/cache/cache.go
```diff
@@ -17,333 +17,97 @@ limitations under the License.
...
-			lbls := labels.Set(meta.GetLabels())
-			if !labelSel.Matches(lbls) {
-				continue
-			}
-		}
-		outItems = append(outItems, obj.DeepCopyObject())
+// New initializes and returns a new Cache
+func New(config *rest.Config, opts Options) (Cache, error) {
+	// Use the default Kubernetes Scheme if unset
+	if opts.Scheme == nil {
```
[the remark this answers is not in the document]
pwittrock: done

pkg/cache/internal/factory.go
```diff
@@ -0,0 +1,212 @@
...
+			return nil, err
+		}
+		ni := cache.NewSharedIndexInformer(lw, obj, ip.resync, cache.Indexers{})
+		i = &MapEntry{
+			Informer: ni,
+			Reader:   Reader{indexer: ni.GetIndexer(), groupVersionKind: gvk}}
+		ip.informersByGVK[gvk] = i
+
+		// Start the Informer if need by
+		if ip.started {
```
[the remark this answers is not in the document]
pwittrock: Added a todo about adding tests and documentation for this.

pkg/cache/cache.go
```diff
@@ -17,333 +17,97 @@ limitations under the License.
...
+	WaitForCacheSync(stop <-chan struct{}) bool
 
-// singleObjectCache is a ReadInterface that retrieves objects
-// from a single local cache populated by a watch.
-type singleObjectCache struct {
-	// Indexer is the underlying indexer wrapped by this cache.
-	Indexer cache.Indexer
-	// GroupVersionKind is the group-version-kind of the resource.
-	GroupVersionKind schema.GroupVersionKind
+	// IndexField adds an index to a field.
```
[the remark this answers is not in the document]
pwittrock: Added a go doc from client-go.  LMK if this is the right one.

pwittrock:
PTAL

pkg/cache/cache.go
```diff
@@ -17,333 +17,110 @@ limitations under the License.
...
+var _ Cache = &informerCache{}
 
-// singleObjectCache is a Reader that retrieves objects
-// from a single local cache populated by a watch.
-type singleObjectCache struct {
-	// Indexer is the underlying indexer wrapped by this cache.
-	Indexer cache.Indexer
-	// GroupVersionKind is the group-version-kind of the resource.
-	GroupVersionKind schema.GroupVersionKind
+// cache is a Kubernetes Object cache populated from InformersMap.  cache wraps a CacheProvider and InformersMap.
```
seans3: cache -> informerCache

seans3: It looks like informerCache only wraps an InformersMap, not a CacheProvider as stated in comment.

pkg/cache/cache.go
```diff
@@ -17,333 +17,110 @@ limitations under the License.
...
-// objectKeyToStorageKey converts an object key to store key.
-// It's akin to MetaNamespaceKeyFunc.  It's separate from
-// String to allow keeping the key format easily in sync with
-// MetaNamespaceKeyFunc.
-func objectKeyToStoreKey(k client.ObjectKey) string {
-	if k.Namespace == "" {
-		return k.Name
+	// Default the resync period to 10 hours if unset
+	if opts.Resync == nil {
+		r := 10 * time.Hour
```
seans3: Consider pulling 10 out into a const as "defaultResyncHours"

pkg/cache/internal/factory.go
```diff
@@ -0,0 +1,227 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+// NewInformersMap returns a new InformersMap
+func NewInformersMap(config *rest.Config,
```
seans3: If this file is mostly about the InformersMap, why is it called factory.go ?

pkg/cache/internal/factory.go
```diff
@@ -0,0 +1,227 @@
...
+// WaitForCacheSync waits until all the caches have been synced
+func (ip *InformersMap) WaitForCacheSync(stop <-chan struct{}) bool {
+	syncedFuncs := make([]cache.InformerSynced, 0, len(ip.informersByGVK))
+	for _, informer := range ip.informersByGVK {
+		syncedFuncs = append(syncedFuncs, informer.Informer.HasSynced)
+	}
+	return cache.WaitForCacheSync(stop, syncedFuncs...)
+}
+
+// Get will create a new Informer and added it to the map of InformersMap if none exists.  Returns
```
seans3: added -> add

pkg/cache/internal/factory.go
```diff
@@ -0,0 +1,227 @@
...
+		defer ip.mu.RUnlock()
+		i, ok := ip.informersByGVK[gvk]
+		return i, ok
+	}()
+	if ok {
+		return i, nil
+	}
+
+	// Do the mutex part in its own function so we can use defer without blocking pieces that don't
+	// to be locked
```
seans3: nit: missing "need" to be locked.

pkg/cache/internal/typed_reader.go
```diff
@@ -0,0 +1,186 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/client-go/tools/cache"
+)
+
+// Reader is a Reader
+var _ client.Reader = &Reader{}
+
+// Reader wraps a cache.Index to implement the client.Reader interface for a single type
+type Reader struct {
```
seans3: Reader struct might be confused with Reader interface. You might want to consider another name.

pkg/cache/reader.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/kubernetes-sigs/controller-runtime/pkg/client"
+	"github.com/kubernetes-sigs/controller-runtime/pkg/client/apiutil"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var _ client.Reader = &informerCache{}
+
+// Get implements Reader
+func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runtime.Object) error {
```
seans3: Even though the informerCache implements the client.Reader interface, would this file more accurately be called informer_cache.go instead of reader.go ?

pkg/cache/cache.go
pwittrock: Done

pwittrock: Done

pkg/cache/internal/factory.go
pwittrock: Done

pkg/cache/internal/factory.go
pwittrock: Done

pkg/cache/cache.go
pwittrock: Done

pkg/cache/internal/factory.go
pwittrock: Done

pkg/cache/internal/typed_reader.go
pwittrock: Done

pkg/cache/reader.go
pwittrock: Done

pwittrock:
Comments addressed in #35

PR #34, opened 2018-06-13
Add runtime test

fanzhangio:
- [x] inject
- [x] log
- [x] signals

### DIFF
--- a/pkg/runtime/inject/inject_suite_test.go
+++ b/pkg/runtime/inject/inject_suite_test.go
@@ -14,12 +14,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package manager
+package inject
 
-// func SetCacheForTest(options *Options, c func(config *rest.Config, opts cache.Options) (cache.Cache, error)) {
-// 	options.newCache = c
-// }
+import (
+	"testing"
 
-// func SetClientForTest(options *Options, c func(config *rest.Config, options client.Options) (client.Client, error)) {
-// 	options.newClient = c
-// }
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/test"
+)
+
+func TestSource(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "Runtime Injection Suite", []Reporter{test.NewlineReporter{}})
+}
--- a/pkg/runtime/inject/inject_test.go
+++ b/pkg/runtime/inject/inject_test.go
@@ -1,0 +1,289 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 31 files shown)


pwittrock:
@fanzhangio please run `test.sh` and address the failures.

pkg/runtime/inject/inject_suite_test.go
```diff
@@ -0,0 +1,57 @@
...
+	RunSpecsWithDefaultAndCustomReporters(t, "Runtime Injection Suite", []Reporter{test.NewlineReporter{}})
+}
+
+var testenv *test.Environment
+var cfg *rest.Config
+var clientset *kubernetes.Clientset
+
+var _ = BeforeSuite(func(done Done) {
+
+	testenv = &test.Environment{}
```
pwittrock: This is expensive to bring up just to get a* rest.Config that isn't used for anything.

Instead just create a dummy config with some values filled in.

fanzhangio:
@pwittrock  I simplified the test suites, and removed the redundant codes.

pkg/runtime/inject/inject_test.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+var instance testSource
+var neg struct{}
+var _ = Describe("runtime inject", func() {
+
+	It("should set informers", func() {
+		instance := testSource{
+			cache: func(c cache.Cache) error {
+				if c != nil {
```
pwittrock: We should validate that the injected argument is the same as the one you provided.

pkg/runtime/inject/inject_test.go
```diff
@@ -0,0 +1,193 @@
...
+	client func(client.Client) error
+	f      func(Func) error
+	stop   func(<-chan struct{}) error
+}
+
+func (s testSource) InjectCache(c cache.Cache) error {
+	if s.cache != nil {
+		return s.cache(c)
+	}
+	return nil
```
pwittrock: This should probably `Fail("unexpected call")` or something right?

pkg/runtime/inject/inject_test.go
```diff
@@ -0,0 +1,193 @@
...
+				if c != nil {
+					return nil
+				}
+				return fmt.Errorf("")
+			},
+		}
+		res, err := CacheInto(&informertest.FakeInformers{}, instance)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res).To(Equal(true))
+		res, err = CacheInto(&informertest.FakeInformers{}, neg)
```
pwittrock: What is this testing?

pkg/runtime/inject/inject_test.go
```diff
@@ -0,0 +1,193 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res).To(Equal(false))
```
pwittrock: We should check an error case for when an error is returned.

pkg/runtime/signals/signal_test.go
```diff
@@ -0,0 +1,42 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("runtime signal", func() {
+
+	Context("SignalHandler Test", func() {
+		It("should close stop channel by signal", func() {
+
+			stop := SetupSignalHandler()
+			p, err := os.FindProcess(syscall.Getpid())
```
pwittrock: This is cool, but the handlers aren't setup right it might doing something weird.  It also changes the signal handler for the whole process.

How about

1.  write a small program and exec it
2. mock out the `signal.Notify` and `os.Exit` calls as functions that we re-assign.

pwittrock:
@fanzhangio just did another pass.  ptal

pwittrock:
@fanzhangio FYI, you'll have to do a small rebase with the package name changes.

fanzhangio:
Needs to run `dep ensure` after the package rebased.
Updated the vendor and deps

PR #37, opened 2018-06-14
Ensure that informers have the namespace index

DirectXMan12:
Most of the lister logic (both here and in Kubernetes generate listers)
assume that the namespace index is present by default, so we should
always set it up.


DirectXMan12:
@pwittrock when we start testing the List operation, this'll cause issues if not in place.

PR #43, opened 2018-06-15
Client tests

seans3:
* First pass at tests for client.go
* Currently 79.6% coverage


pwittrock:
@seans3 error should be legit now.  Looks like a linting error.

pkg/client/client_test.go
```diff
@@ -51,10 +57,24 @@ var _ = Describe("Client", func() {
...
 			Spec:       corev1.NodeSpec{},
 		}
+		// Add Deployment and Node to the Scheme.
+		scheme = runtime.NewScheme()
+		dgv := schema.GroupVersion{Group: "apps", Version: "v1"}
+		ngv := schema.GroupVersion{Group: "", Version: "v1"}
+		dgvk := dgv.WithKind("Deployment")
+		ngvk := ngv.WithKind("Node")
+		scheme.AddKnownTypeWithName(dgvk, dep)
+		scheme.AddKnownTypeWithName(ngvk, node)
```
droot: `client` users should not be required to register core types, by default, `client` should use scheme from the core.

pkg/client/client_test.go
```diff
@@ -134,16 +169,25 @@ var _ = Describe("Client", func() {
 			close(done)
 		})
 
-		It("should create a new object from an unstructured object", func() {
+		It("should create a new object from an unstructured object", func(done Done) {
```
droot: Most of the time, our users are going to be using the core/custom types but unstructured. So not sure if unstructured test really helps ?

pkg/client/client_test.go
```diff
@@ -192,16 +232,34 @@ var _ = Describe("Client", func() {
...
+			Expect(err).To(HaveOccurred())
+
+			close(done)
 		})
 
 		It("should fail if the object cannot be mapped to a GVK", func() {
-			// TODO(seans3): implement these
+			By("creating client with empty Scheme")
+			emptyScheme := runtime.NewScheme()
+			cl, err := client.New(cfg, client.Options{Scheme: emptyScheme})
```
droot: We can make it more closer to the actual user's usecase where: we create the `cl` without specifying the scheme so that client is initialized with default scheme which understands the core-type resources and then use a custom struct (we can define one in this test file) for Create method and it should fail.

pkg/client/client_test.go
```diff
@@ -211,24 +269,47 @@ var _ = Describe("Client", func() {
...
-			By("updating the object")
+			By("updating the Deployment")
 			dep.Annotations = map[string]string{"foo": "bar"}
 			err = cl.Update(context.TODO(), dep)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("writing the result back to the go struct")
+			By("validating updated Deployment exists")
 			actual, err := clientset.AppsV1().Deployments(ns).Get(dep.Name, metav1.GetOptions{})
 			Expect(err).NotTo(HaveOccurred())
```
droot: We aren't validating if the object got updated here.

pkg/client/client_test.go
```diff
@@ -211,24 +269,47 @@ var _ = Describe("Client", func() {
...
 			Expect(actual).NotTo(BeNil())
 
 			close(done)
 		})
 
-		It("should update an existing new object from an unstructured object", func() {
-			// TODO(seans3): implement these
+		It("should update an existing new object from an unstructured object", func(done Done) {
```
droot: As I said earlier, not sure if `unstructured` tests help our usecase ?

pkg/client/client_test.go
```diff
@@ -211,24 +269,47 @@ var _ = Describe("Client", func() {
...
+			dep.Annotations = map[string]string{"foo": "bar"}
+			scheme.Convert(dep, u, nil)
+
+			By("updating the Deployment")
+			err = cl.Update(context.TODO(), u)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("fetching newly created unstructured Deployment")
+			actual, err := clientset.AppsV1().Deployments(ns).Get(dep.Name, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
```
droot: We aren't verifying the update actually happened.

pkg/client/client_test.go
```diff
@@ -269,8 +353,24 @@ var _ = Describe("Client", func() {
 
 		})
 
-		It("should fail if the object cannot be mapped to a GVK", func() {
+		It("should fail if the object cannot be mapped to a GVK", func(done Done) {
+			By("creating client with empty Scheme")
+			emptyScheme := runtime.NewScheme()
+			cl, err := client.New(cfg, client.Options{Scheme: emptyScheme})
```
droot: same comment as earlier about not specifying the scheme.

pkg/client/client_test.go
pwittrock: Its good to check that the provided scheme is actually used though :)

pkg/client/client_test.go
pwittrock: Doesn't this line do that:

> Expect(actual.Annotations["foo"]).To(Equal("bar"))

pkg/client/client_test.go
pwittrock: I think it makes it easier to check for things not in the Scheme if an empty one is passed in

pkg/client/client_test.go
```diff
@@ -192,16 +234,34 @@ var _ = Describe("Client", func() {
...
-			// TODO(seans3): implement these
-		})
+		It("should fail if the object does not pass server-side validation", func(done Done) {
+			cl, err := client.New(cfg, client.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cl).NotTo(BeNil())
+
+			By("creating the pod, since required field Containers is empty")
+			err = cl.Create(context.TODO(), pod)
+			Expect(err).To(HaveOccurred())
```
pwittrock: Would you also check the error message to make sure it is actually the expected message and not something unrelated?

pkg/client/client_test.go
seans3: Phil asked to do the unstructured, which gets slightly better code coverage. So let's go with it for now.

pkg/client/client_test.go
seans3: Phil asked to do the unstructured, which gets slightly better code coverage. So let's go with it for now.

pkg/client/client_test.go
seans3: I just pushed this commit to address Sunil's concern.

pkg/client/client_test.go
seans3: For this one, I'll keep the empty scheme.

pkg/client/client_test.go
seans3: Done

pkg/client/client_test.go
seans3: I'll add a TODO for this. I don't want this PR to take much longer, so I'll address further checks in a future PR.

pkg/client/client_test.go
seans3: Done.

PR #45, opened 2018-06-17
Support SetReconcile on Controller for testing

pwittrock:


pkg/controller/controller.go
```diff
@@ -52,6 +52,9 @@ type Controller interface {
 
 	// Start starts the controller.  Start blocks until stop is closed or a controller has an error starting.
 	Start(stop <-chan struct{}) error
+
+	// SetReconcile sets the Reconcile function on a Controller
+	SetReconcile(reconcile.Reconcile)
```
droot: Was wondering if we really need this ? because one can use channel-source and register their reconciler to trigger arbitrary reconciles ?

pwittrock: This is mainly to make testing easier to stub out the Reconcile method and make sure the watches are setup correctly.

lichuqiang: Any examples for `make testing easier?`, you mean replace the reconcile specification in [existing tests](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller_test.go#L76) with the set func, or for some incoming tests?

droot: > The kubebuilder libraries wrapped the Reconcile method in a Before and After Reconcile function to control timing and test things such as how many times Reconcile is called. We removed these from controller-runtime, but this PR allows users to implement the same thing themselves by wrapping the existing Reconcile in a function that delegates.

I should have probably asked this earlier, is it intended to help write controller tests or controller-runtime tests ?

If it is for the controller-tests, do controller-test really need that sort of testing like how-many times reconcile got called ?  sounds more like controller-runtime tests to me.

> It also makes it possible for users to more easily configure cross-cutting aspects to Controllers - for instance adding logging or monitoring before all Reconcile functions.

hmm.... this is interesting, I want to know more about this because I needed something like this for the example and was thinking using adaptor pattern for that. Wrap the exising reconciler even before invoking `Watch` on the controller. Will chat offline about that.

lichuqiang: I just wonder if we can make the `Do` field internal if we decide to have the Set func, normally the set funcs are provided for outside users to specify internal fields.

pwittrock: @lichuqiang SGTM.

PR #48, opened 2018-06-18
Controller v2 libs review

pwittrock:


pkg/reconcile/reconcile.go
```diff
@@ -0,0 +1,85 @@
...
+		// Implement business logic of reading and writing objects here
+		return controller.Result{}, nil
+	})
+
+Reconciliation is level-based, meaning action isn't driven off changes in individual Events, but instead is
+driven by actual cluster state read from the apiserver or a local cache.
+For example if responding to a Pod Delete Event, the Request won't contain that a Pod was deleted,
+instead the reconcile function observes this when reading the cluster state and seeing the Pod as missing.
+*/
+type Reconcile interface {
```
droot: Should this be called `Reconciler` because anyone who implements the verb `Reconcile` is a Reconciler ? 

Like anyone implementing `Read(...)` is a `io.Reader` ?

pkg/reconcile/reconcile_test.go
```diff
@@ -0,0 +1,69 @@
...
+				Requeue: true,
+			}
+
+			instance := reconcile.Func(func(r reconcile.Request) (reconcile.Result, error) {
+				defer GinkgoRecover()
+				Expect(r).To(Equal(request))
+
+				return result, nil
+			})
+			actualResult, actualErr := instance(request)
```
droot: may be invoke the reconcile using `instance.Reconcile` for clarity ?

pkg/reconcile/example_test.go
```diff
@@ -0,0 +1,44 @@
...
+	})
+
+	r.Reconcile(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: "default", Name: "test"}})
+
+	// Output: Name: test, Namespace: default
+}
+
+// This example declares a simple type that implements reconcile.
+func ExampleReconcile() {
+	type MyReconcileImplementation struct {
```
droot: IMO, default example for demonstrating `Reconcile` should be struct based not adapter based because  99% use-cases involve a struct where an instance of `client` needs to be embedded etc. and speaking from my experience, I have found Adapter pattern a bit difficult to grok for newbies (experience from net/http)

pkg/reconcile/reconcile.go
```diff
@@ -0,0 +1,85 @@
...
+package reconcile
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Result contains the result of a Reconcile invocation.
+type Result struct {
+	// Requeue tells the Controller to requeue the reconcile key.  Defaults to false.
+	Requeue bool
```
droot: One question (not scope of `reconcile` pkg though). What is the semantics for handling the error returned by the `Reconcile()`. Is it going to be simply logging the error and ignoring it ? or using it for re-queuing ? If it is the former, then current controller writer might get surprised a bit (not saying it is a bad thing though).

pkg/event/event.go
```diff
@@ -0,0 +1,68 @@
...
+limitations under the License.
+*/
+
+package event
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
```
droot: May be a few comments explaining where this fit in the over-all picture.  Who will be the producer/consumer of these objects etc ?

pkg/handler/enqueue_owner.go
```diff
@@ -0,0 +1,165 @@
...
+	// groupKind is the cached Group and Kind from OwnerType
+	groupKind schema.GroupKind
+}
+
+var _ inject.Scheme = &EnqueueOwner{}
+
+// InjectScheme is called by the Controller to provide a singleton scheme to the EnqueueOwner.
+func (e *EnqueueOwner) InjectScheme(s *runtime.Scheme) error {
+	return e.parseOwnerTypeGroupKind(s)
+}
```
droot: would suggest moving `InjectScheme` related code at the bottom from code readability perspective, sort of coming in the way :)

pkg/handler/enqueue_owner.go
```diff
@@ -0,0 +1,165 @@
...
+
+// getOwnersReferences returns the OwnerReferences for an object as specified by the EnqueueOwner
+// - if IsController is true: only take the Controller OwnerReference (if found)
+// - if IsController is false: take all OwnerReferences
+func (e *EnqueueOwner) getOwnersReferences(object metav1.Object) []metav1.OwnerReference {
+	if object == nil {
+		return nil
+	}
+
+	// If filtered to Controller use all the OwnerReferences
```
droot: *not* filtered

pkg/handler/enqueue_owner.go
```diff
@@ -0,0 +1,165 @@
...
+		// If the two match, create a Request for the objected referred to by
+		// the OwnerReference.  Use the Name from the OwnerReference and the Namespace from the
+		// object in the event.
+		if ref.Kind == e.groupKind.Kind && refGV.Group == e.groupKind.Group {
+			// Match found - add a Request for the object referred to in the OwnerReference
+			result = append(result, reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: object.GetNamespace(),
+				Name:      ref.Name,
+			}})
+		}
```
droot: Its good that we are not performing lookup against the cluster to match the UUIDs for the owner because that would have made handling the cases of owner itself changing difficult.

pkg/handler/eventhandler.go
```diff
@@ -0,0 +1,104 @@
...
+	// Update is called in response to an update event -  e.g. Pod Updated.
+	Update(workqueue.RateLimitingInterface, event.UpdateEvent)
+
+	// Delete is called in response to a delete event - e.g. Pod Deleted.
+	Delete(workqueue.RateLimitingInterface, event.DeleteEvent)
+
+	// Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
+	// external trigger request - e.g. reconcile Autoscaling, or a Webhook.
+	Generic(workqueue.RateLimitingInterface, event.GenericEvent)
+}
```
droot: Consider switching ordering of the parameters. evt is the primary input to to the handler and q is sort of the output after the handling ?

pkg/handler/eventhandler.go
```diff
@@ -0,0 +1,104 @@
...
+
+var _ EventHandler = Funcs{}
+
+// Funcs allows specifying a subset of EventHandler functions are fields.
```
droot: This comment isn't very clear.

pkg/predicate/predicate.go
```diff
@@ -0,0 +1,83 @@
...
+	return true
+}
+
+// Generic implements Predicate
+func (p Funcs) Generic(e event.GenericEvent) bool {
+	if p.GenericFunc != nil {
+		return p.GenericFunc(e)
+	}
+	return true
+}
```
droot: looks like we don't have any default predicates implemented like one diffing `resourceVersion` field ?

pkg/source/source.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
+	i, err := ks.cache.GetInformer(ks.Type)
+	if err != nil {
+		return err
+	}
+	i.AddEventHandler(internal.EventHandler{Queue: queue, EventHandler: handler, Predicates: prct})
+	return nil
+}
```
droot: consider moving `Kind source` above `Channel source` from readability perspective. Kind source is much more simpler than channel source, so if someone is looking for a canonical example to implement a source, introducing Kind first is a good idea.

droot:
Did a first pass today and reviewed source, handler, predicates, event and reconcile pkgs today. Will review controllers and manager in the second pass today or Monday.

pkg/reconcile/reconcile.go
pwittrock: the latter. will document

pwittrock:
Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/58/files

pkg/reconcile/reconcile.go
DirectXMan12: agreed here -- the name is a bit confusing.

pkg/reconcile/example_test.go
DirectXMan12: agreed

pkg/manager/internal.go
```diff
@@ -0,0 +1,178 @@
...
+			if err := cm.startCache(stop); err != nil {
+				cm.errChan <- err
+			}
+		}()
+
+		// Wait for the caches to sync.
+		// TODO(community): Check the return value and write a test
+		cm.cache.WaitForCacheSync(stop)
+
+		// Start the runnables after the promises
```
droot: comment is outdated (mentions `promises`)

pkg/manager/internal.go
```diff
@@ -0,0 +1,178 @@
...
+func (cm *controllerManager) GetRecorder(name string) record.EventRecorder {
+	return cm.recorderProvider.GetEventRecorderFor(name)
+}
+
+func (cm *controllerManager) Start(stop <-chan struct{}) error {
+	func() {
+		cm.mu.Lock()
+		defer cm.mu.Unlock()
+
+		// Start the Cache.
```
droot: nit: may be move the comment new the `startCache` block below.

pkg/manager/manager.go
```diff
@@ -0,0 +1,167 @@
...
+	GetRecorder(name string) record.EventRecorder
+}
+
+// Options are the arguments for creating a new Manager
+type Options struct {
+	// Scheme is the scheme used to resolve runtime.Objects to GroupVersionKinds / Resources
+	// Defaults to the kubernetes/client-go scheme.Scheme
+	Scheme *runtime.Scheme
+
+	// Mapper is the rest mapper used to map go types to Kubernetes APIs
```
droot: Mapper --> MapperProvider provides the rest mapper......

pkg/manager/manager.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Initialize a rest.config if none was specified
+	if cm.config == nil {
+		return nil, fmt.Errorf("must specify Config")
+	}
+
+	// Use the Kubernetes client-go scheme if none is specified
+	if cm.scheme == nil {
+		cm.scheme = scheme.Scheme
+	}
```
droot: This should be moved to `setOptionsDefaults`

pkg/manager/manager.go
```diff
@@ -0,0 +1,167 @@
...
+		options.MapperProvider = apiutil.NewDiscoveryRESTMapper
+	}
+
+	// Allow newClient to be mocked
+	if options.newClient == nil {
+		options.newClient = client.New
+	}
+
+	// Allow newCache to be mocked
+	// TODO(directxman12): Figure out how to allow users to request a client without requesting a watch
```
droot: Don't know this TODO is still valid ? I remember some discussion where it came as *already* done ?

droot:
@pwittrock completed the entire review today. lgtm. Feel free to close the PR.

pkg/manager/internal.go
pwittrock: Done

pkg/manager/internal.go
pwittrock: Done

pkg/manager/manager.go
pwittrock: Done

pkg/manager/manager.go
pwittrock: Done.  Restructured this fn.

pkg/manager/manager.go
pwittrock: Removed

pwittrock:
Comments addressed in #61

pkg/handler/enqueue.go
```diff
@@ -0,0 +1,89 @@
...
+var enqueueLog = logf.KBLog.WithName("eventhandler").WithName("Enqueue")
+
+var _ EventHandler = &Enqueue{}
+
+// Enqueue enqueues a Request containing the Name and Namespace of the object for each event.
+type Enqueue struct{}
+
+// Create implements EventHandler
+func (e *Enqueue) Create(q workqueue.RateLimitingInterface, evt event.CreateEvent) {
+	if evt.Meta == nil {
```
DirectXMan12: :-/ I think we want to Inject a logger if we can, so that we can track *which* enqueue is having issues.

pkg/handler/enqueue_owner.go
```diff
@@ -0,0 +1,165 @@
...
+func (e *EnqueueOwner) parseOwnerTypeGroupKind(scheme *runtime.Scheme) error {
+	// Get the kinds of the type
+	kinds, _, err := scheme.ObjectKinds(e.OwnerType)
+	if err != nil {
+		log.Error(err, "Could not get ObjectKinds for OwnerType", "OwnerType", e.OwnerType)
+		return err
+	}
+	// Expect only 1 kind.  If there is more than one kind this is probably an edge case such as ListOptions.
+	if len(kinds) != 1 {
+		err := fmt.Errorf("Expected exactly 1 kind for OwnerType")
```
DirectXMan12: add more info to this error (like `kind is probably not a normal API type`)

pkg/handler/enqueue_owner.go
```diff
@@ -0,0 +1,165 @@
...
+func (e *EnqueueOwner) getOwnerReconcileRequest(object metav1.Object) []reconcile.Request {
+	// Iterate through the OwnerReferences looking for a match on Group and Kind against what was requested
+	// by the user
+	var result []reconcile.Request
+	for _, ref := range e.getOwnersReferences(object) {
+		// Parse the Group out of the OwnerReference to compare it to what was parsed out of the requested OwnerType
+		refGV, err := schema.ParseGroupVersion(ref.APIVersion)
+		if err != nil {
+			log.Error(err, "Could not parse OwnerReference GroupVersion",
+				"OwnerReference", ref.APIVersion)
```
DirectXMan12: ownerRef should be the full ownerRef

pkg/handler/eventhandler.go
```diff
@@ -0,0 +1,104 @@
...
+//
+// * Use Enqueue to reconcile the object the event is for
+// - do this for events for the type the Controller Reconciles. (e.g. Deployment for a Deployment Controller)
+//
+// * Use EnqueueOwner to reconcile the owner of the object the event is for
+// - do this for events for the types the Controller creates.  (e.g. ReplicaSets created by a Deployment Controller)
+//
+// * Use EnqueueMappendHandler to transform an event for an object to a reconcile of an object
+// of a different type - do this for events for types the Controller may be interested in, but doesn't create.
+// (e.g. If Foo responds to cluster size events, map Node events to Foo objects.)
```
DirectXMan12: maybe mention using indexers if you need to easily look things up?

pkg/manager/internal.go
```diff
@@ -0,0 +1,178 @@
...
+		cm.started = true
+	}()
+
+	select {
+	case <-stop:
+		// We are done
+		return nil
+	case err := <-cm.errChan:
+		// Error starting a controller
+		return err
```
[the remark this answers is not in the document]
droot: Typically, callers code does `Fatal` when they get first error, so doesn't matter.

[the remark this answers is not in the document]
DirectXMan12: we shouldn't depend on that, though

PR #49, opened 2018-06-18
Fix non-cache client list

DirectXMan12:
This fixes the non-cache client list, which had a null pointer
dereference (opts.Namespace), incorrect resource handling (need the
resource for the items, not the list itself), and incorrect handling of
nil list options (the underlying metav1.ListOptions needs to always be
non-nil, even if empty).


DirectXMan12:
@droot this should fix your issues.  The test in the style of the rest of the tests in this package is currently failing.  Need to figure out why, but this PR should fix your issue.

DirectXMan12:
brainfart on the test.  should be working now

PR #51, opened 2018-06-19
Support installing CRDs from yaml files in test package

pwittrock:
- also rename test package to envtest

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package envtest
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/ghodss/yaml"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+)
+
+// CRDInstallOptions are the options for installing CRDs
+type CRDInstallOptions struct {
+	// Paths is the path to the directory containing CRDs
+	Paths []string
+
+	// CRDs is a list of CRDs to install
+	CRDs []*apiextensionsv1beta1.CustomResourceDefinition
+
+	// maxTime is the max time to wait
+	maxTime time.Duration
+
+	// pollInterval is the interval to check
+	pollInterval time.Duration
+}
+
+const defaultPollInterval = 100 * time.Millisecond
+const defaultMaxWait = 10 * time.Second
+
+// InstallCRDs installs a collection of CRDs into a cluster by reading the crd yaml files from a directory
+func InstallCRDs(config *rest.Config, options CRDInstallOptions) ([]*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	defaultCRDOptions(&options)
+
+	// Read the CRD yamls into options.CRDs
+	if err := readCRDFiles(&options); err != nil {
+		return nil, err
+	}
+
+	// Create the CRDs in the apiserver
+	if err := CreateCRDs(config, options.CRDs); err != nil {
+		return options.CRDs, err
+	}
+
+	// Wait for the CRDs to appear as Resources in the apiserver
+	if err := WaitForCRDs(config, options.CRDs, options); err != nil {
+		return options.CRDs, err
+	}
+
+	return options.CRDs, nil
+}
+
+// readCRDFiles reads the directories of CRDs in options.Paths and adds the CRD structs to options.CRDs
+func readCRDFiles(options *CRDInstallOptions) error {
+	if len(options.Paths) > 0 {
+		for _, path := range options.Paths {
+			new, err := readCRDs(path)
+			if err != nil {
+				return err
+			}
+			options.CRDs = append(options.CRDs, new...)
+		}
+	}
+	return nil
+}
+
+// defaultCRDOptions sets the default values for CRDs
+func defaultCRDOptions(o *CRDInstallOptions) {
+	if o.maxTime == 0 {
+		o.maxTime = defaultMaxWait
+	}
+	if o.pollInterval == 0 {
+		o.pollInterval = defaultPollInterval
+	}
+}
+
+// WaitForCRDs waits for the CRDs to appear in discovery
+func WaitForCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResourceDefinition, options CRDInstallOptions) error {
+	// Add each CRD to a map of GroupVersion to Resource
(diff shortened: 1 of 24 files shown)


pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+func readCRDs(path string) ([]*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	// Get the CRD files
+	var files []os.FileInfo
+	var err error
+	if files, err = ioutil.ReadDir(path); err != nil {
+		return nil, err
+	}
+
+	// White list the file extensions that may contain CRDs
+	crdExts := sets.NewString(".json", ".yaml")
```
fanzhangio: Should `.yml` extension also be supported ?

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return nil, err
+		}
+		crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+		if err = yaml.Unmarshal(b, crd); err != nil {
+			return nil, err
+		}
+
+		// Check that it is actually a CRD
+		if crd.Spec.Names.Kind == "" || crd.Spec.Group == "" {
```
fanzhangio: Should take`Spec.Scope` to consideration as well ?

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+)
+
+// CRDOptions are the options for installing CRDs
+type CRDOptions struct {
```
droot: CRDInstallOptions ? (then it will be self-documenting I guess and we won't need comments)

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+const defaultMaxWait = 10 * time.Second
+
+// InstallCRDs installs a collection of CRDs into a cluster by reading the crd yaml files from a directory
+func InstallCRDs(config *rest.Config, options CRDOptions) ([]*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	defaultCRDOptions(&options)
+
+	// Read the CRD yamls into options.CRDs
+	if err := readCRDFiles(&options); err != nil {
+		return options.CRDs, err
+	}
```
droot: why are we not returning nil (slice of CRDs) in case of error ?

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Create the CRDs in the apiserver
+	if err := CreateCRDs(config, options.CRDs); err != nil {
+		return options.CRDs, err
```
droot: same question as above.

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// Wait for the CRDs to appear as Resources in the apiserver
+	if err := WaitForCRDs(config, options.CRDs, options); err != nil {
+		return options.CRDs, err
```
droot: same question as above.

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+
+// readCRDFiles reads the directories of CRDs in options.Paths and adds the CRD structs to options.CRDs
+func readCRDFiles(options *CRDOptions) error {
+	if len(options.Paths) > 0 {
+		for _, path := range options.Paths {
+			new, err := readCRDs(path)
+			if err != nil {
+				return err
+			}
+			options.CRDs = append(options.CRDs, new...)
```
droot: I get it now why we are returning non-nil list in case of error above. Ignore those comments.

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := WaitForCRDs(config, options.CRDs, options); err != nil {
+		return options.CRDs, err
+	}
+
+	return options.CRDs, nil
+}
+
+// readCRDFiles reads the directories of CRDs in options.Paths and adds the CRD structs to options.CRDs
+func readCRDFiles(options *CRDOptions) error {
+	if len(options.Paths) > 0 {
```
droot: This line is redundant. wondering if metalinter will catch this ?

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+		if _, found := waitingFor[gv]; !found {
+			// Initialize the set
+			waitingFor[gv] = &sets.String{}
+		}
+		// Add the Resource
+		waitingFor[gv].Insert(crd.Spec.Names.Plural)
+	}
+	// Poll until all resources are found
+	return wait.PollImmediate(options.pollInterval, options.maxTime, func() (bool, error) {
+		// Create a new clientset to avoid any client caching of discovery
```
droot: putting it in a separate func will improve the readability and avoid errors with sharing the `waitingFor`.

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+		allFound := true
+		for gv, resources := range waitingFor {
+			// All resources found, do nothing
+			if resources.Len() == 0 {
+				delete(waitingFor, gv)
+				continue
+			}
+
+			// Get the Resources for this GroupVersion
+			resourceList, err := cs.Discovery().ServerResourcesForGroupVersion(gv.Group + "/" + gv.Version)
```
droot: unrelated: I am thinking if we should have a method to discover resources in our new client interface ? or a way to get discoveryClient without getting the `cs`.

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		allFound := true
+		for gv, resources := range waitingFor {
+			// All resources found, do nothing
+			if resources.Len() == 0 {
+				delete(waitingFor, gv)
+				continue
```
droot: I wonder if it is safe to delete an element from map while iterating over it ?

pkg/envtest/crd.go
```diff
@@ -0,0 +1,200 @@
...
+// CreateCRDs creates the CRDs
+func CreateCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResourceDefinition) error {
+	cs, err := clientset.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	// Create each CRD
+	for _, crd := range crds {
+		if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
```
droot: wondering if we can use `pkg/client` do this ?

pkg/envtest/crd.go
pwittrock: Added a todo

pkg/envtest/crd.go
pwittrock: Yes it is.  I had the same question.

https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-golang-map-within-a-range-loop

pkg/envtest/crd.go
pwittrock: We could, but i need this for discovery anyway...

pkg/envtest/crd.go
pwittrock: done

pkg/envtest/crd.go
pwittrock: I don't thin so.  This is really just to check if it was an unrelated type that happened to parse since go doesn't throw an error if the file is valid yaml but the schema doesn't match.

pkg/envtest/crd.go
pwittrock: I cleaned this up a bit to get the linter to pass and the tests working .  PTAL

pkg/envtest/crd.go
pwittrock: done

pkg/envtest/crd.go
pwittrock: Done

pwittrock:
PTAL

PR #52, opened 2018-06-19
Added tests

seans3:
* Added "List" tests
* Added "ListOptions" tests
* Coverage is now 93.0%


pkg/client/client_test.go
```diff
@@ -648,23 +651,184 @@ var _ = Describe("Client", func() {
...
+			deps := &appsv1.DeploymentList{}
+			lo := client.InNamespace("test-namespace-1")
+			Expect(cl.List(context.Background(), lo, deps)).NotTo(HaveOccurred())
+
+			By("only the Deployment in test-namespace-1 is returned")
+			Expect(deps.Items).NotTo(BeEmpty())
+			Expect(1).To(Equal(len(deps.Items)))
+			actual := deps.Items[0]
+			Expect(actual.Name).To(Equal("deployment-frontend"))
+
```
droot: would be great if we can can another List() call here for `all-namespaces` and we get both the deployments. That will cover all the namespace based filtering.

PR #53, opened 2018-06-19
Implement predicate

fanzhangio:
https://github.com/kubernetes-sigs/kubebuilder/issues/240

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -16,7 +16,12 @@ limitations under the License.
 
 package predicate
 
-import "sigs.k8s.io/controller-runtime/pkg/event"
+import (
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.KBLog.WithName("predicate").WithName("eventFilters")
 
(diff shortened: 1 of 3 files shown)


pwittrock:
Why not copy the kubebuilder Predicates over:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/controller/predicates/predicates.go

The main ones we need are the ResourceVersionChanged, and the ability to compose predicates with And/Or

pkg/predicate/predicate.go
```diff
@@ -16,7 +16,12 @@ limitations under the License.
 
 package predicate
 
-import "sigs.k8s.io/controller-runtime/pkg/event"
+import (
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var predicateLog = logf.KBLog.WithName("predicate").WithName("eventFilters")
```
droot: predicateLog --> log ?

fanzhangio: Don't we use unified `runtime log` ?

droot: I was suggesting to rename `predicateLog` --> `log`

PR #54, opened 2018-06-19
Add helper function for adding OwnerReferences to metav1.Objects

pwittrock:


pkg/manager/internal.go
```diff
@@ -132,6 +136,26 @@ func (cm *controllerManager) GetRecorder(name string) record.EventRecorder {
 	return cm.recorderProvider.GetEventRecorderFor(name)
 }
 
+func (cm *controllerManager) SetControllerReference(owner, object v1.Object) error {
```
droot: It feels a bit awkward to add a helper method on `controllerManager` for this. I get the reasoning that this needs access to scheme which is only available at mgr level. Another way would be have this helper as a function which takes scheme as input (not elegant, but isn't too bad either), WDYT ?  (thinking along the lines of `io` and `ioutil`)

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -0,0 +1,52 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+var _ = Describe("Controllerutil", func() {
+	Describe("SetControllerReference", func() {
+		It("should set the OwnerReference if it can find the group version kind", func() {
+			rs := &appsv1.ReplicaSet{}
+			dep := &extensionsv1beta1.Deployment{
```
droot: any specific reason we are not using appsv1 instead of extensionsv1beta1 here ?

PR #59, opened 2018-06-26
First pass at informer cache tests

seans3:
* Initial pass for tests of informer cache.
* Test coverage 76.9%

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -191,10 +191,20 @@ var _ = Describe("Informer Cache", func() {
 
 	Describe("as an Informer", func() {
 		It("should be able to get informer for the object", func(done Done) {
-			By("getting an shared index informer for a pod")
+			By("getting a shared index informer for a pod")
 			pod := &kcorev1.Pod{
-				ObjectMeta: kmetav1.ObjectMeta{Name: "informer-obj", Namespace: "default"},
-				Spec:       kcorev1.PodSpec{Containers: []kcorev1.Container{{Name: "nginx", Image: "nginx"}}},
+				ObjectMeta: kmetav1.ObjectMeta{
+					Name:      "informer-obj",
+					Namespace: "default",
+				},
+				Spec: kcorev1.PodSpec{
+					Containers: []kcorev1.Container{
+						{
+							Name:  "nginx",
+							Image: "nginx",
+						},
+					},
+				},
 			}
 			sii, err := informerCache.GetInformer(pod)
 			Expect(err).NotTo(HaveOccurred())
@@ -203,9 +213,10 @@ var _ = Describe("Informer Cache", func() {
 
 			By("adding an event handler listening for object creation which sends the object to a channel")
 			out := make(chan interface{})
-			sii.AddEventHandler(kcache.ResourceEventHandlerFuncs{AddFunc: func(obj interface{}) {
+			addFunc := func(obj interface{}) {
 				out <- obj
-			}})
+			}
+			sii.AddEventHandler(kcache.ResourceEventHandlerFuncs{AddFunc: addFunc})
 
 			By("adding an object")
 			cl, err := client.New(cfg, client.Options{})
@@ -227,16 +238,27 @@ var _ = Describe("Informer Cache", func() {
 
 			By("adding an event handler listening for object creation which sends the object to a channel")
 			out := make(chan interface{})
-			sii.AddEventHandler(kcache.ResourceEventHandlerFuncs{AddFunc: func(obj interface{}) {
+			addFunc := func(obj interface{}) {
 				out <- obj
-			}})
+			}
+			sii.AddEventHandler(kcache.ResourceEventHandlerFuncs{AddFunc: addFunc})
 
 			By("adding an object")
 			cl, err := client.New(cfg, client.Options{})
 			Expect(err).NotTo(HaveOccurred())
 			pod := &kcorev1.Pod{
-				ObjectMeta: kmetav1.ObjectMeta{Name: "informer-gvk", Namespace: "default"},
-				Spec:       kcorev1.PodSpec{Containers: []kcorev1.Container{{Name: "nginx", Image: "nginx"}}},
+				ObjectMeta: kmetav1.ObjectMeta{
+					Name:      "informer-gvk",
+					Namespace: "default",
+				},
+				Spec: kcorev1.PodSpec{
+					Containers: []kcorev1.Container{
+						{
+							Name:  "nginx",
+							Image: "nginx",
+						},
+					},
+				},
 			}
 			Expect(cl.Create(context.TODO(), pod)).To(Succeed())
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -31,6 +31,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const testNamespaceOne = "test-namespace-1"
+const testNamespaceTwo = "test-namespace-2"
+
 var informerCache cache.Cache
 var stop chan struct{}
 var knownPod1 runtime.Object
@@ -74,9 +77,9 @@ var _ = Describe("Informer Cache", func() {
 		Expect(cfg).NotTo(BeNil())
 
 		By("creating three pods")
-		knownPod1 = createPod("test-pod-1", "test-namespace-1", kcorev1.RestartPolicyNever)
-		knownPod2 = createPod("test-pod-2", "test-namespace-2", kcorev1.RestartPolicyAlways)
-		knownPod3 = createPod("test-pod-3", "test-namespace-2", kcorev1.RestartPolicyOnFailure)
+		knownPod1 = createPod("test-pod-1", testNamespaceOne, kcorev1.RestartPolicyNever)
+		knownPod2 = createPod("test-pod-2", testNamespaceTwo, kcorev1.RestartPolicyAlways)
+		knownPod3 = createPod("test-pod-3", testNamespaceTwo, kcorev1.RestartPolicyOnFailure)
 
 		By("creating the informer cache")
 		var err error
@@ -135,7 +138,7 @@ var _ = Describe("Informer Cache", func() {
 			By("listing pods with a particular label")
 			// NB: each pod has a "test-label" equal to the pod name
 			out := kcorev1.PodList{}
-			Expect(informerCache.List(context.TODO(), client.InNamespace("test-namespace-2").
+			Expect(informerCache.List(context.TODO(), client.InNamespace(testNamespaceTwo).
 				MatchingLabels(map[string]string{"test-label": "test-pod-2"}), &out)).NotTo(HaveOccurred())
 
 			By("verifying the returned pods have the correct label")
@@ -150,21 +153,21 @@ var _ = Describe("Informer Cache", func() {
 			By("listing pods in test-namespace-1")
 			listObj := &kcorev1.PodList{}
 			Expect(informerCache.List(context.Background(),
-				client.InNamespace("test-namespace-1"),
+				client.InNamespace(testNamespaceOne),
 				listObj)).NotTo(HaveOccurred())
 
 			By("verifying that the returned pods are in test-namespace-1")
 			Expect(listObj.Items).NotTo(BeEmpty())
 			Expect(len(listObj.Items)).To(Equal(1))
 			actual := listObj.Items[0]
-			Expect(actual.Namespace).To(Equal("test-namespace-1"))
+			Expect(actual.Namespace).To(Equal(testNamespaceOne))
 			close(done)
 		})
 
 		It("should deep copy the object unless told otherwise", func(done Done) {
 			By("retrieving a specific pod from the cache")
 			out := &kcorev1.Pod{}
-			podKey := client.ObjectKey{Name: "test-pod-2", Namespace: "test-namespace-2"}
+			podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
 			Expect(informerCache.Get(context.Background(), podKey, out)).NotTo(HaveOccurred())
 
 			By("verifying the retrieved pod is equal to a known pod")
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -22,6 +22,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	kcorev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -102,8 +103,8 @@ var _ = Describe("Informer Cache", func() {
 		close(stop)
 	})
 
-	Describe("Cache as a Reader", func() {
-		It("should be able to list objects that haven't been watched previously", func(done Done) {
+	Describe("as a Reader", func() {
+		It("should be able to list objects that haven't been watched previously", func() {
 			By("listing all services in the cluster")
 			listObj := &kcorev1.ServiceList{}
 			Expect(informerCache.List(context.Background(), nil, listObj)).NotTo(HaveOccurred())
@@ -119,10 +120,9 @@ var _ = Describe("Informer Cache", func() {
 				}
 			}
 			Expect(hasKubeService).To(BeTrue())
-			close(done)
 		})
 
-		It("should be able to get objects that haven't been watched previously", func(done Done) {
+		It("should be able to get objects that haven't been watched previously", func() {
 			By("getting the Kubernetes service")
 			svc := &kcorev1.Service{}
 			svcKey := client.ObjectKey{Namespace: "default", Name: "kubernetes"}
@@ -131,10 +131,9 @@ var _ = Describe("Informer Cache", func() {
 			By("verifying that the returned service looks reasonable")
 			Expect(svc.Name).To(Equal("kubernetes"))
 			Expect(svc.Namespace).To(Equal("default"))
-			close(done)
 		})
 
-		It("should support filtering by labels", func(done Done) {
+		It("should support filtering by labels", func() {
 			By("listing pods with a particular label")
 			// NB: each pod has a "test-label" equal to the pod name
 			out := kcorev1.PodList{}
@@ -146,10 +145,9 @@ var _ = Describe("Informer Cache", func() {
 			Expect(len(out.Items)).To(Equal(1))
 			actual := out.Items[0]
 			Expect(actual.Labels["test-label"]).To(Equal("test-pod-2"))
-			close(done)
 		})
 
-		It("should be able to list objects by namespace", func(done Done) {
+		It("should be able to list objects by namespace", func() {
 			By("listing pods in test-namespace-1")
 			listObj := &kcorev1.PodList{}
 			Expect(informerCache.List(context.Background(),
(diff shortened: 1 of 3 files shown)


pkg/cache/cache_test.go
```diff
@@ -22,40 +22,90 @@ import (
...
+	kmetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kcache "k8s.io/client-go/tools/cache"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var informerCache cache.Cache
```
DirectXMan12: I'd move these inside the describe, so it's clear what they're being used by.

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,90 @@ import (
...
-
-			By("running the cache and waiting for it to sync")
-			go func() {
-				defer GinkgoRecover()
-				Expect(reader.Start(stop)).ToNot(HaveOccurred())
-			}()
-			Expect(reader.WaitForCacheSync(stop)).NotTo(BeFalse())
-
-			By("Listing all services in the cluster")
+	Describe("Cache as a Reader", func() {
```
DirectXMan12: personal preference (totally fine to ignore): it's fine to have this just be `as a Reader`, since it's already in the context of `Informer Cache`

pkg/cache/cache_test.go
```diff
@@ -66,188 +116,165 @@ var _ = Describe("Informer Cache", func() {
 				}
 			}
 			Expect(hasKubeService).To(BeTrue())
+			close(done)
```
DirectXMan12: this doesn't do async, right? so it doesn't need the done channel, IIRC

pkg/cache/cache_test.go
```diff
@@ -66,188 +116,165 @@ var _ = Describe("Informer Cache", func() {
...
+		})
+
+		It("should error out for missing objects", func(done Done) {
+			By("getting a service that does not exists")
+			svc := &kcorev1.Service{}
+			svcKey := client.ObjectKey{Namespace: "unknown", Name: "unknown"}
+
+			By("verifying that an error is returned")
+			Expect(informerCache.Get(context.Background(), svcKey, svc)).To(HaveOccurred())
```
DirectXMan12: check that the error is a not found error (kube errors IsNotFound).  we actually want to guarantee that contract.

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,157 @@ var _ = Describe("Informer Cache", func() {
...
+			Expect(informerCache.Get(context.Background(), podKey, out)).NotTo(HaveOccurred())
+
+			By("verifying the retrieved pod is equal to a known pod")
+			Expect(out).To(Equal(knownPod2))
+
+			By("altering a field in the retrieved pod")
+			*out.Spec.ActiveDeadlineSeconds = 4
+
+			By("verifying the pods are no longer equal")
+			Expect(out).NotTo(Equal(knownPod2))
```
droot: +1 for this test.

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,157 @@ var _ = Describe("Informer Cache", func() {
...
+			By("verifying that an error is returned")
+			err := informerCache.Get(context.Background(), svcKey, svc)
+			Expect(err).To(HaveOccurred())
+			Expect(errors.IsNotFound(err)).To(Equal(true))
+		})
+	})
+
+	Describe("as an Informer", func() {
+		It("should be able to get informer for the object", func(done Done) {
+			By("getting an shared index informer for a pod")
```
droot: an -> a

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,157 @@ var _ = Describe("Informer Cache", func() {
...
+			pod := &kcorev1.Pod{
+				ObjectMeta: kmetav1.ObjectMeta{Name: "informer-obj", Namespace: "default"},
+				Spec:       kcorev1.PodSpec{Containers: []kcorev1.Container{{Name: "nginx", Image: "nginx"}}},
+			}
```
droot: IMO, squeezing above two values in single line impact the readability. I think making them multi-line will appear more verbose, but improves the readability.

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,157 @@ var _ = Describe("Informer Cache", func() {
...
+
+		It("should be able to index an object field then retrieve objects by that field", func() {
+			By("creating the cache")
+			informer, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("indexing the restartPolicy field of the Pod object before starting")
+			pod := &kcorev1.Pod{}
+			Expect(informer.IndexField(pod, "spec.restartPolicy", func(obj runtime.Object) []string {
+				return []string{string(obj.(*kcorev1.Pod).Spec.RestartPolicy)}
```
droot: nit: Can move out indexFn as separate variable to make it more readable.

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,157 @@ var _ = Describe("Informer Cache", func() {
...
+				listObj)).NotTo(HaveOccurred())
 
-			By("Verifying that the returned service looks reasonable")
-			Expect(svc.Name).To(Equal("kubernetes"))
-			Expect(svc.Namespace).To(Equal("default"))
+			By("verifying that the returned pods have correct restart policy")
+			Expect(listObj.Items).NotTo(BeEmpty())
+			Expect(len(listObj.Items)).To(Equal(1))
+			actual := listObj.Items[0]
+			Expect(actual.Name).To(Equal("test-pod-3"))
```
droot: +1 for testing the indexer :)

pkg/envtest/server.go
```diff
@@ -34,6 +34,17 @@ const (
...
+// APIServerDefaultArgs are flags necessary to bring up apiserver.
+// TODO: create test framework interface to append flag to default flags.
+var defaultKubeAPIServerFlags = []string{
+	"--etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}",
+	"--cert-dir={{ .CertDir }}",
+	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
+	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
+	"--secure-port=0",
+	"--admission-control=AlwaysAdmit",
+}
```
droot: Didn't see this in any other places where we were using testing framework. Will dig more into it.

seans3: This adds the "--admission-control=AlwaysAdmit" flag to the other default flags (unfortunately it's not possible to append a flag to the defaults) for the kube apiserver. Arguably, this flag should be one of the defaults. Without it, it's not possible to add test Pods without running into the following types of errors:

`pods "test-pod-1" is forbidden: error looking up service account test-namespace-1/default: serviceaccount "default" not found`

We don't see this in other tests, because we either add Deployments or Services.

pkg/cache/cache_test.go
seans3: Thanks

pkg/cache/cache_test.go
seans3: Thanks

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,94 @@ import (
...
 )
 
+const testNamespaceOne = "test-namespace-1"
+const testNamespaceTwo = "test-namespace-2"
+
+var informerCache cache.Cache
+var stop chan struct{}
+var knownPod1 runtime.Object
+var knownPod2 runtime.Object
+var knownPod3 runtime.Object
```
DirectXMan12: move these inside `Describe` (or reply here as to why they're not there).  It makes it clearer where they're used

DirectXMan12: also, use a 

```go
var (

)
```
block

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,180 @@ var _ = Describe("Informer Cache", func() {
...
+		It("should be able to list objects by namespace", func() {
+			By("listing pods in test-namespace-1")
+			listObj := &kcorev1.PodList{}
+			Expect(informerCache.List(context.Background(),
+				client.InNamespace(testNamespaceOne),
+				listObj)).NotTo(HaveOccurred())
+
+			By("verifying that the returned pods are in test-namespace-1")
+			Expect(listObj.Items).NotTo(BeEmpty())
+			Expect(len(listObj.Items)).To(Equal(1))
```
DirectXMan12: you can collapse these into `HaveLen`, IIRC

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,180 @@ var _ = Describe("Informer Cache", func() {
...
+
+		It("should error out for missing objects", func() {
+			By("getting a service that does not exists")
+			svc := &kcorev1.Service{}
+			svcKey := client.ObjectKey{Namespace: "unknown", Name: "unknown"}
+
+			By("verifying that an error is returned")
+			err := informerCache.Get(context.Background(), svcKey, svc)
+			Expect(err).To(HaveOccurred())
+			Expect(errors.IsNotFound(err)).To(Equal(true))
```
DirectXMan12: `BeTrue()` is a bit clearer (very minor nit)

pkg/cache/cache_test.go
```diff
@@ -69,185 +123,180 @@ var _ = Describe("Informer Cache", func() {
...
+				Spec: kcorev1.PodSpec{
+					Containers: []kcorev1.Container{
+						{
+							Name:  "nginx",
+							Image: "nginx",
+						},
+					},
+				},
+			}
+			Expect(cl.Create(context.TODO(), pod)).To(Succeed())
```
DirectXMan12: `context.Background()`, not `context.TODO()`, unless you're really intending on coming back and changing this later (https://golang.org/pkg/context/#Background)

pkg/envtest/server.go
```diff
@@ -58,8 +69,9 @@ func (te *Environment) Stop() error {
 // Start starts a local Kubernetes server and updates te.ApiserverPort with the port it is listening on
 func (te *Environment) Start() (*rest.Config, error) {
 	te.ControlPlane = integration.ControlPlane{}
+	te.ControlPlane.APIServer = &integration.APIServer{Args: defaultKubeAPIServerFlags}
```
pwittrock: This is fine for now.  Would you create an issue to allow users more fine grain control of the flags passed to the apiserver?

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,96 @@ import (
...
+	kcache "k8s.io/client-go/tools/cache"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const testNamespaceOne = "test-namespace-1"
+const testNamespaceTwo = "test-namespace-2"
+
+func createPod(name, namespace string, restartPolicy kcorev1.RestartPolicy) runtime.Object {
```
pwittrock: This seems generally useful to a variety of tests.  Add `TODO(community): ` to pull these into testenv.  Perhaps these should be functions on testenv.Environment.

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			By("verifying that the returned service looks reasonable")
+			Expect(svc.Name).To(Equal("kubernetes"))
+			Expect(svc.Namespace).To(Equal("default"))
+		})
+
+		It("should support filtering by labels", func() {
+			By("listing pods with a particular label")
+			// NB: each pod has a "test-label" equal to the pod name
+			out := kcorev1.PodList{}
+			Expect(informerCache.List(context.Background(), client.InNamespace(testNamespaceTwo).
```
pwittrock: nit: `Expect(...).To(Succeed())` is the same thing, but reads more naturally in the ginkgo output.

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+				MatchingLabels(map[string]string{"test-label": "test-pod-2"}), &out)).NotTo(HaveOccurred())
+
+			By("verifying the returned pods have the correct label")
+			Expect(out.Items).NotTo(BeEmpty())
```
pwittrock: This can also be written:

```go
Expect(out.Items).To(And(
  HaveLen(1),
  HaveKeyWithValue("test-label", "test-pod-2"),
))
```

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			Expect(out.Items).Should(HaveLen(1))
+			actual := out.Items[0]
+			Expect(actual.Labels["test-label"]).To(Equal("test-pod-2"))
+		})
+
+		It("should be able to list objects by namespace", func() {
+			By("listing pods in test-namespace-1")
+			listObj := &kcorev1.PodList{}
+			Expect(informerCache.List(context.Background(),
```
pwittrock: Same comment about Succeed() and And(HaveLen(), HaveKeyWithValue())

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			// NB: each pod has a "test-label" equal to the pod name
+			out := kcorev1.PodList{}
+			Expect(informerCache.List(context.Background(), client.InNamespace(testNamespaceTwo).
+				MatchingLabels(map[string]string{"test-label": "test-pod-2"}), &out)).NotTo(HaveOccurred())
+
+			By("verifying the returned pods have the correct label")
+			Expect(out.Items).NotTo(BeEmpty())
+			Expect(out.Items).Should(HaveLen(1))
+			actual := out.Items[0]
+			Expect(actual.Labels["test-label"]).To(Equal("test-pod-2"))
```
pwittrock: Can you also check that it will get Pods from multiple namespaces in a single list?

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,96 @@ import (
...
+		By("creating the informer cache")
+		var err error
+		informerCache, err = cache.New(cfg, cache.Options{})
+		Expect(err).NotTo(HaveOccurred())
+		By("running the cache and waiting for it to sync")
+		go func() {
+			defer GinkgoRecover()
+			Expect(informerCache.Start(stop)).ToNot(HaveOccurred())
+		}()
+		Expect(informerCache.WaitForCacheSync(stop)).NotTo(BeFalse())
```
pwittrock: We should also test what happens when we don't wait for the cache to sync.  The code *should* wait until the cache syncs before fetching the result even if this isn't called.

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			Expect(out).To(Equal(knownPod2))
+
+			By("altering a field in the retrieved pod")
+			*out.Spec.ActiveDeadlineSeconds = 4
+
+			By("verifying the pods are no longer equal")
+			Expect(out).NotTo(Equal(knownPod2))
+		})
+
+		It("should error out for missing objects", func() {
```
pwittrock: nit: `should return an error if the object is not found`

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			By("getting a service that does not exists")
+			svc := &kcorev1.Service{}
+			svcKey := client.ObjectKey{Namespace: "unknown", Name: "unknown"}
+
+			By("verifying that an error is returned")
```
pwittrock: Also test that type of error is of the not found variety so that when controllers check against the error they will know if it is a deletion error or not.

pkg/cache/cache_test.go
DirectXMan12: No, because we're testing actual.Labels, so that code won't work

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+							Name:  "nginx",
+							Image: "nginx",
+						},
+					},
+				},
+			}
+			sii, err := informerCache.GetInformer(pod)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(sii).NotTo(BeNil())
+			Expect(sii.HasSynced()).To(BeTrue())
```
pwittrock: Nice check.  We should also check this is true when we don't wait for the cache

pkg/cache/cache_test.go
```diff
@@ -69,185 +125,180 @@ var _ = Describe("Informer Cache", func() {
...
+			Expect(cl.Create(context.Background(), pod)).To(Succeed())
+
+			By("verifying the object is received on the channel")
+			Eventually(out).Should(Receive(Equal(pod)))
+			close(done)
+		})
+
+		It("should be able to get an informer by group/version/kind", func(done Done) {
+			By("getting an shared index informer for gvk = core/v1/pod")
+			gvk := schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"}
```
pwittrock: Add a test for when the GVK isn't in the Scheme.  Does the code support getting an informer for an unstructured object?  This should definitely be possible, just not sure if we support it right now.

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,96 @@ import (
...
+		knownPod2     runtime.Object
+		knownPod3     runtime.Object
+	)
 
 	BeforeEach(func() {
 		stop = make(chan struct{})
 		Expect(cfg).NotTo(BeNil())
+
+		By("creating three pods")
+		knownPod1 = createPod("test-pod-1", testNamespaceOne, kcorev1.RestartPolicyNever)
```
pwittrock: Add a comment that the restartpolicy is used by the field indexing test so folks know not to change these.

pkg/cache/cache_test.go
```diff
@@ -22,40 +22,96 @@ import (
...
-			Expect(reader.WaitForCacheSync(stop)).NotTo(BeFalse())
-
-			By("Listing all services in the cluster")
+			By("listing all services in the cluster")
 			listObj := &kcorev1.ServiceList{}
-			Expect(reader.List(context.Background(), nil, listObj)).NotTo(HaveOccurred())
+			Expect(informerCache.List(context.Background(), nil, listObj)).NotTo(HaveOccurred())
 
-			By("Verifying that the returned list contains the Kubernetes service")
+			By("verifying that the returned list contains the Kubernetes service")
```
pwittrock: where does the service come from?  I don't see it created anywhere.

pwittrock:
Mainly nits around using gomega matchers.  One thing to follow up on is more targeted testing around the making sure that caches are always sync'ed before results are fetched from them, even if the user doesn't call wait for sync ahead of time.

pkg/envtest/server.go
droot: Got it. Thanks for clarifying it.

pkg/cache/cache_test.go
seans3: Updated "NB: kubernetes default service is automatically created in testenv. " on line 117.

pkg/cache/cache_test.go
seans3: I put a TODO for this test.

pkg/envtest/server.go
seans3: Done: https://github.com/kubernetes-sigs/testing_frameworks/issues/55

pkg/cache/cache_test.go
seans3: Done

pkg/cache/cache_test.go
seans3: Added Succeed().  Skipping the " And(HaveLen(), HaveKeyWithValue())", since Solly had an issue with it.

pkg/cache/cache_test.go
pwittrock: ah right,

so just `Expect(actual.Labels).To(HaveKeyWithValue("test-label", "test-pod-2"))`

PR #62, opened 2018-06-26
Scheme helper fn

pwittrock:


pkg/runtime/scheme/scheme_test.go
```diff
@@ -0,0 +1,29 @@
...
+	Describe("NewAddToScheme", func() {
+		It("should return a function that will add the types to the scheme", func() {
+			gv := schema.GroupVersion{Group: "core", Version: "v1"}
+			a := scheme.NewAddToScheme(gv, func() []runtime.Object {
+				return []runtime.Object{&v1.Pod{}, &v1.PodList{}}
+			})
+			s := runtime.NewScheme()
+
+			Expect(a(s)).To(Succeed())
+			Expect(s.AllKnownTypes()).To(HaveLen(13))
```
mengqiy: I don't understand why it has 13 items.
Maybe add a comments.

pkg/runtime/scheme/scheme_test.go
```diff
@@ -0,0 +1,29 @@
...
+			Expect(s.AllKnownTypes()[gv.WithKind("Pod")]).To(Equal(reflect.TypeOf(v1.Pod{})))
+			Expect(s.AllKnownTypes()[gv.WithKind("PodList")]).To(BeEquivalentTo(reflect.TypeOf(v1.PodList{})))
```
mengqiy: Why this one is using `BeEquivalentTo` but not `Equal`?
What is the difference between Pod and PodList?

pkg/runtime/scheme/scheme.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// NewAddToScheme returns a function to add go struct types as GroupVersionKinds to a runtime.Scheme.
+// Scheme is used to map go structs to GroupVersionKinds and vice versa.
+func NewAddToScheme(gv schema.GroupVersion, types func() []runtime.Object) func(*runtime.Scheme) error {
```
DirectXMan12: This is a really weird name, and seems like a weird return type. Find a better name, maybe?

pwittrock: It creates a new `AddToScheme` function.

pwittrock:
> I think the structure of this is fairly confusing. Why not have a helper like NewStandardScheme, which returns a new Scheme with these defaults already applied? You could still have this method available, but this seems like a weird entry point for common cases.

The approach of creating a NewAddToScheme function seems like a standard pattern across the kubernetes/kubernetes codebase.  This removes the need to copy-paste the code.  I think some of the code-generators rely upon the AddToScheme package function existing.

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/abac/v1beta1/register.go

mengqiy:
LGTM
golint is failing.

pwittrock:
@mengqiy Fixed the comment :)

PR #65, opened 2018-06-28
Don't fail envtest if directory is missing

pwittrock:


pkg/envtest/crd.go
```diff
@@ -39,6 +39,9 @@ type CRDInstallOptions struct {
 	// CRDs is a list of CRDs to install
 	CRDs []*apiextensionsv1beta1.CustomResourceDefinition
 
+	// ErrorIfPathMissing will cause an error if a Path does not exist
+	ErrorIfPathMissing bool
```
droot: Sensible default in this case would be `true`, but its the opposite currently. Consider renaming it something like `SkipMissingCRDs` ? so that default is *to error if files missing". 

Which case do you need this ?

pwittrock: This is necessary for scaffolded projects without custom resources.  I ran into this when creating a new controller-only project for deployments.  The scaffolded test tried to read the crds, but the directory didn't exist because there were no user defined resources.

PR #71, opened 2018-07-06
fix vendor logr

mengqiy:


grantr:
I think this PR missed updating a reference to `zaplogr`:

https://github.com/kubernetes-sigs/controller-runtime/blob/6865d0a1bfc9afa3bc76a5f2e2c656ddbbd178e2/pkg/runtime/log/log.go#L28

mengqiy:
Not true.
`zaplogr` is referencing the package name https://github.com/go-logr/zapr/blob/7719b12d150e27231f2c113e7d40901f35e92287/zaplogr.go#L38 in `"github.com/go-logr/zapr"` :)

grantr:
Oh I see. The bug is actually in [go-logr/zapr](https://github.com/go-logr/zapr), which is documented with the package name and path`zapr` but is actually called `zaplogr`. I'm surprised Go doesn't consider that a bug, or at least a vet error. Thanks for clearing up my confusion @mengqiy!

PR #73, opened 2018-07-09
Fix un-initialized once that causes panic

anfernee:


mengqiy:
LGTM
Will apply labels after travis passes

pkg/admission/controller_test.go
```diff
@@ -0,0 +1,13 @@
+package admission
+
+import (
+	"testing"
+
+	"k8s.io/api/admissionregistration/v1beta1"
+)
+
+func TestCertProvisionerInit(t *testing.T) {
+	p := &CertProvisioner{}
```
mengqiy: You need to use a fake client like `Client: fake.NewFakeClient()`.
Otherwise, it will try find the kubeconfig which doesn't exist in the test env.

PR #74, opened 2018-07-10
validate cert

mengqiy:


mengqiy:
cc: @anfernee

pkg/admission/cert/writer/certwriter.go
```diff
@@ -91,8 +96,16 @@ func handleCommon(webhook *admissionregistrationv1beta1.Webhook, ch certReadWrit
 		return err
 	}
 
+	dnsName, err := dnsNameForWebhook(&webhook.ClientConfig)
+	if err != nil {
+		return err
+	}
 	// Recreate the cert if it's invalid.
-	if !validCert(certs) {
+	valid, err := validCert(certs, dnsName, 6*month)
```
Liujingfang1: Maybe move `6*month` to a constant.

pkg/admission/cert/writer/certwriter.go
```diff
@@ -141,19 +154,37 @@ type certReadWriter interface {
...
 
-func validCert(certs *generator.Artifacts) bool {
-	// TODO:
-	// 1) validate the key and the cert are valid pair e.g. call crypto/tls.X509KeyPair()
-	// 2) validate the cert with the CA cert
-	// 3) validate the cert is for a certain DNSName
-	// e.g.
-	// c, err := tls.X509KeyPair(cert, key)
-	// err := c.Verify(options)
+func validCert(certs *generator.Artifacts, dnsName string, validTime time.Duration) (bool, error) {
```
Liujingfang1: All the return statements returns error as `nil`. I think you can remove this argument from the return list. Then `validCert` returns only a `bool`

mengqiy: The linter is complaining about parameter `validTime` not being changed :/
It is handy to have this parameter for testing purpose.

mengqiy: Good catch

mengqiy:
PTAL

pkg/admission/cert/writer/certwriter.go
```diff
@@ -39,6 +42,8 @@ const (
 	ServerCertName = "cert.pem"
 )
 
+const sixMonths = 6 * 31 * 24 * time.Hour
```
droot: does it have to be pkg level const ? I see it only being used inside a fn.

pkg/admission/cert/writer/certwriter.go
```diff
@@ -141,18 +151,36 @@ type certReadWriter interface {
...
+		return false
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return false
+	}
+	ops := x509.VerifyOptions{
+		DNSName:     dnsName,
+		Roots:       pool,
+		CurrentTime: time.Now().Add(sixMonths),
```
anfernee: why 6 months? also it's more like `time.Now().AddDate(0, 1,  0)` [doc](https://golang.org/pkg/time/#Time.AddDate)

mengqiy: Good point!
Updated

mengqiy:
PTAL

PR #80, opened 2018-07-17
fix controllerutil_test

mengqiy:
test package name should be consist with https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil_suite_test.go#L1
Otherwise, causing issue when building with bazel.


mengqiy:
Pushed additional change. Reapply labels per https://github.com/kubernetes-sigs/controller-runtime/pull/80#pullrequestreview-137936307

PR #82, opened 2018-07-20
Avoid adding test flags when using the null logger

DirectXMan12:
The null logger from logr is in the same package as the testing logger,
which means it accidentally adds testing flags to all binaries that
import it.  Copy the null logger into the pkg/runtime/log package
to avoid this.

### DIFF
--- a/pkg/runtime/log/log.go
+++ b/pkg/runtime/log/log.go
@@ -6,7 +6,6 @@ import (
 	"log"
 
 	"github.com/go-logr/logr"
-	tlogr "github.com/go-logr/logr/testing"
 	"github.com/go-logr/zapr"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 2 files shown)


mengqiy:
LGTM
gometalinter is not happy. Please fix it :)

DirectXMan12:
This should fix the lint issues.  I always forget to run gometalinter, since it's not written anywhere.  Maybe we should add it to CONTRIBUTING.md or have a makefile target for it, or something?

mengqiy:
Yeah, we can add a section in `CONTRIBUTING.md` for it.

Linter is still complaining.
You can run `./test.sh` to ensure everything is passing before push new changes :)

DirectXMan12:
there, that should fix it.  actually remembered to run the right linter command this time ;-)

PR #84, opened 2018-07-24
config: fix using the KUBECONFIG to build the client

rphillips:
Fix passing the kubeconfig to the client factory.

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -62,7 +62,7 @@ func GetConfig() (*rest.Config, error) {
 	}
 	// If an env variable is specified with the config locaiton, use that
 	if len(os.Getenv("KUBECONFIG")) > 0 {
-		return clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
+		return clientcmd.BuildConfigFromFlags(masterURL, os.Getenv("KUBECONFIG"))
 	}
 	// If no explicit location, try the in-cluster config
 	if c, err := rest.InClusterConfig(); err == nil {


DirectXMan12:
approve bot might be borked right now, need to check.

migueleliasweb:
There appears to be a regression error here. In the master branch the code seem to have been changed back to exactly what this PR fixed initially. Is that right?

Master branch: https://github.com/rphillips/controller-runtime/blob/master/pkg/client/config/config.go

```golang
func GetConfig() (*rest.Config, error) {
	if len(kubeconfig) > 0 {
		return clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
	}
	if len(os.Getenv("KUBECONFIG")) > 0 {
                 // HERE we have kubeconfig instead of os.Getenv("KUBECONFIG") again
		return clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
	}
	if c, err := rest.InClusterConfig(); err == nil {
		return c, nil
	}
	if usr, err := user.Current(); err == nil {
		if c, err := clientcmd.BuildConfigFromFlags(
			"", filepath.Join(usr.HomeDir, ".kube", "config")); err == nil {
			return c, nil
		}
	}

	return nil, fmt.Errorf("could not locate a kubeconfig")
}
```

sbueringer:
There is no master branch anymore on the usptream repository. And the code on the main branch looks different

migueleliasweb:
Oh my god, I'm so sorry. I swear to god I was looking at the upstream kubernetes-sigs repo. My bad. Too many tabs open!

PR #86, opened 2018-07-25
Clean up logging usage (and minor script improvements)

DirectXMan12:
This fixes up a bunch of places where we were using the go `log` package in the standard library, or glog, fixes a bug found in the delegating logger, and introduces proper tests for the delegating logger to prevent that from happening again.

It also contains a couple minor quality-of-life improvements to the tests scripts -- you can now run linters and testing separately, and can simply specify `KUBEBUILDER_ASSETS` instead of exporting three different variables if your kubebuilder lives elsewhere.


droot:
Thanks for the PR @DirectXMan12 . Took a quick look and looks good. 

@mengqiy Can you also pl. take a look at the changes in the admission pkg ?

pkg/admission/cert/writer/fs.go
```diff
@@ -148,7 +147,7 @@ func (f *fsCertWriter) doWrite(webhookName string) (*certgenerator.Artifacts, er
 	if err != nil {
 		return nil, err
 	}
-	aw, err := atomic.NewAtomicWriter(v.path, fmt.Sprintf("processing webhook %q", webhookName))
+	aw, err := atomic.NewAtomicWriter(v.path, log.WithName("atomic-writer").WithValues("task", "processing webhook", "webhook", webhookName))
```
mengqiy: Not in the scope of this PR.
I feel the signature of method `WithValues` is a little confusing, if I don't know that the list of interface with 2*n items is actually representing n key-value pairs.
Why not using a signature that taking a map?

pkg/runtime/log/log.go
```diff
@@ -3,15 +3,20 @@
...
 
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 // ZapLogger is a Logger implementation.
-// if development is true stack traces will be printed for errors
+// If development is true, a Zap development config ill be used
```
mengqiy: s/ill/will

pkg/admission/cert/writer/internal/atomic/atomic_writer.go
```diff
@@ -57,8 +56,8 @@ const (
 // inotify or fanotify to receive events when the content in the volume is
 // updated.
 type AtomicWriter struct {
-	targetDir  string
-	logContext string
+	targetDir string
+	log       logr.Logger
```
mengqiy: Changing this file means that it starts to diverge from upstream.
But I guess it's not a problem.

pkg/admission/cert/writer/fs.go
DirectXMan12: the interface is mostly copied from Zap, so using a map means more overhead in our primary usecase.  Beyond that, Zap specifically does it to allow certain tricks that we don't take advantage of now, but we could consider in the future (i.e. helpers which have their own tag built in, improvements around number of allocations, etc).   Plus, it's fewer keystrokes ;-).  I do understand where you're coming from, though.

pkg/admission/cert/writer/internal/atomic/atomic_writer.go
DirectXMan12: if we actually have our own copy, we should treat it as such, IMO.

DirectXMan12:
@mengqiy fixes should be in place now, so this should be ready to merge.

Issue #89, opened 2018-07-26
dep ensure emits a warning and requires Mercurial


grantr:
When I try to run `dep ensure`, I get this warning, plus the command never returns:

```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  k8s.io/kube-aggregator

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.
```

When I Ctrl-C, I get this error:

```
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export bitbucket.org/ww/goautoneg:
	(1) hg is not installed:
	(2) hg is not installed:
	(3) hg is not installed:
	(4) failed to list versions for https://bitbucket.org/ww/goautoneg: remote: Not Found
fatal: repository 'https://bitbucket.org/ww/goautoneg/' not found
: exit status 128
	(5) failed to list versions for ssh://git@bitbucket.org/ww/goautoneg: : signal: interrupt
	(6) context canceled
	(7) context canceled
```

When I install Mercurial, the warning remains but the command finishes.

DirectXMan12:
Yep, I think we have an incorrectly specify dependency in our `Gopkg.toml`.  Is this from master?  I thought we'd fixed it, but it might have gotten re-introduced.

DirectXMan12:
(the mercurial dep is a known issue -- most of kube depends on it)

grantr:
Yes, this is master.

DirectXMan12:
It's an easy fix for the warning.  I'll have a PR up shortly

DirectXMan12:
PR up -- CCed you in it.

grantr:
Closing this but it would be nice to mention in contributor docs somewhere that Mercurial must be installed.

PR #91, opened 2018-07-26
Remove kube-aggregator from Gopkg.toml

DirectXMan12:
It's not directly used by anything, and the transitive deps should pull
in the proper version if needed.

Partial fix for #89


DirectXMan12:
cc @grantr

grantr:
:+1: works now.

PR #94, opened 2018-07-26
Add options to client.Delete()

### LINKED ISSUE #76: client.Delete doesn't support PropagationPolicy and other delete options
The controller-runtime client doesn't seem to support passing `DeleteOptions` in the body of a delete request.

Use case: when a [Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) is deleted, its pods are normally orphaned by the GC. To delete a Job and GC its pods with the normal client-go, I pass a `DeleteOptions` with `PropagationPolicy` set. I can't do that with the controller-runtime client.

grantr:
Allows Delete requests to carry options like GracePeriodSeconds, Preconditions, and PropagationPolicy. I chose the functional options pattern here so the function signature remains the same for all existing callsites.

To use the default options, call Delete with the same signature as before:

```go
client.Delete(context.TODO(), obj)
```

This Delete call sets a GracePeriodSeconds for the request:

```go
client.Delete(context.TODO(), obj, client.GracePeriodSeconds(1))
```

Multiple options can be set:

```go
client.Delete(context.TODO(), obj,
  client.GracePeriodSeconds(1),
  client.PropagationPolicy(metav1.DeletePropagationForeground))
```

I wrote tests for the DeleteOptions struct, but I wasn't sure how to test the actual request content. With the current testing strategy, we can only test behavior by inspecting what the apiserver did, and in this case that's fairly subtle since it involves timeouts and cascading deletes. It would be easier to use `httptest` or a similar mock to inspect the request that's sent by the client.

Fixes #76.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -111,15 +111,18 @@ func (c *client) Update(ctx context.Context, obj runtime.Object) error {
 }
 
 // Delete implements client.Client
-func (c *client) Delete(ctx context.Context, obj runtime.Object) error {
+func (c *client) Delete(ctx context.Context, obj runtime.Object, opts ...DeleteOptionFunc) error {
 	o, err := c.cache.getObjMeta(obj)
 	if err != nil {
 		return err
 	}
+
+	deleteOpts := DeleteOptions{}
 	return o.Delete().
 		NamespaceIfScoped(o.GetNamespace(), o.isNamespaced()).
 		Resource(o.resource()).
 		Name(o.GetName()).
+		Body(deleteOpts.ApplyOptions(opts).AsDeleteOptions()).
 		Do().
 		Error()
 }
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -978,6 +978,52 @@ var _ = Describe("Client", func() {
 		})
 	})
 
+	Describe("DeleteOptions", func() {
+		It("should allow setting GracePeriodSeconds", func() {
+			do := &client.DeleteOptions{}
+			client.GracePeriodSeconds(1)(do)
+			gp := int64(1)
+			Expect(do.AsDeleteOptions().GracePeriodSeconds).To(Equal(&gp))
+		})
+
+		It("should allow setting Precondition", func() {
+			do := &client.DeleteOptions{}
+			pc := metav1.NewUIDPreconditions("uid")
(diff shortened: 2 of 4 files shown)


DirectXMan12:
Our list options are chained instead of varargs.  I'm not necessary objecting to the varargs approach (I like how it ends up looking, but I think there are merits to both), but I think we need to be consistent.

grantr:
In general, the benefit of functional options is that they're invisible until you need them. This seems more useful for `Delete` than `List`. I expect `Delete` options will be rarely used, while almost all `List` calls will use options. The two operations have different ergonomic needs.

I can create another PR exploring functional `ListOptions` if that seems useful.

DirectXMan12:
> In general, the benefit of functional options is that they're invisible until you need them.

Agreed, I do like that aspect.

> I can create another PR

yeah, I think we should have a common experience.  Please do create a separate PR for varargs list options.

droot:
@grantr will wait for your second PR to compare how it looks.

Also, another major factor to consider is that the changes should be backward compatible. Implementation in current PR meets that criterion.

grantr:
> I would rewrite the tests to capture desired behavior from a user perspective though, instead of being broken out into the smallest possible units.

The list you propose is all the existing tests less one! 😜I folded the `metav1.DeleteOptions` test into the others and refactored a bit. Take a look, let me know what you think.

DirectXMan12:
tests look good.  will do one more pass tonight or tomorrow morning, then merge.

DirectXMan12:
whoops, looks like you missed a fake implementation in one of the tests.

DirectXMan12:
Still *technically* a incompatible change then, I think.  We'll do it once everything's migrated to the pinned deps.

grantr:
Can this change also go forward with https://github.com/kubernetes-sigs/controller-runtime/pull/106#issuecomment-418516452?

DirectXMan12:
Yep, this can go forward now.

DirectXMan12:
Fix golint issues (some missing godocs), and this is good to go.

grantr:
Fixed the golint issues and rebased into a single commit.

grantr:
Looks like the gosec check is failing, but it doesn't seem to be anything related to the changes in this PR.

grantr:
All good now.

PR #95, opened 2018-07-27
Support Patch request in client

grantr:
Patch is similar to Update but submits the patch data in the request body instead of the object content. Patch data is accompanied by a patch type specifying the patch strategy, e.g. `types.MergePatchType`.

A subresource can be patched by passing the subresource path as variadic string arguments.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -122,6 +123,22 @@ func (c *client) Delete(ctx context.Context, obj runtime.Object) error {
 		Name(o.GetName()).
 		Do().
 		Error()
+}
+
+// Patch implements client.Client
+func (c *client) Patch(ctx context.Context, pt types.PatchType, data []byte, obj runtime.Object, subresources ...string) error {
+	o, err := c.cache.getObjMeta(obj)
+	if err != nil {
+		return err
+	}
+	return o.Patch(pt).
+		NamespaceIfScoped(o.GetNamespace(), o.isNamespaced()).
+		Resource(o.resource()).
+		SubResource(subresources...).
+		Name(o.GetName()).
+		Body(data).
+		Do().
+		Into(obj)
 }
 
 // Get implements client.Client
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -18,6 +18,7 @@ package client_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sync/atomic"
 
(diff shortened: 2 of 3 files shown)


DirectXMan12:
I'd like to see this accompanied by (or at least with a plan for) how we'd make it easy for users to use this.  Building patches isn't necessarily obvious or straightforward, and the difference between the different types isn't always clear to users.

I think we want to try and make sure interfaces that we expose like this are useful for people who aren't a-priori versed in how patch works.

Additionally, it might be worth thinking about how this signature would/could/should change when server-side apply merges.

DirectXMan12:
(I also think we want a better story in general for acting on subresources, so it might be worth having a broader discussion about that @pwittrock -- once you hit polymorphic subresources, it gets weird).  I don't have any immediate thoughts off the top of my head -- need to spend some time mulling it over.

grantr:
> I think we want to try and make sure interfaces that we expose like this are useful for people who aren't a-priori versed in how patch works.

:+1: on the intent. In the meantime, could there be a separate client interface that implements methods like this? This gives an escape hatch to users who need it, and makes it clear that this interface doesn't offer the same compatibility guarantees as the "blessed" methods.

```go
type ExperimentalClient interface{
  Patch(...)
  UpdateStatus(...)
  Watch(...)
}
client.Experimental().Patch(...)
```

I haven't thought about it too carefully, but maybe a nice Patch interface could take two objects, the original and the update, and generate a patch internally based on the object diff (and optional filters restricting the patch scope).

```go
client.Patch(ctx, objA, objAPrime)
client.Patch(ctx, objA, objAPrime, client.PatchFilter("metadata", "status"))
```

Not sure how patch types would integrate, but maybe those are going away with server-side apply.

droot:
Good discussion and very timely because I am starting work on adding support for updating "Status" subresource :) because this is turning out to be a very common ask.
`Status` seems to be a very simple one and fits nicely in our current structure
```bash
  err := client.UpdateStatus(ctx, object)
```

Re: `Patch`
I am also trying to understand the use of `Patch` over `Update` from best practices pt. of view. @grantr what is your primary motivation in your use-case ? I concur with @DirectXMan12 (quoting below) and would like steer `controller` authors towards that approach. 
> This already exists in another sense -- you can always manually construct a generated client (not trying to be cheeky here). Patch is mostly a performance optimization -- you avoid the potential for conflicts causing requeues, but it doesn't gain you much else in controller code, since you can always requeue if an update fails.

Initially when we started with controller-runtime, main goal for `client` pkg was to implement an interface which serves common needs of `controller` authors instead of becoming a general purpose kubernetes client. For example, `Watch` method is never be needed in the `client` interface because higher abstractions in controller-runtime eliminates the need for that. I think having clarity on goals/non-goals for the `client` interface will help with some of these discussions ?

grantr:
> @grantr what is your primary motivation in your use-case ?

A controller in https://github.com/knative/serving is using Patch to add a label to a resource it doesn't control. As @DirectXMan12 states, this is a performance optimization, but a useful one as judged by the knative/serving authors. if Patch fit naturally into controller-runtime that would be one less reason not to adopt it.

I understand the counter-arguments, and I'm fine with the answer being "use a normal client-go client" as long as there's a reasonably straightforward way to create/retrieve a client-go client using the controller-runtime config. Documentation to this effect would also be useful. Similar to #88, it should be clear to the user when additional capabilities exist (with greater responsibility).

> I think having clarity on goals/non-goals for the client interface will help with some of these discussions ?

Yes definitely!

DirectXMan12:
let's start adding doc issues for things like this -- we should have a section of "so you're an advanced user that needs an escape hatch until the functionality lands in controller-runtime (if ever)"

Adracus:
Any update on this? I guess patching really is something that is very useful to a broad user base.
I'd be up to continue the work on this PR / file a new one / discuss here so we can get this / similar functionality in.

grantr:
@Adracus I'm not working on this currently but I do think it's a useful feature, especially combined with the recently added unstructured object support.

DirectXMan12:
Ultimately, what we're probably going to want to do here is have a `client.Apply` that uses server-side patch when available, and generates a patch otherwise.  That probably covers most use case for patch

Adracus:
Alright, I'll bring this PR back to speed again :)

adracus:
PR is at https://github.com/kubernetes-sigs/controller-runtime/pull/235

PR #96, opened 2018-07-30
client: update status subresource support

droot:
Sending this (WIP) PR for an early feedback on the interface before I add tests.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -154,3 +154,36 @@ func (c *client) List(ctx context.Context, opts *ListOptions, obj runtime.Object
 		Do().
 		Into(obj)
 }
+
+// Status implements client.StatusClient
+func (c *client) Status() StatusWriter {
+	return &statusWriter{client: c}
+}
+
+// statusWriter is client.StatusWriter that writes status subresource
+type statusWriter struct {
+	client *client
+}
+
+// ensure statusWriter implements client.StatusWriter
+var _ StatusWriter = &statusWriter{}
+
+// Update implements client.StatusWriter
+func (sw *statusWriter) Update(_ context.Context, obj runtime.Object) error {
+	o, err := sw.client.cache.getObjMeta(obj)
+	if err != nil {
+		return err
+	}
+	// TODO(droot): examine the returned error and check if it error needs to be
+	// wrapped to improve the UX ?
+	// It will be nice to receive an error saying the object doesn't implement
+	// status subresource and check CRD definition
+	return o.Put().
+		NamespaceIfScoped(o.GetNamespace(), o.isNamespaced()).
+		Resource(o.resource()).
+		Name(o.GetName()).
+		SubResource("status").
+		Body(obj).
+		Do().
+		Into(obj)
+}
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -145,7 +145,7 @@ var _ = Describe("Client", func() {
 			close(done)
 		})
 
-		It("should use the provided Mapper if provided", func() {
+		PIt("should use the provided Mapper if provided", func() {
 
 		})
 
@@ -269,7 +269,7 @@ var _ = Describe("Client", func() {
 			Expect(err.Error()).To(ContainSubstring("no kind is registered for the type"))
 		})
 
-		It("should fail if the GVK cannot be mapped to a Resource", func() {
+		PIt("should fail if the GVK cannot be mapped to a Resource", func() {
 			// TODO(seans3): implement these
 			// Example: ListOptions
 		})
@@ -360,11 +360,11 @@ var _ = Describe("Client", func() {
 			close(done)
 		})
 
-		It("should fail if the object does not pass server-side validation", func() {
+		PIt("should fail if the object does not pass server-side validation", func() {
 
 		})
 
-		It("should fail if the object doesn't have meta", func() {
+		PIt("should fail if the object doesn't have meta", func() {
 
 		})
 
@@ -388,7 +388,7 @@ var _ = Describe("Client", func() {
 			close(done)
 		})
 
-		It("should fail if the GVK cannot be mapped to a Resource", func() {
+		PIt("should fail if the GVK cannot be mapped to a Resource", func() {
 
 		})
 	})
@@ -497,11 +497,11 @@ var _ = Describe("Client", func() {
 			close(done)
 		})
 
-		It("should fail if the GVK cannot be mapped to a Resource", func() {
+		PIt("should fail if the GVK cannot be mapped to a Resource", func() {
 
 		})
 
-		It("should fail if an API does not implement Status subresource", func() {
+		PIt("should fail if an API does not implement Status subresource", func() {
 
 		})
 	})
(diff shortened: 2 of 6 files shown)


droot:
cc @grantr

grantr:
Seems fine to me. I'm curious what the plan is for other subresources. `Scale` is officially supported by Kubernetes apiserver, so will there be an `UpdateScale` function?

Intuitively, it seems fine to have a top-level `UpdateStatus` function (since this is used by almost every controller) and a separate "update any subresource" function that takes care of updating `Scale`. Though if you have the latter, why would you use the former?

DirectXMan12:
> so trying to fit all of them in some generic interface leads to a very poor UX.

Yeah, that's what I was struggling with.

The dynamic scale client isn't terribly complicated (compared to the typed clients), so we may just want to consider providing an easy handle to it.

droot:
>so ours would probably look like `client.GetScale(&kapi.Pod{}, scaleObj, reconcileKey)` (although the duplicate types there might be confusing).

Why do we need to provide `reconcileKey` when we are providing `&kapi.Pod{}` which contains namespace and name ?

>I'm also wary of having too many methods under the same "namespace" (interface), lest it get a bit confusing. We might want consider something like client.Get/Update/Delete/Whatever for the "normal" methods, and client.Scale(&kapi.Pod{}).Get(scaleInst, reconcileKey) for subresources. That way, if they ever introduce arbitrary CRUD-style subresource support (i.e. not things like exec and attach, but more like scale and status), we could have client.Subresource("mysubresource", &myapi.MyKind{}).Get(subresourceInst, reconcileKey) or what have you.

Same question why `reconcileKey` is needed ?

So going by the proposal above, Status subresource methods will look like:
 - Get: `client.Status(&kapi.Pod{}).Get(....)`
 - Update: `client.Status(&kapi.Pod).Update(....)`

Implementation in current PR looks like:
 - Get: didn't implement because client.Get retrieves Status as well
 - Update: `client.UpdateStatus(ctx, &kapi.Pod{})`

Implementation in the PR looks very intuitive because `Status` is special in the sense it accepts the API object itself as payload. I am sort of leaning towards special-casing `Status` and go with the simpler interface. WDYT ?

pkg/client/fake/client.go
```diff
@@ -123,6 +123,11 @@ func (c *fakeClient) Update(ctx context.Context, obj runtime.Object) error {
 	return c.tracker.Update(gvr, obj, accessor.GetNamespace())
 }
 
+func (c *fakeClient) UpdateStatus(ctx context.Context, obj runtime.Object) error {
+	// TODO(droot): implement for *real*
```
DirectXMan12: we should prob implement this before merging.

pkg/client/fake/client.go
```diff
@@ -123,6 +123,12 @@ func (c *fakeClient) Update(ctx context.Context, obj runtime.Object) error {
 	return c.tracker.Update(gvr, obj, accessor.GetNamespace())
 }
 
+func (c *fakeClient) UpdateStatus(ctx context.Context, obj runtime.Object) error {
+	// TODO(droot): This results in full update of the obj (spec + status). Need
+	// a way to update status field only.
+	return c.Update(ctx, obj)
```
droot: Couldn't figure out how to implement UpdateStatus behavior with tracker object. 
@mengqiy any ideas ?

pkg/client/fake/client.go
droot: I am calling `c.Update` now which will update the Status, but will also update the Spec part, which is wrong. 

Still looking for a way to do it with underneath tracker.

pkg/client/fake/client.go
mengqiy: I don't see an easier way except implementing it upstream.
It seems no one uses fakeclient's UpdateStatus, I think this is OK to be a starting point.

droot: Sounds good.

droot:
@DirectXMan12 added partial implementation of the UpdateStatus, should be ok for now. (fixing it completely will require re-doing the fake implementation)

> Personally would prefer .Status().Update(ctx, object) (gives us a bit more flexibility in the future, IMO).

Keeping flexibility aside for a moment, do you find `UpdateStatus(...)` is more intuitive over `client.Status().Update(...)` ?

droot:
@DirectXMan12 revised the interface/implementation as per our last conversation. PTAL.

pkg/client/client.go
```diff
@@ -154,3 +154,36 @@ func (c *client) List(ctx context.Context, opts *ListOptions, obj runtime.Object
...
+var _ StatusWriter = &statusWriter{}
+
+// Update implements client.StatusWriter
+func (sw *statusWriter) Update(_ context.Context, obj runtime.Object) error {
+	o, err := sw.client.cache.getObjMeta(obj)
+	if err != nil {
+		return err
+	}
+	// TODO(droot): examine the returned error and check if it error needs to be
+	// wrapped to improve the UX ?
```
DirectXMan12: we could always check discovery to see if the `/status` subresource appeared in discovery first, if we really wanted to...

pkg/client/client_test.go
```diff
@@ -391,6 +393,127 @@ var _ = Describe("Client", func() {
...
+			Expect(cl).NotTo(BeNil())
+
+			By("updating status of a non-existent object")
+			err = cl.Status().Update(context.TODO(), dep)
+			Expect(err).To(HaveOccurred())
+
+			close(done)
+		})
+
+		It("should fail if the object does not pass server-side validation", func() {
```
DirectXMan12: I don't think we need to explicitly test server-side behavior.  The implication is that we're using the status subresource here.

droot: +1, removed it.

pkg/client/client.go
droot: Thanks for the tip. Not critical for now, so not addressing this in this PR.

droot:
@DirectXMan12 addressed the comments. PTAL.

pkg/client/client_test.go
```diff
@@ -391,6 +393,119 @@ var _ = Describe("Client", func() {
...
+			By("updating status of the Deployment")
+			dep.Status.Replicas = 1
+			err = cl.Status().Update(context.TODO(), dep)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("no kind is registered for the type"))
+
+			close(done)
+		})
+
+		It("should fail if the GVK cannot be mapped to a Resource", func() {
```
DirectXMan12: you've still got a couple empty tests here.

droot: I wanted to have these placeholders as TODOs so that it can act as good-first-issues for new contributors. 

Lets wait for today, I might be able to add these later today.

DirectXMan12: Please mark them as pending then -- http://onsi.github.io/ginkgo/#pending-specs.

droot:
@DirectXMan12 converted all TODO tests to pending specs (Thanks for the tip). PTAL whenever you get a chance.

DirectXMan12:
/lgtm
/approve

Let's try and fill in the gaps in the tests in a follow-up PR shortly

PR #98, opened 2018-07-31
Add CreateOrUpdate utility method

### LINKED ISSUE #85: [feature request] Reduce dependant resrouce creation boilerplate
Currently, in order to create a dependent resource (eg. a deployment) the code always looks like:
```go
deploy := &appsv1.Deployment{
        Name:      instance.Name + "-deployment",
        Namespace: instance.Namespace,
    },
    Spec: appsv1.DeploymentSpec{
       // ...
    },
}
if err := controllerutil.SetControllerReference(instance, deploy, r.scheme); err != nil {
    return reconcile.Result{}, err
}

// Check if the Deployment already exists
found := &appsv1.Deployment{}
err = r.Get(context.TODO(), types.NamespacedName{Name: deploy.Name, Namespace: deploy.Namespace}, found)
if err != nil && errors.IsNotFound(err) {
    log.Printf("Creating Deployment %s/%s\n", deploy.Namespace, deploy.Name)
    err = r.Create(context.TODO(), deploy)
    if err != nil {
        return reconcile.Result{}, err
    }
} else if err != nil {
    return reconcile.Result{}, err
}

// Update the found object and write the result back if there are any changes
if !reflect.DeepEqual(deploy.Spec, found.Spec) {
    found.Spec = deploy.Spec
    log.Printf("Updating Deployment %s/%s\n", deploy.Namespace, deploy.Name)
    err = r.Update(context.TODO(), found)
    if err != nil {
        return reconcile.Result{}, err
    }
}
```

Which is about 50 lines of boilerplate.

What I've found very handy when working on controllers, were the `CreateOrPatch` methods from https://github.com/appscode/kutil.

Do you think that something similar can be implemented in the generic client?

calind:
Closes #85.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -17,11 +17,15 @@ limitations under the License.
 package controllerutil
 
 import (
+	"context"
 	"fmt"
+	"reflect"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
@@ -97,3 +101,78 @@ func referSameObject(a, b v1.OwnerReference) bool {
 
 	return aGV == bGV && a.Kind == b.Kind && a.Name == b.Name
 }
+
+// OperationResult is the action result of a CreateOrUpdate call
+type OperationResult string
+
+const ( // They should complete the sentence "Deployment default/foo has been ..."
+	// OperationResultNone means that the resource has not been changed
+	OperationResultNone OperationResult = "unchanged"
+	// OperationResultCreated means that a new resource is created
+	OperationResultCreated OperationResult = "created"
+	// OperationResultUpdated means that an existing resource is updated
+	OperationResultUpdated OperationResult = "updated"
+)
+
+// CreateOrUpdate creates or updates the given object obj in the Kubernetes
+// cluster. The object's desired state should be reconciled with the existing
+// state using the passed in ReconcileFn. obj must be a struct pointer so that
+// obj can be updated with the content returned by the Server.
+//
+// It returns the executed operation and an error.
+func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
+	// op is the operation we are going to attempt
+	op := OperationResultNone
+
+	// get the existing object meta
+	metaObj, ok := obj.(v1.Object)
+	if !ok {
+		return OperationResultNone, fmt.Errorf("%T does not implement metav1.Object interface", obj)
+	}
+
+	// retrieve the existing object
+	key := client.ObjectKey{
+		Name:      metaObj.GetName(),
+		Namespace: metaObj.GetNamespace(),
+	}
+	err := c.Get(ctx, key, obj)
+
+	// reconcile the existing object
+	existing := obj.DeepCopyObject()
+	existingObjMeta := existing.(v1.Object)
+	existingObjMeta.SetName(metaObj.GetName())
+	existingObjMeta.SetNamespace(metaObj.GetNamespace())
+
+	if e := f(obj); e != nil {
+		return OperationResultNone, e
+	}
+
+	if metaObj.GetName() != existingObjMeta.GetName() {
+		return OperationResultNone, fmt.Errorf("ReconcileFn cannot mutate objects name")
+	}
+
+	if metaObj.GetNamespace() != existingObjMeta.GetNamespace() {
+		return OperationResultNone, fmt.Errorf("ReconcileFn cannot mutate objects namespace")
+	}
+
+	if errors.IsNotFound(err) {
+		err = c.Create(ctx, obj)
+		op = OperationResultCreated
+	} else if err == nil {
+		if reflect.DeepEqual(existing, obj) {
+			return OperationResultNone, nil
+		}
+		err = c.Update(ctx, obj)
+		op = OperationResultUpdated
+	} else {
+		return OperationResultNone, err
+	}
+
+	if err != nil {
+		op = OperationResultNone
+	}
+	return op, err
+}
+
+// MutateFn is a function which mutates the existing object into it's desired state.
+type MutateFn func(existing runtime.Object) error
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -17,12 +17,18 @@ limitations under the License.
 package controllerutil_test
 
 import (
+	"context"
+	"fmt"
+	"math/rand"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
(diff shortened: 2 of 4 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -46,3 +50,56 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+				return nil, OperationNoop, err
+			}
+			return obj, OperationUpdated, err
+		}
+		return obj, OperationNoop, nil
+	}
+}
+
+// TransformFn is a function which take in a kubernetes object and returns the
+// desired state of that object
```
DirectXMan12: note whether or not it's allowed/expected to mutate the object.  It can be helpful when refactoring later.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -46,3 +50,56 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+const ( // They should complete the sentence "Deployment default/foo has been ..."
+	// OperationNoop means that the resource has not been changed
+	OperationNoop = "unchanged"
+	// OperationCreated means that a new resource has been created
+	OperationCreated = "created"
+	// OperationUpdated means that an existing resource has been updated
+	OperationUpdated = "updated"
+)
+
+// CreateOrUpdate creates or updates a kuberenes resource. It takes in a key and
```
DirectXMan12: typo: "kubernetes"

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -46,3 +50,56 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+// a placeholder for the existing object and returns the modified object
+func CreateOrUpdate(ctx context.Context, c client.Client, key client.ObjectKey, existing runtime.Object, t TransformFn) (runtime.Object, OperationType, error) {
```
DirectXMan12: our client methods tend not to return objects, and instead operate "in-place" on them.  This should do so as well

calind:
Hi @DirectXMan12 I've implemented your feedback and also wrote tests

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -46,3 +50,75 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+	OperationNoop = "unchanged"
+	// OperationCreated means that a new resource has been created
+	OperationCreated = "created"
+	// OperationUpdated means that an existing resource has been updated
+	OperationUpdated = "updated"
+)
+
+// CreateOrUpdate creates or updates a kubernetes resource. It takes in a key and
+// a placeholder for the existing object and returns the operation executed
+func CreateOrUpdate(ctx context.Context, c client.Client, key client.ObjectKey, existing runtime.Object, t TransformFn) (OperationType, error) {
```
DirectXMan12: is the purpose of `OperationType` just logging?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -46,3 +50,75 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+	err := c.Get(ctx, key, existing)
+	var obj runtime.Object
+
+	if errors.IsNotFound(err) {
+		// Create a new zero value object so that the in parameter of
+		// TransformFn is always a "clean" object, with only Name and Namespace
+		// set
+		zero := reflect.New(reflect.TypeOf(existing).Elem()).Interface()
```
DirectXMan12: this is what `existing` is for.  It should never be nil, so you shouldn't have to construct a new zero object.

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -44,10 +48,94 @@ var _ = Describe("Controllerutil", func() {
...
+		It("creates a new object if one doesn't exists", func() {
+			deplKey := types.NamespacedName{Name: "test-create", Namespace: "default"}
+			depl := &appsv1.Deployment{}
+
+			op, err := controllerutil.CreateOrUpdate(context.TODO(), c, deplKey, depl, createDeployment)
+
+			By("returning OperationCreated")
+			Expect(op).Should(BeEquivalentTo(controllerutil.OperationCreated))
+
+			By("returning returning no error")
```
DirectXMan12: typo: duplicate word

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -44,10 +48,94 @@ var _ = Describe("Controllerutil", func() {
...
+		It("update existing object", func() {
+			deplKey := types.NamespacedName{Name: "test-update", Namespace: "default"}
+			d, _ := createDeployment(&appsv1.Deployment{})
+			depl := d.(*appsv1.Deployment)
+			depl.Name = "test-update"
+			depl.Namespace = "default"
+
+			var scale int32 = 2
+
+			Expect(c.Create(context.TODO(), depl)).Should(Succeed())
```
DirectXMan12: style is general `Expect(X).To(Y)` (`Should` is reserved for the omega character, which we don't use)

pkg/controller/controllerutil/controllerutil.go
calind: The reason for creating a new zero value struct is that you can pass in existing an initialized one, eg. with `.spec` filled. I'm not sure that `client.Get` zeroes the object if it doesn't find one.

What I want to enforce here, is that if this function is going to create a new object, your transform callback always gets an empty object with only Name and Namespace filled in.

calind:
@DirectXMan12  what do you think about changing the signature to 
```go
CreateOrUpdate (c client.Client, ctx context.Context, obj runtime.Object, t ...TransformFn)
```

That way you can just pass the desired object and if you need to apply some changes you can define a set of transform functions.

The reason for keeping the transform function is that for example in case of Service another controller fills in the `ClusterIP` and you cannot reset it to null (you will get a validation error).

The `TransformFn` would change the signature to:

```go
type TransformFn func(existing, desired runtime.Object, op OperationType) (runtime.Object, error)
```

calind:
Also, do you thing that I should move this to a `client/clientutil` package?

DirectXMan12:
>  Also, do you thing that I should move this to a `client/clientutil` package

I think it's probably fine where it is.

>  what do you think about changing the signature to ...

I'm fine with the multiple transform functions.  I'm not entirely certain as to the purpose of passing the operation type to the transform function, though.  Same with passing existing and desired pointers -- ideally, the transformation should be idempotent, and should take an object, and turn it into the desired state.

I'd expect a signature of

```go
func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, transforms... TransformFunc) error
type TransformFunc func(obj runtime.Object) error
```

but it's possible I'm missing something here.

calind:
@DirectXMan12 I think I've made some logic mistakes in the previous commit so I rewrote it in 1431a57. Let me try to better illustrate the use cases for transform functions which are actually reconcile functions (I have renamed the type to better illustrate this to `ReconcileFn`).

Probably 90% of times you would use `CreateOrUpdate`  by just giving it a `desired` object. The role of reconcile functions is to cover some corner cases. 

For example if you pass in a `corev1.Service` as a desired object, when the object is created in apiserver, another controller will set it's `spec.clusterIP`. After is set, the field becomes immutable. A reconcile function solves this by copying `spec.clusterIP` from the existing object into the desired object.

So the `ReconcileFn` signature would become:
```go
type ReconcileFn func(desired, existing runtime.Object) error
```

As for the operation I've dropped it from the `ReconcileFn` and mainly it's purpose is for informing the caller what action have been taken (eg. for logging purposes). If you have any feedback on how this should be passed I'm open for changes :)

DirectXMan12:
aaaah, I see what you're saying with the reconcile func.  That makes sense, we just need to make sure that's blindingly obvious.

calind:
@DirectXMan12 after using it in real code, I'm having second thoughts about the current state of `ReconcileFn`:
```go
type ReconcileFn func(desired, existing runtime.Object) error
```
because I allways end up with code like:
```go
existing.DeepCopyInto(desired)
desired.replicas = 2
```
Ideally you want your reconcile function to just do:
```go
existing.replicas = 2
```
And that is because you always want to mutate the existing state towards a desired state.

So the proposal is to change the signatures to:
```go
// CreateOrUpdate a resrouce with the same type, name and namespace as obj 
// and recocile it's state using the passed in ReconcileFn.
// It returns the executed operation and an error.
func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, r ReconcileFn) (OperationType, error) {}

// ReconcileFn is a function which mutates the existing object into it's desired state. 
// If there is no existing object, a zero value object is passed in.
type ReconcileFn func(existing runtime.Object) error
```

Here is an example which always creates/updates a deployment with the scale of 2:
```go
func scale(existing runtime.Object) error {
    var replicas int32 = 2
    out := existing.(*appsv1.Deployment)
    existing.Sepc.Replicas = &replicas
}
CreateOrUpdate(context.TODO(), c, &appsv1.Deployment{Name: "foo", Namespace: "bar"}, scale)
```

DirectXMan12:
So, we just need to be clear on the intended usage in the comments -- the `ReconcileFunc` should actually be doing the work -- it's not just reconciling the differences any more.  In the original proposal,
it seemed like you'd do work on the passed in object before calling `CreateOrUpdate`, and then reconcile the differences if it already exists.

calind:
@DirectXMan12, I've pushed the changes. I've also added an example. It is ready for review.

calind:
I've rebased the changes and also found an issue when the `ReconcileFn` changes the name/namespace of the object.

@DirectXMan12 PTAL

DirectXMan12:
hey, sorry, busy week.  let me try and take a look today.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -97,3 +101,74 @@ func referSameObject(a, b v1.OwnerReference) bool {
...
+	// OperationUpdate means that an existing resource is updated
+	OperationUpdate = "updated"
+)
+
+// CreateOrUpdate a resource with the same type, name and namespace as obj
+// and reconcile it's state using the passed in ReconcileFn.
+// It returns the executed operation and an error.
+func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, r ReconcileFn) (OperationType, error) {
+	// op is the operation we are going to attempt
+	var op OperationType = OperationNoop
```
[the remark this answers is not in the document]
calind: I think needs to be a way to surface the operation. For example, I'm using it to emit events based of that, eg. `Successfully created Deployment default/foo`. Is there a better way of doing it?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -97,3 +101,74 @@ func referSameObject(a, b v1.OwnerReference) bool {
...
+	// get the existing object
+	mo, ok := obj.(v1.Object)
+	if !ok {
+		return OperationNoop, fmt.Errorf("%T does not implement metav1.Object interface", obj)
+	}
+	key := client.ObjectKey{
+		Name:      mo.GetName(),
+		Namespace: mo.GetNamespace(),
+	}
+	existing := reflect.New(reflect.TypeOf(obj).Elem()).Interface().(runtime.Object)
```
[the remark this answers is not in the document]
calind: update: after more reading I see that `client.Create` and `client.Update` mutate the `obj`. I'll update the PR accordingly.

calind:
@DirectXMan12 PTAL.

I've implemented the feedback regarding `obj` parameter. I've also kept `OperationType` in return as a way to surface the operation. I'm open to suggestions on better ways to do this. 

For example I'm using the `OperationType` to send events, like `Successfully created Deployment default/foo` when creating/updating a dependent deployment, logging being the other obvious use case.

DirectXMan12:
I think this looks good.  Let's have a quick peek from @droot and then we'll merge.

droot:
@DirectXMan12 @calind looking at it.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -97,3 +101,78 @@ func referSameObject(a, b v1.OwnerReference) bool {
...
+	}
+
+	if errors.IsNotFound(err) {
+		err = c.Create(ctx, obj)
+		op = OperationCreate
+	} else if err == nil {
+		if reflect.DeepEqual(existing, obj) {
+			return OperationNoop, nil
+		}
+		err = c.Update(ctx, obj)
```
droot: Trying to analyze the semantics of this wrt status sub-resource. If an API implements `status` as sub-resource (which is the recommended approach), then `Update` is updating just the `Spec` part of API (Then the `DeepEqual` should be of the spec's field only ?) @DirectXMan12

calind: @droot  `obj` being a `runtime.Object`, is it possible to get the spec of that? If so what should happen with types not having a `Spec` field?

calind:
@droot PTAL

droot:
@calind looks good to me. Can you pl. squash the commits ?

calind:
@droot done

blaggacao:
@calind Is it correct to assume that I would pass on the _desired_ state object as `runtime.Object` to the `CreateOrUpdate` function?
Like this:
```golang
	// Fetch the OdooCluster instance
	instance := &clusterv1beta1.OdooCluster{}
	// Define the desired DBNamespace object
	objDBNamespace := &clusterv1beta1.DBNamespace{
			Name:      instance.Name + "-dbnamespace",
			Namespace: instance.Namespace,
		},
		Spec: instance.Spec.DBSpec,
	}

	if err := controllerutil.SetControllerReference(instance, objDBNamespace, r.scheme); err != nil {
		return reconcile.Result{}, err
	}

	_, err := controllerutil.CreateOrUpdate(context.TODO(), r, objDBNamespace, scale)
       // Still don't know how to properly handle _ but doesn't matter, here.

```

calind:
The object you pass in is the _receiver_ object, aka the result of `client.Create` or `client.Update`. The desired state must be reconciled in the `MutateFn` (aka. the last parameter) like this:

```go
// Define the object name/namespace
objDBNamespace := &clusterv1beta1.DBNamespace{
		Name:      instance.Name + "-dbnamespace",
		Namespace: instance.Namespace,
	},
}

result, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, objDBNamespace, func (existing runtime.Object) error{
	// mutate here the state of existing object to the desired state
	// note that at this point _objDBNamespace_ and _existing_ point to the same struct
	out := existing.(*clusterv1beta1.DBNamespace)

	out.Spec = instance.Spec.DBSpec  // Desired state

	if err := controllerutil.SetControllerReference(instance, out, r.scheme); err != nil {
		return err
	}

	return nil
})

if err != nil {
	return reconcile.Result{}, err
}
// Your objDBNamespace now contains the object from the culster after Create/Update
// see: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/client#Writer
// Result contains the applied operation: Create/Update/NoOp
log.Info("object reconciled", "operation", result)

```

blaggacao:
Thanks, got it! The inline function idiom is a nice idea to keep logic linear to read!

blaggacao:
@calind This commit enabled a beautiful pattern https://github.com/xoe-labs/odoo-operator/blob/2a0248030b53ad82b99c2c229fdd1db3f3ae38aa/pkg/controller/odoocluster/odoocluster_controller.go#L212-L459
I guess I still have to take care of become-immutable fields manually in the mutate function, right?

calind:
Cool, glad to hear that! You can verify in the in the mutate function that `out.ObjectMeta.CreationTimestamp.IsZero()` to check that an object is going to be created or updated.

PR #99, opened 2018-07-31
Fix duplicate controller references

### LINKED ISSUE #68: controllerutil.SetOwnersReference should do nothing if the reference already exists

calind:
Fixes #68

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -25,9 +25,30 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+// AlreadyOwnedError is an error returned if the object you are trying to assign
+// a controller reference is already owned by another controller Object is the
+// subject and Owner is the reference for the current owner
+type AlreadyOwnedError struct {
+	Object v1.Object
+	Owner  v1.OwnerReference
+}
+
+func (e *AlreadyOwnedError) Error() string {
+	return fmt.Sprintf("Object %s/%s is already owned by another %s controller %s", e.Object.GetNamespace(), e.Object.GetName(), e.Owner.Kind, e.Owner.Name)
+}
+
+func newAlreadyOwnedError(Object v1.Object, Owner v1.OwnerReference) *AlreadyOwnedError {
+	return &AlreadyOwnedError{
+		Object: Object,
+		Owner:  Owner,
+	}
+}
+
 // SetControllerReference sets owner as a Controller OwnerReference on owned.
 // This is used for garbage collection of the owned object and for
 // reconciling the owner object on changes to owned (with a Watch + EnqueueRequestForOwner).
+// Since only one OwnerReference can be a controller, it returns an error if
+// there is another OwnerReference with Controller flag set.
 func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) error {
 	ro, ok := owner.(runtime.Object)
 	if !ok {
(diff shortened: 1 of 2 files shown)


droot:
@calind Thanks for the PR. Looks good, Can we add a few tests to verify the behavior.

calind:
Will do. Can you take a look at #100 as well since the tests fail because of that.

Thanks!

calind:
@droot, since only one `OwnerReference` can have `Controller=true` what should be the semantics of this function?

1. Return an error is another `OwnerReference` is a controller
2. Remove the `Controller=true` flag from existing `OwnerReference`
3. Remove any `OwnerReference` with `Controller=true`

I think that first option is the one to go.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -28,6 +28,8 @@ import (
 // SetControllerReference sets owner as a Controller OwnerReference on owned.
 // This is used for garbage collection of the owned object and for
 // reconciling the owner object on changes to owned (with a Watch + EnqueueRequestForOwner).
+// Since only one OwnerRecerence can be a controller, it returns an error if
```
droot: s/OwnerRecerence/OwnerReference

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -39,10 +41,30 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
 
-	// Add it to the child
-	object.SetOwnerReferences(append(object.GetOwnerReferences(), ref))
+	existingRefs := object.GetOwnerReferences()
+	fi := -1
+	for i, r := range existingRefs {
+		if r.Kind == gvk.Kind && r.Name == owner.GetName() && r.UID == owner.GetUID() {
+			fi = i
+		} else if r.Controller != nil && *r.Controller {
+			return fmt.Errorf("Object %s/%s is already owned by another %s controller %s", object.GetNamespace(), object.GetName(), r.Kind, r.Name)
```
droot: May be we should consider creating an error-type for this error because users would like to detect this error-type and kick-in some business logic for adopting of these objects (re-parenting to new owner). @DirectXMan12 what do you think ?

calind:
@droot implemented the feedback, also cleared out #100 and now travis tests are passing.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -25,9 +25,31 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+// AlreadyOwnedError is an error returned if the object you are trying to assign
+// an controller reference is already owned by another controller
```
droot: s/an/a

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -25,9 +25,31 @@ import (
...
+	Object v1.Object
+	Owner  v1.OwnerReference
+}
+
+func (e *AlreadyOwnedError) Error() string {
+	return fmt.Sprintf("Object %s/%s is already owned by another %s controller %s", e.Object.GetNamespace(), e.Object.GetName(), e.Owner.Kind, e.Owner.Name)
+}
+
+// NewAlreadyOwnedError creates a new AlreadyOwnedError
+func NewAlreadyOwnedError(Object v1.Object, Owner v1.OwnerReference) *AlreadyOwnedError {
```
droot: This fn doesn't need to be exported so pl. lowercase it.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -42,7 +64,37 @@ func SetControllerReference(owner, object v1.Object, scheme *runtime.Scheme) err
...
+		existingRefs[fi] = ref
+	}
+
+	// Update owner references
+	object.SetOwnerReferences(existingRefs)
 	return nil
 }
+
+// Returns true if a and b point to the same object
+func sameOwnerReference(a, b v1.OwnerReference) bool {
```
droot: Since its comparing the GVKs, better name should be `gvkEquals(a,b)`. Current name reads as `a`, `b` has same owner ?

I wonder if something like this exists the apimachinery already ?

droot:
@calind This is good to go. Can you pl. squash the commits ?. Thank you so much for this PR.

calind:
@droot done. Thanks for the feedback.

Issue #33, opened 2018-06-13
Support pkg/client operations for Unstructured objects


pwittrock:

shawn-hurley:
@pwittrock I believe #101 Should fix this issues. WDYT?

DirectXMan12:
fixed a while ago

PR #101, opened 2018-07-31
Adding ability for clients, cache and watcher to work with unstructured

### LINKED ISSUE #17: Support source.Kind for unstructured objects
We should be able to watch for objects against unstructured objects from source.Kind.

- Make the client work
- Make the cache work
- Make watch work

shawn-hurley:
Attempt to fix #17 and I think it also addresses #33 

/cc @DirectXMan12 @pwittrock can you PTAL and let me know if I am even on the right track?

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/cache"
@@ -73,13 +74,25 @@ func (ip *informerCache) List(ctx context.Context, opts *client.ListOptions, out
 	}
 	// we need the non-list GVK, so chop off the "List" from the end of the kind
 	gvk.Kind = gvk.Kind[:len(gvk.Kind)-4]
-
-	// http://knowyourmeme.com/memes/this-is-fine
-	elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
-	cacheTypeValue := reflect.Zero(reflect.PtrTo(elemType))
-	cacheTypeObj, ok := cacheTypeValue.Interface().(runtime.Object)
-	if !ok {
-		return fmt.Errorf("cannot get cache for %T, its element %T is not a runtime.Object", out, cacheTypeValue.Interface())
+	_, isUnstructured := out.(*unstructured.UnstructuredList)
+	var cacheTypeObj runtime.Object
+	if isUnstructured {
+		u := &unstructured.Unstructured{}
+		u.SetGroupVersionKind(gvk)
+		cacheTypeObj = u
+	} else {
+		itemsPtr, err := apimeta.GetItemsPtr(out)
+		if err != nil {
+			return nil
+		}
+		// http://knowyourmeme.com/memes/this-is-fine
+		elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
+		cacheTypeValue := reflect.Zero(reflect.PtrTo(elemType))
+		var ok bool
+		cacheTypeObj, ok = cacheTypeValue.Interface().(runtime.Object)
+		if !ok {
+			return fmt.Errorf("cannot get cache for %T, its element %T is not a runtime.Object", out, cacheTypeValue.Interface())
+		}
 	}
 
 	cache, err := ip.InformersMap.Get(gvk, cacheTypeObj)
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -59,11 +59,6 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 
 // List implements Reader
 func (ip *informerCache) List(ctx context.Context, opts *client.ListOptions, out runtime.Object) error {
-	itemsPtr, err := apimeta.GetItemsPtr(out)
-	if err != nil {
-		return nil
-	}
-
 	gvk, err := apiutil.GVKForObject(out, ip.Scheme)
 	if err != nil {
 		return err
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -158,6 +158,8 @@ func (ip *InformersMap) getMapEntry(gvk schema.GroupVersionKind, isUnstructured 
 // the Informer from the map.
 func (ip *InformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Object) (*MapEntry, error) {
 	_, isUnstructured := obj.(*unstructured.Unstructured)
+	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
+	isUnstructured = isUnstructured || isUnstructuredList
 	// Return the informer if it is found
 	i, ok := ip.getMapEntry(gvk, isUnstructured)
 	if ok {
@@ -199,7 +201,7 @@ func (ip *InformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Object) (*M
 			Informer: ni,
 			Reader:   CacheReader{indexer: ni.GetIndexer(), groupVersionKind: gvk},
 		}
-		ip.informersByGVK[gvk] = i
+		ip.setMap(i, gvk, isUnstructured)
 
 		// Start the Informer if need by
 		// TODO(seans): write thorough tests and document what happens here - can you add indexers?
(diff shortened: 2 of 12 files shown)


pkg/cache/internal/cache_reader.go
```diff
@@ -73,6 +75,21 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
 	// TODO(directxman12): revisit the decision to always deepcopy
 	obj = obj.(runtime.Object).DeepCopyObject()
 
+	// If out is supposed to be unstructured handle correctly.
+	if o, ok := out.(*unstructured.Unstructured); ok {
```
DirectXMan12: I'm not sure I'm following the logic behind the marshalling and unmarshalling... shouldn't the object already be an unstructured?  We should just be able to do the reflection hack...

pkg/client/apiutil/apimachinery.go
```diff
@@ -64,3 +66,30 @@ func RESTClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config, code
...
+	var jsonInfo runtime.SerializerInfo
+	for _, info := range scheme.Codecs.SupportedMediaTypes() {
+		if info.MediaType == runtime.ContentTypeJSON {
+			jsonInfo = info
+			break
+		}
+	}
+	jsonInfo.Serializer = unstructured.UnstructuredJSONScheme
+
+	cfg.NegotiatedSerializer = serializer.NegotiatedSerializerWrapper(jsonInfo)
```
DirectXMan12: the only thing that changes here is the negotiated serializer setup, right?  Refactor the common bits out into a private function so we're not copying and pasting code around.

DirectXMan12: Also, do you have to care about status, like the old dynamic client did?

pkg/client/client_cache.go
```diff
@@ -46,7 +47,10 @@ type clientCache struct {
 
 	// resourceByType caches type metadata
 	resourceByType map[reflect.Type]*resourceMeta
-	mu             sync.RWMutex
+	// resourceByGVK caches type metadat for unstructured
```
DirectXMan12: typo: "metadata"

DirectXMan12: also, call this unstructuredResourceByGVK, so it's clear why it's separate

pkg/client/client_cache.go
```diff
@@ -73,24 +77,44 @@ func (c *clientCache) newResource(obj runtime.Object) (*resourceMeta, error) {
 	return &resourceMeta{Interface: client, mapping: mapping, gvk: gvk}, nil
 }
 
-// getResource returns the resource meta information for the given type of object.
-// If the object is a list, the resource represents the item's type instead.
-func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
+func (c *clientCache) handleResourceByGVK(obj runtime.Object) (*resourceMeta, error) {
```
DirectXMan12: leave these as `get` -- it's a bit clearer, IMO

DirectXMan12:
So, I've been thinking about this a bit more, and I want to make sure we're not introducing a foot-gun with (specifically) the unstructured cache part.  Client I'm totally fine with.

I'm curious as to the use case for unstructured cache.  The normal case for unstructured client is something like "I want to access a bunch of object regardless of the kind".  But if you do that with the caching client, you'll get a new watch that you *might* not want (you might accidentally end up caching the entire cluster, for instance, if you're running something like GC).  You might even get a duplicate cache (one for the concrete object, one for the unstructured).

We should careful here...

shawn-hurley:
@DirectXMan12 and I discussed offline the problems that were identified. I will be re-working this based on the suggestions and help.

shawn-hurley:
@DirectXMan12 @droot @pwittrock PTAL I think this is ready for review

pkg/client/apiutil/apimachinery.go
```diff
@@ -58,9 +88,36 @@ func RESTClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config, code
...
 	}
-	return rest.RESTClientFor(cfg)
+	return cfg
+
+}
+
+//Copied from deprecated-dynamic/bad_debt.go
+// dynamicCodec is a codec that wraps the standard unstructured codec
+// with special handling for Status objects.
+// Deprecated only used by test code and its wrong
```
DirectXMan12: we should have a TODO to figure out *why* this is bad...

droot:
@shawn-hurley will take a look today. Thanks

pkg/cache/internal/informers_map.go
```diff
@@ -228,3 +237,38 @@ func (ip *InformersMap) newListWatch(gvk schema.GroupVersionKind) (*cache.ListWa
 		},
 	}, nil
 }
+
+// createUnstructuredClient is a ClientCreatorFunc for use with structured
```
droot: s/createUnstructuredClient/createStructuredClient

pkg/client/apiutil/apimachinery.go
```diff
@@ -49,6 +56,29 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
...
+func RESTUnstructuredClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config) (rest.Interface, error) {
+	cfg := createRestConfig(gvk, baseConfig)
+	var jsonInfo runtime.SerializerInfo
+	for _, info := range scheme.Codecs.SupportedMediaTypes() {
+		if info.MediaType == runtime.ContentTypeJSON {
+			jsonInfo = info
+			break
+		}
+	}
+	jsonInfo.Serializer = dynamicCodec{}
```
droot: comment below about dynamicCodec doesn't boost my confidence. I thought initially it will be used in test code or so, but I see it is being used here in *real* code.

pkg/client/apiutil/apimachinery.go
```diff
@@ -58,9 +88,36 @@ func RESTClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config, code
...
+// Deprecated only used by test code and its wrong
+type dynamicCodec struct{}
+
+func (dynamicCodec) Decode(data []byte, gvk *schema.GroupVersionKind, obj runtime.Object) (runtime.Object, *schema.GroupVersionKind, error) {
+	obj, gvk, err := unstructured.UnstructuredJSONScheme.Decode(data, gvk, obj)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if _, ok := obj.(*metav1.Status); !ok && strings.ToLower(gvk.Kind) == "status" {
```
droot: interesting we have `Status` resource as well ?

pkg/client/client_cache.go
```diff
@@ -46,12 +47,15 @@ type clientCache struct {
 
 	// resourceByType caches type metadata
 	resourceByType map[reflect.Type]*resourceMeta
-	mu             sync.RWMutex
+	// resourceByGVK caches type metadata for unstructured
+	unstructuredResourceByGVK map[schema.GroupVersionKind]*resourceMeta
+	muByType                  sync.RWMutex
+	muByGVK                   sync.RWMutex
```
droot: consider moving these above the fields they guard.

pkg/cache/informer_cache.go
```diff
@@ -58,11 +59,6 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 
 // List implements Reader
 func (ip *informerCache) List(ctx context.Context, opts *client.ListOptions, out runtime.Object) error {
-	itemsPtr, err := apimeta.GetItemsPtr(out)
-	if err != nil {
-		return nil
-	}
-
 	gvk, err := apiutil.GVKForObject(out, ip.Scheme)
```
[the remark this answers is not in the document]
shawn-hurley: I don't think that it does. unstructured does conform to the runtime.Object interface and has the methods to retrieve the GVK from that interface at least that is my understanding. 

Is there an edge case that I am not considering?

pkg/client/apiutil/apimachinery.go
shawn-hurley: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Status

I believe this is used for errors from the API servers but I don't know for a fact.

DirectXMan12: correct, the `Status` Kind is used to convey API errors.  It doesn't belong to any particular resource, IIRC.

shawn-hurley:
@DirectXMan12 and I had an offline discussion about the dynamic client. 

In the changes, we discussed using the new dynamic client for the cache's list watcher setup. We also talked about a separate client for the unstructured client. This new unstructured client would use the new dynamic client. 

@droot Wondering if the above would help with the concerns about the client?

pkg/cache/informer_cache.go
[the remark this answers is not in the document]
droot: Got it. No, was just curious.

pkg/client/apiutil/apimachinery.go
droot: Ack.

droot:
@shawn-hurley ignore my comment about the tests :) I completely missed them (they were probably collapsed)

shawn-hurley:
Updated to the new dynamic client for the unstructured clients.

PTAL @DirectXMan12 @droot

DirectXMan12:
For what this does, I think it came out pretty sane and readable. Please squash down the commits into logical chunks, and then I think it looks good from my end.  We'll want equivalent docs in the gitbook about unstructured as well.

shawn-hurley:
@DirectXMan12 I tried to make it more clear for each commit. PTAL

DirectXMan12:
better -- generally, commits should not show review or development history -- so for instance, the "refactor structure out of..." commit should probably be squashed into the first commit.  We *can* merge it like this, but I'd prefer to fix that kind of thing up first.

DirectXMan12:
Want to give @droot one more chance to look this over, but I think it looks good from my end.

pwittrock:
@droot will be back Monday to TAL

PR #107, opened 2018-08-10
Clarify SyncPeriod comment

grantr:
Removes the advice about when to change, since sync period is more subtle than can be explained in a comment. Replaces it with a warning to know what you're doing before changing it.

Followup from the conversation with @DirectXMan12 in #88.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -76,12 +76,10 @@ type Options struct {
 	// MapperProvider provides the rest mapper used to map go types to Kubernetes APIs
 	MapperProvider func(c *rest.Config) (meta.RESTMapper, error)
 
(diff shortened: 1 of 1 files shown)


spiffxp:
FYI this isn't picking up @DirectXMan12's /approve because the OWNERS_ALIASES file that his rights should be coming from has syntax errors

Compare [the sample](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners_aliases) to the [actual file used in this repo](https://github.com/kubernetes-sigs/controller-runtime/blob/master/OWNERS_ALIASES)

DirectXMan12:
Ok, fix merged, let's try again

/approve

PR #109, opened 2018-08-13
Update OWNERS_ALIASES to fix approver bot

DirectXMan12:
The OWNERS_ALIASES needs aliases as map keys, not as list items.


DirectXMan12:
@droot @pwittrock FY

PR #117, opened 2018-08-24
glog elimination round 2

DirectXMan12:
glog seems to have a way of creeping back in to our codebase.  If you see it in code review, point it out -- you should be using logr instead.


DirectXMan12:
@droot FYI

PR #123, opened 2018-08-28
add tests for webhook/admission pkg

mengqiy:
Fix several nil pointer issues when testing the unhappy path in the unit test.

### DIFF
--- a/pkg/webhook/admission/admission_suite_test.go
+++ b/pkg/webhook/admission/admission_suite_test.go
@@ -26,7 +26,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-func TestSource(t *testing.T) {
+func TestAdmissionWebhook(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecsWithDefaultAndCustomReporters(t, "application Suite", []Reporter{envtest.NewlineReporter{}})
 }
--- a/pkg/webhook/admission/admission_suite_test.go
+++ b/pkg/webhook/admission/admission_suite_test.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-   http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -14,15 +14,25 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package types
+package admission
 
-// WebhookType defines the type of a webhook
-type WebhookType int
+import (
+	"testing"
 
-const (
-	_ = iota
-	// WebhookTypeMutating represents mutating type webhook
-	WebhookTypeMutating WebhookType = iota
-	// WebhookTypeValidating represents validating type webhook
-	WebhookTypeValidating
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
+
+func TestSource(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "application Suite", []Reporter{envtest.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(logf.ZapLoggerTo(GinkgoWriter, true))
+
+	close(done)
+}, 60)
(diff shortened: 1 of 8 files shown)


pkg/webhook/admission/http_test.go
```diff
@@ -0,0 +1,191 @@
...
+			wh.ServeHTTP(w, req)
+			Expect(w.response).NotTo(BeNil())
+			Expect(w.response).To(Equal(expected))
+			Expect(h.invoked).To(BeTrue())
+			Expect(h.valueFromContext).To(Equal("bar"))
+		})
+	})
+})
+
+type fakeResponseWriter struct {
```
droot: Did you look at `httptest` pkg https://golang.org/pkg/net/http/httptest/ It provides some nice utilities for testing http handlers without running the server.

pkg/webhook/admission/webhook.go
```diff
@@ -130,37 +135,44 @@ func (w *Webhook) kvMapToContext(ctx context.Context) (context.Context, error) {
 }
 
 func (w *Webhook) handleMutating(ctx context.Context, req Request) Response {
-	var resp Response
-	var patches []jsonpatch.JsonPatchOperation
+	r := Response{
```
Liujingfang1: You can move this declaration to line 155

pkg/webhook/admission/admission_suite_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func TestSource(t *testing.T) {
```
droot: `TestSource` or we can call it something like `TestAdmissionWebhook()` ?

pkg/webhook/admission/decode_test.go
```diff
@@ -0,0 +1,81 @@
...
+    }
+}`),
+				},
+			},
+		}
+		BeforeEach(func(done Done) {
+			var err error
+			decoder, err = NewDecoder(scheme.Scheme)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(decoder).NotTo(BeNil())
```
droot: Do we want to create `decoder` in `BeforeEach` ? I think it can be shared for the tests below. WDYT ?

pkg/webhook/admission/http_test.go
```diff
@@ -0,0 +1,191 @@
...
+var _ = Describe("admission webhook http handler", func() {
+	Describe("empty request body", func() {
+		req := &http.Request{Body: nil}
+		wh := &Webhook{
+			Handlers: []Handler{},
+		}
+		w := &fakeResponseWriter{}
+		expected := []byte(`{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"request body is empty","code":400}}}
+`)
+		It("should return a response with an error", func() {
```
droot: `should return an error with bad-request status code` ?

pkg/webhook/admission/http_test.go
```diff
@@ -0,0 +1,191 @@
...
+			Header: http.Header{"Content-Type": []string{"application/foo"}},
+			Body:   nopCloser{Reader: bytes.NewBuffer(nil)},
+		}
+		wh := &Webhook{
+			Handlers: []Handler{},
+		}
+		w := &fakeResponseWriter{}
+		expected := []byte(`{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=application/foo, expect application/json","code":400}}}
+`)
+		It("should return a response with an error", func() {
```
droot: same as above.

pkg/webhook/admission/webhook_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+			It("should deny the request", func() {
+				expected := []byte(`{"response":{"uid":"","allowed":false}}
+`)
+				wh.ServeHTTP(w, req)
+				Expect(w.response).NotTo(BeNil())
+				Expect(w.response).To(Equal(expected))
+				Expect(alwaysDeny.invoked).To(BeTrue())
+				Expect(alwaysDeny.valueFromContext).To(Equal("bar"))
+				Expect(alwaysAllow.invoked).To(BeFalse())
```
Liujingfang1: So the handlers are executed with respect to their order in the `Handlers` slice. Is that something users need to be aware of? Do you need to add some doc for that?

pkg/webhook/admission/webhook_test.go
```diff
@@ -0,0 +1,379 @@
...
+						Response: &admissionv1beta1.AdmissionResponse{
+							Allowed:   true,
+							PatchType: func() *admissionv1beta1.PatchType { pt := admissionv1beta1.PatchTypeJSONPatch; return &pt }(),
+						},
+					}
+				},
+			}
+			wh := &Webhook{
+				Type:     types.WebhookTypeMutating,
+				Handlers: []Handler{patcher1, patcher2},
```
Liujingfang1: Do you check conflict among different handlers?

mengqiy: No. I can't find a easy way to do that.
I added a comment in the struct saying that user should ensure that their handlers should not mutate same field and generate conflicting JSON patches.

mengqiy:
PTAL
I will squash commits before merging.

PR #126, opened 2018-08-29
gas linter renamed to gosec

mengqiy:
This PR fixes an failing linter check due to renaming.

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -39,7 +39,7 @@ gometalinter.v2 --disable-all \
     --enable=errcheck \
     --enable=varcheck \
     --enable=goconst \
-    --enable=gas \
+    --enable=gosec \
     --enable=unparam \
     --enable=ineffassign \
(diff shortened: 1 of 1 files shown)


mengqiy:
It can be found failing in previous Travis run: https://travis-ci.org/kubernetes-sigs/controller-runtime/jobs/422318277

hack/verify.sh
```diff
@@ -39,7 +39,7 @@ gometalinter.v2 --disable-all \
     --enable=errcheck \
     --enable=varcheck \
     --enable=goconst \
-    --enable=gas \
+    --enable=gosec \
```
droot: Does it mean, contributors will have to update the metalinter ?

mengqiy: They don't have to.
The Travis has this problem is because it always get the latest one.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/.travis.yml#L18

PR #129, opened 2018-08-31
 Prompt to install CRD when Kind no found

interma:
Fix: kubernetes-sigs/kubebuilder#337

New output:
```
$ make run
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run ./cmd/manager/main.go
2018/08/31 10:40:04 Registering Components.
2018/08/31 10:40:04 must install CRD {core.hawq.org MyResource} before calling Start
exit status 1
make: *** [run] Error 1
```
Thanks for the review.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1129,7 +1129,7 @@
   revision = "c106140daf37a5d2b7a513726091a5eb8107e8b9"
 
 [[projects]]
-  digest = "1:b08a45828b09ce543aea649464635291107f53fd21465c0ab00a867b4418fa15"
+  digest = "1:5919b80013fe2204af836990cf255fc100d7d7cb077ec524c7518ff272e9a633"
   name = "k8s.io/client-go"
   packages = [
     "discovery",
@@ -1343,12 +1343,14 @@
     "k8s.io/apimachinery/pkg/runtime/serializer",
     "k8s.io/apimachinery/pkg/selection",
     "k8s.io/apimachinery/pkg/types",
+    "k8s.io/apimachinery/pkg/util/intstr",
     "k8s.io/apimachinery/pkg/util/runtime",
     "k8s.io/apimachinery/pkg/util/sets",
     "k8s.io/apimachinery/pkg/util/uuid",
     "k8s.io/apimachinery/pkg/util/wait",
     "k8s.io/apimachinery/pkg/watch",
     "k8s.io/client-go/discovery",
+    "k8s.io/client-go/dynamic",
     "k8s.io/client-go/kubernetes",
     "k8s.io/client-go/kubernetes/scheme",
     "k8s.io/client-go/kubernetes/typed/core/v1",
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -20,16 +20,20 @@ import (
 	"fmt"
 	"sync"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source/internal"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
+
+var log = logf.KBLog.WithName("source")
 
 const (
 	// defaultBufferSize is the default number of event notifications that can be buffered.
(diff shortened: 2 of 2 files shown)


pkg/source/source.go
```diff
@@ -80,7 +81,12 @@ func (ks *Kind) Start(handler handler.EventHandler, queue workqueue.RateLimiting
 	// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
 	i, err := ks.cache.GetInformer(ks.Type)
 	if err != nil {
-		return err
+		switch err.(type) {
+		case *meta.NoKindMatchError:
+			return fmt.Errorf("must install CRD %v before calling Start", err.(*meta.NoKindMatchError).GroupKind)
```
DirectXMan12: maybe use `%s` here instead to get it pretty-printed?

interma: Updated.
They are the same output.

DirectXMan12:
looks good, please squash the commits together

DirectXMan12:
one last thing before merge -- this message could potentially be confusing if the resource isn't a CRD, so we might want to change the wording a bit.  Also, we lose the ability to have users inspect the error themselves, which probably isn't the most useful, but might be worth considering...

Maybe we can just wrap the error so that people can still check, but provide our own error message?

interma:
@DirectXMan12 
Yes, wrap error is better and I modified error message:
```
$ make run
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run ./cmd/manager/main.go
2018/09/10 17:25:27 Registering Components.
2018/09/10 17:25:27 if {core.hawq.org MyResource} is a CRD, should install it before calling Start: no matches for kind "MyResource" in version "core.hawq.org/v1alpha1"
exit status 1
make: *** [run] Error 1
```

PTAL, thanks.

DirectXMan12:
@droot @pwittrock @mengqiy we should decide if we want something like `errors.Wrap` in our codebase -- it's not a terrible idea, but it's probably something we should explicitly decide.

In the mean time, if we want to move this forward, we might just want to log a warning at `V(1)` instead of modifying the error.   Otherwise, we can just wait to make that decision.

mengqiy:
> errors.Wrap

@DirectXMan12 I don't think it's a bad idea. What is your concern if we introduce this into our codebase?

Gopkg.lock
```diff
@@ -3,96 +3,75 @@
 
 [[projects]]
   branch = "default"
-  digest = "1:24df057f15e7a09e75c1241cbe6f6590fd3eac9804f1110b02efade3214f042d"
```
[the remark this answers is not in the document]
interma: Thanks @mengqiy You are right.

before:
```
$ dep version
dep:
 version     : devel
 build date  :
 git hash    :
 go version  : go1.10.3
 go compiler : gc
 platform    : darwin/amd64
```

after upgrade:
```
$ dep version
dep:
 version     : v0.5.0
 build date  : 2018-07-26
 git hash    : 224a564
 go version  : go1.10.3
 go compiler : gc
 platform    : darwin/amd64
 features    : ImportDuringSolve=false
```

mengqiy:
@interma It seems you have a conflict in `Gopkg.lock`.

interma:
Yes, @mengqiy 
Looks I need to rebase from master.
I will fix them after resigning my CLA as an employee.

Thanks for your review.

interma:
@mengqiy 
Rebased and no conflict now.

@DirectXMan12 @droot 
PTAL, thanks.

interma:
Thanks for your review. @droot 
Wait for @DirectXMan12  opinion.

DirectXMan12:
agree with @droot

interma:
Thanks @DirectXMan12 @droot 
A little confused:
You mean I don't need to wrap the error and just refine error message here (like I did before) 
or use another way to wrap error (but I don't know how to do it...) ?

droot:
>You mean I don't need to wrap the error and just refine error message here (like I did before)
or use another way to wrap error (but I don't know how to do it...) ?

Sorry I wasn't clear about the change:
I was suggesting to log the message at error level (so that users gets the issue) and avoiding the wrapping the error for now (because we don't want to lose the original error).

pkg/source/source.go
```diff
@@ -80,6 +84,11 @@ func (ks *Kind) Start(handler handler.EventHandler, queue workqueue.RateLimiting
 	// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
 	i, err := ks.cache.GetInformer(ks.Type)
 	if err != nil {
+		switch err.(type) {
```
DirectXMan12: no need to for the one item switch here --

```go
if kindMatchErr, ok := err.(*meta.NoKindMatchError); ok {
   log.Error(err, /* rest of message goes here */)
}
return err
```

interma: Thanks.

DirectXMan12:
/approve
/lgtm

thanks for your contribution, and for your patience!

PR #131, opened 2018-09-01
Support delaying requeue for objectkey.

pwittrock:
This is a requirement for migrating cluster-api to controller-runtime.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -209,6 +209,9 @@ func (c *Controller) processNextWorkItem() bool {
 		log.Error(err, "Reconciler error", "Controller", c.Name, "Request", req)
 
 		return false
+	} else if result.RequeueAfter > 0 {
+		c.Queue.AddAfter(req, result.RequeueAfter)
+		return true
 	} else if result.Requeue {
 		c.Queue.AddRateLimited(req)
 		return true
--- a/pkg/reconcile/reconcile.go
+++ b/pkg/reconcile/reconcile.go
@@ -17,6 +17,8 @@ limitations under the License.
 package reconcile
 
 import (
+	"time"
+
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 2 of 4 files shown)


DirectXMan12:
/lgtm

We should discuss adding some sort of metadata/emoji to PR titles for "will this cause a major/minor/patch release" to make it easier to visually check on release and/or automate.

DirectXMan12:
/lgtm cancel

Actually...

pkg/reconcile/reconcile.go
```diff
@@ -17,13 +17,18 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/types"
 )
 
 // Result contains the result of a Reconciler invocation.
 type Result struct {
 	// Requeue tells the Controller to requeue the reconcile key.  Defaults to false.
 	Requeue bool
+
+	// RequeueAfter is the duration to wait before adding the key back to the queue.
+	RequeueAfter *time.Duration
```
droot: We could avoid the pointer in this case by considering zero-value as indicator for not-to-requeue-after. I see that pointer introduces unnecessary verbosity in caller's code by having to declare a variable. WDYT ?

pkg/internal/controller/controller.go
```diff
@@ -210,8 +210,12 @@ func (c *Controller) processNextWorkItem() bool {
 
 		return false
 	} else if result.Requeue {
-		c.Queue.AddRateLimited(req)
-		return true
+		if result.RequeueAfter != nil {
+			c.Queue.AddAfter(req, *result.RequeueAfter)
```
DirectXMan12: Should we be forgetting with `AddAfter`?  This currently does...

pwittrock: Good catch.  Done

pkg/reconcile/reconcile.go
pwittrock: Done

pwittrock:
PTAL

PR #134, opened 2018-09-06
skip gosec linting for now

mengqiy:
gosec complains when we load from a file path which is a variable. Example error message:
```
pkg/envtest/crd.go:204::warning: Potential file inclusion via variable,MEDIUM,HIGH (gosec)
``` 
One possible way to make it work is ignoring this kind of error by running `gosec` as a custom linter and exclude G304 errors.

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -39,7 +39,6 @@ gometalinter.v2 --disable-all \
     --enable=errcheck \
     --enable=varcheck \
     --enable=goconst \
-    --enable=gosec \
     --enable=unparam \
     --enable=ineffassign \
     --enable=nakedret \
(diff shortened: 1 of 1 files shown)


pkg/envtest/crd.go
```diff
@@ -201,7 +201,7 @@ func readCRDs(path string) ([]*apiextensionsv1beta1.CustomResourceDefinition, er
 		}
 
 		// Unmarshal the file into a struct
-		b, err := ioutil.ReadFile(filepath.Join(path, file.Name()))
+		b, err := ioutil.ReadFile(filepath.Join(path, file.Name())) // nolint: gosec
```
droot: Can we configure the gosec linter to ignore these types or errors (or specific files) ? Annotating the source code to exclude is not right.

mengqiy:
I have tried to follow the document about how to [add an custom linter](https://github.com/alecthomas/gometalinter#adding-custom-linters).
My plan was to run gosec by ourself and exclude G304 errors.
But I can't really make it work. It seems it doesn't parse the linter output correctly and eats the error messages that it should emit.
I disable it for now. And I will try to fix it next week when get more time :)

PR #135, opened 2018-09-06
inject client and decoder for webhook

mengqiy:
inject client and decoder for webhook server and underlying webhook handlers


mengqiy:
It's strange that Tide (Prow) is missing in the presubmit :/

PR #136, opened 2018-09-09
Option to restrict informer cache to a namespace

### LINKED ISSUE #124: Allow restricting the Cache's ListWatch to be namespaced
#### Problem
Currently the ListWatch for the cache's informers are non-namespaced.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L218-L227

This means the Manager always requires cluster scoped permissions to work. While kubebuilder uses ClusterRole and ClusterRolebinding by default, that assumption isn't always true for an operator/controller (at least not in our context with the operator-sdk).

With just a Role and Rolebinding, the informers fail to list resources at the cluster scope.
```
E0828 23:41:19.472228       1 reflector.go:205] github.com/operator-framework/operator-sdk-samples/app-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:106: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:haseeb:default" cannot list pods at the cluster scope
E0828 23:41:20.141658       1 reflector.go:205] github.com/operator-framework/operator-sdk-samples/app-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:106: Failed to list *v1alpha1.App: apps.app.example.com is forbidden: User "system:serviceaccount:haseeb:default" cannot list apps.app.example.com at the cluster scope
```

#### Proposed Fix
Unless this is already supported or I've missed an easier way to do this, I've found that I can easily pipe down the namespace as an option from the Manager->Cache->InfromersMap->ListWatch.
```Go
mgr, err := manager.New(cfg, manager.Options{Namespace: namespace})
```
Possible fix: https://github.com/hasbro17/controller-runtime/commit/55894c2ce4676c7dcecad103d273b52d3542c034
That fixes the permissions issue as the ListWatch requests are now restricted to the desired namespace.

And in the default case of not specifying a namespace the ListWatch goes back to making cluster-scoped requests.
https://github.com/kubernetes/client-go/blob/master/rest/request.go#L424

hasbro17:
Fixes #124 

A namespace option can be passed from the manager to the cache which restricts the ListWatch for all informers to the desired namespace. 
This way a manager can be run with a Role instead of a ClusterRole.

### DIFF
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -42,7 +42,9 @@ type createListWatcherFunc func(gvk schema.GroupVersionKind, ip *specificInforme
 func newSpecificInformersMap(config *rest.Config,
 	scheme *runtime.Scheme,
 	mapper meta.RESTMapper,
-	resync time.Duration, createListWatcher createListWatcherFunc) *specificInformersMap {
+	resync time.Duration,
+	namespace string,
+	createListWatcher createListWatcherFunc) *specificInformersMap {
 	ip := &specificInformersMap{
 		config:            config,
 		Scheme:            scheme,
@@ -52,6 +54,7 @@ func newSpecificInformersMap(config *rest.Config,
 		paramCodec:        runtime.NewParameterCodec(scheme),
 		resync:            resync,
 		createListWatcher: createListWatcher,
+		namespace:         namespace,
 	}
 	return ip
 }
@@ -102,6 +105,10 @@ type specificInformersMap struct {
 	// and allows for abstracting over the particulars of structured vs
 	// unstructured objects.
 	createListWatcher createListWatcherFunc
+
+	// namespace is the namespace that all ListWatches are restricted to
+	// default or empty string means all namespaces
+	namespace string
 }
 
 // Start calls Run on each of the informers and sets started to true.  Blocks on the stop channel.
@@ -225,14 +232,16 @@ func createStructuredListWatch(gvk schema.GroupVersionKind, ip *specificInformer
 	return &cache.ListWatch{
 		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 			res := listObj.DeepCopyObject()
-			err := client.Get().Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do().Into(res)
+			isNamespaceScoped := ip.namespace != "" && mapping.Scope.Name() != meta.RESTScopeNameRoot
+			err := client.Get().NamespaceIfScoped(ip.namespace, isNamespaceScoped).Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do().Into(res)
 			return res, err
 		},
 		// Setup the watch function
 		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
 			// Watch needs to be set to true separately
 			opts.Watch = true
-			return client.Get().Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Watch()
+			isNamespaceScoped := ip.namespace != "" && mapping.Scope.Name() != meta.RESTScopeNameRoot
+			return client.Get().NamespaceIfScoped(ip.namespace, isNamespaceScoped).Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Watch()
 		},
 	}, nil
 }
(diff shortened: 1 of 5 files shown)


hasbro17:
@DirectXMan12 PTAL
First time writing a test in Ginkgo so I'm not sure if the namespaced cache test case should be in it's own Describe block. Seemed more concise to add it as a sub-case to the Reader.

pkg/manager/manager.go
```diff
@@ -128,7 +132,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 	}
 
 	// Create the cache for the cached read client and registering informers
-	cache, err := options.newCache(config, cache.Options{Scheme: options.Scheme, Mapper: mapper, Resync: options.SyncPeriod})
+	cache, err := options.newCache(config, cache.Options{Scheme: options.Scheme, Mapper: mapper, Resync: options.SyncPeriod, Namespace: options.Namespace})
```
interma: When I need to watch all namespace events, set nil to `options.Namespace` or set other value?
Thanks.

hasbro17: To watch all namespaces you set `options.Namespaces = ""` or [metav1.NamespaceAll](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L278-L279) which is the same thing.
But since the default string value is empty you can just not set the Namespace option in `cache.Options` and it should watch all namespaces by default.

hasbro17:
@DirectXMan12 The request is explicitly namespace scoped now. PTAL.

hasbro17:
@droot can you please review this in case @DirectXMan12 is busy.

droot:
@hasbro17 change looks good to me. @DirectXMan12 , would you like to take a final look ?

hasbro17:
@droot thanks. I think @DirectXMan12 should be back on Monday to take a look.
I'll rebase my changes in the meanwhile since #101 was merged.

hasbro17:
Rebased and added another namespaced cache test for an unstructured object.

pkg/manager/manager.go
```diff
@@ -106,6 +106,10 @@ type Options struct {
 	// will use for holding the leader lock.
 	LeaderElectionID string
 
+	// Namespace if specified restricts the manager's cache to watch the desired namespace
+	// Defaults to all namespaces
```
droot: May be make it more explicit:
> watch `objects` in the desired namespace

Also one more q: if I specify the namespace here and later setup a watch for a cluster scoped resource (say Node), will I get error, what would be the UX ?
We should document that behavior here probably.

hasbro17: I tested it out by specifying the manager's namespace and added a controller to watch nodes:
```sh
$ ./main --namespace=default
{"level":"info","ts":1537598131.6718707,"logger":"kubebuilder.controller","caller":"controller/controller.go:120","msg":"Starting EventSource","Controller":"node-controller","Source":{"Type":{"metadata":{"creationTimestamp":null},"spec":{},"status":{"daemonEndpoints":{"kubeletEndpoint":{"Port":0}},"nodeInfo":{"machineID":"","systemUUID":"","bootID":"","kernelVersion":"","osImage":"","containerRuntimeVersion":"","kubeletVersion":"","kubeProxyVersion":"","operatingSystem":"","architecture":""}}}}}
E0921 23:35:31.674413   15574 reflector.go:205] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1.Node: the server could not find the requested resource
E0921 23:35:32.680154   15574 reflector.go:205] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1.Node: the server could not find the requested resource
```
As expected the ListWatcher fails to List the nodes because when the namespace is set it will form an incorrect request to list Nodes by adding the namespace in the REST path.
https://github.com/kubernetes-sigs/controller-runtime/pull/136/files#diff-fc36bd1640f2ed10bc9e4f4d24789edfR235

The behavior seems consistent with the need for namespacing the manager i.e you don't expect to Watch cluster-scoped resources if your permissions are restricted to a namespace.

I've updated the comment to document this behavior.

pkg/cache/internal/informers_map.go
```diff
@@ -225,14 +232,14 @@ func createStructuredListWatch(gvk schema.GroupVersionKind, ip *specificInformer
 	return &cache.ListWatch{
 		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 			res := listObj.DeepCopyObject()
-			err := client.Get().Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do().Into(res)
+			err := client.Get().NamespaceIfScoped(ip.namespace, ip.namespace != "").Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do().Into(res)
```
DirectXMan12: it occurs to me that there's one last wrinkle for this -- we need to also check if the resource is namespace-scoped to begin with using the rest mapping -- i.e. you should be able to watch nodes when watching pods in a specific namespace.

DirectXMan12: so this should be `NamespaceIfScoped(ip.namespace, ip.namespace != "" && mapping.Scope == apimeta.RESTScopeNameRoot)`

DirectXMan12: and below (in the watch)

pkg/cache/internal/informers_map.go
```diff
@@ -252,12 +259,18 @@ func createUnstructuredListWatch(gvk schema.GroupVersionKind, ip *specificInform
 	// Create a new ListWatch for the obj
 	return &cache.ListWatch{
 		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
+			if ip.namespace != "" {
```
DirectXMan12: ditto here on the "is this namespaced"

hasbro17:
@DirectXMan12 I've updated the ListWatchers to check if the resource is namespace scoped before making a namespaced request.
With this change ,even if the cache is namespaced, it can still be used to read cluster-scoped resources.

I've updated the tests with the additional case of ensuring a namespaced cache will still let you List a cluster-scoped resource(`corev1.Namespace`).
Also updated the manager's `Namespace` option comment to clarify this.

pkg/cache/internal/informers_map.go
hasbro17: Fixed. Minor nit being it's:
```Go
ip.namespace != "" && mapping.Scope != apimeta.RESTScopeNameRoot
```

PR #139, opened 2018-09-11
change how to set webhook type and how we default path in builder

mengqiy:
Address comments came from https://github.com/mengqiy/webhook-scaffolding/pull/1

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -34,7 +34,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/types"
 )
 
 var log = logf.Log.WithName("example-controller")
@@ -76,8 +75,7 @@ func main() {
 	// Setup webhooks
 	mutatingWebhook, err := builder.NewWebhookBuilder().
 		Name("mutating.k8s.io").
-		Type(types.WebhookTypeMutating).
-		Path("/mutating-pods").
+		Mutating().
 		Operations(admissionregistrationv1beta1.Create, admissionregistrationv1beta1.Update).
 		WithManager(mgr).
 		ForType(&corev1.Pod{}).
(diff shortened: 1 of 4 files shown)


pkg/webhook/admission/builder/builder.go
```diff
@@ -165,48 +174,48 @@ func (b *WebhookBuilder) Build() (*admission.Webhook, error) {
...
+		}
 	}
 
-	if b.operations == nil {
-		b.operations = []admissionregistrationv1beta1.OperationType{
-			admissionregistrationv1beta1.Create,
-			admissionregistrationv1beta1.Update,
+	if len(b.path) == 0 {
+		if *b.t == types.WebhookTypeMutating {
+			b.path = "/mutating" + w.Rules[0].Resources[0]
```
DirectXMan12: super-nit: maybe just `/mutate-{resource}` or `/validate-{resource}`?

pkg/webhook/admission/builder/builder.go
```diff
@@ -73,10 +73,19 @@ func (b *WebhookBuilder) Name(name string) *WebhookBuilder {
...
+func (b *WebhookBuilder) Mutating() *WebhookBuilder {
+	m := types.WebhookTypeMutating
+	b.t = &m
+	return b
+}
+
+// Validating sets the type to validating admission webhook
+// Only one of Mutating and Validating can be invoked.
+func (b *WebhookBuilder) Validating() *WebhookBuilder {
+	m := types.WebhookTypeValidating
```
DirectXMan12: do we have a default?  Can we have a default?

mengqiy: No, I don't think we can choose a reasonable defaulting for webhook type.

example/main.go
```diff
@@ -76,8 +75,7 @@ func main() {
 	// Setup webhooks
 	mutatingWebhook, err := builder.NewWebhookBuilder().
 		Name("mutating.k8s.io").
-		Type(types.WebhookTypeMutating).
-		Path("/mutating-pods").
+		Mutating().
```
Liujingfang1: Maybe change the function names to `BeMutating` and `BeValidating`.

example/main.go
```diff
@@ -103,7 +100,7 @@ func main() {
 	}
 
 	as, err := webhook.NewServer("foo-admission-server", mgr, webhook.ServerOptions{
-		Port:    443,
+		Port:    9876,
```
Liujingfang1: Why is this port changed?

example/main.go
mengqiy: IMHO changing methods names to `BeMutating` and `BeValidating` doesn't buy us much. And just call if `Mutating` is more concise.

PR #141, opened 2018-09-12
drop KVMap

mengqiy:
- drop KVMap from Server
- cleanup godoc
- tweak some defaulting regarding webhook name and path

### DIFF
--- a/example/mutatingwebhook.go
+++ b/example/mutatingwebhook.go
@@ -18,11 +18,11 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
 )
@@ -46,7 +46,7 @@ func (a *podAnnotator) Handle(ctx context.Context, req types.Request) types.Resp
 	}
 	copy := pod.DeepCopy()
 
-	err = mutatePodsFn(ctx, copy)
+	err = a.mutatePodsFn(ctx, copy)
 	if err != nil {
 		return admission.ErrorResponse(http.StatusInternalServerError, err)
 	}
@@ -54,16 +54,30 @@ func (a *podAnnotator) Handle(ctx context.Context, req types.Request) types.Resp
 }
 
 // mutatePodsFn add an annotation to the given pod
-func mutatePodsFn(ctx context.Context, pod *corev1.Pod) error {
-	v, ok := ctx.Value(admission.StringKey("foo")).(string)
-	if !ok {
-		return fmt.Errorf("the value associated with %v is expected to be a string", "foo")
+func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error {
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
 	}
-	anno := pod.Annotations
-	if anno == nil {
-		anno = map[string]string{}
-	}
-	anno["example-mutating-admission-webhook"] = v
-	pod.Annotations = anno
+	pod.Annotations["example-mutating-admission-webhook"] = "foo"
+	return nil
+}
+
+// podValidator implements inject.Client.
+// A client will be automatically injected.
+var _ inject.Client = &podValidator{}
+
+// InjectClient injects the client.
+func (v *podAnnotator) InjectClient(c client.Client) error {
+	v.client = c
+	return nil
+}
+
+// podValidator implements inject.Decoder.
+// A decoder will be automatically injected.
+var _ inject.Decoder = &podValidator{}
+
+// InjectDecoder injects the decoder.
+func (v *podAnnotator) InjectDecoder(d types.Decoder) error {
+	v.decoder = d
 	return nil
 }
--- a/pkg/webhook/admission/builder/builder.go
+++ b/pkg/webhook/admission/builder/builder.go
@@ -31,33 +31,40 @@ import (
 
 // WebhookBuilder builds a webhook based on the provided options.
 type WebhookBuilder struct {
-	// name specifies the Name of the webhook. It must be unique in the http
-	// server that serves all the webhooks.
+	// name specifies the name of the webhook. It must be unique among all webhooks.
 	name string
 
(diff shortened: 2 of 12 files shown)


mengqiy:
cc: @droot

mengqiy:
@DirectXMan12 PTAL

example/mutatingwebhook.go
```diff
@@ -46,24 +46,40 @@ func (a *podAnnotator) Handle(ctx context.Context, req types.Request) types.Resp
...
 }
 
 // mutatePodsFn add an annotation to the given pod
-func mutatePodsFn(ctx context.Context, pod *corev1.Pod) error {
-	v, ok := ctx.Value(admission.StringKey("foo")).(string)
-	if !ok {
-		return fmt.Errorf("the value associated with %v is expected to be a string", "foo")
-	}
+func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error {
 	anno := pod.Annotations
```
DirectXMan12: this can be shortened to just
if pod.Annotations == nil {
  pod.Annotations = map[string]string{}
}
pod.Annotations["xyz"] = "qpr"

pkg/webhook/admission/builder/builder.go
```diff
@@ -89,8 +98,11 @@ func (b *WebhookBuilder) Validating() *WebhookBuilder {
 	return b
 }
 
-// Path sets the Path for the webhook.
-// This is optional
+// Path sets the path for the webhook.
+// Path needs to be unique among different webhooks.
+// This is optional. If not set, We will try to default it by deriving from webhook
```
DirectXMan12: > If not set, it will be built from the type and resource name.  For example,
> a webhook that mutates pods has a default path of "/mutate-pods"

pkg/webhook/admission/builder/builder.go
```diff
@@ -89,8 +98,11 @@ func (b *WebhookBuilder) Validating() *WebhookBuilder {
...
+// type and resource name. e.g. "/mutate-pods"
+// If we can't find a unique default path for it, user
```
DirectXMan12: ??

pkg/webhook/admission/builder/builder.go
```diff
@@ -159,6 +171,23 @@ func (b *WebhookBuilder) validate() error {
 	return nil
 }
 
+func (b *WebhookBuilder) setDefaults(w *admission.Webhook) {
+	if len(b.path) == 0 {
+		if *b.t == types.WebhookTypeMutating {
+			b.path = "/mutate-" + w.Rules[0].Resources[0]
```
DirectXMan12: do resources have a group in them?  Will this end up as `resource.group` (meaning we need to do substitution?)

pkg/webhook/admission/webhook.go
```diff
@@ -84,23 +80,30 @@ type Webhook struct {
 }
 
 func (w *Webhook) setDefaults() {
-	if len(w.Name) == 0 {
-		w.Name = "admissionwebhook.k8s.io"
-	}
 	if len(w.Path) == 0 {
+		if len(w.Rules) == 0 || len(w.Rules[0].Resources) == 0 {
```
DirectXMan12: should this logic be extracted to a helper or something?  Can the builder rely on the underlying webhook implementation to do the defaulting?

pkg/webhook/admission/webhook.go
```diff
@@ -84,23 +80,30 @@ type Webhook struct {
...
+			w.Path = "/mutate-" + w.Rules[0].Resources[0]
 		} else if w.Type == types.WebhookTypeValidating {
-			w.Path = "/validatingwebhook"
+			w.Path = "/validate-" + w.Rules[0].Resources[0]
 		}
 	}
+	if len(w.Name) == 0 {
+		reg := regexp.MustCompile("[^a-zA-Z0-9]+")
+		processedPath := strings.ToLower(reg.ReplaceAllString(w.Path, ""))
+		w.Name = processedPath + ".k8s.io"
```
DirectXMan12: is `.k8s.io` always the right suffix?  that seems wrong...

pkg/webhook/admission/builder/builder.go
mengqiy: > do resources have a group in them?

No. Group is in `w.Rules[0].Groups`. Rule has the following struct:
```
type Rule struct {
	APIGroups []string
	APIVersions []string
	Resources []string
}
```
We are not using `group` name in the path to keep it short.

pkg/webhook/admission/webhook.go
mengqiy: In kubebuilder scaffolding, kubebuilder should use the domain name. i.e. `builder.Name("myfeature.mycorp.com")`
But in the library defaulting, there is no such information. IMO using `k8s.io` as default in the library is fine.

pkg/webhook/admission/webhook.go
mengqiy: Either way works. No strong opinion which way to use. @DirectXMan12 WDYT?

mengqiy: > Can the builder rely on the underlying webhook implementation to do the defaulting?

This will make the builder return an incomplete webhook struct. But I guess it's not a big problem.

mengqiy: I have dropped the defaulting in builder and only relying the defaulting in `webhook.go`

pkg/webhook/admission/webhook.go
DirectXMan12: k8s.io is for official kubernetes projects.  We should not encourage people to be misusing that suffix.

DirectXMan12:
> But in the library defaulting, there is no such information. IMO using k8s.io as default in the library is fine.

k8s.io is for official kubernetes projects. We should not encourage people to be misusing that suffix.

mengqiy:
@DirectXMan12 PTAL

pkg/webhook/admission/webhook.go
```diff
@@ -84,23 +80,30 @@ type Webhook struct {
...
-		w.Name = "admissionwebhook.k8s.io"
-	}
 	if len(w.Path) == 0 {
+		if len(w.Rules) == 0 || len(w.Rules[0].Resources) == 0 {
+			// can't do defaulting, skip it.
+			return
+		}
 		if w.Type == types.WebhookTypeMutating {
-			w.Path = "/mutatingwebhook"
+			w.Path = "/mutate-" + w.Rules[0].Resources[0]
```
DirectXMan12: For later on, we should consider adding in `.APIGroups[0]`, but we'll merge as-is for now.

PR #146, opened 2018-09-17
Revert "Convert client.List to use functional options"

mengqiy:
Reverts kubernetes-sigs/controller-runtime#106

#106 is a breaking change to the client interface.
Per suggestion from @pwittrock, we are going to revert it for now before we make a decision about how to deal with it.
We can decide later if we want to [introduce the new signature as a new method and mark the old one as deprecated](https://github.com/kubernetes-sigs/controller-runtime/pull/106) or do a `v0.2.0` release for it.

### DIFF
--- a/pkg/builder/example_test.go
+++ b/pkg/builder/example_test.go
@@ -77,7 +77,7 @@ func (a *ReplicaSetReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
 
 	// List the Pods matching the PodTemplate Labels
 	pods := &corev1.PodList{}
-	err = a.List(context.TODO(), pods, client.InNamespace(req.Namespace), client.MatchingLabels(rs.Spec.Template.Labels))
+	err = a.List(context.TODO(), client.InNamespace(req.Namespace).MatchingLabels(rs.Spec.Template.Labels), pods)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -113,7 +113,7 @@ var _ = Describe("Informer Cache", func() {
 		It("should be able to list objects that haven't been watched previously", func() {
 			By("listing all services in the cluster")
 			listObj := &kcorev1.ServiceList{}
-			Expect(informerCache.List(context.Background(), listObj)).To(Succeed())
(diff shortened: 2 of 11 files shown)


droot:
@grantr @DirectXMan12 This introduces breaking change without any deprecation (which is making harder to make it available in v0.1.x versions of controller-runtime). We need to cut a release to unblock some other features, lets revert it for now, and figure out how we want to roll out this breaking change.

grantr:
@droot has the situation changed since https://github.com/kubernetes-sigs/controller-runtime/pull/106#issuecomment-418516452?

> we have sorted out the versioning story for controller-runtime/tools and Kubebuilder now uses versioned pkgs of controller-*, so we can move forward with these changes now.

FWIW #94 is technically a breaking change to the interface also, though unlike #106 it adds critical functionality to `client.Delete`.

mengqiy:
> FWIW #94 is technically a breaking change to the interface also, though unlike #106 it adds critical functionality to client.Delete.

@grantr we can have 0 `DeleteOptionFunc`, so that's not a breaking change.

droot:
@grantr Though we have started versioning controller-runtime, we still do not have a process for managing multiple revisions like branches for 0.1.x and 0.2.x.. and cherrypicking etc. I was assuming we will introduce "ListVariadic" version without dropping the existing "List" method (mark deprecated) which is not breaking compatibility (in hard way) because interface and implementation are both the provided by `controller-runtime`. So if your code uses `client` interface and same implementation, update should work. And then in major version (0.2.0), we remove rename `ListVariadic` to `List`. 

Sorry, I was OOO last week and couldn't participate in review.

grantr:
> we can have 0 DeleteOptionFunc, so #94 is not a breaking change.

It is a breaking _interface_ change. Any existing implementations of the interface will no longer compile. See https://github.com/kubernetes-sigs/controller-runtime/pull/94#issuecomment-415859063.

mengqiy:
> usage of the client interface (Delete() calls) is backward compatible, but implementation of the client interface is not.

Yes, this is true. From this perspective, even `introducing "ListVariadic" version without dropping the existing "List" method` is still a breaking change.
But IMO not breaking the usages of our client implementation is more important than not breaking user's implementation. I assume there are much more people using our client implementation than implementing their own client.

grantr:
That seems like a reasonable policy to me. Maybe something to add to the compatibility guarantee, if there is one.

PR #148, opened 2018-09-18
Adding stop channel so injecting the stop channel does not inject nil

shawn-hurley:


mhrivnak:
@shawn-hurley looks like the test is expecting the value to be `nil` for some reason, which is why CI failed.

https://github.com/kubernetes-sigs/controller-runtime/blob/8d93cc75ed7086e85ec7de09332a3bfd8603a959/pkg/manager/manager_test.go#L331

DirectXMan12:
/hold

I think this doesn't quite work

PR #150, opened 2018-09-18
Support optional identifier for handlers

mengqiy:
One handler may represent a feature and people may want to disable a feature at runtime.
So a handler should have an optional identifier.

### DIFF
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -42,6 +42,13 @@ type Handler interface {
 	Handle(context.Context, atypes.Request) atypes.Response
 }
 
+// NameGetter gets the name of the Handler
+type NameGetter interface {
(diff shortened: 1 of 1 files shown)


pkg/webhook/admission/webhook.go
```diff
@@ -42,6 +42,13 @@ type Handler interface {
 	Handle(context.Context, atypes.Request) atypes.Response
 }
 
+// NameGetter gets the name of the Handler
+type NameGetter interface {
+	// Name returns the name of the Handler.
+	// The name should be a feature name that can be the identifier for the handler.
+	Name() string
```
droot: This doesn't need to exist in the controller-runtime IMO.

mengqiy:
Had an offline discussion with @droot, I agree we don't need to have the `NameGetter` in the controller-runtime lib at least for now.

PR #158, opened 2018-09-22
 support gen-manifests and disable-installer

mengqiy:
Some test changes make the PR looks big, but it's actually not that big :)

### DIFF
--- a/pkg/webhook/bootstrap.go
+++ b/pkg/webhook/bootstrap.go
@@ -136,7 +136,7 @@ func (s *Server) installWebhookConfig() error {
 	}
 
 	if !s.GenerateManifests && s.DisableInstaller {
-		log.Info("webhook installer is disabled")
+		log.Info("generating webhook manifests is disabled and webhook installer is also disabled")
 		return nil
 	}
 	if s.GenerateManifests {
--- a/pkg/webhook/bootstrap.go
+++ b/pkg/webhook/bootstrap.go
@@ -26,10 +26,9 @@ import (
 	"path"
 	"strconv"
 
-	"github.com/ghodss/yaml"
-
 	"k8s.io/api/admissionregistration/v1beta1"
 	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -93,6 +92,15 @@ func (s *Server) setBootstrappingDefault() {
 		s.Host = &varString
 	}
 
+	if s.Writer == nil {
+		s.Writer = os.Stdout
+	}
+	if s.GenerateManifests {
+		s.Client = &writerClient{
+			writer: s.Writer,
+		}
+	}
+
 	var certWriter writer.CertWriter
 	var err error
 	if s.Secret != nil {
@@ -114,12 +122,10 @@ func (s *Server) setBootstrappingDefault() {
 	s.certProvisioner = &cert.Provisioner{
 		CertWriter: certWriter,
 	}
-	if s.Writer == nil {
-		s.Writer = os.Stdout
-	}
 }
 
-// installWebhookConfig writes the configuration of admissionWebhookConfiguration in yaml format if dryrun is true.
+// installWebhookConfig writes the configuration of admissionWebhookConfiguration in yaml format
+// if GenerateManifests is true.
 // Otherwise, it creates the the admissionWebhookConfiguration objects and service if any.
 // It also provisions the certificate for the admission server.
 func (s *Server) installWebhookConfig() error {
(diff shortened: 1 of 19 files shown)


pkg/webhook/bootstrap.go
```diff
@@ -129,6 +135,14 @@ func (s *Server) installWebhookConfig() error {
 		return s.err
 	}
 
+	if !s.GenerateManifests && s.DisableInstaller {
+		log.Info("webhook installer is disabled")
```
droot: log seems to be incorrect. may be add "and generating webhook manifests is also disabled"

pkg/webhook/bootstrap.go
```diff
@@ -360,3 +352,100 @@ func (s *Server) service() runtime.Object {
...
+						{
+							Name:         "cert",
+							VolumeSource: volumeSource,
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
droot: I was thinking about just printing out the webhook config only. Outputting other manifests means, now we have two different ways to generate the manifests for controller ?

pkg/webhook/internal/cert/generator/selfsigned.go
```diff
@@ -30,24 +32,42 @@ func ServiceToCommonName(serviceNamespace, serviceName string) string {
 
 // SelfSignedCertGenerator implements the certGenerator interface.
 // It provisions self-signed certificates.
-type SelfSignedCertGenerator struct{}
+type SelfSignedCertGenerator struct {
+	CAKey  []byte
+	CACert []byte
```
droot: why are they public ? (if they are public then probably we don't need the `SetCA` method ?

pkg/webhook/internal/cert/generator/selfsigned.go
```diff
@@ -30,24 +32,42 @@ func ServiceToCommonName(serviceNamespace, serviceName string) string {
...
-	}
-	signingCert, err := cert.NewSelfSignedCACert(cert.Config{CommonName: "webhook-cert-ca"}, signingKey)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create the CA cert: %v", err)
+	var signingKey *rsa.PrivateKey
+	var signingCert *x509.Certificate
+	var valid bool
+	var err error
+
+	valid, signingKey, signingCert = cp.validCACert()
```
droot: Can these become invalid due to cert expiry ? Do we need to differentiate between expiry from other scenario ? (in case of expiry, we probably don't need to re-generate the CA ? )

pkg/webhook/internal/cert/writer/certwriter.go
```diff
@@ -39,7 +41,7 @@ const (
 // CertWriter provides method to handle webhooks.
 type CertWriter interface {
 	// EnsureCert provisions the cert for the webhookClientConfig.
-	EnsureCert(dnsName string, dryrun bool) (*generator.Artifacts, bool, error)
+	EnsureCert(dnsName string) (*generator.Artifacts, bool, error)
```
droot: Making these pkgs internal is helping us today :)

pkg/webhook/internal/cert/writer/secret.go
```diff
@@ -127,39 +112,30 @@ func (s *secretCertWriter) overwrite() (
...
 	secret := &corev1.Secret{}
 	err := s.Client.Get(nil, *s.Secret, secret)
 	if apierrors.IsNotFound(err) {
 		return nil, notFoundError{err}
 	}
-	return secretToCerts(secret), err
+	certs := secretToCerts(secret)
+	if certs != nil {
+		// Store the CA for next usage.
+		s.CertGenerator.SetCA(certs.CAKey, certs.CACert)
```
droot: I am assuming secret will be updated only if installer is enabled ?

pkg/webhook/internal/cert/generator/selfsigned.go
mengqiy: The generated CA is valid for 10 years.

I added some check to ensure it won't expire in 1 year.

pkg/webhook/internal/cert/writer/secret.go
mengqiy: Not always true.
Disabling the installer only skip installation during bootstrapping.
The webhook server will refresh its cert if the the cert is expiring.
IMO installer only cover bootstrapping time, after that the server may try to rotate its cert if necessary. If it doesn't have permission it will fail. 
We probably want to reconsider if it should fail or just log a warning.

pkg/webhook/bootstrap.go
mengqiy: IMO using generated Deployment (can also be a StatefulSet with a headless Service) is less error-prone.
There are quite some places need to be set correctly in order to make the webhook server binary work correctly.
We can discuss in person tomorrow :)

DirectXMan12:
What's the main reason that we've added install-webhook-config to controller-runtime, but we removed the equivalent install-crds?  Is there any reason not to have the manifest generation in controller-tools like for CRDs?

mengqiy:
It makes more sense to have the generator in controller-tools.
I will close this PR as a record.

PR #159, opened 2018-09-25
Support for client-go code generated controllers as Controller Watch …

pwittrock:
This is useful for projects that either:

1. Want to migrate to controller-runtime client incrementally from client-go generated libraries
1. Prefer to use the client-go generated libraries

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/source/internal"
 
+	toolscache "k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Can you add a description of "why" for posterity?  Even just "this makes it easier for existing projects to incrementally migrate to controller-runtime" or such.

pwittrock:
@DirectXMan12 Done

pkg/source/source.go
```diff
@@ -221,6 +222,28 @@ func (cs *Channel) syncLoop() {
...
+
+	// Informer should have been specified by the user.
+	if ks.Informer == nil {
+		return fmt.Errorf("must specify Informer.Informer")
+	}
+
+	ks.Informer.AddEventHandler(internal.EventHandler{Queue: queue, EventHandler: handler, Predicates: prct})
+	return nil
+}
+
```
droot: Thinking more about the code in the manager's `main` fn, I am wondering if we need to add the concept of `Source` being ready (in this case, informed being sync'ed) from the perspective of wiring everything. Will have to look at the code to reason more about it.

DirectXMan12:
@droot for your comment above, is that a "I want to look at this pre-merge" or "I want to look at this post-merge"

droot:
> is that a "I want to look at this pre-merge" or "I want to look at this post-merge"

Oh, sorry, I wasn't clear. I meant "post-merge", so I am good with the change.

PR #163, opened 2018-09-27
lets controllerManager.start() return instead of blocking

mhrivnak:
It always gets called as a goroutine, so nothing expects it to block. This
change slightly simplifies the code and lets the goroutine die instead of
keeping it around forever.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -204,42 +204,34 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 }
 
 func (cm *controllerManager) start(stop <-chan struct{}) {
-	func() {
-		cm.mu.Lock()
-		defer cm.mu.Unlock()
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
 
-		cm.stop = stop
+	cm.stop = stop
 
-		// Start the Cache. Allow the function to start the cache to be mocked out for testing
-		if cm.startCache == nil {
-			cm.startCache = cm.cache.Start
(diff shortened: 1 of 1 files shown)


mhrivnak:
The diff looks more complex than it is. It just:
1. removes the `select` from the bottom
2. removes the inner `func()` block

Everything else is untouched except moving the indentation left.

When viewing it, if you click "Diff settings" and select "Hide whitespace changes", you'll see there's not much to it.

droot:
Looks good to me. Would like @JoelSpeed to take a look just to ensure this change plays nicely with the leaderElection code that we have in `manager.Start()`.

droot:
Another thing I noticed that while using leader election, leaderElector passes its own stop channel to `cm.start`, so not sure if that leads to unexpected behavior. Also, the code for leader election in `cm.Start` sort of looks complex with multiple go routines and select blocks. /cc @JoelSpeed @mhrivnak @DirectXMan12

droot:
> I don't think this does. When the leader elector loses leader it closes the channel and sends an error which in turn causes cm.Start to return the error, exiting the controller. If stop passed to cm.Start is closed then the function returns as well in theory closing the controller, we don't at this point then close the leader electors stop channel which I guess could cause problems, this could be avoided with a little refactor though. Want me to do that and make sure we ignore the leader elector's stop channel?

Took a quick look a the leader elector lib, ignoring the leader elector's stop channel is fine as long as we exit when we stop leading (which we do as you explained above). I saw you made another PR, will take a look.

PR #165, opened 2018-09-30
envtest: export DefaultKubeAPIServerFlags & make flags configurable

### LINKED ISSUE #112: Support appending additional flags to defaultKubeAPIServerFlags
There's currently no way to pass additional flags to the API server. For instance, if I want to use an alpha feature (like PriorityClass, prior to 1.11.0), I need to set a flag on the API server.

darkowlzz:
This change exports DefaultKubeAPIServerFlags, allowing a user to
append the default flags with other custom flags and set it to
Environment.KubeAPIServerFlags.
If KubeAPIServerFlags is not set, DefaultKubeAPIServerFlags is used when
the API server starts.

Fixes #112

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -53,9 +53,8 @@ func defaultAssetPath(binary string) string {
 
 }
 
-// APIServerDefaultArgs are flags necessary to bring up apiserver.
-// TODO: create test framework interface to append flag to default flags.
-var defaultKubeAPIServerFlags = []string{
+// DefaultKubeAPIServerFlags are default flags necessary to bring up apiserver.
+var DefaultKubeAPIServerFlags = []string{
 	"--etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}",
 	"--cert-dir={{ .CertDir }}",
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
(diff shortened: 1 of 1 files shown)


DirectXMan12:
I'd prefer we didn't break backwards compat here, since it should be fairly easy to work around it.  Perhaps just make `DefaultAPIServerArgs` public, and then default APIServerFlags if it's not set already?

darkowlzz:
> I'd prefer we didn't break backwards compat here, since it should be fairly easy to work around it. Perhaps just make DefaultAPIServerArgs public, and then default APIServerFlags if it's not set already?

Yeah, that's much better. Removed the constructor and exported `DefaultKubeAPIServerFlags`. Since it's an exported slice, users can append it on their own and set a custom set of flags to `Environment.KubeAPIServerFlags`.

pkg/envtest/server.go
```diff
@@ -46,9 +46,8 @@ func defaultAssetPath(binary string) string {
 
 }
 
-// APIServerDefaultArgs are flags necessary to bring up apiserver.
-// TODO: create test framework interface to append flag to default flags.
-var defaultKubeAPIServerFlags = []string{
+// DefaultKubeAPIServerFlags are default flags necessary to bring up apiserver.
+var DefaultKubeAPIServerFlags = []string{
```
droot: Is there a need to export this anymore given that we have made `KubeAPIServerFlag` configurable (change below) ?

DirectXMan12: yes, because we still want to allow people to make use of the default args if they just want to append.

darkowlzz:
Rebased.

darkowlzz:
Cyclomatic Complexity went above 10 because of the if statement.
I created a new function `Environment.getAPIServerFlags()` to return the set flags or the default flags if not set. That fixed the failure due to CC > 10.
Hope that's the right way to handle it :)

DirectXMan12:
@darkowlzz can you rebase, please (or let us know if you need us to step in and rebase)

darkowlzz:
Rebased!

PR #166, opened 2018-09-30
pkg/client: Pass ctx to typedClient requests

### LINKED ISSUE #40: pkg/client methods should respect given context's deadline
All exported methods on client interface takes in `context.Context` parameter, but the methods don't implement the deadline semantics.

darkowlzz:
This change passes the context in the client methods to the underlying
object requests created using client-go's Request for `typedClient`.
Context isn't passed to the `unstructuredClient` because the
`dynamicResourceClient` in client-go doesn't accepts any context.

Fixes #40

### DIFF
--- a/pkg/client/typed_client.go
+++ b/pkg/client/typed_client.go
@@ -30,7 +30,7 @@ type typedClient struct {
 }
 
 // Create implements client.Client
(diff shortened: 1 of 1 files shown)


DirectXMan12:
/lgtm
/approve

Good catch!

PR #169, opened 2018-10-06
Make start and stop timeouts for test controlplane configurable

alvaroaleman:
This allows configuring the start and stop timeouts for the controlplane. I've created this PR after repeated test failures in the kubernetes-sigs/cluster-api repo due to `timeout waiting for process kube-apiserver to start`, e.G. in https://github.com/kubernetes-sigs/cluster-api/pull/523

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -17,8 +17,10 @@ limitations under the License.
 package envtest
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/client-go/rest"
@@ -32,6 +34,8 @@ const (
 	envEtcdBin             = "TEST_ASSET_ETCD"
 	envKubectlBin          = "TEST_ASSET_KUBECTL"
 	envKubebuilderPath     = "KUBEBUILDER_ASSETS"
+	envStartTimeout        = "KUBEBUILDER_CONTROLPLANE_START_TIMEOUT"
+	envStopTimeout         = "KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT"
 	defaultKubebuilderPath = "/usr/local/kubebuilder/bin"
 	StartTimeout           = 60
 	StopTimeout            = 60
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -132,3 +147,22 @@ func (te *Environment) Start() (*rest.Config, error) {
 	})
 	return te.Config, err
 }
+
+func getTimeouts() (time.Duration, time.Duration, error) {
+	var startTimeout, stopTimeout time.Duration
+	if envVal := os.Getenv(envStartTimeout); envVal != "" {
+		converted, err := strconv.Atoi(envVal)
```
droot: Looks like a good usecase for `parseDuration` https://golang.org/pkg/time/#ParseDuration

pkg/envtest/server.go
```diff
@@ -132,3 +147,22 @@ func (te *Environment) Start() (*rest.Config, error) {
...
+			return startTimeout, stopTimeout, err
+		}
+		startTimeout = time.Duration(converted)
+	}
+	if envVal := os.Getenv(envStopTimeout); envVal != "" {
+		converted, err := strconv.Atoi(envVal)
+		if err != nil {
+			return startTimeout, stopTimeout, err
+		}
+		stopTimeout = time.Duration(converted)
```
droot: Same comment as above.

pkg/envtest/server.go
```diff
@@ -32,6 +35,8 @@ const (
 	envEtcdBin             = "TEST_ASSET_ETCD"
 	envKubectlBin          = "TEST_ASSET_KUBECTL"
 	envKubebuilderPath     = "KUBEBUILDER_ASSETS"
+	envStartTimeout        = "KUBEBUILDER_CONTROLPLANE_START_TIMEOUT"
+	envStopTimeout         = "KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT"
```
droot: I would recommend making these parts of `Environment` struct below as public args, so that callers can override the defaults and lets tweak the defaults on the basis of your experience. Lets avoid passing these through environment variables.

alvaroaleman: hey @droot , thanks for your feedback.

While I think adding them to the struct is a good idea I believe it still makes sense to populate them from the environment, because a controller-manager may contain more than one controller and this is a setting one does not want to set per controller but globally for all controllers, since there is no reason to believe that the test apiserver and etcd will have a different startup time per controller.

What do you think about adding them to the environment struct but populating them from environment variables in case they are unset?

droot: >What do you think about adding them to the environment struct but populating them from environment variables in case they are unset?

Picking the defaults (if unspecified) from environment variables sounds reasonable to me.

alvaroaleman: @droot I've updated the PR to add the start/stop timeout options to the `Environment` struct and only default from environment variables if they are unspecified, please take another look

alvaroaleman:
@droot Is there something else I need to do here? Should I `/assign` this PR to you?

PR #170, opened 2018-10-08
Adding string methods for the sources.

### LINKED ISSUE #153: Starting event source log for Channel contains error
Right now I am getting this log when starting a `watch` on a source of type `Channel`

{"level":"info","ts":1537378068.766453,"logger":"kubebuilder.controller","caller":"controller/controller.go:120","msg":"Starting EventSource","Controller":"database-controller","SourceError":"json: unsupported type: <-chan event.GenericEvent"}

shawn-hurley:
This will set the `String` method for the sources to fix #153

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -95,6 +95,13 @@ func (ks *Kind) Start(handler handler.EventHandler, queue workqueue.RateLimiting
 	return nil
 }
 
+func (ks *Kind) String() string {
+	if ks.Type != nil && ks.Type.GetObjectKind() != nil {
+		return fmt.Sprintf("kind source: %v", ks.Type.GetObjectKind().GroupVersionKind().String())
+	}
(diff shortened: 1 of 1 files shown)


pkg/source/source.go
```diff
@@ -132,6 +136,10 @@ type Channel struct {
 	destLock sync.Mutex
 }
 
+func (c *Channel) String() string {
+	return fmt.Sprintf("channel source: %p", c)
```
shawn-hurley: @DirectXMan12 @droot I chose to print the pointer because I didn't know of a better field to make it identifiable and thoughts?

droot: hmm..can't think of better alternative in this case.

shawn-hurley:
@droot I think if this is good to merge it needs an approved label.

Are we waiting on @DirectXMan12 to take a look too?

PR #174, opened 2018-10-18
Support disable installer

mengqiy:
This supersedes #158.


droot:
@mengqiy Took a quick look, looks good. Will take another detailed look in the afternoon.

mengqiy:
PTAL

example/main.go
```diff
@@ -39,18 +39,24 @@ import (
 var log = logf.Log.WithName("example-controller")
 
 func main() {
+	var enableInstaller bool
+	flag.BoolVar(&enableInstaller, "enable-installer", false,
+		"enable the installer in the webhook server, so it will install webhook related resources during bootstrapping")
+
```
droot: shall we rename the flag to reflect the usage ? for ex. `--install-webhook-config` ?

pkg/webhook/server.go
```diff
@@ -54,10 +54,9 @@ type ServerOptions struct {
 	// Client will be injected by the manager if not set.
 	Client client.Client
 
-	// Dryrun controls if the server will install the webhookConfiguration and service if any.
-	// If true, it will print the objects in yaml format.
-	// If false, it will install the objects in the cluster.
-	Dryrun bool
+	// EnableInstaller controls if the server will automatically create webhook related objects
+	// during bootstrapping. e.g. webhookConfiguration, service and secret.
+	EnableInstaller bool
```
droot: May be rename the flag ? ex. `InstallWebhookConfig` ?

pkg/webhook/server.go
```diff
@@ -187,13 +187,17 @@ func (s *Server) Handle(pattern string, handler http.Handler) {
...
-// Otherwise, it will print the objects in yaml format.
+// Start runs the server.
+// It will install the webhook related resources depend on the server configuration.
 func (s *Server) Start(stop <-chan struct{}) error {
-	err := s.installWebhookConfig()
-	// if encounter an error or it's in dryrun mode, return.
-	if err != nil || s.Dryrun {
-		return err
+	if s.EnableInstaller {
+		log.Info("webhook installer is enabled")
```
droot: We can improve the log --> installing webhook configuration in cluster ? or something for user-friendly.

mengqiy:
Addressed comments.
Pushed one more commit to address an issue due to https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/register.go#L26-L29

pkg/webhook/internal/cert/writer/secret_test.go
```diff
@@ -78,6 +79,10 @@ var _ = Describe("secretCertWriter", func() {
 			//isController := true
 			//blockOwnerDeletion := true
 			secret = &corev1.Secret{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "v1",
+					Kind:       "Secret",
+				},
```
droot: As per my understanding, as long as the client has correct scheme (in this case scheme for core-types which is set by default), we don't need to set typemeta. Has anything changed ?

PR #178, opened 2018-10-24
do not hold the lock after controller has started

droot:
This enables controller to be able to create new watches after it has started.
This is required for cases where you are writing some meta-controller
which discovers new types to watch during the reconcilation.


droot:
@DirectXMan12 did some analysis and found reconciliation workers are accessing thread-safe fields like `queue`, but would like you to take a look. The main motivation is to be able to setup watches after controller has started and holding on to mutex prevents that.

DirectXMan12:
I'll take a look when I'm a bit more coherent -- at a glance this looks ok, but I'd like to take a bit closer look

PR #182, opened 2018-10-25
:sparkles: lightweight dependency injection

pwittrock:


pwittrock:
@DirectXMan12 WDYT?

There are a couple use cases for this:

1. Injecting client-go generated clients and informers into Controllers for use with source.Informer
1. Injecting cluster-api actuators into Controllers instead of hard wiring them.

pwittrock:
@droot thoughts?

droot:
Took a look. In my mind, the primary use-case for dependency injection for a way to provide user-managed-objects like `Reconcilers`, `Sources` with the objects that only manager knows about. Injecting deps which belongs completely in the user-managed-space (which is second use-case you mentioned above), not super sure about the advantages. Discuss offline next week ?

pwittrock:
Here is an example of the problem we are trying to solve:

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/master/cmd/manager/main.go#L78

Take a look at how the actuator is passed around.

droot:
@pwittrock Took a look. our current scaffolding project structure isn't flexible enough for people to initialize `newReconciler` from the entry point `main` (the way stuff is wired using pkg `init()`) which is forcing us to solve this problem at the layer below (`controller-runtime/manager`) which I am not super sure about.

ironcladlou:
I found it easier to implement inversion of control by just skipping the SDK boilerplate and [using  controller-runtime as a library](https://github.com/openshift/cluster-ingress-operator/blob/f17c35e4af04db005f444e1bcf94969274895316/cmd/cluster-ingress-operator/main.go)(part of https://github.com/openshift/cluster-ingress-operator/pull/70).

In this context, I feel like the underlying problem is not lack of a dependency injection system (i.e. *more framework*) but instead a more IoC-friendly architecture.

That said, I acknowledge that IoC might not be important to everyone and there are many ways to go about it.

pwittrock:
@ironcladlou I mostly agree, though I think we still will need an answer for those who want to work within the framework as structured today.

pwittrock:
PTAL

PR #184, opened 2018-10-26
Adding documentation for using unstructured for client and controller

shawn-hurley:


pkg/client/example_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	c client.Client
+)
+
+// This example shows how to use the client with typed and unstrucurted objects to retrieve a objects.
```
droot: May be we can say `This example shows how to use the client to retrieve typed and unstructured objects`

Typo in unstructured.

pkg/client/example_test.go
```diff
@@ -0,0 +1,179 @@
...
+func ExampleClient_get() {
+	// Using a typed object.
+	p := &corev1.Pod{}
+	// c is a created client.
```
droot: I was wondering if we should have an example of constructing the client as well. Because sometime people wants to create a client without using the kubebuilder or manager machinery and they get stuck. Or we can probably have another example to create a client ? something that this gist captures --> https://gist.github.com/droot/3a60b97c8b48f8d470796302e87a01ea

pkg/client/example_test.go
```diff
@@ -0,0 +1,179 @@
...
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apps",
+		Kind:    "Deployment",
+		Version: "v1",
+	})
+	_ = c.Get(context.Background(), client.ObjectKey{
+		Namespace: "namespace",
+		Name:      "name",
+	}, u)
```
droot: Can we rename `p` to `pod`, it helps with readability without increasing verbosity.

pwittrock:
@shawn-hurley Would you update the GitBook as a follow up?

shawn-hurley:
I have not done that before, so I might need a little guidance or a link, but would happily do that :)

droot:
@shawn-hurley You can look at this PR https://github.com/kubernetes-sigs/kubebuilder/pull/465 for reference. There is this small [README](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/README.md) for previewing your changes locally.

PR #186, opened 2018-10-27
Rename InstallWebhookConfig and change it to a pointer

mengqiy:
This PR changes an option introduced in #174.

We always install webhook config prior v0.1.5 releases.
We introduced an option `InstallWebhookConfig` recently and was released on Oct 25 in v0.1.5.

If `InstallWebhookConfig` is not an pointer, it will be defaulted to `false` by golang.
That means existing user who are upgrading from previous controller-runtime version will be break, because they expect webhook config installer to be on.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -54,9 +54,10 @@ type ServerOptions struct {
 	// Client will be injected by the manager if not set.
 	Client client.Client
 
-	// InstallWebhookConfig controls if the server will automatically create webhook related objects
+	// DisableWebhookConfigInstaller controls if the server will automatically create webhook related objects
 	// during bootstrapping. e.g. webhookConfiguration, service and secret.
-	InstallWebhookConfig bool
+	// If false, the server will install the webhook config objects. It is defaulted to false.
+	DisableWebhookConfigInstaller *bool
 
 	// BootstrapOptions contains the options for bootstrapping the admission server.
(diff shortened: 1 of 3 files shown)


droot:
So one recommendation, we should invert the flag (--disable-webhook-config-install ? or something)

mengqiy:
@droot PTAL

mengqiy:
Renamed `InstallWebhookConfig` to `DisableInstallWebhookConfig` as a pointer. LMK if the naming makes sense to you.

pkg/webhook/server.go
```diff
@@ -54,9 +54,10 @@ type ServerOptions struct {
 	// Client will be injected by the manager if not set.
 	Client client.Client
 
-	// InstallWebhookConfig controls if the server will automatically create webhook related objects
+	// DisableInstallWebhookConfig controls if the server will automatically create webhook related objects
 	// during bootstrapping. e.g. webhookConfiguration, service and secret.
-	InstallWebhookConfig bool
+	// If false, the server will install the webhook config objects. It is defaulted to false.
+	DisableInstallWebhookConfig *bool
```
droot: Something about the name feels off while reading it. "Disable" and "Install" not going together. One option could be "DisableWebhookConfigAutoInstall" ?

mengqiy:
Changed it to `DisableWebhookConfigInstaller`.
PTAL

PR #188, opened 2018-10-31
Removes logic that checks for a "port already in use" error and instead retries on any error.

mjnichol2:
The "port already in use" error is only present in
the stderr logs of the separate process (etcd or apiserver) and will
not be surfaced in the go error.

Checking the stderr logs is an alternative solution, but it's
fragile (the message could change in the future and varies across
operating systems).

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -18,10 +18,8 @@ package envtest
 
 import (
 	"fmt"
-	"net"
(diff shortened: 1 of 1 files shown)


mjnichol2:
Please recheck CLA as I've properly registered now.

mjnichol2:
I signed it

PR #189, opened 2018-10-31
Removes logic that checks for a "port already in use" error

mjnichol2:
Instead, retry on any error.

The "port already in use" error is only present in
the stderr logs of the separate process (etcd or apiserver) and will
not be surfaced in the go error.

Checking the stderr logs is an alternative solution, but it's
fragile (the message could change in the future and varies across
operating systems).

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -18,10 +18,8 @@ package envtest
 
 import (
 	"fmt"
-	"net"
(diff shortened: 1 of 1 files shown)


mengqiy:
LGTM
Let's wait for some time before tagging it for merging in case @droot and @DirectXMan12 want to take a look.

PR #203, opened 2018-11-09
Set GVK when getting the object from the cache.

### LINKED ISSUE #202: TypeMeta : Kind + ApiVersion are empty when the object is processed the first time
## Bug

Original ticket : https://github.com/operator-framework/operator-sdk/issues/727

When I create a `Component`'s custom resource, the component is well created and the MetaData well defined (Kind, ApiVersion)

```
oc get cp/fruit-client -o yaml                
apiVersion: component.k8s.io/v1alpha1
kind: Component
metadata:
  name: fruit-client
  namespace: my-spring-app
spec:
  deployment: innerloop
  exposeService: true
  runtime: springboot
  version: 1.5.16
```

but when the reconcile's function of the controller fetch the Component, then the Kind, ApiVersion fields are empty

What is the root cause of this issue ?

screenshot 2018-11-09 12 27 43

## Environment

```
okd : 3.11 (= kubernetes v1.11)
operator-sdk : v0.1.1+git
```

## Project

Code of the API : https://github.com/snowdrop/component-operator/tree/operator-0.1.1/pkg/apis
Code of the controller : https://github.com/snowdrop/component-operator/blob/operator-0.1.1/pkg/controller/component/handler.go#L94

## Hypothesis

As described [here](https://github.com/operator-framework/operator-sdk/issues/727#issuecomment-437343992) I suspect that we have a `deepCopy` issue the first time that we get the object from the cache

shawn-hurley:
* Fixing up a TODO around DeepCopyInto

fixes #202

### DIFF
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -70,18 +70,41 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
 	}
 
 	// deep copy to avoid mutating cache
-	// TODO(directxman12): revisit the decision to always deepcopy
 	obj = obj.(runtime.Object).DeepCopyObject()
 
-	// Copy the value of the item in the cache to the returned value
-	// TODO(directxman12): this is a terrible hack, pls fix (we should have deepcopyinto)
-	outVal := reflect.ValueOf(out)
-	objVal := reflect.ValueOf(obj)
-	if !objVal.Type().AssignableTo(outVal.Type()) {
-		return fmt.Errorf("cache had type %s, but %s was asked for", objVal.Type(), outVal.Type())
+	// TODO: Add DeepCopyInto(runtime.Object) to the code generator.
+	// Create an interface for this and don't use reflection here.
+	oType := reflect.TypeOf(obj)
+	_, ok := oType.MethodByName("DeepCopyInto")
+	if ok {
+		err = deepCopyInto(obj, out)
+	} else {
+		err = copyTheValues(obj, out)
 	}
-	reflect.Indirect(outVal).Set(reflect.Indirect(objVal))
+	if err != nil {
+		return err
+	}
+	out.GetObjectKind().SetGroupVersionKind(c.groupVersionKind)
 
+	return nil
+}
+
+func deepCopyInto(in interface{}, out runtime.Object) error {
+	values := reflect.ValueOf(in).MethodByName("DeepCopyInto").Call([]reflect.Value{reflect.ValueOf(out)})
+	if len(values) > 0 {
+		return fmt.Errorf("could not deep copy the obj, unknown return values")
+	}
+	return nil
+}
+
+func copyTheValues(in interface{}, out runtime.Object) error {
+	// Copy the value of the item in the cache to the returned value
+	outVal := reflect.ValueOf(out)
+	inVal := reflect.ValueOf(in)
+	if !inVal.Type().AssignableTo(outVal.Type()) {
+		return fmt.Errorf("cache had type %s, but %s was asked for", inVal.Type(), outVal.Type())
+	}
+	reflect.Indirect(outVal).Set(reflect.Indirect(inVal))
 	return nil
 }
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -89,6 +89,13 @@ var _ = Describe("Informer Cache", func() {
 		knownPod1 = createPod("test-pod-1", testNamespaceOne, kcorev1.RestartPolicyNever)
 		knownPod2 = createPod("test-pod-2", testNamespaceTwo, kcorev1.RestartPolicyAlways)
 		knownPod3 = createPod("test-pod-3", testNamespaceTwo, kcorev1.RestartPolicyOnFailure)
+		podGVK := schema.GroupVersionKind{
+			Kind:    "Pod",
+			Version: "v1",
+		}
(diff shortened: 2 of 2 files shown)


cmoulliard:
I have applied the patch to test it with my project and when an existing resource is updated now, I get this error during the step to create the resource

```
ERRO[0021] deploymentconfigs.apps.openshift.io "fruit-client" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

cmoulliard:
Forget the error reported in my comment as it is related to a bug into my own code. Sorry.

pkg/cache/internal/cache_reader.go
```diff
@@ -73,14 +73,20 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
...
-	// Copy the value of the item in the cache to the returned value
-	// TODO(directxman12): this is a terrible hack, pls fix (we should have deepcopyinto)
-	outVal := reflect.ValueOf(out)
-	objVal := reflect.ValueOf(obj)
-	if !objVal.Type().AssignableTo(outVal.Type()) {
-		return fmt.Errorf("cache had type %s, but %s was asked for", objVal.Type(), outVal.Type())
+	oType := reflect.TypeOf(obj)
+	_, ok := oType.MethodByName("DeepCopyInto")
+	if ok {
+		reflect.ValueOf(obj).MethodByName("DeepCopyInto").Call([]reflect.Value{reflect.ValueOf(out)})
```
pwittrock: Do we need to check for an error here?

pkg/cache/internal/cache_reader.go
```diff
@@ -73,14 +73,20 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
 	// TODO(directxman12): revisit the decision to always deepcopy
 	obj = obj.(runtime.Object).DeepCopyObject()
 
-	// Copy the value of the item in the cache to the returned value
-	// TODO(directxman12): this is a terrible hack, pls fix (we should have deepcopyinto)
-	outVal := reflect.ValueOf(out)
-	objVal := reflect.ValueOf(obj)
-	if !objVal.Type().AssignableTo(outVal.Type()) {
-		return fmt.Errorf("cache had type %s, but %s was asked for", objVal.Type(), outVal.Type())
+	oType := reflect.TypeOf(obj)
```
pwittrock: We could probably do this without reflection if the deepcopy code generated this function.  That should be relatively straightforward to add.  Would you add a TODO for that?

pkg/cache/internal/cache_reader.go
```diff
@@ -73,14 +73,20 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
...
+	_, ok := oType.MethodByName("DeepCopyInto")
+	if ok {
```
pwittrock: Would you pull this logic into its own function and unit test it?  Verify both code paths work?

pkg/cache/internal/cache_reader.go
```diff
@@ -73,14 +73,20 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
...
+		// Copy the value of the item in the cache to the returned value
+		outVal := reflect.ValueOf(out)
+		objVal := reflect.ValueOf(obj)
+		if !objVal.Type().AssignableTo(outVal.Type()) {
+			return fmt.Errorf("cache had type %s, but %s was asked for", objVal.Type(), outVal.Type())
+		}
+		reflect.Indirect(outVal).Set(reflect.Indirect(objVal))
 	}
-	reflect.Indirect(outVal).Set(reflect.Indirect(objVal))
+	out.GetObjectKind().SetGroupVersionKind(c.groupVersionKind)
```
pwittrock: Would update any tests to verify this is being populated now.

pkg/cache/internal/cache_reader.go
shawn-hurley: Actually, I don't think we do, 

https://github.com/kubernetes/api/blob/master/admission/v1beta1/zz_generated.deepcopy.go#L29

Does that sound correct?

pkg/cache/internal/cache_reader.go
DirectXMan12: Is there not an interface for `DeepCopyInto`?

shawn-hurley: Hey, guys so I don't think that we can use an interface based on how generation works today. 
https://github.com/kubernetes/api/blob/master/core/v1/zz_generated.deepcopy.go#L30

If you notice that the `DeepCopyInto` does not use a `runtime.Object` as its input, it instead uses the type. I believe this means that we can not make a generic interface for `DeepCopyInto`. 

Am I missing something?

shawn-hurley:
Closing in favor of two PRs

pkg/cache/internal/cache_reader.go
pwittrock: Sorry, I was suggesting as a follow up we make a function that takes a runtime.Object and checks its type.  It should be trivial to wrap the DeepCopyInto function in another function that takes runtime.Object and does a type check.

Again, that would be a follow up and something we could put in the 1.14 version

pkg/cache/internal/cache_reader.go
pwittrock: The first one.

cmoulliard:
> Closing in favor of two PRs

What are the 2 PRs you mentioned ? Have they been merged and released ?

PR #204, opened 2018-11-09
fix nil stop value for source.Channel

### LINKED ISSUE #103: Channel Source is borked due to stop channel injection

DirectXMan12:
Creates a stop channel for the manager in New(), which will get passed to any
source.Channel instances that are added. When the manager's start method is
called and a new stop channel is passed in, that channel will be joined in a
goroutine with the manager's existing channel so that if the newer channel gets
closed, so will the manager's.

Fixes #103


DirectXMan12:
@mindprince @droot please take a quick look, but this should be all set

PR #213, opened 2018-11-14
Add scheme support to fake client

### LINKED ISSUE #137: Fake client assumes scheme.Scheme
Break out from https://github.com/kubernetes-sigs/controller-runtime/pull/106#discussion_r216406456

Fake client can only use `scheme.Scheme` as the scheme. It should allow using an arbitrary `Scheme`.

grantr:
Adds a `NewFakeClientWithScheme` function that takes a `*runtime.Scheme` as its first argument. `NewFakeClient` is now a wrapper around `NewFakeClientWithScheme`, passing it the default `scheme.Scheme`.

Fixes #137.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -169,6 +169,10 @@ func getGVKFromList(list runtime.Object, scheme *runtime.Scheme) (schema.GroupVe
 		return schema.GroupVersionKind{}, err
 	}
 
+	if gvk.Kind == "List" {
+		return schema.GroupVersionKind{}, fmt.Errorf("cannot derive GVK for generic List type %T (kind %q)", list, gvk)
+	}
+
 	if !strings.HasSuffix(gvk.Kind, "List") {
 		return schema.GroupVersionKind{}, fmt.Errorf("non-list type %T (kind %q) passed as output", list, gvk)
 	}
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -19,7 +19,9 @@ package fake
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -38,6 +40,7 @@ var (
 
 type fakeClient struct {
 	tracker testing.ObjectTracker
+	scheme  *runtime.Scheme
 }
 
 var _ client.Client = &fakeClient{}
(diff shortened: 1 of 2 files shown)


grantr:
As a bonus, I added the GVK support from https://github.com/kubernetes-sigs/controller-runtime/pull/106/files#diff-62fbcb10a0dc0bc5b0a1bba07b8eefb0R82 to this PR. This allows the fake client to work with typed Lists (e.g. `appsv1.DeploymentList`) instead of requiring `metav1.List` and `opts.Raw.TypeMeta`.The old style still works, but the new style is cleaner. I updated the fake client tests as an example.

pkg/client/fake/client.go
```diff
@@ -134,15 +154,33 @@ func (c *fakeClient) Status() client.StatusWriter {
...
 }
 
+func getGVKFromList(list runtime.Object, scheme *runtime.Scheme) (schema.GroupVersionKind, error) {
+	gvk, err := apiutil.GVKForObject(list, scheme)
+	if err != nil {
+		return schema.GroupVersionKind{}, err
+	}
+
+	if gvk.Kind == "List" {
+		return schema.GroupVersionKind{}, fmt.Errorf("cannot derive GVK for generic List type %T (kind %q)", list, gvk)
```
grantr: This check is necessary to detect the `corev1.List` and `metav1.List` type, which don't conform to the `SingularTypeList` format expected here. Without this check the returned GVK has an empty `Kind`.

It seems worthwhile to consolidate this derivation into a single helper function, since it's also used in varying forms in these places:
https://github.com/kubernetes-sigs/controller-runtime/blob/2ef903b2a5ed4cf8adde14a27994a291388a1789/pkg/client/client_cache.go#L60
https://github.com/kubernetes-sigs/controller-runtime/blob/df7c11e379403b7377ec52e54fd1baf587d3ac29/pkg/client/unstructured_client.go#L115
https://github.com/kubernetes-sigs/controller-runtime/blob/df7c11e379403b7377ec52e54fd1baf587d3ac29/pkg/cache/informer_cache.go#L67

pkg/client/fake/client.go
```diff
@@ -78,7 +89,16 @@ func (c *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.
 }
 
 func (c *fakeClient) List(ctx context.Context, opts *client.ListOptions, list runtime.Object) error {
-	gvk := opts.Raw.TypeMeta.GroupVersionKind()
+	gvk, err := getGVKFromList(list, c.scheme)
+	if err != nil {
+		// The old fake client required GVK info in Raw.TypeMeta, so check there
+		// before giving up
+		if opts.Raw.TypeMeta.APIVersion == "" || opts.Raw.TypeMeta.Kind == "" {
```
estroz: Check if `opts.Raw == nil`, as that's what was causing the panic previously.

grantr: Good spot @estroz, I accepted your suggestion.

PR #215, opened 2018-11-15
pkg/client/fake/client.go: get gvk without potential panic in List()

### LINKED ISSUE #168: Panic in fake client
I tried to write unit tests using the fake client and `List()` panics reliably when accessing `opts.Raw`:

https://github.com/kubernetes-sigs/controller-runtime/blob/ddf0390deebe496cb449b1b8540031da9236afa2/pkg/client/fake/client.go#L80-L82

Why `opts.Raw` must be filled (it's not necessary when using the real client) and why it's `TypeMeta` should have any value?

If I understand the code (and I probably don't), it expects `opts.Raw.TypeMeta.GroupVersionKind()` to be the kind of items in returned list, not the list itself (i.e. Pod instead of PodList). Such requirement looks very odd.

estroz:
Unless `opts.Raw.TypeMeta` is explicitly populated, `client.List()` panics. As noted in #168, population is not necessary outside of the fake client. This simple fix does essentially the same thing as [`unstructuredClient.List()`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/unstructured_client.go#L114-L117).

Closes #168

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


estroz:
Signed

sverrirsig:
Thank you! - List() now works perfectly 👍

grantr:
This is quite similar to #213. :smiley:

estroz:
@grantr yes your solution looks better. I'll close this if @DirectXMan12 @droot give the ok.

droot:
@estroz Thanks for submitting this fix. Took at a look at #213 which also addresses this issue and improves the code/tests also, so lets close this in favor of #213 .

PR #216, opened 2018-11-16
:book: Versioning and Branching Strategy

DirectXMan12:
This is mostly a proposal at this point.  I've started to outline the branch strategy we've loosely discussed in the past, with some motivations and discussion of alternatives.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,236 @@
+# Versioning and Branching in controller-runtime
+
+*NB*: this also applies to controller-tools.
+
+## TL;DR:
+
+### Users
+
+- We follow [Semantic Versioning (semver)](https://semver.org)
+- Use releases with your dependency management to ensure that you get
+  compatible code
+- The master branch contains all the latest code, some of which may break
+  compatibility (so "normal" `go get` is not reccomended)
+
+### Contributors
+
+- All code PR must be labeled with :bug: (patch fixes), :sparkles:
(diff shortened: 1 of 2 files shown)


DirectXMan12:
@droot @pwittrock PTAL and let me know what you think

VERSIONING.md
```diff
@@ -0,0 +1,206 @@
...
+accrued.  Since we don't intend to have a ton of these, the maintainers
+will evaluate when to do a major release as it comes up.
+
+### Exact Steps
+
+***TODO***
+
+### Breaking Changes
+
+Try to avoid breaking changes.  They make life difficult for users, who
```
pwittrock: Can we provide a `sed` operation to automatically fix any breaking changes with defaults that maintain the old behavior or provide some new default?

DirectXMan12: Probably want more intelligent than sed, but it should be possible to do so

VERSIONING.md
```diff
@@ -0,0 +1,236 @@
...
+- We follow [Semantic Versioning (semver)](https://semver.org)
+- Use releases with your dependency management to ensure that you get
+  compatible code
+- The master branch contains all the latest code, some of which may break
+  compatibility (so "normal" `go get` is not reccomended)
+
+### Contributors
+
+- All code PR must be labeled with :bug: (patch fixes), :sparkles:
+  (backwards-compatible features), or :warning: (breaking changes)
```
droot: Github PR template can help us here.

PR #220, opened 2018-11-22
Use typed lists in cert writer tests

grantr:
Now that #213 is in, these tests can be updated to use typed lists instead of `corev1.List`. I find the tests are clearer without the use of `Raw` hacks.

@mengqiy I'm assuming the use of `corev1.List` and JSON marshaling was a workaround for the issue with the fake client fixed in #213, but I could be wrong about that. You can decide what style you prefer. :smile:


mengqiy:
I definitely prefer the cleaner one :)
Thanks for the cleanup!
/lgtm

Issue #222, opened 2018-11-24
Allow setting "UseExistingCluster" via OS Env


ichekrygin:
Currently, `controller-runtime` does not support setting `UseExistingCluster` dynamically via environment variables (similarly to `envKubeAPIServerBin` or `envEtcdBin`, etc.). 

It would be very convenient and helpful to provide a uniform way in which users can dynamically set the option to `UseExistingCluster` via the environment.

DirectXMan12:
done

PR #223, opened 2018-11-24
:sparkles: Allow setting "UseExistingCluster" via environment

ichekrygin:
This PR is an attempt to address feature request: #222 
Non-breaking feature: sparkles (:sparkles:)

### DIFF
--- a/pkg/envtest/envtest_suite_test.go
+++ b/pkg/envtest/envtest_suite_test.go
@@ -33,7 +33,7 @@ var env *Environment
 
 var _ = BeforeSuite(func(done Done) {
 	logf.SetLogger(logf.ZapLoggerTo(GinkgoWriter, true))
-	env = &Environment{}
+	env = NewEnvironment()
 	_, err := env.Start()
 	Expect(err).NotTo(HaveOccurred())
 
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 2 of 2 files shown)


DirectXMan12:
/kind feature

Can you please tag the PR title according to https://github.com/kubernetes-sigs/controller-runtime/blob/master/VERSIONING.md#pr-process ?  Thanks!

DirectXMan12:
We tend to try and favor using structs directly with sane defaults instead of constructors, when possible.  Can you make this work with the existing setup?  If you need to make a breaking change, now would be the time to do it ;-)

ichekrygin:
@DirectXMan12 thank you for looking into this. 

I think I share the sentiment of using structs directly with sane defaults. In this case, however, the struct `Environment` already has a field `UseExistingCluster` which I am attempting to set on "client/user" behalf, hence the "constructor" implementation. 
The motivation is(was) not to "leak" environment variable value testing into the "client" code.

The alternative, "constructor-less" solution may look a bit ugly and convoluted (or maybe not): 

in `server.go`
```go
func UseExistingCluster() bool {
	return strings.ToLower(os.Getenv(envUseExistingCluster)) == "true"
}
```

in client code:

``` go
func TestMain(m *testing.M) {
	t := &envtest.Environment{
		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "..", "..", "config", "crds")},
                UseExistingCluster: envtest.UseExistingCluster(),
	}
...
```

@DirectXMan12 would you be open to "contsturctor" like solion? If so - I will update the PR with...
I guess: "Non-breaking feature: sparkles (:sparkles:)" - is it?

DirectXMan12:
So, the other option is changing `UseExistingCluster bool` to `UseExistingCluster *bool`, since most people probably don't want to set `UseExistingCluster` anyway, the awkwardness of builtin-pointers isn't super onerous.  Sometimes I miss being able to write `Option<bool>`...

ichekrygin:
I think I am coming to a realization that this is best handled in each individual project, rather than being built in. Closing.

PR #227, opened 2018-11-29
:sparkles: Allow user to define how create managed objects

### LINKED ISSUE #226: Overriding Values in Manager
As a user, I would like the ability to specify the cache and client that the manager injects into other objects. 

I am proposing that the manager options expose the new functions for client and cache.

```golang
NewCache  func(config *rest.Config, opts cache.Options) (cache.Cache, error)
NewClient func(config *rest.Config, options client.Options) (client.Client, error)
```

The defaults will stay the same, this will allow users to change the client.

I also believe that their should be a new method `NewDelegatingClient func(cache cache.Cache, client client.Client) (client.Client, error)` that will allow a user to define which delegating client they would like to use. The default should be the same as it is today IMO. 

/cc @DirectXMan12 @droot @ironcladlou

shawn-hurley:
* Allow the user to specify in the options how to create the client,
cache, and how to construct the delegating client.

fixes #226

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -118,14 +118,29 @@ type Options struct {
 	// for serving prometheus metrics
 	MetricsBindAddress string
 
+	// Functions to all for a user to customize the values that will be injected.
+
+	// NewCache is the function that will create the cache to be used
+	// by the manager. If not set this will use the default new cache function.
+	NewCache NewCacheFunc
+
+	// NewClient will create the client to be used by the manager.
+	// If not set this will create the default DelegatingClient that will
+	// use the cache for reads and the client for writes.
+	NewClient NewClientFunc
+
 	// Dependency injection for testing
-	newCache            func(config *rest.Config, opts cache.Options) (cache.Cache, error)
-	newClient           func(config *rest.Config, options client.Options) (client.Client, error)
 	newRecorderProvider func(config *rest.Config, scheme *runtime.Scheme, logger logr.Logger) (recorder.Provider, error)
 	newResourceLock     func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error)
 	newAdmissionDecoder func(scheme *runtime.Scheme) (types.Decoder, error)
 	newMetricsListener  func(addr string) (net.Listener, error)
 }
+
+// NewCacheFunc allows a user to define how to create a cache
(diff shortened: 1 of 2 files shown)


pkg/manager/manager.go
```diff
@@ -118,15 +118,27 @@ type Options struct {
 	// for serving prometheus metrics
 	MetricsBindAddress string
 
+	// Functions to all for a user to customize the values that will be injected.
+	NewCache            NewCacheFunc
+	NewClient           NewClientFunc
+	NewDelegatingClient NewDelegatingClientFunc
```
DirectXMan12: hmm... we try to avoid exposing a difference between these two at this level.  What's the usecase for separating them?  It's probably better/clearer to just have `NewClient`, which means `NewDelegatingClient`, and not leak the fact that we have a difference to the user/require the user to have that distinction.

shawn-hurley: The use case is that I want to not use the `DelegatingReader` for my client. I would like my client to use the cache for unstructured types. 

Because there is no other way to override this client this is the only way that I can think of to update it. Thoughts?

DirectXMan12: but there's not really a use for `NewClient` *except* as a step in constructing the delegating client, so we can just elide that and have:

```go
type Options struct {
    ...
    NewClient func(...) client.Client
    ...
}

func defaultNewClient(...) client.Client {
    directClient := newClient(...)
    return &DelegatingClient{
        Writer: directClient,
        Reader: cacheClient,
    }
}
```

shawn-hurley: I was still thinking about them as two things but I think you are completely correct, I will update now!

pkg/manager/manager.go
```diff
@@ -118,15 +118,27 @@ type Options struct {
 	// for serving prometheus metrics
 	MetricsBindAddress string
 
+	// Functions to all for a user to customize the values that will be injected.
+	NewCache NewCacheFunc
```
DirectXMan12: This needs godoc

shawn-hurley:
@DirectXMan12 I will take as a follow up to update the kubebuilder book around the manager. It looks like there is no section on the options to the manager, WDYT about a new section that describes the fields and examples on how to use them?

DirectXMan12:
@shawn-hurley sounds fine to me.

/lgtm
/approve

DirectXMan12:
When we get our new general dep injection story solve, we may end up tweaking this slightly, but this seems to solve a real usecase right now, and isn't particularly onerous.

@pwittrock FYI

PR #229, opened 2018-12-01
[WIP] :running: External k8s.io References Checker

DirectXMan12:
This introduces a tool for checking all our external references to
Kubernetes in our public API surface, so that we can make sure we don't
accidentally break people in an update due to changing Kubernetes API
surface.

### DIFF
--- a/hack/external-objects/README.md
+++ b/hack/external-objects/README.md
@@ -1,0 +1,34 @@
+External Imports Checker
+========================
+
+This tool knows how to check for external imports, to make sure that we
+don't accidentally break ourselves because of Kubernetes.  It's also
+a general framework for doing this kind of thing, so it could be adapted
+to general type signature checking, etc.
+
+Usage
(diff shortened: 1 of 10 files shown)


DirectXMan12:
cc @droot @mengqiy @pwittrock part one of the tooling

DirectXMan12:
(may need a bit of cleanup before merge, but wanted to get it up for review)

DirectXMan12:
Note to future self: it looks like there's a prebuilt version of something similar to this hiding in https://godoc.org/golang.org/x/exp/apidiff.  I can probably make use of it to simplify this code.

mengqiy:
@DirectXMan12 Just be aware of `x/exp/apidiff` is still experimental and not stable. That mean each time we upgrade to a newer go version, it may break.

DirectXMan12:
closing, but I'll bring this back at some point

PR #221, opened 2018-11-22
:sparkles: Update to Kubernetes 1.12.2

jimmidyson:
Fixes #211

This has no breaking changes, only changes to dependencies and internal changes to work with the new version.

PR #230, opened 2018-12-02
Update to Kubernetes 1.12.3

jimmidyson:
As requested in https://github.com/kubernetes-sigs/controller-runtime/pull/221#issuecomment-443056575

This has no breaking changes, only changes to dependencies and internal changes to work with the new version.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -931,8 +931,8 @@
     "storage/v1beta1",
   ]
   pruneopts = "UT"
-  revision = "d1dc89ebaebe945edbeff52a79f50c791a59ff87"
-  version = "kubernetes-1.12.2"
+  revision = "b503174bad5991eb66f18247f52e41c3258f6348"
+  version = "kubernetes-1.12.3"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


jimmidyson:
Wait - I just noticed there are no differences in vendored files... so what to do here? @DirectXMan12? It's a shame we have to reference a specific tagged version of Kubernetes (e.g. `kubernetes-1.12.3`) rather than a branch (e.g. `release-1.12`). How do you feel about referencing the branch instead?

/assign @directxman12

jimmidyson:
@DirectXMan12 @mengqiy Any time to review?

mengqiy:
We generally pin all the k8s dep to a released version instead of a branch. We upgrade the k8s dep about once per quarter, so these 2 approaches don't have much differences for us.

No breaking changes
/lgtm

jimmidyson:
OK thanks for explaining @mengqiy. This PR follows that convention and pins to released versions. Would be good to merge :)

jimmidyson:
I want to update controller-tools but need a tag to reference there. Can someone do that please?

PR #233, opened 2018-12-04
:sparkles: Expose client-go metrics

DirectXMan12:
This exposes the client-go metrics (client, workqueue, reflector)
metrics in the controller-runtime registry.

It also fixes a typo in the tests, and fixes our Gopkg.toml to prune some unnecessary requires (since I had to do a `dep ensure` anyway, and it was taking forever without the changes).

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -1,0 +1,299 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"net/url"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	reflectormetrics "k8s.io/client-go/tools/cache"
+	clientmetrics "k8s.io/client-go/tools/metrics"
+	workqueuemetrics "k8s.io/client-go/util/workqueue"
+)
+
+// this file contains setup logic to initialize the myriad of places
+// that client-go registers metrics.  We copy the names and formats
+// from Kubernetes so that we match the core controllers.
+
+var (
+	// client metrics
+
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_latency_seconds",
+			Help:    "Request latency in seconds. Broken down by verb and URL.",
+			Buckets: prometheus.ExponentialBuckets(0.001, 2, 10),
+		},
+		[]string{"verb", "url"},
+	)
+
+	requestResult = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "rest_client_requests_total",
+			Help: "Number of HTTP requests, partitioned by status code, method, and host.",
+		},
+		[]string{"code", "method", "host"},
+	)
+
+	// reflector metrics
+
+	// TODO(directxman12): update these to be histograms once the metrics overhaul KEP
+	// PRs start landing.
+
+	reflectorSubsystem = "reflector"
+
+	listsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: reflectorSubsystem,
+		Name:      "lists_total",
+		Help:      "Total number of API lists done by the reflectors",
+	}, []string{"name"})
+
+	listsDuration = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Subsystem: reflectorSubsystem,
+		Name:      "list_duration_seconds",
+		Help:      "How long an API list takes to return and decode for the reflectors",
+	}, []string{"name"})
+
+	itemsPerList = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Subsystem: reflectorSubsystem,
+		Name:      "items_per_list",
+		Help:      "How many items an API list returns to the reflectors",
+	}, []string{"name"})
+
+	watchesTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: reflectorSubsystem,
+		Name:      "watches_total",
+		Help:      "Total number of API watches done by the reflectors",
+	}, []string{"name"})
+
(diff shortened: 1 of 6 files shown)


DirectXMan12:
@brancz btw, this first is why I'm hesitant to call the current client-go setup easily usable.

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
...
+
+func init() {
+	registerClientMetrics()
+	registerReflectorMetrics()
+	registerWorkqueueMetrics()
+}
+
+// registerClientMetrics sets up the client latency metrics from client-go
+func registerClientMetrics() {
+	var ()
```
brancz: This looks like a leftover

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// registerWorkQueueMetrics sets up workqueue (other reconcile) metrics
+func registerWorkqueueMetrics() {
+	workqueuemetrics.SetProvider(workqueueMetricsProvider{})
+}
+
+// this section contains adapters, implementations, and other sundry organic, artisinally
+// hand-crafted syntax trees required to convince client-go that it actually wants to let
+// someone use its metrics.
```
droot: :)

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
...
+	return lastResourceVersion.WithLabelValues(name)
+}
+
+// Workqueue metrics (method #3 for client-go metrics),
+// copied (more-or-less directly) from k8s.io/kubernetes setup code
+// (which isn't anywhere in an easily-importable place).
+
+// NB(directxman12): these are changed to MustRegister from Register.  It's not clear why they weren't
+// MustRegister in the first place, except maybe to not bring down the controller if the metrics fail
+// to register (which shouldn't happen unless there's a duplicate metric).
```
droot: +1

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
...
+	Registry.MustRegister(watchDuration)
+	Registry.MustRegister(itemsPerWatch)
+	Registry.MustRegister(lastResourceVersion)
+
+	reflectormetrics.SetReflectorMetricsProvider(reflectorMetricsProvider{})
+}
+
+// registerWorkQueueMetrics sets up workqueue (other reconcile) metrics
+func registerWorkqueueMetrics() {
+	workqueuemetrics.SetProvider(workqueueMetricsProvider{})
```
droot: Does this means metrics from multiple controllers will get combined ?

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
droot: s/2016/2018 (someone should build github suggestion bot for this :))

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,266 @@
...
+
+func (workqueueMetricsProvider) NewRetriesMetric(name string) workqueuemetrics.CounterMetric {
+	retries := prometheus.NewCounter(prometheus.CounterOpts{
+		Subsystem: name,
+		Name:      "retries",
+		Help:      "Total number of retries handled by workqueue: " + name,
+	})
+	Registry.MustRegister(retries)
+	return retries
+}
```
droot: These workqueue metrics have some overlap with the existing metrics. Shall we remove the higher level controller_runtime metrics ?

pkg/metrics/client_go_adapter.go
DirectXMan12: yes, it is

pkg/metrics/client_go_adapter.go
DirectXMan12: ack

pkg/metrics/client_go_adapter.go
DirectXMan12: no, if you look at how the provider is implemented, there's a label for the controller name

pkg/metrics/client_go_adapter.go
DirectXMan12: There's things that look like duplicates, but it's not clear to me if they actually are completely duplicate.  Lets evaluate a bit.

DirectXMan12: I think we want to kill QueueLength in favor of depth, although the fact that the client-go metrics don't expose queue names as labels is strange to me.

@brancz do you have any insight on that one?

DirectXMan12: on this subject: https://github.com/kubernetes/kubernetes/issues/71165

DirectXMan12:
> I agree that this is quite a mouthful. Nonetheless the way you’re using the metrics registry here it could just as easily be injected and then this could just be a package in client-go.

Oh sure.  I think this particular model is just super clunky ;-)

brancz:
Do you think you could take that on as a follow up, to contribute that to client-go? This is the second time I am reviewing a PR that implements more or less the same thing twice (first time was Prometheus implementing these metrics itself for its Kubernetes service discovery). I imagine many more want to do the same thing. Preferably even first in client-go and then move forward with this PR, but I understand if you first want to make progress with this.

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,299 @@
...
+
+	retries = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem:   workQueueSubsystem,
+		Name:        "retries_total",
+		Help:        "Total number of retries handled by workqueue",
+	}, []string{"name"})
+
+	longestRunning = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: workQueueSubsystem,
+		Name:      "longest_running_processor_microseconds",
```
mengqiy: microseconds?
It seems this is not following the convention.

pkg/metrics/client_go_adapter.go
```diff
@@ -0,0 +1,299 @@
...
+
+func (workqueueMetricsProvider) NewWorkDurationMetric(name string) workqueuemetrics.SummaryMetric {
+	return workDuration.WithLabelValues(name)
+}
+
+func (workqueueMetricsProvider) NewRetriesMetric(name string) workqueuemetrics.CounterMetric {
+	return retries.WithLabelValues(name)
+}
+
+func (workqueueMetricsProvider) NewLongestRunningProcessorMicrosecondsMetric(name string) workqueuemetrics.SettableGaugeMetric {
```
mengqiy: Microseconds?

pkg/metrics/client_go_adapter.go
DirectXMan12: yeah, but the current code requires it atm.  it's fixed in future releases

pkg/metrics/client_go_adapter.go
DirectXMan12: the current code requires it atm (see above)

PR #235, opened 2018-12-05
⚠️ Implement Patch method

adracus:
This adds the `Patch` method to the `Client` interface as well as to the two implementations (`typedClient`, `unstructuredClient`).

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -130,6 +130,15 @@ func (c *client) Delete(ctx context.Context, obj runtime.Object, opts ...DeleteO
 	return c.typedClient.Delete(ctx, obj, opts...)
 }
 
+// Patch implements client.Client
+func (c *client) Patch(ctx context.Context, obj runtime.Object, patch Patch, opts ...PatchOptionFunc) error {
+	_, ok := obj.(*unstructured.Unstructured)
+	if ok {
+		return c.unstructuredClient.Patch(ctx, obj, patch, opts...)
+	}
+	return c.typedClient.Patch(ctx, obj, patch, opts...)
+}
+
 // Get implements client.Client
 func (c *client) Get(ctx context.Context, key ObjectKey, obj runtime.Object) error {
 	_, ok := obj.(*unstructured.Unstructured)
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -928,6 +928,7 @@
   input-imports = [
     "github.com/appscode/jsonpatch",
     "github.com/emicklei/go-restful",
+    "github.com/evanphx/json-patch",
     "github.com/ghodss/yaml",
     "github.com/go-logr/logr",
     "github.com/go-logr/logr/testing",
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -18,8 +18,11 @@ package client_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sync/atomic"
+
+	"k8s.io/apimachinery/pkg/types"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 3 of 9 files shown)


DirectXMan12:
/hold

Unfortunately, server-side apply hasn't landed yet, but I still feel like this is a really awkward structure to expose to users, especially as one that we have to support into perpetuity.

DirectXMan12:
/priority important-soon

had a few more people ask about this.  I think we may need to go forward with *some* patch-like thing.

adracus:
@DirectXMan12 do you have more input on this? I'd be up to change this PR to suit these needs.

pkg/client/client.go
```diff
@@ -121,6 +123,15 @@ func (c *client) Delete(ctx context.Context, obj runtime.Object, opts ...DeleteO
 	return c.typedClient.Delete(ctx, obj, opts...)
 }
 
+// Patch implements client.Client
+func (c *client) Patch(ctx context.Context, pt types.PatchType, data []byte, obj runtime.Object, subresources ...string) error {
```
coderanger: I think it would be better to make this a splat of PatchOptions like we have for ListOptions and DeleteOptions to allow for future expansion. We can then create an option method for subresource.

DirectXMan12: our past discussed subresource plan has been a `.Subresource` method, but I'm open for debate on that.  Otherwise agree that using a splat for subresource isn't the right way to go

DirectXMan12:
Ok, so, to move forward:

* no splat for subresource (see comment)
* move patch type and data into a single type that's an interface.  Something along the lines of

```go
type Patch interface {
    Type() types.PatchType
    Data(obj runtime.Object) []byte
}
```

This way, we can provide nicer abstractions on top of the patch method (e.g. passing `[]jsonpatch.Operation` instead of raw bytes, just using `client.Apply` instead of having to construct an apply patch, etc.

Also (this is a bit more just preference), I feel it looks more natural to have object then patch data for the arguments order.

/hold cancel

Sorry for all the foot-dragging on this :-(

pkg/client/client.go
adracus: I removed the  `subresources` splat and exchanged it for `PatchOptionFunc`. However, since there was no final clarity yet how subresources are handled, the `PatchOptions` are empty for now.
When we finally agree on how to do subresource handling in the future, we can incorporate it without changing the API.

adracus:
I implemented the spec almost as you described it:
For the `Patch` interface I removed the `obj *runtime.Object` part, since then in the implementation of it you'd have to do nasty casting again - see the discussion we had on `controllerutil.CreateOrUpdate`.

To also show that this approach is viable, I went ahead and implemented a `CreateMergePatch` method in `controllerutil` in which you see how to create patches without requiring the `obj` argument to the `Data` method.

DirectXMan12:
This is a breaking change since it modifies an interface, so I've edited the title.

> For the Patch interface I removed the `obj *runtime.Object` part,

I put the runtime.Object part in there so that we could cleanly have server-side apply work nicely:

```go
cl.Patch(ctx, &desiredObject, client.Apply)
```

It also means that you can do stuff like

```go
cl.Patch(ctx, &desiredObject, client.MergedFromOld(&oldObject))
```

Then, `client.Apply` should look something like

```go
type applyPatcher struct {}
func (p applyPatcher) Type() types.PatchType { return types.ApplyPatchType /* or whatever it actually is */ }
func (p applyPatcher) Data(obj runtime.Object) ([]byte, error) {
    return runtime.Encode(unstructured.UnstructuredJSONScheme, obj) /* maybe should actually be an encoder and not the unstructured scheme, but doesn't really matter too much */
}
```

adracus:
Alright I tried to reflect these changes, PTAL

DirectXMan12:
Looks good!  I'll follow up with some more patch types (json, server-side apply).

/lgtm
/approve

DirectXMan12:
pushed some fixes + options support, so we can get this merged.

DirectXMan12:
I'll do the following in a follow-up PR:

- merge patch --> strategic merge patch (you generally want the latter)
- server side apply (you actually want this if available)

DirectXMan12:
oof, that was a weird flake

PR #241, opened 2018-12-07
Minimal addon and declarative pattern implementations for addon operators

justinsb:
@johnsonj and I are working to create operators for kubernetes addons - we aim to use kubebuilder to replace the addon-manager in https://github.com/kubernetes/kubernetes/tree/master/cluster/addons.

We would like to contribute it to controller-runtime & kubebuilder.  This PR is the minimal first step, starting on an example for dashboard, and including the minimal functionality of the supporting library.  We are putting this into an `alpha/patterns` directory, but hopefully we can figure out if this is indeed the most appropriate place, and/or whether this is a good place to start.

We have three main components:

* The skeleton example dashboard controller
* A `declarative` pattern for implementing a reconciler based on a manifest that is manipulated and the `kubectl apply`d (later this will likely use server-side-apply)
* An `addon` pattern building on `declarative` which is opinionated as to how those manifests should be found, and some standard manipulations of the objects and surfacing of status that makes sense for addons.

Future PRs will include:

* Adding an Application CRD to wrap the manifest objects
* Setting labels & owner references on objects
* Incorporating "golden yaml" output testing, which has proved powerful and expressive for us
* Automatically watching deployed objects and surfacing status

It might be that declarative ends up as core functionality of controller-runtime, as the goal is to have it be much less opinionated than the `addon` pattern.

However, we though this was a good place to share the first step and figure out how best to proceed - input is greatly appreciated!

cc @DirectXMan12 @droot @mengqiy @pwittrock


justinsb:
I think that makes sense @DirectXMan12 ... we're doing it under the alpha/patterns directory, but clearly that would be better as a `controller-patterns` repo or similar.  While some of the patterns are likely applicable generally, we can (as you say) promote them as we demonstrate that.

That said, if sig-apimachinery's stance is that fewer repos is better, we can work with that, and just have bigger repos that are less focused.

(Fixed up the tests as well!)

alpha/patterns/addon/examples/dashboard-operator/.gitignore
```diff
@@ -0,0 +1,27 @@
+
```
droot: I am wondering if we should move the `examples` outside the `alpha/patterns/...` to maintain separation between the packages user import and the examples that they refer. This will also help tools because tools like `dep` considers examples to be part of the imported code.

alpha/patterns/addon/examples/dashboard-operator/Gopkg.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[prune]
+  go-tests = true
+
+
+# STANZAS BELOW ARE GENERATED AND MAY BE WRITTEN - DO NOT MODIFY BELOW THIS LINE.
+
+[[constraint]]
+  name="sigs.k8s.io/controller-tools"
+  version="v0.1.1"
```
droot: Does the presence of this file introduces deps on controller-tools ? That can introduce the deadlock between controller-tools and controller-runtime.

alpha/patterns/addon/pkg/loaders/fs.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/alpha/patterns/addon"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var FlagChannel = "./channels"
+
+func init() {
+	// TODO: Yuk - global flags are ugly
+	flag.StringVar(&FlagChannel, "channel", FlagChannel, "location of channel to use")
```
droot: Oh no!!!!

alpha/patterns/declarative/options.go
```diff
@@ -0,0 +1,151 @@
+package declarative
```
droot: We should add pkg level `doc.go` explaining this pattern and the types defined below.

alpha/patterns/declarative/reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	//applicationclient "github.com/kubernetes-sigs/application/pkg/client/clientset/versioned"
+)
+
+var _ reconcile.Reconciler = &Reconciler{}
+
+type Reconciler struct {
```
droot: We should explain this in detail.

johnsonj:
Code review-wise I think we have a few comments to address that should be straightforward. 

The open question is if this PR is destin for this repo or another. Any progress on that decisions @DirectXMan12 @droot ?

mengqiy:
closing this now since we have a separate repo for the add-on pattern.

PR #250, opened 2018-12-13
WIP: Support for addon pattern

justinsb:


droot:
Given that we have a separate repo for it, its safe to close this now.

PR #251, opened 2018-12-14
:sparkles: Fix roundtripping issue when genearting JSON patch

mengqiy:
fixes https://github.com/kubernetes-sigs/kubebuilder/issues/510
The change should be backward compatible 

An alternative is in #256.

PR #256, opened 2018-12-14
:warning: introduce PatchResponseFromRaw and drop PatchResponse

mengqiy:
fixes kubernetes-sigs/kubebuilder#510

This PR supersedes #251.
This PR introduces PatchResponseFromRaw and drops PatchResponse.

This is a breaking change, so should released in v0.2.0.


pkg/webhook/admission/response.go
```diff
@@ -54,7 +56,23 @@ func ValidationResponse(allowed bool, reason string) types.Response {
...
+	return types.Response{
+		Patches: patches,
+		Response: &admissionv1beta1.AdmissionResponse{
+			Allowed:   true,
+			PatchType: func() *admissionv1beta1.PatchType { pt := admissionv1beta1.PatchTypeJSONPatch; return &pt }(),
+		},
+	}
+}
+
+// PatchResponse is DEPRECATED. Use PatchResponseFromRaw instead.
```
droot: It will be good to highlight why is this being deprecated (may be refer to the issue which explains the roundtripping) ?

pkg/webhook/admission/response.go
```diff
@@ -54,7 +56,23 @@ func ValidationResponse(allowed bool, reason string) types.Response {
 	return resp
 }
 
-// PatchResponse returns a new response with json patch.
+// PatchResponseFromRaw takes 2 byte arrays and returns a new response with json patch.
+func PatchResponseFromRaw(original, current []byte) types.Response {
```
droot: I prefer this over the other PR which takes three params `original, current, originalRawBytes` which could get confusing for the end user and also not help user avoid the round-tripping issue.

DirectXMan12:
agree with @droot's comments.

mengqiy:
@droot @DirectXMan12 PTAL

mengqiy:
This should be ready to go.
@droot Can you apply the label?

PR #259, opened 2018-12-18
:sparkles: allow user to specify metrics

### LINKED ISSUE #258: controller histogram buckets configuration
Metrics are created for the controller's workqueue, using Prometheus's [default buckets](https://github.com/kubernetes-sigs/controller-runtime/blob/b497fd58b9d76e0af8a4cba21c595d13cf094307/pkg/internal/controller/metrics/metrics.go#L47-L57). Unfortunately, the default buckets are poorly chosen for event processing.

> The default buckets are tailored to broadly measure the response time (in seconds) of a network service.

This can easily result in metrics that are extremely coarse. In a controller I was working on today, every single reconcile was faster than the smallest bucket:

```
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.005"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.01"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.025"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.05"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.1"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.25"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="0.5"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="1"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="2.5"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="5"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="10"} 31
controller_runtime_reconcile_time_seconds_bucket{controller="application",le="+Inf"} 31
```

It is possible to change the default buckets by modifying [DefBuckets](https://godoc.org/github.com/prometheus/client_golang/prometheus#pkg-variables) in `init`, as my `init` will be called after the `DefBuckets` variable has been initialized, but before the controllermetrics package `init`. But this is a very heavy handed brush, changing the defaults of all histograms.

I propose two paths forward:

1. Allow the user to pass their own metrics to the controller, with the current collectors used if the none are provided.
2. Alternatively, move the current package outside of the internal package, This will allow users to call [Unregister](https://godoc.org/github.com/prometheus/client_golang/prometheus#Registry.Unregister) and then assign a replacement.

My preference would be the first option, as the change to the system is would be more easily understood.

mengqiy:
fixes #258

The tests are failing because currently different test cases try to register same metrics in the shared Registry. I will fix tests if we decide to proceed with this approach.

### DIFF
--- a/pkg/internal/controller/metrics/metrics.go
+++ b/pkg/internal/controller/metrics/metrics.go
@@ -45,13 +45,10 @@ var (
 )
 
 func init() {
-	metrics.Registry.MustRegister(
+	metrics.DefaultMetrics = append(
+		metrics.DefaultMetrics,
 		QueueLength,
 		ReconcileErrors,
 		ReconcileTime,
-		// expose process metrics like CPU, Memory, file descriptor usage etc.
-		prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}),
-		// expose Go runtime metrics like GC stats, memory stats etc.
-		prometheus.NewGoCollector(),
 	)
 }
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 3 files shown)


DirectXMan12:
This case made me a bit nervous when we first introduced this.  I'm not sure this is the right way to solve it, though...

terinjokes:
@DirectXMan12 @mengqiy Is there something I can help with from a user use-case?

I understand the desire elsewhere to make it really easy to have working metrics, but registering default metrics, even if the user provides their own collectors, makes it extremely difficult to use controller-runtime in a larger project.

DirectXMan12:
@terinjokes see my comment on #258 for my comments on the buckets.  For the default collectors, go runtime metrics, etc we might just want to disable them.  Which are you taking about?

mengqiy:
Closing this until we decided how we tackle it.

PR #260, opened 2018-12-18
:bug: fix issue when webhook server refreshing cert

### LINKED ISSUE #191: RefreshCert should be based on the cert expire time, not after 180 days
A server restart will reset the timer but not the certificate. Unless we refresh certificate every time we start the server. 

https://github.com/kubernetes-sigs/controller-runtime/blob/a8ea2056444a5d74d7408e4b8798cbe63b14066b/pkg/webhook/server.go#L226

### LINKED ISSUE #192: RefreshCert loop will quit after first refreshing
```golang
func (s *Server) run(stop <-chan struct{}) error {
	srv := &http.Server{
		Addr:    fmt.Sprintf(":%v", s.Port),
		Handler: s.sMux,
	}
	errCh := make(chan error)
	serveFn := func() {
// FIXME: This return will return after first shutdown
		errCh <- srv.ListenAndServeTLS(path.Join(s.CertDir, writer.ServerCertName), path.Join(s.CertDir, writer.ServerKeyName))
	}

	go serveFn()
	for {
		// TODO(mengqiy): add jitter to the timer
		// Could use https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Jitter
		timer := time.Tick(6 * 30 * 24 * time.Hour)
		select {
		case <-timer:
			changed, err := s.RefreshCert()
			if err != nil {
				log.Error(err, "encountering error when refreshing the certificate")
				return err
			}
			if !changed {
				continue
			}
			log.Info("server is shutting down to reload the certificates.")
			err = srv.Shutdown(context.Background())
			if err != nil {
				log.Error(err, "encountering error when shutting down")
				return err
			}
			go serveFn()
		case <-stop:
			return nil
		case e := <-errCh:
// FIXME: this will break from the loop
			return e
		}
	}
}
```

mengqiy:
fixes #191 
fixes #192

Tests are on the way.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -27,6 +27,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
@@ -35,6 +36,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/types"
 )
+
+// default interval for checking cert is 90 days (~3 months)
+var defaultCertRefreshInterval = 3 * 30 * 24 * time.Hour
 
 // ServerOptions are options for configuring an admission webhook server.
 type ServerOptions struct {
@@ -128,6 +132,9 @@ type Server struct {
 	// manager is the manager that this webhook server will be registered.
 	manager manager.Manager
 
+	// httpServer is the actual server that serves the traffic.
+	httpServer *http.Server
+
 	once sync.Once
 }
 
@@ -209,21 +216,21 @@ func (s *Server) Start(stop <-chan struct{}) error {
 	return s.run(stop)
 }
 
-func (s *Server) run(stop <-chan struct{}) error {
-	srv := &http.Server{
-		Addr:    fmt.Sprintf(":%v", s.Port),
-		Handler: s.sMux,
-	}
+func (s *Server) run(stop <-chan struct{}) error { // nolint: gocyclo
 	errCh := make(chan error)
 	serveFn := func() {
-		errCh <- srv.ListenAndServeTLS(path.Join(s.CertDir, writer.ServerCertName), path.Join(s.CertDir, writer.ServerKeyName))
+		s.httpServer = &http.Server{
+			Addr:    fmt.Sprintf(":%v", s.Port),
+			Handler: s.sMux,
+		}
+		log.Info("starting the webhook server.")
+		errCh <- s.httpServer.ListenAndServeTLS(path.Join(s.CertDir, writer.ServerCertName), path.Join(s.CertDir, writer.ServerKeyName))
 	}
 
+	shutdownHappend := false
+	timer := time.Tick(wait.Jitter(defaultCertRefreshInterval, 0.1))
 	go serveFn()
 	for {
-		// TODO(mengqiy): add jitter to the timer
-		// Could use https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Jitter
-		timer := time.Tick(6 * 30 * 24 * time.Hour)
 		select {
 		case <-timer:
 			changed, err := s.RefreshCert()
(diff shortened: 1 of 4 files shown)


anfernee:
/lgtm

it doesn't fix #191 tho.

mengqiy:
Made some additional changes in pkg/webhook/server.go and added tests.

mengqiy:
> it doesn't fix #191 tho.

@anfernee Explained in #191 :)

@anfernee @droot PTAL

pkg/webhook/server_test.go
```diff
@@ -0,0 +1,148 @@
+/*
```
anfernee: can you add a test case to test the case where the server exceeds cert refresh interval? you can change `defaultCertRefreshInterval` to a smaller interview. I am worried that if you start a server on the same port immediately after shutdown, you will have port conflict.

pkg/webhook/server.go
```diff
@@ -232,19 +239,28 @@ func (s *Server) run(stop <-chan struct{}) error {
...
 			if err != nil {
 				log.Error(err, "encountering error when shutting down")
 				return err
 			}
+			timer = time.Tick(wait.Jitter(defaultCertRefreshInterval, 0.1))
 			go serveFn()
 		case <-stop:
 			return nil
 		case e := <-errCh:
-			return e
```
anfernee: ideally if you start server here, it should just work.

pkg/webhook/server.go
```diff
@@ -232,19 +239,28 @@ func (s *Server) run(stop <-chan struct{}) error {
 				return err
 			}
 			if !changed {
+				log.Info("no need to reload the certificates.")
 				continue
 			}
 			log.Info("server is shutting down to reload the certificates.")
-			err = srv.Shutdown(context.Background())
+			shutdownHappend = true
+			err = s.httpServer.Shutdown(context.Background())
```
anfernee: My understanding is `Shutdown` is async, meaning return of shutdown doesn't necessarily mean the server is already done. But return of `ListenAndServeTLS` guarantees it.

pkg/webhook/server.go
```diff
@@ -232,19 +239,28 @@ func (s *Server) run(stop <-chan struct{}) error {
...
 			if err != nil {
 				log.Error(err, "encountering error when shutting down")
 				return err
 			}
+			timer = time.Tick(wait.Jitter(defaultCertRefreshInterval, 0.1))
 			go serveFn()
```
anfernee: this might have port conflict.

pkg/webhook/server.go
mengqiy: IMHO if there is an unexpected error, we should surface the error instead of keeping retrying.

pkg/webhook/server_test.go
mengqiy: IIUC 
https://github.com/kubernetes-sigs/controller-runtime/pull/260/files#diff-8b0412ec4fd52af1419bd0fcd0f2e101R121 is doing what you ask. The server restarts multiple times and rotates the certs.

pkg/webhook/server.go
mengqiy: > Shutdown gracefully shuts down the server without interrupting any active connections. Shutdown works by first closing all open listeners, then closing all idle connections, and then waiting indefinitely for connections to return to idle and then shut down

Per https://golang.org/pkg/net/http/#Server.Shutdown, it is graceful shutdown and synchronized call. Did I miss anything?

pkg/webhook/server.go
mengqiy: I think the question here is that shutting down the server will unbind the port, but if it will complete in time before the server tries to bind the same port next time.
I have been searching online for this question for quite a while. But I'm still not sure what the 100% correct thing to do here :/
It seems when the Listener is closed, it should have unbinded the port.
Some tests I added in this PR rotates the cert and reloads the server for multiple times. I haven't seen any issue about port conflict.

I'm open to suggestions if you have a better solution. If not, I will probably merge it as is :)

pkg/webhook/server_test.go
```diff
@@ -0,0 +1,160 @@
...
+				ServerOptions: ServerOptions{
+					Port: int32(port),
+					BootstrapOptions: &BootstrapOptions{
+						Host: &cn,
+					},
+				},
+			}
+
+			cg := &generator.SelfSignedCertGenerator{}
+			s.CertDir, err = ioutil.TempDir("/tmp", "controller-runtime-")
```
droot: Are we doing cleanup of this temp dir ?

mengqiy: Good catch.
Added code to cleanup the temp dir.
https://github.com/kubernetes-sigs/controller-runtime/pull/260/files#diff-8b0412ec4fd52af1419bd0fcd0f2e101R159

pkg/webhook/server.go
anfernee: I am confused of those 2 methods, you are right. pls ignore it.

PR #264, opened 2018-12-20
:sparkles: Improve builder UX

pwittrock:


example_test.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// * Start the application.
+func ExampleBuilder() {
+	manager, err := controllers.NewManager(controllers.GetConfigOrDie(), controllers.Options{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = controllers.
+		NewControllerBuilder(manager). // Create the Controller
```
DirectXMan12: :bike: :house: : maybe just `NewControllerIn`?  Reads a bit more fluently

example_test.go
```diff
@@ -0,0 +1,91 @@
...
+		ForType(&appsv1.ReplicaSet{}). // ReplicaSet is the Application API
+		Owns(&corev1.Pod{}).           // ReplicaSet owns Pods created by it
+		Build(&ReplicaSetReconciler{Client: manager.GetClient()})
```
DirectXMan12: what's up with the return of explicit wiring?

pkg/builder/build.go
```diff
@@ -51,35 +44,59 @@ type Builder struct {
...
 }
 
+type watchRequest struct {
+	src          source.Source
+	eventhandler handler.EventHandler
+}
+
+// Watch exposes the lower-level Controller Watch functions through the builder.  Consider using
+// Owns or ForType instead of Watch directly.
+func (b *Builder) Watch(src source.Source, eventhandler handler.EventHandler) *Builder {
```
DirectXMan12: nit: follow the same phrasing as `ForType` or `Owns` here -- `Watches` or something similar

pkg/builder/build.go
```diff
@@ -51,35 +44,59 @@ type Builder struct {
...
+	b.watchRequest = append(b.watchRequest, watchRequest{src: src, eventhandler: eventhandler})
+	return b
+}
+
 // WithConfig sets the Config to use for configuring clients.  Defaults to the in-cluster config or to ~/.kube/config.
+// Deprecated - set through the Manager instead.
```
DirectXMan12: nit: we should prob follow the standard library convention: https://rakyll.org/deprecated/

pkg/builder/build.go
```diff
@@ -114,12 +131,10 @@ func (b *Builder) Build(r reconcile.Reconciler) (manager.Manager, error) {
 		return nil, err
 	}
 
-	a := &application{mgr: b.mgr, ctrl: b.ctrl}
-
 	// Reconcile type
 	s := &source.Kind{Type: b.apiType}
```
DirectXMan12: blech, we should fix these single-letter variables in a follow-up

pkg/builder/build_test.go
```diff
@@ -206,6 +209,104 @@ var _ = Describe("application", func() {
 			close(done)
 		}, 10)
 	})
+
+	Describe("Start with Controller", func() {
+		depName := "deploy-name-2"
+		It("should Reconcile objects", func(done Done) {
+			By("Creating the application")
```
DirectXMan12: these tests effectively do the same thing after setting up the controller.  We should prob refactor out into a help above for easier maintenance

DirectXMan12: or invert and do something like

```go
func DescribeAController(setupTheController func() manager.Manager) {
   It("should reconcile objects", func(done Done) {
       mgr := setupTheController()
       // ...
   })
}
```

pkg/builder/example_test.go
```diff
@@ -19,39 +19,41 @@ package builder_test
 import (
 	"context"
 	"fmt"
-	"os"
+	"log"
```
DirectXMan12: why're we reverting back to std log?

alias.go
```diff
@@ -0,0 +1,80 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller_runtime
+
```
droot: We should add a comment explaining motivation for this file `alias.go`.

alias.go
```diff
@@ -0,0 +1,80 @@
...
+	return config.GetConfigOrDie()
+}
+
+func GetConfig() (*rest.Config, error) {
+	return config.GetConfig()
+}
+
+func NewManager(config *rest.Config, options Options) (manager.Manager, error) {
+	return manager.New(config, options)
+}
```
droot: nit: some fns/vars have comments and some don't, so reading the file feels inconsistent. 

Also this will impact the godocs as well.

example_test.go
```diff
@@ -0,0 +1,91 @@
...
+// * Start the application.
+func ExampleBuilder() {
+	manager, err := controllers.NewManager(controllers.GetConfigOrDie(), controllers.Options{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = controllers.
+		NewControllerBuilder(manager). // Create the Controller
+		ForType(&appsv1.ReplicaSet{}). // ReplicaSet is the Application API
```
droot: wondering if we should drop `Type` ?

example_test.go
pwittrock: I think this works for this case.  It is pretty simple.

pkg/builder/build.go
pwittrock: Done

pkg/builder/build.go
pwittrock: done

example_test.go
pwittrock: Done

alias.go
pwittrock: Done

alias.go
pwittrock: Done.  Removed all comments

example_test.go
pwittrock: Done

pkg/builder/build.go
pwittrock: Done

pkg/builder/build_test.go
pwittrock: Done

pwittrock:
PTAL

example_test.go
DirectXMan12: ack

example_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	controllers "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
DirectXMan12: can just use `controllers.Log`, right?

pkg/controller/controller.go
```diff
@@ -78,15 +78,15 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error)
...
+		Do:       options.Reconciler,
+		Cache:    mgr.GetCache(),
+		Config:   mgr.GetConfig(),
+		Scheme:   mgr.GetScheme(),
+		Client:   mgr.GetClient(),
+		Recorder: mgr.GetRecorder(name),
+		Queue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), name),
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
-		Name:                    name,
+		Name: name,
```
DirectXMan12: did these get unformatted?

pkg/source/source.go
```diff
@@ -241,7 +241,7 @@ func (cs *Channel) syncLoop() {
 	}
 }
 
-// Informer is used to provide a source of events originating inside the cluster from Watches (e.g. Pod Create)
+// Informer is used to provide a source of events originating inside the cluster from Watch (e.g. Pod Create)
```
DirectXMan12: ??

alias.go
```diff
@@ -0,0 +1,126 @@
...
+	SetControllerReference = controllerutil.SetControllerReference
+
+	// SetupSignalHandler registered for SIGTERM and SIGINT. A stop channel is returned
+	// which is closed on one of these signals. If a second signal is caught, the program
+	// is terminated with exit code 1.
+	SetupSignalHandler = signals.SetupSignalHandler
+
+	// Log is the base logger used by kubebuilder.  It delegates
+	// to another logr.Logger.  You *must* call SetLogger to
+	// get any actual logging.
```
DirectXMan12: might want to note that `SetLogger` isn't here ;-)

pkg/controller/controller.go
pwittrock: these are formatted with `go fmt ./pkg/...`

alias.go
pwittrock: Added

example_test.go
pwittrock: Done

pwittrock:
PTAL.  I am going on break.  Can we merge this as is an follow up on any minor / formatting issues.

pkg/builder/build_test.go
```diff
@@ -119,94 +123,157 @@ var _ = Describe("application", func() {
...
-				ObjectMeta: metav1.ObjectMeta{
-					Namespace: "default",
-					Name:      "rs-name",
-					Labels:    dep.Spec.Selector.MatchLabels,
-					OwnerReferences: []metav1.OwnerReference{
+			bldr := ControllerManagedBy(m).
+				For(&appsv1.Deployment{}).
+				Watches( // Equivalent of Owns
+					&source.Kind{Type: &appsv1.ReplicaSet{}},
+					&handler.EnqueueRequestForOwner{OwnerType: &appsv1.Deployment{}, IsController: true})
```
droot: Looking at this example, builder for controller appears to be a very *thin* improvement over current `controller` abstraction because users have to understand `manager`, `watches` and `handlers` to use these. Am I missing something ?

pkg/builder/example_test.go
```diff
@@ -21,41 +21,50 @@ import (
...
+	if err != nil {
+		log.Error(err, "could not create manager")
+		os.Exit(1)
+	}
+
+	_, err = builder.
+		ControllerManagedBy(mgr).  // Create the ControllerManagedBy
+		For(&appsv1.ReplicaSet{}). // ReplicaSet is the Application API
+		Owns(&corev1.Pod{}).       // ReplicaSet owns Pods created by it
+		Build(&ReplicaSetReconciler{})
```
droot: I see it now (ignore my previous comment). Simple use-case is looking simple now :)

pkg/builder/build.go
```diff
@@ -32,145 +32,185 @@ import (
...
-}
-
 // Supporting mocking out functions for testing
 var getConfig = config.GetConfig
 var newController = controller.New
 var newManager = manager.New
 var getGvk = apiutil.GVKForObject
 
-// Builder builds an Application Controller (e.g. Operator) and returns a manager.Manager to start it.
+// Builder builds an Application ControllerManagedBy (e.g. Operator) and returns a manager.Manager to start it.
```
DirectXMan12: find-replace issue

pkg/builder/build_test.go
DirectXMan12: this is just an example of using watches directly.  We need to test the "I want .Owns and one more complicated thing" -- the test just happens to use the Owner enqueuer for convienience

pkg/builder/build.go
pwittrock: Done

pwittrock:
PTAL

PR #267, opened 2018-12-28
:warning: adding multi namespaces cache

shawn-hurley:
This will add the ability to set up a multi-namespace cache. This also has a backward incompatible change in the `Informers` interface which effects the `cache` interface. 

This is based on the discussion from the BoF.
/cc @DirectXMan12 @mhrivnak @hasbro17 

Wanted to get some early feedback before starting on tests.

### DIFF
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -1,0 +1,220 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewCacheFunc - Function for creating a new cache from the options and a rest config
+type NewCacheFunc func(config *rest.Config, opts Options) (Cache, error)
+
+// MultiNamespacedCacheBuilder - Builder function to create a new multi-namespaced cache.
+// This will scope the cache to a list of namespaces. Listing for all namespaces
+// will list for all the namespaces that this knows about.
+func MultiNamespacedCacheBuilder(namespaces []string) NewCacheFunc {
+	return func(config *rest.Config, opts Options) (Cache, error) {
+		opts, err := defaultOpts(config, opts)
+		if err != nil {
+			return nil, err
+		}
+		caches := map[string]Cache{}
+		for _, ns := range namespaces {
+			opts.Namespace = ns
+			c, err := New(config, opts)
+			if err != nil {
+				return nil, err
+			}
+			caches[ns] = c
+		}
+		return &multiNamespaceCache{namespaceToCache: caches, Scheme: opts.Scheme}, nil
+	}
+}
+
+// multiNamespaceCache knows how to handle multiple namespaced caches
+// Use this feature when scoping permissions for your
+// operator to a list of namespaces instead of watching every namespace
+// in the cluster.
+type multiNamespaceCache struct {
+	namespaceToCache map[string]Cache
+	Scheme           *runtime.Scheme
+}
+
+var _ Cache = &multiNamespaceCache{}
+
+// Methods for multiNamespaceCache to conform to the Informers interface
+func (c *multiNamespaceCache) GetInformer(obj runtime.Object) (Informer, error) {
+	informers := map[string]Informer{}
+	for ns, cache := range c.namespaceToCache {
+		informer, err := cache.GetInformer(obj)
+		if err != nil {
+			return nil, err
+		}
+		informers[ns] = informer
+	}
+	return &multiNamespaceInformer{namespaceToInformer: informers}, nil
+}
+
+func (c *multiNamespaceCache) GetInformerForKind(gvk schema.GroupVersionKind) (Informer, error) {
+	informers := map[string]Informer{}
+	for ns, cache := range c.namespaceToCache {
+		informer, err := cache.GetInformerForKind(gvk)
+		if err != nil {
+			return nil, err
+		}
+		informers[ns] = informer
+	}
+	return &multiNamespaceInformer{namespaceToInformer: informers}, nil
+}
+
+func (c *multiNamespaceCache) Start(stopCh <-chan struct{}) error {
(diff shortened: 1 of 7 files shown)


pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,136 @@
...
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type multiNamespaceCache struct {
+	namespaceToCache map[string]Cache
+}
+
+// NewMultiNamespacedCache - retrieve the new function for a multinamespaced cache.
+func NewMultiNamespacedCache(namespaces []string) func(config *rest.Config, opts Options) (Cache, error) {
```
DirectXMan12: `NewMultiNamespaceCacheBuilder` or something, to imply that this doesn't return a multi-namespace-cache directly?

pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,136 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type multiNamespaceCache struct {
```
DirectXMan12: needs godoc

pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (c *multiNamespaceCache) List(ctx context.Context, opts *client.ListOptions, list runtime.Object) error {
+	if opts.Namespace != corev1.NamespaceAll {
+		cache, ok := c.namespaceToCache[opts.Namespace]
+		if !ok {
+			return fmt.Errorf("unable to get: %v because of unknown namespace for the cache", opts.Namespace)
+		}
+		return cache.List(ctx, opts, list)
+	}
+	//TODO(shawn-hurley): handle this use case by listing for each namespace in the cache.
```
DirectXMan12: please handle before merge (just adding this so we don't miss it later)

pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,136 @@
...
+	return fmt.Errorf("unable to list for an all namespaces")
+}
+
+type multiNamespaceInformer struct {
```
DirectXMan12: godoc

sebgl:
Drive-by comment: I think this is great and would love to have it merged 🎉

pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		allItems.Items = append(allItems.Items, items.Items...)
+		// The last list call should have the most correct resource version.
+		allItems.Object = items.Object
+	}
+	data, err := allItems.MarshalJSON()
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(data, list)
```
mengqiy: Can we avoid marshal and then unmarshal to improve performance?

pkg/cache/multi_namespace_cache.go
```diff
@@ -0,0 +1,212 @@
...
+	for ns, cache := range c.namespaceToCache {
+		go func(ns string, cache Cache) {
+			err := cache.Start(stopCh)
+			if err != nil {
+				log.Error(err, "multinamespace cache failed to start namespaced informer", "namespace", ns)
+			}
+		}(ns, cache)
+	}
+	<-stopCh
+	return nil
```
mengqiy: Should the `multiNamespaceCache` main goroutine wait for each cache goroutine to finish before returning?
Or it doesn't matter?

shawn-hurley: I don't think that it matters, just following the pattern that I think we started here: 
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/deleg_map.go#L60-L62

Thoughts?

pkg/cache/multi_namespace_cache.go
shawn-hurley: I don't have a good way to do this when say someone is asking for a `PodList` that I can think of.

Maybe we could do some reflection bits to make sure that the `Items` field is there and use the scheme to create a new Object each time. I don't know how much better performance this would be though, WDYT?

shawn-hurley:
@DirectXMan12 or @mengqiy can you guys re-start the build I think it was a flake.

pkg/cache/multi_namespace_cache.go
shawn-hurley: Something like this: https://github.com/shawn-hurley/controller-runtime/blob/feature/poc-reflection-mulitnamespaced-cache-list/pkg/cache/multi_namespace_cache.go#L153-L172

mengqiy: I think the code snippet you linked look reasonable. I believe it can improve performance by avoiding some unnecessary work.
But what confused me is why you treat `ResourceVersion` specially. How about other `metadata` fields?

shawn-hurley: so here was my thinking:

1.  SelfLink string -  would think that this also does not make sense because we are combining multiple calls into one list. There is not a link that one could call to get the same list. This is also optional, so I figured it was fine to ignore.

2. Continue string - I think this also does not make a ton of sense because we should be handling this value in the cache by getting all the values. I would imagine that this is not necessary to set. 

and the only other field is ResourceVersion which we will need.

Does that make sense?

mengqiy: It makes sense.
Re. `Continue string`, I found we currently don't support it in [controller-runtime's ListOptions](https://github.com/kubernetes-sigs/controller-runtime/blob/bd4677326d3a1e201f986bf798638fd38fbe3680/pkg/client/interfaces.go#L191-L212) comparing to [upstream ListOptions](https://github.com/kubernetes/apimachinery/blob/f05b8decd79cf55449e1e5b2cc9f14496ba6b942/pkg/apis/meta/v1/types.go#L365-L396). So I guess we don't need to worry about it ATM.

mengqiy: @droot @DirectXMan12 thoughts regarding the reflection bits linked above?

shawn-hurley: I am pushing up an even better implementation in my mind. It will use api-machinery to manipulate the list, instead of using reflection directly.

pkg/cache/multi_namespace_cache.go
```diff
@@ -145,28 +164,40 @@ func (c *multiNamespaceCache) List(ctx context.Context, list runtime.Object, opt
 		return cache.List(ctx, list, opts...)
 	}
 
-	// Get all the objects in the namespaces we are watching.
-	gvk, err := apiutil.GVKForObject(list, c.Scheme)
+	listAccessor, ok := list.(metav1.ListInterface)
```
mengqiy: It seems you don't have to do the type assertion yourself.
You can use https://github.com/kubernetes-sigs/controller-runtime/blob/e2d83e883d17206248db563cbc71f023d0344d3b/vendor/k8s.io/apimachinery/pkg/api/meta/meta.go#L69-L93.

pkg/cache/multi_namespace_cache.go
```diff
@@ -145,28 +164,40 @@ func (c *multiNamespaceCache) List(ctx context.Context, list runtime.Object, opt
...
+		err = cache.List(ctx, listObj, opts...)
+		if err != nil {
+			return err
+		}
+		items, err := apimeta.ExtractList(listObj)
 		if err != nil {
 			return err
 		}
-		allItems.Items = append(allItems.Items, items.Items...)
+		accessor, ok := listObj.(metav1.ListInterface)
```
mengqiy: Same comment here.

mengqiy:
/lgtm
Let's wait by EOD before merging just in case others may want to take a look.

PR #273, opened 2019-01-02
:sparkles: adding ability to get a APIReader from the client.

### LINKED ISSUE #180: Why DelegatingClient is needed?
From [the code](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L196),
the Manager.GetClient returns an DelegatingClient which read from cache for structured object, but read directly for unstructured object.

This forbids us to use the client before controller loops(since cache needs to be synced). (Our use case is load configurations from configMap before controller loops).

From my perspective, users should use manager.GetCache if they need caching behavior(and we can do the delegation there to read directly for unstructured object).
For manager.GetClient, it's better to let it always read directly without cache.

### LINKED ISSUE #245: Possible race condition / inconsistency between CRUD operations
`Create`/`Update`/`Delete` operations are performed using non-caching clients and return before any caches are likely to have seen the changes. This means `Get`/`List` can return stale data for an undeterministic amount of time, could be milliseconds or could be longer based on latency/load on both client and apiserver.

Example:
1. List resource type X (see count: 0)
2. Create resource of type X (success)
3. List resource of type X (see count: 0)
4. wait
5. List resource of type X (see count: 1)
6. Update resource of type X (success)
7. Get resource of type X (see old object)
8. wait
9. Get resource of type X (see new object) 
10. Delete resource of type X
11. Get resource of type X (see non-deleted object)
12. List resource of type X (see count: 1)
13. wait
14. Get resource of type X (see tombstoned object)
15. wait
16. List resource of type X (see count: 0)

This is something that seems likely to catch quite a few operator developers off guard and could cause a lot of resources to be created when missing if the identifier for the resource is non-deterministic/random.

This relates a little to #180 because the behavior comes from implicit caching.

shawn-hurley:
Adding a new interface `APIClient` to expose as part of the client
a non-caching reader interface.

fixes #180
fixes #245

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -144,6 +144,10 @@ func (c *client) Status() StatusWriter {
 	return &statusWriter{client: c}
 }
 
+func (c *client) APIReader() Reader {
+	return c
+}
+
 // statusWriter is client.StatusWriter that writes status subresource
 type statusWriter struct {
 	client *client
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -1084,6 +1084,819 @@ var _ = Describe("Client", func() {
 		})
 	})
 
+	Describe("APIReader", func() {
+		Describe("Get", func() {
+			Context("with structured objects", func() {
+				It("should fetch an existing object for a go struct", func(done Done) {
+					By("first creating the Deployment")
+					dep, err := clientset.AppsV1().Deployments(ns).Create(dep)
+					Expect(err).NotTo(HaveOccurred())
+
+					cl, err := client.New(cfg, client.Options{})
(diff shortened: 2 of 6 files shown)


shawn-hurley:
I think this is probably the first step described in #180 

I think the next step would be to talk about the return of the `GetCache` and/or change the name. I think we want the ability to get to the cache for adding event handlers and such, but maybe hiding/changing the reader interface? @droot thoughts?

DirectXMan12:
I don't like this being part of client -- it brings weird implications to how we implement client.  I think it should probably just be exposed directly on the manager.

shawn-hurley:
I think that I agree with @DirectXMan12 I will close this PR and move this to the manager in a separate PR.

Issue #228, opened 2018-11-30
EnqueueRequestForOwner doesn't correctly enqueue cluster-scoped owner


grantr:
_Splitting out the cluster-scoped owner scenario from #214._

As [pointed out](https://github.com/kubernetes-sigs/controller-runtime/issues/214#issuecomment-443343432) by @liggitt, cluster-scoped owners are now [explicitly supported](https://github.com/kubernetes/kubernetes/blob/6e5176b48df8719ab969a233a0411ad6ba35d222/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L288-L290).

`EnqueueRequestForOwner` doesn't support this relationship because it uses the namespace of the owned object as the namespace of the owner. We ran into this in https://github.com/knative/eventing/issues/649.

rajathagasthya:
I'd like to work on this, with some hand-holding. How do I get started with making changes?

DirectXMan12:
looks like #274 is working on this

grantr:
I think this can be closed now that #274 is merged. Reopen if I'm wrong about that.
/close

PR #274, opened 2019-01-03
:bug: EnqueueRequestForOwner correctly enqueue cluster-scoped owner

shawn-hurley:
Determines if the owner object is cluster scoped, and if it is will not use the object's namespace in the reconcile request.

To complete this, it adds dependency injection for the managers RESTMapper.

### DIFF
--- a/pkg/handler/enqueue_owner.go
+++ b/pkg/handler/enqueue_owner.go
@@ -19,6 +19,7 @@ package handler
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -51,6 +52,9 @@ type EnqueueRequestForOwner struct {
 
 	// groupKind is the cached Group and Kind from OwnerType
 	groupKind schema.GroupKind
+
+	// mapper maps GroupVersionKinds to Resources
+	mapper meta.RESTMapper
 }
 
 // Create implements EventHandler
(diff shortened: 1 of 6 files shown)


pwittrock:
Cool.  IIUC - if both the owned and owning objects are namespaced things work correctly.  If both are not-namespaced things work correctly.  This addresses the case where a non-namespaced object owns a namespaced object.  I am curious, are there usecases for this pattern you can share?

pkg/handler/enqueue_owner.go
```diff
@@ -126,10 +130,21 @@ func (e *EnqueueRequestForOwner) getOwnerReconcileRequest(object metav1.Object)
 		// object in the event.
 		if ref.Kind == e.groupKind.Kind && refGV.Group == e.groupKind.Group {
 			// Match found - add a Request for the object referred to in the OwnerReference
-			result = append(result, reconcile.Request{NamespacedName: types.NamespacedName{
-				Namespace: object.GetNamespace(),
-				Name:      ref.Name,
-			}})
+			r := reconcile.Request{NamespacedName: types.NamespacedName{
```
DirectXMan12: please don't use single-letter variable names for anything other than loop indicies and receivers

pkg/handler/enqueue_owner.go
```diff
@@ -126,10 +130,21 @@ func (e *EnqueueRequestForOwner) getOwnerReconcileRequest(object metav1.Object)
...
+				Name: ref.Name,
+			}}
+
+			// if owner is not namespaced then we should set the namespace to the empty
+			mapping, err := e.mapper.RESTMapping(e.groupKind, refGV.Version)
+			if err != nil {
+				log.Error(err, "Could not retrieve rest mapping", "group", e.groupKind.Group, "kindv", e.groupKind.Kind)
```
DirectXMan12: "kindv"?

Also: https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md#groups-versions-and-kinds (just use the "kind" tag with the groupKind directly)

pkg/handler/eventhandler_suite_test.go
```diff
@@ -30,6 +31,18 @@ func TestEventhandler(t *testing.T) {
...
+var testenv *envtest.Environment
+var cfg *rest.Config
+
 var _ = BeforeSuite(func() {
 	logf.SetLogger(logf.ZapLoggerTo(GinkgoWriter, true))
+
+	testenv = &envtest.Environment{}
+	var err error
+	cfg, err = testenv.Start()
+	Expect(err).NotTo(HaveOccurred())
```
DirectXMan12: can we just extract this out onto a helper on `testenv` at some point?  I want to be able to call `BeforeSuite(testenv.Environment{}.BeforeSuite)` or something similar.

droot:
Changes look good. I have the same question as @pwittrock are there any use-cases for `non-namespaced object owns a namespaced object` ?

shawn-hurley:
@pwittrock @droot #228 is the bug that I was attempting to fix.

> EnqueueRequestForOwner doesn't support this relationship because it uses the namespace of the owned object as the namespace of the owner. We ran into this in knative/eventing#649.

This is the part that you might be interested in for they use case.

pkg/handler/eventhandler_suite_test.go
shawn-hurley: Should I create an issue to track this or is it already tracked some other place?

droot:
The change looks good to me. Will leave it to @DirectXMan12 to lgtm it.

DirectXMan12:
/lgtm
/approve
looks like there's a conflict with another pr.  please rebase, and we'll get this merged.

fabxc:
I ran into this issue and upgraded to a version that includes this commit. It doesn't seem to fix the issue for me.

DirectXMan12:
can you post a reproducer or test or something to quickly verify?

PR #280, opened 2019-01-05
:bug: fix interface assertions in example/mutatingwebhook.go such that podA…

### LINKED ISSUE #266: Incorrect interface assertions in example/mutatingwebhook.go
I believe [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/example/mutatingwebhook.go#L67) and [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/example/mutatingwebhook.go#L77) need to be updated to assert `podAnnotator`'s conformance to `inject.Client` and `inject.Decoder`. 

`example/validatingWebhook.go` already does the interface assertions for `podValidator`. 

I can submit the PR you agree this needs done.

perryao:
…nnotator is required to implement both inject.Client and inject.Decoder interfaces

Closes #266

### DIFF
--- a/example/mutatingwebhook.go
+++ b/example/mutatingwebhook.go
@@ -62,9 +62,9 @@ func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error 
 	return nil
 }
 
-// podValidator implements inject.Client.
+// podAnnotator implements inject.Client.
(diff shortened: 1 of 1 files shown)


mengqiy:
Thanks for the patch!
/lgtm

PR #285, opened 2019-01-08
[WIP] :warning: dependency injection rev2 experiment

DirectXMan12:
This is an experiment around improving our dependency injection story.  The implementation is a bit hairy, but the actual usage mechanics are fairly straightforward.

All dependencies are stored in a special dependency context:

```go
ctx := inject.Nothing()
```

New dependencies can be add or overwritten by adding a new link to the context:

```go
ctx = inject.ProvideA(ctx, mySpecialClient)
```

Dependencies (including the context itself, which gets automatically injected) can be injected into a struct:

```go
type Foo struct {
  Client client.Client `inject:""`
  specialDep FooBar
}

func (f *Foo) InjectFooBar(fb FooBar) error {
  f.specialDep = fb
}

...

myFoo := &Foo{}
inject.Into(ctx, myFoo)
```

Dependencies can also be fetched directly:

```go
var myClient client.Client
ctx.PopulateThis("", &myClient) // the first argument is an optional "key" for things like InjectStopChannel, where the type name doesn't correspond to the "name" of the dependency
```

Finally, things wishing to dynamically provide dependencies without constructing a new context every time can implement a link in the `Context` chain themselves (by adding `PopulateThis`, which implements `Context`) -- `manager.Manager` does this.

There's a *lot* of reflection under the hood, but this model has a number of nice properties:

- dependencies can be fetched directly
- dependencies can be "masked" by other dependencies for specific scopes
- dependencies can be added for specific scopes
- adding new dependencies is pretty easy

### DIFF
--- a/example/mutatingwebhook.go
+++ b/example/mutatingwebhook.go
@@ -22,7 +22,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
 )
@@ -62,9 +61,7 @@ func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error 
 	return nil
 }
 
-// podValidator implements inject.Client.
 // A client will be automatically injected.
-var _ inject.Client = &podValidator{}
 
 // InjectClient injects the client.
 func (v *podAnnotator) InjectClient(c client.Client) error {
@@ -72,9 +69,7 @@ func (v *podAnnotator) InjectClient(c client.Client) error {
 	return nil
 }
 
-// podValidator implements inject.Decoder.
 // A decoder will be automatically injected.
-var _ inject.Decoder = &podValidator{}
 
 // InjectDecoder injects the decoder.
 func (v *podAnnotator) InjectDecoder(d types.Decoder) error {
--- a/pkg/inject/context.go
+++ b/pkg/inject/context.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inject
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Context carries dependencies across component boundaries.
+type Context interface {
+	// PopulateThis populates the given pointer with an instance of its type,
+	// if available.  The target *must* be a pointer.  Key is an optional key
+	// to differentiate different values of the same type.  If it's the same
+	// as the type name, pass the empty string.
+	PopulateThis(key string, target interface{}) bool
+}
+
+// Nothing returns a new Context that doesn't provide any dependencies.
+func Nothing() Context {
+	return emptyContext{}
+}
+
+// emptyContext doesn't provide anything
+type emptyContext struct{}
+
+func (c emptyContext) PopulateThis(_ string, _ interface{}) bool { return false }
+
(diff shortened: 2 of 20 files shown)


DirectXMan12:
related to #182 

cc @pwittrock

pkg/inject/context.go
```diff
@@ -0,0 +1,115 @@
...
+func (c emptyContext) PopulateThis(_ string, _ interface{}) bool { return false }
+
+// depContext is a Context that provides a dependency of a given type.
+// It'll call out to next if it can provide the type asked for.
+type depContext struct {
+	typ         reflect.Type
+	value       reflect.Value
+	key         string
+	keyOptional bool
+	next        Context
```
droot: wondering if having `depContext` as a container of multiple values can simplify things.

example/mutatingwebhook.go
```diff
@@ -62,19 +61,15 @@ func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error
 	return nil
 }
 
-// podValidator implements inject.Client.
 // A client will be automatically injected.
-var _ inject.Client = &podValidator{}
 
 // InjectClient injects the client.
 func (v *podAnnotator) InjectClient(c client.Client) error {
```
droot: If I understand correctly, I can get rid of this method by just adding a tag `inject` on client field in `podAnnotator` ?

pkg/internal/controller/controller.go
```diff
@@ -72,8 +70,8 @@ type Controller struct {
 	// the Queue for processing
 	Queue workqueue.RateLimitingInterface
 
-	// SetFields is used to inject dependencies into other objects such as Sources, EventHandlers and Predicates
-	SetFields func(i interface{}) error
+	// Dependencies are used to inject further dependencies into other objects, such as Sources, EventHandlers, and Predicates.
+	Dependencies inject.Context `inject:""`
```
droot: So this is the field users set in their main.go to provide external deps at the entry point.

pkg/inject/inject.go
```diff
@@ -0,0 +1,173 @@
...
+		next:        ctx,
+	}
+}
+
+// Into reads the given object for implementations
+// of injectable pseudo-interfaces and struct tags,
+// and injects dependencies based on those.
+// As a special case, the given context will also "Provide" itself
+// with or without key "Dependencies".
+func Into(ctx Context, target interface{}) (bool, error) {
```
droot: I feel the order of `ctx` and `target` needs to be reversed.

DirectXMan12: yeah.. but everything else takes ctx first.

example/mutatingwebhook.go
droot: Got it. We should document it at an appropriate place, because users will have the same question.

pkg/internal/controller/controller.go
droot: Yes, makes sense.

shawn-hurley:
I am still attempting to use this to in a project to get the feel for it. After looking at it though I like this approach outlined here. 

I have some concerns about the use of Context (because I am worried it be confused with `context.Context`). 

I have been a little busy but hopefully, have some availability opening up and can give a more thorough look.

droot:
>I have some concerns about the use of Context (because I am worried it be confused with context.Context).

Can we call it `DependencyContext`, I know it is a bit longer, but looks appropriate for the functionality. and `dctx` or `depctx` for the variables in the codebase. Thoughts ? @shawn-hurley @DirectXMan12

pkg/manager/manager.go
```diff
@@ -36,22 +36,22 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
 )
 
 // Manager initializes shared dependencies such as Caches and Clients, and provides them to Runnables.
 // A Manager is required to create Controllers.
 type Manager interface {
+	// Every manager knows how to inject dependencies into other things.
+	inject.Context
```
mengqiy: I'm wondering if changing it to `Dependency` or `Dependencies` makes more sense.
So this user-facing piece becomes `inject.Dependency` or `inject.Dependencies`.

eriknelson:
Hey folks, to explain the link, @shawn-hurley directed me towards this PR today while we were chatting about DI in go. I thought there were some interesting ideas presented in this PR. I wonder if there's work here that could be generically useful to other projects?

DirectXMan12:
It's possible some of this is generically useful.  I've generally shifted away from needing this as much in KB-generated projects by default.  Some of this still feels a bit clunky to me.  I'm curious if you had more detailed thoughts, though.

DirectXMan12:
closing for now, since this is pretty complicated and ends up being *way* less necessary with the new KB v2 scaffolding

PR #290, opened 2019-01-09
:warning: Refactor logging structure, avoid mandatory Zap

DirectXMan12:
The way we currently have logging structured, you *have* to pull in Zap, even if you don't use Zap as your logr backend.

In the spirit of avoiding dependency bloat, this separates zap off into its own package, and moves around some of the logging package hierarchy:

- pkg/log contains generic logging helpers, and the generic root logger
  It *only* depends on logr, and thus won't pull in Zap if you don't
  need it.

- pkg/internal/log contains the internal log handle (with the
  appropriate name).  Nothing outside CR should be logging with this
  handle anyway, so this enforces it.

- pkg/log/zap contains Zap-related setup code, and thus CR only has a
  dependency on Zap if you pull in this package.

- pkg/runtime/log remains as a deprecated package to ease refactoring
  and compatibility.  While practically a few types have been removed
  from this package, in reality most consumers shouldn't notice.

This has the practical effect of meaning that while you now have to explicitly import `pkg/log/zap` if you want Zap setup, if you don't pull that in, you don't get Zap listed as a dependency.


DirectXMan12:
@pwittrock do you remember what the idea was behind `pkg/runtime/xyz` vs `pkg/xyz` directly?

pkg/log/log.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// SetLogger sets a concrete logging implementation for all deferred Loggers.
+func SetLogger(l logr.Logger) {
+	Log.Fulfill(l)
+}
+
+// Log is the base logger used by kubebuilder.  It delegates
+// to another logr.Logger.  You *must* call SetLogger to
+// get any actual logging.
```
droot: wondering if we can set stdout-logger as default ? (or we need to ensure scaffolding in kubebuilder is doing that) because stdout-logging is a sane default.

droot: Also, can we include an example of how to enable stdout logger ?

pkg/log/zap/zap.go
```diff
@@ -0,0 +1,68 @@
...
+// Logger is a Logger implementation.
+// If development is true, a Zap development config will be used
+// (stacktraces on warnings, no sampling), otherwise a Zap production
+// config will be used (stacktraces on errors, sampling).
+func Logger(development bool) logr.Logger {
+	return LoggerTo(os.Stderr, development)
+}
+
+// LoggerTo returns a new Logger implementation using Zap which logs
+// to the given destination, instead of stderr.  It otherise behaves like
```
droot: s/otherise/otherwise

PR #292, opened 2019-01-14
Don't Merge: Example of simpler interface for Webhooks

pwittrock:
This code isn't polished enough to be merged.

Sending to show folks the vision.


pkg/webhook/handlers.go
```diff
@@ -0,0 +1,165 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
+)
+
+// Validator defines functions for validating an operation
+type Validator interface {
+	runtime.Object
+	ValidateCreate() error
+	ValidateUpdate(old runtime.Object) error
```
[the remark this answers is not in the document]
pwittrock: Do we need to initially?

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		err = obj.ValidateUpdate(oldObj)
+		if err != nil {
+			return admission.ValidationResponse(false, err.Error()), true
+		}
+	}
+
+	return types.Response{}, false
+}
```
droot: This is fantastic. I like the way it hides all the webhook machinery from the end-users. Resource author just need to define the `validate` and `default` methods on their types and rest is taken care by the machinery (exactly the way it should be). 

This is a huge step in making `defaulting` and `validation` accessible.

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return admission.ErrorResponse(http.StatusBadRequest, err), true
+	}
+
+	// Default the object
+	copy := obj.DeepCopyObject().(Defaulter)
+	copy.Default()
+
+	// Create the patch
+	return admission.PatchResponse(obj, copy), true
```
droot: This causes roundtripping issue, we are moving away from this https://github.com/kubernetes-sigs/controller-runtime/pull/256 .

pkg/webhook/handlers.go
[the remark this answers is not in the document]
mengqiy: > ValidateUpdate(old runtime.Object) error

IIRC if the webhook is registered as [`ValidatingWebhookConfiguration`](https://github.com/kubernetes/kubernetes/blob/69cb938a6ab8a4a281ac0774287efde89d253701/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L83-L94), old object will not be included in the AdmissionReview request.
But a nil object may not be a problem here.

pkg/webhook/handlers.go
pwittrock: +1

droot:
@mengqiy is working on extracting manifest generation out of controller-runtime that has an overlap with this, so has volunteered to tackle this PR.

PR #295, opened 2019-01-16
:book: Add project overview to docs, clean up docs, structure

DirectXMan12:
This adds an overview of the project organization to the root godocs, and cleans up the docs here and there, fixing whitespace issues, cleaning up wording, clarifying, and adding docs where they're lacking.

This also moves things out of pkg/runtime, so that pkg/runtime is (mostly) deprecated paths to other things.


pkg/manager/manager.go
```diff
@@ -84,7 +84,8 @@ type Manager interface {
 // Options are the arguments for creating a new Manager
 type Options struct {
 	// Scheme is the scheme used to resolve runtime.Objects to GroupVersionKinds / Resources
-	// Defaults to the kubernetes/client-go scheme.Scheme
+	// Defaults to the kubernetes/client-go scheme.Scheme, but it's almost always a good
+	// idea to pass your own scheme in.
```
droot: I have seen users struggling with this. I think we can provide some guidance here: For ex. user is writing controllers with built-in types, then default scheme is enough, but if using your own types/external-types, augment the default scheme with the scheme for specific types.

DirectXMan12: I've added more documentation on schemes to pkg/runtime/scheme, which probably should be promoted to `pkg/scheme`.

FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+up state that's no longer referenced.
+
+### Q: My cache might be stale if I read from a cache! How should I deal with that?
+
+**A**: There are several different approaches that can be taken, depending
+on your situation.
+
+- When you can, take advantage of optimistic locking: use deterministic
+  names for objects you create, so that the Kubernetes API server will
+  warn you if the object already exists.  May controllers in Kubernetes
```
rajathagasthya: s/May/Many/

FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+time it runs.  If none of this works for you, you can always construct
+a client that reads directly from the API server, but this is generally
+considered to be a last resort, and the two approaches above should
+generally cover most circumstances.
+
+### Q: Where's the fake client?  How do I use it?
+
+**A**: The fake client
+[exists](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/fake),
+but we generally reccomend using
```
thbkrkr: s/reccomend/recommend/

FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+clients gradually re-implement poorly-written impressions of a real API
+server, which leads to hard-to-maintain, complex test code.
+
+### Q: How should I write tests?  Any suggestions for getting started? 
+
+- Use the aforementioned
+  [envtest.Environment](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest#Environment)
+  to spin up a real API server instead of trying to mock one out.
+
+- Structure your tests to check the that the state of the world is as you
```
thbkrkr: s/check the that/check that/

FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+  may have some delay between write time and reconcile time.
+
+### Q: What are these errors about no Kind being registered for a type?
+
+**A**: You're probably missing a fully-set-up Scheme.  Schemes record the
+mapping between Go types and group-version-kinds in Kubernetes. In
+general, your application should have its own Scheme containing the types
+from the API groups that it needs (be they Kubernetes types or your own).
+See the [scheme builder
+docs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/scheme) for
```
thbkrkr: s|`controller-runtime/pkg/scheme`|`controller-runtime/pkg/runtime/scheme`|

DirectXMan12: see above changes

mengqiy:
/lgtm
/hold
Moving packages is actually breaking changes. Maybe we should change the commit messages to use :warning: instead of :sparkles:

mengqiy:
IIRC we are not going to release again in `release-0.1` branch, so using :warning: or :sparkles: doesn't matter much :)
/hold cancel

DirectXMan12:
The package moves left aliases in place, so they shouldn't actually be breaking yet.

pkg/manager/manager.go
```diff
@@ -74,8 +74,8 @@ type Manager interface {
 	// GetCache returns a cache.Cache
 	GetCache() cache.Cache
 
-	// GetRecorder returns a new EventRecorder for the provided name
-	GetRecorder(name string) record.EventRecorder
+	// GetEventRecorderFor returns a new EventRecorder for the provided name
+	GetEventRecorderFor(name string) record.EventRecorder
```
grantr: This is a breaking change. Was that intentional? If not, can we add the `GetRecorder` method back and deprecate it?

DirectXMan12: That was marked as breaking.  Did it make it into a stable branch?  We should do a patch to the stable branch to restore it.

DirectXMan12: Doesn't look like it made it into a stable branch (it's not on release 0.1.10).  Master is intended to have breaking changes, and yes, this was intended to be breaking (it brings manager in line with other things that provide a recorder).

grantr: Oops, I missed the breaking mark (I wasn't aware of the meaning of ⚠️), so please ignore my comment. Thanks for explaining @DirectXMan12!

PR #298, opened 2019-01-16
:sparkles: Master -> Control Plane

vincepri:
Renames master to controlplane as discussed in the [Jan 31st steering meeting](https://docs.google.com/document/d/1qazwMIHGeF3iUh5xMJIJ6PDr-S3bNkT8tNLRkSiOkOU/edit#)

Related: https://github.com/kubernetes/website/issues/6525

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -29,8 +29,8 @@ import (
 )
 
 var (
-	kubeconfig, masterURL string
-	log                   = logf.RuntimeLog.WithName("client").WithName("config")
+	kubeconfig, apiServerURL string
+	log                      = logf.RuntimeLog.WithName("client").WithName("config")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


vincepri:
Related to https://github.com/kubernetes-sigs/cluster-api/issues/688

coderanger:
:+1: for `--api-server` and similar as being both more accurate and nicer :)

DirectXMan12:
>  I should probably have done more research around the decision beforehand, I didn’t expect it to cause trouble, apologies!

No worries :-)  Sorry if my message above came off a bit harsh.

> community meeting announcement by Joe Beda is from Feb 1st 2018:
https://github.com/kubernetes/community/blob/master/communication/meeting-notes-archive/q1-2_2018_community_meeting_minutes.md#february-1-2018---recording


Perfect, thanks!

> but i'm mostly leaning towards redundant.

Yeah, I'm leaning towards just marking as deprecated (so as not to break people, but it'll hide it in help, etc).

> as outlined in the above announcement, "control plane" encapsulates a composite meaning. if we are targeting an individual component, we should do so in docs and flags for clarity.

:+1: (more or less what I was saying above, I think)

vincepri:
@DirectXMan12 @neolit123 Updated the PR with the latest comment, the flag is marked as deprecated both in code and description suggesting users to switch to --kubeconfig

pkg/client/config/config.go
```diff
@@ -29,21 +29,22 @@ import (
 )
 
 var (
-	kubeconfig, masterURL string
-	log                   = logf.RuntimeLog.WithName("client").WithName("config")
+	kubeconfig, controlPlaneURL string
```
neolit123: switch to `apiServerAddress` as the variable name for the deprecation period?

vincepri: I'm fine either way, @DirectXMan12 ?

vincepri:
Is CI broken? I tested locally on master and it also shows the same failures.

pkg/client/config/config.go
DirectXMan12: yeah, use `apiServerURL` as the variable name sot that we're clear.

DirectXMan12:
> Is CI broken

I think it was fixed in a recent PR.  Rebase on master.

vincepri:
Rebased and updated the PR with the variable name change as suggested above.

PR #300, opened 2019-01-17
:warning: split webhook server and manifest generation

mengqiy:
Code related to running a webhook server stays in CR repo.
Code related to generating cert are dropped.
Code related to generating non-cert manifests (e.g. webhookConfiguration, service) are currently under `pkg/webhookgenerator`,which will be moved to the controller-tools repo.


mengqiy:
The PR is not polished yet.
Sent out the draft for getting feedback before @DirectXMan12 disappears next week :)

pkg/internal/webhookgenerator/admission/webhook.go
```diff
@@ -0,0 +1,105 @@
...
+	"regexp"
+	"strings"
+	"sync"
+
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/internal/webhookgenerator/types"
+)
+
+// Webhook represents each individual webhook.
```
droot: May be we can describe it a bit more:
```
Webhook contains bits needed for generating a Webhook Configuration/manifest ? 
```

pkg/internal/webhookgenerator/server.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+
+	apitypes "k8s.io/apimachinery/pkg/types"
+	types "sigs.k8s.io/controller-runtime/pkg/internal/webhookgenerator/types"
+)
+
+// ServerOptions are options for configuring an admission webhook server.
+type ServerOptions struct {
```
droot: may be we should rename the filename to be also server_options.go ?

pkg/internal/webhookgenerator/server.go
```diff
@@ -0,0 +1,113 @@
...
+	// Port is the port number that the server will serve.
+	// It will be defaulted to 443 if unspecified.
+	Port int32
+
+	// CertDir is the directory that contains the server key and certificate.
+	// If using FSCertWriter in Provisioner, the server itself will provision the certificate and
+	// store it in this directory.
+	// If using SecretCertWriter in Provisioner, the server will provision the certificate in a secret,
+	// the user is responsible to mount the secret to the this location for the server to consume.
+	CertDir string
```
droot: I think most of these comments need to be re-written from the new purely generation perspective. Currently they read as if webhook server itself is going to use these options to install webhook configuration.

pkg/internal/webhookgenerator/server.go
```diff
@@ -0,0 +1,113 @@
...
+	return nil
+}
+
+// Start runs the server.
+// It will install the webhook related resources depend on the server configuration.
+func (s *ServerOptions) Start(stop <-chan struct{}) error {
+	s.setDefault()
+
+	return s.InstallWebhookManifests()
+}
```
droot: Same comment as above. Some of these methods `Register` and `Start` doesn't seem relevant from purely generation perspective.

mengqiy:
PTAL

mengqiy:
Pushed a little more changes on top of the earlier commit.

pkg/webhook/server.go
```diff
@@ -203,105 +174,57 @@ func (s *Server) Start(stop <-chan struct{}) error {
...
-var _ inject.Decoder = &Server{}
-
-// InjectDecoder injects the client into the server
-func (s *Server) InjectDecoder(d atypes.Decoder) error {
-	for _, wh := range s.registry {
-		if _, err := inject.DecoderInto(d, wh.Handler()); err != nil {
-			return err
+	for path := range s.registry {
+		// TODO(mengqiy): remove this in PR #316
+		if wh, ok := s.registry[path].(Webhook); ok {
```
DirectXMan12: this inject code isn't quite right the way it's written, since someone could write their own webhook impl that needed the info.  If webhook isn't an interface, this isn't a problem.

pkg/webhook/server.go
```diff
@@ -159,40 +95,75 @@ func NewServer(name string, mgr manager.Manager, options ServerOptions) (*Server
...
+		s.sMux = http.DefaultServeMux
+	}
+	if s.Port <= 0 {
+		s.Port = 443
+	}
+	if len(s.CertDir) == 0 {
+		s.CertDir = path.Join("k8s-webhook-server", "cert")
+	}
+
+	if s.Client == nil {
```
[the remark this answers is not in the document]
mengqiy: I realized it too when fixing the injector for webhook. I have dropped it in #316.

pkg/webhook/server.go
```diff
@@ -123,30 +66,23 @@ type Server struct {
...
-	// They can be nil, if there is no webhook registered under it.
-	webhookConfigurations []runtime.Object
+	registry map[string]http.Handler
 
 	// manager is the manager that this webhook server will be registered.
 	manager manager.Manager
 
-	// httpServer is the actual server that serves the traffic.
-	httpServer *http.Server
+	// err will be non-nil if there is an error occur during initialization.
```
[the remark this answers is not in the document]
mengqiy: It is kind of the same pattern as [kubectl resource builder](https://github.com/kubernetes/kubernetes/blob/f96849981203ea506ea8fa3f88c173d36cf688aa/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/resource/builder.go#L70).

> e.g. multiple registrations will overwrite errors

You are right. We can probably make it an array of errors i.e. `[]error`.

If you really don't like it, we rethink how to handle it :)

pkg/webhook/server.go
mengqiy: This is fixed in https://github.com/kubernetes-sigs/controller-runtime/pull/316/commits/a4ef92988bb52994fceff913d959f8e7f4162fa7#diff-fbc18bb07cdd05391b7081acc1dfe170R209
I tried to avoid putting everything in the same PR, so the code here may look incorrect.

pkg/webhook/server.go
```diff
@@ -159,40 +95,75 @@ func NewServer(name string, mgr manager.Manager, options ServerOptions) (*Server
...
 		ServerOptions: options,
 		manager:       mgr,
 	}
 
 	return as, nil
 }
 
+// setDefault does defaulting for the Server.
+func (s *Server) setDefault() {
+	if len(s.Name) == 0 {
```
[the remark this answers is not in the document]
mengqiy: Yup.
It was for [the name (identifier) of each individual webhook](https://github.com/kubernetes/kubernetes/blob/f96849981203ea506ea8fa3f88c173d36cf688aa/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L146), but it is no longer used anywhere in CR.

pkg/webhook/server.go
```diff
@@ -159,40 +95,75 @@ func NewServer(name string, mgr manager.Manager, options ServerOptions) (*Server
...
+		s.Name = "default-k8s-webhook-server"
+	}
+	if s.registry == nil {
+		s.registry = map[string]http.Handler{}
+	}
+	if s.sMux == nil {
+		s.sMux = http.DefaultServeMux
```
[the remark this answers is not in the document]
mengqiy: > why is the default mux different from the mux used if you call the constructor?

Thanks for catching this.
I agree we should use the same mux. i.e. use `http.NewServeMux()`.
 
> for that matter, why do we have a constructor if we've got the `setDefaults` style?

Because some users may use the public `Server` struct directly, we need to ensure it get sane defaulting and works.

mengqiy:
PTAL

pkg/webhook/server.go
```diff
@@ -199,110 +142,51 @@ var _ manager.Runnable = &Server{}
...
-func (s *Server) RefreshCert() (bool, error) {
-	cc, err := s.getClientConfig()
-	if err != nil {
-		return false, err
-	}
-	changed, err := s.certProvisioner.Provision(cert.Options{
-		ClientConfig: cc,
-		Objects:      s.webhookConfigurations,
-	})
+	listener, err := tls.Listen("tcp", net.JoinHostPort("", strconv.Itoa(int(s.Port))), cfg)
```
DirectXMan12: yay!  `JoinHostPort` :+1:

mengqiy:
Squashed the commits, no new code change.
Re-applying the lgtm label per https://github.com/kubernetes-sigs/controller-runtime/pull/300#pullrequestreview-203909021

/hold cancel

PR #311, opened 2019-01-25
:bug: fix labels list option in fake client

### LINKED ISSUE #293: Fake client doesn't apply the LabelSelector or FieldSelector options on List
Currently the fake client doesn't apply the `ListOptions.LabelSelector` or the `ListOptions.FieldSelector` to filter results when using the `List` function.

Here is the current fake client's `List` function code:
```golang {.line-numbers}
func (c *fakeClient) List(ctx context.Context, opts *client.ListOptions, list runtime.Object) error {
	gvk, err := getGVKFromList(list, c.scheme)
	if err != nil {
		// The old fake client required GVK info in Raw.TypeMeta, so check there
		// before giving up
		if opts.Raw == nil || opts.Raw.TypeMeta.APIVersion == "" || opts.Raw.TypeMeta.Kind == "" {
			return err
		}
		gvk = opts.Raw.TypeMeta.GroupVersionKind()
	}

	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
	o, err := c.tracker.List(gvr, gvk, opts.Namespace)
	if err != nil {
		return err
	}
	j, err := json.Marshal(o)
	if err != nil {
		return err
	}
	decoder := scheme.Codecs.UniversalDecoder()
	_, _, err = decoder.Decode(j, nil, list)
	return err
}
```

Probably the filter logic should be apply after the `_, _, err = decoder.Decode(j, nil, list)` code line.

sebgl:
This commit allows passing label matches to the fake client `List()` function. 
This feature was not working properly before (no filters applied at all).

This is done in a way similar to the `CachedReader`. Both now share a common call to `listutil.FilterWithLabels`.

One thing worth discussing:

I wasn't sure where to extract this common function in the package hierarchy. Keeping it in `cache/internal` does not work since the fake client cannot use a package named `internal`. Looking at other packages I could not find the right place for it.
Hence the new `cache/listutil` package. It still feels a bit wrong, suggestions for a new location are very welcome!

Partial fix for #293.

### DIFF
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
 )
 
 // CacheReader is a CacheReader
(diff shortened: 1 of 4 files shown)


DirectXMan12:
I'd put this under `pkg/internal` (because it's mostly an internal detail that we don't want people depending on as part of the public, versioned API surface)

DirectXMan12:
otherwise LGTM

sebgl:
@DirectXMan12 thanks for looking at this. You're right, `pkg/internal` makes more sense here.

I also renamed the package name from `listutil` to `objectutil`. I think it leaves more room for other objects manipulation/conversion functions that may exist in the future?

DirectXMan12:
Looks good.  Just squash the commits and I'll merge (our policy is to have "logical" commits at the end, where each commit describes a distinct piece of functionality)

PR #313, opened 2019-01-31
:bug: cleanup and fix CreateOrUpdate

adracus:
This resolves an issue with the `MutateFn` being run even if the initial get call fails with an `error` other than `errors.IsNotFound`. Also, it cleans up the convoluted structure of the method, eliminating lots of `if/else` branches and assignments to a top level (`op`) variable.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -121,57 +121,42 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 //
 // It returns the executed operation and an error.
 func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
-	// op is the operation we are going to attempt
-	op := OperationResultNone
-
-	// get the existing object meta
-	metaObj, ok := obj.(v1.Object)
-	if !ok {
-		return OperationResultNone, fmt.Errorf("%T does not implement metav1.Object interface", obj)
-	}
-
-	// retrieve the existing object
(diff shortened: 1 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -121,57 +121,54 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
...
+		return OperationResultNone, err
 	}
 
-	// retrieve the existing object
-	key := client.ObjectKey{
-		Name:      metaObj.GetName(),
-		Namespace: metaObj.GetNamespace(),
+	if err := c.Get(ctx, key, obj); err != nil {
+		if errors.IsNotFound(err) {
+			if err := safeMutate(key, obj, f); err != nil {
```
mengqiy: When the object is not found on the server, we should directly call `Create`. Why we want to call `safeMutate` here?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -121,57 +121,54 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
...
 	}
 
+	newKey, err := client.ObjectKeyFromObject(obj)
 	if err != nil {
-		op = OperationResultNone
+		return err
+	}
+
+	if key != newKey {
+		return fmt.Errorf("MutateFn cannot mutate object key")
```
mengqiy: Error message will be more useful if we point out which fields are not mutable.
e.g. `MutateFn cannot mutate namespace and (or) object name.`

PR #316, opened 2019-02-04
:bug: fix webhook injector

### LINKED ISSUE #309: Webhook: Need more injection Interface in webhook server
Webhook server only implements `inject.Client` and `inject.Decoder`.  I think the server should implement all of injection interfaces. Then I can use these interfaces in my webhooks.

mengqiy:
fixes #309

Based on #300.

use `mgr.SetFields()` for webhook server and implement `Injector` interface for webhooks.


mengqiy:
cc: @kdada

DirectXMan12:
can you provide a bit more detail in the PR description/commit message?  Otherwise LGTM

pkg/webhook/server.go
```diff
@@ -181,6 +180,11 @@ func (s *Server) Register(webhooks ...Webhook) error {
 		}
 		s.registry[webhook.GetPath()] = webhooks[i]
 		s.sMux.Handle(webhook.GetPath(), webhook.Handler())
+
+		// Inject dependencies into each webhook
+		if err := s.manager.SetFields(webhooks[i]); err != nil {
```
droot: Interesting... Won't manager call `SetField` on webhook server and that should call the inject field for each of the webhooks ? because manager knows when all the deps are resolved ?

DirectXMan12: yeah, this should be receiving an injectfunc, which it should be using to inject things, as opposed to having a reference to the manager.

DirectXMan12: plus, this should be called before trying to call any methods on the webhook

mengqiy:
The first 2 commits are from #300.

example/main.go
```diff
@@ -81,62 +79,33 @@ func main() {
...
-	mutatingWebhook, err := builder.NewWebhookBuilder().
-		Name("mutating.k8s.io").
+	mutatingWebhook := builder.NewWebhookBuilder().
+		Path("/mutate-pods").
 		Mutating().
-		Operations(admissionregistrationv1beta1.Create, admissionregistrationv1beta1.Update).
-		WithManager(mgr).
-		ForType(&corev1.Pod{}).
 		Handlers(&podAnnotator{}).
 		Build()
```
droot: This looks a lot simpler now :)

mengqiy:
Closing since the commit will be in #300.

PR #319, opened 2019-02-08
:warning: Call MutateFn on CreateOrUpdate regardless of the operation

calind:
#313 breaks `CreateOrUpdate` by not calling the `MutateFn` when new objects are created. This patch fixes that and updates the tests to better reflect the expected behavior.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -114,48 +114,39 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 	OperationResultUpdated OperationResult = "updated"
 )
 
-// CreateOrUpdate creates or updates the given object obj in the Kubernetes
-// cluster. The object's desired state should be reconciled with the existing
-// state using the passed in MutateFn. obj is just a struct pointer so that
-// obj can be updated with the content returned by the Server.
+// CreateOrUpdate creates or updates the given object in the Kubernetes
+// cluster. The object's desired state must be reconciled with the existing
+// state inside the passed in callback MutateFn.
 //
 // The MutateFn is called regardless of creating or updating an object.
 //
 // It returns the executed operation and an error.
-func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) { // nolint: gocyclo
+func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
 	key, err := client.ObjectKeyFromObject(obj)
 	if err != nil {
 		return OperationResultNone, err
 	}
 
 	if err := c.Get(ctx, key, obj); err != nil {
-		if errors.IsNotFound(err) {
-			if err := f(obj); err != nil {
-				return OperationResultNone, err
-			}
-			if err := c.Create(ctx, obj); err != nil {
-				return OperationResultNone, err
-			}
-			return OperationResultCreated, nil
+		if !errors.IsNotFound(err) {
+			return OperationResultNone, err
 		}
-		return OperationResultNone, err
+		if err := mutate(f, key, obj); err != nil {
+			return OperationResultNone, err
+		}
+		if err := c.Create(ctx, obj); err != nil {
+			return OperationResultNone, err
+		}
+		return OperationResultCreated, nil
 	}
 
 	existing := obj.DeepCopyObject()
-	if err := f(obj); err != nil {
+	if err := mutate(f, key, obj); err != nil {
 		return OperationResultNone, err
 	}
 
 	if reflect.DeepEqual(existing, obj) {
 		return OperationResultNone, nil
-	}
-
-	newKey, err := client.ObjectKeyFromObject(obj)
-	if err != nil {
-		return OperationResultNone, err
-	}
-	if key != newKey {
-		return OperationResultNone, fmt.Errorf("MutateFn cannot mutate object namespace and/or object name")
 	}
 
 	if err := c.Update(ctx, obj); err != nil {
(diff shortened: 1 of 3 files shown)


adracus:
I had it like that but @mengqiy commented in #313 that this is maybe not the thing we want to do... 
If you're anyways working on this function, then I guess after this change, `MutateFn` is again not the right name but rather sth like `ReconcileFn`. Also, do we always want to have the pointer of the existing object being passed in to our function? This way, since it's `runtime.Object`, we'll have to typecast which somehow eliminates the niceness of this library....

calind:
The behavior was to allways call on `MutateFn`. Check https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/controller/controllerutil#example-CreateOrUpdate.

`obj` is not the desired state and `CreateOrUpdate` cannot reliably work by just passing in a desired object state. That is because some fields are immutable (eg. a deployment selector) and there is not logic to detect which ones. So here is the need to have a user defined behavior for this situations, thus the `MuateFn`.

`MutateFn` needs the pointer passed-in since it need access to existing state omehow as it's role is to mutate the existing state towards desired state.

adracus:
Yes but doesn't the caller of `CreateOrUpdate` anyways have to pass in a pointer to the object that should be created / updated? If yes, then he should also be able to supply a closure that reconciles the object accordingly

calind:
I understand your point now.

I think that can work but it's a breaking change and might deserve another PR. For now you can still get away without casting like this:
```go
controllerutil.CreateOrUpdate(context.TODO(), c, deploy, func(_ runtime.Object) error {
    deploy.Spec.Replicas = 5
    return nil
})
```

adracus:
Ok so I'm not the only one seeing this 😀 anyways then I'm also fine with it

mengqiy:
I think the debate is if `CreateOrUpdate` should do real mutation work.
It seems in the very first implementation it did real mutation work.
IMO `CreateOrUpdate` should behave like `server-side apply` which ensure an object exists in the server with desired state. It should takes a `ReconcileFn`, since we may want to preserve some fields set by the server. `ReconcileFn` should be able to reconcile the provided object and the live object got from the server and `merge` them and update the live object. e.g. when updating a service, I may want to preserve the `.spec.clusterIP` and `.spec.port[*].nodePort`.

#313 has not been released yet, so we still have time to decide what we want `CreateOrUpdate` to do.

@DirectXMan12 what was the initial intend of this method? I will let you make the final call since you are the reviewer of that PR.

calind:
Probably I'm biased here doing the original implementation but my take on this is that is better to keep the things the way they are with `CreateOrUpdate` and when server-side apply is ready provide a method for that.

Otherwise controller-runtime users will probably end up implementing something that's like server-side apply in `MutateFn/ReconcileFn` which is less than ideal. Probably server-side apply is not mainstream yet because it's hard to do it and I think that users having to implement it on their own is very prone to errors.

The breaking change that I'd be inclined to do is to remove or make optional the `obj` from `MutateFn` as per @adracus suggestion since you can allways use the passed-in `obj` to the `CreateOrUpdate`.

Btw. we're relying to the original behavior on 3 operators already and the feedback on the original PR was that it enables good patterns in user's code.

calind:
This is the original PR: #98. I think it's worth revisiting it first.

DirectXMan12:
I'm :+1: on changing `MutateFn` to not take an additional argument.  I think it makes sense, especially since the user already has to pass that in.

I'm also :+1: on fixing this to always call `MutateFn` -- I liked the original behavior, because it prevented bugs where you accidentally put different date in place in create and update.

DirectXMan12:
(now is the time to make breaking changes, if we want to)

DirectXMan12:
P.S. We should have *really* obvious tests that make it clear that this is the desired behavior as well.

calind:
I've removed the parameter in `MutateFn`. The tests were improved in previous commit. @DirectXMan12 PTAL.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -116,18 +116,26 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 
 // CreateOrUpdate creates or updates the given object obj in the Kubernetes
 // cluster. The object's desired state should be reconciled with the existing
-// state using the passed in ReconcileFn. obj must be a struct pointer so that
+// state using the passed in MutateFn. obj is just a struct pointer so that
```
DirectXMan12: nit: docs are wrong now (nothing is passed to mutate function, the sentence about `obj` needs to be updated too)

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -116,18 +116,26 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
...
 // It returns the executed operation and an error.
-func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
+func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) { // nolint: gocyclo
 	key, err := client.ObjectKeyFromObject(obj)
 	if err != nil {
 		return OperationResultNone, err
 	}
 
 	if err := c.Get(ctx, key, obj); err != nil {
 		if errors.IsNotFound(err) {
```
DirectXMan12: can improve nesting and readability here a bit by inverting this:

```suggestion
if !errors.IsNotFound(err) {
    return OperationResultNone, err
}
if err := f(); err != nil {
```

DirectXMan12: will probably fix the cyclo issue too.

calind:
Implemented the feedback and also addressed the gocyclo issue. @DirectXMan12 PTAL.

PR #322, opened 2019-02-11
:bug: Webhook server uses its configured host if set

craigfurman:
**Update**: Since this PR was originally made, the ability to configure the webhook server host was removed from master. Rather than fixing a bug, this PR now introduces a new minor, non-breaking feature.

I would still like to see this merged if possible, as it's handy for integration testing using the webhook server on developer machines that are (perhaps a little overzealously) configured with a firewall to loudly complain if new server processes begin listening on non-loopback interfaces.

~~The webhook server can be configured with a host to bind to, defaulting to all IPs (empty string) if not set. This was not actually used.~~

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -36,6 +36,10 @@ const (
 
 // ServerOptions are options for configuring an admission webhook server.
 type ServerOptions struct {
+	// Address that the server will listen on.
+	// Defaults to "" - all addresses.
+	Host string
+
 	// Port is the port number that the server will serve.
 	// It will be defaulted to 443 if unspecified.
 	Port int32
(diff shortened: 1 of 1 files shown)


craigfurman:
I just signed the CLA

pkg/webhook/server.go
```diff
@@ -218,9 +218,13 @@ func (s *Server) Start(stop <-chan struct{}) error {
...
 func (s *Server) run(stop <-chan struct{}) error { // nolint: gocyclo
 	errCh := make(chan error)
+	host := ""
+	if s.Host != nil {
+		host = *s.Host
+	}
 	serveFn := func() {
 		s.httpServer = &http.Server{
-			Addr:    fmt.Sprintf(":%v", s.Port),
+			Addr:    fmt.Sprintf("%s:%d", host, s.Port),
```
DirectXMan12: This breaks on IPv6.  Use `net.JoinHostPort` instead.

craigfurman:
Cheers, ready for another review now

craigfurman:
(and thanks for the speedy response!)

DirectXMan12:
needs rebase, otherwise looks good :+1:

craigfurman:
Rebased. I've updated the top message to re-state what this PR does, because since I made it, upstream has removed the webhook server_test file, and removed the configurability of the host.

PR #323, opened 2019-02-11
:warning: webhook redesign for generic case

DirectXMan12:
While implementing a webhook, I found some tech debt and rough corners around the webhook code.  This is an attempt to smooth some of that over.


DirectXMan12:
cc @mengqiy

DirectXMan12:
(built on top of #300 )

pkg/webhook/server.go
```diff
@@ -47,27 +48,19 @@ type ServerOptions struct {
...
-	// err will be non-nil if there is an error occur during initialization.
-	err error // nolint: structcheck
-}
-
-// Server is an admission webhook server that can serve traffic and
-// generates related k8s resources for deploying.
-type Server struct {
-	// Name is the name of server
-	Name string
+	// TODO(directxman12): should we make the mux configurable?
```
mengqiy: Is there a use case in your head that the user may want to configure mux?

pkg/webhook/server.go
```diff
@@ -47,27 +48,19 @@ type ServerOptions struct {
...
 
-	// ServerOptions contains options for configuring the admission server.
-	ServerOptions
-
-	sMux *http.ServeMux
-	// registry maps a path to a http.Handler.
-	registry map[string]Webhook
+	// webhookMux is the multiplexer that handles different webhooks.
+	webhookMux *http.ServeMux
+	// hooks keep track of all registered webhooks for dependency injection,
```
mengqiy: nit: do we want to use `hooks` or `webhooks` here?

pkg/webhook/server.go
```diff
@@ -53,256 +50,111 @@ type ServerOptions struct {
...
 
-// Handle registers a http.Handler for the given pattern.
-func (s *Server) Handle(pattern string, handler http.Handler) {
-	s.sMux.Handle(pattern, handler)
-}
+// instrumentedHook adds some instrumentation on top of the given webhook.
+func instrumentedHook(path string, hookRaw http.Handler) http.Handler {
+	return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
+		startTS := time.Now()
+		defer metrics.RequestLatency.WithLabelValues(path).Observe(time.Now().Sub(startTS).Seconds())
```
droot: We need to wrap it with func{} otherwise the args `time.Now().Sub(startTS).Seconds()` will be evaluated immediately reporting almost no latency :)

pkg/webhook/server.go
```diff
@@ -82,72 +75,47 @@ type Webhook interface {
...
-		s.sMux.Handle(webhook.GetPath(), webhook)
+		s.webhooks[webhook.GetPath()] = webhooks[i]
+		s.webhookMux.Handle(webhook.GetPath(), webhook)
 	}
 
 	return nil
 }
 
-// Handle registers a http.Handler for the given pattern.
-func (s *Server) Handle(pattern string, handler http.Handler) {
```
mengqiy: The original intend was to give the user flexibility to add additional handlers. e.g. `statusz`, `healthz` etc.

example/mutatingwebhook.go
```diff
@@ -57,15 +54,6 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 	return admission.PatchResponseFromRaw(req.AdmissionRequest.Object.Raw, marshaledPod)
 }
 
-// mutatePodsFn add an annotation to the given pod
-func (a *podAnnotator) mutatePodsFn(ctx context.Context, pod *corev1.Pod) error {
```
mengqiy: The intend of having this method is that in most cases the user's business logic should be contained in this method. Anything else in `Handle()` method should be unchanged.
So IMO it's better to keep it.

example/validatingwebhook.go
```diff
@@ -44,26 +41,16 @@ func (v *podValidator) Handle(ctx context.Context, req admission.Request) admiss
...
 	}
 
-	allowed, reason, err := v.validatePodsFn(ctx, pod)
-	if err != nil {
-		return admission.ErrorResponse(http.StatusInternalServerError, err)
-	}
-	return admission.ValidationResponse(allowed, reason)
-}
-
-func (v *podValidator) validatePodsFn(ctx context.Context, pod *corev1.Pod) (bool, string, error) {
```
mengqiy: Same comment here.

pkg/webhook/server.go
```diff
@@ -54,7 +56,7 @@ type Server struct {
 	webhookMux *http.ServeMux
 	// hooks keep track of all registered webhooks for dependency injection,
 	// and to provide better panic messages on duplicate webhook registration.
-	webhooks map[string]Webhook
+	webhooks map[string]http.Handler
```
mengqiy: I did this in #300 too.

pkg/webhook/server.go
```diff
@@ -87,25 +77,26 @@ func (s *Server) setDefaults() {
...
-			return fmt.Errorf("can't register duplicate path: %v", webhook.GetPath())
-		}
-		s.webhooks[webhook.GetPath()] = webhooks[i]
-		s.webhookMux.Handle(webhook.GetPath(), webhook)
+// Register marks the given webhook as being served at the given path.
+// It panics if two hooks are registered on the same path.
+func (s *Server) Register(path string, hook http.Handler) {
+	_, found := s.webhooks[path]
+	if found {
+		panic(fmt.Errorf("can't register duplicate path: %v", path))
```
mengqiy: In #300 I have re-order the code here a bit.
Moving `s.webhookMux.Handle(path, instrumentedHook(path, hook))` to the beginning. And it actually does the check for duplicate paths.

pkg/webhook/admission/http.go
```diff
@@ -98,14 +89,6 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func (wh *Webhook) writeResponse(w io.Writer, response Response) {
-	if response.Result.Code != 0 {
-		if response.Result.Code == http.StatusOK {
-			metrics.TotalRequests.WithLabelValues(wh.Path, "true").Inc()
```
mengqiy: It seems metrics `TotalRequests` has been dropped.
Is it not useful?

pkg/webhook/admission/http.go
```diff
@@ -24,38 +24,28 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics"
 )
 
-var admissionv1beta1scheme = runtime.NewScheme()
-var admissionv1beta1schemecodecs = serializer.NewCodecFactory(admissionv1beta1scheme)
+var admissionScheme = runtime.NewScheme()
```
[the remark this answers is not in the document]
mengqiy: This decoder only need to understand `AdmissionReviewRequest`. So IMO no need to be injected.

pkg/webhook/server.go
mengqiy: Good catch!
`The deferred call's arguments are evaluated immediately, but the function call is not executed until the surrounding function returns.`
Ref: https://tour.golang.org/flowcontrol/12

DirectXMan12:
@mengqiy this is ready for more review.  I'll write a better description tomorrow.  Still want to follow it up with a better story around generating patches and returning errors, perhaps.

pkg/webhook/admission/multi.go
```diff
@@ -0,0 +1,124 @@
...
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	"github.com/appscode/jsonpatch"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+)
+
+type multiMutating []Handler
+func (hs multiMutating) Handle(ctx context.Context, req Request) Response {
+	patches := []jsonpatch.JsonPatchOperation{}
+	for _, handler := range hs {
+		resp := handler.Handle(ctx, req)
```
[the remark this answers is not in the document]
DirectXMan12: maybe, but I almost want to deprecate this.

pkg/webhook/server.go
DirectXMan12: ack, yep

example/main.go
```diff
@@ -92,14 +92,15 @@ func main() {
...
-	as, err := webhook.NewServer("foo-admission-server", mgr, webhook.ServerOptions{
+	as, err := webhook.NewServer("foo-admission-server", webhook.ServerOptions{
 		Port:    9876,
 		CertDir: "/tmp/cert",
 	})
 	if err != nil {
 		entryLog.Error(err, "unable to create a new webhook server")
 		os.Exit(1)
 	}
+	mgr.Add(as)
```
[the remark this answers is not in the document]
DirectXMan12: Start does the deps injection, really. Add just records the `setFields` method for later.

pkg/webhook/server.go
DirectXMan12: if you want to serve on an existing handler more easily.  It's not clear that there's a concrete usecase, but I can imagine someone saying something at some points.

DirectXMan12: (e.g. your comment below about `Handle`)

pkg/webhook/server.go
DirectXMan12: right, but they shouldn't be doing it on the webhook server.  They should do it elsewhere, or we should let people pass in their own mux.

example/mutatingwebhook.go
DirectXMan12: IMO, it makes the example harder to follow, which is why I removed it.

pkg/webhook/server.go
DirectXMan12: handle does a check for duplicate paths itself, but the message might not be super-nice -- this presents us a way to have a slightly nicer message.

pkg/webhook/admission/http.go
DirectXMan12: It's not clear how useful it is (accepting or rejecting is a normal occurrence -- you really probably want to see errors and error codes if you're trying to alert, etc), and it was difficult to work in the new code structure.  There may be a way to implement it generically, but I figured it was fine for a follow-up PR + the scream test.

pkg/webhook/admission/http.go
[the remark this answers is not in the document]
droot: Got it.

pkg/webhook/admission/multi.go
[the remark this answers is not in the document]
droot: ok.

droot:
Looks good to me. Will defer it to @mengqiy for the final LGTM.

pkg/webhook/admission/multi.go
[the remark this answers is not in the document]
mengqiy: > maybe, but I almost want to deprecate this.

Reasons?

pkg/webhook/server.go
mengqiy: comment still doesn't match field name

pkg/webhook/server.go
mengqiy: Ha, I got what you were saying from https://github.com/kubernetes-sigs/controller-runtime/pull/323/commits/38e611d7f34a259b36892bbdad7f64800123ca9d#r258328244.
Then IMO we should make `webhookMux` field public.

mengqiy: > (e.g. your comment below about `Handle`)

We probably should make `webhookMux` field configurable.

pkg/webhook/admission/http_test.go
```diff
@@ -151,16 +104,28 @@ type nopCloser struct {
 func (nopCloser) Close() error { return nil }
 
 type fakeHandler struct {
-	invoked bool
-	fn      func(context.Context, atypes.Request) atypes.Response
+	invoked        bool
+	fn             func(context.Context, Request) Response
+	decoder        *Decoder
+	injectedString string
```
mengqiy: Is it actually tested anywhere?

pkg/webhook/admission/decode.go
```diff
@@ -34,6 +36,21 @@ func NewDecoder(scheme *runtime.Scheme) (*Decoder, error) {
 
 // Decode decodes the inlined object in the AdmissionRequest into the passed-in runtime.Object.
 func (d *Decoder) Decode(req Request, into runtime.Object) error {
+	// NB(directxman12): there's a bug/weird interaction between decoders and
```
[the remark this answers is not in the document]
DirectXMan12: ack, need to file the issue

example/mutatingwebhook.go
```diff
@@ -23,67 +23,51 @@ import (
...
-	if pod.Annotations == nil {
-		pod.Annotations = map[string]string{}
-	}
-	pod.Annotations["example-mutating-admission-webhook"] = "foo"
-	return nil
-}
-
 // podAnnotator implements inject.Client.
 // A client will be automatically injected.
-var _ inject.Client = &podAnnotator{}
```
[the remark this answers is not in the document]
DirectXMan12: It doesn't really add much, and necessitates an extra import.  I can add it back in if we really want it, but I think the comment suffices here.

pkg/webhook/admission/http_test.go
DirectXMan12: yep, pkg/webhook/admission/webhook_test.go

PR #324, opened 2019-02-12
[WIP] :warning: Add DeleteCollection to the client Interface.

MadVikingGod:
This Adds the DeleteCollection to the client Writer Interface.  This would be used to delete objects based on label selection, instead of using Delete on each object individually.

#### Things I would appreciate feedback on:
- DeleteCollectionOptions - This is currently a superset of ListOptions and DeleteOptions.  Is there a different way to build the "functional options pattern" for two sets of options?
- The Obj passed to DeleteCollection() is just used to determine the GVK/GVR, and will not delete objects in the list.  Is there a way to signal that?


MadVikingGod:
I signed it.

DirectXMan12:
> * DeleteCollectionOptions - This is currently a superset of ListOptions and DeleteOptions.  Is there a different way to build the "functional options pattern" for two sets of options?

Other than just reproducing the different functions?  Not really :-/.  We could *maybe* make things more complicated elsewhere to avoid this, but I'm not sure its worth it.

> * The Obj passed to DeleteCollection() is just used to determine the GVK/GVR, and will not delete objects in the list.  Is there a way to signal that?

No, we don't really have a good way to signal that.

There is one other design to consider -- something like

`Delete(pod, someDeleteOption, CollectionMatching(listOptions...))` and just having that be the interface to delete-collection (i.e. it would be in the options), similarly to how namespaced vs non-namespaced is determined by an option.

DirectXMan12:
Ping, what's the state on this?  Just want to make sure you're not waiting on anything from me :-)

MadVikingGod:
Sorry for the large break, I had other schedule priorities.

I have updated this so that there is no longer a DeleteCollectionOptions, and instead there is a ListOptions as part of the DeleteOptions.  This is only used by the DeleteCollection method, and it feels more natural.

@DirectXMan12 I would appreciate it if you could look over, and let me know what is needed to move this forward.

DirectXMan12:
I'll take a look.  I'm still wondering if there's a good reason to separate the `DeleteCollection` from `Delete` methods.  Could we just have `DeleteOptions` choose one or the other based on the types and/or the presence of ListOptions in the delete options?  That feels slightly more natural to me.  WDYT?

adracus:
@MadVikingGod / @DirectXMan12 still up to getting this in? Pretty nice feature I'd also like to have :) If you need any help, let me know!

DirectXMan12:
yeah, I'd still like to get it in.  Let's aim for before the 0.2.0 final release

adracus:
So then @MadVikingGod will you finish or should I take over ?

MadVikingGod:
Sorry, I haven't had the time to dedicate to following up on this.

You are more than welcome to take this over.  I have signed the CLA, I don't know why it is not showing.

adracus:
Since I'm not a maintainer I couldn't modify your PR, @MadVikingGod. I now went ahead with a different PR :neutral_face: : #447 . Also mentioned you as a co-author, although the main work really has been done by you, hope this is ok for you. Otherwise, I guess the PR contains the suggested changes by @DirectXMan12 (not having a `DeleteCollection` method but extending the `DeleteOptions` with `CollectionOptions`).

MadVikingGod:
It's all good. Just as long as it makes the project better.

I will close this PR out Tuesday

MadVikingGod:
Lost track of this, I'm very sorry.  Thank you again for the time, and@adracus for taking it up in the other pr.

PR #325, opened 2019-02-12
:docs: Clarify IndexerFunc comment with namespaced objects

vincepri:

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -89,7 +89,8 @@ type Client interface {
 }
 
 // IndexerFunc knows how to take an object and turn it into a series
-// of (non-namespaced) keys for that object.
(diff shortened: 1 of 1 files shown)


pkg/client/interfaces.go
```diff
@@ -89,7 +89,7 @@ type Client interface {
 }
 
 // IndexerFunc knows how to take an object and turn it into a series
-// of (non-namespaced) keys for that object.
+// of non-namespaced keys. Namespaced objects are automatically wrapped.
```
DirectXMan12: maybe make this `Namespaced objects are automatically given namespaced and non-spaced variants, so keys do not need to include namespace`.

vincepri: Fixed!

DirectXMan12:
/lgtm
/approve

thanks!

PR #326, opened 2019-02-13
:bug: Ensure that enqueuemapped's mapper gets deps

DirectXMan12:
Previously, we weren't injecting dependencies into the Mapper in the
enqueue-mapped handler.  This makes more complicated to things that
require clients, etc.

### DIFF
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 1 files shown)


pkg/handler/enqueue_mapped.go
```diff
@@ -68,6 +69,16 @@ func (e *EnqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInter
 	}
 }
 
+// EnqueueRequestsFromMapFunc can inject fields into the mapper.
+
+// InjectFunc implements inject.Injector.
```
mengqiy: I suggest you add something like below to ensure it implements the interface
```
var _ inject.Injector = EnqueueRequestsFromMapFunc(nil)
```

droot: +1

PR #327, opened 2019-02-13
:sparkles: adding APIReader to the manager and injection.

shawn-hurley:
Adding the APIReader onto the manager and allowing users to use dependency injection to add this to their types.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -58,6 +58,9 @@ type controllerManager struct {
 	// client is the client injected into Controllers (and EventHandlers, Sources and Predicates).
 	client client.Client
 
+	// apiReader is the reader that will make requests to the api server and not the cache.
+	apiReader client.Reader
+
 	// fieldIndexes knows how to add field indexes over the Cache used by this controller,
 	// which can later be consumed via field selectors from the injected client.
 	fieldIndexes client.FieldIndexer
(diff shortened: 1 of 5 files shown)


pkg/manager/internal.go
```diff
@@ -61,6 +61,9 @@ type controllerManager struct {
 	// client is the client injected into Controllers (and EventHandlers, Sources and Predicates).
 	client client.Client
 
+	// apiReader is the reader that will make requests to the api server and not the cache.
+	apiReader client.Reader
```
droot: Is there a reason for not making it `client.Client` ? [implementation wise client.Client is already supported]. Though it does get tricky with the naming if we decide to change the type here `apiClient` ? maybe `directClient` ?

shawn-hurley: My thought process was

1. When you need this field you really only need the reader. This still points people to the client for most use cases. I was hoping the fact it did not have any write mechanisms people would only use this when they needed it. 

2. Naming is the other reason. I think the idea that we don't really expose "direct" vs "cache" and try to hide this decision as much as possible is a good thing. This is just a step in adding an escape hatch but not exposing that decision still. 

Thoughts?

DirectXMan12: those reasons @shawn-hurley make a lot of sense -- +1: from me to having it be a Reader.

droot: Sorry for delayed response. Haven't come across `write` use case, so going with the restricted use-case sounds good.

PR #328, opened 2019-02-14
:warning: Simplify the webhook interface

mengqiy:
If a CRD type implements the `Defaulter` and `Validator` interface, we will generate the admission webhook for the CRD type.
For non-CRD type, we provide methods to pass in path and handlers to create the mutating|validting webhook.

godoc and tests are on the way.


pkg/webhook/server.go
```diff
@@ -53,256 +52,145 @@ type ServerOptions struct {
...
+// For registers defaulting and validation webhooks for the provided types that implement either
+// webhooktuil.Validator or webhookutil.Defaulter.
+func (s *Server) For(objects ...runtime.Object) error {
+	// Create a webhook for each type
+	for _, object := range objects {
+		wh := NewResourceWebhook(object)
+		if wh == nil {
+			continue
+		}
+		s.Register("path", wh)
```
droot: We can generate the path using object's GVK and name. And, the default path generation should probably be shared with controller-tools so that generator is in sync with this code ?

pkg/webhook/server.go
```diff
@@ -53,256 +52,145 @@ type ServerOptions struct {
...
-	// manager is the manager that this webhook server will be registered.
-	manager manager.Manager
-
-	// httpServer is the actual server that serves the traffic.
-	httpServer *http.Server
-
-	once sync.Once
+// For registers defaulting and validation webhooks for the provided types that implement either
+// webhooktuil.Validator or webhookutil.Defaulter.
+func (s *Server) For(objects ...runtime.Object) error {
```
droot: `For` name itself doesn't convey the intent. Wondering if we should give it a more friendly name like `EnableValidationFor` or `EnableDefaultsFor` ?

mengqiy:
Addressed comments.
PTAL
If looks OK, I will push more tests and godoc.

crdexample/pkg/resource.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Resource is required by pkg/client/listers/...
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
+
+// TODO: figure out should we encourage the users to do a wrapper for a core type object.
+type MyPod struct {
+	corev1.Pod
+}
```
droot: Its not clear to me in this case whether the webhook gets generated for the `corev1.Pods` (the embedded type) or the wrapped type ?

crdexample/pkg/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,127 @@
...
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new FirstMateStatus.
+func (in *FirstMateStatus) DeepCopy() *FirstMateStatus {
+	if in == nil {
+		return nil
+	}
+	out := new(FirstMateStatus)
+	in.DeepCopyInto(out)
+	return out
+}
```
droot: I have seen these example projects add too much noise to the repo and introduce additional maintenance burden. Is there a way to avoid it ?

pkg/builder/build.go
```diff
@@ -148,6 +148,18 @@ func (blder *Builder) Build(r reconcile.Reconciler) (manager.Manager, error) {
 		return nil, err
 	}
 
+	if blder.mgr.GetWebhookServer() != nil {
```
droot: nit: we can combine the if conditional with the svr assignment.
```Go
 if svr := blder.mgr.GetWebhookServer(); svr != nil {
}
```

pkg/builder/build.go
```diff
@@ -148,6 +148,18 @@ func (blder *Builder) Build(r reconcile.Reconciler) (manager.Manager, error) {
...
+		svr := blder.mgr.GetWebhookServer()
+		err := svr.EnableDefaultsFor(blder.mgr.GetScheme(), blder.apiType)
+		if err != nil {
+			return nil, err
+		}
+		err = svr.EnableValidationFor(blder.mgr.GetScheme(), blder.apiType)
```
droot: Won't `svr` have access to the scheme ? If yes, then we can get rid of scheme parameter from `EnableDefaultsFor` and `EnableValidationFor`.

pkg/manager/internal.go
```diff
@@ -90,6 +91,8 @@ type controllerManager struct {
 	internalStopper chan<- struct{}
 
 	startCache func(stop <-chan struct{}) error
+
+	webhookServer *webhook.Server
```
droot: is manager suppose to handle lifecycle of this webhookServer, then we should probably make it a `Runnable` ?

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+	"net/http"
+
+	"k8s.io/api/admission/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// Defaultor defines functions for setting defaults on resources
+type Defaulter interface {
```
droot: typo: name in comment doesn't match.

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+type mutatingHandler struct {
+	defaulter Defaulter
+	scheme    *runtime.Scheme
+	decoder   *admission.Decoder
+}
+
+var _ inject.Scheme = &mutatingHandler{}
+
+// InjectDecoder injects the decoder into the FirstMateDeleteHandler
+func (h *mutatingHandler) InjectScheme(s *runtime.Scheme) error {
```
droot: comment needs to be rewritten.

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+//var _ admission.DecoderInjector = &mutatingHandler{}
+//
+//// InjectDecoder injects the decoder into the FirstMateDeleteHandler
+//func (h *mutatingHandler) InjectDecoder(d *admission.Decoder) error {
+//	h.decoder = d
+//	return nil
+//}
```
droot: we don't need this ?

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Handle handles admission requests.
+func (h *mutatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	var err error
+	h.decoder, err = admission.NewDecoder(h.scheme)
+	if err != nil {
+		log.Error(err, "unable to get a decoder from the scheme")
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
```
droot: I am sure we must have explored that but wanted to understand why this can't be done `once` for this handler at the start-up or may be at the time of scheme injection ?

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	// Create the patch
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshalled)
+}
+
+// newValidatingWebhookFor creates a new WebhookBuilder for Validation and Defaulting the provided type.  Returns nil
+// if o does not implement either validator or defaulter.
```
droot: May be move the `validation` stuff in a separate file validation.go ?

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+type validatingHandler struct {
+	validator Validator
+	scheme    *runtime.Scheme
+	decoder   *admission.Decoder
+}
+
+var _ inject.Scheme = &validatingHandler{}
+
+// InjectDecoder injects the decoder into the FirstMateDeleteHandler
+func (h *validatingHandler) InjectScheme(s *runtime.Scheme) error {
```
droot: typo: comments need to be rewritten.

pkg/webhook/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+//}
+
+// Handle handles admission requests.
+func (h *validatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	var err error
+	h.decoder, err = admission.NewDecoder(h.scheme)
+	if err != nil {
+		log.Error(err, "unable to get a decoder from the scheme")
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
```
droot: same comment as above.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
 	}
 }
 
+// For registers defaulting and validation webhooks for the provided types that implement either
+// webhooktuil.validator or webhookutil.defaulter.
+func (s *Server) EnableDefaultsFor(scheme *runtime.Scheme, objects ...runtime.Object) error {
```
droot: we can make scheme injected from manager and get rid of this parameter ?

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+		}
+
+		gvks, _, err := scheme.ObjectKinds(object)
+		if err != nil {
+			return err
+		}
+		if len(gvks) != 1 {
+			return fmt.Errorf("expected only GVK returned by scheme.ObjectKinds")
+		}
+		fmt.Println("/mutate-" + generatePath(gvks[0]))
```
droot: logging this info at some verbosity level may not be a bad idea, can help in debugging.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+		if wh == nil {
+			continue
+		}
+
+		gvks, _, err := scheme.ObjectKinds(object)
+		if err != nil {
+			return err
+		}
+		if len(gvks) != 1 {
+			return fmt.Errorf("expected only GVK returned by scheme.ObjectKinds")
```
droot: we should some webhook related context to the error message and also include all the GVKs to help in debugging.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+		}
+		fmt.Println("/mutate-" + generatePath(gvks[0]))
+		s.Register("/mutate-"+generatePath(gvks[0]), wh)
+	}
+	return nil
+}
+
+// For registers defaulting and validation webhooks for the provided types that implement either
+// webhooktuil.validator or webhookutil.defaulter.
+func (s *Server) EnableValidationFor(scheme *runtime.Scheme, objects ...runtime.Object) error {
```
droot: comment is out of sync due to renaming.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+			return err
+		}
+		if len(gvks) != 1 {
+			return fmt.Errorf("expected only GVK returned by scheme.ObjectKinds")
+		}
+		fmt.Println("/validate-" + generatePath(gvks[0]))
+		s.Register("/validate-"+generatePath(gvks[0]), wh)
+	}
+	return nil
+}
```
droot: I am sort of wondering if we should enable validation/defaulting by default when webhook server starts ? (This will save people one step if they want to enable validation/defaulting). And I am thinking of doing the same thing for conversion.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+// PathWithMutateFn creates a mutating webhook with the provided handlers and then
+// register it with the given path.
+func (s *Server) PathWithMutateFn(path string, handlers ...admission.Handler) {
+	var h http.Handler
+	h = &admission.Webhook{
+		Handler: admission.MultiMutatingHandler(handlers...),
+	}
+	// Register all of the types
+	s.Register(path, h)
+}
```
droot: Name of the function doesn't make sense.

pkg/webhook/server.go
```diff
@@ -82,6 +86,74 @@ func (s *Server) setDefaults() {
...
+// PathWithValidateFn creates a validating webhook with the provided handlers and then
+// register it with the given path.
+func (s *Server) PathWithValidateFn(path string, handlers ...admission.Handler) {
+	var h http.Handler
+	h = &admission.Webhook{
+		Handler: admission.MultiValidatingHandler(handlers...),
+	}
+	// Register all of the types
+	s.Register(path, h)
+}
```
droot: Name of the function doesn't make sense.

crdexample/pkg/resource.go
mengqiy: This was some experimental code, but they actually sneak in.
I will remove this piece.
Sorry for the confusion.

pkg/builder/build.go
mengqiy: `svr` won't be able to have the Scheme until the inject time.
But the setup time is earlier than the inject time.

Another option is having 2 rounds of injection:
1st round injects deps for the server
2nd round inject deps to the handlers under the server.

It's not clear to me which way is better.

pkg/manager/internal.go
mengqiy: `*webhook.Server` IS a `Runable`.

pkg/webhook/server.go
mengqiy: Same comment as https://github.com/kubernetes-sigs/controller-runtime/pull/328#discussion_r259543868

pkg/webhook/server.go
mengqiy: My debugging code sneaked in.
I agree it can be converted to logging.
Will fix it

mengqiy:
Addressed comments. PTAL at the interfaces.
2 things are still open for discussion:
- should we do https://github.com/kubernetes-sigs/controller-runtime/pull/328#discussion_r259461431 
- should we have an example for CRD like the one in commit https://github.com/kubernetes-sigs/controller-runtime/pull/328/commits/b51eb44f9b3bca8a827817f71f94e221d793f1e3

@DirectXMan12 WDYT

pkg/builder/build.go
```diff
@@ -230,3 +234,22 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
 }
+
+func (blder *Builder) doWebhook() error {
+	svr := &webhook.Server{}
+	err := svr.EnableWebhooksFor(blder.apiType)
+	// If no error, we add the Server instance to the controller manager.
+	// If we get a NotImplementDefaulterValidatorInterfacesError, we discord the Server instance.
+	// If other type of error, surface it.
+	if err == nil {
+		e := blder.mgr.Add(svr)
```
droot: quick question: shouldn't this be done before calling `svr.EnableWebhooksFor` anyways to ensure mgr injects dependencies for webhooks (mainly scheme) ?

pkg/webhook/defaulter.go
```diff
@@ -0,0 +1,90 @@
...
+	decoder   *admission.Decoder
+}
+
+var _ inject.Scheme = &mutatingHandler{}
+
+// InjectScheme injects the scheme into mutatingHandler and initialize a decoder using the scheme.
+func (h *mutatingHandler) InjectScheme(s *runtime.Scheme) error {
+	h.scheme = s
+	var err error
+	h.decoder, err = admission.NewDecoder(h.scheme)
```
droot: +1 on setting decoder this way.

pkg/webhook/server.go
```diff
@@ -74,14 +92,67 @@ func (s *Server) setDefaults() {
...
 		s.CertDir = path.Join("/tmp", "k8s-webhook-server", "serving-certs")
 	}
 }
 
+// EnableWebhooksFor registers mutating (defaulting) and validation webhooks for the provided types that
+// implement Defaulter and (or) Validator. Passing in core types doesn't work, since the core types implement
+// neither the Defaulter nor the Validator interfaces.
+// It returns a NotImplementDefaulterValidatorInterfacesError error if a provided object implements
+// neither the Defaulter interface nor the Validator interface.
+func (s *Server) EnableWebhooksFor(objects ...runtime.Object) error {
```
DirectXMan12: I'm not convinced this needs to be a method (as opposed to a helper or helpers).  The helper makes it more composable, easier to test individually, and lets us actually have a type signature that requires an object that implements the interfaces (removing the error path in favor of compile-time checking).

pkg/webhook/server.go
```diff
@@ -172,3 +243,15 @@ func (s *Server) InjectFunc(f inject.Func) error {
 	s.setFields = f
 	return nil
 }
+
+func generatePath(gvk schema.GroupVersionKind) string {
+	var pathItems []string
+	splittedGroup := strings.Split(gvk.Group, ".")
```
DirectXMan12: why aren't we just `strings.Replace(gvk.Group, ".", "-")` or something?

pkg/webhook/validator.go
```diff
@@ -0,0 +1,108 @@
...
+	decoder   *admission.Decoder
+}
+
+var _ inject.Scheme = &validatingHandler{}
+
+// InjectDecoder injects the decoder into the FirstMateDeleteHandler
+func (h *validatingHandler) InjectScheme(s *runtime.Scheme) error {
+	h.scheme = s
+	var err error
+	h.decoder, err = admission.NewDecoder(h.scheme)
```
DirectXMan12: decoders should automatically be injected with `InjectDecoder` by the `admission.Webhook` logic.

pkg/builder/build.go
```diff
@@ -218,3 +234,22 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	blder.ctrl, err = newController(name, blder.mgr, controller.Options{Reconciler: r})
 	return err
 }
+
+func (blder *Builder) doWebhook() error {
+	svr := &webhook.Server{}
```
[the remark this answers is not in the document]
DirectXMan12: or some other design -- but it needs to work for multiple controllers in a single binary

pkg/webhook/server.go
```diff
@@ -74,14 +92,67 @@ func (s *Server) setDefaults() {
...
+	}
+
+	// Create a webhook for each type
+	for _, object := range objects {
+		gvk, err := apiutil.GVKForObject(object, s.Scheme)
+		if err != nil {
+			return nil
+		}
+
+		mwh := newDefaultingWebhookFor(object)
```
DirectXMan12: we should opportunistically merge validating and defaulting hooks into a single hook when possible, to avoid the extra round-trip

mengqiy: > we should opportunistically merge validating and defaulting hooks into a single hook when possible, to avoid the extra round-trip

This means you want to put defaulting and validating hooks in one mutatingWebhook. But doing it this way is not safe, since there may be another mutating webhook changes the object after your validation logic. The safe way to do it is making validation logic in a validating webhook, which is always run after all mutating webhooks.

pkg/webhook/server.go
mengqiy: Curret behavior doesn't include domain name in the path:
```
Group: apps.example.com
Version: v1
Kind: Foo
```
I will get `/mutate-apps-v1-foo` instead of `/mutate-apps-example-com-v1-foo`. IMO it's more concise, but there may be a small risk of collision. e.g. with 
```
Group: apps.bar.com
Version: v1
Kind: Foo
```

I don't have a strong opinion about if we should have domain name.
Thoughts?

mengqiy:
`pkg` dir is ready for another round of reviewing interfaces.
I haven't make a lot of changes in `examples/crd`. I'd like to discuss with @DirectXMan12 more before spending more than time on it.

pkg/webhook/server.go
DirectXMan12: I don't think conciseness here is valuable (only ever have to write this once, and it's usually going to be autogenerated), and I'd rather avoid collisions

pkg/webhook/server.go
DirectXMan12: good point :+1:

pkg/webhook/server.go
```diff
@@ -54,6 +58,9 @@ type Server struct {
 	// the user is responsible to mount the secret to the this location for the server to consume.
 	CertDir string
 
+	// Scheme knows the mapping between go struct types and GVKs.
+	Scheme *runtime.Scheme
```
DirectXMan12: what's the scheme used here for?

DirectXMan12: (the comment doesn't actually say)

pkg/webhook/admission/defaulter.go
```diff
@@ -0,0 +1,85 @@
...
+type mutatingHandler struct {
+	defaulter Defaulter
+	scheme    *runtime.Scheme
+	decoder   *Decoder
+}
+
+var _ inject.Scheme = &mutatingHandler{}
+
+// InjectScheme injects the scheme into a mutatingHandler.
+func (h *mutatingHandler) InjectScheme(s *runtime.Scheme) error {
```
DirectXMan12: what's the scheme here for?

examples/crd/main.go
```diff
@@ -0,0 +1,67 @@
...
+
+package main
+
+import (
+	"flag"
+	"os"
+
+	appsv1 "k8s.io/api/apps/v1"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/examples/crd/pkg"
```
DirectXMan12: let's use `ctrl "sigs.k8s.io/controller-runtime"` instead of most of these in the examples

examples/crd/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Build the manager binary
```
DirectXMan12: don't think we should have the deployment stuff here -- this should be only code.  Deployment examples can live in KB

examples/crd/pkg/resource.go
```diff
@@ -0,0 +1,118 @@
...
+
+// FirstMateStatus defines the observed state of firstMate
+type FirstMateStatus struct {
+	Location string
+}
+
+var _ admission.Validator = &FirstMate{}
+
+// ValidateCreate implements webhookutil.validator so a webhook will be registered for the type
+func (f *FirstMate) ValidateCreate() error {
```
DirectXMan12: should have webhook generation annotations?

examples/crd/Dockerfile
mengqiy: Ack

pkg/webhook/admission/validator.go
```diff
@@ -0,0 +1,103 @@
...
+
+// InjectDecoder injects the decoder into a validatingHandler.
+func (h *validatingHandler) InjectDecoder(d *Decoder) error {
+	h.decoder = d
+	return nil
+}
+
+// Handle handles admission requests.
+func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
+	if h.validator == nil {
```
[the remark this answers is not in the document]
mengqiy: Done

pkg/builder/build.go
```diff
@@ -218,3 +240,60 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	blder.ctrl, err = newController(name, blder.mgr, controller.Options{Reconciler: r})
 	return err
 }
+
+func (blder *Builder) doWebhook() error {
+	if blder.mgr.GetScheme() == nil {
```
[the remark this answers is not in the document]
mengqiy: Was being more defensive here. But if you think it's not necessary at all, I will drop it.

mengqiy:
PTAL

mengqiy:
Rebased on HEAD to resolved the conflicts and squashed some commits.
Ready for the final round of review

pkg/builder/build.go
```diff
@@ -218,3 +236,36 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
+	gvk, err := apiutil.GVKForObject(blder.apiType, blder.mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+
+	partialPath := strings.Replace(gvk.Group, ".", "-", -1) + "-" +
+		gvk.Version + "-" + strings.ToLower(gvk.Kind)
+
+	defaulter, isDefaulter := blder.apiType.(admission.Defaulter)
+	if isDefaulter {
```
DirectXMan12: nit: collapse this into a compound statement.

pkg/builder/build.go
```diff
@@ -218,3 +236,36 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
+		mwh := admission.DefaultingWebhookFor(defaulter)
+		if mwh != nil {
+			path := "/mutate-" + partialPath
+			log.Info("registering a mutating webhook", "path", path)
+			blder.mgr.GetWebhookServer().Register(path, mwh)
+		}
+	}
+
+	validator, isValidator := blder.apiType.(admission.Validator)
```
DirectXMan12: nit: collapse this into a compound statement

pkg/builder/build.go
```diff
@@ -218,3 +236,36 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
+	if isValidator {
+		vwh := admission.ValidatingWebhookFor(validator)
+		if vwh != nil {
+			path := "/validate-" + partialPath
+			log.Info("registering a validating webhook", "path", path)
```
DirectXMan12: log the GVK?

pkg/builder/build.go
```diff
@@ -218,3 +236,36 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
+		if mwh != nil {
+			path := "/mutate-" + partialPath
+			log.Info("registering a mutating webhook", "path", path)
+			blder.mgr.GetWebhookServer().Register(path, mwh)
+		}
+	}
+
+	validator, isValidator := blder.apiType.(admission.Validator)
+	if isValidator {
+		vwh := admission.ValidatingWebhookFor(validator)
```
DirectXMan12: when we land conversion, we're going to want to automatically handle all registered versions of a given group-kind.  please add a comment.

DirectXMan12:
/lgtm
/approve

Can you follow up with a PR with slightly more comprehensive tests?

mengqiy:
> Can you follow up with a PR with slightly more comprehensive tests?

Sure!

mengqiy:
Forgot to remove the `do-not-merge` label last Friday, and this PR ended up with a small merge conflict with #352 in the comments.
I rebased and resolved conflicts.

PR #337, opened 2019-02-25
✨ pkg/manager/*: Add MetricsServingDisabled option

### LINKED ISSUE #336: Serve metrics independently of `manager.Start`
Currently metrics are only started to be served when `manager.Start()` is called. In operator-sdk we have a use case as we implemented our own leader election, so we block `Start()` until that pod becomes a leader. Due to this problem we are not getting metrics for all the pods. We do want to still like to collect metrics for all the running pods even if the manager is not started.

Proposed solution:

Add `MetricsServingDisabled` option to ovveride serving of metrics. Expose the `serveMetrics` https://github.com/kubernetes-sigs/controller-runtime/blob/45d652077cba5ccc29621e8db615f25084c701cb/pkg/manager/internal.go#L180 that can be used by users who want to start serving metrics independently of the `Start()` method.

LiliC:
MetricsServingDisabled option when set to true disables serving metrics
by manager.

Closes #336

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -42,10 +42,10 @@ import (
 var log = logf.RuntimeLog.WithName("manager")
 
 type controllerManager struct {
-	// config is the rest.config used to talk to the apiserver.  Required.
+	// config is the rest.config used to talk to the apiserver. Required.
 	config *rest.Config
 
-	// scheme is the scheme injected into Controllers, EventHandlers, Sources and Predicates.  Defaults
+	// scheme is the scheme injected into Controllers, EventHandlers, Sources and Predicates. Defaults
 	// to scheme.scheme.
 	scheme *runtime.Scheme
 
@@ -74,6 +74,10 @@ type controllerManager struct {
 
 	// metricsListener is used to serve prometheus metrics
 	metricsListener net.Listener
+
+	// metricsServingDisabled is used to disable the serving of metrics
+	// by default these are enabled, as the value is false.
+	metricsServingDisabled bool
 
 	mu      sync.Mutex
 	started bool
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Setting the metrics port to zero should disable serving as well, is that not working, or is it just not sufficiently intuitive (it's kind-a weird, so it's presumably easy to miss).

LiliC:
> Setting the metrics port to zero should disable serving as well, is that not working, or is it just not sufficiently intuitive (it's kind-a weird, so it's presumably easy to miss).

Sorry if I didn't make myself clear enough, we would just like to start serving the metrics at an earlier point and not when `.Start()` is called. Meaning call the exposed function in this PR `.ServeMetrics(...)`, but still pass the correct port. Hope that makes more sense. :)

DirectXMan12:
Ah, ok.  The use case makes sense, but I'd kind-of prefer to just extract the general bits of `serveMetrics` out into a helper such that it's callable separately.  Then, you can just disable metrics serving from the manager, can call `ServeMetrics` at your leisure.  Exposing the separate `DisableMetricsServing` and `ServeMetrics` seems confusing, especially since `ServeMetrics` doesn't actually use much of the internals of the manager.

DirectXMan12:
does that work for you and/or make sense?

LiliC:
@DirectXMan12 If I understand correctly we would get rid of calling `serveMetrics()` in `.Start()` and instead the user will always have to call the new helper function `ServiceMetrics` separately?

DirectXMan12:
I mean:

1. Have a helper in `pkg/metrics` called `ServeMetrics` that takes in a  `net.Listener` and does what `.serveMetrics` does (which is serve metrics from the global CR registry)

2. make `Manager#serveMetrics` just call `metrics.ServeMetrics`

3. You could just set `MetricsPort: "0"` in the manager config, and still get the same behavior by calling `metrics.ServeMetrics` manually.

This way, users continue to get the same behavior, but you can customize when you serve metrics without needing to re-implement the logic in `serveMetrics` (which basically just sets up an HTTP server and the Prometheus handler for the global CR registry):

https://github.com/kubernetes-sigs/controller-runtime/blob/6649bdbe3f1f5f9769b91c031e52be253c87e922/pkg/manager/internal.go#L170-L194

LiliC:
@DirectXMan12 Thanks for your suggestion and time to look at this! That implementation would work for our use case. I will close this PR and open a new one (probably not anytime soon as I am off to vacation for a week.) Thanks again!

DirectXMan12:
SGTM.  Enjoy your vacation :-)

PR #341, opened 2019-02-27
✨ Add Limit and Continue functional list options for client

### LINKED ISSUE #310: No list option for Continue or Limit
The new functional list options from #294 don't support setting Limit or Continue with a functional option. If the user wants to use those options, they need to fall back to `UseListOptions`. Not a huge deal, but probably not too hard to fix.

rajathagasthya:
Fixes #310

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -1904,6 +1904,89 @@ var _ = Describe("Client", func() {
 				close(done)
 			}, serverSideTimeoutSeconds)
 
+			It("should filter results using limit and continue options", func() {
+
+				makeDeployment := func(suffix string) *appsv1.Deployment {
+					return &appsv1.Deployment{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: fmt.Sprintf("deployment-%s", suffix),
+						},
+						Spec: appsv1.DeploymentSpec{
+							Selector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{"foo": "bar"},
+							},
+							Template: corev1.PodTemplateSpec{
+								ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"foo": "bar"}},
+								Spec:       corev1.PodSpec{Containers: []corev1.Container{{Name: "nginx", Image: "nginx"}}},
+							},
+						},
+					}
+				}
+
+				By("creating 4 deployments")
+				dep1 := makeDeployment("1")
+				dep1, err := clientset.AppsV1().Deployments(ns).Create(dep1)
+				Expect(err).NotTo(HaveOccurred())
+				defer deleteDeployment(dep1, ns)
+
+				dep2 := makeDeployment("2")
+				dep2, err = clientset.AppsV1().Deployments(ns).Create(dep2)
+				Expect(err).NotTo(HaveOccurred())
+				defer deleteDeployment(dep2, ns)
+
+				dep3 := makeDeployment("3")
+				dep3, err = clientset.AppsV1().Deployments(ns).Create(dep3)
+				Expect(err).NotTo(HaveOccurred())
+				defer deleteDeployment(dep3, ns)
+
+				dep4 := makeDeployment("4")
+				dep4, err = clientset.AppsV1().Deployments(ns).Create(dep4)
+				Expect(err).NotTo(HaveOccurred())
+				defer deleteDeployment(dep4, ns)
+
+				cl, err := client.New(cfg, client.Options{})
+				Expect(err).NotTo(HaveOccurred())
+
+				By("listing 1 deployment when limit=1 is used")
+				deps := &appsv1.DeploymentList{}
+				err = cl.List(context.Background(), deps,
+					client.Limit(1),
+				)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(deps.Items).To(HaveLen(1))
+				Expect(deps.Continue).NotTo(BeEmpty())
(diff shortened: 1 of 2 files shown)


rajathagasthya:
@DirectXMan12 PTAL. Not quite sure about what changes to make for the fake client though, so any pointers are appreciated.

pkg/client/client_test.go
```diff
@@ -1430,6 +1430,130 @@ var _ = Describe("Client", func() {
...
+						},
+						Template: corev1.PodTemplateSpec{
+							ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"foo": "bar"}},
+							Spec:       corev1.PodSpec{Containers: []corev1.Container{{Name: "nginx", Image: "nginx"}}},
+						},
+					},
+				}
+				dep1, err := clientset.AppsV1().Deployments(ns).Create(dep1)
+				Expect(err).NotTo(HaveOccurred())
+				dep2 := &appsv1.Deployment{
```
DirectXMan12: refactor this out into a little helper.  It can live in this test, if you want (e.g. `makeDeployment := func(suffix string) *appsv1.Deployment { ... }`)

pkg/client/client_test.go
```diff
@@ -1430,6 +1430,130 @@ var _ = Describe("Client", func() {
...
+
+				Expect(deps.Items).NotTo(BeEmpty())
+				Expect(2).To(Equal(len(deps.Items)))
+				Expect(deps.Continue).To(BeEmpty())
+				first := deps.Items[0]
+				Expect(first.Name).To(Equal(dep3.Name))
+				second := deps.Items[1]
+				Expect(second.Name).To(Equal(dep4.Name))
+
+				deleteDeployment(dep1, ns)
```
DirectXMan12: add a defer for this, or something, in case the test fails.

pkg/client/interfaces.go
```diff
@@ -290,6 +305,20 @@ func (o *ListOptions) InNamespace(ns string) *ListOptions {
...
+// SetLimit is a convenience function that sets the limit, and
+// then returns the options. It mutates the list options.
+func (o *ListOptions) SetLimit(n int64) *ListOptions {
+	o.Limit = n
+	return o
+}
+
+// SetContinueToken is a convenience function that sets the continuation
+// token, and then returns the options. It mutates the list options.
+func (o *ListOptions) SetContinueToken(token string) *ListOptions {
```
DirectXMan12: should probably match the name of the corresponding `ListOptionFunc` (i.e. the name should be `SetContinue`, or `Continue` should be `ContinueToken`)

pkg/client/client_test.go
```diff
@@ -1430,6 +1430,130 @@ var _ = Describe("Client", func() {
...
+				err = cl.List(context.Background(), deps,
+					client.Continue(continueToken),
+				)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(deps.Items).NotTo(BeEmpty())
+				Expect(2).To(Equal(len(deps.Items)))
+				Expect(deps.Continue).To(BeEmpty())
+				first := deps.Items[0]
+				Expect(first.Name).To(Equal(dep3.Name))
```
DirectXMan12: this'd be more idomatic as

```go
Expect(deps).To(HaveLength(2))
Expect(deps[0].Name).To(Equal(dep3.Name))
...
```

(same elsewhere)

pkg/client/client_test.go
```diff
@@ -1430,6 +1430,130 @@ var _ = Describe("Client", func() {
...
+				second := deps.Items[1]
+				Expect(second.Name).To(Equal(dep4.Name))
+
+				deleteDeployment(dep1, ns)
+				deleteDeployment(dep2, ns)
+				deleteDeployment(dep3, ns)
+				deleteDeployment(dep4, ns)
+
+				close(done)
```
DirectXMan12: no need for a done channel unless you're doing async tests (which this isn't)

DirectXMan12:
if you really want to try and implement limit and continue in the fake client, that'd be ok, but I think since not everything supports limit and continue, we can just leave the fake client as-is.

pkg/client/client_test.go
rajathagasthya: I noticed most of the tests in the file do this and I just followed. Should they be fixed too maybe in a separate issue?

pkg/client/client_test.go
rajathagasthya: Same as above. Most tests do this.

pkg/client/client_test.go
DirectXMan12: yeah, they need to be fixed in a separate issue

rajathagasthya:
@DirectXMan12 Can you take a look when you get a chance?

pkg/client/interfaces.go
```diff
@@ -290,6 +305,20 @@ func (o *ListOptions) InNamespace(ns string) *ListOptions {
...
+	o.Limit = n
+	return o
+}
+
+// SetContinue is a convenience function that sets the continuation
+// token, and then returns the options. It mutates the list options.
+func (o *ListOptions) SetContinue(token string) *ListOptions {
+	o.Continue = token
+	return o
+}
```
DirectXMan12: we don't need these any more -- this style is deprecated.

rajathagasthya:
Rebased on top of #536

gjkim42:
Does `CacheReader` supports `List` function with `Continue` token?

PR #349, opened 2019-03-05
 ✨ update kubernetes 1.13 to the latest patch release

tariq1890:
It is in the best interests of any k8s API consumer that we always vendor in the latest patch release of a minor version.

As controller-runtime acts as the primary source of k8s-dependencies for cluster-API and its providers, best practices mandate that we update the versions at the root so that they percolate down to the Cluster API providers. If this change is not accepted, we would be forced to use override deps.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -620,7 +620,7 @@
   version = "v2.2.1"
 
 [[projects]]
-  digest = "1:3ca4baec1a14e727951288f091badc5d403096d1c00d3ff5299f273a0c6db85d"
+  digest = "1:684a27f12acc943bee24373b4e8f4acde803990add2fc5dde602c27b58207d3b"
   name = "k8s.io/api"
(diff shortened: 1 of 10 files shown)


tariq1890:
@detiber Can we push this in pretty please :) ?

vincepri:
@tariq1890 The PR title should be prefixed with ✨ given that this is a backward compatible change. 

/assign @DirectXMan12

DirectXMan12:
Can you please add a PR description motivating this change -- it's useful for looking back at the merge history to see why something was particularly relevant.

DirectXMan12:
(even just something like "1.13.4 includes a fix for bug XYZ that affects CR users when they do foo")

tariq1890:
@DirectXMan12 I've mentioned the reason.

tariq1890:
Any updates on this?

PR #353, opened 2019-03-07
:sparkles: Trim Gopkg.toml

DirectXMan12:
This fixes some issues in our Gopkg.toml (ghodss/yaml --> sigs.k8s.io/yaml, normalizing 'v' vs no 'v' in version strings), sets more aggressive pruning, and removes the "required" marks, which should cut down on dependency bloat in CR.


droot:
lgtm'd it, but needs a rebase.

droot:
Since we have moved to Go modules, it is probably safe to close this one now.

PR #355, opened 2019-03-09
:sparkles: [WIP] Structured Errors

DirectXMan12:
This is a sketch of what structured error collection could look like.
With stack tracing enabled, it's slower than current format.Errorf, but
still faster than the Go 2 error prototype.  With stack traces disabled,
it's actually faster than fmt.Errorf, with the same number of
allocations, but slightly more bytes.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -482,6 +482,18 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:cc5c2ca087db25b6dcae0e2e2714582bc900f0b711acc42f8bc10c5e2bc5a159"
+  name = "golang.org/x/exp"
+  packages = [
+    "errors",
+    "errors/fmt",
+    "errors/internal",
+  ]
+  pruneopts = "UT"
+  revision = "a1d7652674e8f269b17d9be90d3c4c1d0e8830e8"
+
+[[projects]]
+  branch = "master"
   digest = "1:ff58bc124488348ba2ca02f62de899d07ba3e7dce51da02557bfc9a8052608db"
   name = "golang.org/x/net"
   packages = [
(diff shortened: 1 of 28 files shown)


DirectXMan12:
Will add some metrics later.  This is currently a sketch/experiment right now, but seems like it could be neat going forward.  We could just use the Go 2 errors if we just want stack traces, but that's mandatory stack traces, which may or may not be desirable.

DirectXMan12:
### Without stack traces

```
goos: linux
goarch: amd64
pkg: sigs.k8s.io/controller-runtime/pkg/erreur
BenchmarkWithErreur-8            3000000               487 ns/op             233 B/op          4 allocs/op
BenchmarkExpError-8               500000              2337 ns/op             315 B/op          6 allocs/op
BenchmarkPlainFmt-8              2000000               785 ns/op             185 B/op          4 allocs/op
PASS
ok      sigs.k8s.io/controller-runtime/pkg/erreur       5.614s
```

### With stack traces (may actually double-capture stack traces due to Zap functionality)

```
goos: linux
goarch: amd64
pkg: sigs.k8s.io/controller-runtime/pkg/erreur
BenchmarkWithErreur-8            1000000              1660 ns/op             233 B/op          4 allocs/op
BenchmarkExpError-8               500000              2453 ns/op             315 B/op          6 allocs/op
BenchmarkPlainFmt-8              2000000               801 ns/op             185 B/op          4 allocs/op
PASS
ok      sigs.k8s.io/controller-runtime/pkg/erreur       5.399s
```

DirectXMan12:
follow-ups: a join context-y thing that lets you produce errors or log with a given set of values.

DirectXMan12:
There's some questions about how the values from the error should integrate with log lines.  The currently-commented method looks a bit ore standard, while the currently uncommented version should be the most efficient with the current zap semantics (potentially avoiding an extra allocation/slice resize on logging, afaict).

DirectXMan12:
closing as an experiment.  We can re-open it later

PR #358, opened 2019-03-12
:running: add tests for simplified webhook interface

mengqiy:
Followup PR per https://github.com/kubernetes-sigs/controller-runtime/pull/328#issuecomment-471138778

### DIFF
--- a/pkg/builder/build_test.go
+++ b/pkg/builder/build_test.go
@@ -18,11 +18,12 @@ package builder
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
 	"strings"
-
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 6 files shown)


pkg/builder/build_test.go
```diff
@@ -121,6 +120,93 @@ var _ = Describe("application", func() {
...
+				Owns(&appsv1.ReplicaSet{}).
+				Build(noop)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(instance).NotTo(BeNil())
+			svr := m.GetWebhookServer()
+			Expect(svr).NotTo(BeNil())
+
+			By("trying to register an existing mutating webhook path")
+			path := generateMutatePath(schema.GroupVersionKind{Group: "foo.test.org", Version: "v1", Kind: "TestDefaulter"})
+			Ω(func() { svr.Register(path, nil) }).Should(Panic())
```
DirectXMan12: this is a bit of a weak test -- what if we add validation that handler is not nil, or something?  Maybe we should check that it responds to a valid admission request?

pkg/builder/builder_suite_test.go
```diff
@@ -42,7 +44,88 @@ var cfg *rest.Config
 var _ = BeforeSuite(func(done Done) {
 	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
 
-	testenv = &envtest.Environment{}
+	testenv = &envtest.Environment{
+		CRDs: []*apiextensionsv1beta1.CustomResourceDefinition{
+			{
+				TypeMeta: metav1.TypeMeta{
```
DirectXMan12: refactor this out into a method that takes a kind name

pkg/builder/build_test.go
mengqiy: Done.
PTAL

DirectXMan12:
(as discussed) inject isn't quite it the right spot now (you can register hooks before adding to the manager)

DirectXMan12:
/lgtm
/approve

We'll need to resolve the dep management issue at some point, but this is fine for now.

PR #365, opened 2019-03-19
:sparkles: Default QPS to a saner value in and out of tests

DirectXMan12:
The default QPS on REST clients is 5 (with burst to 10), which is rather slow.  This is especially true when running a test server, since you want to run tests pretty quickly, and the test cluster is default.

For normal clients, we probably want to default to what the KCM defaults to (our situation isn't quite the same, since we actually share the client across controllers, so we should follow up dealing with that at some point).  This gives us 20 QPS with burst to 30.


DirectXMan12:
cc @droot @mengqiy 

We might actually want to cherry-pick this to stable as well.

mengqiy:
It seems `goimports` is not happy, please fix.

PR #366, opened 2019-03-20
[WIP] :warning: Server Side Apply Support

DirectXMan12:
This introduces a server-side apply patch type.  Don't use yet, since it requires Kubernetes 1.14.  Breaking since the structure of PatchOptions needs to change.


DirectXMan12:
Also TODO: switch to specificallyversionedoptions calls

DirectXMan12:
One step in #347

DirectXMan12:
also TODO (eventually): jsonpatch patch type

DirectXMan12:
(this also needs tests before merge)

pkg/client/patch.go
```diff
@@ -72,3 +77,21 @@ func (s *mergeFromPatch) Data(obj runtime.Object) ([]byte, error) {
 func MergeFrom(obj runtime.Object) Patch {
 	return &mergeFromPatch{obj}
 }
+
+// applyPatch uses server-side apply to patch the object.
+type applyPatch struct{}
```
apelisse: You don't use the `var _ SomeInterface = someImpl{}` pattern? I'm assuming this implements something that I don't know

DirectXMan12: See the `Patch` interface above

apelisse:
LGTM from my perspective.

apelisse:
Shame on me for forgetting the fieldmanager field ...

DirectXMan12:
for posterity: part of this went in to #435, the rest needs a bit more work.

DirectXMan12:
we've got minimal support currently, but it's a bit clunky.  We'll add a nice wrapper eventually, but closing for now

Issue #350, opened 2019-03-05
Manager does not wait for Runnables to stop


grantr:
Currently `Manager.Start` immediately returns when stop channel is closed without waiting for Runnables to stop:

https://github.com/kubernetes-sigs/controller-runtime/blob/6649bdbe3f1f5f9769b91c031e52be253c87e922/pkg/manager/internal.go#L216-L219

This makes writing a Runnable with blocking stop logic difficult or impossible. For example, the Manager's internal `serveMetrics` is basically a Runnable (a blocking method that takes a stop channel). When the stop channel is closed, `serveMetrics` tries to shut down its http server:

https://github.com/kubernetes-sigs/controller-runtime/blob/6649bdbe3f1f5f9769b91c031e52be253c87e922/pkg/manager/internal.go#L188-L192

But [in normal usage, the process immediately exits when Manager.Start returns](https://github.com/kubernetes-sigs/kubebuilder/blob/4228ca5e22d680aff3ec499e6451f51b2405f515/test/project/cmd/manager/main.go#L80-L83) so the `Shutdown` call is unlikely to complete.

Adding `sync.WaitGroup` accounting [inside the Runnable wrapper goroutines](https://github.com/kubernetes-sigs/controller-runtime/blob/6649bdbe3f1f5f9769b91c031e52be253c87e922/pkg/manager/internal.go#L249-L251) would allow the addition of a `Wait()` method to the `Manager` interface that blocks until all Runnables have returned (or `WaitWithTimeout(time.Duration)`, `Shutdown(context.Context)` for a termination timeout), without changing either the Runnable contract or the Manager contract:

```go
if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
    log.Error(err, "unable to run the manager")
    os.Exit(1)
}
mgr.Wait()
```

DirectXMan12:
/kind bug
/priority important-soon

good catch

rajathagasthya:
I'll take this!

DirectXMan12:
awesome!

grantr:
While investigating this I discovered another issue: the Manager's `errChan` is unbuffered, so only one Runnable goroutine (including the metrics server) will be able to terminate correctly. Others will block indefinitely. An easy fix for this is making the errChan buffered, but that's not a great long-term solution.

Anyone working on solutions in this area might want to keep #330 in mind too.

rajathagasthya:
/remove-lifecycle stale

[...]

Issue #305, opened 2019-01-24
Consider using OpenCensus for metrics instead of Prometheus


grantr:
Using the Prometheus library to collect metrics works fine mostly, but has some limitations: #258 wants to change the way metrics are aggregated, and #297 wants to add additional handlers to the manager's HTTP endpoint.

Maybe this is a far-out idea, but I wonder if switching to [OpenCensus](https://opencensus.io) for measurement instead of Prometheus client at this early stage would be a good idea. Tl;dr: OpenCensus is a collection of libraries in multiple languages that facilitates the measurement and aggregation of metrics in-process and is agnostic to the export format used. _It doesn't replace Prometheus the service, it just replaces Prometheus the Go library._ OpenCensus can export to Prometheus servers, so this is strictly an in-process change.

The [OpenCensus Go library](https://github.com/census-instrumentation/opencensus-go) is similar to the Prometheus client, but **separates the collection of metrics from their aggregation and export**. This theoretically allows libraries to be instrumented without dictating how users will aggregate metrics (solving #258) and export metrics (solving #297), though default solutions can be provided for both (likely the same as today's default bucketing and Prometheus HTTP exporter).

Here's an example from knative/pkg of defining measures and views (aggregations): https://github.com/knative/pkg/blob/53b1235c2a85e1309825bc467b3bd54243c879e6/controller/stats_reporter.go. The view is defined separate from the measure, giving the library user the ability to define their own views with library-defined metrics.

And an example of exporting metrics to either stackdriver or prometheus: https://github.com/knative/pkg/blob/225d11cc1a40c0549701fb037d0eba48ee87dfe4/metrics/exporter.go. The user of the library can export views in whatever format they wish, independent of the measures and views that are defined.

It additionally has support for exporting traces, which IMO would be a useful debugging tool and a good use for the context arguments in the client interface (mentioned in #265). Threading the trace id into that context would give the controller author a nice overview of the entire reconcile, with spans for each request, cached or not.

DirectXMan12:
/kind feature

I do kind-of like this idea.  Want to put together a PoC?

grantr:
Working on this now!
/assign

negz:
I was just exploring controller-runtime's observability stance, noticed the kubebuilder book mentioned Prometheus metrics, thought "oh man I wish they used Opencensus instead", came here to thumbs-up this issue, then found past me already did that. ;)

[...]

PR #367, opened 2019-03-21
:sparkles: pkg/manager,metrics: Expose ServingMetrics func

### LINKED ISSUE #336: Serve metrics independently of `manager.Start`

LiliC:
Exposes `ServingMetrics()` function and moves it to the metrics package, so serving metrics can be done independently of calling `.Start()`.

Implemented as per the suggestion from [this comment](https://github.com/kubernetes-sigs/controller-runtime/pull/337#issuecomment-469771107 ).

Closes #336

cc @DirectXMan12

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -20,11 +20,9 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/http"
 	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


LiliC:
Noticed once the unit tests failed, but reran the job and they passed. Not sure if this is a documented flake? Don't think it's introduced with this PR? It was the `source_integration_test.go` test.

grantr:
_From https://github.com/kubernetes-sigs/controller-runtime/pull/337#issuecomment-469771107:_
> Have a helper in pkg/metrics called ServeMetrics that takes in a net.Listener and does what .serveMetrics

I propose a slightly different interface: Create a `PrometheusMetricsServer` struct and have it implement `manager.Runnable`. Then the Manager doesn't need any special metrics server handling; the main method can just call `manager.Add(metricsServer)`.

We've been experimenting with this in Knative [here](https://github.com/knative/eventing/blob/b6bd8f4b26d06fb7de692698301cf21fdb6ee25f/cmd/broker/ingress/main.go#L214-L267) where we have two Runnable http servers (one for metrics) managed by the Manager. (note the use of a waitgroup is only required because of #350).

It's also relevant for #305, where we may have multiple metrics exporters running in the process.

LiliC:
> I propose a slightly different interface: 

That works for our use case very well for our use cases as well. Would be happy to implement that if everyone agrees on it. 👍

DirectXMan12:
That seems pretty reasonable.  I'd still like to have "metrics by default" where you can explicitly disable them, but making it a runnable seems fine.

LiliC:
>  That seems pretty reasonable. 

Okay, will close this PR and we can discuss the details in the https://github.com/kubernetes-sigs/controller-runtime/issues/336

Issue #370, opened 2019-03-22
specificInformersMap Get may accidentally return an unynced informer


nvachhar:
If `specificInformersMap` is already started, a call to `Get` that doesn't find a particular GVK will create the corresponding Informer, add it to the map, and then wait for it to sync. Therefore the caller will get a synced Informer. However, if there is a concurrent caller to Get for the same GVK, it will find the Informer in the map and return it without waiting for it to sync. Therefore the concurrent caller will get an unsynced informer.

nvachhar:
I forgot to mention, I have a patch to fix the problem. Just posting the issue to ensure others agree with the problem's existence.

DirectXMan12:
go ahead and submit the PR to fix it, please

/kind bug

PR #371, opened 2019-03-25
:bug: Ensure returned informers are always correctly synced

nvachhar:
`specificInformersMap.Get` may accidentally return an unsynced `Informer` (contained in the returned `MapEntry`). The condition occurs when 2 callers concurrently call `Get` for the same GVK after the `specificInformersMap` has been started. The first won't find the GVK in the cache. It will add it and wait for the informer to sync. The second will find it in the cache and erroneously return it before it syncs.This patch guarantees that all returned informers are synced by checking before return.


nvachhar:
Fixes #370

nvachhar:
I'm working on the contributor agreement. Stay tuned. We had previously signed with Google before things moved over to CNCF, so I just have to get a new signature.

nvachhar:
Changed my github primary email. Maybe that'll help...

nvachhar:
A quick question for future contributions. What is pseudo-UB? I racked my brain trying to figure it out, but came up short...

Issue #354, opened 2019-03-08
NewFakeClientWithScheme dies silently


michaelgugino:
Current master commit: https://github.com/kubernetes-sigs/controller-runtime/blob/45d652077cba5ccc29621e8db615f25084c701cb/pkg/client/fake/client.go#L65

This line does not result in actual logging in my case, just results (confusingly) in a call to os.Exit(1).

Simple test case (replace clusterautoscaler with a suitable object that is not part of default scheme):

```go
package clusterautoscaler

import (
	"testing"
	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
	"k8s.io/apimachinery/pkg/runtime"
)

func TestAvailableAndUpdatedCheck2(t *testing.T) {
	tscheme := runtime.NewScheme()
	// With the following commented, we fail silently.
	//apis.AddToScheme(tscheme)
	ca := NewClusterAutoscaler()
	_ = fakeclient.NewFakeClientWithScheme(tscheme, ca)
}
```

DirectXMan12:
a) that should probably be a panic, not a log-then-exit.  I suspect it was an incorrectly replaced pattern from our purge of fatalf (which doesn't existing in logr).

b) you should generally set up logging in `TestMain`, `BeforeSuite`, each test, or what have you (by calling `"pkg/log".SetLogger`).

anthonyho007:
@DirectXMan12  Can i work on this issue ?

DirectXMan12:
go ahead. Can't officially assign unless you're an org member, but consider it assigned :-)

/assign @anthonyho007

DirectXMan12:
close as per above PRs

PR #373, opened 2019-03-27
:bug: Fake client should fail with panic instead of log and exit

anthonyho007:
Fake client should panic instead of log-then-exit when it fails to create.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/354

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


grantr:
See also https://github.com/kubernetes-sigs/controller-runtime/pull/232

anthonyho007:
@grantr I totally didnt see ur PR. I will close this PR then. It would be nice if you can include the issue in ur PR.

grantr:
@anthonyho007 no problem, I should have created an issue instead of just a PR describing it! I updated #232 to link to #354.

PR #376, opened 2019-03-27
⚠ Bump Kubernetes version to 1.14

adracus:
Bump versions of dependencies to Kubernetes 1.14.
@DirectXMan12 I guess you could base #366  on this to both speed up your development + keep the code review volume for your PR small.


DirectXMan12:
woo hoo!

I'll do a double check of this a bit later, but thanks for putting this together.

justaugustus:
Could we bump these to 1.14.1 instead?

vincepri:
@adracus Are you still working on this update? Otherwise I'm happy to pick this up

adracus:
Sorry for the delay, Easter and GoT was happening :) Bumped the version to 1.14.1 as requested.

justaugustus:
Thanks Axel!

rfranzke:
Can we expect this to be merged anytime soon?

mengqiy:
We will merge https://github.com/kubernetes-sigs/controller-runtime/pull/414 probably sometime this week. And then bump to 1.14 (this PR).

tamalsaha:
@adracus, I wonder if you could rebase and fix the conflicts? We would like to use 1.14.x, too! :)

adracus:
@tamalsaha Done! However I wonder why does controller runtime use both `dep` and go modules... Hope this specifies the right dependencies.

DirectXMan12:
We're keeping around both since dep doesn't understand go modules yet and we don't want to break people.  The plan is to drop dep support when Go modules goes stable in Go 1.13

PR #294, opened 2019-01-15
:warning: Convert client.List to use functional options

droot:
Replaces:

`List(ctx, ListOptions, list)`

with:

`List(ctx, list, ...Option)`

This matches the functional options signature of `client.Delete` in #94.

A few examples (I tried formatting these as a table but it wasn't wide enough):

Select all:
```go
// current
List(ctx, nil, list)
// proposed
List(ctx, list)
```

Filter by namespace:
```go
// current
List(ctx, client.InNamespace(ns), list)
// proposed
List(ctx, list, client.InNamespace(ns))
```

Filter by namespace and labels:
```go
// current
List(ctx, client.InNamespace(ns).MatchingLabels(labels), list)
// proposed
List(ctx, list, client.InNamespace(ns), client.MatchingLabels(labels))
```

To assemble a `ListOptions` with the builder methods or some other strategy, use the `UseListOptions` functional option:

```go
lo := &client.ListOptions{}
client.List(ctx, list, client.UseListOptions(
  lo.InNamespace(ns).MatchingLabels(labels)
))
```

Note: This PR was contributed by @grantr but we couldn't get it merged because of the breaking changes, but now we are ready for a major revision `0.2.0`, so I manually ported https://github.com/kubernetes-sigs/controller-runtime/pull/106

PR #380, opened 2019-04-01
🐛Fix used scheme for fake client in List method

mszostok:
**Description**

Changes proposed in this pull request:

- Really simple fix for used scheme in fake client in `List` method. 

**Background**
 
This bug https://github.com/kubernetes-sigs/controller-runtime/issues/137 was already fixed by this PR: https://github.com/kubernetes-sigs/controller-runtime/pull/213. Unfortunately in this [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/294/files#diff-62fbcb10a0dc0bc5b0a1bba07b8eefb0R92) the bug was reintroduced - probably by mistake.

**NOTE:** 
I also tried to change the tests to never regress but such action will complicate the test suite, so IMO is really not worth to do that. But if you think that tests coverage should be added then let's decide which `types` should be used and how the test suite should look like, e.g. full tests for `NewFakeClientWithScheme` and only one test for `NewFakeClient` ensuring that it's injecting the `"k8s.io/apimachinery/pkg/runtime/schema.Scheme"`? If the decision will be made then I can implement that too :) 

**Related issue**
- https://github.com/kubernetes-sigs/controller-runtime/issues/137

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -93,7 +93,7 @@ func (c *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.
 }
 
 func (c *fakeClient) List(ctx context.Context, obj runtime.Object, opts ...client.ListOptionFunc) error {
-	gvk, err := apiutil.GVKForObject(obj, scheme.Scheme)
+	gvk, err := apiutil.GVKForObject(obj, c.scheme)
 	if err != nil {
 		return err
 	}


grantr:
> Unfortunately in this [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/294/files#diff-62fbcb10a0dc0bc5b0a1bba07b8eefb0R92) the bug was reintroduced - probably by mistake.

Good catch! This may have been accidentally introduced if the code from #294 was copied from #106 (which was written before #213).
/cc @droot

PR #381, opened 2019-04-01
:sparkles: [WIP] A fake client using client-go reactors

grantr:
This is an experiment in writing a fake client using the same reactor mechanism as the generated fake clients.

The `FakeReactorClient` does all the same things the `FakeClient` does, but it also adds all the `testing.Fake` reactor functions that make it possible to mock the response to do something abnormal for testing error conditions:

```go
cl.PrependReactor("get", "*", func(action testing.Action) (bool, runtime.Object, error) {
    return true, nil, fmt.Errorf("mocked %s: %s", action.GetVerb(), action.GetResource())
})
namespacedName := types.NamespacedName{
    Name:      "test-deployment",
    Namespace: "ns1",
}
obj := &appsv1.Deployment{}
err := cl.Get(nil, namespacedName, obj)
Expect(err).To(MatchError("mocked get: apps/v1, Resource=deployments"))
```

This is something we've missed in Knative. We built our own MockClient [here](https://github.com/knative/eventing/blob/4fcaf7e2568ae24bb3d77ce0d775c70cb17413aa/pkg/reconciler/testing/mock_client.go) but it might be easier if one was provided using the familiar reactor mechanisms.

WIP because I'm not sure if this is the right interface.

### DIFF
--- a/pkg/client/fake/reactor_client.go
+++ b/pkg/client/fake/reactor_client.go
@@ -1,0 +1,276 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fake
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/testing"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
+)
+
+var (
+//log = logf.RuntimeLog.WithName("fake-client")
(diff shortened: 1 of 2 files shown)


grantr:
Another useful feature is the ability to get a list of all actions previously invoked. This can be used to determine how many times an action was called and with what arguments:

```go
namespacedName := types.NamespacedName{
    Name:      "test-deployment",
    Namespace: "ns1",
}
obj := &appsv1.Deployment{}
err := cl.Get(nil, namespacedName, obj)

Expect(cl.Actions()).To(HaveLen(1))
action := cl.Actions()[0]

Expect(action.GetNamespace()).To(Equal("ns1"))
Expect(action.GetVerb()).To(Equal("get"))
```

DirectXMan12:
We've been historically hesitant to introduce this for a few reasons (not necessarily saying no, just laying out why we don't have this):

1. I've never seen a concise reactor test -- they almost always involve huge numbers of lines of test setup
2. It's was fairly common in my experience in k/k to encounter reactors that didn't *quite* simulate the API server, and thus broke in weird ways occasionally
3. You almost always end up testing that certain things happen, and not that things end up the right way -- testing the actions is almost always going to lead to brittle tests, IMO (it's occasionally what you want, but generally not).
4. client-go reactors aren't always obvious when the break, or how they're supposed to work (e.g. label selectors get auto-applied to output, input resource arguments can occasionally be strange, fall-through doesn't always work, the return values aren't particularly clear, etc).  The interface needs a bit of work, IMO.

grantr:
I'm definitely open to other interfaces, and I agree reactors are not at all obvious (I struggled to understand when I first encountered them).

The need I'm attempting to meet here is the ability to generate mocked requests. Some code is only evaluated in situations that rarely happen, and it's very difficult to get test coverage without mocks.

The list of actions taken is secondary but useful - ensure the object was updated only once during the reconcile to avoid ResourceVersion conflicts, for example. Yes, you don't need this with an apiserver, but that doesn't apply when mocking is involved.

Reactors seem like the most convenient way of adding mocking to the client with the smallest number of code and behavior changes and the largest number of *contributors* already familiar with the interface.  I originally experimented with https://github.com/vektra/mockery instead, but abandoned that effort (can't remember why). Perhaps it's a more natural interface for users.

If the mocking needs can be met by a different solution without reactors, that would be fine with me. :)

PR #382, opened 2019-04-03
:sparkles: support custom negotiated serializer :bug:

terinjokes:
The Kubernetes client allows for users to configure a serializer which
is used to encode and decode resources exchanged with the Kubernetes API
server.

This changeset modifies the controller-runtime client to only set the
default serializer if the NegotiatedSerializer option is nil.

Signed-off-by: Terin Stock <terin@cloudflare.com>

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -66,10 +66,13 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 }
 
 // RESTClientForGVK constructs a new rest.Interface capable of accessing the resource associated
-// with the given GroupVersionKind.
+// with the given GroupVersionKind. The REST client will be configured to use the negotiated serializer from
+// baseConfig, if set, otherwise a default serializer will be set.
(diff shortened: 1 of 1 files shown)


terinjokes:
I assume, since I cut from broken master, the CI test are failing for the same reason here.

DirectXMan12:
It's a flake.  I think we may need to bump the version of the kubernetes binaries used in our tests.

DirectXMan12:
Seems fairly reasonable, but the godocs need an update to mention this.
Out of curiosity, what's the usecase here?  Proto support?

terinjokes:
@DirectXMan12 I'm happy to update the GoDocs. In which package did you have in mind?

I call `scheme.AddTypeDefaultingFunc` which I noticed weren't being run after converting my project to using controller-runtime.

DirectXMan12:
The GoDocs on RESTClientForGVK should mention this.

DirectXMan12:
> which I noticed wasn't being run after converting my project to using controller-runtime.

ah, ok.

terinjokes:
@DirectXMan12 I added a note, I'm happy to make changes if you have a better way to word it.

PR #383, opened 2019-04-03
:warning: source: make the Informer source compatible with cache.Informer

ironcladlou:
Before fc804a41 (#267), the `cache.Informers` interface methods returned
`k8s.io/client-go/tools/cache.SharedIndexInformer`s which were by default
compatible with the built-in `source.Informer`. Users could create arbitrary
`cache.Cache` instances (or get them from the `Manager`) and then use
`controller.Watch` to drive a controller with a `source.Informer` from the
`cache.Cache`.

With fc804a41, the `cache.Informers` interface was changed to return
`cache.Informer` instances; however, `source.Informer` was not updated to accept
a `cache.Informer`, and so users can no longer use the built-in
`source.Informer` with `cache.Cache`.

The `cache.Informer` interface appears to satisfy the needs of
`source.Informer`. This commit broadens `source.Informer` to accept a
`cache.Informer`, restoring the prior capability while remaining compatible
with `SharedIndexInformer` use.

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -29,7 +29,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/source/internal"
 
-	toolscache "k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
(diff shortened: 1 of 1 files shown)


ironcladlou:
Guess there's no build bot — not sure what's up with the test failure. Second run passed. Flake?

DirectXMan12:
this is technically a breaking change (so I've fixed that) otherwise

/lgtm
/approve

DirectXMan12:
(we've been getting intermittent flakes.  based on a conversion I had the other day, it might actually be a now-fixed bug in kube-apiserver).

ironcladlou:
Thank you!

PR #385, opened 2019-04-04
WIP: Bump dependencies to Kubernetes master ✨⚠

munnerz:
This PR probably shouldn't be merged, as it pins controller-runtime to `master` as of today.

I put this patch together so that I can test out go modules from my project that depends on controller-runtime.

Opened for visibility, the only patch needed here was to the leaderelection package (as you can see in the 3rd commit).

On the topic of go modules - is there any ongoing discussion of kubebuilder and/or controller-runtime switching to it in future? I'd be interested in seeing where I could get involved here 😄


DirectXMan12:
We need to make sure that dependent kubebuilder projects don't break or anything, but we're looking to move to modules, keeping around dep support as well till we get to Go 1.13 when modules are on by default.  We just moved controller-tools over.

draveness:
Hi, I updated the other kubernetes related pkg to v1.14. But controller-runtime seems out of date, when will this PR get merged? :)

mengqiy:
CR is on k8s 1.14 and using go modules.
Closing this.

Issue #390, opened 2019-04-10
Webhook server should reload certs on change


vincepri:
Opening this issue to track [this TODO](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/server.go#L135). Currently the certificates are only loaded on server creation, there should be some logic to watch the directory and update the tls configuration appropriately.

bison:
Would love to get some feedback on #421 if anyone has some time. Seems to work for me.

PR #395, opened 2019-04-12
:sparkles, minor [WIP] Autoload on tls changes

anthonyho007:
To be update description

Fixes:
https://github.com/kubernetes-sigs/controller-runtime/issues/390


pkg/webhook/tls_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	defer watcher.Close()
+
+	done := make(chan error)
+	go func() {
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Op&fsnotify.Write == fsnotify.Write {
```
mengqiy: I doubt if it will work, since when a secret get updated it actually triggers a rename event. Ref: [algorithm of volume writer](https://github.com/kubernetes/kubernetes/blob/b5c34d0c436dc6b95f9b78769d090af1bbbfa629/pkg/volume/util/atomic_writer.go#L89-L119).

pkg/webhook/tls_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+	w.certificate = &certKey
+	return nil
+}
+
+func (w *wrappedTLS) autoloadTLS(cert, key string) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+	defer watcher.Close()
```
mengqiy: function `autoloadTLS` is not waiting for the child goroutine you created in line 66. 
So when it finishes the execution and the `watcher` is closed. The child goroutine is not doing anything useful anymore.

pkg/webhook/tls_wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/go-logr/logr"
+)
+
+type wrappedTLS struct {
+	sync.Mutex
+	certificate *tls.Certificate
+	log         logr.Logger
+}
+
+func (w *wrappedTLS) setLogger(logger logr.Logger) error {
```
mengqiy: It seems this functions is not used anywhere.

mengqiy:
Sorry for the late review.

mengqiy:
Closing in favor of #421

PR #396, opened 2019-04-12
:bug: Reset Failure count for reconcile loops using RequestAfter

### LINKED ISSUE #392: Reconcile loops using RequeueAfter never reset their failure count
Controllers calling reconcile loops that stabilise by calling `return result.Result{RequeueAfter: resyncPeriod}, nil` to periodically re-run reconciliation never call `c.Queue.Forget` to clear the workqueue failure count for a particular request.

As the Failure count is never cleared this count builds up over the lifetime of the CR and quickly results in the back-off timeout reaching its maximum limit for any error even if the reconcile loop is mostly stable. This is not ideal as the maximum limit for the timeout is ~16mins which slows down the reconciliation loop considerably.

We've experimented with using the resyncPeriod set on the manager as a workaround however from other tickets this doesn't seem to be recommended and does have side effects such as causing conflicts when calling client.Update as the object we were operating on is now out of date.

alaypatel07:
This PR fixes #392. Reset the failure count if `return result.Result{RequeueAfter: resyncPeriod}, nil` is returned by the reconcile loop.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -214,6 +214,11 @@ func (c *Controller) processNextWorkItem() bool {
 		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, "error").Inc()
 		return false
 	} else if result.RequeueAfter > 0 {
+		// The result.RequeueAfter request will be lost, if it is returned
+		// along with a non-nil error. But this is intended as
+		// We need to drive to stable reconcile loops before queuing due
+		// to result.RequestAfter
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Hey, I'm sorry about the huge delay reviewing this.  Slipped off my radar due to a number of factors.

DirectXMan12:
/lgtm
/approve

Thanks so much!

alaypatel07:
@DirectXMan12 Thanks for the review :)

Issue #369, opened 2019-03-22
Envtest: support loading multiple CRDs from a single file


nrfox:
Currently envtest supports [installing CRDs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest#CRDInstallOptions) from yaml at specified paths however it assumes that there is only one CRD per file. Given a single file with multiple CRD documents separated by `---`, envtest will read/install the first document but not the rest. This is confusing while testing because the scheme has all of the correct GVKs but tests will fail when [registering the watch](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/controller/controller.go#L175-L181) for the second CRD in the file with the error `no matches for kind "MyKind" in version "mygroup/version"`. When running outside of tests, there are no errors because all the CRDs get installed properly using `kubectl apply`.

I'm not sure the current library used for parsing yaml supports reading multiple documents from a single file however I believe the apimachinery [decoder](https://godoc.org/k8s.io/apimachinery/pkg/util/yaml#YAMLOrJSONDecoder) does.

I discovered this issue because I have an external CRD as a dependency of my project which exposes its definitions as a single yaml file.

DirectXMan12:
/kind bug

Yeah, we should tackle this.  It's not hard -- it just requires a tweak to the CRD loading code

/help-wanted

delp:
Hello,

I am a first time contributor and I wanted to try taking on this issue. I have been working with the operator-sdk, and so have some familiarity with CRDs, watches, and the controller. But I can't figure out how to build this repository, and didn't find what I was looking for in any of the docs for FAQ. Can someone help me get started?

GrigoriyMikhalkin:
Hi, i would like to take this issue, seems like no one is working on it

PR #397, opened 2019-04-13
✨ envtest can load multiple CRDs from single file

GrigoriyMikhalkin:
Updated`envtests` to support loading multiple CRDs from single file

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/369

### DIFF
--- a/pkg/envtest/multiplecrds.yaml
+++ b/pkg/envtest/multiplecrds.yaml
@@ -1,0 +1,25 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: captains.crew.example.com
+spec:
+  group: crew.example.com
+  names:
+    kind: Captain
+    plural: captains
+  scope: Namespaced
+  version: "v1beta1"
+---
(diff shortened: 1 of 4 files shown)


GrigoriyMikhalkin:
I signed it

pkg/envtest/multiplecrds.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: firsts.first.example.com
```
mengqiy: `firsts.first.example.com` and `seconds.second.example.com` looks very weird.
I'd suggest `captains.crew.example.com` and `firstmates.crew.example.com`. These are something we use a lot in kubebuilder testing.

mengqiy:
@GrigoriyMikhalkin sorry for the late review

pkg/envtest/multiplecrds.yaml
GrigoriyMikhalkin: Updated PR

PR #400, opened 2019-04-18
:sparkles: use more reasonable status code for ValidationResponse

mengqiy:
In ValidationResponse, use 200 if allowed, otherwise use 403 (Forbidden).

### DIFF
--- a/pkg/webhook/admission/response.go
+++ b/pkg/webhook/admission/response.go
@@ -62,15 +62,20 @@ func Errored(code int32, err error) Response {
 
 // ValidationResponse returns a response for admitting a request.
 func ValidationResponse(allowed bool, reason string) Response {
+	code := http.StatusForbidden
+	if allowed {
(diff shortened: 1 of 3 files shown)


pkg/webhook/admission/response.go
```diff
@@ -62,15 +62,22 @@ func Errored(code int32, err error) Response {
 
 // ValidationResponse returns a response for admitting a request.
 func ValidationResponse(allowed bool, reason string) Response {
+	var code int32
+	if allowed {
+		code = http.StatusOK
+	} else {
+		code = http.StatusForbidden
+	}
```
droot: We can simplify this (probably):

```Go
code := http.StatusForbidden
if allowed {
   code = http.StatusOK
}
```

mengqiy: Done

mengqiy:
@droot PTAL

PR #405, opened 2019-04-24
Add support for passing namespace using manager cmdline flag

vikaschoudhary16:


vikaschoudhary16:
This seems not a good idea to add changes here for adding a new flag at controller manager. closing this.

PR #409, opened 2019-04-25
Add back http2 support :bug:

### LINKED ISSUE #408: HTTP2 broken in webhook server in v0.2.0-alpha.0
The rework of the certificate generation seems to have removed support for HTTP2 as a side effect.

We now call `http.Server.Serve` instead of the old `ListenAndServeTLS`:
https://github.com/kubernetes-sigs/controller-runtime/blob/856708d78b479900b9d17f98f7f59a2f074fbc35/pkg/webhook/server.go#L166

`Serve` does not default to http2. It must be configured manually in the listener: 
```
// HTTP/2 support is only enabled if the Listener returns *tls.Conn
// connections and they were configured with "h2" in the TLS
// Config.NextProtos.
//
```

It should be a small thing to fix.
I am going to submit a PR if can just get though signing the CLA, unless someone tells me this works as designed, in which case it should probably be mentioned somewhere in the release notes.

lmunro-at-shopify:
Add back support for serving HTTP2 from the webhook server.
Fixes #408

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -140,6 +140,7 @@ func (s *Server) Start(stop <-chan struct{}) error {
 
 	cfg := &tls.Config{
 		Certificates: []tls.Certificate{cert},
+		NextProtos:   []string{"h2"},
 	}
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
/lgtm
/approve

Thanks, good catch!

PR #412, opened 2019-04-29
Make leader election lease duration configurable through manager options

vikaschoudhary16:

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/leaderelection/leader_election.go
```diff
@@ -43,6 +43,16 @@ type Options struct {
 	// LeaderElectionID determines the name of the configmap that leader election
 	// will use for holding the leader lock.
 	LeaderElectionID string
+
+	// LeaseDurationSeconds is the duration that non-leader candidates will
+	// wait to force acquire leadership.
+	LeaseDurationSeconds int
```
sttts: any reason this is not `time.Duration`?

pkg/manager/manager.go
```diff
@@ -113,6 +113,17 @@ type Options struct {
 	// will use for holding the leader lock.
 	LeaderElectionID string
 
+	// LeaseDuration is the duration that non-leader candidates will
+	// wait to force acquire leadership. This is measured against time of
+	// last observed ack.
+	LeaseDurationSeconds *time.Duration
```
sttts: why are these pointers?

vikaschoudhary16: Trying to keep consistent with already existing `syncPeriod`, https://github.com/kubernetes-sigs/controller-runtime/pull/412/files#diff-7d903b129bc883285ffc5068c39c00b0R102

sttts: Please document the defaults.

pkg/manager/manager.go
```diff
@@ -231,22 +242,42 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	stop := make(chan struct{})
 
+	var leaseDuration, renewDeadline, retryPeriod time.Duration
```
sttts: `leastDuration, renewDeadline, retryPeriod := defaultLeastDuration, ...``
Then if clauses without else. Makes it much easier to read.

pkg/manager/internal.go
```diff
@@ -101,6 +108,16 @@ type controllerManager struct {
 	host string
 
 	webhookServer *webhook.Server
+
+	// leaseDurationSeconds is the duration that non-leader candidates will
+	// wait to force acquire leadership.
+	leaseDurationSeconds time.Duration
```
sttts: drop the second suffix, and the `* time.Second` below. Durations have units already.

pkg/manager/internal.go
```diff
@@ -40,6 +40,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+const (
+	// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
+	defaultLeaseDuration = 15
```
sttts: make these durations (`* time.Second`)

sttts:
/assign @droot @DirectXMan12 

For approval.

/lgtm

shawn-hurley:
I think we need to turn off cyclomatic complexity for the `New` function for this to pass CI. 

Personally, I think that makes sense as the new function is still very readable and I don't think it makes sense to make another function just for this.

example_test.go
```diff
@@ -37,7 +37,11 @@ import (
 func Example() {
 	var log = controllers.Log.WithName("builder-examples")
 
-	manager, err := controllers.NewManager(controllers.GetConfigOrDie(), controllers.Options{})
+	manager, err := controllers.NewManager(controllers.GetConfigOrDie(), controllers.Options{
```
shawn-hurley: Can you instead create a new example to show how to change leader election lease duration?

pkg/manager/manager.go
```diff
@@ -173,6 +184,7 @@ func (r RunnableFunc) Start(s <-chan struct{}) error {
 }
 
 // New returns a new Manager for creating Controllers.
+// nolint: gocyclo
```
DirectXMan12: ??

DirectXMan12: please don't just turn off the lints because they're annoying.  This seems to suggest that perhaps leader election setup should be in a separate helper function.

pkg/manager/manager.go
```diff
@@ -231,6 +243,17 @@ func New(config *rest.Config, options Options) (Manager, error) {
...
+	leaseDuration, renewDeadline, retryPeriod := defaultLeaseDuration, defaultRenewDeadline, defaultRetryPeriod
+	if options.LeaseDuration == nil {
+		leaseDuration = defaultLeaseDuration
+	}
+	if options.RenewDeadline == nil {
+		renewDeadline = defaultRenewDeadline
+	}
+	if options.RetryPeriod == nil {
+		retryPeriod = defaultRetryPeriod
+	}
```
DirectXMan12: this should be in setOptionsDefaults, as per everything else that's a default

vikaschoudhary16:
ping @DirectXMan12 @sttts @shawn-hurley .

pkg/manager/manager.go
```diff
@@ -302,6 +316,18 @@ func setOptionsDefaults(options Options) Options {
 	if options.newMetricsListener == nil {
 		options.newMetricsListener = metrics.NewListener
 	}
+	leaseDuration, renewDeadline, retryPeriod := defaultLeaseDuration, defaultRenewDeadline, defaultRetryPeriod
```
droot: We should pull the `options.newResourceLock = leaderelection.NewResourceLock` code fragment above closer to this block so that all leader election bits are in same place.

droot:
>I'm slightly worried about the increase of options in the main controller-manager struct. We might want to consider folding out leader election into its own struct, but I'm not sure we're there yet, and that the ergonomics are worth it.

@DirectXMan12  I agree we are not there yet.

PR #414, opened 2019-05-04
:sparkles: Go modules support

DirectXMan12:
This should be fine, as long as minimal version selection holds and you
don't call `go get -u` on other direct dependencies.


go.mod
```diff
@@ -0,0 +1,61 @@
...
+	golang.org/x/sync v0.0.0-20190423024810-112230192c58 // indirect
+	golang.org/x/sys v0.0.0-20180823144017-11551d06cbcc // indirect
+	golang.org/x/text v0.3.0 // indirect
+	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
+	google.golang.org/appengine v1.1.0 // indirect
+	gopkg.in/fsnotify.v1 v1.4.7 // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
+	gopkg.in/yaml.v2 v2.2.1 // indirect
+	k8s.io/api v0.0.0-20190222213804-5cb15d344471
```
rajathagasthya: Do we want to use `replace` to pin specific versions of k8s deps? k/k uses `replace` to pin _all_ deps, btw.

mengqiy: IIUC k/k uses `replace` to make the `staging` directory work.
Unlike k/k, controller-runtime always uses a published version of k8s deps.

mengqiy:
> you don't call `go get -u` on other direct dependencies.

why this is a problem when using go modules?

hack/common.sh
```diff
@@ -25,6 +25,14 @@ if [ -n "$TRACE" ]; then
   set -x
 fi
 
+(go mod edit -json &>/dev/null)
```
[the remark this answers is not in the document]
rajathagasthya: I think it's just a handy way to check if modules is enabled for the repo, given the next line grabs exit code.

go.mod
rajathagasthya: Using it for `staging` is part of the reason I think. They use it to pin specific versions for all dependencies. Anyway, just wanted to make sure if we want to force specific k8s versions for a given c-r release or just let minimum version selection do its thing.

rajathagasthya:
@mengqiy I think that's because Go mod authors recommend not to use `go get -u` when upgrading direct dependencies. 

From https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies:

> A common mistake is thinking go get -u foo solely gets the latest version of foo. In actuality, the -u in go get -u foo or go get -u foo@latest means to also get the latest versions for all of the direct and indirect dependencies of foo. A common starting point when upgrading foo is instead to do go get foo or go get foo@latest without a -u (and after things are working, consider go get -u=patch foo, go get -u=patch, go get -u foo, or go get -u).

go.mod
tamalsaha: I have been playing with Go mod. It seemed to me that until k8s.io supports go mod, you have to pin using replace to pick up the correct version. So, to set dependency on k8s.io 1.14.x we are using replace.

DirectXMan12: replace only works on the highest-level repository, so when people consume this repo, replace won't help them.

Also, these should be kubernetes-1.14.x -- if you say "k8s.io/api kubernetes-1.14.x", `go mod tidy` will convert it to a version-ish thing like this, but it still should be equivalent afaict (check the date on the commit -- it should match the corresponding version from Gopkg.toml)

hack/common.sh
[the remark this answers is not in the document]
DirectXMan12: yep, exactly!  I'll add a comment.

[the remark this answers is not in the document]
DirectXMan12: (note that we're using the command's return code immediately below :-) )

go.mod
rajathagasthya: Yep, `go mod tidy` will convert to a "pseudo version". Also, your point about downstream consumers of this library makes sense. Thanks for clarifying.

PR #418, opened 2019-05-09
🏃 task: enable running tests in prow

alexeldeib:
Update bash scripts to install gometalinter.v2 and dep if they are missing and SKIP_FETCH_TOOLS is false. This allows installation of the tools in docker containers in Prow.


DirectXMan12:
We should install gometalinter in an image eventually (probably same with tools).  If you put together a file, we'll figure out where it lives.

Thanks so much for this!

/lgtm
/approve

Issue #297, opened 2019-01-16
Support health and liveness probes


j-vizcaino:
I'm adding a liveness probe to a manager built with `controller-runtime`.
The manager already leverages the HTTP service provided natively to serve Prometheus metrics.

I would like to add another endpoint `/live` to provide the liveness of the application, on the same HTTP server that is used for metrics.
There does not seem to be a way to do this today because the http mux is created internally and attached to the HTTP server.

A way to circumvent this is to basically copy/paste the http server logic into a package of my own and add the extra handler but it feels sad that there is no way to reuse the existing code.

DirectXMan12:
we should have a general liveness hook.  I'm not sure we want to make this a general HTTP serving platform, but liveness support would definitely be useful.

/kind feature
/priority important-longterm

GrigoriyMikhalkin:
Hi, i would like to try to implement this.

estroz:
@DirectXMan12 has this been fully implemented in #419 or is there more to do?

DirectXMan12:
I think so

PR #419, opened 2019-05-09
✨ added health probes

GrigoriyMikhalkin:
Attempt to solve https://github.com/kubernetes-sigs/controller-runtime/issues/297.  Adds server, that can serve readiness and liveness probes. Probes return response based on manager's fields `isReady` and `isAlive`, that can be changed via methods `SetReady`/`SetAlive`

### DIFF
--- a/pkg/healthz/healthz.go
+++ b/pkg/healthz/healthz.go
@@ -17,97 +17,175 @@ limitations under the License.
 package healthz
 
 import (
-	"bytes"
 	"fmt"
 	"net/http"
+	"path"
+	"sort"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+// Handler is an http.Handler that aggregates the results of the given
+// checkers to the root path, and supports calling individual checkers on
+// subpaths of the name of the checker.
+//
+// Adding checks on the fly is *not* threadsafe -- use a wrapper.
 type Handler struct {
-	checks []Checker
+	Checks map[string]Checker
 }
 
-// GetChecks returns all health checks.
-func (h *Handler) GetChecks() []Checker {
-	return h.checks
+// checkStatus holds the output of a particular check
+type checkStatus struct {
+	name     string
+	healthy  bool
+	excluded bool
 }
 
-// AddCheck adds new health check to handler.
-func (h *Handler) AddCheck(check Checker) {
-	h.checks = append(h.checks, check)
-}
+func (h *Handler) serveAggregated(resp http.ResponseWriter, req *http.Request) {
+	failed := false
+	excluded := getExcludedChecks(req)
 
-// Build creates http.Handler that serves checks.
-func (h *Handler) Build() http.Handler {
-	return handleRootHealthz(h.checks...)
-}
+	parts := make([]checkStatus, 0, len(h.Checks))
 
-// HealthzChecker is a named healthz checker.
-type Checker interface {
-	Name() string
-	Check(req *http.Request) error
-}
-
-// PingHealthz returns true automatically when checked
-var PingHealthz Checker = ping{}
-
-// ping implements the simplest possible healthz checker.
-type ping struct{}
-
-func (ping) Name() string {
-	return "ping"
-}
-
-// PingHealthz is a health check that returns true.
-func (ping) Check(_ *http.Request) error {
-	return nil
-}
-
-// NamedCheck returns a healthz checker for the given name and function.
-func NamedCheck(name string, check func(r *http.Request) error) Checker {
-	return &healthzCheck{name, check}
-}
-
-// InstallPathHandler registers handlers for health checking on
-// a specific path to mux. *All handlers* for the path must be
-// specified in exactly one call to InstallPathHandler. Calling
-// InstallPathHandler more than once for the same path and mux will
-// result in a panic.
-func InstallPathHandler(mux mux, path string, handler *Handler) {
-	if len(handler.GetChecks()) == 0 {
-		log.V(1).Info("No default health checks specified. Installing the ping handler.")
-		handler.AddCheck(PingHealthz)
+	// calculate the results...
+	for checkName, check := range h.Checks {
+		// no-op the check if we've specified we want to exclude the check
+		if excluded.Has(checkName) {
+			excluded.Delete(checkName)
+			parts = append(parts, checkStatus{name: checkName, healthy: true, excluded: true})
+			continue
+		}
+		if err := check(req); err != nil {
+			log.V(1).Info("healthz check failed", "checker", checkName, "error", err)
+			parts = append(parts, checkStatus{name: checkName, healthy: false})
+			failed = true
+		} else {
+			parts = append(parts, checkStatus{name: checkName, healthy: true})
+		}
 	}
 
-	mux.Handle(path, handler.Build())
-	for _, check := range handler.GetChecks() {
-		log.V(1).Info("installing healthz checker", "checker", check.Name())
-		mux.Handle(fmt.Sprintf("%s/%s", path, check.Name()), adaptCheckToHandler(check.Check))
+	// ...default a check if none is present...
+	if len(h.Checks) == 0 {
+		parts = append(parts, checkStatus{name: "ping", healthy: true})
+	}
+
+	for _, c := range excluded.List() {
+		log.V(1).Info("cannot exclude health check, no matches for it", "checker", c)
+	}
+
+	// ...sort to be consistent...
+	sort.Slice(parts, func(i, j int) bool { return parts[i].name < parts[j].name })
+
+	// ...and write out the result
+	// TODO(directxman12): this should also accept a request for JSON content (via a accept header)
+	_, forceVerbose := req.URL.Query()["verbose"]
+	writeStatusesAsText(resp, parts, excluded, failed, forceVerbose)
+}
+
+// writeStatusAsText writes out the given check statuses in some semi-arbitrary
+// bespoke text format that we copied from Kubernetes.  unknownExcludes lists
+// any checks that the user requested to have excluded, but weren't actually
+// known checks.  writeStatusAsText is always verbose on failure, and can be
+// forced to be verbose on success using the given argument.
+func writeStatusesAsText(resp http.ResponseWriter, parts []checkStatus, unknownExcludes sets.String, failed, forceVerbose bool) {
+	resp.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	resp.Header().Set("X-Content-Type-Options", "nosniff")
+
+	// always write status code first
+	if failed {
+		resp.WriteHeader(http.StatusInternalServerError)
+	} else {
+		resp.WriteHeader(http.StatusOK)
+	}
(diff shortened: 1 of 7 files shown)


pkg/manager/internal.go
```diff
@@ -91,8 +91,19 @@ type controllerManager struct {
...
 	metricsListener net.Listener
 
+	// healthProbeListener is used to serve liveness probe
+	healthProbeListener net.Listener
+
+	// Readiness probe endpoint name
+	readinessEndpointName string
+
+	// Liveness probe endpoint name
+	livenessEndpointName string
```
DirectXMan12: these are pretty standard in kubernetes (e.g. `/healthz` and `/readyz`), so we should default them at the very least

pkg/manager/internal.go
```diff
@@ -252,6 +263,57 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
...
+			} else {
+				w.WriteHeader(503)
+				_, err := w.Write([]byte("error\n"))
+				log.Error(err, "Failed to send readiness probe response")
+			}
+		}
+		mux.Handle(cm.readinessEndpointName, readinessHandler)
+	}
+	if cm.livenessEndpointName != "" {
+		var livenessHandler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
```
DirectXMan12: we should copy the `k8s.io/apiserver` healthz code, and then switch to that whenever it moves to component-base (@luxas ?).  It provides support for pluggable health checks, so we could have multiple in a single manager (e.g. health check per controller).

GrigoriyMikhalkin:
@DirectXMan12 updated code. Now it's using healthz package from `k8s.io/apiserver`.

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,185 @@
...
+// specified in exactly one call to InstallPathHandler. Calling
+// InstallPathHandler more than once for the same path and mux will
+// result in a panic.
+func InstallPathHandler(mux mux, path string, checks ...Checker) {
+	if len(checks) == 0 {
+		log.V(5).Info("No default health checks specified. Installing the ping handler.")
+		checks = []Checker{PingHealthz}
+	}
+
+	log.V(5).Info("Installing healthz checkers:", formatQuoted(checkerNames(checks...)...))
```
DirectXMan12: err... this log line isn't quite right, since it's structured logging.  At an rate, I'd recommend logging once per checker anyway, like:

```go
log.V(1).Info("installing healthz checker", "checker", check.Name
```

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+		}
+		if excluded.Len() > 0 {
+			fmt.Fprintf(&verboseOut, "warn: some health checks cannot be excluded: no matches for %v\n", formatQuoted(excluded.List()...))
+			log.Info(fmt.Sprintf("cannot exclude some health checks, no health checks are installed matching %v",
+				formatQuoted(excluded.List()...)))
+		}
+		// always be verbose on failure
+		if failed {
+			log.V(2).Info(fmt.Sprintf("%vhealthz check failed", verboseOut.String()))
```
DirectXMan12: no `fmt` in our logging.  Log messages should be constant.  Use key-value pairs to attach variable information.

pkg/healthz/healthz_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	contentType = "text/plain; charset=utf-8"
+)
+
+func TestInstallPathHandler(t *testing.T) {
```
DirectXMan12: please adapt these to ginkgo, like below

DirectXMan12: in fact, you can probably prune a few of these, if they overlap heavily with the ginkgo tests below.

pkg/manager/internal.go
mengqiy: Why can't we directly use `k8s.io/apiserver`? Its description is `Library for writing a Kubernetes-style API server.` If it's a lib, why can't we just use it instead of copying it?

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,175 @@
...
+	return &healthzCheck{name, check}
+}
+
+// InstallPathHandler registers handlers for health checking on
+// a specific path to mux. *All handlers* for the path must be
+// specified in exactly one call to InstallPathHandler. Calling
+// InstallPathHandler more than once for the same path and mux will
+// result in a panic.
+func InstallPathHandler(mux mux, path string, checks ...Checker) {
+	if len(checks) == 0 {
```
DirectXMan12: I'm not a huge fan of this logic here -- I'd rather just have something where you call a single call per health check, like `Register` with the webhook server

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,175 @@
...
+			// no-op the check if we've specified we want to exclude the check
+			if excluded.Has(check.Name()) {
+				excluded.Delete(check.Name())
+				fmt.Fprintf(&verboseOut, "[+]%v excluded: ok\n", check.Name())
+				continue
+			}
+			if err := check.Check(r); err != nil {
+				// don't include the error since this endpoint is public.  If someone wants more detail
+				// they should have explicit permission to the detailed checks.
+				log.V(4).Info("healthz check failed", "checker", check.Name(), "error", err)
```
DirectXMan12: We generally don't log at anything above 1 by default in CR.

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,175 @@
...
+				fmt.Fprintf(&verboseOut, "[-]%v failed: reason withheld\n", check.Name())
```
DirectXMan12: shouldn't this be `%q` or `%s`, since name is a string?

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		_, err := verboseOut.WriteTo(w)
+		if err != nil {
+			log.V(2).Info("healthz check failed", "message", verboseOut.String())
+			http.Error(w, fmt.Sprintf("%vhealthz check failed", verboseOut.String()), http.StatusInternalServerError)
+			return
+		}
+		fmt.Fprint(w, "healthz check passed\n")
+	})
```
DirectXMan12: do we not need a WriteHeader here?

pkg/manager/internal.go
```diff
@@ -180,6 +201,32 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	return nil
 }
 
+// AddHealthzChecks allows you to add Healthz checkers
+func (cm *controllerManager) AddHealthzChecks(checks ...healthz.Checker) error {
```
DirectXMan12: I'd probably just make this `AddHealthzCheck` and have it built up the list

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,175 @@
...
+func NamedCheck(name string, check func(r *http.Request) error) Checker {
+	return &healthzCheck{name, check}
+}
+
+// InstallPathHandler registers handlers for health checking on
+// a specific path to mux. *All handlers* for the path must be
+// specified in exactly one call to InstallPathHandler. Calling
+// InstallPathHandler more than once for the same path and mux will
+// result in a panic.
+func InstallPathHandler(mux mux, path string, checks ...Checker) {
```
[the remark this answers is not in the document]
GrigoriyMikhalkin: IIUC, healthz code will be moved to component-base, and we planning to switch to it, right? In that case, isn't it better to leave this function as it is?

pkg/healthz/healthz.go
GrigoriyMikhalkin: It's already set to `http.StatusOK` by default

pkg/healthz/healthz.go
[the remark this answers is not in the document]
DirectXMan12: if we're going to carry an interface in controller-runtime, I'd rather have it up to controller-runtime standards.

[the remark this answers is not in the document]
GrigoriyMikhalkin: Updated PR

DirectXMan12:
Refactored to look closer to what I'd want this to look like.  Thanks for the initial work here!

@mengqiy or @droot can one of you take a look for sanity, please?

droot:
@DirectXMan12 Looks solid to me.

Do we intend to plug in in scaffolded manifests in KB ? (opt-in may be through a kustomize patch ?)

pkg/manager/internal.go
```diff
@@ -188,6 +209,40 @@ func (cm *controllerManager) SetFields(i interface{}) error {
...
+
+	if cm.healthzStarted {
+		return fmt.Errorf("unable to add new checker because healthz endpoint has already been created")
+	}
+
+	if cm.healthzHandler == nil {
+		cm.healthzHandler = &healthz.Handler{Checks: map[string]healthz.Checker{}}
+	}
+
+	cm.healthzHandler.Checks[name] = check
```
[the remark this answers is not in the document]
DirectXMan12: :woman_shrugging: we don't have a way to replace otherwise, so I'm not sure.  Let's leave it like this for the moment, and see what people do with it.

pkg/manager/internal.go
```diff
@@ -188,6 +209,40 @@ func (cm *controllerManager) SetFields(i interface{}) error {
...
+
+	if cm.healthzStarted {
+		return fmt.Errorf("unable to add new checker because readyz endpoint has already been created")
+	}
+
+	if cm.readyzHandler == nil {
+		cm.readyzHandler = &healthz.Handler{Checks: map[string]healthz.Checker{}}
+	}
+
+	cm.readyzHandler.Checks[name] = check
```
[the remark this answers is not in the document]
DirectXMan12: ditto above

DirectXMan12:
@mengqiy ready for another round

pkg/manager/manager.go
```diff
@@ -304,6 +332,19 @@ func defaultNewClient(cache cache.Cache, config *rest.Config, options client.Opt
 	}, nil
 }
 
+// defaultHealthProbeListener creates the default health probes listener bound to the given address
+func defaultHealthProbeListener(addr string) (net.Listener, error) {
+	if addr == "" || addr == "0" {
```
seh: Did we intend for health probes to not be served by default? By my reading, since `Options.HealthProbeBindAddress` has no default value set apart from the empty string, we get no probe server unless we set it explicitly.

omerlh:
Is this documented somewhere? It took me a while to figure how to enable it, and I had to read the code...

GrigoriyMikhalkin:
@omerlh Does `controller-runtime` has documentation at all)? I mean, besides mentions in https://book.kubebuilder.io/

alexeldeib:
Generally, the godoc: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/healthz

https://godoc.org/sigs.k8s.io/controller-runtime is really useful in general

PR #421, opened 2019-05-13
✨ webhook: Handle TLS certificate rotation

bison:
This watches the configured TLS certificate and key files for admission webhooks.  On any change, it attempts to reload them from disk.

Fixes: #390

### DIFF
--- a/pkg/webhook/internal/certwatcher/certwatcher.go
+++ b/pkg/webhook/internal/certwatcher/certwatcher.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package certwatcher
+
+import (
+	"crypto/tls"
+	"sync"
+
+	"gopkg.in/fsnotify.v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
+)
+
+var log = logf.RuntimeLog.WithName("certwatcher")
+
+// CertWatcher watches certificate and key files for changes.  When either file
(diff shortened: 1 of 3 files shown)


pkg/webhook/internal/certwatcher/certwatcher.go
```diff
@@ -0,0 +1,162 @@
...
+limitations under the License.
+*/
+
+package certwatcher
+
+import (
+	"crypto/tls"
+	"sync"
+
+	"gopkg.in/fsnotify.v1"
```
mengqiy: I'd suggest using `https://github.com/fsnotify/fsnotify` directly. CR is now using go modules, we don't need to use `gopkg.in`.

pkg/webhook/internal/certwatcher/certwatcher.go
```diff
@@ -0,0 +1,162 @@
...
+	cw.Lock()
+	cw.currentCert = &cert
+	cw.Unlock()
+
+	log.Info("Updated current TLS certiface")
+
+	return nil
+}
+
+func (cw *CertWatcher) handleEvent(event fsnotify.Event) {
```
mengqiy: When a secret get updated it actually triggers a rename event. Ref: [algorithm of volume writer](https://github.com/kubernetes/kubernetes/blob/b5c34d0c436dc6b95f9b78769d090af1bbbfa629/pkg/volume/util/atomic_writer.go#L89-L119).
I'd like to know what testing has been done to verify if it work.

mengqiy:
@bison Please fix travis build

pkg/webhook/internal/certwatcher/certwatcher.go
bison: Right, this is a bit tricky. So, fsnotify always fully resolves symlinks. In the case of a Kubernetes volume mount, the certificate and key files in the current data directory are watched. On an update, it's actually the remove event after the rename that is picked up here, then the watch is re-added following the new symlink.

I've tested this manually pretty extensively. My use-case is that I have another controller in the cluster generating and rotating short-lived TLS certificates for my controller-runtime based project. With this patch, webhook requests have continued to succeed after the periodic refresh of the TLS assets.

pkg/webhook/internal/certwatcher/certwatcher.go
bison: I used the `gopkg.in` URL because it was already vendored this way as an indirect dependency. I see that the `Gopkg.*` files still exist, and that `dep check` is run as part of the test suite. `Gopkg.toml` references `gopkg.in/fsnotify.v1`.

PR #422, opened 2019-05-13
:running: Remove vendor

DirectXMan12:
Modules don't need vendor any more, so it doesn't serve any purpose.


tamalsaha:
I wonder if the vendor folder should be preserved until GOPROXY becomes common/stable and Go 1.13 is released (when go mod becomes default on). I saw that prometheus-operator project uses `go build -mod=vendor` to keep using the vendor folder.

DirectXMan12:
yeah, that's what we had in controller-tools before, but I'm not convinced of how useful it is.  What's the usecase?

tamalsaha:
Vendoring is useful for folks who are behind slow internet connection. This feels a bit rushed. But I have no objection.

tamalsaha:
Another usefulness of vendor folder is to see what has changed when go.mod file is updated. The version syntax of go.mod is unreadable for k8s.io repos. So that might be useful at times. But that can be done locally, I guess.

DirectXMan12:
> Vendoring is useful for folks who are behind slow internet connection

I am sympathetic to that, but ideally the local shared module cache should help with that (and, not checking in vendor helps keep the overall repo size down).  Additionally, the vendor directory won't actually help if you're consuming controller-runtime, just if you're hacking on CR itself, so for users with slow internet, it doesn't matter either way.

> The version syntax of go.mod is unreadable for k8s.io repos

Yeah, that part's fairly unfortunate, but there's not really much we can do until k/k gets its act together.

PR #424, opened 2019-05-13
support HA (non leader election) components

mengqiy:
controller manager support 2 set of runables
- one set is managed by leader election. i.e. only one replica running.
- one set is not managed by leader election. i.e. always running.

### DIFF
--- a/examples/crd/pkg/resource.go
+++ b/examples/crd/pkg/resource.go
@@ -23,8 +23,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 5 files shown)


examples/crd/pkg/resource.go
```diff
@@ -41,12 +41,9 @@ type ChaosPodStatus struct {
 	LastRun metav1.Time `json:"lastRun,omitempty"`
 }
 
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
DirectXMan12: you need an object-gen marker here, no?

pkg/manager/internal.go
```diff
@@ -237,6 +246,8 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 		go cm.serveMetrics(cm.internalStop)
 	}
 
+	go cm.startHA()
```
DirectXMan12: it's not clear what HA means -- we should say "LeaderElection" or something instead (e.g. `NeedsLeaderElection`, `startWithLeaderElection`, etc).

PR #425, opened 2019-05-13
:bug: List requests should not pass a body.

maisem:
Currently the typedClient.List call passes in the serialized object in the body of the HTTP request. This is unnecessary and should not be done. This PR makes it so that body is not passed in the HTTP request.

@DirectXMan12
cc @ilkercelikyilmaz


DirectXMan12:
/lgtm
/approve

good catch, thanks!

PR #427, opened 2019-05-14
:sparkles: enable conversion webhook by default

droot:
This PR enables conversion webhook by default.

It also logs a warning at startup if a type is a multi-version and doesn't implement conversion. Users will get a false-positive if they build a controller for non-managed multi-version types. Looking for ideas to handle it better.

### DIFF
--- a/pkg/builder/build.go
+++ b/pkg/builder/build.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
 )
 
 // Supporting mocking out functions for testing
@@ -285,7 +286,11 @@ func (blder *Builder) doWebhook() error {
 		}
 	}
 
-	return err
+	err = conversion.CheckConvertibility(blder.mgr.GetScheme(), blder.apiType)
+	if err != nil {
+		log.Error(err, "conversion check failed", "GVK", gvk)
+	}
+	return nil
 }
 
 func generateMutatePath(gvk schema.GroupVersionKind) string {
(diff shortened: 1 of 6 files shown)


pkg/builder/build.go
```diff
@@ -285,7 +286,15 @@ func (blder *Builder) doWebhook() error {
 		}
 	}
 
-	return err
+	err = conversion.CheckConvertibility(blder.mgr.GetScheme(), blder.apiType)
+	if err != nil {
+		log.Error(err, "conversion check failed", "GVK", gvk)
```
DirectXMan12: i'm concerned that this error is going to be a bit spam-y for non-conversion cases (e.g. built-ins).  Maybe we should have an explicit opt-in on the builder?

DirectXMan12: for posterity:

- error out if there's a partial implementation (hub w/o spokes, spokes w/o hub, partial set of spokes, multiple hubs, etc)
- log on success at `V(1).Info("conversioned enabled for kind", "kind", groupKind, "hub", hubVersion, "spokes", spokeVersions)` so that people can look for that line to debug too, or check for it's absence if they've failed to implement anything.

mengqiy: > i'm concerned that this error is going to be a bit spam-y for non-conversion cases (e.g. built-ins).

If kube-apiserver sends us conversion requests for built-ins, it means there is a bug in kube-apiserver, right?

pkg/webhook/conversion/conversion_test.go
```diff
@@ -200,3 +200,39 @@ var _ = Describe("Conversion Webhook", func() {
...
+				Kind:       "ExternalJob",
+				APIVersion: "jobs.example.org/v2",
+			},
+		}
+
+		err := CheckConvertibility(scheme, obj)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should not return error for a built-in multi-version type", func() {
```
DirectXMan12: we should probably have a test for the error case

droot: Yes, we should. It requires adding new type definitions (with partial conversion implementation), I will do that in a follow up PR.

PR #435, opened 2019-05-17
:warning: Minimal Apply Support, Fix Up Client Options

DirectXMan12:
This fixes up client options (collapsing nillary functions to not require parens), makes sure that status update takes options, adds status patch, and adds a patch type and relevant options for server-side apply.


pkg/client/typed_client.go
```diff
@@ -156,6 +156,32 @@ func (c *typedClient) UpdateStatus(ctx context.Context, obj runtime.Object) erro
...
+	if err != nil {
+		return err
+	}
+
+	patchOpts := &PatchOptions{}
+	return o.Patch(patch.Type()).
+		NamespaceIfScoped(o.GetNamespace(), o.isNamespaced()).
+		Resource(o.resource()).
+		Name(o.GetName()).
+		SubResource("status").
```
droot: ok (I get it) server-side apply is implemented using patch on status sub-resource, little intuitive but I am sure they must have some reasons for it. 
So that means, CRD needs to have status sub-resource enabled (I wonder if we should generate marker for status subresource by default or we already do ?)

pkg/client/client_test.go
```diff
@@ -2262,7 +2268,7 @@ var _ = Describe("Patch", func() {
 			patch := client.MergeFrom(cm.DeepCopy())
 
 			By("returning a patch with type MergePatch")
-			Expect(patch.Type()).To(Equal(types.MergePatchType))
+			Expect(patch.Type()).To(Equal(types.StrategicMergePatchType))
 
 			By("retrieving modifying the config map")
 			metav1.SetMetaDataAnnotation(&cm.ObjectMeta, annotationKey, annotationValue)
```
droot: Do we need to add any tests for applyPatch ?

pkg/client/options.go
```diff
@@ -0,0 +1,417 @@
...
+
+	// Force is going to "force" Apply requests. It means user will
+	// re-acquire conflicting fields owned by other people. Force
+	// flag must be unset for non-apply patch requests.
+	// +optional
+	Force *bool
+
+	// FieldManager is the name of the user or component submitting
+	// this request.  It must be set with server-side apply.
+	FieldManager string
```
mengqiy: All of Create|Update|PatchOptions in upstream has `FieldManager`.
In CR, only `PatchOptions` has explicit `FieldManager` field. Should we make it consistent for all of Create|Update|PatchOptions?

pkg/client/client_test.go
```diff
@@ -2157,19 +2157,25 @@ var _ = Describe("Client", func() {
...
+			client.ForceOwnership(po)
 			mpo := po.AsPatchOptions()
 			Expect(mpo.Force).NotTo(BeNil())
 			Expect(*mpo.Force).To(BeTrue())
 		})
 
+		It("should allow setting the field manager", func() {
+			po := &client.PatchOptions{}
+			client.FieldOwner("some-owner")(po)
+			Expect(po.AsPatchOptions().FieldManager).To(Equal("some-owner"))
```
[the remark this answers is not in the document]
DirectXMan12: for posterity: manager is the term used by the kubernetes API, but I feel it's non-intuitive, so I picked fieldowner for the option name.  We do that a couple of places.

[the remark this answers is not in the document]
apelisse: We've had a few discussion, and "official" term is `FieldManager`

pkg/client/typed_client.go
apelisse: > ok (I get it) server-side apply is implemented using patch on status sub-resource, little intuitive [clipped]

Not just status, all resources/sub-resources.

PR #443, opened 2019-05-24
:bug: set status in conversion response

mengqiy:
We need to set "success" status in conversion response when we successfully convert the object.
kube-apiserver rejects it if the status is not set.

### DIFF
--- a/pkg/webhook/conversion/conversion_test.go
+++ b/pkg/webhook/conversion/conversion_test.go
@@ -122,6 +122,7 @@ var _ = Describe("Conversion Webhook", func() {
 		convReview := doRequest(convReq)
 
 		Expect(convReview.Response.ConvertedObjects).To(HaveLen(1))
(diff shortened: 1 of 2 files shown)


pkg/webhook/conversion/conversion_test.go
```diff
@@ -122,6 +122,7 @@ var _ = Describe("Conversion Webhook", func() {
 		convReview := doRequest(convReq)
 
 		Expect(convReview.Response.ConvertedObjects).To(HaveLen(1))
+		Expect(convReview.Response.Result.Status).To(BeIdenticalTo(metav1.StatusSuccess))
 		got, _, err := decoder.Decode(convReview.Response.ConvertedObjects[0].Raw)
```
droot: Equal ?

mengqiy: Done

Issue #364, opened 2019-03-19
support per-controller leader election


DirectXMan12:
It's been pointed out that per-controller leader election can be very useful, since it allows you to safely migrate controllers across binaries.

/kind feature
/priority important-longterm

GrigoriyMikhalkin:
Hi, i would like to try to implement this.

DirectXMan12:
/lifecycle frozen

work is in progress

vincepri:
Closing this given that we're heavily relying on managers today to achieve leader election, and it'd be hard to support per-controller LE settings

/close

PR #444, opened 2019-05-25
✨ added per controller leader election

GrigoriyMikhalkin:
Attempt to solve https://github.com/kubernetes-sigs/controller-runtime/issues/364

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
+	"sigs.k8s.io/controller-runtime/pkg/leaderelection"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/ratelimiter"
@@ -48,8 +49,8 @@ type Options struct {
 
 // Leader Election options
 type LeaderElectionOptions struct {
-	// NeedLeaderElection determines whether or not to use leader election when starting the controller.
-	NeedLeaderElection bool
+	// LeaderElectionMode determines what leader election mode to use when starting the controller.
+	LeaderElectionMode leaderelection.Mode
 
 	// LeaderElectionID determines the name of the configmap that leader election will use for holding the leader lock.
 	LeaderElectionID string
@@ -96,8 +97,9 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 	}
 
 	if options.LeaderElection == nil {
+		// Defaulting to per-manager leader election mode for backwards compatibility
 		options.LeaderElection = &LeaderElectionOptions{
-			NeedLeaderElection: true,
+			LeaderElectionMode: leaderelection.PerManagerLeaderElectionMode,
 			LeaderElectionID:   name,
 		}
 	}
@@ -120,7 +122,7 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 		},
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		Name:                    name,
-		LeaderElection:          options.LeaderElection.NeedLeaderElection,
+		LeaderElectionMode:      options.LeaderElection.LeaderElectionMode,
 		LeaderElectionID:        options.LeaderElection.LeaderElectionID,
 	}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -41,6 +41,18 @@ type Options struct {
 	// Defaults to MaxOfRateLimiter which has both overall and per-item rate limiting.
 	// The overall is a token bucket and the per-item is exponential.
 	RateLimiter ratelimiter.RateLimiter
+
+	// Leader election is by default
+	LeaderElection *LeaderElectionOptions
+}
+
+// Leader Election options
+type LeaderElectionOptions struct {
+	// NeedLeaderElection determines whether or not to use leader election when starting the controller.
+	NeedLeaderElection bool
+
+	// LeaderElectionID determines the name of the configmap that leader election will use for holding the leader lock.
+	LeaderElectionID string
 }
 
 // Controller implements a Kubernetes API.  A Controller manages a work queue fed reconcile.Requests
@@ -83,6 +95,13 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 		options.RateLimiter = workqueue.DefaultControllerRateLimiter()
 	}
 
+	if options.LeaderElection == nil {
+		options.LeaderElection = &LeaderElectionOptions{
+			NeedLeaderElection: true,
+			LeaderElectionID:   name,
+		}
+	}
+
 	// Inject dependencies into Reconciler
 	if err := mgr.SetFields(options.Reconciler); err != nil {
 		return nil, err
@@ -101,6 +120,8 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 		},
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		Name:                    name,
+		LeaderElection:          options.LeaderElection.NeedLeaderElection,
+		LeaderElectionID:        options.LeaderElection.LeaderElectionID,
 	}
 
 	// Add the controller as a Manager components
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -50,6 +50,14 @@ type Controller struct {
 	// MaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run. Defaults to 1.
 	MaxConcurrentReconciles int
 
+	// LeaderElection determines whether or not to use leader election when
+	// starting the controller.
+	LeaderElection bool
+
+	// LeaderElectionID determines the name of the configmap that leader election
+	// will use for holding the leader lock.
+	LeaderElectionID string
+
 	// Reconciler is a function that can be called at any time with the Name / Namespace of an object and
 	// ensures that the state of the system matches the state specified in the object.
 	// Defaults to the DefaultReconcileFunc.
@@ -295,4 +303,12 @@ func (c *Controller) InjectFunc(f inject.Func) error {
 // updateMetrics updates prometheus metrics within the controller
 func (c *Controller) updateMetrics(reconcileTime time.Duration) {
 	ctrlmetrics.ReconcileTime.WithLabelValues(c.Name).Observe(reconcileTime.Seconds())
+}
+
+func (c *Controller) NeedLeaderElection() bool {
+	return c.LeaderElection
+}
+
+func (c *Controller) GetID() string {
+	return c.LeaderElectionID
 }
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
+	"sigs.k8s.io/controller-runtime/pkg/leaderelection"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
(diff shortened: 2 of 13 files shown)


pkg/controller/controller.go
```diff
@@ -18,23 +18,65 @@ package controller
...
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/recorder"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
+	defaultLeaseDuration = 15 * time.Second
```
mengqiy: These are shared bits between `controller` and `manager` packages and they can potentially moved to `sigs.k8s.io/controller-runtime/pkg/leaderelection/internal`.

GrigoriyMikhalkin:
> We should add some validation to check if there are 2 different controller or manager using the same LE namespace and ID when starting the manager.

@mengqiy You mean, check duplicate LE lock names used by manager and controllers that it manages?

mengqiy:
> You mean, check duplicate LE lock names used by manager and controllers that it manages? Or you mean totally unrelated managers and controllers? If later, i'm not sure how we can check that LE lock is created by different controller or manager.

I mean the former.

DirectXMan12:
that would also allow the manager to check for duplicate ids within a given manager

GrigoriyMikhalkin:
@mengqiy @DirectXMan12 Completely rewrote code. Added new runnable interface(called it `SingletonRunnable`, probably not the best naming). Now manager takes care of leader election for controllers, also checks that there's no duplicate LE IDs.

GrigoriyMikhalkin:
> we already have the `LeaderElectionRunnable` interface, we should be able to use that.

`LeaderElectionRunnable` interface used for checking if runnable have to be run before or after manager was elected. Are you implying that we can add `GetLeaderElectionID` method to it and use `NeedLeaderElection` to determine if need to perform leader election on runnable? It seems like it can be a breaking change for some users(who implemented they own runnables), would it be ok?

DirectXMan12:
> interface used for checking if runnable have to be run before or after manager was elected

Correct -- "run before" is equivalent to "don't need leader election".

We could actually just have `NeedLeaderElection` return an ID as well as a bool (or we could have the separate "GetID" method).

> It seems like it can be a breaking change for some users(who implemented they own runnables), would it be ok

We're doing breaking changes this release anyway, but we could still have the default be to run under some standard leader election ID or something.

pkg/manager/internal.go
```diff
@@ -265,8 +278,18 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 
 	go cm.startNonLeaderElectionRunnables()
 
-	if cm.resourceLock != nil {
-		err := cm.startLeaderElection()
+	if resourceLock := cm.resourceLocksMap[cm.leaderElectionID]; resourceLock != nil {
+		err := cm.startLeaderElection(cm.leaderElectionID, leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(_ context.Context) {
+				cm.startLeaderElectionRunnables()
```
DirectXMan12: I don't think we need this double-leader election here -- in fact, it kinda defeats the purpose of per-controller leader election.

DirectXMan12: Really, you want to try to start all of the locks independently.

pkg/manager/internal.go
```diff
@@ -313,7 +336,46 @@ func (cm *controllerManager) startLeaderElectionRunnables() {
...
+
+				// Check that leader election ID is defined
+				if leID == "" {
+					cm.errChan <- errors.New("LeaderElectionID must be configured")
+					return
+				}
+
+				// Check that leader election ID is unique
+				if _, exists := cm.resourceLocksMap[leID]; exists {
+					cm.errChan <- errors.New("LeaderElectionID must be unique")
```
DirectXMan12: that's not entirely true -- a group of controllers could choose to run under the same leader election

DirectXMan12: (which might make sense if they needed the same resources, etc.

GrigoriyMikhalkin:
Updated. Now there's no global leader lock. But i use same`LeaderElectionNamespace`(and different duration parameters) for all per controller locks, is that fine? IMHO, i think it is, seems like specifying different namespaces for locks is redundant.

mengqiy:
> But i use same `LeaderElectionNamespace` (and different duration parameters) for all per controller locks, is that fine? IMHO, i think it is, seems like specifying different namespaces for locks is redundant.

How to handle LE namespace is debatable. Using the same `LeaderElectionNamespace` probably is not flexible enough. What if some controllers want to do LE in the same namespace as the manager and some controllers want to do LE in a dedicated namespace e.g. `leader-election-system` or `kube-system`?
IMHO we can make `GetID` return namespace and object ID or have another function for getting namespace e.g. `GetNamespace()`. If getting an empty namespace, we still default it to the same namespace as the manager.

GrigoriyMikhalkin:
Resolved conflict that was caused by merging https://github.com/kubernetes-sigs/controller-runtime/pull/487 with master. So, now if new controller is added after manager has started:

* If controller don't require leader election -- it's started
* If controller requires leader election and Leader Election ID is already used, we check if this ID was already elected:

    1) If it wasn't -- just add this controller, to array with other runnables that should run under this ID
    2) If it was elected -- start it.

* If controller requires leader election and Leader Election ID is new, we run leader election for it.

pkg/manager/internal.go
```diff
@@ -135,22 +144,47 @@ func (cm *controllerManager) Add(r Runnable) error {
...
+		}
+
+		cm.leaderElectionRunnables[leID] = runnables
+
+		if cm.started {
+			// If Leader Election ID is already used and elected, start the controller.
+			// If it's appeared first time, start leader election for it.
+			if elected, _ := cm.electedLeaderElectionIDs[leID]; elected {
+				go func() {
+					cm.errChan <- r.Start(cm.internalStop)
```
DirectXMan12: can we ever have a case where this races with losing the lease?

pkg/manager/internal.go
```diff
@@ -346,23 +366,51 @@ func (cm *controllerManager) waitForCache() {
...
+			runnables := cm.leaderElectionRunnables[leaderElectionID]
+			for _, r := range runnables {
+				runnable := r
+				go func() {
+					cm.errChan <- runnable.Start(cm.internalStop)
+				}()
+			}
+		},
+		OnStoppedLeading: func() {
+			cm.errChan <- fmt.Errorf("runnable leader election lost")
```
DirectXMan12: this'll kill the entire controller.  Don't we want to continue operating?  What's the intended behavior here?  Do we need per-le-group stop channels?

GrigoriyMikhalkin: Yes, i think per-le-group stop channels would be reasonable.

GrigoriyMikhalkin:
@DirectXMan12 Added per-le-group stop channels

pkg/manager/internal.go
GrigoriyMikhalkin: Currently, if lease is lost we close le-group channel so runnable is started with closed channel. It's expected that runnable will exit immediately if stop channel is closed(this is the case with `controller.Controller` right now)

GrigoriyMikhalkin:
@DirectXMan12 Updated PR, added stop channels for groups and now always starting leader election in `OnStoppedLeading` callback(so that controller could acquire lease after it was lost)

pkg/internal/controller/controller.go
```diff
@@ -256,3 +264,11 @@ func (c *Controller) InjectFunc(f inject.Func) error {
 func (c *Controller) updateMetrics(reconcileTime time.Duration) {
 	ctrlmetrics.ReconcileTime.WithLabelValues(c.Name).Observe(reconcileTime.Seconds())
 }
+
+func (c *Controller) NeedLeaderElection() bool {
```
DirectXMan12: Now that we implement this, I think this needs to return true by default, otherwise this is a fairly surprising change in default behavior (anything that doesn't implement LeaderElectionRunnable is assumed to use leader election to avoid breaking people too badly).

pkg/manager/internal.go
```diff
@@ -135,22 +150,44 @@ func (cm *controllerManager) Add(r Runnable) error {
...
-	var shouldStart bool
+	if cm.leaderElectionRunnables == nil {
+		cm.leaderElectionRunnables = make(map[string][]Runnable)
+	}
 
 	// Add the runnable to the leader election or the non-leaderelection list
-	if leRunnable, ok := r.(LeaderElectionRunnable); ok && !leRunnable.NeedLeaderElection() {
-		shouldStart = cm.started
-		cm.nonLeaderElectionRunnables = append(cm.nonLeaderElectionRunnables, r)
+	if leRunnable, ok := r.(LeaderElectionRunnable); ok && leRunnable.NeedLeaderElection() {
```
DirectXMan12: this was actually intentional (we should've had a comment here) -- we default to "use leader election" intentionally to avoid breaking existing code.

pkg/internal/controller/controller.go
DirectXMan12: (see below)

DirectXMan12:
Thanks for all your hard work on this, btw! :-)

GrigoriyMikhalkin:
@DirectXMan12
> default controllers to use their names as LE IDs

Problem is we can't get name for `Runnable` if it's not implementing `LeaderElectionRunnable` interface.

> Finally, we could assume you want LE by default if you don't specify the interface (as we used to), and default controllers to some global LE identifier if you don't specify another name. This matches existing behavior, but seems meh because per-controller LE now requires explicit overrides.

Seems like best option here

shawn-hurley:
> BIG BOLD LETTERS that you need to make sure you specify LE IDs on all runnables that need leader election, including your controllers. 

I agree this is going to make setup annoying.

> We can assume you want leader election by default if you don't specify the interface (as we used to, which covers basic runnables), and we can also default controllers to use their names as LE IDs, and default to "on" if not explicitly specified. 

I think this probably makes the most sense to me personally. Sorry if this is obvious but was there already a requirement on controller names being unique within a process? I assume you wanted to do that but this could go from annoying logs to certain controllers never start which we should clearly document IMO.


> Finally, we could assume you want LE by default if you don't specify the interface (as we used to), and default controllers to some global LE identifier if you don't specify another name. This matches existing behavior but seems meh because per-controller LE now requires explicit overrides.

+1

DirectXMan12:
> Sorry if this is obvious but was there already a requirement on controller names being unique within a process?

More-or-less -- we don't explicitly error out, but you'll start seeing error messages from a few things that require it (like metrics).

GrigoriyMikhalkin:
@DirectXMan12 Update PR.

1) Controllers use leader election by default

2) If runnable doesn't implement `LeaderElectionRunnable` it either:

    - If `DefaultLeaderElection` set to true in Manager options -- will use leader election
    - If `DefaultLeaderElection` set to false(default) -- will not

I added `DefaultLeaderElection` and `DefaultLeaderElectionID` fields to Manager options. `DefaultLeaderElection` specifies if runnables that don't implement `LeaderElectionRunnable` should use leader election by default, it doesn't affect runnables that implement interface. `DefaultLeaderElectionID` -- is common leID for those runnables.

GrigoriyMikhalkin:
Fixed conflicts

pkg/webhook/server.go
```diff
@@ -95,6 +95,10 @@ func (*Server) NeedLeaderElection() bool {
 	return false
 }
 
+func (*Server) GetID() string {
```
vincepri: Add a comment here and explain why we're returning an empty string

DirectXMan12:
Outcome of the f2f at kubecon was to support this, but default to manager's LE ID.  We need to rebase this and make sure that's the state it's in.  I'll take a poke at it.

vincepri:
@GrigoriyMikhalkin are you still interested in working on this change?

GrigoriyMikhalkin:
> @GrigoriyMikhalkin are you still interested in working on this change?

Yes. I'm gonna update this PR this weekend.

vincepri:
Thank you!

GrigoriyMikhalkin:
> Outcome of the f2f at kubecon was to support this, but default to manager's LE ID.
@DirectXMan12 So, by default, we want to make leader election for controllers?

pkg/webhook/server.go
GrigoriyMikhalkin: added comment

DirectXMan12:
Basically, by default, keep the current behavior unless explicitly overridden by the controller.

GrigoriyMikhalkin:
Updated PR. Now `LeaderElectionRunnable` should implement `GetLeaderElectionMode` method, which returns one of (Non, per-manager, per-controller) leader election modes. Controllers, by default, use per-manager mode(current behavior).

@DirectXMan12, @vincepri please review

GrigoriyMikhalkin:
Planning to fix conflicts this weekend. Is there still plans to merge this?

PR #446, opened 2019-05-25
:warning: Remove defaulting for leader election ID

### LINKED ISSUE #445: Remove defaulting for leader election ID
Currently, the leader election ID defaults to `controller-leader-election-helper`.  Since it is convenient to not configure anything for the leader election ID, it will crash whenever two controller-runtime based controllers run in the same namespace or have the same namespace configured for the leader election (e.G. `kube-system`).

This is especially bad since there is no logging about the lock being held by a different controller, so from a users perspective this looks like the controller just froze.

alvaroaleman:
With the current defaulting for the leader election ID, there is a clash
as soon as two controller that do not have it explicitly configured run
in the same namespace or have the same namespace configured for leader
election.

This is especially bad since there is no logging about the lock being
held by a different controller, so from a users perspective this looks
like the controller just froze.

Fixes #445

/assign @DirectXMan12 
/cc @JoelSpeed

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -143,15 +143,16 @@ var _ = Describe("manger.Manager", func() {
 				m, err := New(cfg, Options{
 					LeaderElection:          true,
 					LeaderElectionNamespace: "default",
+					LeaderElectionID:        "test-leader-election-id",
 					newResourceLock: func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error) {
 						var err error
 						rl, err = leaderelection.NewResourceLock(config, recorderProvider, options)
 						return rl, err
 					},
 				})
-				Expect(m).ToNot(BeNil())
 				Expect(err).ToNot(HaveOccurred())
-				Expect(rl.Describe()).To(Equal("default/controller-leader-election-helper"))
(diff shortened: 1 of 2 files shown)


DirectXMan12:
/hold

till we resolve the discussion on #445 (I'm happy to be corrected on that issue, just want to finish up that discussion first)

DirectXMan12:
I don't think we meant to close this...

alvaroaleman:
@DirectXMan12 re https://github.com/kubernetes-sigs/controller-runtime/issues/445#issuecomment-561860554: Can we add this change to the `breaking-next` milestone?

DirectXMan12:
Added.  I think we've got a breaking one coming up due to #749, which I think we need in for supportability reasons.

shawn-hurley:
@DirectXMan12 bump to get this in as #749 merged

pkg/manager/manager_test.go
```diff
@@ -143,15 +143,16 @@ var _ = Describe("manger.Manager", func() {
 				m, err := New(cfg, Options{
 					LeaderElection:          true,
 					LeaderElectionNamespace: "default",
+					LeaderElectionID:        "test-leader-election-id",
```
DirectXMan12: lol, this test should have a different name now.  I'll fix it elsewhere.

PR #448, opened 2019-05-27
:bug: update internal controller for more clean code and metrics

### LINKED ISSUE #440: controller_runtime_reconcile_time_seconds includes queue wait time
This metric currently includes the time spent waiting for an item to be available in the queue (while `c.Queue.Get()` is blocking):

https://github.com/kubernetes-sigs/controller-runtime/blob/12d98582e72927b6cd0123e2b4e819f9341ce62c/pkg/internal/controller/controller.go#L176-L182

It would be more useful if it measured only the time spent doing reconciliation after an item is returned by the queue.

adohe:
update internal controller process loop, this also fix #440

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -154,10 +154,7 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 	log.Info("Starting workers", "controller", c.Name, "worker count", c.MaxConcurrentReconciles)
 	for i := 0; i < c.MaxConcurrentReconciles; i++ {
 		// Process work items
-		go wait.Until(func() {
-			for c.processNextWorkItem() {
-			}
-		}, c.JitterPeriod, stop)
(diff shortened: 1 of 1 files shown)


pkg/internal/controller/controller.go
```diff
@@ -168,30 +165,32 @@ func (c *Controller) Start(stop <-chan struct{}) error {
...
 		c.updateMetrics(time.Now().Sub(reconcileStartTS))
 	}()
 
-	obj, shutdown := c.Queue.Get()
-	if shutdown {
-		// Stop working
-		return false
-	}
-
-	// We call Done here so the workqueue knows we have finished
```
mengqiy: Why we delete the comment here?

adohe: Ooop~~

adohe:
comments addressed, ptal~ @mengqiy

PR #449, opened 2019-05-27
:sparkles: Allow setting "UseExistingCluster" via environment

adohe:
support setting "UseExistingCluster" via OS Env, address #222

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -87,7 +89,7 @@ type Environment struct {
 	// UseExisting indicates that this environments should use an
 	// existing kubeconfig, instead of trying to stand up a new control plane.
 	// This is useful in cases that need aggregated API servers and the like.
-	UseExistingCluster bool
+	UseExistingCluster *bool
```
mengqiy: IMO we don't have to use a boolean pointer. And this is actually a breaking change, since you change the type of a public struct.

adohe:
//assign @mengqiy

DirectXMan12:
I think we may want this to be a pointer -- this means you can never explicitly opt out of useexistingcluster.

mengqiy:
Discussed this with Solly a bit offline.
In sum, if we need to make breaking changes to make things work well, now it's time (v0.2.0). :)

PR #453, opened 2019-05-29
:sparkles: cleanup CRDs when use existing cluster as test environment

### LINKED ISSUE #420: envtest.Environment should cleanup CRDs on Stop()
`envtest.Environment` does not cleanup CRDs when `Stop()` is called. When you use an `Environment` with `UseExistingCluster = true`, this leaves CRDs hanging around in the cluster. So the second time you run `go test`, it will fail because "can't create CRDs which already exist". So it would be really useful if `Stop()` also deleted the CRDs by default irrespective of whether we use existing cluster or not.

adohe:
Closes #420

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -27,6 +27,7 @@ import (
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


adohe:
this PR fix #420

mengqiy:
/lgtm
Please fix the linter error

mengqiy:
It seems goimports is still not happy.

adohe:
@mengqiy finally CI happy, ptal

pkg/envtest/crd.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
mengqiy: I guess this line was making `goimports` unhappy.
To be consistent with other k8s projects, we should alias it as `metav1`. I believe you already know it. :)

DirectXMan12:
#457 complicates this somewhat, so whichever merges second needs to be aware -- if the CRDs were already installed, wiping them away is the wrong thing to do.

adohe:
sure, wiping the already installed CRDs is always the wrong thing to do, the envtest should just clean up newly installed CRDs when running against existing cluster, if this is want we want, I will update this PR. cc @DirectXMan12 for double confirm

DirectXMan12:
> envtest should just clean up newly installed CRDs when running against existing cluster

Yep, that's what we want

adohe:
I will hold this until #457 get merged.

PR #457, opened 2019-05-30
🐛 Don't error if the CRD is already present

coderanger:
Ignores this one error for cases with UseExistingServer because if using the same control plane for multiple suites, the CRDs will already be loaded in after the first.

A possible tweak would be to do a full create-or-update instead but it seems unlikely to matter.


DirectXMan12:
Should this just be createorupdate?  If it already exists you probably want to make sure it's up-to-date

coderanger:
@DirectXMan12 CreateOrUpdate removes some class of tests around upgrades. I would generally say that if they are already present, it means you probably set them up on purpose :)

DirectXMan12:
ah, hmm... perhaps we should at least log?

lukaszo:
@codenrhoden any update here? I would love to have it merged

codenrhoden:
@lukaszo looks like you were trying to ping @coderanger instead. =)

lukaszo:
@codenrhoden ups, sorry for pinging :)

pires:
I think it should be `CreateOrUpdate`, just in the case you have changed your CRDs in between runs.

PR #458, opened 2019-05-30
:sparkles: Allow setting controller name, fix default

DirectXMan12:
This switches the default controller name in the builder to just
lower(<kind>) which fixes any issues that arrise from it showing up in
the deprecated metrics (which use the workqueue name as part of the
metric name).

It also allows overriding name in case of two controllers reconciling
the same type using the builder.


DirectXMan12:
cc @pwittrock

pwittrock:
woot

PR #459, opened 2019-05-30
:sparkles: more visibility of the testing control plane

mengqiy:
Support bridging the control plane output to the stdout and stderr

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -42,6 +42,7 @@ const (
 	envKubebuilderPath     = "KUBEBUILDER_ASSETS"
 	envStartTimeout        = "KUBEBUILDER_CONTROLPLANE_START_TIMEOUT"
 	envStopTimeout         = "KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT"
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -143,8 +149,28 @@ func (te *Environment) Start() (*rest.Config, error) {
...
+			te.ControlPlane.APIServer = &integration.APIServer{Args: te.getAPIServerFlags()}
+		}
+		if te.ControlPlane.Etcd == nil {
+			te.ControlPlane.Etcd = &integration.Etcd{}
+		}
+
+		if os.Getenv(envAttachOutput) == "true" {
+			te.AttachControlPlaneOutput = true
+		}
+		if te.ControlPlane.APIServer.Out == nil && te.AttachControlPlaneOutput {
```
DirectXMan12: we should make it easier to set these ourselves in case people want to redirect elsewhere.  Maybe we shouldn't clear `te.ControlPlane` above?

mengqiy: > Maybe we shouldn't clear te.ControlPlane above?

Done. Now people can directly set `Out` and `Err` in `ControlPlane.APIServer` and `ControlPlane.Etcd`.

mengqiy:
PTAL

Issue #461, opened 2019-05-31
Make UseExistingCluster a pointer?


DirectXMan12:
#449 allows setting `UseExistingCluster` via an env var, but as discussed in #222, we probably want a pointer.  Without the pointer, we don't have a way to explicitly opt-out in the code (we really want a 3-state enum: true, false, and UseEnvVar).

PR #463, opened 2019-06-02
:warning: make UseExistingCluster a pointer to support explicitly op…

adohe:
…t-out in the code

just as what #461 mentioned, make UseExistingCluster a pointer thus gives a way to explicitly opt-out in the code.

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -90,7 +90,7 @@ type Environment struct {
 	// UseExisting indicates that this environments should use an
 	// existing kubeconfig, instead of trying to stand up a new control plane.
 	// This is useful in cases that need aggregated API servers and the like.
-	UseExistingCluster bool
+	UseExistingCluster *bool
 
 	// ControlPlaneStartTimeout is the maximum duration each controlplane component
 	// may take to start. It defaults to the KUBEBUILDER_CONTROLPLANE_START_TIMEOUT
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -256,3 +253,11 @@ func (te *Environment) defaultTimeouts() error {
 	}
 	return nil
 }
+
+func (te *Environment) useExistingCluster() bool {
+	if te.UseExistingCluster == nil {
+		useExistingCluster := strings.ToLower(os.Getenv(envUseExistingCluster)) == "true"
+		te.UseExistingCluster = &useExistingCluster
```
DirectXMan12: no need to populate this here -- just return the env var without populating

adohe: comment addressed, ptal

pkg/envtest/server.go
```diff
@@ -256,3 +252,10 @@ func (te *Environment) defaultTimeouts() error {
 	}
 	return nil
 }
+
+func (te *Environment) useExistingCluster() bool {
+	if te.UseExistingCluster == nil {
+		return strings.ToLower(os.Getenv(envUseExistingCluster)) == "true"
+	}
+	return *te.UseExistingCluster == true
```
DirectXMan12: nit: no need for `== true` here, since it's a bool

adohe:
comment addressed and CI happy @DirectXMan12 ptal

PR #465, opened 2019-06-04
:running: use :ghost: for no release note commits

### LINKED ISSUE #433: Use :ghost: to not show a commit message
Use :ghost: to prefix the commit message if we don't want it to show up in the release note.

mengqiy:
fixes #433

I was thinking using `:see_no_evil:` :see_no_evil:, but I feel the wording `evil` may be offending. So this PR stay with `:ghost:` :ghost:.


mengqiy:
PTAL

PR #466, opened 2019-06-04
🐛 Fix runnable race in manager

abursavich:
In `controllerManager`, `startLeaderElectionRunnables` doesn't protect its read of `leaderElectionRunnables` and write of `started`. This may cause a Runnable to be run twice:

- `startNonLeaderElectionRunnables` sets `started` to true
- `Add` appends the runnable to `leaderElectionRunnables`, sees that that `started` is true, and starts the runnable
- `startLeaderElectionRunnables` starts all of the runnables in `leaderElectionRunnables`

See https://travis-ci.org/kubernetes-sigs/controller-runtime/jobs/541066866 for an example.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -288,6 +288,41 @@ func (cm *controllerManager) startNonLeaderElectionRunnables() {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
 
+	cm.waitForCache()
+
+	// Start the non-leaderelection Runnables after the cache has synced
+	for _, c := range cm.nonLeaderElectionRunnables {
+		// Controllers block, but we want to return an error if any have an error starting.
+		// Write any Start errors to a channel so we can return them
+		ctrl := c
+		go func() {
+			cm.errChan <- ctrl.Start(cm.internalStop)
+		}()
+	}
+}
+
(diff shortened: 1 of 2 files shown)


abursavich:
I should note that there's another subtle bug with adding leader election runnables, which I noticed when reading this code but this change **does not** fix.

If a runnable is added after the manager has started, the runnable is unconditionally started.  In the case that the runnable **should** wait for leader election, it will start early (without leadership) and then it will be started **again** when the manager wins election.

DirectXMan12:
Thank you for fixing this.  Do you mind also fixing the additional bug you mentioned?

DirectXMan12:
otherwise looks good.  I don't suppose you've found a good way of testing this behavior?  races are very hard to test

abursavich:
I would assume that running the existing tests with the race detector would've caught this. Otherwise, no testing ideas.

I can probably get to the other issue in the next few days. I had planned on tackling them separately to keep the diffs smaller and easier to review.

DirectXMan12:
> I can probably get to the other issue in the next few days

Ack.  Can you file a separate issue then, just so we don't lose track of it?

> I would assume that running the existing tests with the race detector would've caught this

Ah, yeah, need to make sure that's on.

Issue #442, opened 2019-05-24
log/zap: Allow configuring formatter independently of `debug` option


alvaroaleman:
Right now, the formatter/encoder of the `log/zap` package is inferred from the `debug` option, if its enabled a console encoder is used, otherwise a json encoder.  This does not always make sense, e.G. sometimes debug log may help in production, but that should not result in unstructured logging.

It would be awesome, if another option was added to [`RawLoggerTo`](https://github.com/kubernetes-sigs/controller-runtime/blob/96b67f231945a8e8550dbdd8bfbd550a0c68097b/pkg/log/zap/zap.go#L49) to make this configurable. Alternatively, we could use a different type for `opts` that contains an option for the encoder.

hpandeycodeit:
@alvaroaleman @DirectXMan12 made some changes to make this configurable. #467 Still WIP as I need to modify the LoggerTo from other places as well if we are looking to change it dynamically. 

One question here, how are we deciding when to use console encoder and when not?

DirectXMan12:
By default, it was debug vs non-debug, but I think we need to start considering how we want to expose these options.

hpandeycodeit:
@DirectXMan12 can't we use existing `development` flag to use this option in production as well?

https://github.com/kubernetes-sigs/controller-runtime/blob/71cdf355b7a663a0ccb3248dc4699ba10c663b13/pkg/log/zap/zap.go#L33-L35

DirectXMan12:
The development flag control also controls output format (human-readable vs json), so we need more granular options.

DirectXMan12:
Perhaps we should adopt the "functional options" style here too.

PR #467, opened 2019-06-04
 ✨[WIP] Added changes to make console encoder configurable

hpandeycodeit:
This PR will make the console encoder configurable. 
Added an extra parameter in the function RawLoggerTo where a bool value can be passed to use console encoder for Production. 

As per issue #442


DirectXMan12:
/hold
re my comment on the original issue about how to pass options

Issue #415, opened 2019-05-07
unstructured client should key off of the unstructured interface


DirectXMan12:
If we key off of the unstructured interface instead of the type, we can allow for alternate internal unstructured representations, and the underlying machinery works fine with the interface.

vincepri:
/priority awaiting-more-evidence
/help

We might need a spike to understand what we learned from #468.

PR #468, opened 2019-06-05
✨ unstructured client should key off unstructured

anthonyho007:
Make Unstructured client to key off unstructured

Issue : https://github.com/kubernetes-sigs/controller-runtime/issues/415

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -22,7 +22,6 @@ import (
 	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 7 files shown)


pkg/client/client.go
```diff
@@ -105,54 +105,54 @@ type client struct {
 
 // Create implements client.Client
 func (c *client) Create(ctx context.Context, obj runtime.Object, opts ...CreateOptionFunc) error {
-	_, ok := obj.(*unstructured.Unstructured)
+	u, ok := obj.(*unstructured.Unstructured)
```
DirectXMan12: I'm not entirely certain as to what you're trying to do here, but you haven't accomplished what was requested in the linked issue.  That's about using the separate `Unstructured` interface instead of the literal struct implementation `unstructured.Unstructured`

anthonyho007: I have totally misunderstood the ticket. Do you mean by having a custom struct definition for unstructured instead of using unstructured.Unstructured from apimachinery?

DirectXMan12: there's an equivalent interface to `unstructured.Unstructured` that lives in APIMachinery.  It's something like `runtime.Unstructured', and, IIRC, it's what all the underlying methods actually take in APIMachinery

pkg/client/interfaces.go
```diff
@@ -44,10 +45,36 @@ type Patch interface {
...
 	// Data is the raw data representing the patch.
 	Data(obj runtime.Object) ([]byte, error)
+	// DataFromUnstructured is the raw data representing the patch from unstructured
+	DataFromUnstructured(obj Unstructured) ([]byte, error)
 }
 
 // TODO(directxman12): is there a sane way to deal with get/delete options?
 
+// Unstructured is an interface to type unstructured.Unstructured in apimachinery
+type Unstructured interface {
```
DirectXMan12: why aren't we just using `runtime.Unstructured` here?

pkg/client/client.go
anthonyho007: I hope this is what you meant 😅. would it be a good idea use this interface in pkg/cache too ?

pkg/client/interfaces.go
anthonyho007: coz we only use `SetUnstructuredContent(map[string]interface{})` and `Object` interface from `runtime.Unstructured`. since it will save us two lines of code, i will include it.

DirectXMan12:
I suspect I may have been a bit unclear here, so I apologize.  We should have the following goals:

1. We don't need to define our own unstructured interface
2. The unstructured client continues to implement the `client.Client` interface  

This should be possible.  Basically, what we'll need to do is:

1. replace all the type assertions in the client internals, substituting `*unstructured.Unstructured` for `runtime.Unstructured`
2. Adapt to using the `runtime.Unstructured` interface:
   a. calls to `u.GroupVersionKind()` become `u.GetGroupKind().GroupVersionKind()`
   b. calls that use the helpers defined on `unstructured.Unstructured`, like `u.GetNamespace`, will probably need to end up looking like
    ```go
    rawData := &unstructured.Unstructured{Object: u.UnstructuredContent()}
    // and then stuff like
    ns := rawData.Namespace()
    resourceInterface.UpdateStatus(rawData, ...)
   ```
   c. Where before we wrote `u.Object = i.Object`, we'll now write `u.SetUnstructuredContent(i.Object)`
   d. Similarly, where we used unstructured list helpers, we can make use of `u.EachListItem` and the aforementioned pattern of `rawData := &unstructured.UnstructuredList{}; rawData.SetUnstructuredContent(u.UnstructuredContent())`

We may end up eventually rewriting the dynamic client anyway, since it doesn't support contexts, so we may need less shuffling back and forth between runtime.Unstructured and unstructured.Unstructured.  If you'd like to tackle that as well, that's fine, otherwise we can put it off till later.

anthonyho007:
👋 @DirectXMan12 ! Thanks for your very detailed explanations and i think i have a clear picture of what you intended to do in the issue. I should have familiarize myself with apimachinery codebase first 😅. I would love to tackle the dynamic client issue but i wanna finish this issue first! and thanks for the code review !!!

DirectXMan12:
(and, fwiw, I should probably appologize for putting good first issue on there -- I missed that there was a bit of a mismatch in exposed methods between the struct and the interface when I first added the label)

PR #487, opened 2019-06-12
:bug: fix runnable run twice issue

### LINKED ISSUE #478: Runnable added after manager started skips leader election and may run twice
If a runnable is added after the manager has started, the runnable is unconditionally started. In the case that the runnable should wait for leader election, it will start early (without leadership) and then it will be started again when the manager wins election.

https://github.com/kubernetes-sigs/controller-runtime/pull/466#issuecomment-498816582

adohe:
fix #478 @DirectXMan12 @abursavich ptal

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -91,9 +91,10 @@ type controllerManager struct {
 	// metricsListener is used to serve prometheus metrics
 	metricsListener net.Listener
 
-	mu      sync.Mutex
-	started bool
-	errChan chan error
+	mu            sync.Mutex
+	started       bool
+	startedLeader bool
+	errChan       chan error
 
 	// internalStop is the stop channel *actually* used by everything involved
 	// with the manager as a stop channel, so that we can pass a stop channel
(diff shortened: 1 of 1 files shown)


pkg/manager/internal.go
```diff
@@ -347,6 +352,7 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 		RetryPeriod:   cm.retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
+				cm.isLeader = true
```
abursavich: Setting `isLeader` needs to be protected by the mutex. I might suggest moving it into `startLeaderElectionRunnables` and changing its name to something like `startedLeader`?

pkg/manager/internal.go
```diff
@@ -135,14 +136,18 @@ func (cm *controllerManager) Add(r Runnable) error {
...
 	}
 
+	shouldStart := false
+
 	// Add the runnable to the leader election or the non-leaderelection list
 	if leRunnable, ok := r.(LeaderElectionRunnable); ok && !leRunnable.NeedLeaderElection() {
+		shouldStart = cm.started
 		cm.nonLeaderElectionRunnables = append(cm.nonLeaderElectionRunnables, r)
 	} else {
+		shouldStart = cm.started && (cm.isLeader || cm.resourceLock == nil)
```
abursavich: If you push take the suggestion from the other comment, this could be simplified to `shouldStart = cm.startedLeader`.

pkg/manager/internal.go
adohe: yeap, that makes more sense. 👍

DirectXMan12:
fyi #444

abursavich:
The code looks good to me, but it would be nice to see the tests pass -- but that failure doesn't look like your fault.

adohe:
@DirectXMan12 kind ping~

adohe:
@mengqiy @DirectXMan12 the CI is happy, kind ping~

PR #489, opened 2019-06-14
:sparkles: add GetConfigWithContext to retrieve a config with a specific context

### LINKED ISSUE #479: Helper to fetch config using a particular kubeconfig context
It would be nice if a `--context` flag would be supported (same semantics as in kubectl).

Use Case:
I'm developing locally on a ClusterAPI controller (which uses controller-runtime). It's kind of dangerous, if it uses whatever the current context is in my local kubeconfig. It would be nice to just "pin" the context via command-line flag.

I would like to open a PR, if this feature makes sense to you.

sbueringer:
This PR adds a new function to the config package. Before it was only possible to retrieve the current context of a kubeconfig file via `GetConfig`. Now there's an additional `GetConfigWithContext` where a specific context can be retrieved. 

The advantage is when developing locally (run controllers via IDE) with a specified context, the controller is not dependent on whatever the current context is of the used kubeconfig file.

Fixes #479

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -970,6 +970,7 @@
     "k8s.io/client-go/testing",
     "k8s.io/client-go/tools/cache",
     "k8s.io/client-go/tools/clientcmd",
+    "k8s.io/client-go/tools/clientcmd/api",
     "k8s.io/client-go/tools/leaderelection",
     "k8s.io/client-go/tools/leaderelection/resourcelock",
     "k8s.io/client-go/tools/metrics",
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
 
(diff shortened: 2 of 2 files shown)


sbueringer:
@DirectXMan12 How do I fix the following build error? Couldn't find anything in the docs. `dep ensure` changes a lot in Gpkg.lock, go.mod & go.sum.
````
k8s.io/client-go/tools/clientcmd/api: imported or required, but missing from Gopkg.lock's input-imports
The command "GO111MODULE=on TRACE=1 ./hack/check-everything.sh" exited with 1.
````

DirectXMan12:
`dep ensure --no-vendor` should fix dep, and `go mod tidy` should fix go.mod.

DirectXMan12:
Can you add a bit more details to the PR message beyond just the fixes?  It's useful when going through with `git log`.

DirectXMan12:
otherwise looks reasonable

sbueringer:
> Can you add a bit more details to the PR message beyond just the fixes? It's useful when going through with `git log`.

I added some more text here in the PR. The commit message itself already is: `add GetConfigWithContext to retrieve a config with a specific context`. Do I have to squash the `fix gopkg check` commit by myself or is that done by the automation?

DirectXMan12:
We don't automatically squash, no.  It's up to the submitter to produce a chain of logical commits (we don't enforce 1 commit per PR -- we just prefer that each commit stands on its own).

DirectXMan12:
/lgtm
/approve

thanks!

PR #497, opened 2019-06-24
:warning: split the webhook builder out as a separate builder.

mengqiy:
`build.go`=> `controller.go` + `webhook.go`
`build_test.go`=> `controller_test.go` + `webhook_test.go`

TODO: figure out if `For` method in webhook builder should be variadic?

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package builder
+
+import (
+	"net/http"
+	"net/url"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
+)
+
+// WebhookBuilder builds a Webhook.
+type WebhookBuilder struct {
+	apiType runtime.Object
+	gvk     schema.GroupVersionKind
+	mgr     manager.Manager
+	config  *rest.Config
+}
+
+func WebhookManagedBy(m manager.Manager) *WebhookBuilder {
+	return &WebhookBuilder{mgr: m}
+}
+
+// TODO(droot): update the GoDoc for conversion.
+
+// For takes a runtime.Object which should be a CR.
+// If the given object implements the admission.Defaulter interface, a MutatingWebhook will be wired for this type.
+// If the given object implements the admission.Validator interface, a ValidatingWebhook will be wired for this type.
+func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
+	blder.apiType = apiType
+	return blder
+}
+
+// Complete builds the webhook.
+func (blder *WebhookBuilder) Complete() error {
+	// Set the Config
+	if err := blder.loadRestConfig(); err != nil {
+		return err
+	}
+
+	// Set the Webook if needed
+	return blder.registerWebhooks()
+}
+
+func (blder *WebhookBuilder) loadRestConfig() error {
+	if blder.config != nil {
+		return nil
+	}
+	if blder.mgr != nil {
+		blder.config = blder.mgr.GetConfig()
+		return nil
+	}
+	var err error
+	blder.config, err = getConfig()
+	return err
+}
+
+func (blder *WebhookBuilder) registerWebhooks() error {
+	// Create webhook(s) for each type
+	var err error
+	blder.gvk, err = apiutil.GVKForObject(blder.apiType, blder.mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+
+	blder.registerDefaultingWebhook()
+	blder.registerValidatingWebhook()
+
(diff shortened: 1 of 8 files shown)


pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
+)
+
+// WebhookBuilder builds a Controller.
```
droot: s/Controller/Webhook

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+// If the passed in object has implemented the admission.Defaulter interface, a MutatingWebhook will be wired for this type.
+// If the passed in object has implemented the admission.Validator interface, a ValidatingWebhook will be wired for this type.
+// TODO: figure out if we support only one object or multiple objects
+func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
+	blder.apiType = apiType
+	return blder
+}
+
+// Complete builds the Application ControllerManagedBy.
+func (blder *WebhookBuilder) Complete() error {
```
DirectXMan12: do we really need this?  Can't we just do the equivalent in `For`?

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Set the Webook if needed
+	if err := blder.doWebhook(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (blder *WebhookBuilder) doConfig() error {
```
DirectXMan12: this should have a better name than `doConfig` -- perhaps `loadRestConfig`?

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	return &WebhookBuilder{mgr: m}
+}
+
+// For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
+// update events by *reconciling the object*.
+// This is the equivalent of calling
+// Watches(&source.Kind{Type: apiType}, &handler.EnqueueRequestForObject{})
+// If the passed in object has implemented the admission.Defaulter interface, a MutatingWebhook will be wired for this type.
+// If the passed in object has implemented the admission.Validator interface, a ValidatingWebhook will be wired for this type.
+// TODO: figure out if we support only one object or multiple objects
```
droot: I think we can simplify this comment now.

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	if blder.mgr != nil {
+		blder.config = blder.mgr.GetConfig()
+		return nil
+	}
+	var err error
+	blder.config, err = getConfig()
+	return err
+}
+
+func (blder *WebhookBuilder) doManager() error {
```
DirectXMan12: not needed -- was only in the old builder for legacy reasons (before we switched to `ManagedBy`)

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	if blder.mgr != nil {
+		return nil
+	}
+	var err error
+	blder.mgr, err = newManager(blder.config, manager.Options{})
+	return err
+}
+
+func (blder *WebhookBuilder) doWebhook() error {
```
DirectXMan12: ditto here -- maybe `registerRelatedWebhooks` or something?

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	// 2) convert it to the hub version
+	// 3) do defaulting
+	// 4) convert it back to the same bespoke version
+	// 5) calculate the JSON patch
+	//
+	// A potential workflow for validating webhook
+	// 1) a bespoke (non-hub) version comes in
+	// 2) convert it to the hub version
+	// 3) do validation
+	if defaulter, isDefaulter := blder.apiType.(admission.Defaulter); isDefaulter {
```
DirectXMan12: split these into helpers

DirectXMan12:
Need to split out conversion registration as well (in a separate PR) @droot

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+			if !blder.isAlreadyHandled(path) {
+				log.Info("Registering a validating webhook",
+					"GVK", gvk,
+					"path", path)
+				blder.mgr.GetWebhookServer().Register(path, vwh)
+			}
+		}
+	}
+
+	err = conversion.CheckConvertibility(blder.mgr.GetScheme(), blder.apiType)
```
droot: I will make a separate PR on top of this to enable conversion only when type implements convertible.

pkg/builder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	// 1) a bespoke (non-hub) version comes in
+	// 2) convert it to the hub version
+	// 3) do defaulting
+	// 4) convert it back to the same bespoke version
+	// 5) calculate the JSON patch
+	//
+	// A potential workflow for validating webhook
+	// 1) a bespoke (non-hub) version comes in
+	// 2) convert it to the hub version
+	// 3) do validation
```
droot: Do we need the above comments ?

mengqiy:
PTAL

pkg/builder/webhook.go
```diff
@@ -0,0 +1,146 @@
...
+	config  *rest.Config
+}
+
+func WebhookManagedBy(m manager.Manager) *WebhookBuilder {
+	return &WebhookBuilder{mgr: m}
+}
+
+// TODO(droot): update the GoDoc for conversion.
+
+// For takes a runtime.Object which should be a CR.
```
droot: Not necessarily this PR, but we should consider adding an example for this that shows in godoc. They are super useful.

mengqiy: Done!

mengqiy: https://github.com/kubernetes-sigs/controller-runtime/pull/497/files#diff-8b3e76472aacd449b1d1dda72eefbc32

mengqiy:
PTAL

pkg/builder/webhook.go
droot: I realized without having `Complete()` method, the build chain looks incomplete and its not clear whether we are done with the builder or not. So I would suggest not dropping this method. It also keeps it consistent with the controller builder.

mengqiy: We decided to bring `Complete` back.

PR #504, opened 2019-06-27
:sparkles: enable conversion webhook in webhook builder

droot:
Main changes:
 - Disable conversion webhook by default and enable only if a type
 implements convertible
 - Enhanced the test coverage to include spoke to spoke conversion


mengqiy:
And please also fix the linter failure.

PR #508, opened 2019-06-28
:bug: IsConvertible should not error on uninitialized struct.

mengqiy:
IsConvertible used to use obj.GetObjectKind().GroupVersionKind() to get
GVK which will not work if apiVersion and kind fields are not set. Now
it uses scheme.ObjectKinds(obj).


droot:
@mengqiy conversion_test.go needs to be run through gofmt I guess.

mengqiy:
fixed goimport linter error
adding label back per https://github.com/kubernetes-sigs/controller-runtime/pull/508#event-2448441756

PR #511, opened 2019-07-02
:sparkles Add env support for setting ErrorIfCRDPathMissing during testing

### LINKED ISSUE #481: envtest.Environment.Start() silently fails for missing CRD paths
There is no way to specify `envtest.CRDInstallOptions.ErrorIfPathMissing = true` when calling `envtest.Environment.Start()`. As a result, silent failure in the case of missing CRD directories is the default and only behavior when using `.Start()`.

Workaround:

Call `envtest.InstallCRDs()` directly:

```go
	t.Start()
	// ...
	if _, err := envtest.InstallCRDs(t.Config, envtest.CRDInstallOptions{
		Paths: []string{
			filepath.Join("..", "..", "..", "config", "crds"),
			filepath.Join("..", "..", "..", "third_party", "crds"),
		},
		ErrorIfPathMissing: true,
	}); err != nil {
```

connor1989:
A simpler interface for CRDInstallOptions.ErrorIfPathMissing. Adds ability to control, through the env, how errors are propagated when InstallCRDs is invoked with an invalid CRD path.

Closes #481

### DIFF
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -58,10 +58,28 @@ var _ = Describe("Test", func() {
 	})
 
 	Describe("InstallCRDs", func() {
+
+		It("should set ErrorIfCRDPathMissing through env settings", func(done Done) {
+
+			testEnv := &Environment{
+				ErrorIfCRDPathMissing: true,
+			}
+
+			crds, err = InstallCRDs(env.Config, CRDInstallOptions{
+				Paths:              []string{"fake"},
+				ErrorIfPathMissing: testEnv.ErrorIfCRDPathMissing,
+			})
+
+			Expect(err).To(HaveOccurred())
+
+			close(done)
+		})
+
 		It("should install the CRDs into the cluster", func(done Done) {
 
 			crds, err = InstallCRDs(env.Config, CRDInstallOptions{
-				Paths: []string{filepath.Join(".", "testdata")},
+				Paths:              []string{filepath.Join(".", "testdata")},
+				ErrorIfPathMissing: env.ErrorIfCRDPathMissing,
 			})
 			Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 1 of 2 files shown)


pkg/envtest/envtest_test.go
```diff
@@ -58,10 +58,28 @@ var _ = Describe("Test", func() {
 	})
 
 	Describe("InstallCRDs", func() {
+
+		It("should set ErrorIfCRDPathMissing through env settings", func(done Done) {
```
DirectXMan12: it's unclear to me what this is testing, since it doesn't seem to be testing that the setting gets automatically propogated from `Environment` to `InstallCRDs`

connor1989: Yes, it only validates the new option. This was to keep the other tests unchanged here.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/envtest_test.go#L155 

My other thought was to make this change directory in [BeforeSuite](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/envtest_suite_test.go#L37) and update all the tests accordingly.

```
	env = &Environment{
++   	    ErrorIfCRDPathMissing: true,
        }
	_, err := env.Start()
```
Any opposition to this?

pkg/envtest/envtest_test.go
```diff
@@ -58,10 +58,28 @@ var _ = Describe("Test", func() {
...
+
+			close(done)
+		})
+
 		It("should install the CRDs into the cluster", func(done Done) {
 
 			crds, err = InstallCRDs(env.Config, CRDInstallOptions{
-				Paths: []string{filepath.Join(".", "testdata")},
+				Paths:              []string{filepath.Join(".", "testdata")},
+				ErrorIfPathMissing: env.ErrorIfCRDPathMissing,
```
connor1989: @DirectXMan12 to your first point, although the change is minimal, I did make use of the new option here

DirectXMan12: right, but it's not plumbed through to any non-test logic, afaict

PR #512, opened 2019-07-02
:running: ff release-0.2 branch to master HEAD

mengqiy:
Prepare for next CR beta release.


mengqiy:
Travis apparently has a huge backlog: https://www.traviscistatus.com/incidents/g9xm3l02dknv
This is all from master HEAD which is already tested.
I'm not sure if we really need to wait for Travis to become normal.

droot:
> I'm not sure if we really need to wait for Travis to become normal.

We can merge  it manually. But lets run tests locally before cutting the release.

Issue #317, opened 2019-02-07
We should surface controller options when using builder


DirectXMan12:
It's hard to set controller options if you're using builder, which means you can't set stuff like max concurrent workers.

/kind feature
/priority important-soon

PR #520, opened 2019-07-17
:bug: surface controller options when using builder

awesomenix:
Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/317

- Currently there is no way to override MaxConcurrentReconciles using builder
- Provide a method to override controller options
- Another approach was to provide MaxConcurrentReconciles itself, but it isnt future proof incase additional options are added

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -47,6 +47,7 @@ type Builder struct {
 	watchRequest   []watchRequest
 	config         *rest.Config
 	ctrl           controller.Controller
+	ctrlOptions    controller.Options
 	name           string
 }
 
@@ -122,6 +123,12 @@ func (blder *Builder) WithManager(m manager.Manager) *Builder {
 // Defaults to the empty list.
 func (blder *Builder) WithEventFilter(p predicate.Predicate) *Builder {
 	blder.predicates = append(blder.predicates, p)
+	return blder
+}
+
+// WithOptions overrides the controller options use in doController. Defaults to empty.
+func (blder *Builder) WithOptions(options controller.Options) *Builder {
+	blder.ctrlOptions = options
 	return blder
 }
 
@@ -241,6 +248,8 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-	blder.ctrl, err = newController(name, blder.mgr, controller.Options{Reconciler: r})
+	ctrlOptions := blder.ctrlOptions
+	ctrlOptions.Reconciler = r
+	blder.ctrl, err = newController(name, blder.mgr, ctrlOptions)
 	return err
 }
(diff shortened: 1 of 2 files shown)


DirectXMan12:
For future reference: try to avoid github handles in PR messages, since they'll trigger notifications every time the commit moves around in a fork.

DirectXMan12:
hmm... can we future proof this so additional fields don't need extra work?

pkg/builder/controller.go
```diff
@@ -125,6 +126,14 @@ func (blder *Builder) WithEventFilter(p predicate.Predicate) *Builder {
 	return blder
 }
 
+// WithControllerOptions overrides the controller options use in doController. Defaults to empty.
+//
+// Deprecated: Use ControllerManagedBy(Manager) and this isn't needed.
+func (blder *Builder) WithControllerOptions(options controller.Options) *Builder {
```
DirectXMan12: `WithOptions`?

pkg/builder/controller.go
```diff
@@ -125,6 +126,14 @@ func (blder *Builder) WithEventFilter(p predicate.Predicate) *Builder {
 	return blder
 }
 
+// WithControllerOptions overrides the controller options use in doController. Defaults to empty.
+//
+// Deprecated: Use ControllerManagedBy(Manager) and this isn't needed.
```
DirectXMan12: why is this immediately deprecated?  It's unclear why this isn't needed with `ControllerManagedBy`...

pkg/builder/controller.go
```diff
@@ -241,6 +250,10 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-	blder.ctrl, err = newController(name, blder.mgr, controller.Options{Reconciler: r})
+	ctrlOptions := controller.Options{
+		Reconciler:              r,
+		MaxConcurrentReconciles: blder.ctrlOptions.MaxConcurrentReconciles,
```
DirectXMan12: probably should invert this so that we don't have to manually copy over fields

DirectXMan12: namely:

```go
ctrlOptions := builder.ctrlOptions
ctrlOptions.Reconciler = r
```

pkg/builder/controller.go
awesomenix: Sorry, copy pasted the comments and tried changing it and accidently left it there. Removed.

pkg/builder/controller.go
awesomenix: Done

awesomenix: Actually wasnt sure about WithOptions and WithControllerOptions since Manager too has options and it could be overriden in the future maybe?

pkg/builder/controller.go
awesomenix: Done.

PR #525, opened 2019-07-18
:warning: support delete validation in validator interface

awesomenix:
- Currently validator interface supports create and update
- Add support for delete as well, which could be a breaking change
- Users need to implement ValidateDelete since interface is updated

### DIFF
--- a/pkg/builder/webhook_test.go
+++ b/pkg/builder/webhook_test.go
@@ -269,6 +269,103 @@ var _ = Describe("application", func() {
 			Expect(w.Body).To(ContainSubstring(`"allowed":true`))
 			Expect(w.Body).To(ContainSubstring(`"code":200`))
 		})
+
+		It("should scaffold a validating webhook if the type implements the Validator interface to validate deletes", func() {
+			By("creating a controller manager")
+			m, err := manager.New(cfg, manager.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("registering the type in the Scheme")
+			builder := scheme.Builder{GroupVersion: testValidatorGVK.GroupVersion()}
+			builder.Register(&TestValidator{}, &TestValidatorList{})
+			err = builder.AddToScheme(m.GetScheme())
+			Expect(err).NotTo(HaveOccurred())
+
+			err = WebhookManagedBy(m).
+				For(&TestValidator{}).
+				Complete()
+			Expect(err).NotTo(HaveOccurred())
+			svr := m.GetWebhookServer()
+			Expect(svr).NotTo(BeNil())
+
+			reader := strings.NewReader(`{
+  "kind":"AdmissionReview",
+  "apiVersion":"admission.k8s.io/v1beta1",
+  "request":{
+    "uid":"07e52e8d-4513-11e9-a716-42010a800270",
+    "kind":{
+      "group":"",
+      "version":"v1",
+      "kind":"TestValidator"
+    },
+    "resource":{
+      "group":"",
+      "version":"v1",
+      "resource":"testvalidator"
+    },
+    "namespace":"default",
+    "operation":"DELETE",
+    "object":null,
+    "oldObject":{
+      "replica":1
+    }
+  }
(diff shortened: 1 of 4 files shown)


mengqiy:
@awesomenix I'd like to hear your use case more before we decide to make change to the interface.

mengqiy:
For breaking change, we should use `:warning:` I have updated the title.

awesomenix:
@mengqiy In our case we use annotations to validate deletes on certain resource so that none of the deletes are accidental. We used Validating Webhooks which support validating DELETE operations as well (this was easy to override in kubebuilder v1), hence requesting the same in v2 as well.

BTW love the easy scaffolding of webhooks v2, way less code than the original

mengqiy:
I'm thinking if we should support CONNECT operation. Trying to find if there is a common use case for it.

awesomenix:
was thinking if we should add support for CONNECT as well, but wasnt sure of any use case. Honestly i do not know when it will be used, during resource get?

mengqiy:
@awesomenix 
Re. DELETE operation, I saw this [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/00xx-admission-webhooks-to-ga.md#plumbing-existing-objects-to-delete-admission).

It seems upstream doesn't support it yet. No objects will be in the `AdmissionReview`.

pkg/webhook/admission/validator.go
```diff
@@ -89,5 +90,17 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 		}
 	}
 
+	if req.Operation == v1beta1.Delete {
+		err := h.decoder.Decode(req, obj)
```
mengqiy: It should be `err := h.decoder.DecodeRaw(req.OldObject, oldObj)` here

awesomenix:
> KEP

Thanks for the reference. I see that 1.15 has the fix https://github.com/kubernetes/kubernetes/pull/76346.

Updating PR to reflect the same.

pkg/builder/webhook_test.go
```diff
@@ -269,6 +269,104 @@ var _ = Describe("application", func() {
...
+    },
+    "namespace":"default",
+    "operation":"DELETE",
+    "object":null,
+    "oldObject":{
+      "replica":0
+    }
+  }
+}`)
+			By("sending a request to a mutating webhook path with correct request")
```
mengqiy: validating

pkg/builder/webhook_test.go
```diff
@@ -269,6 +269,104 @@ var _ = Describe("application", func() {
...
+			path = generateValidatePath(testValidatorGVK)
+			req = httptest.NewRequest("POST", "http://svc-name.svc-ns.svc"+path, reader)
+			req.Header.Add(http.CanonicalHeaderKey("Content-Type"), "application/json")
+			w = httptest.NewRecorder()
+			svr.WebhookMux.ServeHTTP(w, req)
+			Expect(w.Code).To(Equal(http.StatusOK))
+			Expect(w.Code).To(Equal(http.StatusOK))
```
mengqiy: Duplicate line

awesomenix:
@mengqiy thanks for the feedback. Updated PR.

pkg/webhook/admission/validator.go
```diff
@@ -89,5 +90,19 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 		}
 	}
 
+	if req.Operation == v1beta1.Delete {
+		// In reference to PR: https://github.com/kubernetes/kubernetes/pull/76346
+		// OldObject contains the object being deleted
+		err := h.decoder.DecodeRaw(req.OldObject, obj)
```
mengqiy: `req.OldObject` is an empty object when kube-apiserver is v1.14 or earlier.
`obj` will be a object with zero values everywhere.
`DecodeRaw` really should return an error when there is an empty object.
I created #529. We should merge that first.

PR #529, opened 2019-07-22
:sparkles: decoder now errors out when encountering an empty runtime.RawExtension

mengqiy:
When user tries to decode an object, it should always not be empty. This
error gives people hints when they running into this kind of issues. e.g.
kube-apiserver doesn't include the object in the admission review for
deletion events.

We needs this before landing https://github.com/kubernetes-sigs/controller-runtime/pull/525


mengqiy:
cc: @awesomenix

PR #530, opened 2019-07-22
:bug: stop using hardcoded temp dir

mengqiy:
We now use https://golang.org/pkg/os/#TempDir to get the os specific
temp dir. Otherwise, it won't work for Windows users.
We also switch to use the os-aware `Join` method.


Azadehkhojandi:
Thanks @mengqiy :)  as you suggested issue #531 created

Issue #534, opened 2019-07-23
expose newResourceLock in manager.Options so that controllers can use customized resourceLock


jzhoucliqr:
I have a use case to run a standalone aggregation-apiserver without native kube-apiserver. To enable leader selection for controllers, currently controller-runtime hardcode to use configmapLocks, which won't work for this case because configmap resource is not available in the aggregation-apiserver. 

If we expose newResourceLock outside of manager.Options, then controllers can their own customized resourcelocks, and leader selection can work without native kube-apiserver. 

This is a minor change, but will be very beneficial for standalone apiservers/controllers. 

What do you guys think? I can have a PR if we agree this is a valid use case.

PR #535, opened 2019-07-24
:sparkles: expose NewResourceLock in manager.Options 

jzhoucliqr:
This is to enable controllers to use customized resource locks for leader election. 

It is for the use case that a standalone aggregation-like apiserver running without native kube-apiserver, thus the corev1 resources are not available through apiserver REST API. 

With this, Mangers can use customized resourceLocks, without relying on corev1 resources.

Closes https://github.com/kubernetes-sigs/controller-runtime/issues/534

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -325,7 +325,6 @@ func setOptionsDefaults(options Options) Options {
 		options.newRecorderProvider = internalrecorder.NewProvider
 	}
 
-	// Allow NewResourceLock to be mocked
 	if options.NewResourceLock == nil {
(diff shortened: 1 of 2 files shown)


jzhoucliqr:
@DirectXMan12 @pwittrock Do you have time to take a quick look at this PR? Thanks.

DirectXMan12:
I'm not sure this is the interface we want to expose -- the particulars of the resource lock are mainly an implementation detail at this point, and this adds yet another piece of client-go to our public API, which increases the chance that things will arbitrarily break in the future.

That being said, I'll have to take a closer look to see what our options are.

DirectXMan12:
closing in favor of #600

PR #536, opened 2019-07-24
:warning: Functional Options Redux

### LINKED ISSUE #503: The XYZDryRunAll functions are frustratingly numerous
Our current "functional options" scheme works well for providing fluent options, but isn't so great when multiple function calls need to share a particular option -- see `PatchDryRunAll`, `UpdateDryRunAll`, and `CreateDryRunAll`.

AFAICT, the only way around this that provides type proper type checking & doesn't break backwards compatibility if we need to extend the set of things the option applies to looks like this:

```go
type Writer interface {
     Patch(ctx context.Context, obj runtime.Object, patch Patch, opts ...PatchOption) error 
     Create(ctx context.Context, obj runtime.Object, patch Patch, opts ...CreateOption) error
     Update(ctx context.Context, obj runtime.Object, patch Patch, opts ...UpdateOption) error
}

type PatchOption interface {
    ApplyToPatch(*PatchOptions)
}
type CreateOption interface {
    ApplyToCreate(*CreateOptions)
}
type UpdateOption interface {
    ApplyToUpdate(*UpdateOptions)
}

type DryRunAllOption struct {}
func (_ DryRunAllOption) ApplyToPatch(opts *PatchOptions) {
    opts.DryRun = []string{metav1.DryRunAll}
}
func (_ DryRunAllOption) ApplyToCreate(opts *CreateOptions) {
    opts.DryRun = []string{metav1.DryRunAll}
}
func (_ DryRunAllOption) ApplyToUpdate(opts *UpdateOptions) {
    opts.DryRun = []string{metav1.DryRunAll}
}
func (_ DryRunAllOption) ApplyToDeleteCollection(opts *DelCollOptions) {
    opts.DryRun = []string{metav1.DryRunAll}
}

var DryRunAll DryRunAllOption = DryRunAllOption{}
```

It yields the same usage:

```go
// Today:
err := cl.Update(ctx, obj, client.UpdateDryRunAll)

// Probably not the darkest timeline:
err := cl.Update(ctx, obj, client.DryRunAll)
```

It's slightly more verbose from a options-writing perspective, and puts another public type in our API surface, but means that we can safely add methods to that type as they come along w/o breaking compatibility (because neither function signature or interface signature need to change -- we just add methods to the struct's method set, which is a compatible operation).

It's something to think about.  If anyone else has any ideas, please write them here.

/kind feature

#### Rejected alternatives

* Similar to above, but `var DryRunAll interface{CreateOption; PatchOption} = dryRunAllOption`: hides the public type from our API surface, but changing that interface to add another (say, `UpdateOption`) causes a change in the type signature for functions that return one of these (e.g. `func GracePeriodSeconds(gp int64) interface{ListOption; DeleteOption}` --> `func GracePeriodSeconds(gp int64) interface{ListOption; DeleteOption; UpdateOption}`)

DirectXMan12:
This switches over our functional options to actually be interfaced-based.  Practically, this feels incredibly similar for users, to the point where I expect nobody will have to change their code (with the exception of Preconditions, which was a pointer).

The upsides are slightly more ergonomic options in some cases (like `client.MatchingLabels{"foo": "bar"}` instead of needing `client.MatchingLabels(map[string]string{"foo": "bar"})`, although the latter still works), as well as the removal of the need for separate variants for options that apply across different operations.

Technically, this is a breaking change, since there are type signature changes.

Fixes #503 (see that issue for more details)

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -58,7 +58,7 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 }
 
 // List implements Reader
-func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...client.ListOptionFunc) error {
+func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...client.ListOption) error {
 	gvk, err := apiutil.GVKForObject(out, ip.Scheme)
 	if err != nil {
 		return err
--- a/pkg/cache/informertest/fake_cache.go
+++ b/pkg/cache/informertest/fake_cache.go
@@ -136,6 +136,6 @@ func (c *FakeInformers) Get(ctx context.Context, key client.ObjectKey, obj runti
 }
 
 // List implements Cache
-func (c *FakeInformers) List(ctx context.Context, list runtime.Object, opts ...client.ListOptionFunc) error {
+func (c *FakeInformers) List(ctx context.Context, list runtime.Object, opts ...client.ListOption) error {
 	return nil
 }
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -88,7 +88,7 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out runtime.O
 }
 
 // List lists items out of the indexer and writes them to out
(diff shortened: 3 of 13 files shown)


rajathagasthya:
Ooh, this is going to affect #341 :-)

DirectXMan12:
should be a relatively minor tweak though.  lmk if you have any issues.

rajathagasthya:
No worries, I can rebase.

PR #542, opened 2019-07-26
:warning: Allow reconcilers to add watches dynamically

### LINKED ISSUE #540: Support adding new watches dynamically at runtime
This feature was added before (and requested in #246), but the updated Manager & Builder patterns don't support this any more because they take care of creating a Controller instead of having user-written code do it, and the Controller isn't accessible to your reconciler.

There was a discussion about this in Slack (https://kubernetes.slack.com/archives/CAR30FCJZ/p1561759175084100). I'm thinking we could modify the inject package like so (names TBD - I know they don't follow the current pattern 100% but we need to scope down the Controller interface to avoid import cycles):
```go
type Watcher interface {
  Watch(src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
}

type WatcherInjector interface {
  InjectWatcher(watcher Watcher)
}
```

Do you think something like this would work @DirectXMan12?

ncdc:
Allow reconcilers to add watches dynamically by getting a handle to the controller created by the builder via `Build` (changes the signature of `Build`).

Fixes #540

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -33,9 +32,7 @@ import (
 )
 
 // Supporting mocking out functions for testing
-var getConfig = config.GetConfig
 var newController = controller.New
-var newManager = manager.New
 var getGvk = apiutil.GVKForObject
 
 // Builder builds a Controller.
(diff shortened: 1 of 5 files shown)


DirectXMan12:
replied a bit on the linked issue.

DirectXMan12:
/hold

(see the linked issue)

ncdc:
@DirectXMan12 I've pushed up the changes we discussed yesterday. For the time being, I've kept in my original injector commit along with another commit that reverts it (just in case we decided to go back to it 😄). Once you're happy with the new changes, I'll rebase and remove the injector commits.

vincepri:
LGTM

pkg/builder/controller.go
```diff
@@ -138,26 +135,24 @@ func (blder *Builder) Named(name string) *Builder {
 // Complete builds the Application ControllerManagedBy.
 func (blder *Builder) Complete(r reconcile.Reconciler) error {
 	_, err := blder.Build(r)
-	return err
+	if err != nil {
```
mcristina422: This change isn't necessary. The error is already being handled

ncdc: Reverted

DirectXMan12:
can you just drop the `WithManager` and `SimpleApplication` (both of which are deprecated)?  If that's too much putzing with the tests, I can push an additional commit to do it.

otherwise LGTM

ncdc:
@DirectXMan12 I removed `WithManager` and `SimpleController`. PTAL.

DirectXMan12:
(added `:warning:` since this is now breaking)

PR #544, opened 2019-07-28
:sparkles: Expose CRDInstallOptions via envtest.Environment

### LINKED ISSUE #541: envtest: expose CRDInstallOptions via Environment
This is related to #481. Currently, there is no way to specify `CRDInstallOptions` using `envtest.Environment`, which means you need to workaround like this:

```
cfg, _ := t.Start()

_, err := envtest.InstallCRDs(cfg, envtest.CRDInstallOptions{
	Paths: []string{
		filepath.Join("..", "..", "..", "test", "integration", "crds"),
	},
	ErrorIfPathMissing: true,
})
```
But even with this workaround there is no way to control CRD install timeout/interval fields that are part of `CRDInstallOptions` since they are not exported (hitting an error with CRD install timing out and default timeout is too small):

https://github.com/kubernetes-sigs/controller-runtime/blob/59b131b7cd54d56ec74a66b084a53dd1c3e4843f/pkg/envtest/crd.go#L49-L53

So at the very least making those fields exported would be good. Another (better?) solution would be to just include `CRDInstallOptions` as a field of `Environment` struct and deprecate `CRDs` and `CRDDirectoryPaths` fields since they are already part of `CRDInstallOptions` struct. 

I'm happy to help with a PR if you think this is feasible!

rajathagasthya:
This change provides better control for installing CRDs in `envtest`.

Fixes #541

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -81,10 +81,17 @@ type Environment struct {
 	// loading.
 	Config *rest.Config
 
-	// CRDs is a list of CRDs to install
+	// CRDInstallOptions are the options for installing CRDs.
+	CRDInstallOptions CRDInstallOptions
+
+	// CRDs is a list of CRDs to install.
+	// If both this field and CRDs field in CRDInstallOptions are specified, the
+	// values are merged.
 	CRDs []*apiextensionsv1beta1.CustomResourceDefinition
 
 	// CRDDirectoryPaths is a list of paths containing CRD yaml or json configs.
+	// If both this field and Paths field in CRDInstallOptions are specified, the
+	// values are merged.
 	CRDDirectoryPaths []string
 
 	// UseExisting indicates that this environments should use an
@@ -203,10 +210,9 @@ func (te *Environment) Start() (*rest.Config, error) {
 	}
 
 	log.V(1).Info("installing CRDs")
-	_, err := InstallCRDs(te.Config, CRDInstallOptions{
-		Paths: te.CRDDirectoryPaths,
-		CRDs:  te.CRDs,
-	})
+	te.CRDInstallOptions.CRDs = mergeCRDs(te.CRDInstallOptions.CRDs, te.CRDs)
+	te.CRDInstallOptions.Paths = mergePaths(te.CRDInstallOptions.Paths, te.CRDDirectoryPaths)
+	_, err := InstallCRDs(te.Config, te.CRDInstallOptions)
 	return te.Config, err
 }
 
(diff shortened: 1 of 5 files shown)


pkg/envtest/server.go
```diff
@@ -81,12 +81,23 @@ type Environment struct {
...
+	// value this ignored.
+	//
+	// Deprecated: Use CRDInstallOptions
 	CRDs []*apiextensionsv1beta1.CustomResourceDefinition
 
 	// CRDDirectoryPaths is a list of paths containing CRD yaml or json configs.
+	// If both this field and Paths field in CRDInstallOptions are specified, this
+	// value this ignored.
+	//
+	// Deprecated: Use CRDInstallOptions
```
DirectXMan12: let's not deprecate these -- they're way more convenient when constructing the object directly -- constructing embedded objects is a pain.

pkg/envtest/server.go
```diff
@@ -203,10 +214,13 @@ func (te *Environment) Start() (*rest.Config, error) {
...
 	log.V(1).Info("installing CRDs")
-	_, err := InstallCRDs(te.Config, CRDInstallOptions{
-		Paths: te.CRDDirectoryPaths,
-		CRDs:  te.CRDs,
-	})
+	if len(te.CRDInstallOptions.CRDs) == 0 {
+		te.CRDInstallOptions.CRDs = te.CRDs
+	}
+	if len(te.CRDInstallOptions.Paths) == 0 {
+		te.CRDInstallOptions.Paths = te.CRDDirectoryPaths
```
DirectXMan12: can't we just append here?

pkg/envtest/server.go
rajathagasthya: Sounds reasonable.

pkg/envtest/server.go
rajathagasthya: Sure, will do. Do we need to filter out "duplicate" paths and CRDs though? In case they same values are passed both directly and via `CRDInstallOptions`?

DirectXMan12: yeah, probably.

rajathagasthya: Just to clarify, "duplicate" CRDs mean they contain the same name right? I don't have to DeepEqual or anything on those CRD objects?

DirectXMan12: identical name is sufficient.  Duplicate name with different contents is an error, so we don't need to try and support that.

DirectXMan12:
let's see if we can re-trigger the tests

/retest

Issue #485, opened 2019-06-11
Implement helpers functions for using finalizers


droot:
Code for implementing finalizers can be captured in couple of helpers functions in controller-runtime. Helpers functions can help in examining, adding and removing finalizers as suggested in [the PR](https://github.com/kubernetes-sigs/kubebuilder/pull/778/files#r292646664)

negz:
We have a few finalizer helper functions in our controller-runtime project Crossplane - I'd be happy to submit a PR to add them to controller-runtime.

```go
// AddFinalizer to the supplied Kubernetes object's metadata.
func AddFinalizer(o metav1.Object, finalizer string) {
	f := o.GetFinalizers()
	for _, e := range f {
		if e == finalizer {
			return
		}
	}
	o.SetFinalizers(append(f, finalizer))
}

// RemoveFinalizer from the supplied Kubernetes object's metadata.
func RemoveFinalizer(o metav1.Object, finalizer string) {
	f := o.GetFinalizers()
	for i, e := range f {
		if e == finalizer {
			f = append(f[:i], f[i+1:]...)
		}
	}
	o.SetFinalizers(f)
}
```

droot:
@negz Thanks for the offer. Pl. do. But API should take `runtime.Object` as input instead of `metav1.Object` because thats the abstraction all other CR's APIs operate on. Also, pl. take a look at @DirectXMan12 's suggestion in the linked PR.

negz:
> API should take runtime.Object as input instead of metav1.Object because thats the abstraction all other CR's APIs operate on.

I believe taking `runtime.Object` would require us to attempt to transform the object via `metav1.Accessor()` inside the utility function, which means we'd have to return an error and thus the caller would be stuck handling (or ignoring) an error every time they wanted to set a finalizer.

Is it actually common for controllers to be treating objects under reconciliation as `runtime.Object`? Speaking personally, I'd opt not to use a utility function that took `runtime.Object` simply to set finalizers because in 99% of cases I'm reconciling an object that I know satisfies `metav1.Object`.

DirectXMan12:
The main reason to use runtime.Object is not to force people to type assert or call `meta.Accessor` themselves.  We pass around `runtime.Object` because it's what most of the low-level core kube apimachinery interfaces work with.  It's mostly a side-effect of us consuming the low-level apimachinery.  That being said, it's fairly uncommon for someone *not* to have something that implements `metav1.Object`, so it may be ok here.

[...]

PR #545, opened 2019-07-29
✨ implement helpers for add/remove finalizer

alexeldeib:
https://github.com/kubernetes-sigs/controller-runtime/issues/485

I don't have a particular preference between the runtime.Object and metav1.Object in the signature. I see both reasons from the linked issue as valid. Figured i'd open as such and ask for feedback.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -168,3 +169,47 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
 
 // MutateFn is a function which mutates the existing object into it's desired state.
 type MutateFn func() error
+
+// AddFinalizer accepts a metav1 object and adds the provided finalizer if not present.
+func AddFinalizer(o metav1.Object, finalizer string) {
+	f := o.GetFinalizers()
+	for _, e := range f {
+		if e == finalizer {
+			return
+		}
+	}
+	o.SetFinalizers(append(f, finalizer))
+}
+
+// AddFinalizerWithError tries to convert a runtime object to a metav1 object and add the provided finalizer.
+// It returns an error if the provided object cannot provide an accessor.
+func AddFinalizerWithError(o runtime.Object, finalizer string) error {
+	m, err := meta.Accessor(o)
+	if err != nil {
+		return err
+	}
+	AddFinalizer(m, finalizer)
+	return nil
+}
+
+// RemoveFinalizer accepts a metav1 object and removes the provided finalizer if present.
+func RemoveFinalizer(o metav1.Object, finalizer string) {
+	f := o.GetFinalizers()
+	for i, e := range f {
+		if e == finalizer {
+			f = append(f[:i], f[i+1:]...)
+		}
+	}
+	o.SetFinalizers(f)
+}
+
+// RemoveFinalizerWithError tries to convert a runtime object to a metav1 object and remove the provided finalizer.
+// It returns an error if the provided object cannot provide an accessor.
+func RemoveFinalizerWithError(o runtime.Object, finalizer string) error {
+	m, err := meta.Accessor(o)
+	if err != nil {
+		return err
+	}
+	RemoveFinalizer(m, finalizer)
+	return nil
+}
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -265,9 +265,62 @@ var _ = Describe("Controllerutil", func() {
 			Expect(err).To(HaveOccurred())
 		})
 	})
+
+	Describe("Finalizers", func() {
+		var obj runtime.Object = &errRuntimeObj{}
+		var deploy *appsv1.Deployment
+
+		Describe("AddFinalizerWithError", func() {
+			It("should return an error if object can't provide accessor", func() {
+				Expect(controllerutil.AddFinalizerWithError(obj, testFinalizer)).To(HaveOccurred())
+			})
+		})
+
+		Describe("RemoveFinalizerWithError", func() {
+			It("should return an error if object can't provide accessor", func() {
+				Expect(controllerutil.RemoveFinalizerWithError(obj, testFinalizer)).To(HaveOccurred())
+			})
+		})
+
+		Describe("AddFinalizer", func() {
+			deploy = &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Finalizers: []string{},
+				},
+			}
+
+			It("should add the finalizer when not present", func() {
+				controllerutil.AddFinalizer(deploy, testFinalizer)
+				Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{testFinalizer}))
+			})
+
+			It("should not add the finalizer when already present", func() {
+				controllerutil.AddFinalizer(deploy, testFinalizer)
+				Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{testFinalizer}))
+			})
+		})
+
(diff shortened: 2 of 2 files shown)


alexeldeib:
It actually might make sense for the `WithError` version to call `client.Update()` too. that way, you get one error to handle, or you can do the granular management on your own metav1 object, or implement your own layer with runtime object.

alexeldeib:
dunno why I assigned @pwittrock, blindly following the robot I guess

/unassign @pwittrock

DirectXMan12:
squash the fixes, otherwise lgtm.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +169,47 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
...
+	AddFinalizer(m, finalizer)
+	return nil
+}
+
+// RemoveFinalizer accepts a metav1 object and removes the provided finalizer if present.
+func RemoveFinalizer(o metav1.Object, finalizer string) {
+	f := o.GetFinalizers()
+	for i, e := range f {
+		if e == finalizer {
+			f = append(f[:i], f[i+1:]...)
```
guymguym: Theoretically this is not a stable iteration since you change the slice during the iteration but the index keeps increasing - see this code for example:
```
package main

import "fmt"

func main() {
	a := []int{1, 2, 1, 2}
	for i, v := range a {
		fmt.Printf("a = %+v\n", a)
		fmt.Printf("i = %d\n", i)
		fmt.Printf("v = %d\n", v)
		fmt.Printf("a[i] = %d\n", a[i])
		if a[i] == 1 {
			a = append(a[:i], a[i+1:]...)
		}
	}
}
```
And the output:
```
a = [1 2 1 2]
i = 0
v = 1
a[i] = 1
a = [2 1 2]
i = 1
v = 1
a[i] = 1
a = [2 2]
i = 2
v = 2
panic: runtime error: index out of range
exit status 2
```

How about using the filter example from here: 
https://github.com/golang/go/wiki/SliceTricks#filter-in-place

alexeldeib: dropped a `return`, but yeah that works too

alexeldeib: will get it when I squash

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +169,48 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
...
+		return err
+	}
+	AddFinalizer(m, finalizer)
+	return nil
+}
+
+// RemoveFinalizer accepts a metav1 object and removes the provided finalizer if present.
+func RemoveFinalizer(o metav1.Object, finalizer string) {
+	f := o.GetFinalizers()
+	for i, e := range f {
```
alvaroaleman: Nit: This is nicer with `k8s.io/apimachinery/pkg/util/sets.String`

alexeldeib: found this today coincidentally. can swap it out 👍

alexeldeib: better?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +170,39 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
 
 // MutateFn is a function which mutates the existing object into it's desired state.
 type MutateFn func() error
+
+// AddFinalizer accepts a metav1 object and adds the provided finalizer if not present.
+func AddFinalizer(o metav1.Object, finalizer string) {
+	finalizers := sets.NewString(o.GetFinalizers()...)
```
DirectXMan12: I hate to go back an forth here, but this was actually probably fine with the linear search -- namely, the list of finalizers is almost always tiny, which means the linear search is probably fast, and we avoid what's actually probably more "work" plus an allocation building the set (plus, we don't re-order the list of finalizers).  FWIW, k8s uses a normal linear search like the old code -- look for NeedsLBFinalizer and the finalizer code in the volume controller.

pkg/util/slice in apimachinery has a RemoveString and ContainsString that we can use to still make this simple to look at.

alexeldeib: makes sense. I noticed that most controller implementations (including the function you mentioned) make a DeepCopy to avoid mutating the informer cache. Is that something I should add here?

alexeldeib: err, i'm also not seeing those methods (or a pkg/util/slice in apimachinery). I think you mean these? https://github.com/kubernetes/kubernetes/blob/4d9e4f0b458c746f43836f558ebe61ec68bb7133/pkg/util/slice/slice.go

Happy to slice those out of k/k into apimachinery or somewhere else (pun heavily intended).

alexeldeib: I don't really understand the proper usage of controllers w.r.t informer cache because I saw this a lot pre-controller runtime, but don't see it as much in controllers using CR. I thought it might be under the hood somehow in CR (copying the object for you to avoid mutating the cache), but I didn't find it in internal/controller or in the manager code for injecting the cache. 

Would love to understand more deeply to make sure I'm doing the right thing.

DirectXMan12: > I noticed most controller implementations make a DeepCopy

We automatically deepcopy before returning the object to you from the cache to avoid these sorts of issues: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L75

> I  think you mean these?

Yeah, those are the ones that I meant :-)

> CR (copying the object for you to avoid mutating the cache), but I didn't find it in internal/controller or in the manager code for injecting the cache.

See above :-)

alexeldeib:
I think...this should work now?

/test

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +169,61 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
 
 // MutateFn is a function which mutates the existing object into it's desired state.
 type MutateFn func() error
+
+// AddFinalizer accepts a metav1 object and adds the provided finalizer if not present.
+func AddFinalizer(o metav1.Object, finalizer string) {
+	newFinalizers := AddString(o.GetFinalizers(), finalizer)
+	o.SetFinalizers(append(newFinalizers, finalizer))
```
DirectXMan12: huh?  Doesn't this add the finalizer twice?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +169,61 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
...
+	if err != nil {
+		return err
+	}
+	RemoveFinalizer(m, finalizer)
+	return nil
+}
+
+// AddString returns a []string with s appended if it is not already found in the provided slice.
+func AddString(slice []string, s string) []string {
+	newSlice := make([]string, 0)
```
DirectXMan12: No reason to make a copy here:

```go
for i, item := range slice {
  if item == s {
     return slice
  }
}
return append(slice, s)
```

DirectXMan12: also make these private

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,3 +169,61 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
...
+		if item == s {
+			return append(newSlice, slice[i+1:]...)
+		}
+		newSlice = append(newSlice, item)
+	}
+	return append(slice, s)
+}
+
+// RemoveString returns a newly created []string that contains all items from slice that are not equal to s.
+func RemoveString(slice []string, s string) []string {
```
DirectXMan12: make this private too

pkg/controller/controllerutil/controllerutil.go
alexeldeib: i was wondering if there was a reason the k/k version made a copy, I stared at it a bit and decided no => already reduce to what you posted

pkg/controller/controllerutil/controllerutil.go
alexeldeib: fixed

pkg/controller/controllerutil/controllerutil.go
alexeldeib: pushed early, fixed

PR #549, opened 2019-08-04
:sparkles: Pass HTTP request Context through to admission Webhook handlers

seh:
Rather than synthesizing a fresh `context.Context` for each admission Webhook request, instead pass through a `Context` obtained from the original `http.Request` passed in to the `http.Handler`. This way, we can wrap middleware `http.Handler`s around the `webhook.Admission` handlers, such as those that establish tracing spans, instrumentation, or logging context, and have our `admission.Handler`s continue to use that same context.

Note that if an `http.Request` had no `Context` bound explicitly with `http.WithContext`, [the `Request.Context` method returns `context.Background`](https://golang.org/src/net/http/request.go?s=12215:12258#L339).

### DIFF
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -92,6 +92,10 @@ func (r *Response) Complete(req Request) error {
 
 // Handler can handle an AdmissionRequest.
 type Handler interface {
+	// Handle yields a response to an AdmissionRequest.
+	//
+	// The supplied context is extracted from the received http.Request, allowing wrapping
+	// http.Handlers to inject values into and control cancelation of downstream request processing.
 	Handle(context.Context, Request) Response
 }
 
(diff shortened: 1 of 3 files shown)


DirectXMan12:
good catch, thanks!

/lgtm
/approve

pkg/webhook/admission/webhook.go
```diff
@@ -92,6 +92,10 @@ func (r *Response) Complete(req Request) error {
 
 // Handler can handle an AdmissionRequest.
 type Handler interface {
+	// Handle yields a response to an AdmissionRequest.
+	//
+	// The supplied context is extracted from the received http.Request, allowing wrapping
+	// http.Handlers to inject values into and control cancelation of downstream request processing.
```
seh: In proxy server terms, the inner handlers are analogous to "upstream" servers. Given that, if my use of "downstream" is confusing, please let me know and I can try again.

DirectXMan12: eh, I think it's clear enough if you're vaguely familiar with context.  We'll see if anyone complains :-)

PR #553, opened 2019-08-07
✨ Add predicate for Generation change on update event

hasbro17:
Adds a `GenerationChangedPredicate` predicate similar to the `ResourceVersionChangedPredicate` that will filter out update events if an object's `metadata.generation` is unchanged.

This predicate is needed since the CRD status subresource is more commonly enabled now which allows users to filter out updates that have no spec changes.

Since this is a common ask, this predicate is better suited as a helper in the predicate library rather than having users write it themselves or being somewhere [downstream in the operator-sdk](https://github.com/operator-framework/operator-sdk/blob/6c7039c373244e8e485edd447adbf6b86bffeda6/pkg/predicate/predicate.go).

**Side note:** For CRD objects the Generation increment is skipped on metatdata updates as expected. But this behavior isn't consistent for other APIs e.g Deployments where annotation updates will cause a Generation bump. https://github.com/kubernetes/kubernetes/issues/67428#issuecomment-413493025
I wasn't sure if this is worth pointing out in the comments though.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -40,6 +40,7 @@ type Predicate interface {
 
 var _ Predicate = Funcs{}
 var _ Predicate = ResourceVersionChangedPredicate{}
+var _ Predicate = GenerationChangedPredicate{}
 
 // Funcs is a function that implements Predicate.
 type Funcs struct {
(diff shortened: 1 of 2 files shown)


pkg/predicate/predicate.go
```diff
@@ -116,3 +116,32 @@ func (ResourceVersionChangedPredicate) Update(e event.UpdateEvent) bool {
 	}
 	return true
 }
+
+// GenerationChangedPredicate implements a default update predicate function on generation change
```
droot: It will be good to add the details that you specified in the PR description especially mentioning `status` sub-resource. That sort of explains `when` this predicate is useful.

pkg/predicate/predicate_test.go
```diff
@@ -308,4 +308,134 @@ var _ = Describe("Predicate", func() {
...
+						Name:       "baz",
+						Namespace:  "biz",
+						Generation: 1,
+					}}
+
+				failEvnt := event.UpdateEvent{
+					MetaNew:   new.GetObjectMeta(),
+					ObjectNew: new,
+				}
+				Expect(instance.Create(event.CreateEvent{})).Should(BeTrue())
```
droot: `Expect` notation should be used with `To` instead of `Should` https://onsi.github.io/gomega/#getting-gomega to be idiomatic. I think our older code has this problem, but we can fix this for newer tests.

`Expect(instance.Create(event.CreateEvent{})).To(BeTrue())`

This comment applies to all the `Expect` notations in the tests below. [avoiding typing ditto :)]

pkg/predicate/predicate.go
shawn-hurley: I would also like to see an update to the examples to show how to use the predicate? WDYT?

pkg/predicate/predicate.go
```diff
@@ -116,3 +116,32 @@ func (ResourceVersionChangedPredicate) Update(e event.UpdateEvent) bool {
...
+	}
+	if e.MetaNew == nil {
+		log.Error(nil, "Update event has no new metadata", "event", e)
+		return false
+	}
+	if e.MetaNew.GetGeneration() == e.MetaOld.GetGeneration() {
+		return false
+	}
+	return true
+}
```
shawn-hurley: I would also like to see us add a line similar to [this](https://github.com/kubernetes-sigs/controller-runtime/pull/553/files#diff-025d468fb006e180e278f2e9863fc7dfR42)

pkg/predicate/predicate.go
hasbro17: Did you mean showing how to use one of the pre-defined predicates as an argument to the controller's Watch()?
I guess I could add that since I don't see it in any of the other examples:
https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/controller#ex-Controller

There's also an existing example in the predicate godoc, outlining how to define a new predicate.
Interestingly enough it's using generation change as the example.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/predicate/example_test.go#L27-L33
https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/predicate#ex-Funcs

I could update that to be more complete like the one written here. Or leave that as it is if it's meant to be just a short example.

@droot Let me know what you think.

DirectXMan12: I'd prefer to keep that example as-is.

In addition to a "why", I think this needs a "caveats" section -- like "this means you won't self-correct when someone else wipes your status".

hasbro17:
Updated the godocs for `GenerationChangedPredicate` to explain how the Generation field works and the purpose of the predicate, along with a caveats section on the gotchas with using the predicate.
PTAL to see if the comments need to be more elaborate or concise. Or if there are any other caveats.

@shawn-hurley Based on the comments above I haven't updated the existing example. Let me know if you think we still need another example.

PR #559, opened 2019-08-09
✨ ZapLoggerTo points to zap.LoggerTo

### LINKED ISSUE #558: Probably a typo in runtime/log
In `runtime/log` we have 2 variables [`ZapLogger`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/log/log.go#L35) and [`ZapLoggerTo`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/log/log.go#L40). 

They both point to the same function `zap.Logger`. In case of  `ZapLoggerTo` it probably should point to `zap.LoggerTo`.

GrigoriyMikhalkin:
In `runtime/log` we have 2 variables [`ZapLogger`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/log/log.go#L35) and [`ZapLoggerTo`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/log/log.go#L40). 

They both point to the same function `zap.Logger`. In case of  `ZapLoggerTo` it probably should point to `zap.LoggerTo`.

Fixes #558

### DIFF
--- a/pkg/runtime/log/log.go
+++ b/pkg/runtime/log/log.go
@@ -37,7 +37,7 @@ var (
 	// ZapLoggerTo returns a new Logger implementation using Zap which logs
 	// to the given destination, instead of stderr.  It otherise behaves like
 	// ZapLogger.
-	ZapLoggerTo = zap.Logger
+	ZapLoggerTo = zap.LoggerTo
 
 	// SetLogger sets a concrete logging implementation for all deferred Loggers.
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Yep, that looks like a typo.

/lgtm
/approve

PR #560, opened 2019-08-10
:sparkles: Allow fine-grained configuration of log/zap

### LINKED ISSUE #442: log/zap: Allow configuring formatter independently of `debug` option

alvaroaleman:
This PR changes the logger in `pkg/log/zap` to additionally allow configuring the Encoder and the LogLevel. This is done using the functional style opts pattern.

Backwards-compatibility is kept by keeping the existing public interface and extending it with a `New` to get a `logr.Logger` and a `NewRaw` to get a `*zap.Logger`.

Fixes #442
Closes #467 

/assign @DirectXMan12


pkg/log/zap/zap.go
```diff
@@ -29,6 +29,11 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// New returns a brand new Logger configured with Opts
```
DirectXMan12: please note that this (and `NewRaw`) use `KubeAwareEncoder`

DirectXMan12:
also need a comment explaining why we have `LoggerTo` and `New`, and what the differences are.

alvaroaleman:
> Should we have options for the comment settings?

What do you mean by "comment settings"?

> also need a comment explaining why we have LoggerTo and New, and what the differences are.

The reason I left those is just compatibility, other than that I do not see a good reason for them to exist. Do you know one? Else I would just add that as a comment.

DirectXMan12:
lol, sorry, "common settings" :-P.

Compatibility is the big reason for the others existing.  You could also just add

```go
// <comment>
//
// Deprecated: Use New(Development) instead
```
or similar

alvaroaleman:
@DirectXMan12 I've updated and expanded the godocs and also added the possibility to configure the log level for stracktraces. The only common thing left now that can not be configured is enablement of the `KubeAwareDecoder`, reasoning was that this project is very much intended to be used in conjunction with Kubernetes APIs so there is no good reason to not want that or WDYT?

DirectXMan12:
Seems reasonable to just always enable it

alvaroaleman:
I _think_ this is good to go, ptal

PR #566, opened 2019-08-15
:running: Add test for typed Patch after Status.Patch

vincepri:

This PR adds a simple test to make sure that running Patch after Status.Patch.


pkg/client/client_test.go
```diff
@@ -1178,6 +1178,34 @@ var _ = Describe("Client", func() {
 				close(done)
 			})
 
+			It("should patch correctly after a status patch", func(done Done) {
```
DirectXMan12: err... what does "correctly" mean here?  I feel like I'm missing some context for this.

vincepri: Apologies, we've noticed that the other way around, doing Status.Patch after Patch makes the object lose status information. I added this test to ensure that, going forward, this behavior is correct.

I can also add the opposite (failing) test and mark it to be Skipped if you prefer

DirectXMan12: ah, I think you should just elaborate a bit in the test description

vincepri: Sounds good, I reworded a little yesterday but can add more comments.

FYI, we ended up adding a package to wrap the Patch logic for us https://github.com/kubernetes-sigs/cluster-api/pull/1275

DirectXMan12:
hmm... the problem is that what you're describing is kinda intended behavior, but is a bit of a footgun :-/.  I think it kinda makes sense that if you update status, you only expect the status bits to be filled in from the server.

vincepri:
Definitely agree, do you think it might be worth discussing in a separate issue or should we leave as it is and maybe add only some docs?

DirectXMan12:
let's discuss in a separate issue

DirectXMan12:
I think the conclusion was that we could close this

PR #569, opened 2019-08-19
should support webhook certDir args

### LINKED ISSUE #568: we should support webhook certDir args in cm args
now we only support default certDir:
https://github.com/kubernetes-sigs/controller-runtime/blob/d7467fcf9e1c5817231f2768ffd626a244f2221b/pkg/webhook/server.go#L88

xrmzju:
add webhook certDir args. 
fixes #568

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -149,6 +149,10 @@ type Options struct {
 	// It is used to set webhook.Server.Host.
 	Host string
 
+	// CertDir is the directory that contains the server key and certificate.
+	// if not set, webhook server would look up the server key and certificate in
+	// {TempDir}/k8s-webhook-server/serving-certs
+	CertDir string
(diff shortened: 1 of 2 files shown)


xrmzju:
I signed it

pkg/manager/manager.go
```diff
@@ -149,6 +149,8 @@ type Options struct {
 	// It is used to set webhook.Server.Host.
 	Host string
 
+	// CertDir is the directory that contains the server key and certificate.
```
mengqiy: We should make the GoDoc more clear that the CertDir is optional. If not set, default will be used.
Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/7b08fd6a95a05eec0bc9c62bb5e77c11bac605c4/pkg/webhook/server.go#L87-L89

xrmzju: done

PR #578, opened 2019-08-26
:sparkles: pkg/client/options.go: MatchingSelector can add a fields.Selector to ListOptions

### LINKED ISSUE #576: client.MatchingFields should be a fields.Selector
As far as I know, there's no way to parse a field selector string, ex. from [`v1.ListOptions.FieldSelector`](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions) into a [`fields.Set`](https://godoc.org/k8s.io/apimachinery/pkg/fields#Set), making it hard to create [`client.MatchingFields`](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#MatchingFields) from a string. You could directly construct the Set (which is just a map) by manually splitting out keys and values, but the accepted way to do this is to directly parse the string into a [`fields.Selector`](https://godoc.org/k8s.io/apimachinery/pkg/fields#Selector) using [`fields.ParseSelector()`](https://godoc.org/k8s.io/apimachinery/pkg/fields#ParseSelector). You can also create a Selector from a Set using [`fields.SelectorFromSet()`](https://godoc.org/k8s.io/apimachinery/pkg/fields#SelectorFromSet), so a map can still be used directly without much extra effort. Additionally a Selector has more matching options than a Set, which only checks for equality.

I suggest either changing the underlying type of `client.MatchingFields` to `fields.Selector`, or creating a separate `client.MatchingFieldsSelector` that implements `client.ListOption` and `client.DeleteAllOfOption` with an underlying type of `fields.Selector`.

/cc @DirectXMan12 @mengqiy

estroz:
`MatchingSelector` provides the full [`fields.Selector`](https://godoc.org/k8s.io/apimachinery/pkg/fields#Selector) API for listing/deleting resources. See #576 for more details.

Closes #576

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -313,6 +313,21 @@ func (m MatchingLabels) ApplyToDeleteAllOf(opts *DeleteAllOfOptions) {
 	m.ApplyToList(&opts.ListOptions)
 }
 
+// MatchingLabelsSelector filters the list/delete operation on the given label
+// selector (or index in the case of cached lists). A struct is used because
+// labels.Selector is an interface, which cannot be aliased.
+type MatchingLabelsSelector struct {
+	labels.Selector
+}
+
+func (m MatchingLabelsSelector) ApplyToList(opts *ListOptions) {
+	opts.LabelSelector = m
+}
+
+func (m MatchingLabelsSelector) ApplyToDeleteAllOf(opts *DeleteAllOfOptions) {
+	m.ApplyToList(&opts.ListOptions)
+}
+
 // MatchingField filters the list operation on the given field selector
 // (or index in the case of cached lists).
 //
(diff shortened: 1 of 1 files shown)


pkg/client/options.go
```diff
@@ -335,6 +335,20 @@ func (m MatchingFields) ApplyToDeleteAllOf(opts *DeleteAllOfOptions) {
 	m.ApplyToList(&opts.ListOptions)
 }
 
+// MatchingSelector filters the list/delete operation on the given field
+// selector (or index in the case of cached lists).
+type MatchingSelector struct {
```
DirectXMan12: Should just be able to use a type alias here:

```suggestion
type MatchingSelector fields.Selector
```

joelanford: I had the same thought. Turns out `fields.Selector` is an interface, and Go doesn't allow interface types to be used as method receivers.

estroz: I can make the type:
```go
type MatchingFieldsSelector struct {
	fields.Selector
}
```
to basically give it the same calling UX (not exactly the same of course).

DirectXMan12: ah, yeah, forgot about that.  Good call.

DirectXMan12: maybe add a note

DirectXMan12:
`MatchingLabelSelector` SGTM

PR #580, opened 2019-09-01
[WIP] :sparkles: Adding timeout while waiting for cache to sync

### LINKED ISSUE #562: Create timeout and error when informer set up in the cache is hanging
It may be desirable to have a timeout when the informer is established so that this does not hang the operator and instead can send back an error. This is helpful in diagnosing permission issues and I think is a better user experience. 

/kind feature

shawn-hurley:
fixes #562 

Adding a timeout based on the context while waiting for the cache to sync. The default value is `30 seconds` as I didn't have a better guess on what it should be. 

Discussion:
1. What should the correct default be?
2. Should we create a specific error that one could check, I would think we would ultimately want [this](https://godoc.org/k8s.io/apimachinery/pkg/api/errors#IsTimeout) to return true.
3. Need help adding a test for this. All I can think to do is change the user that is making the calls to a user that does not have access to a resource. This seemed daunting, is there some other way to trigger the cache sync failure?

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -50,7 +51,11 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 		return err
 	}
 
-	cache, err := ip.InformersMap.Get(gvk, out)
+	ctx, cancelFunc := addTimeout(ctx)
+	if cancelFunc != nil {
+		defer cancelFunc()
+	}
+	cache, err := ip.InformersMap.Get(ctx, gvk, out)
 	if err != nil {
 		return err
 	}
@@ -90,7 +95,11 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
 		}
 	}
 
-	cache, err := ip.InformersMap.Get(gvk, cacheTypeObj)
+	ctx, cancelFunc := addTimeout(ctx)
+	if cancelFunc != nil {
+		defer cancelFunc()
+	}
+	cache, err := ip.InformersMap.Get(ctx, gvk, cacheTypeObj)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/cache/informer_cache.go
```diff
@@ -90,35 +95,59 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
...
 	if err != nil {
 		return err
 	}
 
 	return cache.Reader.List(ctx, out, opts...)
 }
 
+// addTimeout adds a default 30 second timeout to a child contxt
+// if one does not exist.
+func addTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
```
DirectXMan12: maybe change this to be `addTimeoutIfUnset` or something.

Also, do we want a default timeout?  I'm unsure

shawn-hurley: > Also, do we want a default timeout? I'm unsure

That could also be an option it might be easier to punt the decision to the implementer as they probably know their operator and what it would expect better. 

On the other hand, I think a case where we are never going to sync permissions error especially and will hang a worker is a bug that should be fixed. 

I am torn :)

DirectXMan12: > That could also be an option it might be easier to punt the decision to the implementer as they probably know their operator and what it would expect better.

That's my opinion -- defaults might fail weirdly in certain conditions, but hanging forever is bad too.  The problem is that with defaults, there's not way to say "no, never time out" except to go astronomically large.  Let's punt on the default for now (do it in a separate PR) so we can at least get the *capability* in and unbreak people that want to use this.

Relatedly: what's the practical timeout on client->server requests?  If you just eat everything after the handshake on a direct client call, how long will we wait?  Forever?  If not, we can match that by default

joelanford: I agree with @DirectXMan12. My opinion is that when `context.Context` is exposed to callers, its generally to explicitly hand over full control of deadlining. I'd be little confused if I passed `context.Background()` and it timed out after 30s.

If we're not going to default the timeout, how do we handle `GetInformerForKind` and `GetInformer`, which don't accept a `context.Context`?
* Interally use `context.Background()`
* Make a breaking change to the `Informers` interface to include a `Context` parameter
* Add a new interface that has methods that accept a `Context` and check whether the cache implementation supports it when calling.

Then there's the concern of how that propagates up the call stack. For example, what would we pass [here](https://github.com/kubernetes-sigs/controller-runtime/blob/ab6131a999ca9e6ff5bf83dba70dc7751751135f/pkg/source/source.go#L85)?

> On the other hand, I think a case where we are never going to sync permissions error especially and will hang a worker is a bug that should be fixed.

Is it possible to propagate these errors back up to the caller so that a worker doesn't hang?

shawn-hurley: I will remove the defaults and add the timeout error. 

> Is it possible to propagate these errors back up to the caller so that a worker doesn't hang?

We would have to change the underlying informer to do this, I don't know how we could do that without a backward incompatabile change ATM? 

I like @DirectXMan12 idea of using the client's default timeout for all of these as it should be expected. *note* we should add go doc that one of the reasons this can timeout is permissions

DirectXMan12: > we pass here

`context.TODO`, with a note & bug that when we go to make breaking changes, we fix the signature of start to include a context

DirectXMan12: (any internal uses where you say "we really should have a real context here, but can't because of the interface" should be `TODO`, not `Background`)

DirectXMan12: as for propagating those errors up, it would be nice eventually, but we'd need to get wrapping errors from 1.13 before that becomes even slightly close to possible.

shawn-hurley:
Hopefully getting back to this today, sorry about the delay

alvaroaleman:
@shawn-hurley Anything one can do to help out here? The current behavior of just continuing to run but not doing anything when CRDs or RBAC are wrong is pretty terrible IMO.

Maybe a simple approach to avoiding that case would be to extend the `manager.Options` with a `cacheSyncTimeout` that if non-zero will add another runnable that just does a `GetCache().WaitForCacheSync(ctxWithTimeout.Done())`?

djzager:
> Anything one can do to help out here? The current behavior of just continuing to run but not doing anything when CRDs or RBAC are wrong is pretty terrible IMO.

@alvaroaleman I'm working on creating a PR to finish up this patch based on existing comments.

DirectXMan12:
close this for now, if anyone wants to pick it up, feel free

(IIRC, it's just a matter for folks being busy)

PR #581, opened 2019-09-01
:running: Add script to run tests in Prow

### LINKED ISSUE #276: Run tests via prow
We should be running our test suite in prow instead of Travis, so the normal kube workflow commands work.

alvaroaleman:
This PR adds a script to run tests in Prow. Done to resolve #276

### DIFF
--- a/hack/ci-check-everything.sh
+++ b/hack/ci-check-everything.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+#  Copyright 2018 The Kubernetes Authors.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -e
+
+export GO111MODULE=on
+export TRACE=1
+
+# Not included or existing by default in Prow
+export PATH=$(go env GOPATH)/bin:$PATH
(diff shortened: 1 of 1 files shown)


alvaroaleman:
Requires https://github.com/kubernetes/test-infra/pull/14162 to merge, so we can verify this works. I tested this on a different Prow instance so I am reasonably confident there wont be issues.

Should I already remove the `.travis.yaml` or do we want to merge this -> Make Presubmit mandatory -> Remove `.travis.yaml`?

/assign @DirectXMan12

DirectXMan12:
yeah, let's get mandatory presubmit merged before we remove travis.  Better to run tests twice than not at all.

hack/ci-check-everything.sh
```diff
@@ -0,0 +1,37 @@
...
+export TRACE=1
+
+# Not included by default in Prow
+export PATH=$(go env GOPATH)/bin:$PATH
+
+echo "Installing dep"
+(
+	# Doesn't work with gomodules turned on, see https://github.com/golang/dep/issues/2055
+	export GO111MODULE=off
+	go get -u github.com/golang/dep/cmd/dep
```
DirectXMan12: should we just be downloading the compiled binary here?  Prob *way* quicker.

alvaroaleman:
@DirectXMan12 updated the script to install a binary version of `dep` instead of using `go get`.

Could you also have a look on the pull in test-infra? It adds the new presubmit as optional, so it is safe to merge already. Once it is merged, we can verify here if everything works :)

DirectXMan12:
yep, looked at it.  Looks like @droot isn't in the kubernetes org, but otherwise LGTM

DirectXMan12:
hold off on approving this till that merges?

alvaroaleman:
Okay, so the presubmit passed, with a timeout on the first attempt. My best guess as why that happens is CPU starvation, which is why I opened https://github.com/kubernetes/test-infra/pull/14184 to make the job request CPU.

IMHO this is good to go, PTAL

DirectXMan12:
/lgtm
/approve

awesome

excited for this

PR #587, opened 2019-09-06
:sparkles: Event broadcast opts

liztio:
New method to set correlator options for event recording
Unverified


liztio:
I don't think I fully understand the implications of bumping client-go, is that likely to break other things?

detiber:
/assign @DirectXMan12 @fanzhangio 

This issue is affecting us in cluster-api where the Spam Filter is eating events we are generating.

DirectXMan12:
> implications of bumping client-go

Yeah, this needs to be done carefully otherwise we accidentally break backwards compat

DirectXMan12:
besides the aforementioned bump issues (might just pull that in in #588 anyway), this needs at least one of

- good docs
- a sane wrapper

Alternatively, maybe we should just bump this up a level and let you set the recorder-provider directly?

DirectXMan12:
That way we don't need to carry a rather niche dependency on some highly-implementation-specific options

liztio:
> Alternatively, maybe we should just bump this up a level and let you set the recorder-provider directly?

that makes sense to me, let me mock up a PR rq

liztio:
@DirectXMan12 does #591 look any better to you?

DirectXMan12:
yeah, let's switch to that one

PR #589, opened 2019-09-08
:sparkles: Client option types implement client option interfaces

alvaroaleman:
Currently setting multiple `ListOption`s is pretty annoying, because it is not defined as a function but as an interface, which can not be defined inline.

This PR extends the existing `ListOptions` struct to implement the `ListOption` interface to allow using it as a `ListOption`.

If anyhow possible, I'd be super, super glad if we could get this into the `release-0.2` branch as well :grimacing: 

/assign @DirectXMan12

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -878,11 +878,12 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:14e8a3b53e6d8cb5f44783056b71bb2ca1ac7e333939cc97f3e50b579c920845"
+  digest = "1:ee6063c66a2b3be97487a8df6f359cd5f38b3a0d1fc2442b49f214f24622d321"
   name = "k8s.io/utils"
   packages = [
     "buffer",
     "integer",
+    "pointer",
     "trace",
   ]
   pruneopts = "UT"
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Happy to backport (IIRC, we haven't made any breaking changes yet, so releasing is going to be more-or-less just FF to master for a bit).  Do you mind doing this for the rest of the options types too?  That way we have consistency.

DirectXMan12:
But yeah, this seems pretty reasonable

alvaroaleman:
@DirectXMan12 updated the PR to make all the option types implement the corresponding interfaces.

Obviously I am pretty late to the party to mention that, but I think there is a reason ppl commonly use option funcs and not option interfaces..We can probably think about that for a v0.3

alvaroaleman:
@DirectXMan12 should I create a cherrypick PR for the release-0.2 branch?

DirectXMan12:
We'll pull it in as part of the next release -- we generally don't need to cherry-pick until we start breaking things, and I don't think we've merged anything breaking.

You can do a release PR though -- just ask for a merge of master into release-0.2, and check the release notes by running

```bash
$ git checkout release-0.2
$ git merge master
$ ./hack/release/release-notes.sh
```

PR #591, opened 2019-09-09
:sparkles: Let users specify their own EventBroadcaster for the manager

liztio:
A less-invasive alternative to #587. Lets the user optionally specify their own EventBroadcaster, which allows specifying options for event distribution


DirectXMan12:
Yeah, this looks much better

PR #592, opened 2019-09-10
🐛 Fix #537 - Add apiutil option WithDiscoveryFilter() for NewDynamicRESTMapper()

### LINKED ISSUE #537: client.New() takes more than 10 sec when far from apiserver (apiutil.NewDiscoveryRESTMapper)
I am developing and running the https://github.com/noobaa/noobaa-operator as a CLI tool and remote operator for testing.

When I work with a local cluster like minikube, I don't notice anything, but when I am working to a cluster across the Atlantic, it suddenly takes long time just to initialize the client.

I wrote a short program that shows exactly where the problem is:

```
package main

import (
	"log"

	"k8s.io/apimachinery/pkg/api/meta"
	"k8s.io/client-go/discovery"
	"k8s.io/client-go/rest"
	"k8s.io/client-go/restmapper"
)

func main() {
	log.Printf("Calling config.GetConfigOrDie() ...")
	config := config.GetConfigOrDie()
	log.Printf("Calling apiutil.NewDiscoveryRESTMapper() ...")
	NewDiscoveryRESTMapper(config)
	log.Printf("Done.")
}

func NewDiscoveryRESTMapper(c *rest.Config) (meta.RESTMapper, error) {
	// Get a mapper
	log.Printf("Calling discovery.NewDiscoveryClientForConfigOrDie() ...")
	dc := discovery.NewDiscoveryClientForConfigOrDie(c)
	log.Printf("Calling restmapper.GetAPIGroupResources() ...")
	gr, err := GetAPIGroupResources(dc)
	if err != nil {
		return nil, err
	}
	log.Printf("Calling restmapper.NewDiscoveryRESTMapper() ...")
	return restmapper.NewDiscoveryRESTMapper(gr), nil
}

// GetAPIGroupResources uses the provided discovery client to gather
// discovery information and populate a slice of APIGroupResources.
func GetAPIGroupResources(cl discovery.DiscoveryInterface) ([]*restmapper.APIGroupResources, error) {
	log.Printf("Calling cl.ServerGroups() ...")
	apiGroups, err := cl.ServerGroups()
	if err != nil {
		if apiGroups == nil || len(apiGroups.Groups) == 0 {
			return nil, err
		}
		// TODO track the errors and update callers to handle partial errors.
	}
	var result []*restmapper.APIGroupResources
	for _, group := range apiGroups.Groups {
		groupResources := &restmapper.APIGroupResources{
			Group:              group,
		}
		for _, version := range group.Versions {
			log.Printf("Calling cl.ServerResourcesForGroupVersion(%s) ...", version.GroupVersion)
			resources, err := cl.ServerResourcesForGroupVersion(version.GroupVersion)
			if err != nil {
				// continue as best we can
				// TODO track the errors and update callers to handle partial errors.
				if resources == nil || len(resources.APIResources) == 0 {
					continue
				}
			}
			groupResources.VersionedResources[version.Version] = resources.APIResources
		}
		result = append(result, groupResources)
	}
	return result, nil
}
```

Any this is the output running it on a far away cluster - notice the jump of seconds on the log print times -
```
$ go run a.go
2019/07/24 23:27:21 Calling config.GetConfigOrDie() ...
2019/07/24 23:27:21 Calling apiutil.NewDiscoveryRESTMapper() ...
2019/07/24 23:27:21 Calling discovery.NewDiscoveryClientForConfigOrDie() ...
2019/07/24 23:27:21 Calling restmapper.GetAPIGroupResources() ...
2019/07/24 23:27:21 Calling cl.ServerGroups() ...
2019/07/24 23:27:22 Calling cl.ServerResourcesForGroupVersion(v1) ...
2019/07/24 23:27:22 Calling cl.ServerResourcesForGroupVersion(apiregistration.k8s.io/v1) ...
2019/07/24 23:27:22 Calling cl.ServerResourcesForGroupVersion(apiregistration.k8s.io/v1beta1) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(extensions/v1beta1) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(apps/v1) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(apps/v1beta2) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(apps/v1beta1) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(events.k8s.io/v1beta1) ...
2019/07/24 23:27:23 Calling cl.ServerResourcesForGroupVersion(authentication.k8s.io/v1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(authentication.k8s.io/v1beta1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(authorization.k8s.io/v1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(authorization.k8s.io/v1beta1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(autoscaling/v1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(autoscaling/v2beta1) ...
2019/07/24 23:27:24 Calling cl.ServerResourcesForGroupVersion(autoscaling/v2beta2) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(batch/v1) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(batch/v1beta1) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(certificates.k8s.io/v1beta1) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(networking.k8s.io/v1) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(policy/v1beta1) ...
2019/07/24 23:27:25 Calling cl.ServerResourcesForGroupVersion(rbac.authorization.k8s.io/v1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(rbac.authorization.k8s.io/v1beta1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(storage.k8s.io/v1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(storage.k8s.io/v1beta1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(admissionregistration.k8s.io/v1beta1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(apiextensions.k8s.io/v1beta1) ...
2019/07/24 23:27:26 Calling cl.ServerResourcesForGroupVersion(scheduling.k8s.io/v1beta1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(coordination.k8s.io/v1beta1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(apps.openshift.io/v1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(authorization.openshift.io/v1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(build.openshift.io/v1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(image.openshift.io/v1) ...
2019/07/24 23:27:27 Calling cl.ServerResourcesForGroupVersion(oauth.openshift.io/v1) ...
2019/07/24 23:27:28 Calling cl.ServerResourcesForGroupVersion(project.openshift.io/v1) ...
2019/07/24 23:27:28 Calling cl.ServerResourcesForGroupVersion(quota.openshift.io/v1) ...
2019/07/24 23:27:28 Calling cl.ServerResourcesForGroupVersion(route.openshift.io/v1) ...
2019/07/24 23:27:28 Calling cl.ServerResourcesForGroupVersion(security.openshift.io/v1) ...
2019/07/24 23:27:28 Calling cl.ServerResourcesForGroupVersion(template.openshift.io/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(user.openshift.io/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(packages.operators.coreos.com/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(config.openshift.io/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(operator.openshift.io/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(autoscaling.openshift.io/v1) ...
2019/07/24 23:27:29 Calling cl.ServerResourcesForGroupVersion(autoscaling.openshift.io/v1beta1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(ceph.rook.io/v1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(cloudcredential.openshift.io/v1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(imageregistry.operator.openshift.io/v1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(k8s.cni.cncf.io/v1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(machineconfiguration.openshift.io/v1) ...
2019/07/24 23:27:30 Calling cl.ServerResourcesForGroupVersion(monitoring.coreos.com/v1) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(network.openshift.io/v1) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(operators.coreos.com/v1) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(operators.coreos.com/v1alpha2) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(operators.coreos.com/v1alpha1) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(samples.operator.openshift.io/v1) ...
2019/07/24 23:27:31 Calling cl.ServerResourcesForGroupVersion(tuned.openshift.io/v1) ...
2019/07/24 23:27:32 Calling cl.ServerResourcesForGroupVersion(velero.io/v1) ...
2019/07/24 23:27:32 Calling cl.ServerResourcesForGroupVersion(healthchecking.openshift.io/v1alpha1) ...
2019/07/24 23:27:32 Calling cl.ServerResourcesForGroupVersion(noobaa.io/v1alpha1) ...
2019/07/24 23:27:32 Calling cl.ServerResourcesForGroupVersion(rook.io/v1alpha2) ...
2019/07/24 23:27:32 Calling cl.ServerResourcesForGroupVersion(machine.openshift.io/v1beta1) ...
2019/07/24 23:27:33 Calling cl.ServerResourcesForGroupVersion(metrics.k8s.io/v1beta1) ...
2019/07/24 23:27:33 Calling restmapper.NewDiscoveryRESTMapper() ...
2019/07/24 23:27:33 Done.
```

So it's pretty clear that with high latency network to the apiserver this loop over api groups and versions is really wasteful. we can do much better if we just run these in parallel. 

I can suggest a PR to process this list with go routine workers if it sounds good to you.

Also perhaps a variant of `apiutil.NewDiscoveryRESTMapper()` that takes a filter function that can reduce the number of group-versions that will be read from the server.

Perhaps there's a better way to avoid all this network-chatty initialization?

Thanks!

guymguym:
See #537 

This new option provides a custom filter that is used to select only the needed 
api groups that the controller really needs, instead of the default discover client 
that will discover all the groups on the server.

The motivation is since controllers typically use a fixed small number of api groups,
while servers have many dynamic api groups, and in such cases if the controller 
has some non-zero latency to the api server, or the server is loaded, 
the full groups discovery can take long time (10+ seconds).
The filter can reduce the number of groups drastically.

We might want to consider contributing the filter option to
client-go/discovery client.

Usage:
```
mapper, err := apiutil.NewDiscoveryFastRESTMapper(config,
    apiutil.WithDiscoveryFilter(func(groupVersion string) bool {
        return strings.HasPrefix(groupVersion, "my.awesome.group")
    }),
)
cl, err := client.New(config, client.Options{Mapper: mapper})
```

You can see a similar approach being used in https://github.com/noobaa/noobaa-operator/blob/f11f69e5f3d201b81f79ef8a959031c8820e7945/pkg/util/util.go#L91-L115

### DIFF
--- a/pkg/client/apiutil/dynamicrestmapper.go
+++ b/pkg/client/apiutil/dynamicrestmapper.go
@@ -54,10 +54,11 @@ func DelayIfRateLimited(err error) (time.Duration, bool) {
 // dynamicRESTMapper is a RESTMapper that dynamically discovers resource
 // types at runtime.
 type dynamicRESTMapper struct {
-	mu           sync.RWMutex // protects the following fields
-	staticMapper meta.RESTMapper
-	limiter      *dynamicLimiter
-	newMapper    func() (meta.RESTMapper, error)
+	mu              sync.RWMutex // protects the following fields
+	staticMapper    meta.RESTMapper
+	limiter         *dynamicLimiter
+	newMapper       func() (meta.RESTMapper, error)
+	discoveryClient discovery.DiscoveryInterface
 
 	lazy bool
 	// Used for lazy init.
@@ -94,25 +95,40 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
 	}
 }
 
+// WithDiscoveryFilter sets a groupVersion filter for the discovery client
+// that allows to limit the relevant groups for the mapper
+func WithDiscoveryFilter(filter FilterGroupFunc) DynamicRESTMapperOption {
+	return func(drm *dynamicRESTMapper) error {
+		drm.discoveryClient = &FilterDiscoveryClient{
+			Discovery: drm.discoveryClient,
+			Filter:    filter,
+		}
+		return nil
+	}
+}
+
 // NewDynamicRESTMapper returns a dynamic RESTMapper for cfg. The dynamic
 // RESTMapper dynamically discovers resource types at runtime. opts
 // configure the RESTMapper.
 func NewDynamicRESTMapper(cfg *rest.Config, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
-	client, err := discovery.NewDiscoveryClientForConfig(cfg)
+	defaultDiscoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
 	drm := &dynamicRESTMapper{
+		discoveryClient: defaultDiscoveryClient,
 		limiter: &dynamicLimiter{
 			rate.NewLimiter(rate.Limit(defaultRefillRate), defaultLimitSize),
 		},
-		newMapper: func() (meta.RESTMapper, error) {
-			groupResources, err := restmapper.GetAPIGroupResources(client)
-			if err != nil {
-				return nil, err
-			}
-			return restmapper.NewDiscoveryRESTMapper(groupResources), nil
-		},
+	}
+	drm.newMapper = func() (meta.RESTMapper, error) {
+		// NOTICE that we're using the struct's drm.discoveryClient and not the defaultDiscoveryClient
+		// because we want to allow overridding it by WithDiscoveryFilter()
+		groupResources, err := restmapper.GetAPIGroupResources(drm.discoveryClient)
+		if err != nil {
+			return nil, err
+		}
+		return restmapper.NewDiscoveryRESTMapper(groupResources), nil
 	}
 	for _, opt := range opts {
 		if err = opt(drm); err != nil {
--- a/pkg/client/apiutil/filterdiscoveryclient.go
+++ b/pkg/client/apiutil/filterdiscoveryclient.go
@@ -1,0 +1,118 @@
+package apiutil
+
+import (
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+)
+
+// FilterGroupFunc is a plugin function to FilterDiscoveryClient
+// It should return true for groups that should be included in the discovery,
+// and false for groups that should be excluded.
+type FilterGroupFunc func(groupVersion string) bool
+
+// FilterDiscoveryClient is a wrapper to a discovery interface that adds group filtering
+// in order to reduce the number of groups that are read from the server.
+// It should help to reduce the amount of api calls for clients that require just a few number of groups.
+// Perhaps this can be later added as an optional filter to client-go discovery client to avoid the need for wrapping.
+type FilterDiscoveryClient struct {
+	Discovery discovery.DiscoveryInterface
+	Filter    FilterGroupFunc
+}
+
+// ServerResourcesForGroupVersion is doing the actual filtering
+func (d *FilterDiscoveryClient) ServerResourcesForGroupVersion(groupVersion string) (*metav1.APIResourceList, error) {
+	if d.Filter(groupVersion) {
+		return d.Discovery.ServerResourcesForGroupVersion(groupVersion)
+	}
+	return &metav1.APIResourceList{
+		GroupVersion: groupVersion,
+		APIResources: []metav1.APIResource{},
+	}, nil
+}
+
+// ServerGroupsAndResources calls the public module function with d as the interface
+// so that it will callback to d.ServerResourcesForGroupVersion(groupVersion) for filtering
+func (d *FilterDiscoveryClient) ServerGroupsAndResources() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+	// withRetries was copied from client-go/discovery in order to behave like the standard client
+	// but with the discovery interface calling back to our wrapper
+	return withRetries(defaultRetries, func() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+		return discovery.ServerGroupsAndResources(d)
+	})
+}
+
+// ServerPreferredResources is copied from DiscoveryClient
+func (d *FilterDiscoveryClient) ServerPreferredResources() ([]*metav1.APIResourceList, error) {
+	// withRetries was copied from client-go/discovery in order to behave like the standard client
+	// but with the discovery interface calling back to our wrapper
+	_, rs, err := withRetries(defaultRetries, func() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+		rs, err := discovery.ServerPreferredResources(d)
+		return nil, rs, err
+	})
+	return rs, err
+}
+
+// ServerResources is returning just the resources from the ServerGroupsAndResources() response
+func (d *FilterDiscoveryClient) ServerResources() ([]*metav1.APIResourceList, error) {
+	_, rs, err := d.ServerGroupsAndResources()
+	return rs, err
+}
+
+// ServerPreferredNamespacedResources is a proxy func
+func (d *FilterDiscoveryClient) ServerPreferredNamespacedResources() ([]*metav1.APIResourceList, error) {
+	return discovery.ServerPreferredNamespacedResources(d)
+}
+
+// ServerGroups is delegated to the underlying discovery client
+func (d *FilterDiscoveryClient) ServerGroups() (*metav1.APIGroupList, error) {
+	return d.Discovery.ServerGroups()
+}
+
+// RESTClient is delegated to the underlying discovery client
+func (d *FilterDiscoveryClient) RESTClient() rest.Interface {
+	return d.Discovery.RESTClient()
+}
+
+// ServerVersion is delegated to the underlying discovery client
+func (d *FilterDiscoveryClient) ServerVersion() (*version.Info, error) {
+	return d.Discovery.ServerVersion()
+}
+
+// OpenAPISchema is delegated to the underlying discovery client
+func (d *FilterDiscoveryClient) OpenAPISchema() (*openapi_v2.Document, error) {
+	return d.Discovery.OpenAPISchema()
+}
+
+// defaultRetries is the number of times a resource discovery is repeated if an api group disappears on the fly (e.g. ThirdPartyResources).
+const defaultRetries = 2
+
+// withRetries retries the given recovery function in case the groups supported by the server change after ServerGroup() returns.
+// This was copied from client-go/discovery in order to wrap the discovery client.
+func withRetries(
+	maxRetries int,
+	f func() (
+		[]*metav1.APIGroup,
+		[]*metav1.APIResourceList,
+		error,
+	),
+) (
+	[]*metav1.APIGroup,
+	[]*metav1.APIResourceList,
+	error,
+) {
+	var result []*metav1.APIResourceList
+	var resultGroups []*metav1.APIGroup
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		resultGroups, result, err = f()
+		if err == nil {
+			return resultGroups, result, nil
+		}
+		if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); !ok {
+			return nil, nil, err
+		}
+	}
+	return resultGroups, result, err
+}
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/go-logr/zapr v0.1.0
 	github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef // indirect
-	github.com/googleapis/gnostic v0.3.1 // indirect
+	github.com/googleapis/gnostic v0.3.1
 	github.com/imdario/mergo v0.3.6 // indirect
 	github.com/onsi/ginkgo v1.11.0
 	github.com/onsi/gomega v1.8.1


guymguym:
I have signed the CLA

guymguym:
Changed the commit comment to avoid automatically close issues.

guymguym:
I was not sure what kind of test would be appropriate.
Any pointers would be helpful.
Thanks

DirectXMan12:
This should be on top of #554, since it partially overlaps

DirectXMan12:
/hold
for that

alexeldeib:
#554 merged, I think this is unblocked?

DirectXMan12:
can you rebase this on #554?  It already has some of this logic, so I think we can probably integrate a lot of this into dynamicrestmapper (e.g. concurrent fetches can probably be on by default), and that PR already does laziness IIRC.

guymguym:
Hey @DirectXMan12 

I am seeing that client-go discovery on master seem to already fetch group from api server with concurrency - here is the stack of calls:
https://github.com/kubernetes/client-go/blob/bc9b51d240b2f5000adc2fa96231ab219ab0d702/discovery/discovery_client.go#L345-L346
https://github.com/kubernetes/client-go/blob/bc9b51d240b2f5000adc2fa96231ab219ab0d702/discovery/discovery_client.go#L254
https://github.com/kubernetes/client-go/blob/bc9b51d240b2f5000adc2fa96231ab219ab0d702/restmapper/discovery.go#L145-L147

So concurrency is already in.

Now I am thinking how to add optional filtering - perhaps as a wrapper `DiscoveryInterface` that will do the filtering.

Any thoughts?

DirectXMan12:
If the filter happens after discovery is run though, what's it actually doing?  We'd need to run the filter before fetching the resource info.  Same with if we did fully-lazy (lazy per group) loading instead of filters.

guymguym:
@DirectXMan12 Take a look on the updated diff.
Seems like a hassle that could be prevented if we could add this optional filter to DiscoveryClient itself, but perhaps this is OK as a temporary solution.

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -94,25 +95,38 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
...
-		newMapper: func() (meta.RESTMapper, error) {
-			groupResources, err := restmapper.GetAPIGroupResources(client)
-			if err != nil {
-				return nil, err
-			}
-			return restmapper.NewDiscoveryRESTMapper(groupResources), nil
-		},
+	}
+	drm.newMapper = func() (meta.RESTMapper, error) {
+		groupResources, err := restmapper.GetAPIGroupResources(drm.discoveryClient)
```
DirectXMan12: add a comment emphasizing that we're using the struct's `drm.discoveryClient` and not just `discoveryClient` because it might get overridden below

pkg/client/apiutil/filterdiscoveryclient.go
```diff
@@ -0,0 +1,104 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+)
+
+// defaultRetries is the number of times a resource discovery is repeated if an api group disappears on the fly (e.g. ThirdPartyResources).
+const defaultRetries = 2
+
+type FilterDiscoveryClient struct {
```
DirectXMan12: Needs godoc

DirectXMan12: should this be internal?

pkg/client/apiutil/filterdiscoveryclient.go
```diff
@@ -0,0 +1,104 @@
...
+	Discovery discovery.DiscoveryInterface
+	Filter    func(groupVersion string) bool
```
DirectXMan12: especially this: does `true` mean `keep`, or `throw away`?

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -94,25 +95,38 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
 	}
 }
 
+// WithDiscoveryFilter sets a groupVersion filter for the discovery client
+// that allows to limit the relevant groups for the mapper
+func WithDiscoveryFilter(filter func(groupVersion string) bool) DynamicRESTMapperOption {
```
DirectXMan12: This needs better godoc -- does `true` mean `keep` or `discard`?

pkg/client/apiutil/filterdiscoveryclient.go
DirectXMan12: something something boolean blindness something something

pkg/client/apiutil/filterdiscoveryclient.go
```diff
@@ -0,0 +1,104 @@
...
+	return &metav1.APIResourceList{
+		GroupVersion: groupVersion,
+		APIResources: []metav1.APIResource{},
+	}, nil
+}
+
+// ServerGroupsAndResources calls the public module function with d as the interface
+// so that it will callback to d.ServerResourcesForGroupVersion(groupVersion) for filtering
+func (d *FilterDiscoveryClient) ServerGroupsAndResources() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+	return withRetries(defaultRetries, func() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
```
DirectXMan12: add a note here and below that this is copied from the client-go discovery client, in case we're wondering what's going on here.

DirectXMan12: do we even needs the retries, since this is being used in the dynamic restmapper, which'll retry anyway?

pkg/client/apiutil/filterdiscoveryclient.go
```diff
@@ -0,0 +1,104 @@
...
+
+// ServerResources is returning just the resources from the ServerGroupsAndResources() response
+func (d *FilterDiscoveryClient) ServerResources() ([]*metav1.APIResourceList, error) {
+	_, rs, err := d.ServerGroupsAndResources()
+	return rs, err
+}
+
+// ServerPreferredResources is copied from DiscoveryClient
+func (d *FilterDiscoveryClient) ServerPreferredResources() ([]*metav1.APIResourceList, error) {
+	_, rs, err := withRetries(defaultRetries, func() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
```
DirectXMan12: ditto here WRT to retries

vincepri:
I'm a little confused by the PR contents and the description.

The description says that there is a new version of a DiscoveryClient that is "Fast" by the use of concurrency, although I don't see any code that does that. It seems the code changes only add a filtered dynamic rest mapper.

@DirectXMan12 not sure what you think about this, but I was thinking that we should probably improve with the dynamic rest mapper with optional flags instead of providing more options. Having a "fast" vs "non-fast" can introduce confusion for users.

guymguym:
@vincepri See this comment that explains that concurrency is already in master -
https://github.com/kubernetes-sigs/controller-runtime/pull/592#issuecomment-563204694

So I updated the PR to only add optional group filtering.

vincepri:
Thanks @guymguym, makes more sense now, I'd update the PR description though.

DirectXMan12:
@vincepri that's kinda what I wanted here.  The dynamicrestmapper does most of this already, so adding filtering shouldn't be too bad.

pkg/client/apiutil/dynamicrestmapper.go
guymguym: done

pkg/client/apiutil/filterdiscoveryclient.go
guymguym: Added godoc. I personally prefer to keep it public because it is a generic wrapper to discovery interface, which might be useful in a different manner if someone wants to set a custom mapper.

pkg/client/apiutil/filterdiscoveryclient.go
guymguym: Added a type FilterGroupFunc with godoc that explains the api.

pkg/client/apiutil/dynamicrestmapper.go
guymguym: done in godoc

pkg/client/apiutil/filterdiscoveryclient.go
guymguym: We needed to copy the retries logic only for two cases.
For most of the wrapped functions we just call the underlying discovery interface directly and therefore do not need to add retries.
But for the two calls that we override to implement the filtering which are `ServerGroupsAndResources()` and `ServerPreferredResources()` I had to copy the retries logic too.

I certainly didn't want to do it, but also didn't want to leave these calls to fail too easily.

Other suggestions are certainly welcome.

guymguym:
Hey @DirectXMan12 @vincepri 
I addressed your comments and rebased.
Let me know how this looks now.

guymguym:
CI fails on:
```
go module files are out of date
make: *** [Makefile:100: verify-modules] Error 1
```
@DirectXMan12 Is it related?

DirectXMan12:
Did you run `go mod tidy`?

guymguym:
@DirectXMan12 I did now, and also rebased again. Lets see if that resolved the CI issue. Thanks

guymguym:
Hey @DirectXMan12 @vincepri
Do you guys still think this is relevant to be merged?
A better alternative would be to suggest a PR to DiscoveryClient in client-go with optional group filtering support.
WDYT?

PR #597, opened 2019-09-10
Update gnostic to v0.2.2

fejta:
This is the same commit as v0.2.0 but since v0.2.2 is past the zombie v0.2.1 release go module patch upgrades work correctly.

/assign @DirectXMan12 

If we accept this can we tag it as v0.2.1?


fejta:
Alternatively we could tag master as `v0.3.0` or `v0.2.1`, but this will be the smallest change

DirectXMan12:
since gnostic isn't in our public API anywhere, just tagging master as v0.2.1 should be fine (see #596 )

Issue #590, opened 2019-09-09
Go 1.13 Meta-Issue


DirectXMan12:
Go 1.13 [is out](https://blog.golang.org/go1.13) :tada: !

## Things we can now do:

- [x] Banish Gopkg.toml into the [shadow realm](https://yugioh.fandom.com/wiki/Shadow_Realm) (#598)
- [x] Clean up the various scripts referencing Gopkg.toml (#598)

## Things we can do now-ish

For everything below, we'll need to figure out a potential impact of aggressively switching to Go 1.13 as the min-supported-go-version.  I expect Kubernetes will soon-ish, so it might not be that bad for us to do so now.  That being said, it's almost certainly a breaking change.

- [x] switch go.mod over to go1.13, so we can use new features
- [ ] Prune any uses of xerrors
- [ ] Figure out where we want to use `%w` instead of `%v` in `Errorf`
- [ ] Figure out if there are spots where we wanted to use wrapping errors but couldn't, and come up with a plan for switching over (relatedly, provide transition plans for folks using `apierrors.IsXYZ`) -- will be a :warning: breaking change
- [ ] revive the structured errors code/figure out what we want to do with that
- [ ] figure out if we want to prune uses of `GO111MODULE=on` from our scripts?

DirectXMan12:
/kind feature
/priority important soon

/good-first-issue

(the script stuff is a good first issue, a few of the other items are help-wanted -- feel free to ask if you're not sure)

vaibhavk:
Is it still open as `good-first-issue`?  If so I would like to learn & work on it :)
The script clean-up is merged, except for the `GO111MODULE=on` part.

DirectXMan12:
I think the last bits of the script cleanup are good-first-issue.  The rest is probably requires a bit of thinking through.  If you're comfortable putting together a plan, etc, then go ahead and take any of the other bullet points as well.

PR #598, opened 2019-09-11
:warning: Remove dep usage and Gopkg.*

joelanford:
This removes the `Gopkg.toml` and `Gopkg.lock` files, as well as all references to `dep` in the test/build scripts since we've moved to Go modules.

See first two steps in #590


DirectXMan12:
/lgtm
/approve

can you rebase?

I've marked this as "breaking" so that it shows up prominently in the release notes, since it's kinda-breaking.

Issue #460, opened 2019-05-31
use leader election API if available


DirectXMan12:
We should use the new leader election api when available.  We can check if it's around via discovery, and then use it if available, falling back to configmaps if not

/kind feature
/priority backlog

adohe:
here leader election api means coordination.k8s.io api? @DirectXMan12

DirectXMan12:
yeah

mszostok:
hi, I can provide a PR for that 🙂 
but are you still interested in having that?

DirectXMan12:
yeah, still interested.  Will have to coordinate with #444

timebertt:
What's the status on this issue?
I see there were some attempts at solving this but none of them made it, and kind of related PR #444 is still open.

If using the lease API based on whether it's available or not is not an option, I would still be interested in having the LE resource lock configurable in `leaderelection.Options`.
We could add `Options.LeaderElectionResourceLock` which can be set to one of the [locks supported by client-go](https://github.com/kubernetes/client-go/blob/00dbcca6ee44c678754d3f5fda1bd0e704b26fe2/tools/leaderelection/resourcelock/interface.go#L31-L35) and would be passed directly to `resourcelock.New` [here](https://github.com/kubernetes-sigs/controller-runtime/blob/a3abd2657cbbbeb9fbf97391044f3a265600672b/pkg/leaderelection/leader_election.go#L84).
WDYT?

This would allow controllers to make the LE lock configurable just like in kcm, ccm, etc. (ref `--leader-elect-resource-lock`).
It would very much help in our usecase ([Project Gardener](https://github.com/gardener/gardener)), as we have some scalability issues, when running hundreds of controllers in one cluster, which all use configmap locks by default (results in many `PUT configmap` requests, which in turn flood the API server's watch cache).

If you like the proposed approach, I would definitely be willing to try to solve it in a PR :)

timebertt:
@DirectXMan12 @vincepri 
WDYT about making the lock resource configurable via `leaderelection.Options`?

vincepri:
cc @alvaroaleman, given that there was some discussion on #600.

How we approach this problem depends on what we want our compatibility story look like. I'm fine discussing the best way forward that makes sense for most of our users, although I want to make sure we don't break the world at the same time. 

If we can get a pluggable leader election strategy that'd be the best way to keep backward compatible behaviors.

alvaroaleman:
[...]

PR #600, opened 2019-09-11
✨Add option to use Lease lock if coordination group is available

mszostok:
**Description**

- This PR adds new strategy for `NewResourceLock` method:  _Choose the Lease lock if `lease.coordination.k8s.io` is available otherwise, use ConfigMaps._
- Add tests coverage for implemented logic using ginko & gomega 
- Remove outdated information about cert generation (cert generation was removed in this PR: https://github.com/kubernetes-sigs/controller-runtime/pull/300)

**Notes to reviewer:**
- the `NewResourceLock` method is not easily testable. To do it you need to hack it via injected rest config. I did it in a similar way as in kubectl pkg.
- e2e tested with k8s 1.15 ([kind](https://github.com/kubernetes-sigs/kind))
- unit tests can be changed to `table tests`, see this: https://gist.github.com/mszostok/11278dbc9316d526cd8a2e09f063e50f
but IMO is not worth it.

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/460

---
needs to be sync with https://github.com/kubernetes-sigs/controller-runtime/pull/444

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -935,6 +935,7 @@
     "k8s.io/api/apps/v1",
     "k8s.io/api/apps/v1beta1",
     "k8s.io/api/autoscaling/v1",
+    "k8s.io/api/coordination/v1",
     "k8s.io/api/core/v1",
     "k8s.io/api/extensions/v1beta1",
     "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1",
--- a/pkg/leaderelection/doc.go
+++ b/pkg/leaderelection/doc.go
@@ -19,6 +19,9 @@ Package leaderelection contains a constructors for a leader election resource lo
 This is used to ensure that multiple copies of a controller manager can be run with
 only one active set of controllers, for active-passive HA.
 
-It uses built-in Kubernetes leader election APIs.
+It uses built-in Kubernetes leader election APIs. The Lease lock type takes precedence
+as edits to Leases are less common and fewer objects in the cluster watch "all Leases".
+If the Lease API is not available, the ConfigMap resource lock is used.
+
 */
 package leaderelection
--- a/pkg/leaderelection/leader_election.go
+++ b/pkg/leaderelection/leader_election.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 
+	coordinationv1 "k8s.io/api/coordination/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -37,16 +38,17 @@ type Options struct {
 	LeaderElection bool
 
 	// LeaderElectionNamespace determines the namespace in which the leader
-	// election configmap will be created.
+	// election will be created.
 	LeaderElectionNamespace string
 
-	// LeaderElectionID determines the name of the configmap that leader election
+	// LeaderElectionID determines the name of the resource lock that leader election
 	// will use for holding the leader lock.
 	LeaderElectionID string
 }
 
-// NewResourceLock creates a new config map resource lock for use in a leader
-// election loop
+// NewResourceLock creates a new resource lock to use in a leader
+// election loop. Choose the Lease lock if `lease.coordination.k8s.io` is available.
+// Otherwise, the ConfigMap resource lock is used.
 func NewResourceLock(config *rest.Config, recorderProvider recorder.Provider, options Options) (resourcelock.Interface, error) {
 	if !options.LeaderElection {
 		return nil, nil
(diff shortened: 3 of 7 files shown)


mszostok:
I've checked the #444 and there is no merge conflict because the logic that I'm adding in this PR is directly in `NewResourceLock` func and for #444 it should be like a "blackbox"

one thing that needs to be changed are comments:
```
	// LeaderElectionID determines the name of the configmap that leader election
``` 
we should remove the information about the `configmaps` cause it's implementation details
so on my PR, I've changed that into

```
	// LeaderElectionID determines the name of the resource lock that leader election
```

If we agree on that then on #444 it should be adjusted too

/cc @DirectXMan12

mszostok:
@DirectXMan12 thanks for comments, please check my answers

> Should we try pass in the RESTMapper instead of adding yet another call to discovery?

tbh I didn’t take this into account because this is only one call to api-server and only at the beginning. Using client allows us to encapsulate new logic inside the 	`NewResourceLock`. Personally, for me it’s not a case for optimization but here is draft PR with RESTMapper: https://github.com/mszostok/controller-runtime/pull/1/files

if you think that it's better approach then I will rewrite the current proposition and change test coverage. 

---

> Should this be configurable?

Good question, in the task there was only info that new API should be used when available and fall back to configmaps if not, so I didn't customized that

What kind of customization level you were thinking about?
1) User should be able to provide own `NewResourceLock` func? Like in this PR https://github.com/kubernetes-sigs/controller-runtime/pull/535?
2) Add field `LeaderElectionLockType` which would be an enum:
   - `EndpointsResourceLock`
   - `ConfigMapsResourceLock`
   - `LeasesResourceLock`
   If the field will be not set by a user, then we set preferred one, example: https://github.com/mszostok/controller-runtime/pull/1/files

3) both 1&2

--- 

one thing that comes to my mind which I want to clarify. This impl will be a little "breaking change", right? 

**Scenario 1:**
User has a controller with old impl version, then updates the controller-runtime library. In previous version the CM was used, in the new one, the Lease will be used. Problem is that the old ConfigMap will not be garbage collected, so we will have orphaned object

**Scenario 2:**
User updates the controller-runtime library and performs the rolling update of controller Deployment. Problem:

- The previous version has an old strategy (ConfigMap) - still deployed
- The new one has a new strategy (Lease) - newly deployed

Both will work at the same time (reconcile process). Because the old one will check the ConfigMap and the new one will check the Lease. So user needs to know controller Deployment should have the Recreate strategy to prevent such a situation.

So what about the migration path? Should it be automated? Documented?

DirectXMan12:
cc @JoelSpeed (not sure how active you want to be anymore -- lmk if you don't want to be tagged),  @GrigoriyMikhalkin, @alvaroaleman (all of whom have had thoughts on LE in the past)

JoelSpeed:
This seems sensible to me, glad someone's gotten around to doing this! 

What happens during rolling updates when changing from the `ConfigMap` to `ResourceLock` implementations? I guess it's pretty unavoidable that the leader election will fail temporarily if two different versions of the controller are running?

leongor:
Is this PR still relevant? Can it be merged?

alvaroaleman:
As Joel mentioned, we need to think about how we  migrate existing controllers that currently use a configmap

leongor:
I understand, but meantime this option can be used by those who do not need to migrate.
In our environment, it's needed desperately as we cannot use configmap.

alvaroaleman:
> I understand, but meantime this option can be used by those who do not need to migrate.
In our environment, it's needed desperately as we cannot use configmap.

Unfortunately not, because with the changes in this PR controllers will automatically use the lease lock if its available. This means that if a controller gets updated to this version of c-r and then deployed, there are two replicas that use a different mechanism for leader election which means they can hold a leader lock simultaneously. This means we very silently break valid expectations ("If I enable leader election, there will always at most be one active replica") and that is not an option.

Someone has to come up with a concept for what we can do to prevent that. Possible approaches include:
* Default to configmap and only use leader lock if explicitly requested
* Check if there is an existing configmap leader lock in the cluster and if yes, default to using configmap

leongor:
I think it's a reasonable approach to stick with configmap as default and provide an option to change to lease.
The only thing the prevent to use lease today is that `newResourceLock` field in manager options is private and cannot be overridden.

alvaroaleman:
Yeah, generally that sounds reasonable. It is subpar insofar that we basically say the recommended leader lock is not the default one due to these compatibility concerns but I guess its the best we can do and we can add a TODO to the manager to change the default when we break its interface.

leongor:
Sounds good for me.

leongor:
So, can we reopen this PR or do we need a new one?

alvaroaleman:
Well, the PR got closed due to staleness which seems to indicate that the
original author doesn't have the cycles to finish this up so opening a new
PR may make sense.

PR #606, opened 2019-09-17
:warning: Bump to Go 1.13

joelanford:
This PR builds on #598 by completing more of the tasks listed in #590:

- [x] Prune any uses of xerrors (89320998b0928b69df3b5fa255a3eb428333213c)
- [x] Use %w instead of %v in Errorf (fc42a0a06fc146f0a1d314ca254160409617295d)
- [x] Prune uses of GO111MODULE=on from our scripts (e0a1fbb718d55f200c27d2a7dfde7f91901cd8ec)

(note: we already did a soft-bump to 1.13 the previous release, this just starts using the features)

### DIFF
--- a/hack/check-everything.sh
+++ b/hack/check-everything.sh
@@ -70,11 +70,18 @@ function is_installed {
   return 1
 }
 
+function golangci_lint_has_version {
+  # Trim "v" from version prefix since golangci-lint --version
+  # sometimes does not include it, depending on how it's built
+  golangci-lint --version | grep --quiet --fixed-strings "${1#"v"}"
+}
+
 function fetch_go_tools {
-  header_text "Checking for gometalinter.v2"
-  if ! is_installed golangci-lint; then
+  header_text "Checking for golangci-lint"
+  local golangci_lint_version="v1.21.0"
+  if ! is_installed golangci-lint || ! golangci_lint_has_version "${golangci_lint_version}"; then
     header_text "Installing golangci-lint"
-    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
+    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin "${golangci_lint_version}"
   fi
 }
 
--- a/hack/ci-check-everything.sh
+++ b/hack/ci-check-everything.sh
@@ -22,8 +22,4 @@ export TRACE=1
 export PATH=$(go env GOPATH)/bin:$PATH
 mkdir -p $(go env GOPATH)/bin
 
-echo "Installing golangci-lint"
-curl --location --silent --retry 5 --fail  https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.21.0
-echo "Finished installing golangci-lint"
-
 $(dirname ${BASH_SOURCE})/check-everything.sh
--- a/hack/ci-check-everything.sh
+++ b/hack/ci-check-everything.sh
@@ -16,7 +16,6 @@
 
 set -e
 
-export GO111MODULE=on
 export TRACE=1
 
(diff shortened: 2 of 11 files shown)


alvaroaleman:
/ok-to-test

In order for the tests to use go 1.13, the image needs to be bumped here: https://github.com/kubernetes/test-infra/blob/e0a1e8e0d0f947087e7b8295f49b21d772f7a070/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits.yaml#L13

DirectXMan12:
(removed my at-mention from the PR description -- that causes all sorts of issues when it merges)

hack/check-everything.sh
```diff
@@ -70,7 +70,7 @@ function fetch_go_tools {
   header_text "Checking for gometalinter.v2"
```
rajathagasthya: nit: while you are changing this file, `s/gometalinter.v2/golangci-lint/`

hack/check-everything.sh
```diff
@@ -70,7 +70,7 @@ function fetch_go_tools {
...
   if ! is_installed golangci-lint; then
     header_text "Installing golangci-lint"
-    GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.15.0
+    go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.15.0
```
rajathagasthya: I know this isn't changing current behavior, but maybe this tool should be installed with `GO111MODULE=off`? The current version will add this as a dependency to `go.mod` cluttering the file. We don't really want to carry a lint tool as a dependency and someone might accidentally commit it.

Related https://github.com/golang/go/issues/30515

joelanford: I agree that it isn't great to clutter and inadvertantly change `go.mod`. This approach doesn't work though:

```
$ ./hack/check-everything.sh
using tools
Checking for golangci-lint
Installing golangci-lint
go: cannot use path@version syntax in GOPATH mode
```

I updated this to use the `curl`/`goreleaser` installation method that's used elsewhere. Was there some reason we weren't using that here?

rajathagasthya: > The GO111MODULE=off approach doesn't work though

Ah, yes, because of the version. Got it. 

> Was there some reason we weren't using that here?

If I have to guess, it was done before go modules were introduced and it never changed.

joelanford:
No huge rush here, but this is still waiting on https://github.com/kubernetes/test-infra/pull/14373.

I _think_ that should pass CI and be able to be merged without breaking `master` on controller-runtime, but I'm not positive.

@DirectXMan12 can you add `ok-to-test` to https://github.com/kubernetes/test-infra/pull/14373 when you get a chance?

alvaroaleman:
@joelanford Does it make sense to have two presubmits, one with go 1.12 for the release-0.2 branch and one with go 1.13 for the master branch or can we safely assume that the presubmit with go 1.13 will work on both?

joelanford:
@alvaroaleman I see no reason it would not work on both. But the question in my mind is whether this PR will necessitate moving to `v0.3.0` (I think not, but I'm not sure).

If we jump to `v0.3.0` after this merges, it would probably make sense to continue using go1.12 for the `release-0.2` branch.

If we just continue to `v0.2.3` after this merges, then the next commit on `release-0.2` will need go1.13, so I think there would be no reason to have a separate presubmit, unless there's a chance that we backport a bug fix to `release-0.2` without releasing everything else on `master`.

DirectXMan12:
We *probably* want to make sure we don't break go1.12 on release-0.2 -- IMO, we'll want to go to v0.3.0 for go1.13, but I'm not certain exactly how we want to deal with language version bumps.

DirectXMan12:
What do other people think, especially end consumers of CR?  Is go-1.13 in all the major distros at this point yet?

joelanford:
Just updated kubernetes/test-infra#14373 to have separate builds for each branch and to bump `master` to use go 1.13.

I also updated this PR to move from `xerrors` (which we introduced in the DynamicRESTMapper PR) to `errors`.

Anything else we should handle here?

PR #609, opened 2019-09-18
 add apiClient make requests to the api server and not the cache

fanux:
:sparkles:

https://github.com/kubernetes-sigs/kubebuilder/issues/947

If we want to query and write CRD in some agents, like CNI. or create a command line (not long live process). apiClient without cache is needed

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -73,6 +73,9 @@ type controllerManager struct {
 	// apiReader is the reader that will make requests to the api server and not the cache.
 	apiReader client.Reader
 
+	// apiClient is the client that will make requests to the api server and not the cache.
+	apiClient client.Client
+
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Please see my response on the corresponding issue

/hold

fanux:
client.New is not good enough to use. See my response.

fanux:
Using client:
```
cl, err := client.New(config.GetConfigOrDie(), client.Options{})
if err != nil {
    fmt.Println("failed to create client")
    os.Exit(1)
}

podList := &corev1.PodList{}

err = cl.List(context.Background(), podList, client.InNamespace("default"))
if err != nil {
    fmt.Printf("failed to list pods in namespace default: %v\n", err)
    os.Exit(1)
}
```
https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#New

PR #620, opened 2019-10-01
✨ fakeClient: bump ResourceVersion on write

zaneb:
In a real system, the ResourceVersion of a resource will be incremented whenever it is written to. To allow tests to rely on this, fakeClient should do the same.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -61,11 +66,33 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.O
...
 	}
 	return &fakeClient{
-		tracker: tracker,
+		tracker: versionedTracker{tracker},
 		scheme:  clientScheme,
 	}
 }
 
+func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Object, ns string) error {
+	if accessor, err := meta.Accessor(obj); err == nil {
```
alvaroaleman: Wouldn't it make more sense to return the error here if its non-nil? Granted, that shouldn't happen in the first place, but if it does it would IMHO be nicer to return that instead of silently degrading

PR #624, opened 2019-10-02
📖 Document environment variables supported by envtest

mariantalla:
This is part of https://github.com/kubernetes-sigs/kubebuilder/issues/1000 and a followup to the work in https://github.com/kubernetes-sigs/kubebuilder/pull/1055.

Adds information on envtest environment variables to the GoDoc.

/kind documentation

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -33,7 +33,18 @@ import (
 
 var log = logf.RuntimeLog.WithName("test-env")
 
-// Default binary path for test framework
+/*
+It's possible to override some defaults, by setting the following environment variables:
+	USE_EXISTING_CLUSTER (boolean): if set to true, envtest will use an existing cluster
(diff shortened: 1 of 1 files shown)


DirectXMan12:
do those actually show up in the godoc in that position?  If so, lgtm, otherwise, maybe just move to the package doc comment.

mariantalla:
They show up as part of the documentation on `Constants`: 

image

It would be nice to be able to show the actual defaults (`defaultKubebuilderPath`, `defaultKubebuilderControlPlaneStart/StopTimeout` too, but I think that's only possible if they become exported.

mariantalla:
Added a small note on defaults, which now shows up like this. Let me know if it doesn't work and I'll revert that one.

image

mariantalla:
👋 Any thoughts on this one?

/assign @DirectXMan12 @pwittrock @droot

DirectXMan12:
awesome, thanks!

/lgtm
/approve

PR #625, opened 2019-10-03
:sparkles: Avoid extra deep copy in CacheReader.List

### LINKED ISSUE #621: Avoid deep copying objects twice in the CacheReader List method
In the [`List` method of the `CacheReader`](https://github.com/kubernetes-sigs/controller-runtime/blob/801e12a501606245062a6ff77ac91eef8936d9fb/pkg/cache/internal/cache_reader.go#L122-L136), the list of objects retrieved from the index is added to the `runtimeObjs` array by deep copying each object. This array is then passed to the `objectutil.FilterWithLabels` which again creates a deep copy of the objects that match the label selector. 

It would be more efficient to inline the label filtering logic as follows to avoid this extra work and reduce heap usage of controllers that work with large number of items.

```patch
diff --git a/pkg/cache/internal/cache_reader.go b/pkg/cache/internal/cache_reader.go
index a9b7ae34..8564a2da 100644
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
-	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
 )
 
 // CacheReader is a CacheReader
@@ -125,15 +124,22 @@ func (c *CacheReader) List(_ context.Context, out runtime.Object, opts ...client
 		if !isObj {
 		}
+		meta, err := apimeta.Accessor(obj)
+		if err != nil {
+			return err
+		}
+		if labelSel != nil {
+			lbls := labels.Set(meta.GetLabels())
+			if !labelSel.Matches(lbls) {
+				continue
+			}
+		}
+
 		outObj := obj.DeepCopyObject()
 		outObj.GetObjectKind().SetGroupVersionKind(c.groupVersionKind)
 		runtimeObjs = append(runtimeObjs, outObj)
 	}
-	filteredItems, err := objectutil.FilterWithLabels(runtimeObjs, labelSel)
-	if err != nil {
-		return err
-	}
-	return apimeta.SetList(out, filteredItems)
+	return apimeta.SetList(out, runtimeObjs)
 }
 
 // objectKeyToStorageKey converts an object key to store key.
```

Are there any downsides to patching the method as suggested? If not, I'll be happy to raise a PR.

charith-elastic:
`CacheReader.List` method creates deep copies of objects retrieved from the index and then passes them on to `objectutil.FilterWithLabels` which again deep copies objects matching the label selector. This PR inlines the label filtering logic to avoid the double deep copy.

Fixes #621

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
 	k8s.io/kube-openapi v0.0.0-20180731170545-e3762e86a74c // indirect
-	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5 // indirect
+	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5
 	sigs.k8s.io/testing_frameworks v0.1.1
 	sigs.k8s.io/yaml v1.1.0
 )
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
 )
 
 // CacheReader is a CacheReader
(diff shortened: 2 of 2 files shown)


charith-elastic:
I signed it

alvaroaleman:
/ok-to-test
/lgtm
/assign @DirectXMan12 
/check-cla

Nice change, thanks for doing it!

Should you have time for that it would be nice to have some tests that verify that LabelSelector work properly here. But that didn't exist before your change either.

charith-elastic:
> Should you have time for that it would be nice to have some tests that verify that LabelSelector work properly here. But that didn't exist before your change either.

I think there are some label selector tests for the cache that ends up exercising the code here. The coverage report indicates that this bit is covered pretty well. If you have any suggestions about additional tests that should be added, I'll  be happy to look into those.

image

/check-cla

alvaroaleman:
> I think there are some label selector tests for the cache that ends up exercising the code here. The coverage report indicates that this bit is covered pretty well. If you have any suggestions about additional tests that should be added, I'll be happy to look into those

Oh right, I guess that's fine then, I just saw the missing tests in the packages themselves when reviewing.

PR #639, opened 2019-10-14
✨ Add helpers for remaining Zap options

hasbro17:
This PR adds predefined functional options for more easily setting the logger options in `pkg/log/zap`.

Follow up to #560 per https://github.com/kubernetes-sigs/controller-runtime/pull/560#discussion_r317664717

/cc @DirectXMan12 @alvaroaleman

### DIFF
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -64,14 +64,9 @@ func LoggerTo(destWriter io.Writer, development bool) logr.Logger {
 //
 // Deprecated: use NewRaw() and the functional opts pattern instead:
 //
-// NewRaw(UseDevMode(development))
+// NewRaw(UseDevMode(development), WriteTo(destWriter), RawZapOpts(opts...))
 func RawLoggerTo(destWriter io.Writer, development bool, opts ...zap.Option) *zap.Logger {
-	o := func(o *Options) {
-		o.DestWritter = destWriter
-		o.Development = development
-		o.ZapOpts = opts
-	}
-	return NewRaw(o)
+	return NewRaw(UseDevMode(development), WriteTo(destWriter), RawZapOpts(opts...))
 }
 
 // Opts allows to manipulate Options
@@ -87,10 +82,43 @@ func UseDevMode(enabled bool) Opts {
 }
 
 // WriteTo configures the logger to write to the given io.Writer, instead of standard error.
-// See Options.WriterTo.
+// See Options.DestWritter
 func WriteTo(out io.Writer) Opts {
 	return func(o *Options) {
 		o.DestWritter = out
+	}
+}
+
+// Encoder configures how the logger will encode the output e.g JSON or console.
+// See Options.Encoder
+func Encoder(encoder zapcore.Encoder) func(o *Options) {
+	return func(o *Options) {
+		o.Encoder = encoder
+	}
+}
+
+// Level sets the the minimum enabled logging level e.g Debug, Info
+// See Options.Level
+func Level(level *zap.AtomicLevel) func(o *Options) {
+	return func(o *Options) {
+		o.Level = level
+	}
+}
+
+// StacktraceLevel configures the logger to record a stack trace for all messages at
+// or above a given level.
+// See Options.StacktraceLevel
+func StacktraceLevel(stacktraceLevel *zap.AtomicLevel) func(o *Options) {
+	return func(o *Options) {
+		o.StacktraceLevel = stacktraceLevel
+	}
+}
+
+// RawZapOpts allows appending arbitrary zap.Options to configure the underlying zap logger.
+// See Options.ZapOpts
+func RawZapOpts(zapOpts ...zap.Option) func(o *Options) {
+	return func(o *Options) {
+		o.ZapOpts = append(o.ZapOpts, zapOpts...)
 	}
 }
 


hasbro17:
This covers the all the provided options but I was wondering if we should expose more predefined options for the [EncoderConfig](https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig)?
I guess you can still set that via the Encoder option.
```Go
import (
	"os"

	uzap "go.uber.org/zap"
	"go.uber.org/zap/zapcore"
	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
)

var log = logf.Log.WithName("cmd")

func main() {
	ecfg := uzap.NewDevelopmentEncoderConfig()
	ecfg.EncodeTime = zapcore.EpochTimeEncoder
	ecfg.EncodeLevel = zapcore.LowercaseLevelEncoder
	encoder := zapcore.NewConsoleEncoder(ecfg)

	zapLogr := zap.New(zap.Development(true), zap.DestWritter(os.Stdout), zap.Encoder(encoder))
	logf.SetLogger(zapLogr)

	log.Info("Test message", "value", 1)

}
```
And having options for both Encoder and EncoderConfig might be confusing if you could set two different configs via both options.

hasbro17:
@alvaroaleman I also just realized that the destination writer option is named `DestWritter`
https://github.com/kubernetes-sigs/controller-runtime/blob/16c93b06b0a8cb99e002529cb1b43dce5e489baf/pkg/log/zap/zap.go#L97

Is that supposed to be with a single T `DestWriter`? I could change that, but that might be a breaking change.

alvaroaleman:
/ok-to-test

> Is that supposed to be with a single T DestWriter? I could change that, but that might be a breaking change.

Yes, seems there is a typo. Please change it, I guess Solly will defer changing that to a new release anyways.

pkg/log/zap/zap.go
```diff
@@ -149,6 +137,48 @@ func (o *Options) addDefaults() {
 	o.ZapOpts = append(o.ZapOpts, zap.AddStacktrace(o.StacktraceLevel))
 }
 
+// Development sets the Development option for the logger
+func Development(development bool) func(o *Options) {
```
alvaroaleman: Would it make sense for this to not have the boolean arg, considering it wont do anything if the arg is set to `false`?

pkg/log/zap/zap.go
```diff
@@ -149,6 +137,48 @@ func (o *Options) addDefaults() {
...
+	return func(o *Options) {
+		o.Development = development
+	}
+}
+
+// Encoder sets the Encoder option for the logger
```
alvaroaleman: Should we probably mention the defaults in the godocs for the new funcs?

pkg/log/zap/zap.go
```diff
@@ -149,6 +137,48 @@ func (o *Options) addDefaults() {
...
+
+// StacktraceLevel sets the StacktraceLevel option for the logger
+func StacktraceLevel(stacktraceLevel *zap.AtomicLevel) func(o *Options) {
+	return func(o *Options) {
+		o.StacktraceLevel = stacktraceLevel
+	}
+}
+
+// ZapOpts appends zapOpts to the raw zap.Options option of the logger
+func ZapOpts(zapOpts []zap.Option) func(o *Options) {
```
alvaroaleman: Maybe `RawZapOpts` to avoid the linter error?

pkg/log/zap/zap.go
hasbro17: You're right. This is pretty minor but I'm on the fence about this.
Without the arg it's an implicit assumption on the default being false.
And as the caller you can't be explicit about setting a Production config without the arg. 

Not sure if there is a convention for functional options on this.

pkg/log/zap/zap.go
hasbro17: We have them defined already for the Option fields so I wasn't sure if I should repeat those here again.
https://github.com/kubernetes-sigs/controller-runtime/blob/16c93b06b0a8cb99e002529cb1b43dce5e489baf/pkg/log/zap/zap.go#L93-L94

pkg/log/zap/zap.go
hasbro17: Actually scratch that. I completely misread what you were saying :sweat_smile:
You're right, but it's a moot point now after https://github.com/kubernetes-sigs/controller-runtime/pull/646#issuecomment-542958563

hasbro17:
Didn't see #646 get merged which has the first two options. I'll rebase this to only add the remaining option helpers.

pkg/log/zap/zap.go
```diff
@@ -82,18 +77,51 @@ type Opts func(*Options)
 // See Options.Development
 func UseDevMode(enabled bool) Opts {
 	return func(o *Options) {
-		o.Development = true
+		o.Development = enabled
```
hasbro17: @DirectXMan12 Did you mean to fix this?
https://github.com/kubernetes-sigs/controller-runtime/pull/646#issuecomment-543872875

joelanford: @hasbro17 I noticed this as well, and fixed it and added tests in #653.

hasbro17:
PTAL @mengqiy or @droot if you get a chance.

There's a small overlap with #653 so we can rebase either PR depending on what gets merged first.

pkg/log/zap/zap.go
```diff
@@ -82,18 +77,51 @@ type Opts func(*Options)
...
+// See Options.StacktraceLevel
+func StacktraceLevel(stacktraceLevel *zap.AtomicLevel) func(o *Options) {
+	return func(o *Options) {
+		o.StacktraceLevel = stacktraceLevel
+	}
+}
+
+// RawZapOpts allows appending arbitrary zap.Options to configure the underlying zap logger.
+// See Options.RawZapOpts
+func RawZapOpts(zapOpts []zap.Option) func(o *Options) {
```
DirectXMan12: nit: make this variadic

PR #642, opened 2019-10-16
:sparkles: Add support for multi-value KUBECONFIG

### LINKED ISSUE #564: GetConfig does not respect multi-value KUBECONFIG
`GetConfig` doesn't respect multi-value KUBECONFIG, and it should, according to the [docs on kubeconfig](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable).

`clientcmd` has code to do this -- we should audit our config loading logic and see if we can lean on clientcmd more agressively to match desired behavior.

re-filed from kubernetes-sigs/kubebuilder#936

/kind bug
/priority important-soon

joelanford:
This PR adds support in `pkg/client/config.GetConfig()` and `pkg/client/config.GetConfigWithContext()` to handle multi-value KUBECONFIG environment variable values, as documented in the [docs about kubeconfig](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable).

It also adds unit tests, which test config loading for the kubeconfig flag, environment variable, and default home directory location. In-cluster configurations are not tested because they require files to exist in hard-coded file system locations.

Closes #564

### DIFF
--- a/pkg/client/config/config_test.go
+++ b/pkg/client/config/config_test.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/client/config/config_test.go
```diff
@@ -0,0 +1,243 @@
...
+		var err error
+		dir, err = ioutil.TempDir("", "cr-test")
+		Expect(err).NotTo(HaveOccurred())
+
+		// override $HOME/.kube/config
+		clientcmd.RecommendedHomeFile = filepath.Join(dir, ".kubeconfig")
+	})
+
+	AfterEach(func() {
+		os.Unsetenv(clientcmd.RecommendedConfigPathEnvVar)
```
DirectXMan12: please add a note about where this is set

PR #645, opened 2019-10-16
[wip] :warning: Secure Serving & Webhook Support in envtest

DirectXMan12:
***This is currently a prototype -- the code is pretty terribly organized***

I've hacked together support for testing webhooks in envtest.  Along the way, I also added support for authn in envtest, which will eventually become necessary when we have to stop using the insecure port and switch to the secure port.

This is pretty hacky, and includes all of testing_frameworks (as suggested in https://github.com/kubernetes-sigs/testing_frameworks/pull/68), but I wanted to get it up just in case people were interested.

I might not have time to clean it up soon, but will most likely get back to it eventually.


DirectXMan12:
gonna close this one, partially picked up in #787, secure serving bits need a new owner though.

PR #646, opened 2019-10-16
:sparkles: Add helpers for common Zap options

DirectXMan12:
This adds helpers for common Zap options, so that users don't have to
write them themselves.

It also fixes the Deprecation warnings to use the standard syntax, and
fixes the options godocs to follow proper godocs format.

### DIFF
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -41,11 +41,9 @@ func New(opts ...Opts) logr.Logger {
 // (stacktraces on warnings, no sampling), otherwise a Zap production
 // config will be used (stacktraces on errors, sampling).
 //
-// Deprecated, use New() and the functional opts pattern instead:
+// Deprecated: use New() and the functional opts pattern instead:
 //
-// New(func(o *Options){
-//    o.Development = development
-// })
+// New(UseDevMode(development))
 func Logger(development bool) logr.Logger {
 	return LoggerTo(os.Stderr, development)
 }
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cc @mengqiy

pkg/log/zap/zap.go
```diff
@@ -54,9 +52,9 @@ func Logger(development bool) logr.Logger {
 // to the given destination, instead of stderr.  It otherwise behaves like
 // ZapLogger.
 //
-// Deprecated, use New() and the functional opts pattern instead:
+// Deprecated: use New() and the functional opts pattern instead:
 //
-// New(func(o *Options){
+// New(UseDevMode(development), WriteTo(writer))
 //    o.Development = development
```
mengqiy: Please drop the following 3 lines.

joelanford:
@DirectXMan12 it doesn't look like the `enabled` variable is used in [`func UseDevMode(enabled bool)`](https://github.com/kubernetes-sigs/controller-runtime/blob/ecdbe54c7cc55f9dfcc6cb56fc1622d182aa2219/pkg/log/zap/zap.go#L83-L87).

Did we mean to use `enabled` to set the value of `o.Development`? Or should we just get rid of the `enabled` parameter, such that the function always sets `o.Development` to `true`?

DirectXMan12:
Yeah, that's an accident.  I was torn, since it makes it more annoying to have a flag or something for dev mode if you don't have the parameter

PR #647, opened 2019-10-17
:sparkles: add 10% jitter to ResyncPeriod

### LINKED ISSUE #648: use ResyncPeriod func in case of multi controllers list simultaneously
like [kube-controller-manger](https://github.com/kubernetes/kubernetes/blob/f7091992c04e2d4f16f6b462d4a711e5b9eb0f22/cmd/kube-controller-manager/app/controllermanager.go#L151) do

xrmzju:
all informers use the same resync period, this may cause multi controllers send list requests to apiserver simultaneously, add a jitter to resync period would be better.
fixes #648 

this pr uses `ResyncPeriod ` to add a jitter to the `SyncPeriod` option before finally pass to the `NewInformer` func

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -93,7 +93,10 @@ type Options struct {
 	// Mapper is the RESTMapper to use for mapping GroupVersionKinds to Resources
 	Mapper meta.RESTMapper
 
-	// Resync is the resync period. Defaults to defaultResyncTime.
+	// Resync is the base frequency the informers are resynced.
+	// Defaults to defaultResyncTime.
+	// A 10 percent jitter will be added to the Resync period between informers
+	// So that all informers will not send list requests simultaneously.
 	Resync *time.Duration
 
 	// Namespace restricts the cache's ListWatch to the desired namespace
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -111,6 +111,8 @@ type Options struct {
 	// reconciled. A lower period will correct entropy more quickly, but reduce
 	// responsiveness to change if there are many watched resources. Change this
 	// value only if you know what you are doing. Defaults to 10 hours if unset.
+	// there will a 10 percent jitter between the SyncPeriod of all controllers
+	// so that all controllers will not send list requests simultaneously.
 	SyncPeriod *time.Duration
 
 	// LeaderElection determines whether or not to use leader election when
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -18,6 +18,7 @@ package internal
 
 import (
 	"fmt"
+	"math/rand"
 	"sync"
 	"time"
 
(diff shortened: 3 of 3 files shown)


pkg/manager/manager.go
```diff
@@ -107,11 +107,12 @@ type Options struct {
...
 
-	// SyncPeriod determines the minimum frequency at which watched resources are
+	// MinResyncPeriod determines the minimum frequency at which watched resources are
 	// reconciled. A lower period will correct entropy more quickly, but reduce
 	// responsiveness to change if there are many watched resources. Change this
-	// value only if you know what you are doing. Defaults to 10 hours if unset.
-	SyncPeriod *time.Duration
+	// value only if you know what you are doing. Defaults to 5 hours if unset.
+	// the resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod
+	MinResyncPeriod *time.Duration
```
rajathagasthya: This is a breaking change.

DirectXMan12:
closing for now -- see my response on the associated issue

DirectXMan12:
I misunderstood the purpose earlier -- this is about adding jitter to the resync interval.  We'll need to make that more clear (and not make this a breaking change)

xrmzju:
> I misunderstood the purpose earlier -- this is about adding jitter to the resync interval. We'll need to make that more clear (and not make this a breaking change)

does [not make this a breaking change] means we should keep the `Resync` option and add a new `MinResync` option? or we just reuse the `Resync` option

DirectXMan12:
Reusing the resync option and using it as input for the jitter seems reasonable to me, but if you have other thoughts let me know.  More review inline

pkg/cache/cache.go
```diff
@@ -93,23 +93,23 @@ type Options struct {
 	// Mapper is the RESTMapper to use for mapping GroupVersionKinds to Resources
 	Mapper meta.RESTMapper
 
-	// Resync is the resync period. Defaults to defaultResyncTime.
-	Resync *time.Duration
+	// The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod.
+	MinResyncPeriod *time.Duration
```
DirectXMan12: just as a reminder: this is breaking

pkg/cache/internal/deleg_map.go
```diff
@@ -39,19 +40,27 @@ type InformersMap struct {
 	Scheme *runtime.Scheme
 }
 
+// resyncPeriod returns a function which generates a duration each time it is
+// invoked; this is so that multiple controllers don't get into lock-step and all
+// hammer the apiserver with list requests simultaneously.
+func resyncPeriod(resync time.Duration) func() time.Duration {
+	return func() time.Duration {
+		factor := rand.Float64() + 1
```
DirectXMan12: this seems like a *really* big variance.

IMO, it probably makes more sense to say `Resync +/- some percentage`, like `Resync +/- 10%`.

DirectXMan12:
Also, please delete the comments (the stuff between `<!--` and `-->`) from the PR template (as mentioned in the PR template text)

xrmzju:
> Also, please delete the comments (the stuff between `<!--` and `-->`) from the PR template (as mentioned in the PR template text)

done

DirectXMan12:
editted title to reflect current state

/lgtm
/approve

PR #651, opened 2019-10-19
:bug: prevent controller.New and manager's errChan from leaking goroutines

### LINKED ISSUE #649: Clean up controller.New workqueue goroutine leaks
When we create a workqueue in the internals of controller.New, it spins up some goroutines (which can then be shut down with Shutdown).

We should not do this, and delay creating the workqueue till we start the controller, so that we don't leak goroutines.

We'll have to also delay watch starting until we start the controller (since watch start takes a handle to the queue), but that seems sensible anyway.

DirectXMan12:
This prevents a couple of goroutine leaks related to the controller and manager logic.

The first is in controller: `controller.New` previously created a new workqueue, which launched goroutines expecting to be stopped with `Shutdown`.  This is semi-unexpected in CR, since we generally state that things don't happen until Start is called (and, indeed, we don't call shutdown unless start is called).  This means that calling `controller.New` without ever starting the controller will cause a goroutine leak.  We fix this by delaying workqueue initialization (and consequently source starting) until the controller's Start is called.

The second is in the manager's handling of errors -- we used to use an error channel to report errors from internal goroutines and runnables.  However, we only ever read the first error from the channel, and it was a non-buffered channel.  This meant that if multiple things tried to return results on the channel, all but one would block forever.  This was exacerbated by the fact that the goroutines that launched controllers tried to return all results (nil values) on the channel, not just errors.  We fix this by using a mutex to save a single error, using a channel's closing to signal shutdown *only*, and by only ever returning non-nil errors via the error signaler.  This means that the only blocking is ever on a short-duration mutex.

Fixes #649

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -455,6 +455,9 @@ func (cm *controllerManager) startNonLeaderElectionRunnables() {
 			if err := ctrl.Start(cm.internalStop); err != nil {
 				cm.errSignal.SignalError(err)
 			}
+			// we use %T here because we don't have a good stand-in for "name",
+			// and the full runnable might not serialize (mutexes, etc)
+			log.V(1).Info("non-leader-election runnable finished", "runnable type", fmt.Sprintf("%T", ctrl))
 		}()
 	}
 }
@@ -474,6 +477,9 @@ func (cm *controllerManager) startLeaderElectionRunnables() {
 			if err := ctrl.Start(cm.internalStop); err != nil {
 				cm.errSignal.SignalError(err)
 			}
+			// we use %T here because we don't have a good stand-in for "name",
+			// and the full runnable might not serialize (mutexes, etc)
+			log.V(1).Info("leader-election runnable finished", "runnable type", fmt.Sprintf("%T", ctrl))
 		}()
 	}
 
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -114,7 +114,11 @@ type controllerManager struct {
 	started        bool
 	startedLeader  bool
 	healthzStarted bool
-	errChan        chan error
+
+	// NB(directxman12): we don't just use an error channel here to avoid the situation where the
+	// error channel is too small and we end up blocking some goroutines waiting to report their errors.
+	// errSignal lets us track when we should stop because an error occurred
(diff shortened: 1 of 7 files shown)


DirectXMan12:
cc @droot

DirectXMan12:
FYI, there's a subtle change in behavior here.  I think it's the correct one and the old one was a bug, but we used to send nil results from controllers on the error channel, meaning a returned controller is treated as an error.  This should never happen except in graceful shutdown, but it's still a behavior change.

pkg/manager/internal.go
```diff
@@ -150,6 +154,45 @@ type controllerManager struct {
 	retryPeriod time.Duration
 }
 
+type errSignaler struct {
+	// errSignal indicates that an error occurred, when closed.  It shouldn't
+	// be written to.
+	errSignal chan struct{}
```
mengqiy: Why it's `chan struct{}` instead of `chan error`?
`nil` can be treated as `error`, right?

pkg/manager/internal.go
```diff
@@ -150,6 +154,45 @@ type controllerManager struct {
...
+
+	// err is the received error
+	err error
```
mengqiy: Does it make sense to change it to `[]error`?
In `SignalError()`, we append errors.
In `Error()`, we aggregate errors into one error.

pkg/manager/internal.go
droot: Since this is being used purely for signaling that some error has occurred, it is fine.

pkg/manager/internal.go
droot: hmm...since mgr bails out on seeing first error, converting to slice of errors is not going to be of much help.

pkg/manager/internal.go
```diff
@@ -150,6 +154,45 @@ type controllerManager struct {
...
+
+	mu sync.Mutex
+}
+
+func (r *errSignaler) SignalError(err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
```
droot: we should handle if this gets called with `nil` err. That can lead to calling `close` on a closed channel and cause panic.

DirectXMan12: fixed

PR #653, opened 2019-10-21
:bug: pkg/log/zap: use enabled in UseDevMode, add tests

joelanford:
This PR fixes a bug where `func UseDevMode(enabled bool)` always sets the `Development` option to `true`, regardless of the value of `enabled`.

This PR also adds tests for the `New()` function and the `Opts`-related functions.


DirectXMan12:
/lgtm
/approve

thanks!

PR #655, opened 2019-10-22
:bug: Fix builder `For` and `Watches` method conflict

willise:
If `For` method has been used, there is no way to specify own eventhandler. But if `Watches` method is used, an error of lacking object will happend when no apiType is passed.
If both `For` and `Watches` are used with same apiType, the later one will be ignored as it has been watched. Just add nil assertion to avoid object type error.

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -157,11 +157,13 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
 
 func (blder *Builder) doWatch() error {
 	// Reconcile type
-	src := &source.Kind{Type: blder.apiType}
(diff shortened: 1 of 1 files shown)


DirectXMan12:
this seems like a bit of a footgun now, since it'd be pretty easy to skip the `For` call by accident.  What's the usecase here?  I'd probably say if you don't have a `For`, you should just call `controller.New` directly...

willise:
I do think the `builder` use more convenient :-).
The usecase is that I want to pass my own event handler like `EnqueueRequestsFromMapFunc` so I have to call `Watch` func. This func requires to set apiType too.  However, if I have set `For`,  the aipType has been already set. So in the `blder.doWatch`,  my own handler will not work as the `ctrl.Watch` has been called before with the same apiType.

PR #658, opened 2019-10-24
:sparkles: webhook/TLS: use a read lock for new connections

JulienBalestra:
This allows to have multiple new client connections at the same time, without locking.

The certificate updater, use the regular lock for writes.

### DIFF
--- a/pkg/webhook/internal/certwatcher/certwatcher.go
+++ b/pkg/webhook/internal/certwatcher/certwatcher.go
@@ -30,7 +30,7 @@ var log = logf.RuntimeLog.WithName("certwatcher")
 // changes, it reads and parses both and calls an optional callback with the new
 // certificate.
 type CertWatcher struct {
-	sync.Mutex
+	sync.RWMutex
 
 	currentCert *tls.Certificate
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Do you have performance numbers on this?  It makes sense (since the cert is updated relatively rarely, so we've got very few writes), but I'm curious if you were seeing actual contention.

DirectXMan12:
Also, can you clarify the PR title a bit to describe what the change actually is?

JulienBalestra:
@DirectXMan12 I doubt there is any real/noticeable performance improvement with this change.

In the other, it is an optimisation with a very small technical cost/change.

JulienBalestra:
@DirectXMan12 Do you think this PR still has some value regarding the changes involved ?

DirectXMan12:
I'd probably fine, but I'd prefer to see some actual numbers or a reason to do this -- rwlocks are occasionally more tricky performance wise than you might think, and I don't think we've seen any issues about this.

DirectXMan12:
closing till we get some data.

PR #659, opened 2019-10-25
:sparkles: envtest set service-cluster-ip-range if user not set

### LINKED ISSUE #641: envtest needs to set service-cluster-ip-range
service-cluster-ip-range is no longer going to be defaulted after 1.16 (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#deprecations-and-removals), so we'll need to set it in envtest for 1.17 compat.  There's no harm in setting it now, AFAICT

/kind feature
/priority important-soon
/good-first-issue

Dingshujie:
fix #641 
service-cluster-ip-range is no longer going to be defaulted after 1.16, so if user not set service-cluster-ip-range, add default value(10.0.0.0/24) to envtest set


DirectXMan12:
please prove a PR description beyond just the issue number

Dingshujie:
> please prove a PR description beyond just the issue number
sorry, already update PR description

PR #663, opened 2019-10-28
✨ Take context when getting informer.

### LINKED ISSUE #562: Create timeout and error when informer set up in the cache is hanging

djzager:
Follow up to #580 
Fixes #562 

If the context's deadline is exceeded waiting for the cache to sync, then we should return an appropriate error.

This intentionally punts on what a sensible default should be for this timeout based on [this comment](https://github.com/kubernetes-sigs/controller-runtime/pull/580#discussion_r320429256)
> Let's punt on the default for now (do it in a separate PR) so we can at least get the capability in and unbreak people that want to use this.

**What this PR needs**

- [x] Testing. Currently pursuing faking out the  `HasSynced` method.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -261,6 +261,20 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					err := informerCache.Get(context.Background(), svcKey, svc)
 					Expect(err).To(HaveOccurred())
 				})
+
+				It("should return an error when context is cancelled", func() {
+					By("creating a context and cancelling it")
+					ctx, cancel := context.WithCancel(context.Background())
+					cancel()
+
+					By("listing pods in test-namespace-1 with a cancelled context")
+					listObj := &kcorev1.PodList{}
+					err := informerCache.List(ctx, listObj, client.InNamespace(testNamespaceOne))
+
+					By("verifying that an error is returned")
+					Expect(err).To(HaveOccurred())
+					Expect(errors.IsTimeout(err)).To(BeTrue())
+				})
 			})
 			Context("with unstructured objects", func() {
 				It("should be able to list objects that haven't been watched previously", func() {
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -57,7 +57,7 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 		return err
 	}
 
-	started, cache, err := ip.InformersMap.Get(gvk, out)
+	started, cache, err := ip.InformersMap.Get(ctx, gvk, out)
 	if err != nil {
 		return err
 	}
@@ -76,7 +76,7 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
 		return err
 	}
 
-	started, cache, err := ip.InformersMap.Get(*gvk, cacheTypeObj)
+	started, cache, err := ip.InformersMap.Get(ctx, *gvk, cacheTypeObj)
 	if err != nil {
 		return err
 	}
@@ -128,7 +128,6 @@ func (ip *informerCache) objectTypeForListObject(list runtime.Object) (*schema.G
 	}
 
 	return &gvk, cacheTypeObj, nil
-
 }
 
 // GetInformerForKind returns the informer for the GroupVersionKind
@@ -138,7 +137,10 @@ func (ip *informerCache) GetInformerForKind(gvk schema.GroupVersionKind) (Inform
 	if err != nil {
 		return nil, err
 	}
-	_, i, err := ip.InformersMap.Get(gvk, obj)
+
+	// TODO(djzager): before a context can be passed down, the Informers interface
+	// must be updated to accept a context when getting an informer
+	_, i, err := ip.InformersMap.Get(context.TODO(), gvk, obj)
 	if err != nil {
 		return nil, err
 	}
@@ -151,7 +153,10 @@ func (ip *informerCache) GetInformer(obj runtime.Object) (Informer, error) {
 	if err != nil {
 		return nil, err
 	}
-	_, i, err := ip.InformersMap.Get(gvk, obj)
+
+	// TODO(djzager): before a context can be passed down, the Informers interface
+	// must be updated to accept a context when getting an informer
+	_, i, err := ip.InformersMap.Get(context.TODO(), gvk, obj)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -17,11 +17,13 @@ limitations under the License.
 package internal
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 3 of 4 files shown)


pkg/cache/internal/informers_map.go
```diff
@@ -162,9 +164,22 @@ func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Obj
 	}
 
 	if started && !i.Informer.HasSynced() {
+		syncReturn := make(chan bool)
```
DirectXMan12: make this buffered so that we don't leak the goroutine when we time out

```suggestion
		syncReturn := make(chan bool, 1 /* don't leak goroutines on timeout */)
```

pkg/cache/internal/informers_map.go
```diff
@@ -162,9 +164,22 @@ func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Obj
...
+			syncReturn <- cache.WaitForCacheSync(done, i.Informer.HasSynced)
+		}()
+
 		// Wait for it to sync before returning the Informer so that folks don't read from a stale cache.
-		if !cache.WaitForCacheSync(ip.stop, i.Informer.HasSynced) {
-			return started, nil, fmt.Errorf("failed waiting for %T Informer to sync", obj)
+		select {
+		case <-ctx.Done():
+			//end the polling for cache to sync
+			done <- struct{}{}
```
DirectXMan12:
```suggestion
			close(done)
```

pkg/cache/internal/informers_map.go
```diff
@@ -162,9 +164,22 @@ func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Obj
 	}
 
 	if started && !i.Informer.HasSynced() {
+		syncReturn := make(chan bool)
+		done := make(chan struct{})
+		go func() {
+			syncReturn <- cache.WaitForCacheSync(done, i.Informer.HasSynced)
```
DirectXMan12: actually, why isn't think whole thing just 

```go
cache.WaitForCacheSync(ctx.Done(), i.Informer.HasSynced)
```

plus checking for the timeout?

We probably want to wire up `ip.stop` too via a `WithCancel` or somesuch.

djzager: The first part of this comment seems totally reasonable to me. My gut says that the, "plus checking for the timeout?" is handled by leaving the select alone. Unfortunately though, the part about wiring up `ip.stop` I don't understand well enough to act on it.

DirectXMan12: > plus checking for the timeout

I just meant checking the error from the context, IIRC

wrt to `ip.stop` -- we want to stop when one of the following things happens:

- the timeout fires from the context
- `ip.stop` is closed

the former we get for free from passing the context's done to WaitForCacheSync.  The latter I think we need something like

```go
// at the top
ctx, cancel := context.WithCancel(ctx)
defer cancel()

go func() {
  select {
    case <-ip.stop:
      cancel()
    case <- ctx.Done():
  }
}

// later
cache.WaitForSync(ctx.Done(), i.Informer.HasSynced)
```

pkg/cache/internal/informers_map.go
```diff
@@ -162,9 +164,19 @@ func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Obj
...
 		// Wait for it to sync before returning the Informer so that folks don't read from a stale cache.
-		if !cache.WaitForCacheSync(ip.stop, i.Informer.HasSynced) {
-			return started, nil, fmt.Errorf("failed waiting for %T Informer to sync", obj)
+		select {
+		case <-ctx.Done():
+			return started, nil, apierrors.NewTimeoutError(fmt.Sprintf("timeout waiting for %T Informer to sync", obj), 0)
+		case syncSuccess := <-syncReturn:
+			if !syncSuccess {
+				return started, nil, fmt.Errorf("failed waiting for %T Informer to sync", obj)
+			}
```
joelanford: Is there a race condition here between `cache.WaitForCacheSync()` and the `<-ctx.Done()` select case?

When `ctx.Done()` is closed, `cache.WaitForCacheSync()` returns false, and we send that boolean value to the `syncReturn` channel. That effectively ties `ctx.Done()` to `syncReturn`, which means our select block cases are waiting for the same condition.

Could this be simplified to the following?
```go
canceled := cache.WaitForCacheSync(ctx.Done(), i.Informer.HasSynced)
if canceled {
	return started, nil, apierrors.NewTimeoutError(fmt.Sprintf("timeout waiting for %T Informer to sync", obj), 0)
}
```

DirectXMan12:
hey, are you still interested in working on this?

djzager:
> hey, are you still interested in working on this?

Yes I am. Apologies for letting this fall off my radar. Will take today to catch up on the comments and start adding the test.

pkg/cache/internal/informers_map.go
djzager: I implemented this. What you said makes sense.

pkg/cache/internal/informers_map.go
djzager: I believe that implementing the change that @joelanford suggested resolves this.

vincepri:
/ok-to-test
@djzager Can you squash commits down to 1?

djzager:
We are no longer adding a timeout to an unset context. Updated the commit and PR title to reflect that change.

pkg/cache/cache_test.go
```diff
@@ -261,6 +261,19 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+					By("creating a context and cancelling it")
+					ctx, cancel := context.WithCancel(context.Background())
+					cancel()
+
+					By("listing pods in test-namespace-1 with a cancelled context")
+					listObj := &kcorev1.PodList{}
+					err := informerCache.List(ctx, listObj, client.InNamespace(testNamespaceOne))
+
+					By("verifying that an error is returned")
+					Expect(err).To(HaveOccurred())
```
alvaroaleman: I've made very bad experience with `check err is not nil` tests, can we check we actually get the specific error we wanted? Just checking the error string should be good

pkg/cache/cache_test.go
```diff
@@ -261,6 +261,20 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+					Expect(err.Error()).To(Equal(fmt.Sprintf("Timeout: failed waiting for %T Informer to sync", &kcorev1.Pod{})))
```
DirectXMan12: In case this comes up in posterity: error messages are generally not part of our api guarantees.

In cases like this we should be checking `errors.Is(err, context.DeadlineExceeded)`, which is a more type-checkable thing that we actually can guarantee.

I generally think of tests as doubling as the human-readable part of the API contract, and so it's good to stick as close to what we guarantee as possible

pkg/cache/informer_cache.go
```diff
@@ -120,7 +120,8 @@ func (ip *informerCache) GetInformerForKind(gvk schema.GroupVersionKind) (Inform
 	if err != nil {
 		return nil, err
 	}
-	_, i, err := ip.InformersMap.Get(gvk, obj)
+
+	_, i, err := ip.InformersMap.Get(context.TODO(), gvk, obj)
```
DirectXMan12: can you add a TODO here explaining what we'd need to do to plumb this through?

djzager: My best attempt at an explanation. WDYT?

DirectXMan12: sure, looks good.  Do you mind putting together a follow-up to do that?

Issue #429, opened 2019-05-14
Bug in manager_test


mengqiy:
/kind bug

https://github.com/kubernetes-sigs/controller-runtime/blob/d6324a464299673cd0f8edd456f7a9f62fd22c90/pkg/manager/manager_test.go#L254-L286

This piece of code should expect an error in the end but actually it's not.

This test case should not panic, it should not use `GinkgoRecover`.

Probably there is a bug in the test.

mengqiy:
cc: @droot @DirectXMan12

DirectXMan12:
You almost always want to use GinkgoRecover when spawing goroutines (which a runnable is), and just about anything *can* panic (yay side effects not captured by the type system).  In this case, for instance, if that channel is somehow nil, the test could panic.

vincepri:
Closing in favor of https://github.com/kubernetes-sigs/controller-runtime/issues/764

/close

PR #664, opened 2019-10-29
:bug: Wait for runnables to stop fix for #350 and #429

### LINKED ISSUE #350: Manager does not wait for Runnables to stop

dbenque:
This PR  fixes :bug: #350  and :bug: #429 

The manager.Start function now returns only when all Runnables have properly returned or timeout.

It is possible to define the Timeout value with the manager.Options

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -48,6 +48,8 @@ const (
 	defaultRenewDeadline = 10 * time.Second
 	defaultRetryPeriod   = 2 * time.Second
 
+	defaultRunnableTearDownTimeout = 10 * time.Second
+
 	defaultReadinessEndpoint = "/readyz"
 	defaultLivenessEndpoint  = "/healthz"
 )
@@ -114,11 +116,7 @@ type controllerManager struct {
 	started        bool
 	startedLeader  bool
 	healthzStarted bool
-
-	// NB(directxman12): we don't just use an error channel here to avoid the situation where the
-	// error channel is too small and we end up blocking some goroutines waiting to report their errors.
-	// errSignal lets us track when we should stop because an error occurred
-	errSignal *errSignaler
+	errChan        chan error
 
 	// internalStop is the stop channel *actually* used by everything involved
 	// with the manager as a stop channel, so that we can pass a stop channel
@@ -129,6 +127,9 @@ type controllerManager struct {
 	// internalStopper is the write side of the internal stop channel, allowing us to close it.
 	// It and `internalStop` should point to the same channel.
 	internalStopper chan<- struct{}
+
+	// stop procedure engaged. In other words, we should not add anything else to the manager
+	stopProcedureEngaged bool
 
 	startCache func(stop <-chan struct{}) error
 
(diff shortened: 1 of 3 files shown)


alvaroaleman:
@dbenque can you fix the conflicts?

dbenque:
@alvaroaleman the PR has been rebase and conflict resolved.

Note that I had to rework part of https://github.com/kubernetes-sigs/controller-runtime/commit/fe4ada0fc0ca8c7a97113349352f46bf529541be

This PR #664 is giving same result but it is not black-holing runnable errors:https://github.com/kubernetes-sigs/controller-runtime/commit/fe4ada0fc0ca8c7a97113349352f46bf529541be#diff-77faf6b20512574869434402d5c5b6a2R179

This is important because we want to wait for runnable to stop, and so we must catch and handle errors while runnable are stopping.

mengqiy:
This reverts many changes in #651
/assign @DirectXMan12

dbenque:
@DirectXMan12 regarding #651 , this PR is achieving the same and also catch all errors of runnables during the teardown period. There should not be any error silently dropped with that PR and at the same time we wait for all runnables stop (or timeout).
I hope that help for the review.

alexeldeib:
I like this approach. It seems to me like the error draining works nicely here. The same blocking mentioned in https://github.com/kubernetes-sigs/controller-runtime/pull/651#issue-329963182 can occur if more than one controller errors out around L392, but since the stop routine drains the channel while it handles proper shutdown, it won't actually block the controllers from exiting. I like this more than the error signaler, personally. 

One thing I noticed, the runnables are all wired up but manager brings up the metrics endpoint and healthprobes separately. Do we care about gracefully terminating those as well?

dbenque:
> One thing I noticed, the runnables are all wired up but manager brings up the metrics endpoint and healthprobes separately. Do we care about gracefully terminating those as well?

@alexeldeib , you are right and to avoid that on my side I explicitly disable the metrics `ManagerOptions.MetricsBindAddress = "0"` embedded in the manager: I have have create a dedicated `runnable` for metrics and also one for the healthprobes. Then I start them like any other `runnable` and benefit from the stop sequence implemented in that PR.

pkg/manager/internal.go
```diff
@@ -433,10 +406,50 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
...
+		return cm.engageStopProcedure(stopComplete)
+	case err := <-cm.errChan:
+		// Error starting or running a runnable
+		if errStop := cm.engageStopProcedure(stopComplete); errStop != nil {
+			log.Error(errStop, "some runnables could not be stopped after error occurred starting/running the manager.")
+		}
+		return err
+	}
+}
+func (cm *controllerManager) engageStopProcedure(stopComplete chan struct{}) error {
```
DirectXMan12: Godoc: https://www.startrek.com//sites/default/files/images/2019-01/d7b431b1a0cc5f032399870ff4710743.jpeg

pkg/manager/internal.go
```diff
@@ -433,10 +406,50 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
...
+				}
+			case <-stopComplete:
+				return
+			}
+		}
+	}()
+	return cm.waitForRunnableToEnd()
+}
+
+func (cm *controllerManager) waitForRunnableToEnd() error {
```
DirectXMan12: this whole section needs an overview comment of the stop procedure stuff

pkg/manager/internal.go
```diff
@@ -433,10 +406,50 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
...
+	runnableTearDownTimer := time.NewTimer(cm.runnableTearDownTimeout)
+	defer runnableTearDownTimer.Stop()
+	allStopped := make(chan struct{})
+
+	go func() {
+		cm.waitForRunnable.Wait()
```
DirectXMan12: this'll leak through a timeout

DirectXMan12: not sure there's a good way around it though

pkg/manager/internal.go
```diff
@@ -497,7 +503,7 @@ func (cm *controllerManager) waitForCache() {
 	}
 	go func() {
 		if err := cm.startCache(cm.internalStop); err != nil {
-			cm.errSignal.SignalError(err)
+			cm.errChan <- err
```
DirectXMan12: I kinda feel like we should never have anything writing to the error channel directly like this, and instead just wrap everything in a runnable to avoid accidentally forgetting to increment the runnable counter.

DirectXMan12:
(as a follow up PR for someone -- a test that ensured we don't add any additional leaked goroutines in new code would be nice -- just start the manager then stop it, and use `runtime` to check the goroutine count)

DirectXMan12:
(I'll file an issue)

DirectXMan12:
(#724)

DirectXMan12:
hey, are you still interested in working on this

vincepri:
@dbenque are you still interested in working on this change?

vincepri:
Closing for inactivity, feel free to reopen if necessary.

/close

PR #667, opened 2019-10-31
🐛 Fix informer cache creating pointers to pointers

### LINKED ISSUE #656: Unable to List Objects with pointer slices
I've been working on a controller that interacts with another project's CRDs. All was going well until I encountered an error when trying to list owned objects using the reconciler's client with the code below:
```goland
if err := r.List(ctx, &ownedRules, client.MatchingField(ruleOwnerKey, req.Name)); err != nil {
	log.Error(err, "unable to list owned rules")
	return ctrl.Result{}, err
}
```

This caused the following error: 
```
2019-10-23T12:41:43.568+0100    ERROR   controller-runtime.controller   Reconciler error        {"controller": "servicelevelobjective", "request": "default/energy-fastly", "error": "cannot get cache for *v1.PrometheusRuleList, its element **v1.PrometheusRule is not a runtime.Object"}
```

Looking over the definition of the `PrometheusRuleList` CRD I noted that the element type for the `Items` field is `*PrometheusRule`.
https://github.com/coreos/prometheus-operator/blob/v0.33.0/pkg/apis/monitoring/v1/types.go#L679

It appears the code in `controller-runtime/pkg/cache/informer_cache.go` tries to determine whether the items can be cast to `runtime.Object` and in the process assumes that the slice's elements are not pointers.

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.2.2/pkg/cache/informer_cache.go#L84-L85
```golang
elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
cacheTypeValue := reflect.Zero(reflect.PtrTo(elemType))
```

I've fixed this locally by adding a check before converting the elemType to a Ptr which seems sufficient to fix the error in my use case.
```golang
elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
if elemType.Kind() != reflect.Ptr {
	elemType = reflect.PtrTo(elemType)
}

cacheTypeValue := reflect.Zero(elemType)
```

I'm happy to put together a PR with this fix in it but wanted to raise the issue for discussion first in case there may be knock-on effects that I've not yet encountered as I haven't done extensive testing on my end yet.

Any thoughts from the owners on this one?

DewaldV:
Fixes #656

It appears the code in controller-runtime/pkg/cache/informer_cache.go tries to determine whether the items can be cast to runtime.Object and in the process assumes that the slice's elements are not pointers.

This means that if you have a List whose Items type is `[]*Item` the informer cache will create a `cacheTypeValue` of `**Item` which then cannot be case to `runtime.Object` resulting in an error of:

```
"error": "cannot get cache for *v1.PrometheusRuleList, its element **v1.PrometheusRule is not a runtime.Object"}
```

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -93,7 +93,11 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
 		}
 		// http://knowyourmeme.com/memes/this-is-fine
 		elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
-		cacheTypeValue := reflect.Zero(reflect.PtrTo(elemType))
+		if elemType.Kind() != reflect.Ptr {
+			elemType = reflect.PtrTo(elemType)
+		}
+
+		cacheTypeValue := reflect.Zero(elemType)
 		var ok bool
 		cacheTypeObj, ok = cacheTypeValue.Interface().(runtime.Object)
 		if !ok {


DewaldV:
Hmm, my org has a signed CLA so I expected that check to pass. I'll try and see what's up with that before proceeding here.

DewaldV:
Ok, that should be good now. Apologies for the lapse, I had missed a step when linking my GH account to our orgs.

alvaroaleman:
@DewaldV thanks for your PR. Could you add a test for this so we don't regress?
/ok-to-test

And more of a comment, but pretty much all `List` types I've seen use a list of literals, not a list of pointers. The [api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) don't mention anything about that so I guess its fine.

DewaldV:
@alvaroaleman Happy to, I'll get another commit with a test done in a bit.

DewaldV:
Apologies for the delay on this, I've got a basic spike with the tests local but I've been delayed due to some work pressures. I'll try and get some time this week to get the test cleaned up and pushed.

DirectXMan12:
hey, are you still interested in working on this

DewaldV:
Apologies for the delay on this.

The testing turned out to be rather complicated as it required creating some CRDs with Lists that use pointers so I stalled in Nov and then it dropped off my radar.

I'm going to put some time aside to get the tests finished up over the coming week so we can get this merged.

We do still need it.

alvaroaleman:
@DewaldV I ended up needing as well, do you mind if I take your commit from here, add another one with a test on top of it and create a new PR so this gets done in a timely manner?

DewaldV:
@alvaroaleman I'm happy with that. Appreciate you jumping in on this and apologies for not getting this merged before. :smile:

vincepri:
Closing this in favor of 796

/close

Issue #669, opened 2019-11-02
Envtest fails without a network connection


wallrj:
In https://github.com/improbable-eng/etcd-cluster-operator/issues/63 I noticed that our test suite would fail if I was offline.

Looking through the attached control plane logs I can see that the kube-apiserver fails to start if the machine doesn't have a default route set.
Part of the automatic  `--advertise-address` lookup mechanism.
One solution is to supply a fixed `--advertise-address`.

```
richard   master  ~  projects  kubernetes-sigs  controller-runtime  130  KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true KUBEBUILDER_ASSETS=$PWD/bin go test -v ./pkg/envtest/...
=== RUN   TestSource
Running Suite: EnvTest Suite
============================
Random Seed: 1572698196
Will run 5 of 5 specs

2019-11-02 12:36:36.633433 I | etcdmain: etcd Version: 3.3.11
...
Error: unable to find suitable network address.error='no default routes found in "/proc/net/route" or "/proc/net/ipv6_route"'. Try to set the AdvertiseAddress directly or provide a valid BindAddress to fix this
...
2019-11-02 12:37:16.968894 C | etcdmain: listen tcp 127.0.0.1:43979: bind: address already in use
2019-11-02T12:36:36.621Z	DEBUG	controller-runtime.test-env	starting control plane	{"api server flags": ["--etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}", "--cert-dir={{ .CertDir }}", "--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}", "--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}", "--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}", "--admission-control=AlwaysAdmit"]}
2019-11-02T12:36:56.954Z	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 0, "error": "timeout waiting for process kube-apiserver to start"}
...
2019-11-02T12:37:16.955Z	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 1, "error": "timeout waiting for process etcd to start"}
github.com/go-logr/zapr.(*zapLogger).Error
...
Timeout [60.000 seconds]
[BeforeSuite] BeforeSuite 
[...]

PR #670, opened 2019-11-02
:bug: Allow kube-apiserver to start without a default route

wallrj:
Allows envtest based tests to run offline.

Fixes: #669


DirectXMan12:
heh, good catch

/approve

PR #674, opened 2019-11-07
🐛 Fix delete example not setting metadata properly for CRDs

vadasambar:
`name` and `namespace` should go into the metadata of the `unstructured.Unstructured` object.


DirectXMan12:
good catch.  Switched to :book: because it's a docs bug.

/lgtm
/approve

PR #675, opened 2019-11-07
🐛 cross-namespace owner references should be disallowed

boylee1111:
Based on https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents, cross-namespace owner references are disallowed by design. The `SetControllerReference()` should prevent owner reference from setting on dependents in different namespace.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -60,6 +60,17 @@ func SetControllerReference(owner, object metav1.Object, scheme *runtime.Scheme)
 		return fmt.Errorf("%T is not a runtime.Object, cannot call SetControllerReference", owner)
 	}
 
+	ownerNs := owner.GetNamespace()
+	if ownerNs != "" {
+		objNs := object.GetNamespace()
+		if objNs == "" {
+			return fmt.Errorf("cluster-scoped resource must not have a namespace-scoped owner, owner's namespace %s", ownerNs)
+		}
+		if ownerNs != objNs {
+			return fmt.Errorf("cross-namespace owner references are disallowed, owner's namespace %s, obj's namespace %s", owner.GetNamespace(), object.GetNamespace())
+		}
+	}
+
 	gvk, err := apiutil.GVKForObject(ro, scheme)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -60,6 +60,10 @@ func SetControllerReference(owner, object metav1.Object, scheme *runtime.Scheme)
 		return fmt.Errorf("%T is not a runtime.Object, cannot call SetControllerReference", owner)
 	}
 
+	if owner.GetNamespace() != "" && owner.GetNamespace() != object.GetNamespace() {
+		return fmt.Errorf("cross-namespace owner references are disallowed, owner's namespace %s, obj's namespace %s", owner.GetNamespace(), object.GetNamespace())
```
alvaroaleman: This happens to (accidentally?) also cover the case of "cluster-scoped resource must not have a namespace-scoped owner". However in that case the error is confusing, because its not a cross-namespace ref.

Would you mind returning a different error message in case `object.GetNamesapce() == ""`?

boylee1111: my intention was to cover both cases. I had separated them now, it's more clear. thx for suggestion.

boylee1111:
Sorry, I accidentally pushed something to this pr, I had removed it now. PTAL.

DirectXMan12:
/lgtm
/approve

thanks!

FaKod:
I stumbled upon this PR while migrating from Operator Framework to Kubebuilder (there is no such limitation).

I understand why this cross namespace limitation makes sense for standard K8s resources like StatefulSets or Deployments. However, why has this to be **enforced** in a general framework, where also more abstract functionalities are implemented? 

In my case an operator creates developer workspaces, implemented as Namespaces and a Pod in each of them. These workspaces are defined by a namespaced CRs in the operator namespace. I do not see any requirements to introduce a cluster CRD for that.

alvaroaleman:
@FaKoD this is a limitation of Kubernetes itself, check the link in the PR description

FaKod:
@alvaroaleman A call to `SetControllerReference` worked fine with the Operator framework, even if the CR and the resource to watch are of different namespaces. Thats what this PR is disallowing, right?

alvaroaleman:
Yes, prior to this PR it was possible to create a defunct owner Ref.

PR #678, opened 2019-11-11
:bug: InformerCache: Do not require leader lease

### LINKED ISSUE #677: Cache-Backed reader only works when the leader lease was acquired
Currently, the cache-backed reader returned from the `Manager` only works when the leader lease was acquired, because otherwise it doesn't get started.

There are cases in which it is useful to have it working even without the leader lease, basically every runnable that itself doesn't require leader election.

alvaroaleman:
This PR makes the `informerCache` implement the `LeaderElectionRunnable` interface to signal that it doesn't need leader election. This is done so other `runnables` that also do not need leader election can use the cache-backed reader.

Fixes #677

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -140,6 +140,12 @@ func (ip *informerCache) GetInformer(obj runtime.Object) (Informer, error) {
 	return i.Informer, err
 }
 
+// NeedLeaderElection implements the LeaderElectionRunnable interface
+// to indicate that this can be started without requiring the leader lock
+func (ip *informerCache) NeedLeaderElection() bool {
+	return false
+}
+
 // IndexField adds an indexer to the underlying cache, using extraction function to get
 // value(s) from the given field.  This index can then be used by passing a field selector
 // to List. For one-to-one compatibility with "normal" field selectors, only return one value.
--- a/pkg/cache/informer_cache_test.go
+++ b/pkg/cache/informer_cache_test.go
@@ -1,0 +1,28 @@
+package cache_test
+
+import (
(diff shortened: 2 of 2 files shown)


pkg/cache/informer_cache.go
```diff
@@ -140,6 +140,12 @@ func (ip *informerCache) GetInformer(obj runtime.Object) (Informer, error) {
 	return i.Informer, err
 }
 
+// NeedLeaderElection implements the LeaderElectionRunnable interface
```
alvaroaleman: Unfortunately we can not verify the interface implementation via a `var _ manager.LeaderElectionRunnable = &informerCache{}` as that would result in an import cycle

pkg/cache/informer_cache_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+var _ = Describe("informerCache", func() {
+	It("should not require LeaderElection", func() {
+		cfg := &rest.Config{}
+
+		mapper, err := apiutil.NewDynamicRESTMapper(cfg, apiutil.WithLazyDiscovery)
+		Expect(err).ToNot(HaveOccurred())
+
+		c, err := cache.New(cfg, cache.Options{Mapper: mapper})
+		Expect(err).ToNot(HaveOccurred())
```
alvaroaleman: I couldn't find a matcher for "type X implements interface Y", the closes to that I found is `BeAssignableToTypeOf` but that errors out if one of the two is nil.

DirectXMan12: this is clear enough.  If you're really worried, you can always add an explanation after the "be true" part as a format string, but I think this is readable enough for now.

Issue #687, opened 2019-11-20
Ability to create watches for different namespaces


lulf:
We have a use case where we have 2 namespaced resources, and a controller watching resources of type A in all namespace, but it would also like to watch resources of type B for a single specific namespace (for instance based on which namespace an instance of 'A' exists). At present, this is not possible to watch instances of B in a specific namespace. The alternative of watching resource B in all namespaces is not desired (lets say B is the Secret type, and we would like to avoid overly wide privileges for the operator).

Right now namespace is configured at the manager level, but ideally for our use case, we should be able to specify a namespace when watching resources, similar to what was discussed here: https://github.com/kubernetes-sigs/controller-runtime/issues/124#issuecomment-417438810

What would be required to allow the above, and is a contribution of that something that would get accepted?

alvaroaleman:
IMHO this is a duplicate of #244 . Feel free to re-open if you disagree.
/close

lulf:
@alvaroaleman Reading that issue, I agree its similar, but my main requirement is to avoid the need to grant more permissions to the controller than what is needed. I.e. filtering by namespace client-side would not be sufficient. If you want to keep this closed, I can perhaps add a comment to that issue instead.

Issue #244, opened 2018-12-12
Enable filtered list watches as watches


nkvoll:
When setting up watches during initialization it's currently not possible to filter by any selectors (which is possible using list watches).
    
For example it is not possible to only watch pods with specific labels (e.g having the label `pod-type: my-controller-type`). The current behavior results in very broad caching, which might not be desirable for large Kubernetes deployments.

In some scenarios an operator could contain multiple controllers, and they all share caches, so keying caches on GVK's alone might be problematic if they want to watch the same resource type, but with different filters.

When doing `List`/`Get`, how would one decide which of the caches to use? It seems that perhaps this needs to be an explicit choice by the operator developers?

jeefy:
+1

I've run into a similar issue with a controller I'm writing. I wound up starting a goroutine when the controller gets added to watch/act on certain objects that are created and labeled by another controller outside of mine.

DirectXMan12:
we'd have to specify this at the manager level, but it should be possible to pass restricting items.

DirectXMan12:
(since caches are shared)

DirectXMan12:
we could also let users manually specify auxiliary caches

DirectXMan12:
if anyone has ideas for a design doc, I'd love to see them.

DirectXMan12:
/kind feature
/priority important-soon

this has been coming up a lot lately, and seems like a good thing to tackle in the list of related issues.

DirectXMan12:
i.e. one sketch:

Have a list-options watch predicate, smartly detect the presence of that (in some generic way) and pass it down to `GetInformer`.  If we already have a global informer, just use that.  If we don't have a global informer, create a new filtered informer.  We delay creation of informers to figure out if one person needs a global informer, and just use that, to allow splitting and combining seamlessly.  The question is how to deal with the cached client:

- :x:  cached client always tries for a global informer (seems like a performance footgun)
- :x: only one filtered informer max (breaks multi-controller managers)
- :x: cache client refuses to use global informer if filtered informers exist and a global informer does not (breaks multi-controller informers)
- :ok:  keep track of which controller a given client belongs to and which watches that client set up, and try to scope those gets to the informer it belongs to, in the case of filtered informers (complicated, serious internal refactors)
[...]

PR #692, opened 2019-11-22
✨ Add support for creating namespaced watches

lulf:
Extends the Source type with an optional Namespace parameter, which will override the default namespace by adding the namespace as a key in the informer map. Existing usage will remain as is, sharing the same key as before, whereas if namespace is specified, a new informer will be created for that case. 

This allows creating an operator watching resources in multiple namespaces without requiring cluster-wide privileges for that resource, as desired in #687 .

Simple tests suggest that it works for me, but there are probably pitfalls I have not considered.

I'd love some feedback on this approach and if you think this is the right or wrong direction.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -53,9 +53,17 @@ type Informers interface {
 	// API kind and resource.
 	GetInformer(obj runtime.Object) (Informer, error)
 
+	// GetInformer fetches or constructs an informer for the given object that corresponds to a single
+	// API kind and resource.
+	GetInformerInNamespace(obj runtime.Object, namespace *string) (Informer, error)
+
 	// GetInformerForKind is similar to GetInformer, except that it takes a group-version-kind, instead
 	// of the underlying object.
 	GetInformerForKind(gvk schema.GroupVersionKind) (Informer, error)
+
+	// GetInformerForKindInNamespace is similar to GetInformerInNamespace, except that it takes a group-version-kind, instead
+	// of the underlying object.
+	GetInformerForKindInNamespace(gvk schema.GroupVersionKind, namespace *string) (Informer, error)
 
 	// Start runs all the informers known to this cache until the given channel is closed.
 	// It blocks.
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -57,7 +57,7 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 		return err
 	}
 
-	started, cache, err := ip.InformersMap.Get(gvk, out, &key.Namespace)
+	started, cache, err := ip.InformersMap.Get(gvk, out, nil)
 	if err != nil {
 		return err
 	}
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -57,7 +57,7 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 		return err
 	}
 
-	started, cache, err := ip.InformersMap.Get(gvk, out)
+	started, cache, err := ip.InformersMap.Get(gvk, out, &key.Namespace)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


lulf:
@alvaroaleman Thanks for taking a look. 

> Edit: Also thinking about this, I am not sure if this really make sense because the client will still create a non-namespaced watch once you start using it for the given GVK. This means that you end up with both a namespaced and a non-namespaced watch and use the former for the client and the latter for the watch. This effectively makes the whole thing have the same effect as a Predicate except that you create another cache.

I'm very open to alternative approaches. The main motivation is the ability to create watches in both the 'controller namespace' as well as watches for other namespaces, without requiring cluster-wide access to a particular resource. For example, with this PR, I'm able to create a manager watching deployments in any namespace, but for secrets, I can create a watch in a specific namespace (because I don't want to give my controller access to secrets all over the cluster).

Its not 100% clear to me what you mean by a non-namespaced watch. It doesn't seem to create a non-namespaced watch in my testing (at least it doesn't throw any permission error), could you point me to where this happens?

Happy to add some tests for this, just want to get some feedback on the approach and suggestions.

Thanks again!

alvaroaleman:
> Its not 100% clear to me what you mean by a non-namespaced watch. It doesn't seem to create a non-namespaced watch in my testing (at least it doesn't throw any permission error), could you point me to where this happens?

Seems I was too quick with the comment, as you updated `*informerCache.Get` to use the namespace when getting  an informer. However that means that we create an informer for every namespace from which we requested something if the object is namespaced and never use a non-namespaced informer. This results in a lot of cache duplication when someone uses a non-namespaced version of `source.Kind`.

Setting the `namespace` on the `mgr` does not work for your use-case?

lulf:
@alvaroaleman Yes, it would lead to cache duplication, that is the 'cost' of the added flexibility I guess, but I think it only impact users of this functionality, and existing usage should remain as is. Maybe if there was a smarter way to share cache between informers for different namespaces...

Setting namespace on the mgr does not cover my use case, because it allows only watching resources in a single namespace. The specific flexibility I need is the ability to watch some resources (that the controller "owns", i.e. CRD) in all namespaces, and then watch resources created by the controller in the same namespace (such as secrets, deployments etc). The main point here is that I don't want my controller to read secrets in other namespaces than the same as where the controller is running.

I'm looking at your initial idea, maybe changing GetInformers to

```GetInformers(runtime.Object, opts...client.ListOption)``` and somehow base the informer lookup key on those options. In the same way, if options are not set, as today, no cache duplication is done, its only for the cases that actually set those options.

alvaroaleman:
> Yes, it would lead to cache duplication, that is the 'cost' of the added flexibility I guess, but I think it only impact users of this functionality, and existing usage should remain as is

I think it impacts everyone because whenever a namespaced object is retrieved, we create a namespaced cache for the given type even if we already have a non-namespaced one.

lulf:
> I think it impacts everyone because whenever a namespaced object is retrieved, we create a namespaced cache for the given type even if we already have a non-namespaced one.

Just pushed a change so that informer_cache.Get() passes nil as the namespace argument instead of object.key, which should make cache behavior identical to earlier.

What I was thinking is that the existing usages of `cache.GetInformerForKind*` pass `nil` as the namespace argument, which should give the exact same key in the informer map, as if the map was only keyed by GVK as before. The only code path that will cause cache duplication, is if namespace is set to something non-nil (i.e. in this case source.Namespace is set when creating the watch).

Does this address your concern?

alvaroaleman:
Yeah it addresses the concern but it unfortunately means this PR won't solve your problem anymore as it now uses the non-namespaced cache for reading again :grimace:

This is unfortunate a tricky problem, IMHO it would be best to continue discussing possible approaches in #244

lulf:
Good point. I will close this for now, but might have another go 😉

lulf:
@alvaroaleman I was unable to reopen this due to force-pushing, but I have raised #694 which takes a slighly different approach, and overrides the namespace if cache is configured to watch all namespace. Also, the interface is a bit different.

PR #694, opened 2019-11-25
✨ Add support for creating namespaced watches

lulf:
Another variant of #692 . Extend source.Kind with taking ListOptions, and pass these on to the informer map. In this PR, only namespace is used, but I think it could be extended to labels as well.

* If the informer cache is configured to handle all namespaces (i.e. "") - use it as key in all lookups to avoid cache duplication. This preserves existing lookup behavior.
* If the informer cache is configured to handle a specific namespace - create separate cache which will be used for lookups in those cases. This allows creating watches for multiple namespaces dynamically. 

I don't know yet if this works, its just a prototype of an API that I as a user would like to have supported. Will add an integration test, but early feedback is welcomed.

### DIFF
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -31,11 +31,12 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // clientListWatcherFunc knows how to create a ListWatcher
-type createListWatcherFunc func(gvk schema.GroupVersionKind, ip *specificInformersMap) (*cache.ListWatch, error)
+type createListWatcherFunc func(gvk schema.GroupVersionKind, namespace string, ip *specificInformersMap) (*cache.ListWatch, error)
 
 // newSpecificInformersMap returns a new specificInformersMap (like
 // the generical InformersMap, except that it doesn't implement WaitForCacheSync).
@@ -49,7 +50,7 @@ func newSpecificInformersMap(config *rest.Config,
 		config:            config,
 		Scheme:            scheme,
 		mapper:            mapper,
-		informersByGVK:    make(map[schema.GroupVersionKind]*MapEntry),
+		informers:         make(map[MapKey]*MapEntry),
 		codecs:            serializer.NewCodecFactory(scheme),
 		paramCodec:        runtime.NewParameterCodec(scheme),
 		resync:            resync,
@@ -57,6 +58,12 @@ func newSpecificInformersMap(config *rest.Config,
 		namespace:         namespace,
 	}
 	return ip
+}
+
+// MapKey is the key for looking up informers
+type MapKey struct {
+	gvk       schema.GroupVersionKind
+	namespace string
 }
 
 // MapEntry contains the cached data for an Informer
@@ -80,8 +87,8 @@ type specificInformersMap struct {
 	// mapper maps GroupVersionKinds to Resources
 	mapper meta.RESTMapper
 
-	// informersByGVK is the cache of informers keyed by groupVersionKind
-	informersByGVK map[schema.GroupVersionKind]*MapEntry
+	// informers is the cache of informers keyed by groupVersionKind and namespace
+	informers map[MapKey]*MapEntry
 
 	// codecs is used to create a new REST client
 	codecs serializer.CodecFactory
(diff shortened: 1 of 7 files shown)


pkg/cache/internal/informers_map.go
```diff
@@ -136,27 +143,42 @@ func (ip *specificInformersMap) Start(stop <-chan struct{}) {
...
-func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Object) (bool, *MapEntry, error) {
+func (ip *specificInformersMap) Get(gvk schema.GroupVersionKind, obj runtime.Object, opts ...client.ListOption) (bool, *MapEntry, error) {
+	// Set default list options and override
+	listOpts := client.ListOptions{
+		Namespace: ip.namespace,
+	}
+	listOpts.ApplyOptions(opts)
+
+	// If configured using a global namespace, use its cache to avoid duplication
+	if ip.namespace == "" {
```
alvaroaleman: So generally this looks like it will work and I like the overall idea of "use specific cache unless configured to use global one" :)

There are however some things to consider:
* If I read this correctly, using a `source.Kind` with a `Namespace` setting and a non-namespaced `specificInformersMap` will result in getting events from all namespaces, which is not expected
* The current approach means that as soon as any namespace is configured, caches will get established for all namespaces an object was ever requested from which is also unexpected
* When thinking about this approach in the context of label selectors I think just a binary "cache for everything if global cache, else distinct cache for every label selector ever used" is not enough, as it may lead to quite a bit of duplication again. Maybe a list of label selectors for which we want a distinct cache and then have the client internally either use the specific cache or do the label-selecting after receiving the objects from the cache?
* More of a nit, but I think it will be better to not re-use `client.ListOptions` for the `Get` on the cache map, as that forces us to keep their feature set on par


At the end of the day thought, this is the point where I can not give you any further feedback if the approach is feasible or has more issues I didn't consider, we will have to wait for @DirectXMan12 to chime in.

But thanks a lot for pushing this forward, its really great to see some movement here :)

shawn-hurley:
@lulf would this solve your issue?

https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#example-New--MultinamespaceCache

I think that this is a large enough feature with enough gotchas that it might be nice to write down the design for the proposed changes and maybe another option so that we can compare? I would use the 2 ok's [here](https://github.com/kubernetes-sigs/controller-runtime/issues/244#issuecomment-451336202) as a starting point.

lulf:
@shawn-hurley What I need specifically is a way to use a global cache for some objects, and a specific namespace for other resources. The multi-namespace cache would not give me the ability to select which cache to use when creating a watch for a particular resource.

That being said, I have attempted another approach to solve my use case, where I create a 'delegating cache', which delegates either to global or a default namespaced cache based on the object type : https://github.com/EnMasseProject/enmasse/blob/2db9469529254274c90117b1f6fad360b8081857/pkg/cache/delegate_cache.go

So, I think I have a workaround for now. I could generalize the delegate cache and submit a PR for that if you think that would be interesting.

shawn-hurley:
That delegating cache starts to sound like the first option in the list of OK's to me.

We would have to add more scenarios to cover because dynamic watches could change what is being watched during run time. We also would have to handle the case where 1 controller asks for a namespaced watch/cache BUT we already had that in the global lists. I think you would want to add a predicate for that at that time to do the correct filter for that watch. 

I think there are some other scenarios where we need to protect folks from scenarios where the behavior maybe not obvious. But it seems like it could be a good start! I second what @alvaroaleman said, thank you for driving this forward!

Is there any way to add what that code does to a proposal that we can iterate on in words to make sure we can cover these different scenarios?

lulf:
> I think there are some other scenarios where we need to protect folks from scenarios where the behavior maybe not obvious. But it seems like it could be a good start! I second what @alvaroaleman said, thank you for driving this forward!

I've been focusing our own needs for this functionality so far, as we had to investigate whether or not something like this could be done in controller-runtime and how much time it would take. With the delegate cache, we have a solution that allows us to simply implement the cache interface until a more generalized approach is available in controller-runtime, so I'm happy with that. However, I'd love to improve on that and make it into something that could be included in controller-runtime itself.

> Is there any way to add what that code does to a proposal that we can iterate on in words to make sure we can cover these different scenarios?

Is that best driven through #244 ? I can make a more general design document as a starting point, if desired.

shawn-hurley:
@lulf Sorry I missed the last comment. 

I think that #244 might be the best place and a more general design document will really help me think through the problem :) 

Thanks!

lulf:
Closing this for now based on feedback, thanks!

PR #696, opened 2019-11-26
:sparkles: update CRD if already present

### LINKED ISSUE #342: Envtest is not idempotent when installing CRDs, an issue when UseExistingCluster=true
We want to use envtest against an existing cluster and are having issues when installing CRDs. Envtest will fail on the second running of the tests because the CRDs already exist in the cluster.  This is only an issue when using an existing cluster because normally envtest starts a new control plane each time. 

Example setup we want to use:
```
			CRDDirectoryPaths:  []string{filepath.Join("..", "config", "crds")},
			UseExistingCluster: true,
		}
```

And error we get:
```  
Expected error:
      <*errors.StatusError | 0xc00012a2d0>: {
          ErrStatus: {
              ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
              Status: "Failure",
              Message: "customresourcedefinitions.apiextensions.k8s.io \"mycrd.meep.io\" already exists",
              Reason: "AlreadyExists",
              Details: {
                  Name: "mycrd.meep.io",
                  Group: "apiextensions.k8s.io",
                  Kind: "customresourcedefinitions",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
      customresourcedefinitions.apiextensions.k8s.io "mycrd.meep.io" already exists
  not to have occurred
```

Is it possible to have envtest reinstall CRDs on every run rather than fail?

Thanks
Ed & @samze & @ablease

pires:
In testenv, update the CRD if it's already present on the cluster instead of failing to continue.

Fixes #342
Closes #457

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -225,8 +225,19 @@ func CreateCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResource
 	// Create each CRD
 	for _, crd := range crds {
 		log.V(1).Info("installing CRD", "crd", crd.Name)
-		if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
-			return err
+		if existingCrd, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return err
+			}
+			if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
+				return err
+			}
+		} else {
+			log.V(1).Info("CRD already exists, updating", "crd", crd.Name)
+			crd.ResourceVersion = existingCrd.ResourceVersion
+			if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Update(crd); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -226,7 +226,284 @@ var _ = Describe("Test", func() {
 
 			close(done)
 		}, 5)
+
+		It("should reinstall the CRDs if already present in the cluster", func(done Done) {
+
+			crds, err = InstallCRDs(env.Config, CRDInstallOptions{
+				Paths: []string{filepath.Join(".", "testdata")},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			// Expect to find the CRDs
+
+			crd := &v1beta1.CustomResourceDefinition{}
+			err = c.Get(context.TODO(), types.NamespacedName{Name: "foos.bar.example.com"}, crd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.Spec.Names.Kind).To(Equal("Foo"))
+
+			crd = &v1beta1.CustomResourceDefinition{}
+			err = c.Get(context.TODO(), types.NamespacedName{Name: "bazs.qux.example.com"}, crd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.Spec.Names.Kind).To(Equal("Baz"))
+
+			crd = &v1beta1.CustomResourceDefinition{}
+			err = c.Get(context.TODO(), types.NamespacedName{Name: "captains.crew.example.com"}, crd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.Spec.Names.Kind).To(Equal("Captain"))
+
+			crd = &v1beta1.CustomResourceDefinition{}
+			err = c.Get(context.TODO(), types.NamespacedName{Name: "firstmates.crew.example.com"}, crd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.Spec.Names.Kind).To(Equal("FirstMate"))
+
+			crd = &v1beta1.CustomResourceDefinition{}
+			err = c.Get(context.TODO(), types.NamespacedName{Name: "drivers.crew.example.com"}, crd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.Spec.Names.Kind).To(Equal("Driver"))
+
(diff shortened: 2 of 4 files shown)


pkg/envtest/crd.go
```diff
@@ -190,8 +192,17 @@ func CreateCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResource
 	// Create each CRD
 	for _, crd := range crds {
 		log.V(1).Info("installing CRD", "crd", crd.Name)
-		if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
-			return err
+		var err error
+		if _, err = cs.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
```
alvaroaleman: Could we have a test for the "IsAlreadyExists" case succeeding? Other than that lgtm, thanks!
/ok-to-test

pires: On it. Thanks for taking the time to review.

pires: I ended up adding a re-install step and validation to the existing _install CRDs_ test. Does that suffice?

DirectXMan12:
add a log line to this warning about it (and switch the title to say "update CRD if already present" to make it more likely that people'll notice it in the release notes), then I think we're good here.

I think it'd be fine even without, but better to provide feedback than give someone a debugging headache.

pires:
Done, PTAL.

DirectXMan12:
did the test change disappear?  I seem to recall that there was a change to the existing test here before...

otherwise lgtm

pires:
Rebased `master` onto this branch and added a couple tests. I think there were no tests before, sorry for that.

pires:
I have a question, though. Why is `WaitForCRDs` explicitly called in tests that call `InstallCRDs`, when the latter func already calls the former?

pires:
As is tradition around here, I forgot to include two new files and that's why I force-pushed. Need caffeine.

DirectXMan12:
> As is tradition around here, I forgot to include two new files and that's why I force-pushed. Need caffeine.

heh, definitely had those days.  Need a lint that checks if we've had caffeine yet for the day :-P

As for the test, did some poking around, and found the failure.  It's another caffeine error :-P  Commented inline.

pkg/envtest/crd.go
```diff
@@ -225,8 +225,18 @@ func CreateCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResource
...
+				_, err = cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd)
+			}
+			if apierrors.IsAlreadyExists(err) {
+				log.V(1).Info("CRD already exists, updating", "crd", crd.Name)
+				_, err = cs.ApiextensionsV1beta1().CustomResourceDefinitions().Update(crd)
+			}
+			if err != nil {
+				return err
+			}
```
DirectXMan12: You also want some logic outside the get check, otherwise if `err == nil` (it already exists), we'll never update.

In fact, this is probably better as a `Create` check here still.

DirectXMan12: we also need to load the resource version to do the update -- apparently we can't do an unconditional update here.

pkg/envtest/envtest_test.go
```diff
@@ -226,8 +226,228 @@ var _ = Describe("Test", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Expect to find updated CRD
+
+		crd = &v1beta1.CustomResourceDefinition{}
+		err = c.Get(context.TODO(), types.NamespacedName{Name: "drivers.crew.example.com"}, crd)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(crd.Spec.Names.Kind).To(Equal("Driver"))
+		Expect(len(crd.Spec.Versions)).To(BeEquivalentTo(3))
+		Expect(crd.ResourceVersion).NotTo(BeEquivalentTo(firstRV))
```
DirectXMan12: you'll also need a `close(done)` at the end here, or just remove the done argument and the timeout

pires: thank you for putting me out of my misery 🙇

pkg/envtest/crd.go
pires: I really needed caffeine when working on this. I am very sorry for that. Clearly, a conflict would never be the result of a GET 🤦‍♂
Now, I didn't know we'd forcibly need to set the resourceVersion when updating so TIL. THANKS!

pires:
Are we good @DirectXMan12?

pkg/envtest/crd.go
```diff
@@ -225,8 +225,18 @@ func CreateCRDs(config *rest.Config, crds []*apiextensionsv1beta1.CustomResource
...
 	for _, crd := range crds {
 		log.V(1).Info("installing CRD", "crd", crd.Name)
-		if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
-			return err
+		if existingCrd, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				if _, err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd); err != nil {
+					return err
+				}
+			}
```
DirectXMan12: should we have a `return err` here?

DirectXMan12: i.e. `if there was an error fetching the existing CRD, and the error wasn't not found, treat it as an error`

pires: You're right. Fixing.

pires: Done.

gerred:
/assign @directxman12

looks like everything has been addressed - I reviewed the code and nothing else stands out. thanks @pires

DirectXMan12:
Added a bit more of a PR description -- in general, I prefer to see a bit of a description beyond the bugs it fixes and the title, even if it's relatively minimal.

otherwise looks good :-)

/lgtm
/approve

DirectXMan12:
Also, since this is a bit of a behavioral change, I've made it a feature instead of a bugfix, just to play it safe

PR #697, opened 2019-11-26
:sparkles: Clean-up installed CRDs on Stop()

### LINKED ISSUE #420: envtest.Environment should cleanup CRDs on Stop()

pires:
But only if `USE_EXISTING_CLUSTER` is set to `true` as there's no point in doing it if the control-plane is going away.

Fixes #420

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -27,6 +27,7 @@ import (
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -204,7 +205,10 @@ func UninstallCRDs(config *rest.Config, options CRDInstallOptions) error {
 	for _, crd := range options.CRDs {
 		log.V(1).Info("uninstalling CRD", "crd", crd.Name)
 		if err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Delete(crd.Name, &metav1.DeleteOptions{}); err != nil {
-			return err
+			// If CRD is not found, we can consider success
+			if !apierrors.IsNotFound(err) {
+				return err
+			}
 		}
 	}
 
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -27,6 +27,7 @@ import (
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -51,6 +52,11 @@ type CRDInstallOptions struct {
 
 	// PollInterval is the interval to check
 	PollInterval time.Duration
+
+	// CleanUpAfterUse will cause the CRDs listed for installation to be
+	// uninstalled when terminating the test environment.
+	// Defaults to false.
+	CleanUpAfterUse bool
 }
 
 const defaultPollInterval = 100 * time.Millisecond
@@ -178,6 +184,31 @@ func (p *poller) poll() (done bool, err error) {
 		}
 	}
 	return allFound, nil
+}
+
+// UninstallCRDs uninstalls a collection of CRDs by reading the crd yaml files from a directory
+func UninstallCRDs(config *rest.Config, options CRDInstallOptions) error {
+
+	// Read the CRD yamls into options.CRDs
+	if err := readCRDFiles(&options); err != nil {
+		return err
+	}
+
+	// Delete the CRDs from the apiserver
+	cs, err := clientset.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	// Uninstall each CRD
+	for _, crd := range options.CRDs {
+		log.V(1).Info("uninstalling CRD", "crd", crd.Name)
+		if err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Delete(crd.Name, &metav1.DeleteOptions{}); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // CreateCRDs creates the CRDs
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -325,7 +325,28 @@ var _ = Describe("Test", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 
+			// Expect to NOT find the CRDs
+
+			crds := []string{
+				"foos.bar.example.com",
+				"bazs.qux.example.com",
+				"captains.crew.example.com",
+				"firstmates.crew.example.com",
+				"drivers.crew.example.com",
+			}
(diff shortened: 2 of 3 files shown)


pires:
Maybe you can take a look at this one, @alvaroaleman? Please!

pkg/envtest/server.go
```diff
@@ -135,10 +135,10 @@ type Environment struct {
 }
 
 // Stop stops a running server.
-// If USE_EXISTING_CLUSTER is set to true, this method is a no-op.
+// If USE_EXISTING_CLUSTER is set to true, uninstall whatever CRDs were installed with Start().
 func (te *Environment) Stop() error {
 	if te.useExistingCluster() {
-		return nil
+		return UninstallCRDs(te.Config, te.CRDInstallOptions)
```
alvaroaleman: I don't think we should be doing this by default because:

* This may be unexpected for users that re-use an existing cluster they have workloads on that use the given CRD they care
* This breaks parallel tests that use the same cluster

Maybe make this an opt-in option of the `Environment`?

pkg/envtest/crd.go
```diff
@@ -180,6 +181,32 @@ func (p *poller) poll() (done bool, err error) {
...
+	// Create each CRD
+	for _, crd := range options.CRDs {
+		log.V(1).Info("uninstalling CRD", "crd", crd.Name)
+		if err := cs.ApiextensionsV1beta1().CustomResourceDefinitions().Delete(crd.Name, &metav1.DeleteOptions{}); err != nil {
+			return err
+		}
+	}
+	return nil
+
+	return nil
```
alvaroaleman: I believe one `return nil` is sufficient here :)

pkg/envtest/envtest_test.go
```diff
@@ -216,4 +216,105 @@ var _ = Describe("Test", func() {
...
+								Served:  true,
+							},
+						}},
+				},
+			},
+				CRDInstallOptions{MaxTime: 50 * time.Millisecond, PollInterval: 15 * time.Millisecond},
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = UninstallCRDs(env.Config, CRDInstallOptions{
```
alvaroaleman: Apart from the functionality itself, maybe also test when its being executed, i.E. probably not by default?

pkg/envtest/crd.go
pires: 🤦‍♂ will fix.

pkg/envtest/envtest_test.go
pires: Indeed! Will do.

pkg/envtest/server.go
pires: This makes total sense! I'm _ashamed_ it didn't cross my mind. Will do.

pkg/envtest/envtest_test.go
pires: Actually, I just realized this test will only pass if `USE_EXISTING_CLUSTER` is `true` - which was the only scenario I actually tested against. So I realize now that uninstalling only when the aforementioned environment variable is used was a stupid pre-optimization that I am removing now.

pires: I also just realized the existing tests do not cover starting/stopping the environment but only installing CRDs by explicitly call `InstallCRDs`, so I'm unsure how to proceed here. Thoughts?

gerred:
After the fixes, this looks good! Thanks @pires 

/lgtm

pkg/envtest/envtest_test.go
gerred: Might be beyond the scope of this PR imo, maybe we have external e2e tests that do this? In KUDO we're working on abstracting out `kudo test` into [kuttl](https://github.com/kudobuilder/kuttl) which can start external environments with kind.

gerred:
looks like the teardown code is removing the CRDs as well which is why that test is failing? so uninstallCRDs runs but the suite harness is trying to do the same thing.

/lgtm cancel

PR itself looks ok pending fixing that test. :)

pires:
Well, I can re-install the CRDs so the `AfterEach` is happy or I can add a validation to said func that tries to delete only if CRD is present. I'll go with the latter for now and change later if you're not pleased with that solution.

gerred:
Latter sounds good :+1: thanks @pires!

pires:
Sorry for the few force-pushes but stupid me was working from a different machine that didn't had my latest changes and that took me a while to figure out 🤦‍♂

pkg/envtest/envtest_test.go
pires: I'd say it's beyond the scope of this PR.

pkg/envtest/envtest_test.go
```diff
@@ -216,4 +227,105 @@ var _ = Describe("Test", func() {
...
+				Paths: []string{filepath.Join(".", "testdata")},
+			})
+			Expect(err).NotTo(HaveOccurred())
```
alvaroaleman: I think we should also test if they are actually gone, not just if we didn't return an error :P

pires: Agreed.

pires: Done.

pires: Done.

pkg/envtest/envtest_test.go
```diff
@@ -216,4 +227,126 @@ var _ = Describe("Test", func() {
...
+				"bazs.qux.example.com",
+				"captains.crew.example.com",
+				"firstmates.crew.example.com",
+				"drivers.crew.example.com",
+			}
+			placeholder := &v1beta1.CustomResourceDefinition{}
+			Eventually(func() bool {
+				for _, crd := range crds {
+					err = c.Get(context.TODO(), types.NamespacedName{Name: crd}, placeholder)
+					notFound := err != nil && apierrors.IsNotFound(err)
```
alvaroaleman: Nit: AFAIK you don't need to nilcheck the errors before passing it into apierrors.IsNotFound

pkg/envtest/server.go
```diff
@@ -135,8 +135,15 @@ type Environment struct {
...
 
 // Stop stops a running server.
-// If USE_EXISTING_CLUSTER is set to true, this method is a no-op.
+// Previously installed CRDs, as listed in CRDInstallOptions.CRDs, will be uninstalled
+// if CRDInstallOptions.CleanUpAfterUse are set to true.
 func (te *Environment) Stop() error {
+	if te.CRDInstallOptions.CleanUpAfterUse {
+		if err := UninstallCRDs(te.Config, te.CRDInstallOptions); err != nil {
+			// This error should be harmless so just log it.
+			log.Error(err, "failed to uninstall CRDs")
```
alvaroaleman: IMHO the decision if this error should be handled or not belongs to the caller, hence we should return it

pkg/envtest/envtest_test.go
pires: Maybe I'm misunderstanding what the code does but there's no check for `nil` in https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L485-L488, so my understanding is if I pass `nil`, I'll get a panic.

pkg/envtest/server.go
pires: Done.

pkg/envtest/envtest_test.go
alvaroaleman: Its a bit buried, underneath it uses https://github.com/kubernetes/apimachinery/blob/adc6f4cd9e7d28fc1c1e5efc658d940d55c0f356/pkg/api/errors/errors.go#L642 which does a type assertion which doesn't match on nil, so a default is returned that is not a `NotFound` so you'll just get back a `false` and no panic. But again, just a nit.

pires: TIL, thanks!

But if it's a nit, I ask you approve this as is as it will make the code easier to read for newbies like myself.

DirectXMan12:
/approve

thanks!

PR #701, opened 2019-11-28
📖: improve MultiNamespacedCacheBuilder info

camilamacedo86:
**Description**

- Make clear that scenario as "watch all less one namespace" are not recommended for it and recommend the ideal solution for this scenario.
- Make clear that it may face performance issues if your api-server is super flooded or you are watching like 10000 namespaces

**Motivation**
Closes: https://github.com/operator-framework/operator-sdk/issues/2008

### DIFF
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -36,7 +36,9 @@ type NewCacheFunc func(config *rest.Config, opts Options) (Cache, error)
 
 // MultiNamespacedCacheBuilder - Builder function to create a new multi-namespaced cache.
 // This will scope the cache to a list of namespaces. Listing for all namespaces
-// will list for all the namespaces that this knows about.
+// will list for all the namespaces that this knows about. Note that this is not intended
+// to be used for excluding namespaces, this is better done via a Predicate. Also note that
+// you may face performance issues when using this with a high number of namespaces.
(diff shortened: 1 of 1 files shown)


pkg/cache/multi_namespace_cache.go
```diff
@@ -35,8 +35,11 @@ import (
 type NewCacheFunc func(config *rest.Config, opts Options) (Cache, error)
 
 // MultiNamespacedCacheBuilder - Builder function to create a new multi-namespaced cache.
-// This will scope the cache to a list of namespaces. Listing for all namespaces
-// will list for all the namespaces that this knows about.
+// This will scope the cache to a list of namespaces. Note that its usage is NOT recommended
```
alvaroaleman: I think we should keep the `Listing for all namespaces will list for all the namespaces that this cache knows about. ` sentence.

Maybe `Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate. Also note that you may face performance issues when using this with a high number of namespaces.`  is as a more precise description?

And regarding the performance issue, is that much of a difference compared to using a global cache?

camilamacedo86: Applied suggestions.

camilamacedo86:
Hi @droot, 

I saw that you added the lgtm but it still missing the approved one to get merged. Could you please add this label?

PR #704, opened 2019-12-02
 🐛 use semantic equality in CreateOrUpdate

### LINKED ISSUE #464: Consider changing controllerutil#CreateOrUpdate to use equality.Semantic.DeepEqual instead of reflect.DeepEqual
I tried to use CreateOrUpdate with a NetworkPolicy object and it kept resulting in OperationResultUpdated even though nothing has changed. It seems like using reflect.DeepEqual sometimes gives false positives.

I have created a copy of CreateOrUpdate that uses equality.Semantic.DeepEqual from the apimachinery package (`Semantic can do semantic deep equality checks for api objects`). It seems to resolve the issue for me. Should this be changed upstream?

alexeldeib:
plucking some low hanging fruit

fixes #464

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -19,8 +19,8 @@ package controllerutil
 import (
 	"context"
 	"fmt"
-	"reflect"
 
+	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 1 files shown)


alexeldeib:
server side apply is probably better here, actually.

/close

DirectXMan12:
We should probably do both, actually.  This is a pretty easy fix for some weird issues.

PR #710, opened 2019-12-04
:sparkles: more visibility of the testing control plane

Random-Liu:
Support bridging the control plane output to the stdout and stderr

This is very useful for debugging in CI. Let's cherrypick it to release-0.1 which is the only branch that doesn't have this useful thing. :)


mengqiy:
@Random-Liu  It seems the linter is not happy.
I guess we can simply diable that linter, we don't expect much more code change into release-0.1 branch.

Random-Liu:
@mengqiy Added a nolint. Hope it will work.

Issue #210, opened 2018-11-13
Remove metrics global registry


LiliC:
I would like to suggest to remove the global registry as currently the controller-runtime global registry does not capture metrics registered onto `prometheus.DefaultRegisterer`. As a step forward and in line with the prosed [KEP](https://github.com/kubernetes/community/pull/2909), the controller-runtime should be able to have the registry injected. (for a start with `prometheus.DefaultRegisterer`, with the long term goal of removing the global registry in the first place, I am open for suggestions on that.)

As per the proposal https://github.com/kubernetes/community/pull/2909:

> Kubernetes also makes extensive use of a global metrics registry to register metrics to be exposed. Aside from general shortcomings of global variables, Kubernetes is seeing actual effects of this, causing a number of components to use sync.Once or other mechanisms to ensure to not panic, when registering metrics. Instead a metrics registry should be passed to each component in order to explicitly register metrics instead of through init methods or other global, non-obvious executions.

Lastly, this may be my own opinion, but introducing globals as a library hides its dependencies, therefore I think having the registry be injected makes the API more clear.

DirectXMan12:
*Disclaimer: the below represents my current thoughts, but I'm generally open to being convinced otherwise :-)*

The original goal was to make registering metrics "easy", while also not necessarily polluting the default Prometheus registry.  IMO, this is a pattern that the default Prometheus docs get right, especially for smaller projects -- since the metrics registry is in a global, there's very little code or cognitive overhead in registering a new metric -- you just declare it at the top of your file, and start using it.  Fundamentally, most programmers are lazy, and so the easier we can make it to add metrics, the more likely people are to do it.  I don't want to give people excuses to put off writing metrics until later :-).  Overhead leads to all sorts of excuses -- "oh, we can't put metrics here without refactoring to plumb through the registry", "oh we can't put metrics here because this is a free-standing function and not a method", etc.  You might still get metrics, but you'll miss out on some that people didn't think were worth the effort.

With that motivation in mind, I think our current setup is a decent compromise -- it doesn't use the default registry -- it's actually separate, so you can choose to use it or not use it.  However, you still get a similar feel: instead of just importing `prometheus` you have to import `prometheus` and `metrics`, but you get a similar feel -- declaring `metrics.Registry.MustRegister` at the top of your file, and then using the metrics as normal.

LiliC:
/remove-lifecycle rotten 

Personally, still think we should move away from it :)

[...]

PR #713, opened 2019-12-05
✨ Define metrics.Registry as an interface

xrstf:
It's currently not possible to replace the predefined Registry with a custom implementation.

I need this because I re-create controllers with the same name at runtime and this causes the default registry to rightfully complain that the metrics for this label combination are already registered. When calling `controller.New()` I have no influence over the metrics creation and errors are always logged with a logger I can also not control.

With this PR I can register my own Registry implementation that silently drops the AlreadyRegistered errors.

We could have used a custom subset of the two predefined Registerer/Gatherer interfaces, but both are so short and only contain what we actually need, so it seemed not worth the effort to replicate the method definitions.

I can see that there are some long-standing intentions to migrate to both [non-global metrics definitions](https://github.com/kubernetes-sigs/controller-runtime/issues/210) and the [use of OpenTelemetry](https://github.com/kubernetes-sigs/controller-runtime/issues/305), but in the meantime this would make life easier [for me]. :-)

### DIFF
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -18,6 +18,13 @@ package metrics
 
 import "github.com/prometheus/client_golang/prometheus"
 
+// RegistererGatherer combines both parts of the API of a Prometheus
+// registry, both the Registerer and the Gatherer interfaces.
+type RegistererGatherer interface {
+	prometheus.Registerer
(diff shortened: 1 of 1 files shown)


DirectXMan12:
/retest

o_O looks like a flake?

Issue #676, opened 2019-11-11
[Discuss] Move sigs.k8s.io/testing_frameworks/integration to controller-runtime


hoegaarden:
Proposal: Move testing_frameworks/integration into controller-runtime (& deprecate the former)

xref: https://github.com/kubernetes-sigs/testing_frameworks/pull/68#issuecomment-536760791

Advantages:
* Some configuration and implementation gets simpler; e.g. the equivalent of "kubebuilder assets"
* Smaller change/PR overhead - changes that involve both testing_frameworks/integration and controller-runtime can be done in one place
* Until now, testing_frameworks/integration has made the decision to not vendor any k/k packages. controller-runtime, on the other hand, is using k8s.io packages. That sometimes creates friction (e.g. https://github.com/kubernetes-sigs/testing_frameworks/pull/68), which could be eliminated if the 2 lived in the same place.

Drawbacks:
* Anyone who is using testing_frameworks/integration would be forced to vendor parts of k8s.io, which might not be desirable. In practice, we have only seen references to testing_frameworks/integration as part of controller-runtime, so that should be a pretty slim possibility.
* The resulting codebase would not be able to be vendored into k/k (as is the case with controller-runtime today anyway).

Todos:
- [x] Move code to controller-runtime
- [x] Check on deprecation of repos with sig-contribex (or whoever is responsible for that)
- [x] Deprecate sigs.k8s.io/testing_frameworks

/cc @DirectXMan12 @mengqiy @apelisse 
/sig testing
/kind cleanup

mengqiy:
We probably should be starting by moving it under `pkg/internal/testingframework` and then switch controller-runtime code to depend on that.

DirectXMan12:
yep, sgtm.  It'll be a breaking change I believe, since we expose some bits of testing_frameworks in our public API.  Likely low impact though.

hoegaarden:
Code has been moved over, the original repo has been archived, so we can
/close
this issue. Thanks all!

danieldn:
@hoegaarden With `sigs.k8s.io/testing_frameworks/integration` deprecated, what is the best way for us to import this integration library if we want to use it in our projects? I just found out the following doesn't work
```
imports sigs.k8s.io/controller-runtime/pkg/internal/testing/integration: use of internal package sigs.k8s.io/controller-runtime/pkg/internal/testing/integration not allowed
```

PR #714, opened 2019-12-06
⚠ Bring in testing framework

hoegaarden:
This brings in https://github.com/kubernetes-sigs/testing_frameworks/tree/master/integration (and all it's history) as discussed in https://github.com/kubernetes-sigs/controller-runtime/issues/676.

The important commits are:
- ffbd93def06ee935f34f45f0c34cafcaf8d9d8f3 -- subtree add
- 56c5ea7adda1b66a4a7b5f27e917799e8dd2f9df -- change import path
- 8a6a7f28ad92df0d9f70441d6ebb4de8e8ea0b9b -- cleanup & integrate
- d492d7ac0c379635cda846561d83385a5e37e40a -- fix golangci issues

/assign @DirectXMan12 @mengqiy


gerred:
Looks good code-wise! Thanks for doing this I'm assuming once this is squashed the invalid commit messages will go away. :)

Separately, looks like we need to spruce up CI and get these and unit tests running as part of these checks. Also, looks like someone in this list hasn't signed the CLA?

/assign gerred

hoegaarden:
- I would have preferred to not squash, to keep the history ... 🤷‍♂ 
- Every contributor should have signed the CLA, we had the CLA bot running over there too
- The tests for `./pkg/internal/testing/integration/...` are already running by default and on prow, see [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/714/pull-controller-runtime-test/1202959904759877632/build-log.txt)

gerred:
yeah IDK, may want to have the contributors email the CNCF since the CLA bot doesn't seem happy at the moment?

Thanks for the prow link, did not see that. :+1: Pending an answer on squash from an approver and the CLA fix (idk why prow is unhappy), I'm /lgtm

hoegaarden:
> Pending an answer on squash 

Yeah ... let me know if you prefer to squash or not. I can definitely squash if you prefer that.

> idk why prow is unhappy

Maybe a misunderstanding, but where is prow not happy? I just see the status `pull-controller-runtime-test — Job succeeded.`. Isn't it only the CLA bot that is unhappy?

gerred:
Both the CLA bot and Prow.

See https://github.com/kubernetes-sigs/controller-runtime/pull/714#issuecomment-562594808 - not sure why it thinks those commit messages are invalid?

And then the CLA bot separately. :) I need :coffee:.

hoegaarden:
> See #714 (comment) - not sure why it thinks those commit messages are invalid and added the hold label.

Because they say "Closes" and there was a plugin enabled globally (all k8s orgs) more or less recently which does not allow that. Those commits seem to have made it into the original repo, before that plugin was enabled.

> And then the CLA bot separately

Yeah, no idea why that is :man_shrugging:

DirectXMan12:
CLA problem is on 8c458a464b68bdae239348e2ee04c966af836454 -- it's missing a GitHub user, according to the CLA bot.

DirectXMan12:
(GitHub isn't sure what to do with one "Josh Hill" either, so it looks like that's genuinely a problem)

DirectXMan12:
This PR: Add script to run tests on prow

DirectXMan12:
checking with contribex, looks like we'll have to file at ticket with the linuxfoundation.  Will keep y'all posterd

hoegaarden:
Ok ... So Josh changed companies somewhen after that commit and therefore also changed his email address. I guess the CLA bot is now not able to connect that email address to a github or a linuxfoundation.org user. However it seems he had signed the CLA at some point, the original PR has the `cncf-cla: yes` label. I would guess that the fact that he changed jobs / emails does not void the accordance of this commit with the CLA. Or does this change when a commit is moved to a different repo?

DirectXMan12:
It shouldn't void the CLA I think, but CNCF got back to me and (nicely) said "sorry, sucks to be you", so I think we'll need to replace the commit or :ghost: force-merge :ghost: .  Let's get the rest of the issues out of the way (like the commit messages with "Fixes" / "Closes" in them).

hoegaarden:
I can do that (rewording the offending commit messages).

However, my approach with using `git subtree` and other similar methods (mainly with the goal to preserve history) causes other issues, e.g.: After all things have been merged in and integrated, if I do a simple `git rebase upstream/master` I run into all sorts of merge conflicts. Resolvable, but annoying & to some extend surprising.

I wonder if this will cause more pain in future than necessary and if we should just squash the commits and bring the thing in in one big ~swoosh~ commit. That would def. resolve the offending commit messages, might help with the CLA bot (because I'd put all the people of all commits on a `Co-authored-by:` trailer instead of having them as authors/committers). Obviously, we'd loose the history. Not sure how much we care?

PR #715, opened 2019-12-07
:sparkles: ErrorIfCRDPathMissing on Environment

### LINKED ISSUE #481: envtest.Environment.Start() silently fails for missing CRD paths

knabben:
This PR overwrites the ErrorIfPathMissing from CRDInstallOptions via Environment.ErrorIfCRDPathMissing on Start() method.

Similar to PR #511 but propagates the flag and test the Start() with invalid directories.

Closes #481.

### DIFF
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -36,6 +36,9 @@ var _ = Describe("Test", func() {
 	var s *runtime.Scheme
 	var c client.Client
 
+	var validDirectory = filepath.Join(".", "testdata")
+	var invalidDirectory = "fake"
+
 	// Initialize the client
 	BeforeEach(func(done Done) {
 		crds = []*v1beta1.CustomResourceDefinition{}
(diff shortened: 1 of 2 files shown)


pkg/envtest/envtest_test.go
```diff
@@ -349,4 +354,20 @@ var _ = Describe("Test", func() {
 			close(done)
 		}, 30)
 	})
+
+	Describe("Start", func() {
+		It("should raise an error", func(done Done) {
```
alvaroaleman: Nit: Elaborate in the name why it should (not) raise an error, i.E. "should raise an error on invalid dir`

PR #716, opened 2019-12-07
:sparkles: CRD file path on CRDInstallOptions.Paths

### LINKED ISSUE #394: Test environment should support passing crd file path in additional to directories
When building a controller which watches another CRD we need to add CRD file to the test environment, those CRD files usually resides in your project's vendor directory. However test environment struct only supports directories not file path, if we point to the vendor directory which contains files other than crd files, the tests would fail.  	
```
t := &envtest.Environment{
    CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "crds")},
}
```

knabben:
This PR enables the usage of files path on CRDInstallOptions.Paths, allowing users to register only a single CRD or an entire directory.

Next step is to bring this to Environment (maybe CRDFilePaths?).

Starts to fix #394

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -39,7 +39,7 @@ import (
 
 // CRDInstallOptions are the options for installing CRDs
 type CRDInstallOptions struct {
-	// Paths is a list of paths to the directories containing CRDs
+	// Paths is a list of paths to the directories or files containing CRDs
 	Paths []string
 
 	// CRDs is a list of CRDs to install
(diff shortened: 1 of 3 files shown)


DirectXMan12:
/lgtm
/approve

awesome

DirectXMan12:
Flake here: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/713/pull-controller-runtime-test/1204197535384604672

@knabben Can you take a look please?

knabben:
sure. on it

PR #720, opened 2019-12-08
:warning: Envtest: Do not pull in ginkgo outside of _test.go files

alvaroaleman:
I noticed that introducing `envtest` as a dependency also adds `ginkgo` as a transient dependency, this PR fixes that.

/assign @DirectXMan12


DirectXMan12:
This is a breaking change.  We actually meant to make this a while ago, IIRC, but never got around to finalizing the breaking change.

I'm fine with doing this, but let's leave it for the next set of breaking changes.

DirectXMan12:
IIRC, this has been deprecated for a while, but is a :warning: none-the-less.  Updating

DirectXMan12:
ooo, this is gonna be a bit painful b/c it hits test suites.

we need to update KB to start scaffolding new projects with this change too.

gerred:
kicking off one retest bc this is a type alias and shouldn't trigger random timeouts.

this should be generally OK though in KB and elsewhere though until we remove the type alias, yeah?

/retest

PR #725, opened 2019-12-10
:running: Adding another folder with the CRD avoiding late deleted conflicts

knabben:
Another folder inside testdata with the CRD file to avoid working with the ones still being deleted.

This fixes the flake test.


DirectXMan12:
switched to :running: since it's a test failure only

PR #731, opened 2019-12-17
✨Make workqueue rate limiter configurable during controller creation

### LINKED ISSUE #631: Allow controller authors to configure queue rate limits
Controllers created using controller-runtime [implicitly use](https://github.com/kubernetes-sigs/controller-runtime/blob/43a3cd//pkg/controller/controller.go#L89) the [`DefaultControllerRateLimiter`](https://github.com/kubernetes/client-go/blob/911ef75/util/workqueue/default_rate_limiters.go#L39) for the work queue they use to process reconcile requests. If I follow the logic correctly, this means that under the following conditions...

* A watched object has changed.
* A reconcile attempt has returned an error.
* A reconcile attempt has returned `Result{Requeue: true}`

...the earliest time the resulting reconcile will be processed will be the maximum of:

* The enqueue time plus a duration increasing exponentially from 5ms to ~16 minutes per object until said object is reconciled successfully (i.e. a reconcile attempt for that object returns `Result{Requeue: false}, nil` or `Result{RequeueAfter: N}, nil`).
* The enqueue time plus a duration calculated to limit the controller to 10 requeues per second on average.

While building [Crossplane](http://github.com/crossplaneio/crossplane) we've found these requeue semantics are not ideal for reconciling Kubernetes resources with cloud provider APIs. There are two parts to this:

* Typically we don't want to retry as soon as 5ms from the time we encountered an error, and also don't want to reach a point where we're waiting over 16 minutes between each reconcile.
* There is no way for users who do not have access to our controller manager pod logs (i.e. most users) to learn that the object they're waiting on is suffering exponential backoff. Ideally upon encountering an error we'd emit an event explaining the error and informing the user that we'll try again after a particular time.

I believe we could solve both of the above issues if we were able to override the default rate limiter, perhaps by adding it to `controller.Options`. This would allow us to tweak the exponential backoff parameters, and to plumb the rate limiter through to our Reconciler so it could inform concerned parties when an object was next going to be reconciled.

Relates to https://github.com/kubernetes-sigs/controller-runtime/issues/195 and https://github.com/kubernetes-sigs/controller-runtime/issues/417.

rajathagasthya:
Exposes rate limiter as a controller option in case users want to use
a different rate limiter than the default one provided by
controller-runtime.

Fixes #631

### DIFF
--- a/pkg/builder/controller_test.go
+++ b/pkg/builder/controller_test.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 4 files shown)


DirectXMan12:
This seems reasonable.  I'm iffy on the naming, but it currently has the nice property that the signatures match the client-go one, making the cast straightforward.

/ok-to-test

rajathagasthya:
Is just copying and maintaining those client-go rate limiters a bad idea? That way we are insulated from changes to those interfaces and their implementations. But maybe we would like those changes as well?

DirectXMan12:
The idea is that we didn't want a change in client-go to force us to do a major revision.

rajathagasthya:
@DirectXMan12 @shawn-hurley Any thoughts on how to proceed with this change? I'd love to get this merged soon and use it since we sorely need this option.

DirectXMan12:
I think we're good to go forward with this the way it is now

jimmidyson:
Was just looking for this... would love to see this merged :pray: @DirectXMan12. Thank you for adding this @rajathagasthya!

gerred:
this is great, thank you @rajathagasthya 

/approve

mars1024:
@rajathagasthya This is a wonderful PR, could you help to backport this to older versions like 0.4.0 and 0.5.0?

rajathagasthya:
@mars1024 I'm happy to backport. Thoughts @DirectXMan12 @vincepri?

mars1024:
friendly ping @DirectXMan12 @vincepri , a configurable **MakeQueue** is strongly needed anywhere ~

Issue #610, opened 2019-09-18
move off of github actions HCL syntax


BenTheElder:
@k8s-ci-robot received the following warning for this repo (excerpt)
> ...
> With these updates, GitHub Actions now uses YAML syntax in workflow files. The HCL syntax in GitHub Actions will stop working on September 30, 2019. We are contacting you because you’ve run workflows using HCL syntax in the last week in your account @kubernetes-sigs with the following repos: kubernetes-sigs/cluster-api-provider-docker, kubernetes-sigs/controller-runtime. 
>
> To continue using workflows that you created with the HCL syntax, you'll need to [migrate the workflow files to the new YAML syntax](https://help.github.com/en/articles/migrating-github-actions-from-hcl-syntax-to-yaml-syntax
). Once you have your YAML workflows ready, visit your repositories and follow the prompts to upgrade. Once you upgrade, your HCL workflows will stop working.
> ...

DirectXMan12:
Cool, I'll try again with the new syntax -- nothing was happening last time I tried.

/priority important-soon

PR #735, opened 2019-12-18
:running: use new github action syntax

alexeldeib:
fixes https://github.com/kubernetes-sigs/controller-runtime/issues/610

get those sweet, sweet emoji verified

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,12 @@
+on: pull_request
+
+jobs:
+    pr_checks:
+        runs-on: ubuntu-latest
+        name: verify emoji
+        steps:
+        - uses: actions/checkout@v2
+        - name: verify
+          uses: ./hack/release
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,4 +1,6 @@
-on: pull_request
+on:
+  pull_request:
+    types: [opened, closed, reopened, edited, synchronize]
 
 jobs:
     pr_checks:
--- a/hack/release/verify-emoji.sh
+++ b/hack/release/verify-emoji.sh
@@ -25,4 +25,10 @@ base_link=$(jq -r '.pull_request.url' < ${GITHUB_EVENT_PATH})
 head_commit=$(curl -H "Authorization: Bearer ${GITHUB_TOKEN}" -H 'Accept: application/vnd.github.antiope-preview+json' -q ${base_link} | jq -r '.head.sha')
 echo "head commit is ${head_commit}"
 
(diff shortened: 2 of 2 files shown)


alexeldeib:
I think github is unhappy because I am trying to merge into the main workflow: https://github.community/t5/GitHub-Actions/GitHub-action-Resource-not-accessible-by-integration/td-p/19707

However I have an up to date fork where this action works fine: https://github.com/alexeldeib/controller-runtime/commit/f4d028132839d6df1e3d815d0f016ca4a4c17724/checks?check_suite_id=365691723

alexeldeib:
I'm not sure that's the issue, actually. But since actions show up as a check anyway, maybe it's sufficient to leave it like this without the CURL to set status? I guess the text won't pretty print for the unicode emoji, but the check will work at least **shrugs**

DirectXMan12:
Nice, let's get the check back in place, then I'll play around with the pretty-printing.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,12 @@
+on: pull_request
```
DirectXMan12: This changed since I originally wrote this, but I think we want to explicitly list types so we get `edit` as a type here as well: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-event-pull_request

hack/release/verify-emoji.sh
```diff
@@ -25,5 +25,10 @@ base_link=$(jq -r '.pull_request.url' < ${GITHUB_EVENT_PATH})
 head_commit=$(curl -H "Authorization: Bearer ${GITHUB_TOKEN}" -H 'Accept: application/vnd.github.antiope-preview+json' -q ${base_link} | jq -r '.head.sha')
 echo "head commit is ${head_commit}"
 
-curl https://api.github.com/repos/${GITHUB_REPOSITORY}/check-runs -XPOST -H "Authorization: Bearer ${GITHUB_TOKEN}" -H 'Accept: application/vnd.github.antiope-preview+json' -H 'Content-Type: application/json' -q --data-raw '{"name": "Verify Emoji", "head_sha": "'${head_commit}'", "conclusion": "'${conclusion}'", "status": "completed", "completed_at": "'$(date -Iseconds)'", "output": {"title": "Verify Emoji", "summary": "'"${summary}"'"}}'
+echo ${summary}
```
DirectXMan12: Maybe it's just me, but I'm not actually seeing the output of this in the result.  Let me try manually re-running.

DirectXMan12: ah, it just shows up way at the bottom.  Need to scroll.

alexeldeib:
I updated the triggers, but I feel like something weird is happening with the display of the check still..seems like it's showing twice?

alexeldeib:
This is stale and doesn't produce the desired behavior anyway, going to close. I think we probably just want to run this script in prow and exit 1 if failed, or else fail the action directly rather than post back to the checks API.

DirectXMan12:
I played around a bit and got it almost working, but got side-tracked :-/

The problem with prow is that it won't recheck on edits to the title, IIRC -- you'd have to manually run "retest".  We'd need a plugin or something.

Issue #734, opened 2019-12-18
Client initialization doesn't use DynamicRESTMapper by default


yanniszark:
Related to issue: https://github.com/kubernetes-sigs/controller-runtime/issues/321

When using a manager, it uses the DynamicRESTMapper and everything works as expected.
However, when using just the client package, the default RESTMapper is the one from client-go, which caches the discovery endpoint result once and never retries.

I would expect it to use the DynamicRESTMapper by default, unless I am missing something.

alvaroaleman:
> I would expect it to use the DynamicRESTMapper by default, unless I am missing something.

Yes, definitely. Want to prepare a pull for that?

yanniszark:
@alvaroaleman absolutely!

PR #737, opened 2019-12-19
:sparkles: client: Use dynamic RESTMapper by default

yanniszark:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/734

This commit changes the client package to use the dynamic RESTMapper by
default.
The default RESTMapper caches the discovery endpoint results once and
never updates them.
Because of that, if an api-resource is added after the client is
started, the client will mistakenly think the api-resource doesn't exist.

Signed-off-by: Yannis Zarkadas <yanniszark@arrikto.com>

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -64,7 +64,7 @@ func New(config *rest.Config, options Options) (Client, error) {
 	// Init a Mapper if none provided
 	if options.Mapper == nil {
 		var err error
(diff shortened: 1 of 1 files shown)


yanniszark:
@alenkacz thanks for the review :)
It seems that the PR needs an ok-to-test label in order to move forward.

DirectXMan12:
updated to :sparkles: because this is more of a feature introduction (even if we intended to turn it on originally).

Issue #724, opened 2019-12-10
Test that we don't leak goroutines on start and stop


DirectXMan12:
We should have tests for simple operations that make sure they don't leak goroutines.  Namely, `manager.Start & stop` and `controller.New`, both of which leaked in the past.

We can use the runtime package to get the goroutine count.

/good-first-issue

knabben:
Any ideas on how to filter the goroutines from manager/controller, using the NumGoroutines from runtime returns the ones from tests ~43, do makes sense to iterate on these list and extract by file path maybe?

alvaroaleman:
> Any ideas on how to filter only the goroutines from manager/controller?

Maybe by counting them before starting & stopping the manager or a controller, then checking if the count is back at where it was initially?

hypnoglow:
Is there anything else to be done here? The PR is merged so I believe the issue can be closed.

vincepri:
Closing given that #738 has been merged

/close

PR #738, opened 2019-12-20
:running: Verification of leak goroutine on controller and manager

knabben:
This PR adds tests both for manager and controller checking the leak of goroutines on start and stop, I put the threshold of 5 in comparison because it has 3 leaks after stop:

```
goroutine 29 [chan receive]:
k8s.io/apimachinery/pkg/watch.(*Broadcaster).loop(0xc00021fa80)

goroutine 166 [chan receive]:
k8s.io/client-go/tools/record.(*eventBroadcasterImpl).StartEventWatcher.func1(0x291e580, 0xc000212ab0, 0xc000212a80)

goroutine 167 [chan receive]:
k8s.io/client-go/tools/record.(*eventBroadcasterImpl).StartEventWatcher.func1(0x291e580, 0xc000212ae0, 0xc0003fb350)
```

Related with #724

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -18,9 +18,11 @@ package controller_test
 
 import (
 	"fmt"
+	rt "runtime"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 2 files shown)


alenkacz:
Hmm are the leaked goroutines expected? Or should the test rather not have a treshold of 5 but rather 0 and we should fix those leaks before landing this?

DirectXMan12:
Those are semi-expected.  We don't have a good way to work around them yet.  I *think* we should finally have a `shutdown` method, but that doesn't alleviate the problem entirely.  Ideas welcome.

For now, the threshold should be whatever exists currently (so the three), with a comment to ratchet down as we fix the issues.

pkg/controller/controller_test.go
```diff
@@ -90,6 +92,26 @@ var _ = Describe("controller.Controller", func() {
 
 			close(done)
 		})
+
+		It("should not leak goroutines when stop", func(done Done) {
+			// TODO(community) - After closing the proper leaks on watch this must be reduced to 0
```
alenkacz: should we maybe also document why are those leaks expected and what goroutines are leaking for future reference?

DirectXMan12: yeah.  We can at least link to the relevant issue somewhere around here.

DirectXMan12:
gonna push the note, otherwise we should be good

pkg/controller/controller_test.go
DirectXMan12: (also, for future reference, `community` is not actually a valid TODO target -- this isn't supposed to be who does the TODO, it's supposed to be who has context)

PR #739, opened 2019-12-20
📖Replace NullLogger with a zap logger in the ExampleBuilder

erwinvaneyk:
The `ExampleBuilder` uses the default logger, which is the `NullLogger`.  Since this example is (I assume) meant as a comprehensive example, it would be more helpful to print the actual logs.  This PR replaces the default null logger with a zap logger in the example.

### DIFF
--- a/pkg/builder/example_test.go
+++ b/pkg/builder/example_test.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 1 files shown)


pkg/builder/example_test.go
```diff
@@ -40,6 +41,8 @@ import (
 //
 // * Start the application.
 func ExampleBuilder() {
+	logf.SetLogger(zap.New())
+
```
alvaroaleman: Nit: There is a stale tab here in the otherwise empty line (interesting that github doesn't show that), can you remove that? Other than that the PR looks good.

erwinvaneyk:
@alvaroaleman thanks for the quick review! Fixed your feedback - should be good to go now. 🙂

DirectXMan12:
/retest
#725 I believe

PR #742, opened 2019-12-26
📖 Rename "object" to "owned" to match godoc comment.

evankanderson:
I was reading the godoc and noticed that the actual argument didn't match the godoc:

> SetControllerReference sets owner as a Controller OwnerReference on **owned**. This is used for garbage collection of the **owned** object and for reconciling the owner object on changes to **owned** (with a Watch + EnqueueRequestForOwner). Since only one OwnerReference can be a controller, it returns an error if there is another OwnerReference with Controller flag set.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -54,7 +54,7 @@ func newAlreadyOwnedError(Object metav1.Object, Owner metav1.OwnerReference) *Al
 // reconciling the owner object on changes to owned (with a Watch + EnqueueRequestForOwner).
 // Since only one OwnerReference can be a controller, it returns an error if
 // there is another OwnerReference with Controller flag set.
-func SetControllerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
+func SetControllerReference(owner, owned metav1.Object, scheme *runtime.Scheme) error {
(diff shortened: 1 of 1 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -95,7 +95,7 @@ func SetControllerReference(owner, object metav1.Object, scheme *runtime.Scheme)
 	}
 
 	// Update owner references
-	object.SetOwnerReferences(existingRefs)
+	owner.SetOwnerReferences(existingRefs)
```
DirectXMan12: typo here.  I think this is a good reason that we should maybe stick with `owner` and `object`, or come up with something different -- maybe parent and child?  Otherwise it's too easy to typo

gerred: :+1: would also like to see a test catching this.

evankanderson: It turns out that the first test ("should set the OwnerReference if it can find the group version kind") catches this; I think this was a stray unlucky keystroke in overwrite mode which happened to be valid.

I've renamed `owned` to `controlled`, which should avoid that problem.

DirectXMan12:
/lgtm
/approve

removed sparkles since this isn't a user-facing change

PR #743, opened 2019-12-27
✨ Add httpserver.Server and `pkg/debug` for pprof support

clamoriniere:
The idea is to have an `Http.Server` as a `manager.Runnable` component
On which other components can register `http.Handler`.

The main benefice is to have only one http.Listener with multiples
`http.Handler`:

* Add new httpserver.Server Runnable component
* Add metrics handler registration function
* Add debug package with pprof Handler registration function

I saw a discussion on another PullRequest that seems to agree with this approach: https://github.com/kubernetes-sigs/controller-runtime/pull/664#issuecomment-562601357

### DIFF
--- a/pkg/debug/register.go
+++ b/pkg/debug/register.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2020 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/debug/register.go
+++ b/pkg/debug/register.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package debug
+
+import (
+	"net/http/pprof"
+
+	"sigs.k8s.io/controller-runtime/pkg/httpserver"
+)
+
+// Options use to provide configuration option
+type Options struct {
+	CmdLine bool
+	Profile bool
+	Symbol  bool
+	Trace   bool
+}
+
+// DefaultOptions returns default options configuration
+func DefaultOptions() *Options {
+	return &Options{
+		CmdLine: true,
+		Profile: true,
+		Symbol:  true,
+		Trace:   true,
+	}
+}
+
+// Register use to register the different debug endpoint
+func Register(mux httpserver.Server, options *Options) {
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	if options == nil {
+		options = DefaultOptions()
+	}
+	if options.CmdLine {
+		mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	}
+	if options.Profile {
+		mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	}
+	if options.Symbol {
+		mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	}
+	if options.Trace {
+		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	}
+}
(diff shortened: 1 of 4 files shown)


DirectXMan12:
We've thus far resisted a generic "http server" runnable, mainly because we didn't want to be responsible for generic HTTP serving in CR.

clamoriniere:
Hi @DirectXMan12,
thanks for looking into this PR.

> We've thus far resisted a generic "http server" runnable, mainly because we didn't want to be responsible for generic HTTP serving in CR.

I can update the PR to have the `httpserver` more conservative and allow to register/activate only internal CR handlers.
And allow the usage of "internal" CR handlers (like `/metrics`) with an external `http.Server`

What do you think?

DirectXMan12:
It should totally be possible to use "internal" handlers (like `/metrics`) with an external HTTP server.  That's something I'd like to see.

Unifying all internal handlers into a single HTTP server seems like it could be reasonable.  We'd have to figure out if we want to allow people to put different things on different ports though (probably).

gerred:
Is this something that should be handled at by CR rather than the end controller level level (re: different ports @DirectXMan12?) it seems like CR may be able to offer an http.Server interface that can be used by end controllers, but I'm not so certain about having CR actually be the thing that calls ListenAndServe in this case.

DirectXMan12:
So, I think I'd like to see this, off the top of my head:

1. if you do nothing, you get sane defaults -- metrics, health, etc started & managed by CR for you
2. If you want, you can customize what port (ports?) they serve on.  For compat reasons, we might have to make this separate ports
3. If you want to do anything more complicated, you have to manage your own HTTP server and call ListenAndServe yourself, but we expose the helpers to add things like `/metrics` and `/healthz` to those servers.

DirectXMan12:
(so I think I'm agreeing with you, at least partially, @gerred)

dbenque:
> So, I think I'd like to see this, off the top of my head:
> 
> 1. if you do nothing, you get sane defaults -- metrics, health, etc started & managed by CR for you
> 2. If you want, you can customize what port (ports?) they serve on.  For compat reasons, we might have to make this separate ports
> 3. If you want to do anything more complicated, you have to manage your own HTTP server and call ListenAndServe yourself, but we expose the helpers to add things like `/metrics` and `/healthz` to those servers.

I would really like to see 3. Somehow I already achieve it by `ManagerOptions.MetricsBindAddress = "0" ` and running my own server in a Runnable. But as CR continues to add features I would really love to be able to register handler of CR like `/healthz` to my own server.

clamoriniere:
Thanks all for your comments.
I will work on it soon.

alenkacz:
@clamoriniere are you still working on this or may I take over?

clamoriniere:
> @clamoriniere are you still working on this or may I take over?

Hi @alenkacz ,
Unfortunately, I'm not able to work on it in the coming weeks. So if you have time to take over feel free to do it :)

pkg/debug/register.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
camilamacedo86, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

alenkacz:
@DirectXMan12 I am working on this functionality and I was following the discussion here. We currently expose `metricsExtraHandlers` which is basically "server anything you want on the metrics server". Do we want to remove it? Because if we keep it, it's basically what this PR was doing...

DirectXMan12:
@alenkacz yeah, I think we should deprecate that in favor of the piece of functionality described by bullet 3 above.  I had completely forgotten that existed, thanks for the reminder :-P

pkg/httpserver/interface.go
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
clamoriniere, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/httpserver/runner.go
```diff
@@ -0,0 +1,103 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
clamoriniere, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/metrics/register.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
clamoriniere, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

DirectXMan12:
closing as per https://github.com/kubernetes-sigs/controller-runtime/pull/743#issuecomment-638452425

hectorj2f:
I know this issue got closed but I am wondering why pprof is not considered as a by `default` server to serve as part of controller-runtime. It'd be beneficial for everyone here, 1st for controller-runtime to keep track of the cpu.mem profiles and avoid potential regressions in that regard, and controller-runtime users to have that server available (or enabled=false) when needed.

PR #746, opened 2019-12-31
:running: Minor nit in predicates

ialidzhikov:
Simplify the `if` statement with single `return`.


DirectXMan12:
(changed to `:running:` because this is not a functional user-facing change)

PR #751, opened 2020-01-08
📖 docs: document the expected names of the TLS key and certificate

bhcleek:
Document the expected names of TLS key and certificate so that users don't have dig through code to discover it.


gerred:
This should be implicit for any secret, but more docs couldn't hurt, so :+1:

DirectXMan12:
OMG YES.  UUUGH, I've had to look this up myself a few times, and never had time to fix it :-).

/approve
(per @gerred's lgtm)

PR #753, opened 2020-01-08
:bug: When loading client config, check os/user.HomeDir if $HOME is unset

### LINKED ISSUE #748: config.loadConfig Breaking Change introduced in 0.4.0
As part of https://github.com/kubernetes-sigs/controller-runtime/commit/840ac8a2ef0d9bab7e4f721970bda72cae3b2979 there was a switch for the fallback behaviour of `config.loadConfig` (when not specifying via flag, env var or internal config) from identifying the default home directory with `os/user` to the `k8s.io/client-go/util/homedir` package (called via `clientcmd.NewDefaultClientConfigLoadingRules()`).

This has the effect of switching from grabbing the os home config to relying on the `HOME` env var to be set which is not always the case.

This is not listed as a breaking change within the v0.4.0 release notes however it will effect anyone using the `GetConfig` or `GetConfigOrDie` functions.

Let me know if there are any other details needed!

Cheers

cc @gmrodgers

joelanford:
Following up on the regression introduced in #642 and discussed in #748, this PR detects if $HOME is empty and, if so, automatically sets $HOME to user.Current before using client-go's default loading rules, which only checks $HOME.

As a follow-on, we could submit an issue upstream to see if it makes sense for client-go to use user.Current as a fallback so that we don't have to do it here in controller-runtime.

Closes #748

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/user"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -117,6 +118,22 @@ func loadConfig(context string) (*rest.Config, error) {
 
 	// If the recommended kubeconfig env variable is set, or there
 	// is no in-cluster config, try the default recommended locations.
+	//
+	// NOTE: For default config file locations, upstream only checks
+	// $HOME for the user's home directory, but we can also use
+	// os/user.HomeDir when $HOME is unset.
+	//
+	// TODO(jlanford): could this be done upstream?
+	if _, ok := os.LookupEnv("HOME"); !ok {
+		u, err := user.Current()
+		if err != nil {
+			return nil, fmt.Errorf("could not get current user: %v", err)
+		}
+		if err := os.Setenv("HOME", u.HomeDir); err != nil {
+			return nil, fmt.Errorf("could not set HOME env var: %v", err)
+		}
+		defer func() { _ = os.Unsetenv("HOME") }()
+	}
 	if c, err := loadConfigWithContext(apiServerURL, clientcmd.NewDefaultClientConfigLoadingRules(), context); err == nil {
 		return c, nil
 	}
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"os/user"
+	"path"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


gerred:
/lgtm

This should be upstream but makes sense here for now.

pkg/client/config/config.go
```diff
@@ -117,6 +118,22 @@ func loadConfig(context string) (*rest.Config, error) {
 
 	// If the recommended kubeconfig env variable is set, or there
 	// is no in-cluster config, try the default recommended locations.
+	//
+	// NOTE: For default config file locations, upstream only checks
+	// $HOME for the user's home directory, but we can also use
+	// os/user.HomeDir when $HOME is unset.
+	//
+	// TODO(jlanford): could this be done upstream?
+	if _, ok := os.LookupEnv("HOME"); !ok {
```
DirectXMan12: does this actually work?  It looks like homedir is called from a static variable initializer -- https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L51

joelanford: Oh good catch. Yeah this won't work. It looks like changing `RecommendedHomeFile` will work though since that's what actually ends up getting used: https://github.com/kubernetes/client-go/blob/7ec8a74ae980246e9d1a5734f45bd44fde93a5c1/tools/clientcmd/loader.go#L153

I'll update to just set that directly instead of the HOME env var.

DirectXMan12: is that thread-safe?

I'm worried that we may have to do a more invasive solution here, like customizing the result of (or not using at all) the NewDefaultBlahBlahBlah method.

DirectXMan12:
/hold

thread safety

kensipe:
@DirectXMan12 could you elaborate on the thread safety concern?   I'm not seeing it.     There is the concern that `$HOME` does not exist by default on windows.   

since this is a bug fix... it would also be good to have a test

joelanford:
@kensipe I think the thead-safety concern is that this would change a global variable `clientcmd.RecommendedHomeFile` without a lock.

If multiple callers call any of the methods that use `loadConfig` concurrently without HOME set, it is possible that we would not correctly reset `clientcmd.RecommendedHomeFile` back to its original value:
1. caller A sets"
    - `oldRecommendedHomeFile = <original>`
    - `clientcmd.RecommendedHomeFile = <foobarA>`
2. caller B sets 
    - `oldRecommendedHomeFile = <foobarA>`
    - `clientcmd.RecommendedHomeFile = <foobarB>`
3. caller A resets `clientcmd.RecommendedHomeFile = <original>`
4. caller B resets `clientcmd.RecommendedHomeFile = <foobarB>`

Final value of `clientcmd.RecommendedHomeFile` is `<foobarB>` when it should be `<original>`.

@DirectXMan12 I'll look into alternatives.

pkg/client/config/config.go
joelanford: I _believe_ this is now thread-safe. Now, we no longer change any global variables. Instead we create an instance of `loadingRules` and then append the extra file derived from `u.HomeDir` to the precedence list.

I manually tested this and it still works as expected.

Suggestions on adding a unit test for this? I'm not opposed, but the use of `user.Current()` and `u.HomeDir` makes it a little awkward. I don't love the idea of creating temporary files and directories in the home directories of anyone that runs the unit tests.

PR #755, opened 2020-01-10
:bug: wait for cache start before returning from WaitForCacheSync

joelanford:
There is currently an unintuitive race condition between starting
a cache, waiting for it to sync, and using it the first time.

The crux of the problem is that WaitForCacheSync can return true
BEFORE the Start function has had a chance to set started to true.

This means that Get and List can return ErrCacheNotStarted even
after WaitForCacheSync has returned true.

This commit adds a channel and a wait function that is called from
WaitForCacheSync to ensure that started has been set to true before
returning.


DirectXMan12:
nit: can you update the godoc on waitforcachesync to say "waits till started and synced" or somesuch?  Otherwise lgtm

PR #760, opened 2020-01-16
:sparkles: Parameterize webhook key/cert name

qinqon:
A user may want to use different names for server key/cert, for
example, if it's using the client-go cert manager util, the
certificate is just a unify pem file with both on it.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -33,11 +33,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics"
 )
 
-const (
-	certName = "tls.crt"
-	keyName  = "tls.key"
-)
-
 // DefaultPort is the default port that the webhook server serves.
 var DefaultPort = 443
 
@@ -53,9 +48,14 @@ type Server struct {
 	Port int
 
 	// CertDir is the directory that contains the server key and certificate. The
-	// server key and certificate must be named tls.key and tls.crt,
-	// respectively.
+	// server key and certificate.
 	CertDir string
+
+	// CertName is the server certificate name. defaults to tls.crt
+	CertName string
+
+	// CertName is the server key name. defaults to tls.key
+	KeyName string
 
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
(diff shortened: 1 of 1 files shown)


qinqon:
I have the CA signed for quique.llorente@gmail.com

pkg/webhook/server.go
```diff
@@ -53,10 +48,15 @@ type Server struct {
 	Port int
 
 	// CertDir is the directory that contains the server key and certificate. The
-	// server key and certificate must be named tls.key and tls.crt,
-	// respectively.
+	// server key and certificate.
 	CertDir string
 
+	// CertName is the server certificate name. default tls.crt
```
alenkacz: rather "defaults to 'tls.crt'" ?

pkg/webhook/server.go
```diff
@@ -53,10 +48,15 @@ type Server struct {
...
+	CertName string
+
+	// CertName is the server key name. default tls.key
```
alenkacz: rather "defaults to 'tls.key'" ?

pkg/webhook/server.go
qinqon: Done.

pkg/webhook/server.go
qinqon: Done.

qinqon:
@alenkacz Fixed the nits, can we re-run the tests ?

pkg/webhook/server.go
```diff
@@ -53,10 +48,15 @@ type Server struct {
 	Port int
 
 	// CertDir is the directory that contains the server key and certificate. The
-	// server key and certificate must be named tls.key and tls.crt,
-	// respectively.
+	// server key and certificate.
 	CertDir string
 
+	// CertName is the server certificate name. defaults to tls.crt
```
Adirio: NIT: Choose the one you like most:

```suggestion
	// CertName is the server certificate name. Defaults to tls.crt.
```
```suggestion
	// CertName is the server certificate name, defaults to tls.crt.
```

pkg/webhook/server.go
```diff
@@ -53,10 +48,15 @@ type Server struct {
...
+	CertName string
+
+	// CertName is the server key name. defaults to tls.key
```
Adirio: NIT: Choose the one you like most:
```suggestion
	// CertName is the server key name. Defaults to tls.key.
```
```suggestion
	// CertName is the server key name, defaults to tls.key.
```

qinqon:
/cc @alexeldeib 

Changed the stuff from @Adirio

DirectXMan12:
/approve

as per reviewers' LGTMs

PR #749, opened 2020-01-06
⚠ Bring in testing framework

hoegaarden:
This is bringing in https://github.com/kubernetes-sigs/testing_frameworks/tree/694245ac7b60bff90cc8b159fbd8ddb7efe3f218.

The package is fully integrated into the local tooling, e.g.:
- updated import path
- uses the same binary assets for testing
- runs its tests as part of `make test` and on prow
- updated OWNERS

In the original repo the CLA bot was enabled, thus all commits that have been
squashed here should be fine regarding the CLA.
All authors and committers of the squashed commits have been added here as
co-authors.

Superseds https://github.com/kubernetes-sigs/controller-runtime/pull/714 (see: https://github.com/kubernetes-sigs/controller-runtime/pull/714#issuecomment-564578313) 
/assign @DirectXMan12 @gerred

PR #762, opened 2020-01-18
 📖 (:book:, docs) Remove references to 'download-binaries.sh'

alenkacz:
When bringing in the testing framework in #749 download-binaries.sh was not moved over but the references to it remain in the docs. This PR removes them.

### DIFF
--- a/pkg/internal/testing/integration/assets/bin/.gitkeep
+++ b/pkg/internal/testing/integration/assets/bin/.gitkeep
@@ -1,1 +1,1 @@
-This directory will be the home of some binaries which are downloaded with `pkg/framework/test/scripts/download-binaries`.
+This directory will be the home of some binaries which are downloaded with `make test` or `.../hack/check-everything.sh`.
--- a/pkg/internal/testing/integration/doc.go
+++ b/pkg/internal/testing/integration/doc.go
@@ -8,9 +8,7 @@ needed to provide this API is managed by this framework.
 
 Quickstart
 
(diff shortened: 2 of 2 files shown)


pkg/internal/testing/integration/assets/bin/.gitkeep
```diff
@@ -1 +0,0 @@
-This directory will be the home of some binaries which are downloaded with `pkg/framework/test/scripts/download-binaries`.
```
hoegaarden: I think you might want to keep that, for now, as the assets are currently still being downloaded to `.../pkg/internal/testing/integration/assets/bin/`, see: https://github.com/kubernetes-sigs/controller-runtime/blob/8c39906c77cdb574482c01fcd74a76b105b06522/hack/check-everything.sh#L86.

alenkacz: yep, you're right, thanks. Updated...

hoegaarden: nit: If you want to remove references to 'download-binaries.sh' you might want to change the content of `.gitkeep` to something like:
```suggestion
This directory will be the home of some binaries which are downloaded with `make test` or `.../hack/check-everything.sh`.
```
🤷‍♂

alenkacz: oh, right 🙈 thanks

PR #763, opened 2020-01-18
✨ (:sparkles, minor) APIServer defaults deduplication

alenkacz:
After #749 the defaults to API server are now defined in two places - deep into the imported test framework and in envtest. I think it's enough to keep the defaults at one place - that is what this PR does.

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -72,21 +72,6 @@ func defaultAssetPath(binary string) string {
 
 }
 
-// DefaultKubeAPIServerFlags are default flags necessary to bring up apiserver.
-var DefaultKubeAPIServerFlags = []string{
-	// Allow tests to run offline, by preventing API server from attempting to
-	// use default route to determine its --advertise-address
-	"--advertise-address=127.0.0.1",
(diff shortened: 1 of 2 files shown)


pkg/envtest/server.go
```diff
@@ -156,10 +156,11 @@ func (te *Environment) Stop() error {
 }
 
 // getAPIServerFlags returns flags to be used with the Kubernetes API server.
+// it returns empty slice for api server defined defaults to be applied if no args specified
 func (te Environment) getAPIServerFlags() []string {
 	// Set default API server flags if not set.
 	if len(te.KubeAPIServerFlags) == 0 {
-		return DefaultKubeAPIServerFlags
+		return []string{}
```
DirectXMan12: should we alias the above global variable too?

vincepri: Isn't the `DefaultKubeAPIServerFlags` unused after this change?

alenkacz: I forgot to remove it 🤦‍♀ thanks

alenkacz:
@DirectXMan12 @gerred can you please take a look?

gerred:
/approve

looks good!

gerred:
hmm thought these tests were passing when I approved? @alenkacz cc

PR #767, opened 2020-01-20
 ✨  Add helpers to configure logger options via pflags

bharathi-tenneti:
This PR allows logging flags(/pkg/log/zap),  to be passed from command line in addition to functional calls.
Follow up from #639


pkg/log/zap/zap.go
```diff
@@ -207,3 +208,39 @@ func NewRaw(opts ...Opts) *zap.Logger {
 	log = log.WithOptions(o.ZapOpts...)
 	return log
 }
+
+// BindToFlagSet func allows CLI passed flags to be parsed.
+func (o *Options) BindFlags(fs *pflag.FlagSet) {
```
DirectXMan12: since there's not short flags here, can we use use "flag" instead?  That was we're more compatible with non-pflag setups.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+type encoderConfigFunc func(*zapcore.EncoderConfig)
+
+type encoderValue struct {
+	set     bool
+	setFunc func(zapcore.Encoder)
```
hasbro17: How about `setFunc`=>`setOptionFunc`? 
`setFunc` alone seems a bit vague.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (v encoderValue) String() string {
+	return v.str
+}
+
+func (v encoderValue) Type() string {
+	return "encoder"
+}
+
+func (v *encoderValue) Set(e string) error {
```
hasbro17:
```suggestion
func (v *encoderValue) Set(flagValue string) error {
```
I'd suggest being more verbose with the args names here and below so its easier to comprehend later.

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,39 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+	var stackVal stackTraceValue
+	stackVal.setFunc = func(fromFlag zap.AtomicLevel) {
+		o.StacktraceLevel = &fromFlag
+	}
+	fs.Var(&stackVal, "zap-stacktrace-level", "Zap log level (one of 'warn',  'error'")
+
+}
+
+// UseFlagOptions to set logger with CLI passed flags.
+func UseFlagOptions(in *Options) func(o *Options) {
```
hasbro17: @bharathi-tenneti How is this being used?
An example in the godocs for this would help since this exported. Plus I think we have an alias defined for the functional option type `func(o *Options)`:
```suggestion
func UseFlagOptions(in *Options) Opts {
```
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go#L73

pkg/log/zap/flags.go
joelanford: `setFunc` is the callback function that is called when `encoderValue.Set()` is called.

A `setFunc` callback is defined in `BindFlags` so that the `options.Encoder` value can be assigned when the flag is parsed. Just wanted to make sure the intent of the function was clear.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+	case "console":
+		v.setFunc(newConsoleEncoder())
+	default:
+		return fmt.Errorf("invalid encoder value \"%s\"", e)
+	}
+	v.str = e
+	return nil
+}
+
+func newJSONEncoder(ecfs ...encoderConfigFunc) zapcore.Encoder {
```
joelanford: In operator-sdk, we're using `encoderConfigFunc` to allow other flags to control other aspects of the log encoding (e.g. [the timestamp format](https://github.com/operator-framework/operator-sdk/blob/633b92c888ad3eeb9704441b7f6d6bd15f933d0e/pkg/log/zap/flags.go#L211-L215)). We're not doing that here (yet?), so maybe we should remove the code that uses it for now?

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+	switch lower {
+	case "debug":
+		v.setFunc(zap.NewAtomicLevelAt(zap.DebugLevel))
+	case "info":
+		v.setFunc(zap.NewAtomicLevelAt(zap.InfoLevel))
+	default:
+		return fmt.Errorf("invalid log level \"%s\"", l)
+	}
+	v.str = l
+	return nil
```
joelanford: In operator-sdk, we're allowing integer values as well (for increasing debug verbosity). We also hook this flag into the glog/klog verbosity as well.

https://github.com/operator-framework/operator-sdk/blob/633b92c888ad3eeb9704441b7f6d6bd15f933d0e/pkg/log/zap/flags.go#L105-L137

Do we want any of that here?

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
joelanford:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type levelValue struct {
+	set     bool
+	setFunc func(zap.AtomicLevel)
+	str     string
+}
+
+func (v *levelValue) Set(l string) error {
+	v.set = true
```
hasbro17: @joelanford Is `set` actually needed? 
E.g we're not currently checking this to see if the flag has already been set once.
And I don't see a method in the [Value](https://github.com/spf13/pflag/blob/HEAD/flag.go#L187:6) interface using this as well.

joelanford: Good catch. That's leftover from how we were doing defaulting in SDK (e.g. `if !set { v = defaultVal }`)

Here, we're just calling the setFunc and then deferring to the existing `addDefaults` function.

So I agree. I don't think `set` is needed with this implementation.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+	case "console":
+		ev.setFunc(newConsoleEncoder())
+	default:
+		return fmt.Errorf("invalid encoder value \"%s\"", flagValue)
+	}
+	ev.value = flagValue
+	return nil
+}
+
+func newJSONEncoder(ecfs ...encoderConfigFunc) zapcore.Encoder {
```
hasbro17:
```suggestion
func newJSONEncoder() zapcore.Encoder {
```
As @joelanford suggested let's remove the `encoderConfigFunc` type for now.

I think we'll first want to make the [EncoderConfig](https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig) more generally configurable via fields in the Options struct, and helper functions to set those options. 
And in a way so they don't get overridden by other Option fields like Development.

Then based on the fields in the Options struct that set EncoderConfig fields (e.g timestamp format and maybe more) we can add flags to configure those.

If it's simple enough to do, we could add it in this PR but I think a follow up might be better if not.

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,39 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+		o.Encoder = fromFlag
+	}
+	fs.Var(&encVal, "zap-encoder", "Zap log encoding ('json' or 'console')")
+
+	// Set the Log Level
+	var levelVal levelFlag
+	levelVal.setFunc = func(fromFlag zap.AtomicLevel) {
+		o.Level = &fromFlag
+	}
+	fs.Var(&levelVal, "zap-log-level", "Zap log level (one of 'debug', 'info', 'error' or any integer value > 0)")
```
hasbro17:
```suggestion
	fs.Var(&levelVal, "zap-log-level", "Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity")
```
Also we're not currently handling the cases for `error` or integer custom debug values in the `Set()` func above
for this flag.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (ev *stackTraceFlag) Set(flagValue string) error {
+	lower := strings.ToLower(flagValue)
+	switch lower {
+	case "warn":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.WarnLevel))
+	case "error":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.ErrorLevel))
+	default:
```
hasbro17: Handle the full range of [Zap Levels](https://godoc.org/go.uber.org/zap#pkg-constants) here.

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,39 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+	}
+	fs.Var(&stackVal, "zap-stacktrace-level", "Zap log level (one of 'warn',  'error'")
+
+}
+
+// UseFlagOptions to set logger with CLI passed flags.
+func UseFlagOptions(in *Options) Opts {
+	return func(o *Options) {
+		*o = *in
+		o.addDefaults()
```
hasbro17: Any reason to call `o.addDefaults()` here when it will be called in [`NewRaw()`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go#L200) as well?

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,39 @@ func NewRaw(opts ...Opts) *zap.Logger {
 	log = log.WithOptions(o.ZapOpts...)
 	return log
 }
+
+// BindToFlagSet func allows CLI passed flags to be parsed.
+func (o *Options) BindFlags(fs *pflag.FlagSet) {
+	// Set Development mode value
+	fs.BoolVar(&o.Development, "zap-devel", false, "Enable zap development mode (changes defaults to console encoder, debug log level, and disables sampling)")
```
hasbro17: `changes defaults to ...` We should also outline out what the default settings are if this is not set, i.e Production settings.

pkg/log/zap/flags.go
hasbro17: Still need to update this.

pkg/log/zap/zap.go
bharathi-tenneti: For example: from operator-SDK, we plan to call the function UseFlagOptions() directly.
See: https://github.com/learnoperators/memcached-operator/blob/master/cmd/manager/main.go#L66

bharathi-tenneti:
@hasbro17 Please take a look.

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,151 @@
...
+		ev.setFunc(zap.NewAtomicLevelAt(zap.InfoLevel))
+	case "error", "2":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.ErrorLevel))
+	default:
+		i, err := strconv.Atoi(lower)
+		if err != nil {
+			return fmt.Errorf("invalid log level \"%s\"", flagValue)
+		}
+		if i > 0 {
+			fmt.Println("Iam here")
```
camilamacedo86:
```suggestion
```

camilamacedo86: Could we add a log msg in this scenario such as: `setting log level as %s: lvl`

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,151 @@
...
+			lvl = -1 * i
+			ev.setFunc(zap.NewAtomicLevelAt(zapcore.Level(int8(lvl))))
```
camilamacedo86: WDYT about use level instead of lvl to make it more readable?

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,151 @@
...
+	var lvl int
+	switch lower {
+	case "debug", "-1":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.DebugLevel))
+	case "info", "0":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.InfoLevel))
+	case "error", "2":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.ErrorLevel))
+	default:
+		i, err := strconv.Atoi(lower)
```
camilamacedo86: WDYT about using a name which described its value?

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,41 @@ func NewRaw(opts ...Opts) *zap.Logger {
 	log = log.WithOptions(o.ZapOpts...)
 	return log
 }
+
+// BindToFlagSet func allows CLI passed flags to be parsed.
+func (o *Options) BindFlags(fs *flag.FlagSet) {
+
+	// Set Development mode value
+	fs.BoolVar(&o.Development, "zap-devel", false, "Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn)."+
```
estroz:
```suggestion
	fs.BoolVar(&o.Development, "zap-devel", false,
		"Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). "+
```

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,41 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+		o.Encoder = fromFlag
+	}
+	fs.Var(&encVal, "zap-encoder", "Zap log encoding ('json' or 'console')")
+
+	// Set the Log Level
+	var levelVal levelFlag
+	levelVal.setFunc = func(fromFlag zap.AtomicLevel) {
+		o.Level = &fromFlag
+	}
+	fs.Var(&levelVal, "zap-log-level", "Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error',"+
```
estroz:
```suggestion
	fs.Var(&levelVal, "zap-log-level",
		"Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', "+
```

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,41 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+		"or any integer value > 0 which corresponds to custom debug levels of increasing verbosity")
+
+	// Set the StrackTrace Level
+	var stackVal stackTraceFlag
+	stackVal.setFunc = func(fromFlag zap.AtomicLevel) {
+		o.StacktraceLevel = &fromFlag
+	}
+	fs.Var(&stackVal, "zap-stacktrace-level", "Zap Level at and above which stacktraces are captured (one of 'warn' or 'error'")
```
estroz:
```suggestion
	fs.Var(&stackVal, "zap-stacktrace-level",
		"Zap Level at and above which stacktraces are captured (one of 'warn' or 'error')")
```

pkg/log/zap/flags.go
```diff
@@ -0,0 +1,151 @@
...
+	value   string
+}
+
+var _ pflag.Value = &stackTraceFlag{}
+
+func (ev *stackTraceFlag) Set(flagValue string) error {
+	lower := strings.ToLower(flagValue)
+	switch lower {
+	case "debug":
+		ev.setFunc(zap.NewAtomicLevelAt(zap.DebugLevel))
```
estroz: These could be memoized in a map:

```go
var levelStrings = map[string]zapcore.Level{
	"debug": zap.DebugLevel,
	"-1":    zap.DebugLevel,
	...
}

func (ev *stackTraceFlag) Set(flagValue string) error {
	level, validLevel := levelStrings[strings.ToLower(flagValue)]
	if !validLevel {
		return fmt.Errorf("invalid stacktrace level \"%s\"", flagValue)
	}
	ev.setFunc(zap.NewAtomicLevel(level))
	ev.value = flagValue
	return nil
}
```

Applies to `levelFlag.Set` too; they can use the same map.

pkg/log/zap/zap.go
hasbro17: `zap.New()` calls `zap.NewRaw()` which calls `o.addDefaults()` after applying the flag options. 
So it ends up being the same thing.
So not sure why we'd call it here before that.

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,44 @@ func NewRaw(opts ...Opts) *zap.Logger {
 	log = log.WithOptions(o.ZapOpts...)
 	return log
 }
+
+// BindToFlagSet func allows CLI passed flags to be parsed.
```
hasbro17:
```suggestion
// BindFlagSet will parse the given flagset for zap option flags and set the log options accordingly 
```
Also for the godoc comment here, can you add the list of flag names that BindFlags currently parses.

pkg/log/zap/zap.go
```diff
@@ -207,3 +208,44 @@ func NewRaw(opts ...Opts) *zap.Logger {
...
+	// Set the StrackTrace Level
+	var stackVal stackTraceFlag
+	stackVal.setFunc = func(fromFlag zap.AtomicLevel) {
+		o.StacktraceLevel = &fromFlag
+	}
+	fs.Var(&stackVal, "zap-stacktrace-level",
+		"Zap Level at and above which stacktraces are captured (one of 'warn' or 'error')")
+}
+
+// UseFlagOptions to set logger with CLI passed flags.
```
hasbro17: Add an example code block to show how this is used in combination with `opts.BindFlags`.
```suggestion
// UseFlagOptions configures the logger to use the Options set by parsing zap option flags from the CLI.
//  opts := zap.Options{}
//  opts.BindFlags(flag.CommandLine)
//  log := zap.New(zap.UseFlagOptions(&opts))
```

hasbro17:
@gerred This got merged with the commits unsquashed.
https://github.com/kubernetes-sigs/controller-runtime/commits/master

@DirectXMan12 I don't think we can clean this up now but is there any way to avoid this in the future? e.g have the bot squash them for us or a linter to make sure we clean up before merging.

gerred:
@hasbro17 Yeah, this ended up being a little lax.

@shawn-hurley maybe we reverse the relationship in the future as well. As an approver I'm typically looking for those squashed commits when I approve, but _I_ didn't take the due diligence to look back and see that it had already been approved before I /lgtmed. I'll do so in the future, but we should work together to make sure the reviewer is looking at the changes in and of itself and the approver is doing the same work, but _also_ making sure it meets the criteria for being merged. I think it'd nbd but we can always stand to improve project hygiene!

DirectXMan12:
oh yeah.  We should add a CI check or something for this

PR #773, opened 2020-01-22
:running: Update linter and lint fixes

mcristina422:
This updates golangci-lint, moves the configuration all to `.golangci.yml` and then fixes the linter errors

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,27 @@
+linters:
+  disable-all: true
+  enable:
+    - misspell
+    - structcheck
+    - golint
+    - deadcode
+    - errcheck
+    - varcheck
+    - goconst
+    - unparam
+    - ineffassign
+    - nakedret
+    - interfacer
+    - gocyclo
+    - lll
+    - dupl
+    - goimports
+
 linters-settings:
   lll:
     line-length: 170
   dupl:
     threshold: 400
+
+run:
+  timeout: 5m
--- a/hack/check-everything.sh
+++ b/hack/check-everything.sh
@@ -78,7 +78,7 @@ function golangci_lint_has_version {
 
 function fetch_go_tools {
   header_text "Checking for golangci-lint"
-  local golangci_lint_version="v1.21.0"
+  local golangci_lint_version="v1.23.6"
   if ! is_installed golangci-lint || ! golangci_lint_has_version "${golangci_lint_version}"; then
     header_text "Installing golangci-lint"
     curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin "${golangci_lint_version}"
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -24,30 +24,6 @@ go vet ${MOD_OPT} ./...
 
 header_text "running golangci-lint"
 
-golangci-lint run --disable-all \
-    --deadline 5m \
-    --enable=misspell \
-    --enable=structcheck \
-    --enable=golint \
-    --enable=deadcode \
-    --enable=errcheck \
-    --enable=varcheck \
-    --enable=goconst \
-    --enable=unparam \
-    --enable=ineffassign \
-    --enable=nakedret \
-    --enable=interfacer \
-    --enable=misspell \
-    --enable=gocyclo \
-    --enable=lll \
-    --enable=dupl \
-    --enable=goimports \
-    --enable=bodyclose \
-    ./pkg/... ./examples/... .
-
-# TODO: Enable these as we fix them to make them pass
-#    --enable=gosec \
-#    --enable=maligned \
-#    --enable=safesql \
+golangci-lint run ./pkg/... ./examples/... .
 
 GO111MODULES=on go list -mod=readonly ./...
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package fake
 
 import (
+	"context"
 	"encoding/json"
 
 	. "github.com/onsi/ginkgo"
@@ -85,7 +86,7 @@ var _ = Describe("Fake client", func() {
 				Namespace: "ns1",
 			}
 			obj := &appsv1.Deployment{}
-			err := cl.Get(nil, namespacedName, obj)
+			err := cl.Get(context.Background(), namespacedName, obj)
 			Expect(err).To(BeNil())
 			Expect(obj).To(Equal(dep))
 		})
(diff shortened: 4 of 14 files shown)


pkg/internal/testing/integration/etcd.go
```diff
@@ -86,7 +86,7 @@ func (e *Etcd) setProcessState() error {
 		return err
 	}
 
-	e.processState.StartMessage = internal.GetEtcdStartMessage(e.processState.URL)
+	e.processState.HealthCheckEndpoint = "/health"
```
DirectXMan12: o_O

pkg/manager/signals/signal_test.go
```diff
@@ -36,7 +36,7 @@ var _ = Describe("runtime signal", func() {
 			task := &Task{
 				ticker: time.NewTicker(time.Second * 2),
 			}
-			c := make(chan os.Signal)
+			c := make(chan os.Signal, 1)
```
DirectXMan12: context on this one?

pkg/internal/testing/integration/etcd.go
mcristina422: `StartMessage` is deprecated. So staticcheck is failing. I was iffy on this change too, we can add `//nolint` to revert this change

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/testing/integration/internal/process.go#L40

pkg/manager/signals/signal_test.go
mcristina422: `SA1017: the channel used with signal.Notify should be buffered`
https://staticcheck.io/docs/checks#SA1017

.golangci.yml
```diff
@@ -3,3 +3,5 @@ linters-settings:
     line-length: 170
   dupl:
     threshold: 400
+run:
+  deadline: 5m
```
kensipe: `deadline` is deprecated... we should switch to `timeout` 

details: https://github.com/golangci/golangci-lint/pull/822

hack/verify.sh
```diff
@@ -24,29 +24,6 @@ go vet ${MOD_OPT} ./...
...
-    --enable=lll \
-    --enable=dupl \
-    --enable=goimports \
-    ./pkg/... ./examples/... .
-
-# TODO: Enable these as we fix them to make them pass
-#    --enable=gosec \
-#    --enable=maligned \
-#    --enable=safesql \
+golangci-lint run ./pkg/... ./examples/... .
```
kensipe: The following linters have been removed and are NOT enabled by default

```
--enable=misspell \	   
   --enable=golint \	
   --enable=goconst \	
   --enable=unparam \	
   --enable=nakedret \	
   --enable=interfacer \	
   --enable=gocyclo \	
   --enable=lll \	
   --enable=dupl \	
   --enable=goimports \
```

Is this intended?   If we plan to keep them... it would be great to add them to the `.golangci.yml` file.

https://github.com/golangci/golangci-lint#enabled-by-default-linters

mcristina422: No, this was not my intention. I misinterpreted how dropping the `--disable-all` flag worked. I'll make sure to explicitly enable the same ones in the config file

mcristina422:
@kensipe I've reverted the etcd change. Since `staticcheck` isn't enabled currently this will pass linting

In a follow up we can debate enabling more like `staticcheck`, `stylecheck`, `whitespace` and others found in the more strict configuration that golangci uses themselves https://github.com/golangci/golangci-lint/blob/master/.golangci.yml

kensipe:
@DirectXMan12 could we get a 2nd look on this?

pkg/manager/signals/signal_test.go
gerred: ah, prevents missing signals, ok.

hack/check-everything.sh
```diff
@@ -74,7 +74,7 @@ function fetch_go_tools {
   header_text "Checking for gometalinter.v2"
   if ! is_installed golangci-lint; then
     header_text "Installing golangci-lint"
-    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
+    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.23.1
```
gerred: long term maybe we introduce a tools.go to manage these dependencies with go modules - a few other tools do this.

kensipe:
@mcristina422 looks like it needs a rebase... lets see if we can get some attention on this after the rebase... thx

pkg/client/fake/client_test.go
```diff
@@ -235,7 +236,7 @@ var _ = Describe("Fake client", func() {
 					Namespace: "ns2",
 				}
 				obj := &corev1.ConfigMap{}
-				err = cl.Get(nil, namespacedName, obj)
```
camilamacedo86: Why this change is required?

pkg/controller/controller_integration_test.go
```diff
@@ -152,7 +152,7 @@ var _ = Describe("controller", func() {
 			By("Invoking Reconciling for an OwnedObject when it is updated")
 			newReplicaset := replicaset.DeepCopy()
 			newReplicaset.Labels = map[string]string{"foo": "bar"}
-			newReplicaset, err = clientset.AppsV1().ReplicaSets("default").Update(newReplicaset)
+			_, err = clientset.AppsV1().ReplicaSets("default").Update(newReplicaset)
```
camilamacedo86: 👍 cool

hack/verify.sh
```diff
@@ -24,29 +24,6 @@ go vet ${MOD_OPT} ./...
 
 header_text "running golangci-lint"
 
-golangci-lint run --disable-all \
```
camilamacedo86: 👍 cool

.golangci.yml
```diff
@@ -1,5 +1,27 @@
...
+    - goimports
+
 linters-settings:
   lll:
     line-length: 170
   dupl:
     threshold: 400
+
+run:
+  timeout: 5m
```
camilamacedo86: 👍 cool

.golangci.yml
```diff
@@ -1,5 +1,27 @@
+linters:
+  disable-all: true
```
vincepri: I'd suggest to flip these. Enable all linters and explicitly disable the others, so we know which ones we are disabling and folks can pick up the task of removing a disabled linter from this list if they wanted to.

pkg/client/fake/client_test.go
mcristina422: When running the linter with staticcheck enabled (not enforced by this PR) it fails with

`SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)`

https://staticcheck.io/docs/checks#SA1012

.golangci.yml
mcristina422: This is explicitly called out here because enable-all is deprecated  

https://github.com/golangci/golangci-lint/blob/b9eef79121fff235d0d794c176ffa2b3d9bd422f/.golangci.yml#L59-L60

vincepri: Oh well, it seems a bit counter intuitive to me, but nothing major

pkg/internal/testing/integration/etcd.go
camilamacedo86: Before it was `internal.GetEtcdStartMessage(e.processState.URL)`, however,, now it is receiving a fixed value. Is it accurate? Could not it cause issues?

PR #777, opened 2020-01-23
🏃 (:running:, other) Try to stabilize the flaky process_test

alenkacz:
The test is flaky, see https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/592/pull-controller-runtime-test-master/1220082018474266624 or https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/776/pull-controller-runtime-test-master/1220286274334625792

My hunch was that 20 millis is not enoughso I tried fixing the test by using much bigger poll interval.


alenkacz:
@gerred can you please re-approve?

PR #779, opened 2020-01-23
:running: Use scheme.Convert to convert unstructrured objects

vincepri:

This PR updates the envtest helper to use a local scheme and the `Convert` method, which should preserve the TypeMeta information.

`runtime.DefaultUnstructuredConverter` doesn't know how to populate `TypeMeta` information from the Go Struct, while `scheme.Convert` uses internal helper methods to figure that out. 

From a UX point of view, this change has the benefit that `envtest` users can use the Go structs without worrying about TypeMeta. If TypeMeta isn't present when we install CRDs, the installation will fail and `envtest` won't start.

### DIFF
--- a/pkg/envtest/helper.go
+++ b/pkg/envtest/helper.go
@@ -3,9 +3,21 @@ package envtest
 import (
 	"reflect"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 )
+
+var (
+	crdScheme = runtime.NewScheme()
+)
+
+// init is required to correctly initialize the crdScheme package variable.
+func init() {
+	_ = apiextensionsv1.AddToScheme(crdScheme)
+	_ = apiextensionsv1beta1.AddToScheme(crdScheme)
+}
 
 // mergePaths merges two string slices containing paths.
 // This function makes no guarantees about order of the merged slice.
@@ -30,10 +42,10 @@ func mergePaths(s1, s2 []string) []string {
 // This function makes no guarantees about order of the merged slice.
 func mergeCRDs(s1, s2 []runtime.Object) []runtime.Object {
 	m := make(map[string]*unstructured.Unstructured)
-	for _, obj := range runtimeListToUnstructured(s1) {
+	for _, obj := range runtimeCRDListToUnstructured(s1) {
 		m[obj.GetName()] = obj
 	}
-	for _, obj := range runtimeListToUnstructured(s2) {
+	for _, obj := range runtimeCRDListToUnstructured(s2) {
 		m[obj.GetName()] = obj
 	}
 	merged := make([]runtime.Object, len(m))
@@ -57,21 +69,20 @@ func existsUnstructured(s1, s2 []*unstructured.Unstructured) bool {
 	return false
 }
 
-func runtimeListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
+func runtimeCRDListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
 	res := []*unstructured.Unstructured{}
 	for _, obj := range l {
-		m, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
-		if err != nil {
+		u := &unstructured.Unstructured{}
+		if err := crdScheme.Convert(obj, u, nil); err != nil {
+			log.Error(err, "error converting to unstructured object", "object-kind", obj.GetObjectKind())
 			continue
 		}
-		res = append(res, &unstructured.Unstructured{
-			Object: m,
-		})
+		res = append(res, u)
 	}
 	return res
 }
 
-func unstructuredListToRuntime(l []*unstructured.Unstructured) []runtime.Object {
+func unstructuredCRDListToRuntime(l []*unstructured.Unstructured) []runtime.Object {
 	res := []runtime.Object{}
 	for _, obj := range l {
 		res = append(res, obj.DeepCopy())
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -114,7 +114,7 @@ func defaultCRDOptions(o *CRDInstallOptions) {
 func WaitForCRDs(config *rest.Config, crds []runtime.Object, options CRDInstallOptions) error {
 	// Add each CRD to a map of GroupVersion to Resource
 	waitingFor := map[schema.GroupVersion]*sets.String{}
-	for _, crd := range runtimeListToUnstructured(crds) {
+	for _, crd := range runtimeCRDListToUnstructured(crds) {
 		gvs := []schema.GroupVersion{}
 		crdGroup, _, err := unstructured.NestedString(crd.Object, "spec", "group")
 		if err != nil {
@@ -230,7 +230,7 @@ func UninstallCRDs(config *rest.Config, options CRDInstallOptions) error {
 	}
 
 	// Uninstall each CRD
-	for _, crd := range runtimeListToUnstructured(options.CRDs) {
+	for _, crd := range runtimeCRDListToUnstructured(options.CRDs) {
 		log.V(1).Info("uninstalling CRD", "crd", crd.GetName())
 		if err := cs.Delete(context.TODO(), crd); err != nil {
 			// If CRD is not found, we can consider success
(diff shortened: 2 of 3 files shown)


pkg/envtest/helper.go
```diff
@@ -3,10 +3,21 @@ package envtest
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var (
+	crdScheme = runtime.NewScheme()
+)
+
+func init() {
+	_ = apiextensionsv1.AddToScheme(crdScheme)
```
rajathagasthya: You could use:

```go
import apiextensionsinstall "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/install"

apiextensionsinstall.Install(crdScheme)
```

vincepri: Ah nice, didn't know that existed. Fixing soon!

gerred: wow TIL this affects a lot of my own code :joy:

pkg/envtest/helper.go
```diff
@@ -3,10 +3,19 @@ package envtest
...
+	apiextensionsinstall "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/install"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var (
+	crdScheme = runtime.NewScheme()
+)
+
+func init() {
```
gerred: in general I at least want to discuss why we have an `init` anywhere that this is required. This is for integration tests, and I _strongly believe_ it generally makes sense, but maybe some inline documentation that this is an important external dependency for tests would help for future PRs that affect this code.

I've just been bit enough by init that when encountering it we should:

1) remove it
2) if not remove it, have very strong documentation for it that is well maintained.

gerred:
@vincepri my comments are generally around documentation because I agree with the organization here. Given I'm in queue to be an approver, I will save my actions for an approval and would ask @alenkacz, @shawn-hurley, @alvaroaleman, @kensipe, or @joelanford to provide the proper lgtm before I approve.

pkg/envtest/helper.go
rajathagasthya: Just FYI that `Install` will panic on an error, which is okay since this is envtest.

pkg/envtest/helper.go
```diff
@@ -57,21 +66,19 @@ func existsUnstructured(s1, s2 []*unstructured.Unstructured) bool {
...
 }
 
-func runtimeListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
+func runtimeCRDListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
 	res := []*unstructured.Unstructured{}
 	for _, obj := range l {
-		m, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
-		if err != nil {
+		u := &unstructured.Unstructured{}
+		if err := crdScheme.Convert(obj, u, nil); err != nil {
```
alenkacz: should we at least log something here? we're pretty much just silently skipping resources here, right?

vincepri:
Fixed both comments, ptal

vincepri:
ping @gerred

pkg/envtest/helper.go
```diff
@@ -3,10 +3,21 @@ package envtest
...
 )
 
+var (
+	crdScheme = runtime.NewScheme()
+)
+
+// init is required to correctly initialize the crdScheme package variable.
+// Installs the CRDv1 and CRDv1beta1 from the apiextensions package.
+func init() {
+	apiextensionsinstall.Install(crdScheme)
```
DirectXMan12: prob better to explicitly add understood versions here

pkg/envtest/helper.go
```diff
@@ -57,21 +68,20 @@ func existsUnstructured(s1, s2 []*unstructured.Unstructured) bool {
...
 }
 
-func runtimeListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
+func runtimeCRDListToUnstructured(l []runtime.Object) []*unstructured.Unstructured {
 	res := []*unstructured.Unstructured{}
 	for _, obj := range l {
-		m, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
-		if err != nil {
+		u := &unstructured.Unstructured{}
+		if err := crdScheme.Convert(obj, u, nil); err != nil {
```
DirectXMan12: can avoid scheme setup entirely if we use `k8s.io/apimachinery/pkg/runtime.DefaultUnstructuredConverter`, which does the equivalent of the JSON encoder instead.

vincepri: Yeah that's what it was there before, although `DefaultUnstructuredConverter` loses TypeMeta information when converting from struct -> unstructured

vincepri: Quick test with a Cluster API object:

```
        spew.Dump(runtime.DefaultUnstructuredConverter.ToUnstructured(&v1alpha3.Cluster{}))
	out := &unstructured.Unstructured{}
	scheme.Convert(&v1alpha3.Cluster{}, out, nil)
	spew.Dump(out)
```

prints:

```
(map[string]interface {}) (len=3) {
 (string) (len=8) "metadata": (map[string]interface {}) (len=1) {
  (string) (len=17) "creationTimestamp": (interface {}) <nil>
 },
 (string) (len=4) "spec": (map[string]interface {}) (len=1) {
  (string) (len=20) "controlPlaneEndpoint": (map[string]interface {}) (len=2) {
   (string) (len=4) "host": (string) "",
   (string) (len=4) "port": (int64) 0
  }
 },
 (string) (len=6) "status": (map[string]interface {}) (len=2) {
  (string) (len=19) "infrastructureReady": (bool) false,
  (string) (len=23) "controlPlaneInitialized": (bool) false
 }
}
```

```
(*unstructured.Unstructured)(0xc000104028)({
 Object: (map[string]interface {}) (len=5) {
  (string) (len=8) "metadata": (map[string]interface {}) (len=1) {
   (string) (len=17) "creationTimestamp": (interface {}) <nil>
  },
  (string) (len=4) "spec": (map[string]interface {}) (len=1) {
   (string) (len=20) "controlPlaneEndpoint": (map[string]interface {}) (len=2) {
    (string) (len=4) "host": (string) "",
    (string) (len=4) "port": (int64) 0
   }
  },
  (string) (len=6) "status": (map[string]interface {}) (len=2) {
   (string) (len=19) "infrastructureReady": (bool) false,
   (string) (len=23) "controlPlaneInitialized": (bool) false
  },
  (string) (len=10) "apiVersion": (string) (len=25) "cluster.x-k8s.io/v1alpha3",
  (string) (len=4) "kind": (string) (len=7) "Cluster"
 }
})
```

gerred: i've seen this too. is there a relevant apimachinery issue (or do we know if this is intended behavior?)

pkg/envtest/helper.go
rajathagasthya: It was explicit before, but I recommended to use this one-liner :/

pkg/envtest/helper.go
vincepri: https://github.com/kubernetes/kubernetes/issues/80609

DirectXMan12: > Yeah that's what it was there before,

Lol, I can definitely read diffs :-P

I think at least in your example, it's not actually clearing it, it's just not set and it doesn't know how to populate it (since Go struct <-> gvk isn't technically 1-to-1, for some reason that I'm never managed to extract from the folks who might know).  This makes more sense then.

DirectXMan12: Can you add these details to the PR description?

vincepri: Sure thing

vincepri: Done!

pkg/envtest/helper.go
vincepri: Any preferences here?

gerred: Avoiding buzzwords, but here we are since I couldn't type it without them --

controller-runtime is an abstraction. We should use the highest form of abstraction that our dependencies afford us. I'm :+1: on using that version, and including that information in our release notes (as of this version of CR on Kubernetes 1.17, our API code targets CRDv1)

gerred: @vincepri wdyt?

gerred:
Pending @DirectXMan12 and @vincepri updating on our API versions, I'm:

/approve

pkg/envtest/helper.go
gerred: To be a bit more concrete, too - I don't think CR tooling should bleed across the wire. Otherise, we have the API server, client-go, CR, and others all managing this concern -- which in this case should be determined by our pinned version of this dependency, not necessarily by us.

In a larger scope, this gets into minimal version selection (of which that given version, in our case CRD/v1beta1, will be eventually deprecated and removed). I'd lean toward choosing the "storage" version indicated by a given upstream dependency version, which I would assume that these helper methods/funcs would choose.

If they don't, that's when we override.

vincepri:
@gerred @DirectXMan12 imports have been fixed to be explicit, ptal :)

Issue #764, opened 2020-01-20
UMBRELLA: design and refactor graceful termination


tapih:
To shutdown controller gracefully, containers should be terminated after endpoints deletion.

If `time.Sleep(someSeconds)` is added between [these lines](https://github.com/kubernetes-sigs/controller-runtime/blob/45d652077cba5ccc29621e8db615f25084c701cb/pkg/manager/signals/signal.go#L36-L37), applications made with controller runtime can wait for endpoint deletion as default.
What do you think?

alexeldeib:
See some of the previous discussions about graceful termination:
- https://github.com/kubernetes-sigs/controller-runtime/issues/730
- https://github.com/kubernetes-sigs/controller-runtime/issues/695
- https://github.com/kubernetes-sigs/controller-runtime/issues/350

I think a better approach than sleeping (and one that's been discussed a few times) is properly wiring up all the contexts (well, currently stop channels) and using either a wait group (for manager shutdown) or or some sort of map + lock + shutdown mechanism (for manager shutdown + dynamically (un)loading controllers). 

There's a tentative desire to replace all the stop channels with context plumbed all the way through -- graceful manager termination could plausibly be done in tandem with that change (or not, but seems natural).

tapih:
Sorry for my poor explanation.  What I meant to propose is that, by inserting sleep just after the signal handler, we can omit `/bin/sleep` in `preStop` hook of Kubernetes Pods.

When a Pod is shutting down, pod termination and endpoint deletion are executed at the same time.
The problem is that some requests cannot be accepted if the endpoint deletion does not finish while the pod termination finishes.
I meant by the word "graceful shutdown" in the previous comment to avoid this situation by ensuring that the pod is terminated after the endpoint is deleted.

One way to wait for endpoint deletion is to add the `sleep` command into `preStop`, but if a base image does not have the `sleep` command, this approach cannot be taken.
So, I would like to add `time.Sleep(someSeconds)` to resolve this problem at the framework level.

vincepri:
/kind design

We'd need a design document in form of PR to the controller-runtime repository.

/help
/priority important-soon

seh:
Presumably the duration to sleep there would need to be configurable, with zero meaning not to sleep at all. But sleeping like this is a brittle cover for the lack of proper coordination. The chosen duration always turns out to be wrong.

alexeldeib:
[...]

PR #782, opened 2020-01-28
✨ Enable signal handler to close channel after sleeping for some seconds

tapih:
This PR is an implementation of https://github.com/kubernetes-sigs/controller-runtime/issues/764.
By inserting sleep just after the signal handler, we can wait for the endpoint of a k8s pod to be deleted without `/bin/sleep` in `preStop` hook.

The default sleeping period is 0sec, so this PR does not change the current behavior as long as the environment variable is not set.

### DIFF
--- a/pkg/manager/signals/doc.go
+++ b/pkg/manager/signals/doc.go
@@ -18,3 +18,9 @@ limitations under the License.
 // shutdown the manager in combination with Kubernetes pod graceful termination
 // policy.
 package signals
+
+import (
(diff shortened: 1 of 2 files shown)


tapih:
I signed it

tapih:
I signed it

DirectXMan12:
/hold

this needs discussion on the related issue I think

vincepri:
@tapih Thanks for the contribution, we're taking a step back in #764 to design stop behaviors. Closing this for now.

/close

PR #784, opened 2020-01-29
:running: Clean up OWNERS file

DirectXMan12:
This cleans up the owners file, making sure that only active members
will be assigned and bringing it in line with the latest docs on the
OWNERS file from the main k/community repo.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,11 +1,19 @@
-# See the OWNERS docs: https://git.k8s.io/community/contributors/devel/owners.md
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
 
 aliases:
+  # active folks who can be contacted to perform admin-related
+  # tasks on the repo, or otherwise approve any PRS.
   controller-runtime-admins:
   - directxman12
   - droot
-  - pwittrock
-  controller-runtime-maintainers: []
+  - mengqiy
+
+  # non-admin folks who can approve any PRs in the repo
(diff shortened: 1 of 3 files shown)


OWNERS_ALIASES
```diff
@@ -1,18 +1,34 @@
...
+  # tasks on the repo, or otherwise approve any PRS.
   controller-runtime-admins:
   - directxman12
   - droot
-  - pwittrock
-  controller-runtime-maintainers: []
+  - mengqiy
+
+  # non-admin folks who can approve any PRs in the repo
+  controller-runtime-approvers: []
```
vincepri: Should we add @gerred as well as part of this PR?

OWNERS_ALIASES
```diff
@@ -1,18 +1,34 @@
...
+
+  # folks who can review and LGTM any PRs in the repo (doesn't
+  # include approvers & admins -- those count too via the OWNERS
+  # file)
   controller-runtime-reviewers:
   - alvaroaleman
```
vincepri: Should we add other folks here as well?

DirectXMan12:
> If we do the suggested changes, we can probably close the other PRs and do a batch change to the owners file

I was gonna do them separately to have a good PR/commit trail

OWNERS_ALIASES
DirectXMan12: this is just copied from the existing stuff

vincepri: sgtm

PR #786, opened 2020-01-31
🐛 Integration test should close http body

vincepri:
This PR makes sure the `pollURLUntilOK` function closes the body from `http.Get` if there is no error, and adds `bodyclose` linter to the enabled linters.

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -42,6 +42,7 @@ golangci-lint run --disable-all \
     --enable=lll \
     --enable=dupl \
     --enable=goimports \
+    --enable=bodyclose \
     ./pkg/... ./examples/... .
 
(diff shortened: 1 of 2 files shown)


alenkacz:
@vincepri I fixed that flake 🤞  - it's merged on master, you should rebase and rerun the tests :)

alvaroaleman:
@alenkacz Prow merges the PR prior to running tests, so the last run did have your change. You can see that by looking at the test page, then clicking on "ProwJob Yaml", then basesha, which is 5e2c4c885d68f1282531f42dc9cecc908ad5da3f, which is the latest master

alenkacz:
@alvaroaleman oooh, nice. but not nice that the fix did not work 😢

PR #752, opened 2020-01-08
⚠ Add testenv support for CRDv1/CRDv1beta1 + update to k8s 1.16.4

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>

This PR adds support for testenv to load CRDv1 and CRDv1beta1 by using `runtime.Object` and controller-runtime unstructured client.

PR #787, opened 2020-02-01
⚠ (:warning:, major) Webhook support in envtest

### LINKED ISSUE #563: Envtest does not register or run webhooks
Currently the kubebuilder book shows how to write a defaulting/validating webhook, but it doesn't give any envtest examples for how to test it (only a light explanation of how to deploy it to a real cluster).

We should provide a method for testing webhooks the same way we provide a means of testing the controller and types using envtest.

alenkacz:
Ability to test webhooks with envtest

lots of code is reused from PoC done in #645

Fixes #563

Co-authored-by: Solly Ross <sollyross@google.com>

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -96,6 +96,14 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			Expect(cfg).NotTo(BeNil())
 
 			By("creating three pods")
+			cl, err := client.New(cfg, client.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			err = ensureNamespace(testNamespaceOne, cl)
+			Expect(err).NotTo(HaveOccurred())
+			err = ensureNamespace(testNamespaceTwo, cl)
+			Expect(err).NotTo(HaveOccurred())
+			err = ensureNamespace(testNamespaceThree, cl)
+			Expect(err).NotTo(HaveOccurred())
 			// Includes restart policy since these objects are indexed on this field.
 			knownPod1 = createPod("test-pod-1", testNamespaceOne, kcorev1.RestartPolicyNever)
 			knownPod2 = createPod("test-pod-2", testNamespaceTwo, kcorev1.RestartPolicyAlways)
@@ -111,7 +119,6 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			knownPod4.GetObjectKind().SetGroupVersionKind(podGVK)
 
 			By("creating the informer cache")
-			var err error
 			informerCache, err = createCacheFunc(cfg, cache.Options{})
 			Expect(err).NotTo(HaveOccurred())
 			By("running the cache and waiting for it to sync")
@@ -680,4 +687,22 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			})
 		})
 	})
+}
+
+// ensureNamespace installs namespace of a given name if not exists
+func ensureNamespace(namespace string, client client.Client) error {
+	ns := kcorev1.Namespace{
+		ObjectMeta: kmetav1.ObjectMeta{
+			Name: namespace,
+		},
+		TypeMeta: kmetav1.TypeMeta{
+			Kind:       "Namespace",
+			APIVersion: "v1",
+		},
+	}
+	err := client.Create(context.TODO(), &ns)
+	if errors.IsAlreadyExists(err) {
+		return nil
+	}
+	return err
 }
--- a/pkg/envtest/envtest_suite_test.go
+++ b/pkg/envtest/envtest_suite_test.go
@@ -21,6 +21,10 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	admissionv1 "k8s.io/api/admissionregistration/v1"
+	admissionv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
@@ -35,11 +39,100 @@ var env *Environment
 var _ = BeforeSuite(func(done Done) {
 	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
 	env = &Environment{}
+	// we're initializing webhook here and not in webhook.go to also test the envtest install code via WebhookOptions
+	initializeWebhookInEnvironment()
 	_, err := env.Start()
 	Expect(err).NotTo(HaveOccurred())
 
 	close(done)
 }, StartTimeout)
+
+func initializeWebhookInEnvironment() {
+	namespacedScopeV1Beta1 := admissionv1beta1.NamespacedScope
+	namespacedScopeV1 := admissionv1.NamespacedScope
+	failedTypeV1Beta1 := admissionv1beta1.Fail
+	failedTypeV1 := admissionv1.Fail
+	equivalentTypeV1Beta1 := admissionv1beta1.Equivalent
+	equivalentTypeV1 := admissionv1.Equivalent
+	noSideEffectsV1Beta1 := admissionv1beta1.SideEffectClassNone
+	noSideEffectsV1 := admissionv1.SideEffectClassNone
+	webhookPathV1 := "/failing"
+
+	env.WebhookInstallOptions = WebhookInstallOptions{
+		ValidatingWebhooks: []runtime.Object{
+			&admissionv1beta1.ValidatingWebhookConfiguration{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "deployment-validation-webhook-config",
+				},
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ValidatingWebhookConfiguration",
+					APIVersion: "admissionregistration.k8s.io/v1beta1",
+				},
+				Webhooks: []admissionv1beta1.ValidatingWebhook{
+					{
+						Name: "deployment-validation.kubebuilder.io",
+						Rules: []admissionv1beta1.RuleWithOperations{
+							{
+								Operations: []admissionv1beta1.OperationType{"CREATE", "UPDATE"},
+								Rule: admissionv1beta1.Rule{
+									APIGroups:   []string{"apps"},
+									APIVersions: []string{"v1"},
+									Resources:   []string{"deployments"},
+									Scope:       &namespacedScopeV1Beta1,
+								},
+							},
+						},
+						FailurePolicy: &failedTypeV1Beta1,
+						MatchPolicy:   &equivalentTypeV1Beta1,
+						SideEffects:   &noSideEffectsV1Beta1,
+						ClientConfig: admissionv1beta1.WebhookClientConfig{
+							Service: &admissionv1beta1.ServiceReference{
+								Name:      "deployment-validation-service",
+								Namespace: "default",
+								Path:      &webhookPathV1,
+							},
+						},
+					},
+				},
+			},
+			&admissionv1.ValidatingWebhookConfiguration{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "deployment-validation-webhook-config",
+				},
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ValidatingWebhookConfiguration",
+					APIVersion: "admissionregistration.k8s.io/v1beta1",
+				},
+				Webhooks: []admissionv1.ValidatingWebhook{
+					{
+						Name: "deployment-validation.kubebuilder.io",
+						Rules: []admissionv1.RuleWithOperations{
+							{
+								Operations: []admissionv1.OperationType{"CREATE", "UPDATE"},
+								Rule: admissionv1.Rule{
+									APIGroups:   []string{"apps"},
+									APIVersions: []string{"v1"},
+									Resources:   []string{"deployments"},
+									Scope:       &namespacedScopeV1,
+								},
+							},
+						},
+						FailurePolicy: &failedTypeV1,
+						MatchPolicy:   &equivalentTypeV1,
+						SideEffects:   &noSideEffectsV1,
+						ClientConfig: admissionv1.WebhookClientConfig{
+							Service: &admissionv1.ServiceReference{
+								Name:      "deployment-validation-service",
+								Namespace: "default",
+								Path:      &webhookPathV1,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
 
 var _ = AfterSuite(func(done Done) {
 	Expect(env.Stop()).NotTo(HaveOccurred())
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -779,16 +779,18 @@ var _ = Describe("Test", func() {
 
 	Describe("Start", func() {
 		It("should raise an error on invalid dir when flag is enabled", func(done Done) {
-			env = &Environment{ErrorIfCRDPathMissing: true, CRDDirectoryPaths: []string{invalidDirectory}}
+			env := &Environment{ErrorIfCRDPathMissing: true, CRDDirectoryPaths: []string{invalidDirectory}}
 			_, err := env.Start()
 			Expect(err).To(HaveOccurred())
+			Expect(env.Stop()).To(Succeed())
 			close(done)
 		}, 30)
 
 		It("should not raise an error on invalid dir when flag is disabled", func(done Done) {
-			env = &Environment{ErrorIfCRDPathMissing: false, CRDDirectoryPaths: []string{invalidDirectory}}
+			env := &Environment{ErrorIfCRDPathMissing: false, CRDDirectoryPaths: []string{invalidDirectory}}
 			_, err := env.Start()
 			Expect(err).NotTo(HaveOccurred())
+			Expect(env.Stop()).To(Succeed())
 			close(done)
 		}, 30)
 	})
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -86,6 +86,9 @@ type Environment struct {
 	// CRDInstallOptions are the options for installing CRDs.
 	CRDInstallOptions CRDInstallOptions
 
+	// CRDInstallOptions are the options for installing webhooks.
+	WebhookInstallOptions WebhookInstallOptions
+
 	// ErrorIfCRDPathMissing provides an interface for the underlying
 	// CRDInstallOptions.ErrorIfPathMissing. It prevents silent failures
 	// for missing CRD paths.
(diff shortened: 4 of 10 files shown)


alenkacz:
/assign DirectXMan12 gerred

/hold

@DirectXMan12 I want to add you as co-author, what mail should I use? :-)

alenkacz:
strange, the test failure seems like it should be unrelated but it fails consistently so I'll check it out...

pkg/cache/cache_test.go
```diff
@@ -96,6 +96,14 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			Expect(cfg).NotTo(BeNil())
 
 			By("creating three pods")
+			cl, err := client.New(cfg, client.Options{})
```
alenkacz: this is because of https://github.com/kubernetes-sigs/controller-runtime/pull/787#issuecomment-581039898

pkg/envtest/server.go
```diff
@@ -82,7 +82,8 @@ var DefaultKubeAPIServerFlags = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--admission-control=AlwaysAdmit",
+	"--enable-admission-plugins=ValidatingAdmissionWebhook",
+	"--disable-admission-plugins=ServiceAccount",
```
alenkacz: This is because of https://github.com/kubernetes-sigs/controller-runtime/pull/787#issuecomment-581039898

pkg/envtest/envtest_suite_test.go
```diff
@@ -35,12 +37,61 @@ var env *Environment
 var _ = BeforeSuite(func(done Done) {
 	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
 	env = &Environment{}
+	initializeWebhookInEnvironment()
```
DirectXMan12: we should either have a note here ("needs to be done here because it has to be done during env setup") or call the relevant install function separately during the test that needs it.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,355 @@
...
+	"path/filepath"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	admissionreg "k8s.io/api/admissionregistration/v1beta1"
```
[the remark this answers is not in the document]
alenkacz: what was that issue? I was talking with joe about proper support of v1 webhooks, I'll be creating an issue for that. But I assume this is something more subtle... :-)

pkg/envtest/envtest_suite_test.go
alenkacz: my reasoning for having this here is that it tests the envtest initialization as a sideproduct (and the install function). If I did it in the test itself, I would not test that part. I'll put a not here

pkg/envtest/envtest_test.go
```diff
@@ -777,14 +777,14 @@ var _ = Describe("Test", func() {
...
-			env = &Environment{ErrorIfCRDPathMissing: true, CRDDirectoryPaths: []string{invalidDirectory}}
+			env := &Environment{ErrorIfCRDPathMissing: true, CRDDirectoryPaths: []string{invalidDirectory}}
 			_, err := env.Start()
 			Expect(err).To(HaveOccurred())
 			close(done)
 		}, 30)
 
 		It("should not raise an error on invalid dir when flag is disabled", func(done Done) {
-			env = &Environment{ErrorIfCRDPathMissing: false, CRDDirectoryPaths: []string{invalidDirectory}}
+			env := &Environment{ErrorIfCRDPathMissing: false, CRDDirectoryPaths: []string{invalidDirectory}}
```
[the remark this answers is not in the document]
alenkacz: oh yeah... I had to change this btw. because it was overriding my env and making tests flaky (like tests in webhook.go that also rely on that env)

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,355 @@
...
+	LocalServingPort int
+
+	// LocalServingCertDir is the allocated directory for serving certificates.
+	// it will be automatically populated by the local temp dir
+	LocalServingCertDir string
+}
+
+// WebhooksExists returns true if the path to webhook dir is specified or webhook definition provided
+// this does not mean that actual webhook will be installed in case someone specifies a directory with no webhook definition
+func (o WebhookInstallOptions) WebhooksExists() bool {
```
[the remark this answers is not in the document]
alenkacz: ha this was actually unused code so removed it instead :)

alenkacz:
@vincepri @DirectXMan12 should be ready for re-review

pkg/envtest/webhook.go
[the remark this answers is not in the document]
vincepri: I was more suggesting we start with webhooks/v1 instead of v1beta1. If we don't, we might have to add support later in the future.

We had the same issue for `envtest` CRD, and to be backward compatible we have to handle both versions, which makes the whole code a little tricky. https://github.com/kubernetes-sigs/controller-runtime/pull/752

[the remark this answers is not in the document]
alenkacz: oh I see, yeah that makes sense to me if @DirectXMan12 is OK with that...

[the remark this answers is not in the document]
gerred: i agree with this. if there's an interface here too, can we introduce a helper that uses this interface? if not....ok then.

basically I want to abstract the version from the object as much as possible and generally select the client's storage version (as we would with CRDs).

generally :+1: on latest version selection from our level of abstraction, and binding CR release versions to certain core GVK versions.

pkg/envtest/server.go
```diff
@@ -255,6 +259,13 @@ func (te *Environment) Start() (*rest.Config, error) {
 	te.CRDInstallOptions.ErrorIfPathMissing = te.ErrorIfCRDPathMissing
 	crds, err := InstallCRDs(te.Config, te.CRDInstallOptions)
 	te.CRDs = crds
+	if err != nil {
```
gerred: small nit, swap line 261 and 262-264. even if this doesn't break now, it could break in the futrue based on the te struct changing it's guarantees.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+	for _, hookSet := range o.ValidatingWebhooks {
+		for i := range hookSet.Webhooks {
+			cfg := &hookSet.Webhooks[i].ClientConfig
+			cfg.CABundle = caData
+			var path string
+			if cfg.Service != nil && cfg.Service.Path != nil {
+				path = *cfg.Service.Path
+			}
+			cfg.Service = nil // we cannot use service in integration tests since we're running controller outside cluster
```
gerred: with this the comment, is the intent here to generate a panic if it's used incorrectly? what happens in existing tests if this is not nil but is invalid (and if invalid, are we missing certain other validations?)

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to set up webhook CA: %v", err)
+	}
+
+	hookCert, err := hookCA.NewServingCert()
+	if err != nil {
+		return nil, fmt.Errorf("unable to set up webhook serving certs: %v", err)
+	}
+
+	localServingCertsDir, err := ioutil.TempDir("", "envtest-serving-certs-")
```
gerred: silly, but can we use os.Getwd() here instead of ""? It makes it immediately clear, and more specifically bubbles up permissions or other disk issues in testing environments. been bitten by this in the past. very unlikely flake, but it's clear and it hedges against us not understanding why that's happening (and getting a different error later when we try to write files)

pkg/envtest/webhook.go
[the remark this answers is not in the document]
DirectXMan12: I think probably the safest method here is some sort of helper that can deal with v1 and v1beta1 webhooks like we do with CRDs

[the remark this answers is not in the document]
DirectXMan12: since we're prob stuck with v1beta1 for at least a bit longer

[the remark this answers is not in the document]
DirectXMan12: (basically should be able to copy the code from the CRD handling logic, more-or-less)

pkg/envtest/webhook.go
DirectXMan12: the intent here is that we swap out service for raw address because we don't have an actually standard kube service network.  We want to you to be able to use your standard config though (which probably has a service set, b/c that's how you deploy in prod), so we just blank it out here.

DirectXMan12: I.e. you specify

```yaml
clientConfig:
  service:
    path: /foo
    name: somesvc
    namespace: somens
```

and we effectively convert internally to

```yaml
clientConfig:
  url: https://localhost:<some-random-port>/foo
```

so we can use it.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,358 @@
...
+	if err := WaitForWebhooks(config, o.ValidatingWebhooks, o.MutatingWebhooks, *o); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WaitForWebhooks waits for the Webhooks to be available through API server
+func WaitForWebhooks(config *rest.Config,
+	validatingHooks []*admissionreg.ValidatingWebhookConfiguration,
```
mengqiy: Not excited about this API. It only handles v1beta1 webhooks.
It will be a breaking change later, if we want to support v1.
It will be great if it takes a list of objects and then handle them differently depends on the version and kind.

pkg/internal/testing/integration/internal/apiserver.go
```diff
@@ -9,7 +9,8 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
```
mengqiy: Since we have `populateAPIServerCerts`, we can cleanup code related to apiserver serving on insecure port.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,358 @@
...
+	DirectoryPaths []string
+
+	// MutatingWebhooks is a list of MutatingWebhookConfigurations to install
+	MutatingWebhooks []*admissionreg.MutatingWebhookConfiguration
+
+	// ValidatingWebhooks is a list of ValidatingWebhookConfigurations to install
+	ValidatingWebhooks []*admissionreg.ValidatingWebhookConfiguration
+
+	// ErrorIfPathMissing will cause an error if a DirectoryPath does not exist
+	ErrorIfPathMissing bool
```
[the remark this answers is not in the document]
alenkacz: makes sense to me, I was just trying to be consistent

pkg/internal/testing/integration/internal/apiserver.go
```diff
@@ -9,7 +9,8 @@ var APIServerDefaultArgs = []string{
...
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--admission-control=AlwaysAdmit",
+	"--enable-admission-plugins=ValidatingAdmissionWebhook",
+	"--disable-admission-plugins=ServiceAccount",
```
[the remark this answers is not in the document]
alenkacz: should we though? this is what people start their apiserver envtest with. whatever we disable is disabled for them. it then leads to surprises like calling create in tests on non-existing ns and not getting an error. someone was confused by that recently on slack, I can try to find that thread if you want

pkg/envtest/webhook.go
alenkacz: makes total sense, thanks

pkg/envtest/webhook.go
alenkacz: yep, I made the comment more verbose to make this more obvious

pkg/internal/testing/integration/internal/apiserver.go
alenkacz: I am not sure we're there yet. And even if we are, I would rather do that in separate PR

pkg/internal/testing/integration/internal/apiserver.go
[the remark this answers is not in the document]
mengqiy: > I think we prob want to move towards having a set of admission controllers closer to the defaults for k8s

It worth pointing out to the users that this is a breaking change for them.
If they don't setup namespaces in their existing test, tests will break when upgrade to this new envtest package.

If we eventually want to enable serviceAccount admission controller, it'd better to release it with enabling namespace admission controller.
Thoughts?

pkg/envtest/envtest_suite_test.go
```diff
@@ -19,8 +19,13 @@ package envtest
 import (
 	"testing"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
```
mengqiy: nit: No newline needed here

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,399 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
```
mengqiy: nit: please group the imports properly. Stdlib and k8s related should be grouped respectively.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,399 @@
...
+	if err := WaitForWebhooks(config, o.ValidatingWebhooks, o.MutatingWebhooks, *o); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WaitForWebhooks waits for the Webhooks to be available through API server
+func WaitForWebhooks(config *rest.Config,
+	validatingWebhooks []runtime.Object,
```
mengqiy: nit: To be consistent with other functions signature, I'd suggest to make `mutatingWebhooks` comes first, then `validatingWebhooks`.

pkg/envtest/webhook_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
```
mengqiy: nit: no newline needed here.

pkg/envtest/webhook_test.go
```diff
@@ -0,0 +1,81 @@
...
+								{
+									Name:  "nginx",
+									Image: "nginx",
+								},
+							},
+						},
+					},
+				},
+			}
+			err = c.Create(context.TODO(), obj)
```
mengqiy: How can we ensure the server is up and ready to accept incoming requests? It may cause flakes here.
`Eventually` may help here.

pkg/envtest/webhook.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to set up webhook CA: %v", err)
+	}
+
+	hookCert, err := hookCA.NewServingCert()
+	if err != nil {
+		return nil, fmt.Errorf("unable to set up webhook serving certs: %v", err)
+	}
+
+	localServingCertsDir, err := ioutil.TempDir("", "envtest-serving-certs-")
```
mengqiy: Have we ever cleaned up this directory?

alenkacz: added cleanup method, great point 👍

pkg/envtest/webhook_test.go
alenkacz: I think at this point it has to be ready, envtest_suite_test before already run, we wait for the webhook configs to be accessible, I think we should be fine here. Correct me if I am missing something :)

pkg/internal/testing/integration/internal/apiserver.go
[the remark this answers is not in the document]
alenkacz: Yeah, it's a breaking change, that's true... but I think we should do it. Maybe we need to move it out of this PR and submit separately and make sure we highlight it in time of release?

alenkacz:
@mengqiy thanks for the review, you can do another pass :)

pkg/envtest/webhook_test.go
mengqiy: Sorry about not being super clear.
I was talking about the webhook server, but not the kube-apiserver.
```
go func() {
	_ = server.Start(stopCh)
}()
```
We spin up the webhook server in a goroutine. The webhook server is not guaranteed to be ready when we try to create the deployment here.

pkg/internal/testing/integration/internal/apiserver.go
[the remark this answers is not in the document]
mengqiy: I saw @vincepri added this to the `v0.5.x` milestone, where we should not do breaking change. We have to leave it in `v0.6.0`.

> Maybe we need to move it out of this PR and submit separately and make sure we highlight it in time of release?

To get this PR merged soon, I'd suggest split the breaking change bits out to a separate PR, which will be merged when we get close to v0.6.0.

pkg/envtest/webhook_test.go
alenkacz: oh, you're right 🙈 added eventually, thanks. Did not realize that

pkg/internal/testing/integration/internal/apiserver.go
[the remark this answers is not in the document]
DirectXMan12: :+1:

pkg/internal/testing/integration/internal/apiserver.go
```diff
@@ -9,7 +9,8 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--admission-control=AlwaysAdmit",
+	"--enable-admission-plugins=MutatingAdmissionWebhook,ValidatingAdmissionWebhook",
+	"--disable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,TaintNodesByCondition,Priority,DefaultTolerationSeconds,DefaultStorageClass,StorageObjectInUseProtection,PersistentVolumeClaimResize,RuntimeClass,ResourceQuota",
```
alenkacz: this is to be backward compatible

mengqiy:
@alenkacz Don't forget to create the followup PR with the breaking change after this PR merges.
/lgtm
/approve

PR #754, opened 2020-01-10
:warning: Bump to Kubernetes 1.17 dependencies

joelanford:
This PR bumps controller-runtime to use Kubernetes 1.17 libraries, which will enable controller/operator developers to take advantage of the latest 1.17 client-go and API updates in their projects.

Due to an upstream change in `k8s.io/api`, there is one breaking change in controller-runtime's `pkg/leaderelection/fake` package, which is used only in `pkg/manager/manager_test.go`, at least internally. Should we deprecate this fake ResourceLock implementation from the public API and plan to move this into an `internal` tree?

Here's the notes and output from `go-apidiff`:

- There's the one aforementioned breaking change in `sigs.k8s.io/controller-runtime`
- As is expected with a version bump to the next major Kubernetes version, there are a raft of breaking API changes to libraries that controller-runtime imports and uses. For users of controller-runtime that also use these libraries, there may need to be code refactorings to address these breakages.

```
$ go-apidiff master k8s-1.17 --compare-imports

sigs.k8s.io/controller-runtime/pkg/leaderelection/fake
  Incompatible changes:
  - (*ResourceLock).Get: changed from func() (*k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord, error) to func() (*k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord, []byte, error)

k8s.io/client-go/kubernetes
  Incompatible changes:
  - Interface.DiscoveryV1beta1: added
  - Interface.FlowcontrolV1alpha1: added

github.com/onsi/ginkgo/config
  Incompatible changes:
  - VERSION: value changed from "1.8.0" to "1.10.1"

k8s.io/apimachinery/pkg/apis/meta/v1
  Incompatible changes:
  - Convert_Slice_string_To_v1_DeletionPropagation: removed

k8s.io/client-go/rest
  Incompatible changes:
  - (*Request).WatchWithSpecificDecoders: removed
  - NewRESTClient: changed from func(*net/url.URL, string, ContentConfig, float32, int, k8s.io/client-go/util/flowcontrol.RateLimiter, *net/http.Client) (*RESTClient, error) to func(*net/url.URL, string, ClientContentConfig, k8s.io/client-go/util/flowcontrol.RateLimiter, *net/http.Client) (*RESTClient, error)
[...]

PR #732, opened 2019-12-17
✨ Update dependencies to Kubernetes v1.16.4

ialidzhikov:
Bump dependencies for Kubernetes 1.16.4.

PR #789, opened 2020-02-05
🏃 Bump go.mod to use versions instead of SHAs.

armandgrillet:
We use SHAs to specify versions of our dependencies, e.g. `k8s.io/utils v0.0.0-20190801114015-581e00157fb1`. Instead, where possible, I have replaced the SHAs with the latest version using the same major version (in our case, `v0.16.6`). The intent is to make the `go.mod` more readable/upgradable.


gerred:
Holding this until we decide what to do with #754 and #732

vincepri:
Closing this now that #795 has merged

/close

Issue #343, opened 2019-02-28
Cannot Get after Create using Client with Custom Resource


michaelkipper:
Hi!

I'm tearing my hair out trying to debug this issue.

```go
func testCreation(c client.Client) error {
	config := getTestCuscoServiceConfig()

	err := c.Create(context.Background(), &config)
	if err != nil {
		return errors.Wrap(err, "Could not create CuscoServiceConfig")
	}

	new_csc := v1.CuscoServiceConfig{}
	err = c.Get(context.Background(), client.ObjectKey{Name: "test-config", Namespace: "cusco"}, &new_csc)
	if err != nil {
		return errors.Wrap(err, "Could not get CuscoServiceConfig")
	}

	return nil
}
```

When run, I get this error:

```
Could not get CuscoServiceConfig: CuscoServiceConfig.cusco.shopify.io \"test-config\" not found"
```

Outside the program, though:

```yaml
$ kubectl get cuscoserviceconfigs test-config -n cusco -o yaml
apiVersion: cusco.shopify.io/v1
kind: CuscoServiceConfig
metadata:
  creationTimestamp: "2019-02-28T15:35:09Z"
  generation: 1
  name: test-config
  namespace: cusco
  resourceVersion: "21614"
  selfLink: /apis/cusco.shopify.io/v1/namespaces/cusco/cuscoserviceconfigs/test-config
  uid: 6e8bb6e3-3b6e-11e9-932e-02423b5580eb
spec:
  init-container:
    enabled: false
    percentage: 75
  istio-version: 1.1.0
  proxy:
    enabled: true
    percentage: 25
```

Stepping through the code, my theory is that the `Create` call doesn't store the object in the cache, and the `Get` call only looks at the cache.

How should I go about debugging this?
Is this because `CuscoServiceConfig` is a CustomResource?

DirectXMan12:
Yeah, you always write to the API server and read from the cache.  You can't expect to read the object you wrote immediately (and, in fact, there's generally a bit of delay).  What are you trying to accomplish with the immediate `Get`?  You'll get back the created object as part of the call to create.

michaelkipper:
@DirectXMan12: This was a snippet of something that's happening in an integration test that's failing. I would have expected read-after-write to be strongly consistent.

If this is not a guarantee, I can always drop a `WaitForResource` into the test.

DirectXMan12:
how did you get the client for the integration test?

DirectXMan12:
[...]

PR #790, opened 2020-02-05
📖: Add note about split client cache inconsistency

theishshah:
Adds documentation to cover discussion seen in #343


gerred:
Thanks for your contribution!

/ok-to-test
/lgtm
/approve

PR #794, opened 2020-02-10
✨ Create a supported way of setting a cache on source.Kind

### LINKED ISSUE #650: Define a supported way of setting a cache on source.Kind
In order to be able to write a controller that has informers in multiple clusters, one has to inject a cache into `source.Kind`. Options to do this include:

* Just call `kind.InjectCache`, it has a [nilcheck](https://github.com/kubernetes-sigs/controller-runtime/blob/ecdbe54c7cc55f9dfcc6cb56fc1622d182aa2219/pkg/source/source.go#L109-L111) that will prevent the `injectCache` call from the controller to overwrite the cache. However this behavior doesn't have tests and may go away in the future
* Write a wrapper around `kind.Source` that implements `InjectCache` with the nilcheck
* Write a wrapper around `kind.Source` that does not implement `inject.Cache`

We should agree on one option, declare it the supported way to build multicluster controllers and probably document it.

jkbschmid:
Create a wrapper for Kind without `InjectCache` to create a supported way of setting a cache on source.Kind. This can used, e.g., for watching resources on another cluster.

fixes #650

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -55,17 +55,18 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
-// CreateSourceWithFixedCache creates a Source without InjectCache, so that it is assured that the given cache is used
-// and not overwritten
-func CreateSourceWithFixedCache(object runtime.Object, cache cache.Cache) Source {
-	return &sourceWithFixedCache{kind: Kind{Type: object, cache: cache}}
+// NewKindWithFixedCache creates a Source without InjectCache, so that it is assured that the given cache is used
+// and not overwritten. It can be used to watch objects in a different cluster by passing the cache
+// from that other cluster
+func NewKindWithFixedCache(object runtime.Object, cache cache.Cache) Source {
+	return &kindWithFixedCache{kind: Kind{Type: object, cache: cache}}
 }
 
-type sourceWithFixedCache struct {
+type kindWithFixedCache struct {
 	kind Kind
 }
 
-func (ks *sourceWithFixedCache) Start(handler handler.EventHandler, queue workqueue.RateLimitingInterface,
+func (ks *kindWithFixedCache) Start(handler handler.EventHandler, queue workqueue.RateLimitingInterface,
 	prct ...predicate.Predicate) error {
 	return ks.kind.Start(handler, queue, prct...)
 }
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -55,18 +55,18 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
-// NewKindWithFixedCache creates a Source without InjectCache, so that it is assured that the given cache is used
+// NewKindWithCache creates a Source without InjectCache, so that it is assured that the given cache is used
 // and not overwritten. It can be used to watch objects in a different cluster by passing the cache
 // from that other cluster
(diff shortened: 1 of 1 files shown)


pkg/source/source.go
```diff
@@ -55,6 +55,22 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
+// NewKindWithFixedCache creates a Source without InjectCache, so that it is assured that the given cache is used
+// and not overwritten. It can be used to watch objects in a different cluster by passing the cache
+// from that other cluster
+func NewKindWithFixedCache(object runtime.Object, cache cache.Cache) Source {
```
vincepri:
```suggestion
func NewKindWithCache(object runtime.Object, cache cache.Cache) Source {
```

pkg/source/source.go
```diff
@@ -55,6 +55,22 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
+// NewKindWithFixedCache creates a Source without InjectCache, so that it is assured that the given cache is used
```
vincepri:
```suggestion
// NewKindWithCache creates a Source without relying on InjectCache, so that it is assured that the given cache is used
```

vincepri:
Minor comments, otherwise lgtm

gerred:
agreed with @alvaroaleman - if all a constructor doing here is instantiating a struct, I don't think we should use one. Why can't this be a public struct, or why can't we just change Cache to be exported as well?

https://github.com/kubernetes-sigs/controller-runtime/blob/f515d1e05bc071898beba7fd5da7a9ea10cf6263/pkg/source/source.go#L80

Furthermore, if that's the direction we're going, we should probably embed the struct into kindWithCache rather than creating an entirely new struct. The only difference between the two is the extra bytes for the reference.

Simplifying this enables us to create a DefaultKind that in effect preserves the existing functionality, rather than new structs and new constructors.

gerred:
I just read the linked thread and while I think we should look for a cleaner long term solution, I understand the problem and the reason we have this solution.

/hold cancel
/approve

c0d1ngm0nk3y:
> Why can't this be a public struct, or why can't we just change Cache to be exported as well

But then we would end up with a struct that cannot be used until the cache is set. You would have to document this. With the constructor, it is safe to use and you cannot use it in a wrong way.

alvaroaleman:
> But then we would end up with a struct that cannot be used until the cache is set. You would have to document this. With the constructor, it is safe to use and you cannot use it in a wrong way.

I agree with that. Maybe we can add constructors for all sources and mark the direct usage of the structs as deprecated, so this is consistent.

pkg/source/source.go
```diff
@@ -55,6 +55,22 @@ type Source interface {
 	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
 }
 
+// NewKindWithCache creates a Source without InjectCache, so that it is assured that the given cache is used
```
alvaroaleman: Nit: the first sentence of this godoc is basically for ppl who know how its implemented and not explaining the purpose which is far more important. It would be nice to just say sth like `NewKindWithCache is used to create a Wach for a different cluster`, then go on with the implementation details like "this is achieved by intentionally not implementing InjectCache" etc. Would be great if someone could create a followup for that.

/lgtm

alvaroaleman:
/retest

We should probably throw more CPU at this job and hope that stabilizes it.

PR #796, opened 2020-02-13
🐛 Fix informer cache creating pointers to pointers 

### LINKED ISSUE #656: Unable to List Objects with pointer slices

alvaroaleman:
Makes the cache cope with list types that have a slice of pointers as `Items` field.

Fixes #656
Supersedes and thereby closes #667

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -93,7 +93,11 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
 		}
 		// http://knowyourmeme.com/memes/this-is-fine
 		elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
-		cacheTypeValue := reflect.Zero(reflect.PtrTo(elemType))
+		if elemType.Kind() != reflect.Ptr {
+			elemType = reflect.PtrTo(elemType)
+		}
+
+		cacheTypeValue := reflect.Zero(elemType)
 		var ok bool
 		cacheTypeObj, ok = cacheTypeValue.Interface().(runtime.Object)
 		if !ok {
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -70,26 +70,48 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out runt
 
 // List implements Reader
 func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...client.ListOption) error {
-	gvk, err := apiutil.GVKForObject(out, ip.Scheme)
+
+	gvk, cacheTypeObj, err := ip.objectTypeForListObject(out)
 	if err != nil {
 		return err
 	}
 
+	started, cache, err := ip.InformersMap.Get(*gvk, cacheTypeObj)
+	if err != nil {
+		return err
+	}
+
+	if !started {
+		return &ErrCacheNotStarted{}
+	}
+
+	return cache.Reader.List(ctx, out, opts...)
+}
+
+// objectTypeForListObject tries to find the runtime.Object and associated GVK
+// for a single object corresponding to the passed-in list type. We need them
+// because they are used as cache map key.
+func (ip *informerCache) objectTypeForListObject(list runtime.Object) (*schema.GroupVersionKind, runtime.Object, error) {
+	gvk, err := apiutil.GVKForObject(list, ip.Scheme)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	if !strings.HasSuffix(gvk.Kind, "List") {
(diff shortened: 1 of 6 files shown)


pkg/controller/controller_suite_test.go
```diff
@@ -62,3 +76,51 @@ var _ = AfterSuite(func() {
...
+
+// UnconventionalListType is used to test CRDs with List types that
+// have a slice of pointers rather than a slice of literals.
+type UnconventionalListType struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              string `json:"spec,omitempty"`
+}
+
+// DeepCopyObject implements runtime.Object
```
alvaroaleman: I handwrote these mostly because I was too lazy to figure out how to plug deepcopy-gen into a  hackscript and have it verified by CI. Also it may not be able to cope with a folder that contains multiple packages (`controller` and `controller_test`). Since the correctness of the DeepCopies is not really relevant for the test, thats IMHO good enough

DewaldV:
Awesome job on the test @alvaroaleman I got about 50% of the way there when I attempted it last year but had brought a lot more code along so this looks much cleaner than my test implementation. :smile:

alvaroaleman:
> Awesome job on the test @alvaroaleman! I got about 50% of the way there when I attempted it last year but had brought a lot more code along so this looks much cleaner than my test implementation. smile


Yeah it actually turned out to be a lot more involved than anticipated, initially I thought I could directly test the `informerCache` but that isn't really possible as it embedds a `*InformersMap`, so mocking that would require some more refactorings. So I ended up just writing an integration test :upside_down_face:

alvaroaleman:
@vincepri Mind having a look?

pkg/controller/controller_suite_test.go
vincepri: Can you add a comment with the above? :)

pkg/cache/informer_cache.go
```diff
@@ -93,7 +93,11 @@ func (ip *informerCache) List(ctx context.Context, out runtime.Object, opts ...c
 		}
 		// http://knowyourmeme.com/memes/this-is-fine
 		elemType := reflect.Indirect(reflect.ValueOf(itemsPtr)).Type().Elem()
```
vincepri: For posterity, can we also add a comment here explaining what we're trying to do and why? I checked out the code and understood we're pretty much trying to figure out the type of the list passed in, but that might be a lot to ask for folks not familiar with the codebase.

Let's also move the logic to a local function and add some unit tests for it.

alvaroaleman: Done, PTAL. I had to move the `UnconventionalListType` into a different package, because its now used by both unit and integration tests and would otherwise have caused an import cycle.

Issue #572, opened 2019-08-21
Cannot specify per-watch predicates on builders


magicsong:
however, it had no problem in v1

DirectXMan12:
need to change the builder interface ever so slightly to allow for per-watch predicates.

mszostok:
Hi,

I can provide a PR for that, are you interested?

DirectXMan12:
/lifecycle frozen

waiting to make a set of breaking changes

PR #602, opened 2019-09-12
✨Add predicates as variadic args for Owns, For, and Watches func

mszostok:
**Description**

- This PR adds options for defining the `Predicate`. Watches, Owns, For, and Watches 
- Add tests coverage for implemented logic using ginko & gomega 

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/572

PR #799, opened 2020-02-14
:warning: Add predicates as variadic args for Owns, For, and Watches func

mszostok:
**Description**

- This PR adds options for defining the `Predicate`. Watches, Owns, For, and Watches
- Add tests coverage for implemented logic using ginko & gomega 

Implemented with Functional Options pattern as described here: https://github.com/kubernetes-sigs/controller-runtime/pull/602#discussion_r377337991

I decided to go with `Input` struct for each method to have separation and make those methods independent of each other. 

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/572

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -37,15 +37,15 @@ var getGvk = apiutil.GVKForObject
 
 // Builder builds a Controller.
 type Builder struct {
-	apiType        runtime.Object
-	mgr            manager.Manager
-	predicates     []predicate.Predicate
-	managedObjects []runtime.Object
-	watchRequest   []watchRequest
-	config         *rest.Config
-	ctrl           controller.Controller
-	ctrlOptions    controller.Options
-	name           string
+	forInput         ForInput
+	ownsInput        []OwnsInput
+	watchesInput     []WatchesInput
+	mgr              manager.Manager
+	globalPredicates []predicate.Predicate
+	config           *rest.Config
+	ctrl             controller.Controller
+	ctrlOptions      controller.Options
+	name             string
 }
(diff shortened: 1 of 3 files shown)


DirectXMan12:
FYI, this'll prob mean some cherry-picking for the next v0.5.x release

vincepri:
did we say that varargs are breaking changes and need to go into the next release?

DirectXMan12:
Yeah -- it changes the type signature.

For instance, if you had `func callWithPod(cb func(runtime.Object) ctrl.Builder)`, you could no longer pass `For` or `Owns` to that function.

vincepri:
Sounds good, I'll put a hold on this once we want to open master for v0.6. Given that there are a few PRs in flight for v0.5.x, I'm thinking we can wait for those to be merged, cut v0.5.1 and then open master to v0.6.

/hold
until breaking change on master are open

pkg/builder/controller.go
```diff
@@ -63,32 +63,62 @@ func (blder *Builder) ForType(apiType runtime.Object) *Builder {
 	return blder.For(apiType)
 }
 
+// ForInput represents the information set by For method.
+type ForInput struct {
```
seh: Can we get away without exporting this type?
Likewise for `OwnsInput` and `WatchesInput`?

DirectXMan12: If we want to allow custom implementations of options (which is nice), this has to be public

seh: Ah, so that others can implement `ForOption.ApplyToFor` and company. I see.

PR #802, opened 2020-02-17
✨enable webhook server set-up mTLS service to verify client's certificate

answer1991:
Enable webhook server set-up mTLS service to verify client's certificate according to the given CAName

Not sure how to add UT, this package seems no UT for now?

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -19,7 +19,9 @@ package webhook
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -56,6 +58,10 @@ type Server struct {
 
 	// CertName is the server key name. Defaults to tls.key.
 	KeyName string
+
+	// ClientCAName is the CA certificate name which server used to verify remote(client)'s certificate.
+	// Defaults to "", which means server does not verify client's certificate.
+	ClientCAName string
 
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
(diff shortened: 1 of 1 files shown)


pkg/webhook/server.go
```diff
@@ -57,6 +59,10 @@ type Server struct {
 	// CertName is the server key name. Defaults to tls.key.
 	KeyName string
 
+	// CAName is the CA certificate name which server used to verify remote(client)'s certificate.
+	// Defaults to "", which means server does not verify client's certificate.
+	CAName string
```
mengqiy: Name is a little confusing.
Probably `ClientCAName` is better.

pkg/webhook/server.go
```diff
@@ -168,6 +174,23 @@ func (s *Server) Start(stop <-chan struct{}) error {
 		GetCertificate: certWatcher.GetCertificate,
 	}
 
+	// load CA to verify client certificate
+	if len(s.CAName) > 0 {
```
awgreene: Is there a reason you used `len(s.CAName)` over `s.CAName != ""`? Is there some optimization that I'm unaware of?

answer1991: No, just personal coding habit.. Fixed, thanks!

pkg/webhook/server.go
answer1991: done, thanks!

pkg/webhook/server.go
```diff
@@ -57,6 +59,10 @@ type Server struct {
 	// CertName is the server key name. Defaults to tls.key.
```
mengqiy: Just realize there is a typo.
It should be `KeyName`. Can you please also change that?

answer1991: done

djzager:
Changes LGTM @answer1991 . I wonder if it would be reasonable to add this to the godoc in https://github.com/kubernetes-sigs/controller-runtime/blob/39ee87b530648b5fd7c3943c0d0cbe9d8f528894/pkg/webhook/example_test.go#L52 to provide a meaningful example of starting the server with mTLS.

mengqiy:
/lgtm
/approve
@answer1991 @djzager Feel free to add mTLS example in godoc example in a followup PR.

answer1991:
sure, I will finish it ASAP

PR #805, opened 2020-02-20
✨manager graceful exit with release lease

answer1991:
Best Practices of exit controllers

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -44,9 +44,10 @@ import (
 
 const (
 	// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
-	defaultLeaseDuration = 15 * time.Second
-	defaultRenewDeadline = 10 * time.Second
-	defaultRetryPeriod   = 2 * time.Second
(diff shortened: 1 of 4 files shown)


pkg/manager/internal.go
```diff
@@ -44,9 +44,10 @@ import (
 
 const (
 	// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
-	defaultLeaseDuration = 15 * time.Second
-	defaultRenewDeadline = 10 * time.Second
-	defaultRetryPeriod   = 2 * time.Second
+	defaultLeaseDuration   = 15 * time.Second
+	defaultRenewDeadline   = 10 * time.Second
+	defaultRetryPeriod     = 2 * time.Second
+	defaultReleaseOnCancel = false
```
answer1991: I suggest set this `defaultReleaseOnCancel ` to be true, but which will break now behavior.

vincepri:
/close

We're taking a step back in https://github.com/kubernetes-sigs/controller-runtime/issues/764 and trying to come up with a design around stop/close behaviors.

Closing this for now until we re-evaluate.

Issue #518, opened 2019-07-10
Provide a ComponentConfig to tweak the Manager


johnsonj:
Controllers would like to tweak the settings of the manager (eg, change logging verbosity) through a ComponentConfig. 

Additionally, individual controllers may want to expose settings that relate to their usecase (eg, a whitelist of git repos an operator can pull from). I'm not sure if that's a usecase worth splitting off but we would like to use this for [addon-operators](https://github.com/kubernetes-sigs/addon-operators)

/cc @stealthybox @shawn-hurley @DirectXMan12

stealthybox:
Just to add context, ComponentConfig implies using API Machinery for the types used in a config file. (Not just unmarshalling a Yaml file into a Go struct)

More context in the component-base KEP:
https://github.com/luxas/enhancements/blob/f3005915981cf3e590dc0744ad0256cdf76970aa/keps/sig-cluster-lifecycle/0032-create-a-k8s-io-component-repo.md#part-1-componentconfig

stealthybox:
I would imagine we would have some fields to configure the manager's SharedInformer?

DirectXMan12:
Yeah, and the various other bits of the manager that can easily be configured with constant values (like metrics and webhook serving ports)

christopherhein:
@DirectXMan12 @johnsonj @stealthybox I'm interested in looking into what this might take, has anyone made any progress or have thoughts about this?

Loosely looking through the codebase it seems like we could make it so that `controller-runtime.Manager` could support a separate `New` function specifically for pulling in ComponentConfig type resources. We could make this and interface for which the ComponentConfig stuct would have to adhere too.

I'm thinking something along the lines of:

```go
// ManagerConfiguration defines what the ComponentConfig object for ControllerRuntime needs to support
type ManagerConfiguration interface {
	GetSyncPeriod() *time.Duration

	GetLeaderElection() bool
	GetLeaderElectionNamespace() string
	GetLeaderElectionID() string

	GetLeaseDuration() *time.Duration
	GetRenewDeadline() *time.Duration
	GetRetryPeriod() *time.Duration

	GetNamespace() string
	GetMetricsBindAddress() string
	GetHealthProbeBindAddress() string

	GetReadinessEndpointName() string
	GetLivenessEndpointName() string

	GetPort() int
	GetHost() string

	GetCertDir() string
}

func NewFromComponentConfig(config *rest.Config, managerconfig ManagerConfiguration) (Manager, error) {
[...]

PR #809, opened 2020-02-20
✨📖 Adding designs folder and sample template

### LINKED ISSUE #807: Proposals folder and template
/assign

christopherhein:
Per the discussion in the controller-runtime this fixes #807 adding so that I can add the design document for #518 

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Documentation:
 - [Creating a manager](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/manager#example-New)
 - [Creating a controller](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/controller#example-New)
 - [Examples](https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples)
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Let's not do the full KEP template.  I'd like to see something closer to the KB template (in fact, I kinda just think we should copy it).  IMO, at least for us, KEPs are too complicated, intimidating, and paperwork-heavy.

DirectXMan12:
Might be worth having a "is this a breaking change" section for controller-runtime too.

What I'm looking for when I read a proposal is:

- What does it do?
- Is it breaking (and did we consider making it non-breaking), and how does it evolve in the future (additional options, dependency changes, etc, if relevant)?
- How do we teach this (see Rust RFCs) and how does an example look?

That's pretty much it in terms of structure.  At least in my opinion, anything more than that just gets in the way of people actually doing things at this scale.

Key point from the KB design template:

> we want enough information to evaluate the design, but not so much that you're
> annoyed by the overall design process and decide to bake cookies instead.

We should copy the designs README from KB too (adjusting slightly as necessary).

PR #234, opened 2018-12-04
:sparkles: add metrics for webhooks

mengqiy:
Add 2 metrics for webhook:
- Number of total requests
- Latency of processing admission requests

PR #811, opened 2020-02-20
📖 document exposed metrics

### LINKED ISSUE #262: Document what metrics are registered in godoc
From https://github.com/kubernetes-sigs/controller-runtime/pull/234#issuecomment-448248502

djzager:
Update documentation for the metrics package to enumerate all exposed
metrics with an explanation of what they capture.

Fixes #262

### DIFF
--- a/pkg/metrics/workqueue.go
+++ b/pkg/metrics/workqueue.go
@@ -19,106 +19,112 @@ package metrics
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/client-go/util/workqueue"
-	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
-
-var log = logf.RuntimeLog.WithName("metrics")
 
 // This file is copied and adapted from k8s.io/kubernetes/pkg/util/workqueue/prometheus
 // which registers metrics to the default prometheus Registry. We require very
 // similar functionality, but must register metrics to a different Registry.
 
-func init() {
-	workqueue.SetProvider(workqueueMetricsProvider{})
-}
+// Metrics subsystem and all keys used by the workqueue.
+const (
+	WorkQueueSubsystem         = "workqueue"
+	DepthKey                   = "depth"
+	AddsKey                    = "adds_total"
+	QueueLatencyKey            = "queue_duration_seconds"
+	WorkDurationKey            = "work_duration_seconds"
+	UnfinishedWorkKey          = "unfinished_work_seconds"
+	LongestRunningProcessorKey = "longest_running_processor_seconds"
+	RetriesKey                 = "retries_total"
+)
 
-func registerWorkqueueMetric(c prometheus.Collector, name, queue string) {
-	if err := Registry.Register(c); err != nil {
-		log.Error(err, "failed to register metric", "name", name, "queue", queue)
-	}
+var (
+	depth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: WorkQueueSubsystem,
+		Name:      DepthKey,
+		Help:      "Current depth of workqueue",
+	}, []string{"name"})
+
+	adds = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: WorkQueueSubsystem,
+		Name:      AddsKey,
+		Help:      "Total number of adds handled by workqueue",
+	}, []string{"name"})
+
+	latency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Subsystem: WorkQueueSubsystem,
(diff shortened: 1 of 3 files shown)


djzager:
I see that the [original comment](https://github.com/kubernetes-sigs/controller-runtime/pull/234#issuecomment-448248502) was referring to the `pkg/webhook/internal/metrics` package but it already documents the exposed metrics.

 I could follow the same pattern as [`pkg/webhook/internal/metrics`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/metrics/metrics.go#L26-L28) but that would require exporting a bunch of stuff that is currently not exported.

djzager:
Looking around the project at the other metrics packages (ie. [controller metrics](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics) and [webhook metrics](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics)) I feel that I took the wrong approach in this PR.

What _I would like to do_ would be to export the variables in this package where metrics are defined (making them `var` where they are now hidden inside a function) and add the appropriate comments.

@DirectXMan12 @vincepri WDYT?

vincepri:
That sounds good to me as well

pkg/metrics/workqueue.go
```diff
@@ -19,106 +19,118 @@ package metrics
...
 
-var log = logf.RuntimeLog.WithName("metrics")
-
 // This file is copied and adapted from k8s.io/kubernetes/pkg/util/workqueue/prometheus
 // which registers metrics to the default prometheus Registry. We require very
 // similar functionality, but must register metrics to a different Registry.
 
+var (
+	// WorkQueueSubsystem defines the subsystem for the workqueue metrics.
+	WorkQueueSubsystem = "workqueue"
```
djzager: went ahead and followed what was being done in [kubernetes metrics](https://github.com/kubernetes/kubernetes/blob/c7d84ac77e3064c80c7f9c9e11a017b0bf78acd9/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go)

djzager:
Went ahead and left it as 2 commits to make it easier to go back to the original if necessary. If this works I'll rebase and squash.

DirectXMan12:
Hey, so I'm sorry to jump in a bit later here, but: 

> What I would like to do would be to export the variables in this package where metrics are defined (making them var where they are now hidden inside a function) and add the appropriate comments.

I'm not a huge fan of this approach.  Namely, it means we have out metrics compatibility boundary not at the metrics we expose, but via the *exact* method we expose them by.  For instance, under this model, switching to OpenTelemetry is *necessarily* a breaking change, even if we expose the same metrics in the same format with the same names, since the Go types change.  Considering that the Go types are mostly an implementation detail, that seems suboptimal.

As I see it, there are two ways around this: move the actual variables to an internal package so that users can't depend on them, or make the variables private and document as a block.

Thoughts?

/hold

(while we're discussing)

djzager:
> Due to the way this is currently structured, it means we have out metrics compatibility boundary not at the metrics we expose, but via the _exact_ method we expose them by. For instance, under this model, switching to OpenTelemetry is _necessarily_ a breaking change, even if we expose the same metrics in the same format with the same names, since the Go types change. Considering that the Go types are mostly an implementation detail, that seems suboptimal.

My idea of making unexported `var`s exported made me uneasy, but I didn't really understand why. Thank you for the explanation.

I have updated the metrics package to expose the subsystems and specific keys used without exporting the metrics themselves.

I would be tempted to make similar changes in https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/metrics/metrics.go and https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/metrics/metrics.go but I suspect making exported things unexported would also be a breaking change. It's also possible that `pkg/internal/controller/metrics` and `pkg/webhook/internal/metrics` are fundamentally different than `pkg/metrics` but I'm not certain. Hoping to gain better understanding of what changes make sense and when.

DirectXMan12:
> It's also possible that pkg/internal/controller/metrics and pkg/webhook/internal/metrics are fundamentally different than pkg/metrics but I'm not certain

Ah, yes, your intuition is correct here :-).  In Go, and package with `/internal/` in the path as visibility limited to the parent package of `/internal` (as well as any subpackages of the parent package).  So anything in `/pkg/webhook/internal/...` can only be accessed by things in `/pkg/webhook/...`.  This is enforced by the compiler's import logic: https://golang.org/doc/go1.4#internalpackages

shawn-hurley:
Should we cancel the hold on this?

Issue #41, opened 2018-06-15
pkg/cache methods should implement deadline semantics for context parameter


droot:
All exported methods on client interface takes in `context.Context` parameter, but the methods don't implement the deadline semantics.

DirectXMan12:
/remove-lifecycle rotten

IIRC, this is still a problem for the unstructured client

DirectXMan12:
will be fixed by #580

vincepri:
/help
/assign @alexeldeib 

/priority awaiting-more-evidence 
might or might not be fixed by #663

alexeldeib:
#663 fixes the cache semantics, but the unstructured client doesn't plumb through context. We can either track that here or in a separate issue, since there's already so many around context. I can start working on plumbing the context for the unstructured client.

alexeldeib:
#663 and #812 both merged, so:

/close

PR #812, opened 2020-02-21
🐛 respect context in unstructured client

alexeldeib:
ref: https://github.com/kubernetes-sigs/controller-runtime/issues/41#issuecomment-589237195

aligns the unstructured client with the typed client, sharing a cache between them for object metadata. plumbs context through to the restclient for unstructured client in doing so.

@shawn-hurley mentioned there were previously some issues with this, but as-is the implementation passes all current tests?

also, idk how hacky people consider this 😄 feels like I probably need to fix up some GVKs somewhere in the lists...

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -827,6 +827,51 @@ var _ = Describe("Client", func() {
 				close(done)
 			})
 
+			It("should update status and preserve type information", func(done Done) {
+				cl, err := client.New(cfg, client.Options{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(cl).NotTo(BeNil())
+
+				By("initially creating a Deployment")
+				dep, err := clientset.AppsV1().Deployments(ns).Create(dep)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("updating the status of Deployment")
+				u := &unstructured.Unstructured{}
+				dep.Status.Replicas = 1
+				Expect(scheme.Convert(dep, u, nil)).To(Succeed())
+				err = cl.Status().Update(context.TODO(), u)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("validating updated Deployment has type information")
+				Expect(u.GroupVersionKind()).To(Equal(depGvk))
+
+				close(done)
+			})
+
+			It("should patch status and preserve type information", func(done Done) {
+				cl, err := client.New(cfg, client.Options{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(cl).NotTo(BeNil())
+
+				By("initially creating a Deployment")
+				dep, err := clientset.AppsV1().Deployments(ns).Create(dep)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("patching the status of Deployment")
+				u := &unstructured.Unstructured{}
+				depPatch := client.MergeFrom(dep.DeepCopy())
+				dep.Status.Replicas = 1
+				Expect(scheme.Convert(dep, u, nil)).To(Succeed())
+				err = cl.Status().Patch(context.TODO(), u, depPatch)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("validating updated Deployment has type information")
+				Expect(u.GroupVersionKind()).To(Equal(depGvk))
+
+				close(done)
+			})
+
 			It("should not update spec of an existing object", func(done Done) {
 				cl, err := client.New(cfg, client.Options{})
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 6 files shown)


alexeldeib:
cc @alvaroaleman for some additional 👀

pkg/client/unstructured_client.go
```diff
@@ -126,105 +161,119 @@ func (uc *unstructuredClient) Patch(_ context.Context, obj runtime.Object, patch
...
 		return err
 	}
-	u.Object = i.Object
-	return nil
+
+	result := r.Get().
+		NamespaceIfScoped(key.Namespace, r.isNamespaced()).
+		Resource(r.resource()).
+		Context(ctx).
+		Name(key.Name).Do().Into(obj)
```
alvaroaleman: Nit: Please keep the linebreaking the same as in the other methods

pkg/client/unstructured_client.go
```diff
@@ -22,100 +22,135 @@ import (
...
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 )
 
 // client is a client.Client that reads and writes directly from/to an API server.  It lazily initializes
 // new clients at the time they are used, and caches the client.
 type unstructuredClient struct {
+	cache      *clientCache
 	client     dynamic.Interface
 	restMapper meta.RESTMapper
```
alvaroaleman: From what I can see, both the client and the restMapper are now unused

pkg/client/codec.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"errors"
+	"net/url"
+
+	"k8s.io/apimachinery/pkg/conversion/queryparams"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type parameterCodec struct{}
```
alvaroaleman: Could we add a `var _ runtime.ParamterCodec = parameterCodec` to make this easier to understand? Also, please add some godocs to it clarifying how exactly it differs from the `ParamterCodec` obtained via the scheme

alexeldeib:
Thanks for the review, good points 👍 I think I might be able to make the parameter codec a bit cleaner, still poking at it. If not I'll drop a comment explaining.

For posterity, we [chatted on slack](https://kubernetes.slack.com/archives/CAR30FCJZ/p1582425850000500?thread_ts=1582227737.478900&cid=CAR30FCJZ) and agreed this approach makes sense

pkg/client/codec.go
alexeldeib: I think a more correct approach to take is setting the serializer on the restclient when we fetch it from the cache. I think that will allow us to remove this code and streamline further. 

If that doesn't work out, i'll add some comments that this is for no-conversion serialization of create/delete/etc params.

For comparison sake, I looked at the dynamic client go client which forces everything to metav1. So it's technically correct right now due to metav1 being the only option, but setting the serializer is the more future-proof option.

alexeldeib: added a comment and renamed for clarity

pkg/client/client_test.go
```diff
@@ -827,6 +827,51 @@ var _ = Describe("Client", func() {
...
+
+				By("updating the status of Deployment")
+				u := &unstructured.Unstructured{}
+				dep.Status.Replicas = 1
+				Expect(scheme.Convert(dep, u, nil)).To(Succeed())
+				err = cl.Status().Update(context.TODO(), u)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("validating updated Deployment has type information")
+				Expect(u.GroupVersionKind()).To(Equal(depGvk))
```
alvaroaleman: I think we should also get the deployment with the generated clientset and verify that its status got updated/patched :upside_down_face:

alexeldeib: added it for patch, I think it's there for the update calls already?

alexeldeib: not sure why they're organized this way, but I guess the tests are broken out fairly granular -- one test only does the type validation but the other two validate the actual status server side?

alvaroaleman: Oh you are right. Indeed very granular. Well, that definitely helps when debugging :)

alexeldeib:
/assign @DirectXMan12 

I imagine you'd want to review this before merging

vincepri:
LGTM as well

DirectXMan12:
awesome

/approve

alexeldeib:

50 millisecond timeout feels aggressive, not sure if there's something hiding causing a flake or if it's just a low threshold.

PR #813, opened 2020-02-23
update options.go: add a nil check

### LINKED ISSUE #806: Add nil check when applying variadic client options
All of the `Create`, `List`, `Update` and `Delete` client methods are variadic in that it takes any number of options. For example:

```
Create(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error
```

It's too easy to pass `nil` as the last argument for these methods and then it blows up with a NPE in `ApplyOptions` method. Example: https://github.com/kubernetes-sigs/controller-runtime/blob/19e72eb88faf07f1ab7f3dd7f951bc6ae85c8d9c/pkg/client/options.go#L141

There needs to be a nil check when options are applied.

boddumanohar:
:sparkles: 

This PR fixes #806 
All of the Create, List, Update and Delete client methods are variadic in that it takes any number of options.

It's too easy to pass nil as the last argument for these methods and then it crashes with a NPE.
So added a nil check.

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -138,6 +138,9 @@ func (o *CreateOptions) AsCreateOptions() *metav1.CreateOptions {
 // and then returns itself (for convenient chaining).
 func (o *CreateOptions) ApplyOptions(opts []CreateOption) *CreateOptions {
 	for _, opt := range opts {
+		if opt == nil {
+			continue
+		}
 		opt.ApplyToCreate(o)
 	}
 	return o


boddumanohar:
@alvaroaleman all the those functions, finally makes a call to `ApplyOptions` function. So I added check in that function. So I think Adding the check in `ApplyOptions` function would suffice.

alvaroaleman:
> @alvaroaleman all the those functions, finally makes a call to ApplyOptions function. So I added check in that function. So I think Adding the check in ApplyOptions function would suffice.

`ApplyOptions` is implemented as a method on each  of `CreateOptions` (there you added the check), `DeleteOptions`, `ListOptions`, `UpdateOptions`, `PatchOptions` and `DeleteAllOfOptions` (on those you missed it).

Also please extend the tests, so we don't regress on this in the future.

alvaroaleman:
/hold
Actually stop, I don't think we should be doing this, because methods of a nullpointer are still callable and we have a couple of examples where this works without issue, e.G. `MatchingLabels`, `HasLabels`, `MatchingFields`. Those examples here are a no-op if nil, but there could be examples out there where this is not the case. With this change, we silently stop applying them which is surprising at best.

IMHO a method that does not explicitly state that it may be called with nil args can expect the caller to make sure that this doesn't happen.

boddumanohar:
I agree with @alvaroaleman whoever has passed the nil value to the `Create` `Update` ...functions should handle the error and behave accordingly.

rajathagasthya:
(commenting on PR instead of issue)

> methods of a nullpointer are still callable and we have a couple of examples where this works without issue, e.G. MatchingLabels, HasLabels, MatchingFields.

@alvaroaleman This will continue to work for those examples even if there is a `if o == nil` because those values won't be `nil` even if unallocated; only their `reflect.ValueOf(o).IsNil()` will be `true` (interface values in Go). I've tried to demo here: https://play.golang.org/p/P6U4KzzR-Mm

What I'm suggestion is guarding against explicit `nil` argument.

alvaroaleman:
Yeah I know that the implementation here doesn't guard against anything other than an explicit `nil`,  althought I assumed you wanted a guard against a typed nil pointer.

Why do you think ppl would pass in an explicit nil?

rajathagasthya:
Just an easy mistake I made last week when we were upgrading an ancient controller-runtime version to latest and addressing breakage wrt to client CRUD methods and options. 

Previously we used to pass `nil` in place of `*client.ListOptions` struct to say "I don't want any options":
```
client.List(ctx, nil, serviceList)
```

The breakage meant I switched the argument order without realizing options argument type also changed from a struct to an interface:
```
client.Create(ctx, obj, nil)
```

Compiled didn't complain, of course, but it panic'ed at runtime.

It's not a big deal, but I am of the opinion that if we (or the compiler) allow for "bad values" to be sent, we should also guard against it.

alvaroaleman:
The fact that just swapping the arguments compiles successfully is indeed unfortunate.

I still think that we should not add a check for this, as it will only result in an unnecessary code being kept. Also, the next person to get a panic by passing a typed nil pointer may see the existing nilcheck and think that extending it is a good idea.

Ideally ppl have some unittests in their code that will discover this issue before rolling out anything.

PR #815, opened 2020-02-24
🐛 Use GenerateName during fake client's Create method

tvs:
When an object is created without a `Name` but with a `GenerateName`, the fake client will set a `Name` by appending a randomly generated UID to the end of the `GenerateName` prefix.

This change replicates the behavior of the [API Server's NameGenerator](https://github.com/kubernetes/apiserver/blob/master/pkg/storage/names/generate.go).

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -171,6 +171,35 @@ var _ = Describe("Fake client", func() {
 			Expect(obj.ObjectMeta.ResourceVersion).To(Equal("1"))
 		})
 
+		It("should be able to Create with GenerateName", func() {
+			By("Creating a new configmap")
+			newcm := &corev1.ConfigMap{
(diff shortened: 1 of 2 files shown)


tvs:
CLA should be signed now.

pkg/client/fake/client_test.go
```diff
@@ -171,6 +171,35 @@ var _ = Describe("Fake client", func() {
 			Expect(obj.ObjectMeta.ResourceVersion).To(Equal("1"))
 		})
 
+		It("should be able to Create with GenerateName", func() {
+			By("Creating a new configmap")
+			newcm := &corev1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
```
alvaroaleman: Nit: Setting the type meta is not needed

alvaroaleman:
/lgtm
too used to "approve acts as lgtm"...

PR #818, opened 2020-02-25
📖Adding ComponentConfig proposal

christopherhein:
This is a proposal for how we can integrate `ComponentConfig` into `controller-runtime` without breaking existing implementations and allows controller authors an easier way to configure their controllers.

This was originally documented in https://github.com/kubernetes-sigs/controller-runtime/issues/518

This uses the new design proposals from `designs/template.md` for a working example see https://github.com/christopherhein/github-controller/pull/5


designs/component-config.md
```diff
@@ -0,0 +1,241 @@
...
+## Summary
+
+Currently all any controller that uses `controller-runtime` needs to configure the `ctrl.Manager` by using flags or hard coding values into the initialization methods. Core Kubernetes resources have started to move away from using flags as a mechanism for configuring the component and standardized along the pattern of [`ComponentConfig` or Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/edit). This proposal is to bring `ComponentConfig` patterns into `controller-runtime` to allow controller authors to make `go` types backed by apimachinery to unmarshal and configure the `ctrl.Manager` reducing the flags and allowing code based tools to easily configure controllers instead of requiring them to mutate CLI args.
+
+## Motivation
+
+This change is important because: 
+- it will help make it easier for controllers to be configured by other machine processes
+- it will reduce the upfront flags required to start a controller
+- allow for more configuration types which flags don't natively support
```
stealthybox:
```suggestion
- allow for more configuration types which flags don't natively support
- allow using and upgrading older configurations avoiding breaking changes in flags
```

designs/component-config.md
```diff
@@ -0,0 +1,241 @@
...
+            * [Controller User with modifications to config](#controller-user-with-modifications-to-config)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Alternatives](#alternatives)
+      * [Implementation History](#implementation-history)
+
+<!--te-->
+
+## Summary
+
+Currently all any controller that uses `controller-runtime` needs to configure the `ctrl.Manager` by using flags or hard coding values into the initialization methods. Core Kubernetes resources have started to move away from using flags as a mechanism for configuring the component and standardized along the pattern of [`ComponentConfig` or Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/edit). This proposal is to bring `ComponentConfig` patterns into `controller-runtime` to allow controller authors to make `go` types backed by apimachinery to unmarshal and configure the `ctrl.Manager` reducing the flags and allowing code based tools to easily configure controllers instead of requiring them to mutate CLI args.
```
stealthybox:
```suggestion
Currently controllers that use `controller-runtime` need to configure the `ctrl.Manager` by using flags or hard coding values into the initialization methods. Core Kubernetes has started to move away from using flags as a mechanism for configuring components and standardized along the pattern of [`ComponentConfig` or Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/edit). This proposal is to bring `ComponentConfig` patterns into `controller-runtime` to allow controller authors to make `go` types backed by apimachinery to unmarshal and configure the `ctrl.Manager` reducing the flags and allowing code based tools to easily configure controllers instead of requiring them to mutate CLI args.
```

designs/component-config.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Non-Goals/Future Work
+
+- `kubebuilder` implementation and design in another PR
+- Changing the default `controller-runtime` implementation
+- Dynamically reloading ComponentConfig object
+
+## Proposal
+
+The `ctrl.Manager` _SHOULD_ to be able to support load configurations from an interface with getters to pull the specific configuration parameters out of it so that we can use `ComponentConfig` like objects to load the configuration from.
```
stealthybox:
```suggestion
The `ctrl.Manager` _SHOULD_ support loading configurations from `ComponentConfig` like objects.
An interface for that object with getters for the specific configuration parameters is created to bridge existing patterns.
```

designs/component-config.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+### Risks and Mitigations
+
+- Given that this isn't changing the core Manager initialization for `controller-runtime` it's fairly low risk
+- If the underlaying `ctrl.Options{}`
+
+## Alternatives
+
+* `NewFromComponentConfig()` could load the object from disk and hydrate the `ComponentConfig` type
```
stealthybox: Would the `<ComponentConfig>{}` then be used directly within the manager code, thus deprecating `ctrl.Options{}` for this alternative solution?

christopherhein: Good question, originally when I wrote this I was thinking that `NewFromComponentConfig()` should also be responsible for parsing out the config file but when I started to implement this as a demo I found it slightly difficult and would require `flags` to also be standardized which is something `controller-runtime` shouldn't care about. 

Here is an example implementation - https://github.com/christopherhein/github-controller/pull/5/files#diff-7ddfb3e035b42cd70649cc33393fe32cR72-R79

Which uses - https://github.com/christopherhein/controller-runtime/commit/c3954c0a2f5a#diff-7d903b129bc883285ffc5068c39c00b0R392-R403

I'm still debating if the file parsing should be done in `controller-runtime` especially since this example doesn't support flag overrides, I would love to see these functions make their way into https://git.k8s.io/component-base

christopherhein: Regarding deprecating `ctrl.Options{}` for backwards compatibility reasons I'm concerned about trying to tackle that in the same proposal, it's currently what is used to set up any `ctrl.NewManager()` which is scaffolded w/ `kubebuilder`

christopherhein: Yeah, great idea @shawn-hurley let me write something up and we can discuss here.

christopherhein: The nice thing I think we can leverage is the `consts` that are defined for `NewManager` which would just require us to change what that references. As of now it's `manager.New()` path wise we'd be able to leverage that and change it's references to `manager.NewFromComponentConfig()`.

designs/component-config.md
```diff
@@ -0,0 +1,244 @@
...
+```golang
+package config
+
+import (
+  "time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+) 
+
+type ControllerNameConfigurationSpec struct {
```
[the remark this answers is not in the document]
christopherhein: I think this could work, I'm currently exploring before writing it all up. My main concerns are making sure we have a clear path for using flags as overrides for the ComponentConfig type. I'm hesitant to make it so that `controller-runtime` has to have a notion of `Flags` since it's such an opinionated space.

christopherhein:
@stealthybox @shawn-hurley updated with new `DefaultControllerConfig` struct, and using unmarshalling built into the `NewForComponentConfig()` func.

designs/component-config.md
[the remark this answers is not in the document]
christopherhein: Oops, I read this wrong and made a type `DefaultControllerConfig` type. Let me roll this back slightly and make it just the embeddable, or provide an alternative we can also discuss.

christopherhein:
@stealthybox @shawn-hurley updated again to include the embeddable type of `metav1alpha1.RuntimeConfig` and adding how that supports `DefaultControllerConfiguration`.

Thanks again for taking a look at this.

designs/component-config.md
```diff
@@ -0,0 +1,312 @@
...
+}
+mgr, err := ctrl.NewManagerFromComponentConfig(ctrl.GetConfigOrDie(), scheme, configname, )
+if err != nil {
+	// ...
+}
+```
+
+#### Kubebuilder Scaffolding Example
+
+Within expanded in a separate design _(link once created)_ this will require controller authors to generate a type that implements the `ManagerConfiguration` interface. The following is a sample of what this looks like:
```
shawn-hurley: Just a note, that within this design we would need to expose the component config from the manager then so that a user could retrieve the custom type to make the custom choices they would like to make I think?

designs/component-config.md
```diff
@@ -0,0 +1,312 @@
...
+
+	return New(config, options)
+}
+```
+
+#### ComponentConfig Load Order
+
+![ComponentConfig Load Order](/designs/images/component-config-load.png)
+
+#### Embeddable ComponentConfig Type
```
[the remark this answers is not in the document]
christopherhein: I think that makes sense. I'm hoping to pull as much of the work from https://github.com/kubernetes/component-base/blob/master/config/v1alpha1/types.go for example pulling the `LeaderElectionConfiguration` potentially this is a better place to push more of those configurations. 

Fair point that `meta` might not be the right place long term for it, have any suggestions? maybe `config` -> `pkg/api/config/v1alpha1/types.go` like I had used for the `DefaultControllerConfiguration`…

UX wise I would expect that someone could embed this into their type as a natural way to extend to contain non-manager specific components. Eg:

```
package config

import (
  "time"

  metav1alpha1 "sigs.k8s.io/controller-runtime/pkg/apis/meta/v1alpha1"
) 

type ControllerNameConfigurationSpec struct {
  metav1alpha1.RuntimeConfig `json:",inline"`

  // ... Other configurations that are specific to the custom controller
}

type ControllerNameConfiguration struct {
  metav1.TypeMeta

  Spec ControllerNameConfigurationSpec `json:"spec"`
}
```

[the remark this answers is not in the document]
shawn-hurley: Should we call these `pkg/api/config/v1alpha1/types.go` then?

designs/component-config.md
```diff
@@ -0,0 +1,312 @@
...
+	GetReadinessEndpointName() string
+	GetLivenessEndpointName() string
+
+	GetPort() int
+	GetHost() string
+
+	GetCertDir() string
+}
+
+func NewFromComponentConfig(config *rest.Config, scheme *runtime.Scheme, filename string, managerconfig ManagerConfiguration) (Manager, error) {
```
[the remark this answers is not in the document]
christopherhein: Yeah, to initialize the `ctrl.Manager{}` we need to have a `config` which is currently done something like `New(config *rest.Config, options Options)` or for an implemented version:

```
    Scheme:                      scheme,
    MetricsBindAddress: metricsAddr,
    LeaderElection:          enableLeaderElection,
    Port:                             9443,
    SyncPeriod:                &resyncTimeout,
})
```

Since this this is just passed through into the `Options{}` we could make the user set this on the `ManagerConfiguration` type before passing into `NewFromComponentConfig` my concern is that they we're representing non-`yaml` declarable resources into the ComponentConfig type. For example it could be:

```
mgr, err := ctrl.NewManagerFromComponentConfig(configname, &defaultv1alpha1.DefaultControllerConfiguration{
  Client: ctrl.GetConfigOrDie(),
  Scheme:  scheme, 
})
```

But given it currently separates `config` in `New(config *rest.Config, options Options) (Manager, error)`  I'm favoring not packing those fields into the `ComponentConfig` type.

thoughts?

designs/component-config.md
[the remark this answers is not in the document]
lavalamp: This doesn't belong in the meta api.

designs/component-config.md
```diff
@@ -0,0 +1,312 @@
...
+package v1alpha1
+
+import (
+	"time"
+
+	configv1alpha1 "k8s.io/component-base/config/v1alpha1"
+)
+
+// RuntimeConfig defines the embedded RuntimeConfiguration for controller-runtime clients.
+type RuntimeConfig struct {
```
lavalamp: "RuntimeConfig" isn't great. Maybe "ControllerConfig" or "ProcessConfig"...

designs/component-config.md
christopherhein: @shawn-hurley given that you pass in a pointer the `decodeComponentConfigFileInto()` can hydrate the ComponentConfig type. In my examples this is implemented as:

```
// decodeComponentConfigFileInto will load the component config into the runtime.Object
func decodeComponentConfigFileInto(codecs serializer.CodecFactory, filePath string, obj runtime.Object) error {
	content, err := ioutil.ReadFile(filePath)
	if err != nil {
		return err
	}
	// Regardless of if the bytes are of any external version,
	// it will be read successfully and converted into the internal version
	return runtime.DecodeInto(codecs.UniversalDecoder(), content, obj)
}
```

Does that cover what you are saying? I could see someone then passing that object into the controllers for use of the configs from the reconcilers (at least in kubebuilder's case). Potentially it makes sense to also make `GetComponentConfig()` on the `Manager` interface so that everything is self contained.

Thoughts?

christopherhein:
> 2\. A new design on how a user could extend the default stuff including adding scaffolding for that?

Correct and #2 will be another PR into Kubebuilder as I'd assume the scaffolding will live there. I'm holding off submitting that until we have this one ironed out.

designs/component-config.md
christopherhein: Yeah, those sound much better.

designs/component-config.md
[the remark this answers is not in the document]
mtaufen: ComponentConfig types should only contain types that can be expressed in yaml (just like the other API definitions). They need to be serializable. Once you load them, fine to transform/copy them into runtime structures with more advanced interfaces.

designs/component-config.md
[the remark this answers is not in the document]
mtaufen: Why put this in k/k at all? Is there a reason we can't offer these types, which are for controller-runtime users, from the controller-runtime repo?

Wrt. component-base, I'd like to take a conservative approach to what we put in there. The purpose of that repo is to consolidate foundational code that was spread out across k/k, and improve some of those frameworks within the confines of the core project. Since the controllers in k/k (to my knowledge...which isn't comprehensive) aren't (yet) implemented on controller-runtime, I'm not sure it makes sense to put these types in component-base (yet).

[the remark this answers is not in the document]
christopherhein: sorry that import is from `sigs.k8s.io/controller-runtime/pkg/apis/meta/v1alpha1` in `contoller-runtime` I was not expecting that to end up in k/k.

designs/component-config.md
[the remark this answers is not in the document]
christopherhein: Perfect that was my understanding, so I think it makes sense to pass config and scheme since those are needed to setup the manager properly as params into this func. Agree?

christopherhein:
@mtaufen @lavalamp @shawn-hurley @stealthybox thanks for all the feedback, this has been updated with the most recent changes:

* Renamed embedded `RuntimeConfig` to `ControllerConfiguration`
* Moved the embedded types into `pkg/api/config/v1alpha1/types.go` so they'll now will live with the `DefaultControllerConfiguration`
* Cleaned up some wording.

Super appreciate the time reviewing this 😃

designs/component-config.md
shawn-hurley: I was thinking about how a user could use their custom component config in their reconciler, I was thinking of someting like  `GetComonentConfig(<Interface>)` such that they could pass in their component config and we would populate it? 

IDK WDYT?

christopherhein: Ah, since the `NewFromComponentConfig()` func does this when you call it with a custom type or the default type you'll already have that populated within your `main.go` or where ever you init'ed the manager, so passing it into the reconciler as another struct tag seems like it would work the same? If it was just `GetComponentConfig() ManagerConfiguration` I could understand but I don't think providing another want to unmarshal is necessary.

christopherhein: The example looks something like this:

```
componentConfig := &custom.HighlyCustomizedControllerConfiguration{}
mgr, err := ctrl.NewManagerFromComponentConfig(ctrl.GetConfigOrDie(), scheme, configname, componentConfig)

fmt.Printf("%+v", componentConfig)
fmt.Printf("GetPort() %d", componentConfig.GetPort())
```

and this will return the information from the config passed in via `configname`.

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+
+	Spec   configv1alpha1.ControllerConfiguration   `json:"spec,omitempty"`
+}
+```
+
+This would allow a controller author to use this struct with any config that supports the json/yaml structure. For example a controller author could define their `Kind` as `FoobarControllerConfiguration` and have it defined as the following.
+
+```yaml
+# config.yaml
+apiVersion: somedomain.io/v1alpha1
```
djzager: I see in [Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/edit#), it mentions that core cluster components should define an API group of the form `{component}.config.k8s.io` but it doesn't mention anything about non-core controllers. Should we have an opinion on what the API group _should_ be?

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+kind: FoobarControllerConfiguration
```
djzager: I noticed in [Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/edit#) as suggesting the kind to be of this form `{Component}Configuration` -- which you have done here -- and I wonder if we should specifically call this out.

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+The `ctrl.Manager` _SHOULD_ support loading configurations from `ComponentConfig` like objects. 
+An interface for that object with getters for the specific configuration parameters is created to bridge existing patterns.
+
+Without breaking the current `ctrl.NewManager` which uses an exported `ctrl.Options{}` the `manager.go` can expose a new func, `NewFromComponentConfig()` this would be able to loop through the getters to hydrate an internal `ctrl.Options{}` and pass that into `New()`.
+
+```golang
+//pkg/manager/manager.go
+
+// ManagerConfiguration defines what the ComponentConfig object for ControllerRuntime needs to support
+type ManagerConfiguration interface {
```
djzager: I am weak on semantic versioning rules but it looks to me that adding something to the [`Options` struct](https://github.com/kubernetes-sigs/controller-runtime/blob/edf717755ce2f4f6c8794c976b0184fc65676230/pkg/manager/manager.go#L101) would not be a breaking change now but a new argument for `ManagerConfiguration` would always be a breaking change. Am I understanding that correctly? Do we care?

designs/component-config.md
christopherhein: Since we technically can't use `___.k8s.io` I don't personally think we should… unless it's something like telling folks to use `config.[theirdomain]`… for now I'm changing this example to add the config subdomain.

For the second half of this proposal, the integration in Kubebuilder, I think this will be smart to call out in the Kubebuilder documentation, less so in controller-runtime. 

Does that seem reasonable?

designs/component-config.md
christopherhein: Same comment regarding that I think this should be included in the second half about integration with Kubebuilder and put into that design doc. I'm holding off pushing that proposal until this is fully merged.

designs/component-config.md
christopherhein: his is a really good point and you are totally correct adding a field would be a breaking change for anyone that generated their own `ComponentConfig` object and didn't just use the embedded types. Although if they used the embedded type it should be okay.

For example if someone wrote a custom type like so:

```type ControllerNameConfigurationSpec struct {
	configv1alpha1.ControllerConfiguration `json:",inline"`
}

type ControllerNameConfiguration struct {
  metav1.TypeMeta

  Spec ControllerNameConfigurationSpec `json:"spec"`
}```

… and we added a new func and field to `configv1alpha1.ControllerConfiguration` they would be covered out of the box, but if they implemented then own Getters then it would break. I _personally_ think that is would be fine since it centralizes the `controller-runtime` specific configurations and still allows me to add my own fields for my custom controller.

What do you think? @djzager

File: designs/component-config.md [hunk 13]
```diff
@@ -0,0 +1,318 @@
...
+
+	GetPort() int
+	GetHost() string
+
+	GetCertDir() string
+}
+
+func NewFromComponentConfig(config *rest.Config, scheme *runtime.Scheme, filename string, managerconfig ManagerConfiguration) (Manager, error) {
+	codecs := serializer.NewCodecFactory(scheme)
+    if err := decodeComponentConfigFileInto(codecs, filename, componentconfig); err != nil {
```
djzager: Should `componentconfig` here be `managerconfig`?
```suggestion
	if err := decodeComponentConfigFileInto(codecs, filename, componentconfig); err != nil {
```

christopherhein: Yes, good catch!

designs/component-config.md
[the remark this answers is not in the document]
mtaufen: You already know which types you need in your scheme, so there isn't necessarily a reason to leave room for error there. When we parse Kubelet configs we construct the scheme internally. We have helpers like  https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/scheme/scheme.go#L31 to avoid duplicating scheme construction boilerplate.

designs/component-config.md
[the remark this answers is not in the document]
mtaufen: Ah, makes sense :)
Thanks for the update.

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+      * [Motivation](#motivation)
+         * [Links to Open Issues](#links-to-open-issues)
+         * [Goals](#goals)
+         * [Non-Goals/Future Work](#non-goalsfuture-work)
+      * [Proposal](#proposal)
+            * [ComponentConfig Load Order](#componentconfig-load-order)
+            * [Embeddable ComponentConfig Type](#embeddable-componentconfig-type)
+            * [Default ComponentConfig Type](#default-componentconfig-type)
+            * [Using Flags w/ ComponentConfig](#using-flags-w-componentconfig)
+            * [Kubebuilder Scaffolding Example](#kubebuilder-scaffolding-example)
```
djzager: Noticed it was displaying funky:

```suggestion
         * [ComponentConfig Load Order](#componentconfig-load-order)
         * [Embeddable ComponentConfig Type](#embeddable-componentconfig-type)
         * [Default ComponentConfig Type](#default-componentconfig-type)
         * [Using Flags w/ ComponentConfig](#using-flags-w-componentconfig)
         * [Kubebuilder Scaffolding Example](#kubebuilder-scaffolding-example)
```

designs/component-config.md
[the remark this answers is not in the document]
christopherhein: Does that work because we already know the `kubeletconfig` scheme where as in this it could be custom from that specific controller, maybe I'm not just clear enough on how al that is working… 

For example in a kubebuilder scaffolded project we'd have this - https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/main.go#L39-L44 to register the custom scheme, and they are typically passed into the Manager like so - https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/main.go#L57-L63 given this currently done it seems like we'll need this internally anyways for controller-runtime to work.

So while part of the reason to include in the `NewForComponentConfig()` was to pass in for decoding from a file we also need it to init the actual Manager.

christopherhein:
@shawn-hurley @mtaufen @stealthybox @vincepri any more feedback, I'd love to start moving this forward and get a implementation in place.

designs/component-config.md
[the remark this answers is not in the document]
mtaufen: That's fair. SGTM.

mtaufen:
Forge ahead! :) 

Do take a spin through the [Versioned Component Configuration Files](https://docs.google.com/document/d/1FdaEJUEh091qf5B98HM6_8MS764iXrxxigNIdwHYW9c/) doc (2018) while you're implementing and see how this compares (kubernetes-dev membership grants access).

/lgtm

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+	// ...
+}
+```
+
+The above example uses `configname` which is the name of the file to load the configuration from and uses `scheme` to get the specific serializer, eg `serializer.NewCodecFactory(scheme)`. This will allow the configuration to be unmarshalled into the `runtime.Object` type and passed into the 
+`ctrl.NewManagerFromComponentConfig()` as a `ManagerConfiguration` interface.
+
+#### Using Flags w/ ComponentConfig
+
+Since this design still requires setting up the initial `ComponentConfig` type and passing in a pointer to `ctrl.NewFromComponentConfig()` if you want to allow for the use of flags, your controller can use any of the different flagging interfaces. eg [`flag`](https://golang.org/pkg/flag/), [`pflag`](https://godoc.org/github.com/spf13/pflag), [`flagnum`](https://godoc.org/github.com/luci/luci-go/common/flag/flagenum) and set values on the `ComponentConfig` type prior to passing the pointer into the `ctrl.NewFromComponentConfig()`, example below.
```
mtaufen: Just FYI for backwards compatibility flags usually need to take precedence over config (unless you can force users to use exclusively flags or config).

christopherhein: Good callout, since we're decoding in the same object that the flags could have already been set on there isn't a way I'm aware of to not override existing values while decoding. Have any suggestions?

designs/component-config.md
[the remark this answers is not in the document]
vincepri: From `controller-runtime` user perspective having a meta package in our repository makes sense to offer basic types that are configurable when users implement ComponentConfig in their managers, +1

designs/component-config.md
shawn-hurley: save off the flags and apply them after the decode? (merge the options structs together more or less)?

christopherhein: Yea, that probably makes the most sense. I personally don't think this needs to be documented in the design doc, do you? 

Feeling okay with the rest of this @shawn-hurley ?

joelanford: Instead of `ctrl.NewFromComponentConfig()` that returns `Manager`, would it be reasonable to have `ctrl.NewOptionsFromComponentConfig()`  that returns `manager.Options`. And then let callers call `manager.New()` themselves so that they have more flexibility in updating `manager.Options` after it has been populated?

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+
+	MetricsBindAddress string `json:"metricsBindAddress,omitempty"`
+
+	Health ControllerConfigurationHealth `json:"health,omitempty"`
+
+	Port *int   `json:"port,omitempty"`
+	Host string `json:"host,omitempty"`
+
+	CertDir string `json:"certDir,omitempty"`
+}
```
joelanford: The original issue (#518) suggested being able to use ComponentConfig to tweak logging settings. In controller-runtime, loggers are handled outside the scope of `manager.Options`. Should we consider expanding the scope of the default ComponentConfig to be opinionated about more than just `manager.Options` configuration?

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+package v1alpha1
+
+import (
+	"time"
+
+	configv1alpha1 "k8s.io/component-base/config/v1alpha1"
+)
+
+// ControllerConfiguration defines the embedded RuntimeConfiguration for controller-runtime clients.
+type ControllerConfiguration struct {
```
joelanford: Maybe I'm missing something, but should this struct include `Namespace` to align with the [`ManagerConfiguration` interface](https://github.com/kubernetes-sigs/controller-runtime/pull/818/files#diff-52330260780d994c6c68211c4961e13cR88)?

Also, controller-runtime supports [multi-namespace caches](https://github.com/kubernetes-sigs/controller-runtime/blob/a11a908d91e0c9aeabad65c8788ed4616a864ced/pkg/cache/multi_namespace_cache.go#L42). Do we expect users to be able to use ComponentConfig for this use case, or will we require users to set [`Options.NewCache`](https://github.com/kubernetes-sigs/controller-runtime/blob/a11a908d91e0c9aeabad65c8788ed4616a864ced/pkg/manager/manager.go#L180) manually.

designs/component-config.md
christopherhein: Totally a possibility. I'd originally started with that path and ended up combining the funcs to make the new API quicker to implement. 

@shawn-hurley @mtaufen @DirectXMan12 @stealthybox  Do you have a preference? The options would be:

1.  Use a single function `ctrl.NewFromComponentConfig()` which takes the `ComponentConfig` type and returns a `ctrl.Manager`.

### or

2. Separate the funcs and make it `ctrl.NewOptionsFromComponentConfig()` which returns `manager.Options` that implementation wise would be passed into `manager.New()`.

designs/component-config.md
christopherhein: Good catch, yes it should. I glanced over that since `configv1alpha1.LeaderElectionConfiguration` has `.ResourceNamespace` but yes we should support `.Namespace` on this type.

WRT multi-namespace caches, I would hope we can support this but I'm not very familiar with the current way this is used if you have more info that would be helpful.

designs/component-config.md
christopherhein: I'm concerned about this being too opinionated and something we should avoid from the controller-runtime side. We'll at least have a path now for folks like `kubebuilder` to build their own custom type or even if @johnsonj would want to they could build a custom `ComponentConfig` type for the `addon` pattern allowing them to standardize.

WDYT?

johnsonj: Yeah echoing/adding to that, I'd like to see downstream tools follow this same pattern but it does not need to be in the same type.

designs/component-config.md
```diff
@@ -0,0 +1,318 @@
...
+	}
+	options := Options{}
+
+	if scheme != nil {
+		options.Scheme = scheme
+	}
+
+	// Loop through getters
+	if managerconfig.GetLeaderElection() {
+			managerconfig.GetLeaderElection()
```
johnsonj: I know this is pseudocode here, but I think you want something like

```go
optiosn.LeaderElection = ..
```

designs/component-config.md
joelanford: For single namespaces or all namespaces, using `opts.Namespace` does the right thing. For multiple explicit namespaces, it is necessary to use:

```go
namespaces := []string{"ns1", "ns2"}
opts.NewCache = cache.MultiNamespacedCacheBuilder(namespaces)
```

designs/component-config.md
joelanford: :+1: Sounds good. I agree extending the type seems like the best approach.

designs/component-config.md
joelanford: Another thought would be for the ComponentConfig-populating function to accept a pointer to a `manager.Options` struct so that something like this would be possible:

```go
func main() {
   opts := manager.Options{}
   _ = flag.StringVar(&opts.Namespace, "namespace", "", "Namespace to watch.")

   ctrl.UpdateOptionsFromComponentConfig(&opts)

   // Parse flags after populating from ComponentConfig so that flags
   // take precedence
   _ = flag.Parse()
   _ = manager.New(ctrl.GetConfigOrDie(), opts)
}

designs/component-config.md
christopherhein: Thanks @johnsonj you are absolutely correct 🤦‍♂

designs/component-config.md
christopherhein: I feel like this should be possible as long as we have a `Namespaces` field in the type. 

What do you think about doing this as an addition after this is initial design is implemented? Since it already requires developers to add custom implementations it would be the same after this first implementation then maybe you and I can figure the edge cases we might run into and we can add that feature after?

designs/component-config.md
christopherhein: Fixed.

designs/component-config.md
christopherhein: Yeah, I think both are fair flows. I think it might be easier to guarantee that flags take precedence by decoupling the parsing and new manager initialization logic. This also opens the door later to add the `ctrl.NewManagerFromComponentConfig()`. Where as it would be hard to walk backwards from having both manager init and populating the `Options{}` struct.

Does anyone have strong opinions against this? @stealthybox @johnsonj @joelanford @djzager @shawn-hurley @mtaufen

designs/component-config.md
joelanford: Yep, this could definitely be done as a non-breaking change as a new feature. Sounds good :+1:

shawn-hurley:
/approve

If any other followups we can still discuss here, but let's get an implementation started and if more things come up we can discuss.

PR #821, opened 2020-02-25
🏃 Add verification tools, linter and modules

vincepri:

This PR cleans up the use of linters, uses Go Modules to build specific version of external tools (golangci-lint), and provides a way to check that go modules are always up to date.


Makefile
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-#  Copyright 2019 The Kubernetes Authors.
+#  Copyright 2020 The Kubernetes Authors.
```
shawn-hurley: more for my own knowledge, do we usually update copyright year in kubernetes when a file is updated?

hypnoglow:
It seems `hack/tools/hack/tools/bin/golangci-lint` binary is committed unintentionally.

The rest LGTM

vincepri:
@hypnoglow I force pushed and removed it, I also don't see it in the file list, am I missing something?

Makefile
vincepri: I don't have a good answer, I changed it because it was mostly rewritten

vincepri:
/assign @alenkacz @gerred 

would love to get more 👀 and feedback on this one

gerred:
:+1: I'll take a look - was out sick for the past few days

alenkacz:
/lgtm

this actually looks great, love the cleaned up makefile. Should we update some documentation or use the new targets somewhere? 🤔

Issue #684, opened 2019-11-14
Expose HTTP Endpoints listeners used for metrics


salaboy:
Right now the `/metrics` endpoint is automatically created and hosted using port `8080` it will be nice for users to be able to register other HTTP endpoints using the same listener under `8080`. Today, looking at the code this is not possible.

DirectXMan12:
/kind feature
/priority backlog

We usually just serve other things on a different port, but yeah, this could be useful.

PR #824, opened 2020-02-25
✨Expose metrics http server for extra endpoints

hypnoglow:
Allows users to add extra http handlers served on custom path to the http server that serves metrics. Might be useful to register some diagnostic endpoints e.g. pprof or custom debug info.

Ref: #684

---

Some additional thoughts: since we expose this, naming the server "metrics server" is not 100% correct anymore. Probably we might call it "diagnostic server" or something, that describes the purpose more clearly. Of course renaming would be a breaking change. As far as I found, the only metrics-related name we expose in the API is `MetricsBindAddress` in manager options.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -50,6 +50,7 @@ const (
 
 	defaultReadinessEndpoint = "/readyz"
 	defaultLivenessEndpoint  = "/healthz"
+	defaultMetricsEndpoint   = "/metrics"
 )
 
 var log = logf.RuntimeLog.WithName("manager")
@@ -94,6 +95,9 @@ type controllerManager struct {
 
 	// metricsListener is used to serve prometheus metrics
 	metricsListener net.Listener
+
+	// metricsExtraHandlers contains extra handlers to register on http server that serves metrics.
+	metricsExtraHandlers map[string]http.Handler
 
 	// healthProbeListener is used to serve liveness probe
 	healthProbeListener net.Listener
@@ -260,6 +264,25 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	return nil
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
+func (cm *controllerManager) AddMetricsExtraHandler(path string, handler http.Handler) error {
+	if path == defaultMetricsEndpoint {
+		return fmt.Errorf("overriding builtin %s endpoint is not allowed", defaultMetricsEndpoint)
+	}
+
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+
+	_, found := cm.metricsExtraHandlers[path]
+	if found {
+		return fmt.Errorf("can't register extra handler by duplicate path %q on metrics http server", path)
+	}
+
+	cm.metricsExtraHandlers[path] = handler
+	log.V(2).Info("Registering metrics http server extra handler", "path", path)
+	return nil
+}
+
 // AddHealthzCheck allows you to add Healthz checker
 func (cm *controllerManager) AddHealthzCheck(name string, check healthz.Checker) error {
 	cm.mu.Lock()
@@ -341,19 +364,28 @@ func (cm *controllerManager) GetWebhookServer() *webhook.Server {
 }
 
 func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
-	var metricsPath = "/metrics"
 	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{
 		ErrorHandling: promhttp.HTTPErrorOnError,
 	})
 	// TODO(JoelSpeed): Use existing Kubernetes machinery for serving metrics
 	mux := http.NewServeMux()
-	mux.Handle(metricsPath, handler)
+	mux.Handle(defaultMetricsEndpoint, handler)
+
+	func() {
+		cm.mu.Lock()
+		defer cm.mu.Unlock()
+
+		for path, extraHandler := range cm.metricsExtraHandlers {
+			mux.Handle(path, extraHandler)
+		}
+	}()
+
 	server := http.Server{
 		Handler: mux,
 	}
 	// Run the server
 	go func() {
-		log.Info("starting metrics server", "path", metricsPath)
+		log.Info("starting metrics server", "path", defaultMetricsEndpoint)
 		if err := server.Serve(cm.metricsListener); err != nil && err != http.ErrServerClosed {
 			cm.errSignal.SignalError(err)
 		}
@@ -367,6 +399,8 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
 }
 
 func (cm *controllerManager) serveHealthProbes(stop <-chan struct{}) {
+	// TODO(hypnoglow): refactor locking to use anonymous func in the similar way
+	// it's done in serveMetrics.
 	cm.mu.Lock()
 	mux := http.NewServeMux()
 
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -412,7 +412,7 @@ var _ = Describe("manger.Manager", func() {
 				Expect(resp.StatusCode).To(Equal(200))
 			})
 
-			It("should not serve anything other than metrics endpoint", func(done Done) {
+			It("should not serve anything other than metrics endpoint by default", func(done Done) {
 				opts.MetricsBindAddress = ":0"
 				m, err := New(cfg, opts)
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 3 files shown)


pkg/manager/internal.go
```diff
@@ -260,6 +263,17 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	return nil
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
+func (cm *controllerManager) AddMetricsExtraHandler(path string, handler http.Handler) {
+	_, found := cm.metricsExtraHandlers[path]
+	if found {
+		panic(fmt.Errorf("can't register extra handler by duplicate path %q on metrics http server", path))
```
hypnoglow: Not sure if panic is preferred here over an explicit error, but I followed the way similar implemented in [webhook server](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/server.go#L110)

pkg/manager/internal.go
```diff
@@ -260,6 +263,20 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	return nil
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
+func (cm *controllerManager) AddMetricsExtraHandler(path string, handler http.Handler) {
```
vincepri: I'd return an error here instead of adding the panic

pkg/manager/internal.go
```diff
@@ -260,6 +263,20 @@ func (cm *controllerManager) SetFields(i interface{}) error {
...
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+
+	_, found := cm.metricsExtraHandlers[path]
+	if found {
+		panic(fmt.Errorf("can't register extra handler by duplicate path %q on metrics http server", path))
+	}
+
+	cm.metricsExtraHandlers[path] = handler
+	log.Info("registering metrics http server extra handler", "path", path)
```
vincepri:
```suggestion
	log.Info("Registering metrics http server extra handler", "path", path)
```

also maybe use V(2)?

pkg/manager/internal.go
vincepri: I'd return an error as suggested above

pkg/manager/internal.go
```diff
@@ -348,6 +365,13 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
 	// TODO(JoelSpeed): Use existing Kubernetes machinery for serving metrics
 	mux := http.NewServeMux()
 	mux.Handle(metricsPath, handler)
+
+	cm.mu.Lock()
```
vincepri: Do we need the lock at this point? I'd assume it's safe to read when `serveMetrics` is called. If we do, I'd use 
```
cm.mu.Lock()
defer cm.mu.Unlock()
```
pattern for consistency with the other methods

alvaroaleman:
I am not really sure if I like the idea of allowing to re-use the builtin server for arbitrary handlers rather than just adding a new listener as `Runnable` to the manager. The next persons usecase might need a more complex routing logic than just path->handler and that is IMHO a problem that doesn't really belong here (here as in: The `manager` package).

If this is about pprof/other diagnostics in particular we could probably add an option to enable those specifically. WDYT?

vincepri:
That's a fair statement, although seems there might be some potential useful use cases, and we can leave it up to the users. I don't have strong opinions here, I think either way can be fine

pkg/manager/internal.go
hypnoglow: Normally we don't need the lock, but I think it is theoretically possible to trigger race condition if `Start`-ing and `AddMetricsExtraHandler`-ing at the same time. So probably it's better just to have the mutex.

Regarding

```
cm.mu.Lock()
defer cm.mu.Unlock()
```

It is not suitable for this case, because `serveMetrics` blocks on receiving from `stop` channel thus the method does not return until the manager is stopped. This is the reason `serveHealthProbes` also does not follow the pattern.

pkg/manager/internal.go
hypnoglow: Done

pkg/manager/internal.go
hypnoglow: Yep, done

pkg/manager/internal.go
```diff
@@ -260,6 +263,21 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	return nil
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
```
alvaroaleman: Could you extend this comment to mention that if the simple `path -> handler` mapping offered here is not enough, a new httpListener should be added as `Runnable` to the manager? (My impression is that most ppl are not aware that the manager can run anything, not just controllers)

alvaroaleman: And probably also that this endpoint is somewhat sensitive and shouldn't be exposed publicly

pkg/manager/internal.go
rajathagasthya: Think you can wrap the locking/unlocking and the loop inside a `func` and use `defer` there. That ensures that even if something bad happens during the loop, lock is released. 

```go
func() {
    c.mu.Lock()
    defer c.mu.Unlock()
    for path, extraHandler := range cm.metricsExtraHandlers {
        mux.Handle(path, extraHandler)
    }
}()
```

pkg/manager/internal.go
hypnoglow: Thank you for the suggestion,  done that. Please check if I formulated the comment correctly.

pkg/manager/internal.go
hypnoglow: Nice tip, thanks! Changed according to your snippet.

pkg/manager/internal.go
hypnoglow: Forgot to mention that I've added the comment to the interface, not the implementation.

pkg/manager/internal.go
rajathagasthya: @hypnoglow Could you also a TODO comment to make sure we do this in `serveHealthProbes` function too? Thanks.

pkg/manager/internal.go
alvaroaleman: sounds good, thank you

pkg/manager/internal.go
```diff
@@ -260,6 +263,21 @@ func (cm *controllerManager) SetFields(i interface{}) error {
...
+func (cm *controllerManager) AddMetricsExtraHandler(path string, handler http.Handler) error {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+
+	_, found := cm.metricsExtraHandlers[path]
```
alvaroaleman: This allows to overwrite the builtin metrics endpoint as its not in the `metricsExtraHandlers` map or did I miss something?

pkg/manager/internal.go
hypnoglow: Do you suggest adding this comment in `serveHealthProbes` or in `serveMetrics`?

pkg/manager/internal.go
hypnoglow: @alvaroaleman  Yeah, I missed this point, thanks. If we try to override `/metrics` endpoint, `serveMetrics` will actually panic.

What could you suggest here? Options that come to my mind:
- When adding extra handler, if we see `/metrics`, we log this attempt now, ignoring this overriden endpoint
- When trying to register `/metrics` override in `AddMetricsExtraHandler`, we could return an error saying this is not allowed.

Also we need to document this nuance on `AddMetricsExtraHandler` method.

pkg/manager/internal.go
rajathagasthya: Wherever applicable, please.

pkg/manager/internal.go
alvaroaleman: Its a bit unfortunate that this can only be discovered at runtime. IMHO we should return an error, otherwise we probably hide bugs and keep ppl wondering why their metrics do not appear.

pkg/manager/internal.go
hypnoglow: Added comment to `serveHealthProbes`, thanks.

pkg/manager/internal.go
hypnoglow: Added the check to `AddMetricsExtraHandler`, PTAL

pkg/manager/manager_test.go
```diff
@@ -469,6 +469,40 @@ var _ = Describe("manger.Manager", func() {
...
+				err = m.AddMetricsExtraHandler("/debug", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+					_, _ = w.Write([]byte("Another debug info"))
+				}))
+				Expect(err).To(HaveOccurred())
+
+				s := make(chan struct{})
+				defer close(s)
+				go func() {
+					defer GinkgoRecover()
+					Expect(m.Start(s)).NotTo(HaveOccurred())
```
alvaroaleman: Doesn't the start race with the the Get request we do later on?

pkg/manager/internal.go
alvaroaleman: thanks

pkg/manager/manager_test.go
hypnoglow: I've done this in a similar way as in other tests where HTTP requests are involved. It seems to work, but I see no strong argument why this could not race potentially, maybe I'm missing something.

alvaroaleman: hm okay, interesting. Well, we can still look at it if we actually see it flake.

alvaroaleman:
/lgtm

Thanks for your work!

alvaroaleman:
@vincepri is this currently still missing something from your POV?

vincepri:
I think we're good, I was hoping to get the TODOs in issues though

/approve

PR #830, opened 2020-02-27
⚠ update Informers interface to accept context

### LINKED ISSUE #828: Update Informers interface to accept context on `GetInformer(ForKind)`
Follow-up based on https://github.com/kubernetes-sigs/controller-runtime/pull/663#discussion_r384794159

https://github.com/kubernetes-sigs/controller-runtime/blob/9f8aab6551a3ecc211b6c6078245899495b7e0f5/pkg/cache/informer_cache.go#L141-L142

djzager:
Update all functions on the Informers interface (and IndexField on
FieldIndexer interface) to accept a context.

Fixes #828

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cache
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -51,11 +52,11 @@ type Cache interface {
 type Informers interface {
 	// GetInformer fetches or constructs an informer for the given object that corresponds to a single
 	// API kind and resource.
-	GetInformer(obj runtime.Object) (Informer, error)
+	GetInformer(ctx context.Context, obj runtime.Object) (Informer, error)
 
 	// GetInformerForKind is similar to GetInformer, except that it takes a group-version-kind, instead
 	// of the underlying object.
-	GetInformerForKind(gvk schema.GroupVersionKind) (Informer, error)
+	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind) (Informer, error)
 
 	// Start runs all the informers known to this cache until the given channel is closed.
 	// It blocks.
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -17,6 +17,7 @@ limitations under the License.
 package source
 
 import (
+	"context"
 	"fmt"
 	"sync"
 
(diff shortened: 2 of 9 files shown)


djzager:
@DirectXMan12 follow-up to #663

djzager:
Made this :warning: since it changes the function signatures on public interfaces.

pkg/cache/cache.go
```diff
@@ -51,11 +52,11 @@ type Cache interface {
 type Informers interface {
 	// GetInformer fetches or constructs an informer for the given object that corresponds to a single
 	// API kind and resource.
-	GetInformer(obj runtime.Object) (Informer, error)
+	GetInformer(ctx context.Context, obj runtime.Object) (Informer, error)
```
alvaroaleman: @DirectXMan12 what is our policy for breaking changes? IMHO this change would be extremely great, if we want to give a longer deprecation period we could probably add a new `InformersV2` interface and start using that internally?

djzager:
@alvaroaleman I made it `WIP` wanting to add some form of test (outside of the tests that were update so it would build) but nothing stands out as a good candidate. Any ideas?

alvaroaleman:
> @alvaroaleman I made it WIP wanting to add some form of test (outside of the tests that were update so it would build) but nothing stands out as a good candidate. Any ideas?

A testcase that tries to get an informer from an envtest cluster with a context with super low timeout (1 ms or sth), then verifying the error is `context canceled`?

pkg/cache/cache.go
vincepri: We'll probably need to add this to v0.6 if it's a breaking change

pkg/source/source.go
```diff
@@ -98,7 +99,7 @@ func (ks *Kind) Start(handler handler.EventHandler, queue workqueue.RateLimiting
 	}
 
 	// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
-	i, err := ks.cache.GetInformer(ks.Type)
+	i, err := ks.cache.GetInformer(context.TODO(), ks.Type)
```
joelanford: Do we need to add `context.Context` to the `Source` interface's `Start` method to make this cancellable?

alvaroaleman: I guess, but maybe we can do that in a follow-up?

vincepri:
yep, we discussed at yesterday's meeting and we opened master to v0.6

PR #832, opened 2020-02-27
:warning: Fakeclient: Reject Update with outdated ResourceVersion

alvaroaleman:
This PR changes the fakeclient to:
* Refuse Create requests with RV set
* Refuse Update requests with outdated RV

Note that while this brings the fakeclient more in line with a real client, it is a breaking change and has a good chance of breaking ppls tests.

/assign @vincepri

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -19,10 +19,12 @@ package fake
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -79,28 +81,42 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.O
...
-		if accessor.GetResourceVersion() == "" {
-			accessor.SetResourceVersion("1")
-		}
-	} else {
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
 		return err
 	}
+	if accessor.GetResourceVersion() != "" {
+		return kerrors.NewBadRequest("resourceVersion can not be set for Create requests")
```
alvaroaleman: I am not 100% sure if this is the correct error, can anybody confirm?

pkg/client/fake/client.go
```diff
@@ -19,10 +19,12 @@ package fake
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
```
djzager: nit: everywhere I've seen this package renamed in c-r it's been to `apierrors`.

```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

pkg/client/fake/client_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
```
djzager: nit

```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

pkg/client/fake/client.go
djzager: Am I understanding this correctly...This check is basically doing what the API server would do?

djzager: I found https://github.com/kubernetes/kubernetes/issues/71435 which makes me think this should be a [`NewInvalid` instead](https://godoc.org/k8s.io/apimachinery/pkg/api/errors#NewInvalid)

pkg/client/fake/client.go
```diff
@@ -79,28 +81,42 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.O
...
-	if accessor, err := meta.Accessor(obj); err == nil {
-		if accessor.GetResourceVersion() == "" {
-			accessor.SetResourceVersion("1")
-		}
-	} else {
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
 		return err
 	}
+	if accessor.GetResourceVersion() != "" {
```
djzager: I think a reasonable additional check on Create|Update would be:

```
	name := accessor.GetName()
	if len(name) == 0 {
		return fmt.Errorf("name is required")
	}
```

pkg/client/fake/client.go
alvaroaleman: Yes, exactly

djzager: turns out I can't read https://github.com/kubernetes/kubernetes/issues/71435#issuecomment-486688084 bad request is right.

djzager: turns out I can't read https://github.com/kubernetes/kubernetes/issues/71435#issuecomment-486688084 bad request is right.

pkg/client/fake/client.go
alvaroaleman: Not objecting to the idea, but I think it would be better in a dedicated PR so we have both changes appear in the changelog

vincepri:
Holding for v0.6

pkg/client/fake/client.go
```diff
@@ -79,28 +81,42 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.O
...
+	oldObject, err := t.ObjectTracker.Get(gvr, ns, accessor.GetName())
+	if err != nil {
+		return err
+	}
+	oldAccessor, err := meta.Accessor(oldObject)
+	if err != nil {
 		return err
 	}
+	if accessor.GetResourceVersion() != oldAccessor.GetResourceVersion() {
+		return apierrors.NewConflict(gvr.GroupResource(), accessor.GetName(), errors.New("object was modified"))
```
mengqiy: `NewResourceExpired` may be the right one.
https://godoc.org/k8s.io/apimachinery/pkg/api/errors#NewResourceExpired

pkg/client/fake/client.go
```diff
@@ -79,28 +81,42 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.O
...
+	}
+	if oldAccessor.GetResourceVersion() == "" {
+		oldAccessor.SetResourceVersion("0")
+	}
+	intResourceVersion, err := strconv.Atoi(oldAccessor.GetResourceVersion())
```
mengqiy: Probably better to use `unit64`.
https://github.com/kubernetes/apimachinery/blob/4147c925140e4e3aa51d6b46b41650821712d457/pkg/apis/meta/fuzzer/fuzzer.go#L186

pkg/client/fake/client.go
alvaroaleman: I don't think so, because this is the error message you see on conflicts: https://github.com/kubernetes/apimachinery/blob/6584f51ae935c6ffdc5abd5fc5cd1e640aab8a7f/pkg/api/errors/errors.go#L207

pkg/client/fake/client.go
vincepri: @djzager did you want to open an issue to track that change separately?

PR #834, opened 2020-02-28
🏃fetch_ext_bins now checks existing version before downloading

vincepri:
A slight improvement to the script that fetches kubebuilde-tools to avoid re-downloading the binaries if the existing version already matches our expectations.

### DIFF
--- a/hack/check-everything.sh
+++ b/hack/check-everything.sh
@@ -50,6 +50,15 @@ SKIP_FETCH_TOOLS=${SKIP_FETCH_TOOLS:-""}
 function fetch_kb_tools {
   local dest_dir="${1}"
 
+  # use the pre-existing version in the temporary folder if it matches our k8s version
+  if [[ -x "${dest_dir}/kubebuilder/bin/kube-apiserver" ]]; then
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cool.  Can you add a bit more of a description to the body (i.e. "this avoids extra downloads when unecessary) and remove the HTML comments from the PR description?  Just unhold when ready.

/lgtm
/hold

vincepri:
@DirectXMan12 done!

hack/check-everything.sh
```diff
@@ -50,6 +50,14 @@ SKIP_FETCH_TOOLS=${SKIP_FETCH_TOOLS:-""}
 function fetch_kb_tools {
   local dest_dir="${1}"
 
+  # use the pre-existing version in the temporary folder if it matches our k8s version
+  if [[ -x "${dest_dir}/kubebuilder/bin/kube-apiserver" ]]; then
+    version=$("${dest_dir}"/kubebuilder/bin/kube-apiserver --version)
+    if [[ $version == *"${k8s_version}"* ]]; then
+      return 0
```
alenkacz: should we write something to stdout in this case? That we're using already existing version

Issue #681, opened 2019-11-13
Incorporate API compatibility checks into CI


joelanford:
When reviewing PRs and deciding on a release version, it would be helpful to know if branches of `controller-runtime` contain any breaking changes:

- For PRs: the diff between `master` and the PR branch
- For releases: the diff between the previous semver release tag and `master`

For the recent bumps for Kubernetes [1.15](https://github.com/kubernetes-sigs/controller-runtime/pull/599#issuecomment-538556926) and [1.16](https://github.com/kubernetes-sigs/controller-runtime/pull/618#issuecomment-540450049), we used a tool I wrote ([go-apidiff](https://github.com/joelanford/go-apidiff)) based on Go's experimental [apidiff](https://github.com/golang/exp/tree/master/apidiff) package that checks:
- The backwards compatibility of controller-runtime itself.
- The backwards compatibility of controller-runtime's dependency changes.

In it's current state, the tool is still a proof-of-concept and probably needs improvements. I haven't had a ton of time to keep up with this, so I'm opening this up to the community to ask for help improving it, finding/fixing bugs, and to potentially incorporate it into controller-runtime's CI to give reviewers a signal of breaking public API changes.

joelanford:
From @vincepri:

> We're adding go-apidiff prow job in cluster api, we should do the same in controller-runtime at some point https://github.com/kubernetes-sigs/cluster-api/pull/2923

We can use this as an example for incorporating into controller-runtime. Thanks @vincepri!

alexeldeib:
I was looking at this yesterday, missed the open issue. If no one has taken it I can. 

/assign

PR #836, opened 2020-03-03
🏃 Bump dependencies for Kubernetes v1.18.0

### LINKED ISSUE #874: Update to v0.18 Kubernetes dependencies
Currently we are depending on 1.17.2 libraries. With 1.18, there are a lot of breaking changes in client-go that will block people from upgrading to newer client-go versions. It would be useful if this library is updated once 1.18 is officially published.

munnerz:
This PR bumps k8s.io dependencies to v0.18.0-rc.1.

The typed clientsets were changes in 1.18 to include a `context` field and to consistently use `UpdateOptions` and `CreateOptions` throughout the generated clients.

I've had to update a fair bit of test code for this, but it doesn't appear to have changed any user facing interfaces 😄 🎉

This is required in order for a downstream project to depend on v1.18.X dependencies as well as controller-runtime.

We may want to hold this until v1.18.0 final is released, but perhaps it's acceptable to get it into the master branch ahead of the next controller-runtime release.

/kind cleanup
/cc @DirectXMan12 @tamalsaha 

Fixes #874


vincepri:
/hold 

for community discussion

munnerz:
@vincepri is this to be discussed at a community meeting?

pkg/cache/internal/informers_map.go
```diff
@@ -245,20 +245,23 @@ func createStructuredListWatch(gvk schema.GroupVersionKind, ip *specificInformer
...
+	// TODO: the functions that make use of this ListWatch should be adapted to
+	//  pass in their own contexts instead of relying on this fixed one here.
+	ctx := context.TODO()
 	// Create a new ListWatch for the obj
 	return &cache.ListWatch{
 		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 			res := listObj.DeepCopyObject()
 			isNamespaceScoped := ip.namespace != "" && mapping.Scope.Name() != meta.RESTScopeNameRoot
-			err := client.Get().NamespaceIfScoped(ip.namespace, isNamespaceScoped).Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do().Into(res)
+			err := client.Get().NamespaceIfScoped(ip.namespace, isNamespaceScoped).Resource(mapping.Resource.Resource).VersionedParams(&opts, ip.paramCodec).Do(ctx).Into(res)
```
munnerz: Since this area of code didn't pass a context at all in the past, this is a net-zero change in behaviour 😄

pkg/internal/testing/integration/internal/integration_tests/integration_test.go
```diff
@@ -19,6 +20,7 @@ import (
 
 var _ = Describe("The Testing Framework", func() {
 	var controlPlane *integration.ControlPlane
+	ctx := context.TODO()
```
munnerz: I opted to initially introduce contexts into test code in the most naive way possible, denoting it with `context.TODO()` so go linting tools can flag this as needing doing and it won't be forgotten.

pkg/recorder/example_test.go
```diff
@@ -48,13 +45,3 @@ func Example_eventf() {
 	recorder.Eventf(somePod, corev1.EventTypeNormal,
 		"DislikesCheese", "Not even %s?", mildCheese)
 }
-
-func Example_pastEventf() {
-	// recorderProvider is a recorder.Provider
-	recorder := recorderProvider.GetEventRecorderFor("my-controller")
-
-	// emit a backdated event (potentially with variable message)
-	recorder.PastEventf(somePod, metav1.Time{Time: time.Now().Add(-5 * time.Minute)},
```
munnerz: PastEventf seems to have been removed from the recorder interface

vincepri:
@munnerz Yeah, at the next scheduled community meeting. Does that sound good to you?

munnerz:
👋 unfortunately I had a scheduling conflict for the last meeting - it looks like the next one is April 8th?

Given the 1.18 release is imminent, can we begin discussion on the concerns around this PR here so we have something concrete to discuss/to attempt to asynchronously resolve this?

From an 'exposed interfaces' PoV, this PR does not introduce any breaking changes to controller-runtime's own interfaces/structures. Is there a specific concern you have in mind? We've previously bumped the `k8s.io/*` dependencies in controller-runtime without too much concern, and though this one does involve some further reaching changes (adding `context.Context` to typed clients), controller-runtime users utilise `client.Client` which means they will be largely unaffected (and `client-go`'s own versioning shenanigans are IMO 'out of scope' for controller-runtime to handle (as there's not really anything controller-runtime _can_ do!))

I'll get this PR rebased anyway so we have something 'current' to discuss :)

liggitt:
> Given the 1.18 release is imminent, can we begin discussion on the concerns around this PR here so we have something concrete to discuss/to attempt to asynchronously resolve this?

Agree. controller-runtime is a transitive dependency of many projects, so getting it onto v1.18 libraries is important.

vincepri:
we can open master to v0.6.0 and merge this in

aledbf:
Any update on this?

liggitt:
I expect this will be updated after v0.18.0 tags are pushed

munnerz:
Once the v0.18.0 tags are available in the respective repositories I'll update both this and https://github.com/kubernetes-sigs/controller-tools/pull/424 😄

munnerz:
I've updated this PR to use the `v0.18.0` tags now that they have been released. I think this should be good to go now (assuming it passes CI!) 😄 

/assign @vincepri

vincepri:
@munnerz if we haven't already, we should run the breaking change checker so we can put them in releases

munnerz:
I’ve not run that, no - could you provider a pointer to it and I’ll post the results here? 😄

alexeldeib:
I think it's https://github.com/joelanford/go-apidiff

ref also: https://github.com/kubernetes-sigs/controller-runtime/issues/681

howardjohn:
```
$ go-apidiff v0.5.2 fetch-836

sigs.k8s.io/controller-runtime/pkg/leaderelection/fake
  Incompatible changes:
  - (*ResourceLock).Create: changed from func(k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord) error to func(context.Context, k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord) error
  - (*ResourceLock).Get: changed from func() (*k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord, []byte, error) to func(context.Context) (*k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord, []byte, error)
  - (*ResourceLock).Update: changed from func(k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord) error to func(context.Context, k8s.io/client-go/tools/leaderelection/resourcelock.LeaderElectionRecord) error

```

pkg/cache/internal/informers_map.go
alvaroaleman: It feels like this is something we should fix, now that we do a breaking release anyways. From a quick look this seems to not be a super trivial change.

Issue #831, opened 2020-02-27
Add DryRun option to the client


alvaroaleman:
In order to safely check on what a controller does in a production environment, it would be useful to have a `DryRun` option for the client that makes all mutating operations (`Update`, `Patch` and `Delete`) a no-op and just logs them.

/kind feature

PR #839, opened 2020-03-05
✨ Add DryRunClient wrapper

alvaroaleman:
This PR adds a DryRunClient wrapper that enforced DryRun on all mutating operations and plugs that into the client and manager constructors. This is useful to check what a controller does in a production environment without actually doing it.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/831

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -33,9 +33,9 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	kscheme "k8s.io/client-go/kubernetes/scheme"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const serverSideTimeoutSeconds = 10
--- a/pkg/client/dryrun_test.go
+++ b/pkg/client/dryrun_test.go
@@ -1,0 +1,264 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client_test
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("DryRunClient", func() {
+	var dep *appsv1.Deployment
+	var count uint64 = 0
+	var replicaCount int32 = 2
+	var ns = "default"
+	ctx := context.Background()
+
+	getClient := func() client.Client {
+		nonDryRunClient, err := client.New(cfg, client.Options{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(nonDryRunClient).NotTo(BeNil())
+		return client.NewDryRunClient(nonDryRunClient)
+	}
+
+	BeforeEach(func() {
+		atomic.AddUint64(&count, 1)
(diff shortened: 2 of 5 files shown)


pkg/client/client_test.go
```diff
@@ -33,9 +33,9 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	kscheme "k8s.io/client-go/kubernetes/scheme"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
djzager: Only curious because it looks intentional and not the kind of thing I would expect `gofmt` to do. Why this change?

pkg/client/dryrun_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+		opts := &client.PatchOptions{DryRun: []string{"Bye", "Pippa"}}
+
+		Expect(getClient().Status().Patch(ctx, changedDep, client.MergeFrom(dep), opts)).ToNot(HaveOccurred())
+
+		actual, err := clientset.AppsV1().Deployments(ns).Get(dep.Name, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(actual).NotTo(BeNil())
+		Expect(actual).To(BeEquivalentTo(dep))
+	})
```
djzager: Reading https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/ I see that one of the advantages of supporting a dry-run, besides not being persisted, is

> the request is still processed as typical request: the fields are defaulted, the object is validated, it goes through the validation admission chain, and through the mutating admission chain, and then the final object is returned to the user as it normally would, without being persisted.

I think a test case where a create or an update failed validation would be a good addition.

pkg/client/client_test.go
alvaroaleman: Because we usually put imports form our own repo in a dedicated group, rather than mixing them with other imports in a common group

pkg/client/dryrun_test.go
```diff
@@ -0,0 +1,266 @@
...
+		Expect(apierrors.IsNotFound(err)).To(BeTrue())
+	})
+
+	It("should refuse a create request for an invalid object", func() {
+		changedDep := dep.DeepCopy()
+		changedDep.Spec.Template.Spec.Containers = nil
+
+		err := getClient().Create(ctx, changedDep)
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
```
djzager: nit for consistency.

```suggestion
```

pkg/client/dryrun_test.go
```diff
@@ -0,0 +1,266 @@
...
+		Expect(actual).To(BeEquivalentTo(dep))
+	})
+
+	It("should refuse an update request for an invalid change", func() {
+		changedDep := dep.DeepCopy()
+		changedDep.Spec.Template.Spec.Containers = nil
+
+		err := getClient().Update(ctx, changedDep)
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
```
djzager: nit for consistency.

```suggestion
```

pkg/client/client.go
```diff
@@ -37,6 +37,9 @@ type Options struct {
 
 	// Mapper, if provided, will be used to map GroupVersionKinds to Resources
 	Mapper meta.RESTMapper
+
+	// DryRun specifies whether dry run mode should be enforced
+	DryRun bool
```
DirectXMan12: What's the main usecase for having this as an option here vs just letting people call `NewDryRunClient`?  Just want to avoid too many options here, esp since this isn't strictly related to creating the client

alvaroaleman: Good point, its not really necessary there, I removed it.

I would like to keep it in the `manager` though, because the `defaultNewClient` is not exported and replacing the `NewClient` func is not very intuitive and an option ppl that are not deeply familiar with c-r will most likely not become aware of. WDYT?

alvaroaleman:
/assign @DirectXMan12 
Could you have another look, please?

alvaroaleman:
@DirectXMan12 any chance you can have a look?

pkg/client/client.go
mengqiy: Not super excited about growing manager's Options.

One alternative we can try here is that:
Making `defaultNewClient` exportable, but under `pkg/manager/internal` or `pkg/internal/manager` or some place similar.
Then you can have a helper method that satisfy `NewClientFunc`
```
func NewDryRunClientFn(cache cache.Cache, config *rest.Config, options client.Options) (client.Client, error) {
  cl, err := managerinternal.DefaultNewClient(cache, config, options) 
  return &dryRunClient{cl}, err
}
```
User can pass this method to manager Options.

@DirectXMan12 Thoughts?

alvaroaleman: @mengqiy the reason I dislike that is that its pretty hard to discover that these kind of build-in opts exists

DirectXMan12:
hey, sorry for the lag

/lgtm
/approve

alvaroaleman:
@DirectXMan12 could you give a new lgtm as I had to rebase?

DirectXMan12:
FWIW, I'd love to eventually figure out how to make it easy to sub in the dryrun client instead of needing an option on manager, but I don't really think I want to block on that now

Issue #835, opened 2020-03-01
scheme_test doesn't check for the existence of CreateOptions


Capstan:
Although it exists as a base type alongside UpdateOptions, DeleteOptions, etc., it is not explicitly tested for. Because the test only tests length of the returned list and existence of specific entries rather than a concrete set of expected entries, it seems to have been left out.

alvaroaleman:
Fixed in https://github.com/kubernetes-sigs/controller-runtime/pull/842
/close

PR #842, opened 2020-03-06
🏃 Use whole list matching for known types.

Capstan:
The scheme test knows how many types should be returned, but omitted `CreateOptions` from being explicitly checked.

### DIFF
--- a/pkg/scheme/scheme_test.go
+++ b/pkg/scheme/scheme_test.go
@@ -21,6 +21,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


Capstan:
I think this whole set could be better written using [`MatchAllKeys`](https://godoc.org/github.com/onsi/gomega/gstruct#MatchAllKeys), a la

```
Expect(s.AllKnownTypes()).To(MatchAllKeys(Keys{
    gv.WithKind("Pod"): Equal(reflect.TypeOf(corev1.Pod{}),
    …
    gv.WithKind("CreateOptions"): Ignore(),
    …
}))
```

The length check would go away, being implicit to the exact match of keys. The values that aren't inspected would be `Ignore()`d.

If you prefer I do this approach I can either replace this or commit and follow-on.

mengqiy:
I seems using `MatchAllKeys` is better. Can you please change it in this PR?
But I feel we still need to keep the length checking, since it can give us signals when upstream introduce more stuffs in the scheme and we can make appropriate changes.
e.g. https://github.com/kubernetes-sigs/controller-runtime/commit/71cf6f6cf312e9111fc0ff1b411e19b91726359d#diff-b8b47f9ed3f5e8f94b54d518a7693971

Capstan:
`MatchAllKeys` does length checking implicitly, as the set of keys must exactly match that which is provided.

Capstan:
@mengqiy PTAL

Capstan:
Addresses #835.

Capstan:
Still need @mengqiy or another reviewer to review and lgtm.

Capstan:
Is there some other action people are waiting for me to perform?

PR #844, opened 2020-03-09
🏃Rename EnsureOwnerReference to SetOwnerReference for consistency

vincepri:
To be consistent with `SetControllerReference`, rename EnsureOwnerReference to SetOwnerReference.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -89,10 +89,10 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
-// EnsureOwnerReference is a helper method to make sure the given object contains
-// an object reference to the object provided.
-// If a reference already exists, it'll be overwritten with the newly provided version.
-func EnsureOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
+// SetOwnerReference is a helper method to make sure the given object contains an object reference to the object provided.
+// This allows you to declare that owner has a dependency on the object without specifying it as a controller.
+// If a reference to the same object already exists, it'll be overwritten with the newly provided version.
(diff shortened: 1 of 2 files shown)


vincepri:
Given that we haven't released v0.5.1 yet, we can consider this as non-breaking in the current cycle

alvaroaleman:
Why do we actually have these two, does the controller boolean make any difference at all?

vincepri:
In some cases you might want multiple OwnerReferences, only one can be set (when using these or metav1 utilities). Multiple OwnerRefs can be useful when you want to create a tree of dependencies between objects

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -89,10 +89,10 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
-// EnsureOwnerReference is a helper method to make sure the given object contains
+// SetOwnerReference is a helper method to make sure the given object contains
 // an object reference to the object provided.
-// If a reference already exists, it'll be overwritten with the newly provided version.
-func EnsureOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
+// If a reference to the same object already exists, it'll be overwritten with the newly provided version.
```
djzager: With this change, maybe it would be good to explain in the comments why you would use `SetOwnerReference` over `SetControllerReference`.

vincepri: What would you like that to look like?

djzager: :scream: words escape me. Here is my best attempt:

> This allows you to declare that owner has a dependency on the object without specifying it as a controller.

DirectXMan12:
/hold cancel

as per above approval

PR #848, opened 2020-03-10
⚠ (:warning:, major) Do not disable admission plugins by default

alenkacz:
If we enable ServiceAccount plugin, tests fail with this error. That is because controller normally takes care of creating default service account but that controller is not running in the integration environment.

```
 "pods \"test-pod-1\" is forbidden: error looking up service account test-namespace-1/default: serviceaccount \"default\" not found",
```

We either keep that disabled by default or envtest would have to create that SA for the users? Otherwise all users would have to create that SA which I think is not desired.

### DIFF
--- a/pkg/internal/testing/integration/internal/apiserver.go
+++ b/pkg/internal/testing/integration/internal/apiserver.go
@@ -9,7 +9,9 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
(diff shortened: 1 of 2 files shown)


pkg/internal/testing/integration/internal/apiserver.go
```diff
@@ -9,7 +9,7 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--disable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,TaintNodesByCondition,Priority,DefaultTolerationSeconds,DefaultStorageClass,StorageObjectInUseProtection,PersistentVolumeClaimResize,ResourceQuota", //nolint
+	"--disable-admission-plugins=ServiceAccount", //nolint
```
DirectXMan12: can we remove the `// nolint`?

alenkacz: yes! :) thank you

alenkacz:
@DirectXMan12 fixed, I also added comment

PR #850, opened 2020-03-10
✨CreateOrPatch

akutz:
This patch introduces a variation on the `controllerutil.CreateOrUpdate` function named `CreateOrPatch`. Instead of issuing update calls, the new function uses a patch to perform a more surgical update to the remote data. Additionally, the implementation relies on logic similar to the [`patch.Helper`](https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go) mechanics from Cluster API. The resource is converted to unstructured data first in order to patch the resource and any potential status separately.

Two new OperationResult values were added:

1. `OperationResultUpdatedStatus` - the resource and its status were updated
2. `OperationResultUpdatedStatusOnly` - only the status was updated

There are tests implemented for `CreateOrPatch` that follow those implemented for `CreateOrUpdate` with the addition of two, new tests that validate the new result possibilities.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -19,11 +19,13 @@ package controllerutil
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/pointer"
@@ -180,6 +182,10 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 	OperationResultCreated OperationResult = "created"
 	// OperationResultUpdated means that an existing resource is updated
 	OperationResultUpdated OperationResult = "updated"
+	// OperationResultUpdatedStatus means that an existing resource and its status is updated
+	OperationResultUpdatedStatus OperationResult = "updatedStatus"
+	// OperationResultUpdatedStatusOnly means that only an existing status is updated
+	OperationResultUpdatedStatusOnly OperationResult = "updatedStatusOnly"
 )
 
 // CreateOrUpdate creates or updates the given object in the Kubernetes
@@ -221,6 +227,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f 
 		return OperationResultNone, err
 	}
 	return OperationResultUpdated, nil
+}
+
+// CreateOrPatch creates or patches the given object in the Kubernetes
+// cluster. The object's desired state must be reconciled with the before
+// state inside the passed in callback MutateFn.
+//
+// The MutateFn is called regardless of creating or updating an object.
+//
+// It returns the executed operation and an error.
+func CreateOrPatch(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
+	key, err := client.ObjectKeyFromObject(obj)
+	if err != nil {
+		return OperationResultNone, err
+	}
+
+	if err := c.Get(ctx, key, obj); err != nil {
+		if !errors.IsNotFound(err) {
+			return OperationResultNone, err
+		}
+		if f != nil {
+			if err := mutate(f, key, obj); err != nil {
+				return OperationResultNone, err
+			}
+		}
+		if err := c.Create(ctx, obj); err != nil {
+			return OperationResultNone, err
+		}
+		return OperationResultCreated, nil
+	}
+
+	// Create patches for the object and its possible status.
+	objPatch := client.MergeFrom(obj.DeepCopyObject())
+	statusPatch := client.MergeFrom(obj.DeepCopyObject())
+
+	// Create a copy of the original object as well as converting that copy to
+	// unstructured data.
+	before, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
+	if err != nil {
+		return OperationResultNone, err
+	}
+
+	// Attempt to extract the status from the resource for easier comparison later
+	beforeStatus, hasBeforeStatus, err := unstructured.NestedFieldCopy(before, "status")
+	if err != nil {
+		return OperationResultNone, err
+	}
+
+	// If the resource contains a status then remove it from the unstructured
+	// copy to avoid unnecessary patching later.
+	if hasBeforeStatus {
+		unstructured.RemoveNestedField(before, "status")
+	}
+
+	// Mutate the original object.
+	if f != nil {
+		if err := mutate(f, key, obj); err != nil {
+			return OperationResultNone, err
+		}
+	}
+
+	// Convert the resource to unstructured to compare against our before copy.
+	after, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return OperationResultNone, err
+	}
+
+	// Attempt to extract the status from the resource for easier comparison later
+	afterStatus, hasAfterStatus, err := unstructured.NestedFieldCopy(after, "status")
+	if err != nil {
+		return OperationResultNone, err
+	}
+
+	// If the resource contains a status then remove it from the unstructured
+	// copy to avoid unnecessary patching later.
+	if hasAfterStatus {
+		unstructured.RemoveNestedField(after, "status")
+	}
+
+	result := OperationResultNone
+
+	if !reflect.DeepEqual(before, after) {
+		// Only issue a Patch if the before and after resources (minus status) differ
+		if err := c.Patch(ctx, obj, objPatch); err != nil {
+			return result, err
+		}
+		result = OperationResultUpdated
+	}
+
+	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
+		// Only issue a Status Patch if the resource has a status and the beforeStatus
+		// and afterStatus copies differ
+		if err := c.Status().Patch(ctx, obj, statusPatch); err != nil {
+			return result, err
+		}
+		if result == OperationResultUpdated {
+			result = OperationResultUpdatedStatus
+		} else {
+			result = OperationResultUpdatedStatusOnly
+		}
+	}
+
+	return result, nil
 }
 
 // mutate wraps a MutateFn and applies validation to its result
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -406,6 +406,195 @@ var _ = Describe("Controllerutil", func() {
 		})
 	})
 
+	Describe("CreateOrPatch", func() {
+		var deploy *appsv1.Deployment
+		var deplSpec appsv1.DeploymentSpec
+		var deplKey types.NamespacedName
+		var specr controllerutil.MutateFn
+
+		BeforeEach(func() {
+			deploy = &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("deploy-%d", rand.Int31()),
+					Namespace: "default",
+				},
+			}
+
+			deplSpec = appsv1.DeploymentSpec{
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"foo": "bar"},
+				},
+				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"foo": "bar",
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  "busybox",
+								Image: "busybox",
+							},
+						},
+					},
+				},
+			}
+
+			deplKey = types.NamespacedName{
+				Name:      deploy.Name,
+				Namespace: deploy.Namespace,
+			}
+
+			specr = deploymentSpecr(deploy, deplSpec)
+		})
+
+		It("creates a new object if one doesn't exists", func() {
+			op, err := controllerutil.CreateOrPatch(context.TODO(), c, deploy, specr)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning OperationResultCreated")
+			Expect(op).To(BeEquivalentTo(controllerutil.OperationResultCreated))
+
+			By("actually having the deployment created")
+			fetched := &appsv1.Deployment{}
+			Expect(c.Get(context.TODO(), deplKey, fetched)).To(Succeed())
+
+			By("being mutated by MutateFn")
+			Expect(fetched.Spec.Template.Spec.Containers).To(HaveLen(1))
+			Expect(fetched.Spec.Template.Spec.Containers[0].Name).To(Equal(deplSpec.Template.Spec.Containers[0].Name))
+			Expect(fetched.Spec.Template.Spec.Containers[0].Image).To(Equal(deplSpec.Template.Spec.Containers[0].Image))
+		})
+
+		It("patches existing object", func() {
+			var scale int32 = 2
+			op, err := controllerutil.CreateOrPatch(context.TODO(), c, deploy, specr)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(op).To(BeEquivalentTo(controllerutil.OperationResultCreated))
+
+			op, err = controllerutil.CreateOrPatch(context.TODO(), c, deploy, deploymentScaler(deploy, scale))
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning OperationResultUpdated")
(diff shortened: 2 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
+
+	// If the resource contains a status then remove it from the unstructured
+	// copy to avoid unnecessary patching later.
+	if hasAfterStatus {
+		unstructured.RemoveNestedField(after, "status")
+	}
+
+	result := OperationResultNone
+
+	if !reflect.DeepEqual(before, after) {
```
alexeldeib: nit: equality.Semantic.DeepEqual from apimachinery

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
+		// Only issue a Patch if the before and after resources (minus status) differ
+		if err := c.Patch(ctx, obj, objPatch); err != nil {
+			return result, err
+		}
+		result = OperationResultUpdated
+	}
+
+	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
```
alexeldeib: equality.Semantic.DeepEqual here too I think

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -180,6 +182,10 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 	OperationResultCreated OperationResult = "created"
 	// OperationResultUpdated means that an existing resource is updated
 	OperationResultUpdated OperationResult = "updated"
+	// OperationResultUpdatedStatus means that an existing resource and its status is updated
+	OperationResultUpdatedStatus OperationResult = "updatedStatus"
```
alexeldeib: These are currently only used for patch operations. The naming may be confusing since an update implies a full replace, while a patch does not. Consider `OperationResultPatched` + `OperationResultPatchedStatus`?

alexeldeib: I guess it doesn't really matter since we're diffing the object after mutation and deriving the patch based on that, though.

alexeldeib:
Separating out the status diverges from the existing CreateOrUpdate implementation. I'm not opposed to differentiating spec/status updates, but I think it would be cleaner to be consistent. 

WDYT? Do you think it makes sense to add that logic in CreateOrUpdate too?

pkg/controller/controllerutil/controllerutil.go
akutz: Hi @alexeldeib,

> These are currently only used for patch operations. The naming may be confusing since an update sets the whole object, while a patch does not. Consider `OperationResultPatched `+ `OperationResultPatchedStatus`?

The reason for the current approach is to be consistent with the existing constants.

pkg/controller/controllerutil/controllerutil.go
akutz: Hi @alexeldeib,

I am not quite sure what you mean. I copied this [directly](https://github.com/kubernetes-sigs/cluster-api/blob/f11383ea28f46e10c7738e160aa4ec25b4b97dca/util/patch/patch.go#L118) from Cluster API.

Maybe @vincepri and @detiber can help explain the difference as I believe they were responsible for the Cluster API version.

pkg/controller/controllerutil/controllerutil.go
akutz: Hi @alexeldeib,

I also copied this [directly](https://github.com/kubernetes-sigs/cluster-api/blob/f11383ea28f46e10c7738e160aa4ec25b4b97dca/util/patch/patch.go#L125) from Cluster API.

Once again I will defer to @vincepri and @detiber.

akutz:
> Separating out the status diverges from the existing CreateOrUpdate implementation. I'm not opposed to differentiating spec/status updates, but I think it would be cleaner to be consistent.
> 
> WDYT? Do you think it makes sense to add that logic in CreateOrUpdate too?

I am going to ping @detiber and @vincepri to help add context on the reason for the current approach, but if I recall correctly, it is because we all ran into trouble last year when issuing `Patch` against a resource whose status was not yet present. That is why we ended up removing the status pre-patch and then adding it back for the secondary status patch call. 

I remember prior to unstructured data literally setting the `.status = Status{}` (or whatever the resource's status struct was) to zero it out. Ah, [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a9573d1ac52665c6523e118055b9e7c5f6265da2/pkg/cloud/vsphere/context/cluster_context.go#L311-L314) is what I mean:

```golang
	// Make sure the status isn't part of the JSON patch.
	newStatus := c.Cluster.Status.DeepCopy()
	c.Cluster.Status = clusterv1.ClusterStatus{}
	c.ClusterCopy.Status.DeepCopyInto(&c.Cluster.Status)
```

Anyway, that is why the status is handled the way it is.

pkg/controller/controllerutil/controllerutil.go
alexeldeib: see https://github.com/kubernetes-sigs/controller-runtime/issues/464#issue-451392663

akutz: Hi @alexeldeib,

Perhaps it is more accurate when operating against a map instead of nested structs? I'll still defer to @vincepri and @detiber on whether they have found the same and think we should change things. I am not disagreeing with you, but rather that this means we need to update Cluster API as well :)

pkg/controller/controllerutil/controllerutil.go
alexeldeib: what's the difference between `OperationResultUpdatedStatus` and `OperationResultUpdated`? does it make sense to expose this distinction in `CreateOrUpdate` as well? If not, does it make sense to remove one?

akutz: > what's the difference between `OperationResultUpdatedStatus` and `OperationResultUpdated`? does it make sense to expose this distinction in `CreateOrUpdate` as well? If not, does it make sense to remove one?

Hi @alexeldeib,

I suppose if we do not create `Patched` variants, then there is no real distinction between `OperationResultUpdated` and `OperationResultUpdatedStatus`. Good catch. Here are my general thoughts on the matter:

* To me `Updated` means the result, not the method used to achieve it, another reason I did not create the variants
* It was important to me to have a result that indicated the status sub-resource was updated, but not the primary resource itself.

Beyond the above, I am really fine doing whatever here.

alexeldeib: my initial reaction was because of a race condition like this:

thread a - createOrUpdate
1. get object
2. mutate and extract patch
3. send patch to API server

thread b - racer
1. get obj
2. mutate 
3. send to apiserver

semantically I would want to tell the user their patch was successfully applied but I cannot guarantee that it was applied against the same object they provided. This is a slightly less strong guarantee than `OperationResultUpdated`, so I would have matched the HTTP methods and reached for `OperationResultPatched`. 

To me `OperationResultUpdatedStatus` means what I think `OperationResultUpdatedStatusOnly` is supposed to mean -- that a patch was applied but only the status was updated. Neither of them actually means an update operation was what was executed, vs the naming of `OperationResultCreated`, `OperationResultUpdated`

pkg/controller/controllerutil/controllerutil.go
alexeldeib: sorry, I see these are on the unstructured not structured forms.

pkg/controller/controllerutil/controllerutil.go
alexeldeib: ignore, same as above

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
 }
 
+// CreateOrPatch creates or patches the given object in the Kubernetes
+// cluster. The object's desired state must be reconciled with the before
+// state inside the passed in callback MutateFn.
+//
+// The MutateFn is called regardless of creating or updating an object.
+//
+// It returns the executed operation and an error.
+func CreateOrPatch(ctx context.Context, c client.Client, obj runtime.Object, f MutateFn) (OperationResult, error) {
```
negz: https://github.com/crossplane/crossplane-runtime/blob/c092201a/pkg/resource/resource.go#L274

We have a similar function that we use in various Crossplane controllers. We've found that passing a function that takes the existing and the desired resource can be useful, for example to ensure that the existing resource has the same controller reference as the desired resource before we mutate it.

vincepri:
@alexeldeib @akutz What's the status of this PR? Regarding having Status patched, it's a really nice to have, not sure if it's worth to do it separately though.

akutz:
> @alexeldeib @akutz What's the status of this PR? Regarding having Status patched, it's a really nice to have, not sure if it's worth to do it separately though.

Hi @vincepri,

As far as I am concerned this PR is fine as-is. There was a healthy discussion above, but nothing that resulted in me wanting to make any major changes to the PR. I do agree that perhaps the operation results could be cleaned up, but I am also unsure to what. Do you have any feedback here?

akutz:
> Hi @alexeldeib,
> 
> First, thank you for the review. I appreciate it :)
> 
> > Separating out the status diverges from the existing CreateOrUpdate implementation. I'm not opposed to differentiating spec/status updates, but I think it would be cleaner to be consistent.
> > WDYT? Do you think it makes sense to add that logic in CreateOrUpdate too?
> 
> I am going to ping @detiber and @vincepri to help add context on the reason for the current approach, but if I recall correctly, it is because we all ran into trouble last year when issuing `Patch` against a resource whose status was not yet present. That is why we ended up removing the status pre-patch and then adding it back for the secondary status patch call.
> 
> I remember prior to unstructured data literally setting the `.status = Status{}` (or whatever the resource's status struct was) to zero it out. Ah, [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a9573d1ac52665c6523e118055b9e7c5f6265da2/pkg/cloud/vsphere/context/cluster_context.go#L311-L314) is what I mean:
> 
> ```go
> 	// Make sure the status isn't part of the JSON patch.
> 	newStatus := c.Cluster.Status.DeepCopy()
> 	c.Cluster.Status = clusterv1.ClusterStatus{}
> 	c.ClusterCopy.Status.DeepCopyInto(&c.Cluster.Status)
> ```
> 
> Anyway, that is why the status is handled the way it is.

Hi @vincepri,

I had pinged both you and @detiber for your feedback on March 14th. Any thoughts on this? I remember the reason we separated the status was because at the time the `Patch` operation could not patch the status with the primary resource as it was a sub-resource. IMO to separate it into a distinct function is orthogonal to the behavior of `CreateOrUpdate`.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
+	}
+
+	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
+		// Only issue a Status Patch if the resource has a status and the beforeStatus
+		// and afterStatus copies differ
+		if err := c.Status().Patch(ctx, obj, statusPatch); err != nil {
+			return result, err
+		}
+		if result == OperationResultUpdated {
+			result = OperationResultUpdatedStatus
```
alexeldeib, line 325: @vincepri fyi since this was one we wanted more 👀 on 

IMO my comment in https://github.com/kubernetes-sigs/controller-runtime/pull/850/files#r392628297 is still accurate, either `OperationResultUpdatedStatus` should be removed or there should be real differentiation between `OperationResultUpdatedStatus` and `OperationResultUpdated`.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
+		} else {
+			result = OperationResultUpdatedStatusOnly
+		}
+	}
+
+	return result, nil
```
alexeldeib, line 331: for illustrative purposes:

```suggestion
	result := OperationResultNone

	if !reflect.DeepEqual(before, after) {
		// Only issue a Patch if the before and after resources (minus status) differ
		if err := c.Patch(ctx, obj, objPatch); err != nil {
			return result, err
		}
		result = OperationResultUpdatedSpec
	}

	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
		// Only issue a Status Patch if the resource has a status and the beforeStatus
		// and afterStatus copies differ
		if err := c.Status().Patch(ctx, obj, statusPatch); err != nil {
			return result, err
		}
		if result == OperationResultUpdatedSpec {
			result = OperationResultUpdated
		} else {
			result = OperationResultUpdatedStatus
		}
	}

	return result, nil
```

alexeldeib:
If no one else is bothered by the naming, it's fine with me. I don't think users should depend on the return values here anyway, but if we provide it publicly we should make it as correct as possible.

akutz:
> If no one else is bothered by the naming, it's fine with me. I don't think users should depend on the return values here anyway, but if we provide it publicly we should make it as correct as possible.

Hi @alexeldeib,

I am *totally* fine changing the naming. I pinged Vince because of the order logic question. If you would like to suggest the preferred naming, I'll update the PR.

alexeldeib:
My mistake, scanned through the existing comments quickly. Agree with your comment as well though, would be nice to have clarity :)

pkg/controller/controllerutil/controllerutil.go
vincepri: Yeah that makes sense, I'd suggest to just keep one for now and revisit in a future version. +1 removing the updated status one

akutz:
Hi @alexeldeib,

Is there a time when we could hop on a Zoom and discuss the result values in a high-bandwidth medium? I want to unblock this PR, and I am still unsure what to do about those return values. Thanks!

alexeldeib:
TBH, I'm inclined to merge as-is and fix forward. I think this is useful functionality and my only objections are nits. For the CreateOrUpdate case, we can add the same differentiation if we find it's actuall useful in practice.

@alvaroaleman @DirectXMan12 any thoughts from either of you here?

adding lgtm based on ^ leaving hold for one of those fine folks. i'll be free this afternoon after 2ish and tomorrow in the middle of the day from 11:30-1ish if you still want to chat about it, but I'd rather get us unstuck

/lgtm

akutz:
Hi @alexeldeib,

If everyone is happy with this as-is, then I'm more than happy to remove the hold (since I am the one that added it). It already has two lgtms, so I'm not sure we need to wait any longer if the plan is to iterate anyway.

pkg/controller/controllerutil/controllerutil.go
alvaroaleman: Uhm what if both the object and the status were updated? Then we basically need to return two results or not?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -223,6 +229,108 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj runtime.Object, f
...
+		if err := c.Patch(ctx, obj, objPatch); err != nil {
+			return result, err
+		}
+		result = OperationResultUpdated
+	}
+
+	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
+		// Only issue a Status Patch if the resource has a status and the beforeStatus
+		// and afterStatus copies differ
+		if err := c.Status().Patch(ctx, obj, statusPatch); err != nil {
```
alvaroaleman, line 321: This will always fail if the status subresource is not enabled or not?

akutz: So update to gracefully handle the resources that lack a status sub resource?

pkg/controller/controllerutil/controllerutil.go
akutz: If both were updated, then `OperationResultUpdated` would imply both were updated. 

Again, I didn't change the names of the results to `Patched` because I was trying to preserve compatibility with the existing result values. I'm happy to adjust the names.

pkg/controller/controllerutil/controllerutil.go
alvaroaleman: > So update to gracefully handle the resources that lack a status sub resource?

Sorry I don't understand what you mean by this comment?

We could probably handle this by just optimistically updating and handling the error we get in case its not enabled (I believe its a 404 but might be wrong)?

vincepri: Are you referring to the if condition above?

pkg/controller/controllerutil/controllerutil.go
alvaroaleman: But the fact that the `CreateOrPatch` tries to update the `Status` whereas `CreateOrUpdate` does not is just an implementation detail and not inherent to doing patch vs update or am I missing something?

May I ask for the reason you need to differentiate if the status resource was updated? I am asking because I believe that whatever we do here should also be added to `CreateOrUpdate` and I know ppl use the result sometimes to determine how to proceed. If we extend this we might break them which is why maybe returning a `[]result` may be better (but nor completely sure).

vincepri:
I'm fine merging this as-is 👍 and improve the comments above later

akutz:
Sgtm.

PR #853, opened 2020-03-12
:running: Allow specifying custom TLS Config for webhooks

alastor-erinyes:
* This implements a Manager option WebhookTLSConfig which allows for
  specifying a custom TLS configuration when running webhooks using
  controller-runtime
* Added CustomTLSConfig attribute to WebhookInstallOptions
* Added test to show this config overrides the config generated by
  CertDir

### DIFF
--- a/pkg/envtest/webhook.go
+++ b/pkg/envtest/webhook.go
@@ -15,6 +15,7 @@ package envtest
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
@@ -63,6 +64,11 @@ type WebhookInstallOptions struct {
 	// LocalServingCertDir is the allocated directory for serving certificates.
 	// it will be automatically populated by the local temp dir
 	LocalServingCertDir string
+
+	// CustomTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
+	CustomTLSConfig *tls.Config
 
 	// MaxTime is the max time to wait
 	MaxTime time.Duration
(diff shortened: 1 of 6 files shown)


pkg/manager/manager.go
```diff
@@ -307,6 +313,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		port:                  options.Port,
 		host:                  options.Host,
 		certDir:               options.CertDir,
+		webhookTLSConfig:      options.WebhookTLSConfig,
```
alvaroaleman: What is your usecase for needing this and not being able to use the `certDir`?

alastor-erinyes: We have an internal library that we're required to call which produces a `tls.Config` that complies to a company standard.  Rather than going through a song-and-dance trying to convert that to files which would just be converted back to a `tls.Config` I'd prefer to just use the config directly.

pytimer: @alvaroaleman I want to setting custom TLS config, such as CipherSuites, so i think this feature is helpful for me. Or have other way can do it?

pkg/envtest/webhook.go
```diff
@@ -64,6 +65,11 @@ type WebhookInstallOptions struct {
 	// it will be automatically populated by the local temp dir
 	LocalServingCertDir string
 
+	// CustomTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
+	CustomTLSConfig *tls.Config
```
alvaroaleman: Can we change this here and in the manager to return an error when both `CertDir` and `CustomTLSConfig` are specified to avoid hard to debug issues?

varshaprasad96:
Hi @alastor-erinyes,

Thank you so much for contributing this feature. This is very useful in solving the issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1135. It would be really helpful to have this merged. Would you like to reopen this PR and work on this further. Or if you are busy or have time constraints, can I clone this PR and move this forward with your inputs.

PR #854, opened 2020-03-12
Documenting compatibility matrix between controller-runtime and kubernetes

yue9944882:
/cc @DirectXMan12 @pwittrock 

this pull adds a compatibility matrix to the README page so that users wont stumble on matching the releases w/ kubernetes dependencies..

i got the matrix by running unit-tests w/ the variety of kubernetes dependencies. i.e. k8s.io/{apiserver,apimachinery,api,client-go}. by `NOT COMPATIBLE`, means compiling will be failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,21 @@ Contributors:
   * [Documentation Changes](/.github/PULL_REQUEST_TEMPLATE/docs.md)
   * [Test/Build/Other Changes](/.github/PULL_REQUEST_TEMPLATE/other.md)
 
+##### Compatibility Matrix
+
+| Controller-Runtime      | release-0.2 | release-0.3 | release-0.4 | release-0.5 |
+|-------------------------|-------------|-------------|-------------|-------------|
+| kubernetes/release-1.17 |     -       |     -       |     ✓       |     ✓       |
+| kubernetes/release-1.16 |     -       |     -       |     ✓       |     ✓       |
+| kubernetes/release-1.15 |     ✓       |     ✓       |     -       |     -       |
+| kubernetes/release-1.14 |     ✓       |     ✓       |     -       |     -       |
+| kubernetes/release-1.13 |     -       |     -       |     -       |     -       |
+| kubernetes/release-1.12 |     -       |     -       |     -       |     -       |
+
+Key: 
+
+* `✓`: Compatible, `-`: Incompatible.
+  
 ## FAQ
 
 See [FAQ.md](FAQ.md)


README.md
```diff
@@ -39,6 +39,21 @@ Contributors:
...
+##### Compatibility Matrix
+
+| Controller-Runtime      | release-0.2 | release-0.3 | release-0.4 | release-0.5 |
+|-------------------------|-------------|-------------|-------------|-------------|
+| kubernetes/release-1.17 |     -       |     -       |     ✓       |     ✓       |
+| kubernetes/release-1.16 |     -       |     -       |     ✓       |     ✓       |
+| kubernetes/release-1.15 |     ✓       |     ✓       |     -       |     -       |
+| kubernetes/release-1.14 |     ✓       |     ✓       |     -       |     -       |
+| kubernetes/release-1.13 |     -       |     -       |     -       |     -       |
+| kubernetes/release-1.12 |     -       |     -       |     -       |     -       |
```
vincepri: I'd say we can probably remove these given that there is no support, wdyt?

mars1024:
Looking forward to this for a long time! Thanks!

There used to be a hard time for me to downgrade version and compile project again and again and again...

README.md
yue9944882: yeah by the policy, we provides support for three minor releases at a time. do you think if it's an improvement to clarify the compatibility for these out-of-support releases explicitly so that users wont even try spend time on it? given the fact that a portion of clusters aren't always catching up w/ the time windows of supports.. otherwise we can also rename it like `kubernetes/release-1.12 (unsupported)` for clarification?

vincepri: I don't think we necessarily need to for now, but also don't wanna block on it. When I first read the matrix I immediately noticed that the last two versions were there with no supported version. If we have  `1.14` as the first one, it'd probably going to be clear to users that's the minimum supported version

DirectXMan12:
This matrix is, at best, confusing.  If we're gonna do this, we should move this out into a separate document and explain what we mean.

Specifically: this implies that these are the supported Kubernetes versions that controller-runtime _works against, which is almost certainly wrong, and given the PR description, not what you meant.  This is particularly confusing, since client-go's compatibility matrix documents which versions of Kubernetes client-go can be used against.

AFAIK, the most recent CR works against most, if not all, of the versions of Kubernetes listed.

If we mean "what kubernetes libraries can this be compiled with", then we can say that.  However, that's a *wildly* different statement than "what versions of Kubernetes does CR work with", and as it is, this will only confuse the matter and reinforce the idea that you need kubernetes-1.Y dependencies to talk to a Kubernetes 1.Y cluster.

(this isn't intended to sound mean or antagonistic -- I just want to make sure that we don't confuse people more than they already are).

DirectXMan12:
/hold

for discussion

README.md
yue9944882: convinced. i will shorten the rows to 1.14 on master. how do you think if i pick the matrix to release-0.[1-5]?:
- for matrix upon `release-0.1`, it will cover k8s `release-1.1[0-3]`
- for matrix upon `release-0.2`, it will cover k8s `release-1.1[1-4]`
- for matrix upon `release-0.3`, it will cover k8s `release-1.1[2-6]`
- for matrix upon `release-0.4`, it will cover k8s `release-1.1[3-6]`
- for matrix upon `release-0.5`, it will cover k8s `release-1.1[4-7]`

yue9944882:
> If we mean "what kubernetes libraries can this be compiled with", then we can say that. However, that's a wildly different statement than "what versions of Kubernetes does CR work with"

@DirectXMan12 agree. given the fact that controller-runtime is doing well in keeping backward compatibility in terms of exported/public interfaces and doesnt have many breaking changes in functionalities ever since, so it's supposed to be always "working" w/ most kubernetes releases. while the compiling can basically be failing due to following reasons AFAIK:
- client-go completely deprecated/removed exported/public definitions at a point. e.g. https://github.com/kubernetes/kubernetes/issues/82859
- controller-runtime uses those definitions introduced in later client-go e.g. coordination api, workqueue metrics, etc.

and these compiling issues will also lead to some troubles when users trying to solve/match the dependencies w/ the vendor/mod management toolings in various usecase. the point of the matrix, is actually suggesting the users a proper range of k8s-ish dependency versions to try. how do you think about renaming it to `version-skew compiling matrix`?

Issue #855, opened 2020-03-12
Document controller health probes


GrigoriyMikhalkin:
Was mentioned in comments to https://github.com/kubernetes-sigs/controller-runtime/pull/419, that it's not quite clear how to set health probes for controller. 

Would be nice to add minimal example of manager with health/readiness checks.

P.S. I would gladly do that

alexeldeib:
I really like the examples provided by controller runtime, for example the package level one: https://github.com/kubernetes-sigs/controller-runtime/blob/master/example_test.go

and the webhook specific one: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/example_test.go

perhaps we could had a runnable main.go example with healthz?

alexeldeib:
If that's already what you meant, i'd go for it! Or are you thinking about something else like a gitbook for controller runtime?

GrigoriyMikhalkin:
@alexeldeib Was thinking more about adding example to https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/example_test.go(please check PR).

But adding runnable example(in addition), also a good idea. Maybe we can add health probes to examples in https://github.com/kubernetes-sigs/controller-runtime/tree/master/examples, what you think?

LeoLiuYan:
Maybe I can work for it?

GrigoriyMikhalkin:
@LeoLiuYan There's already PR: https://github.com/kubernetes-sigs/controller-runtime/pull/856

stevekuznetsov:
/reopen

This is not fixed and it's still entirely unclear how this works for a newcomer to this framework. Why do health and readiness probes have names? When do they start running? How do multiple interact?

GrigoriyMikhalkin:
@stevekuznetsov There's [now] [closed PR](https://github.com/kubernetes-sigs/controller-runtime/pull/856/files) with simple probes example. Also, i could write chapter in kubebuilder on health probes if that would be helpful.

taylormgeorge91:
> @stevekuznetsov There's [now] [closed PR](https://github.com/kubernetes-sigs/controller-runtime/pull/856/files) with simple probes example. Also, i could write chapter in kubebuilder book on health probes if that would be helpful.

That would be helpful if still an option

PR #856, opened 2020-03-13
📖 CR-855: added example of probes usage with controller

GrigoriyMikhalkin:
Solves #855


alexeldeib:
makes sense to me! dropped this into a main.go and tested locally, seems good.

/ok-to-test
/lgtm

omerlh:
What is `healthz.Ping`? I was wondering what I should put there :)

christopherhein:
@omerlh https://github.com/kubernetes-sigs/controller-runtime/blob/v0.5.1/pkg/healthz/healthz.go#L188

omerlh:
I would consider changing the name - Ping is not what I was expecting this function to do. Maybe just `ok` or `dummy`?

Issue #804, opened 2020-02-20
DynamicRestMapper dose not cancel previous rate.Reservation when rate limiter returns non-zero delay?


yuxiangqian:
Ran into an issue today while creating hundreds of resources within short period of time, DynamicRestMapper kept returning ErrRateLimited from Create call, and even if I put time.Sleep into the code with the returned duration from [DelayIfRateLimited](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/dynamicrestmapper.go#L46), the situation persists.
Did some digging, and found there might be a bug in the implementation [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/dynamicrestmapper.go#L202).

The function tries to obtain a token from rate.Limiter, and did not cancel the Reservation before returning error. My suspect is that even though the caller retries to call the API next time later, because the previous Reservation has not been released yet, it would cause rate.Limiter to keep on returning a Reservation with Duration != 0. Here is a simple code snippet which could prove that theory:

package main
import (
	"fmt"
	"time"
	"golang.org/x/time/rate"
)
type myLimiter struct {
	limiter *rate.Limiter
}
func (l *myLimiter) Reserve() *rate.Reservation {
	return l.limiter.Reserve()
}
func newLimiter() *myLimiter {
	l := &myLimiter{}
	l.limiter = rate.NewLimiter(2, 2)
	return l
}
type serverWithLimiter struct {
	l *myLimiter
}
func NewServer() *serverWithLimiter {
	s := &serverWithLimiter{}
	s.l = newLimiter()
	return s
}
func (s *serverWithLimiter) serve(i int) time.Duration {
	r := s.l.Reserve()
	if r.Delay() != 0 {
		// r.Cancel()
		return r.Delay()
	}
	fmt.Printf("process int %d\n", i)
	return 0
}
}
}

func main() {
	s := NewServer()
	for i := 0; i < 5; i++ {
		if d := s.serve(i); d != 0 {
			fmt.Printf("sleeping: %d\n", d)
			time.Sleep(d)
			i--
		}
	}
}

vincepri:
Could you submit a PR with a test case for this behavior?

/priority awaiting-more-evidence

alexeldeib:
[...]

PR #859, opened 2020-03-15
🐛 fix conservative delay in rate limiter

alexeldeib:
fixes https://github.com/kubernetes-sigs/controller-runtime/issues/804

this test case fails on master, and should succeed


vincepri:
/approve

Great oneliner fix!

PR #861, opened 2020-03-16
🐛 fix version check when reading webhooks from file

### LINKED ISSUE #867: envtest webhook version check b0rked
When we go to load the webhook in envtest from a file, we check the API version to make sure it's something we can work with.  Unfortunately, we don't check the API version correctly:

```
diff --git a/pkg/envtest/webhook.go b/pkg/envtest/webhook.go
index 9bcb5707..8fe65ba1 100644
--- a/pkg/envtest/webhook.go
+++ b/pkg/envtest/webhook.go
@@ -375,9 +375,13 @@ func readWebhooks(path string) ([]runtime.Object, []runtime.Object, error) {
 				return nil, nil, err
 			}
 
+			const (
+				admissionregv1 = "admissionregistration.k8s.io/v1beta1"
+				admissionregv1beta1 = "admissionregistration.k8s.io/v1"
+			)
 			switch {
 			case generic.Kind == "MutatingWebhookConfiguration":
-				if generic.APIVersion != "v1beta1" && generic.APIVersion != "v1" {
+				if generic.APIVersion != admissionregv1beta1 && generic.APIVersion != admissionregv1 {
 					return nil, nil, fmt.Errorf("only v1beta1 and v1 are supported right now for MutatingWebhookConfiguration (name: %s)", generic.Name)
 				}
 				hook := &unstructured.Unstructured{}
@@ -386,7 +390,7 @@ func readWebhooks(path string) ([]runtime.Object, []runtime.Object, error) {
 				}
 				mutHooks = append(mutHooks, hook)
 			case generic.Kind == "ValidatingWebhookConfiguration":
-				if generic.APIVersion != "v1beta1" && generic.APIVersion != "v1" {
+				if generic.APIVersion != admissionregv1beta1 && generic.APIVersion != admissionregv1 {
 					return nil, nil, fmt.Errorf("only v1beta1 and v1 are supported right now for ValidatingWebhookConfiguration (name: %s)", generic.Name)
 				}
 				hook := &unstructured.Unstructured{}
```

We should have a test for this, too, like the one for install CRDs.

cc @alenkacz do you have bandwidth to port the test?  If not, I'll try to get to it.

databus23:
At the moment the `DirectoryPaths` option of the `WebhookInstallOptions` is broken. This PR fixes it.

Fixed #867

### DIFF
--- a/pkg/envtest/testdata/webhooks/manifests.yaml
+++ b/pkg/envtest/testdata/webhooks/manifests.yaml
@@ -1,0 +1,101 @@
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-v1beta1
+  failurePolicy: Fail
+  name: mpods.kb.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - pods
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration2
+webhooks:
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-v1
+  failurePolicy: Fail
+  name: mpods2.kb.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - pods
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: validating-webhook-configuration
+webhooks:
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-v1beta1
+  failurePolicy: Fail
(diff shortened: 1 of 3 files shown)


vincepri:
/ok-to-test

Can we add or adjust existing unit tests?

databus23:
@vincepri Here you go

pkg/envtest/testdata/webhooks/manifests.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - CREATE
+    - UPDATE
+    resources:
+    - pods
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: validating-webhook-configuration
```
alenkacz: could you do a unique name here, please? it's the same as the previous one

pkg/envtest/webhook.go
```diff
@@ -377,7 +377,7 @@ func readWebhooks(path string) ([]runtime.Object, []runtime.Object, error) {
 
 			switch {
 			case generic.Kind == "MutatingWebhookConfiguration":
-				if generic.APIVersion != "v1beta1" && generic.APIVersion != "v1" {
+				if generic.APIVersion != "admissionregistration.k8s.io/v1beta1" && generic.APIVersion != "admissionregistration.k8s.io/v1" {
```
alenkacz: could you use the comparison @DirectXMan12 suggested here? https://github.com/kubernetes-sigs/controller-runtime/issues/867 instead of string

databus23:
@alenkacz Thanks for the feedback. All done.

alenkacz:
/lgtm

thanks for the fix

DirectXMan12:
We should follow up later with more specific robust tests that clearly highlight what's supported an what's not.  These are fine for now, but we should have ones that test the negative cases, and specifically test what we care about ("v1 works", "v1beta1 works", etc).

databus23:
@alenkacz Now that this is merged in master can we include this in a 0.5.x release as a bugfix?

Issue #730, opened 2019-12-17
Stopping individual controllers in a manager


suskin:
Hi controller-runtime friends! Sorry if this is a dumb question; I'm a bit new to Kubernetes and controller patterns!

I'm experimenting with ways to configure and set up controllers dynamically for user-specified Kinds (as part of the [crossplane project](https://github.com/crossplaneio/crossplane)). I'm interested in stopping a running controller if, for example, the desired Kind no longer exists. I'm using a manager; a single manager may be managing multiple controllers, each of which watches a different Kind.

I already have some code which creates a controller and adds it to an already-started manager, which appears to work. For stopping that controller, I've looked into the manager logic, and the controller logic, and there [appears to be support](https://github.com/kubernetes-sigs/controller-runtime/blob/efbeb2728794b105247bbf27c76a5c4949a73010/pkg/internal/controller/controller.go#L146) for stopping a controller, but it also appears to not be accessible if I am using a manager. Because a manager [reuses a single stop channel](https://github.com/kubernetes-sigs/controller-runtime/blob/efbeb2728794b105247bbf27c76a5c4949a73010/pkg/manager/internal.go#L226) for all of its controllers, and because it [starts all of the controllers without checking whether they're already started](https://github.com/kubernetes-sigs/controller-runtime/blob/efbeb2728794b105247bbf27c76a5c4949a73010/pkg/manager/internal.go#L212), and because controllers [appear to not check whether they've already been started](https://github.com/kubernetes-sigs/controller-runtime/blob/efbeb2728794b105247bbf27c76a5c4949a73010/pkg/internal/controller/controller.go#L146) when `Start` is called, I haven't found a way to do this with today's managers. It seems like one of the easier ways to add support for this would be to add another variation of the manager's `Add` [function](https://github.com/kubernetes-sigs/controller-runtime/blob/efbeb2728794b105247bbf27c76a5c4949a73010/pkg/manager/internal.go#L203). The variation would take in a stop channel as a parameter.

Am I doing a Weird Thing that I shouldn't be doing? Is there a recommended, better way of doing this that I am missing? If possible, it seems like it'd be the most convenient to continue to use the manager pattern and be able to just stop an individual controller under that manager, rather than having to reimplement some of the nice management code in order to call into a controller or into a source directly.

Thanks in advance for your help!

DirectXMan12:
[...]

PR #863, opened 2020-03-17
✨Allow controllers to be started and stopped separately from the manager

negz:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/730

Increasingly in [Crossplane](https://github.com/crossplane) we find that we'd like to use one kind of custom resource to describe how another, arbitrary, kind of custom resource should be reconciled. For example a `kind: Template` resource may configure how a `kind: Widget` resource should be reconciled. This involves the controller that watches for `kind: Template` managing the lifecycles of another set of controllers, including the controller that watches for `kind: Widget`. This is not currently possible in controller-runtime because all controllers are implicitly added to and started by the controller manager, which has no facility for stopping or removing a controller.

This PR makes it possible to start, run, _and stop_ a controller without ever adding the controller to the controller manager. My goal is not to arrive at the ideal API for this concept, but rather to enable the use case with the smallest possible change to controller-runtime. I imagine this will be niche functionality, and functionality that is slated for a broader refactoring at some point per https://github.com/kubernetes-sigs/controller-runtime/issues/764.

I've tested this functionality using https://github.com/negz/crossplane/commit/28e76ecd6fa285929370a8daa380d1fa27fbe9d8, which contains an example implementation of a controller that starts and stops other controllers.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -111,6 +111,7 @@ func Configure(name string, mgr manager.Manager, options Options) (Controller, e
 			return workqueue.NewNamedRateLimitingQueue(options.RateLimiter, name)
 		},
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
+		SetFields:               mgr.SetFields,
 		Name:                    name,
 	}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -67,7 +67,7 @@ type Controller interface {
 // New returns a new Controller registered with the Manager.  The Manager will ensure that shared Caches have
 // been synced before the Controller is Started.
 func New(name string, mgr manager.Manager, options Options) (Controller, error) {
-	c, err := Configure(name, mgr, options)
+	c, err := NewUnmanaged(name, mgr, options)
 	if err != nil {
 		return nil, err
 	}
@@ -76,8 +76,9 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 	return c, mgr.Add(c)
 }
 
-// Configure a new controller without starting it or adding it to the manager.
-func Configure(name string, mgr manager.Manager, options Options) (Controller, error) {
+// NewUnmanaged returns a new controller without adding it to the manager. The
+// caller is responsible for starting the returned controller.
+func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller, error) {
 	if options.Reconciler == nil {
 		return nil, fmt.Errorf("must specify Reconciler")
 	}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -67,6 +67,17 @@ type Controller interface {
 // New returns a new Controller registered with the Manager.  The Manager will ensure that shared Caches have
 // been synced before the Controller is Started.
 func New(name string, mgr manager.Manager, options Options) (Controller, error) {
+	c, err := Configure(name, mgr, options)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add the controller as a Manager components
+	return c, mgr.Add(c)
+}
+
+// Configure a new controller without starting it or adding it to the manager.
+func Configure(name string, mgr manager.Manager, options Options) (Controller, error) {
 	if options.Reconciler == nil {
 		return nil, fmt.Errorf("must specify Reconciler")
 	}
@@ -103,6 +114,5 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error) 
 		Name:                    name,
 	}
 
-	// Add the controller as a Manager components
-	return c, mgr.Add(c)
+	return c, nil
 }
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -130,7 +130,10 @@ type controllerManager struct {
 	// It and `internalStop` should point to the same channel.
 	internalStopper chan<- struct{}
 
-	elected chan struct{}
+	// leading is closed when this manager becomes the leader of a group of
+	// managers, either because it won a leader election or because no leader
+	// election was configured.
+	leading chan struct{}
 
 	startCache func(stop <-chan struct{}) error
 
@@ -425,8 +428,8 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 			return err
 		}
 	} else {
-		// Treat not having an election the same as being elected.
-		close(cm.elected)
+		// Treat not having leader election enabled the same as being elected.
+		close(cm.leading)
 		go cm.startLeaderElectionRunnables()
 	}
 
@@ -515,7 +518,7 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 		RetryPeriod:   cm.retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
-				close(cm.elected)
+				close(cm.leading)
 				cm.startLeaderElectionRunnables()
 			},
 			OnStoppedLeading: func() {
@@ -544,6 +547,6 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 	return nil
 }
 
-func (cm *controllerManager) Elected() <-chan struct{} {
-	return cm.elected
+func (cm *controllerManager) Leading() <-chan struct{} {
+	return cm.leading
 }
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -130,10 +130,10 @@ type controllerManager struct {
 	// It and `internalStop` should point to the same channel.
 	internalStopper chan<- struct{}
 
-	// leading is closed when this manager becomes the leader of a group of
+	// elected is closed when this manager becomes the leader of a group of
 	// managers, either because it won a leader election or because no leader
 	// election was configured.
-	leading chan struct{}
+	elected chan struct{}
 
 	startCache func(stop <-chan struct{}) error
 
@@ -429,7 +429,7 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 		}
 	} else {
 		// Treat not having leader election enabled the same as being elected.
-		close(cm.leading)
+		close(cm.elected)
 		go cm.startLeaderElectionRunnables()
 	}
 
@@ -518,7 +518,7 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 		RetryPeriod:   cm.retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
-				close(cm.leading)
+				close(cm.elected)
 				cm.startLeaderElectionRunnables()
 			},
 			OnStoppedLeading: func() {
@@ -547,6 +547,6 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 	return nil
 }
 
-func (cm *controllerManager) Leading() <-chan struct{} {
-	return cm.leading
+func (cm *controllerManager) Elected() <-chan struct{} {
+	return cm.elected
 }
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -130,6 +130,8 @@ type controllerManager struct {
 	// It and `internalStop` should point to the same channel.
 	internalStopper chan<- struct{}
 
+	elected chan struct{}
+
 	startCache func(stop <-chan struct{}) error
 
 	// port is the port that the webhook server serves at.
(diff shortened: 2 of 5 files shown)


alexeldeib:
/ok-to-test

love this, really elegant solution

vincepri:
This seems a breaking change for clients, unless I'm missing something?

negz:
@vincepri Could you elaborate on what makes it a breaking change? It's not immediately obvious to me. All existing functions, and methods continue to exist and function close to identically.

Running through the changes:

1. We add `controller.Configure` as an alternative to `controller.New`
2. `controller.New` will now explicitly configure `c.SetFields`, because `controller.New` now wraps `controller.Configure`. This should be a no-op, as this field was already injected by `manager.Add`.
3. We add a `manager.Elected()` method to the manager interface.

Is it the third change you're concerned about? I typically would not consider _new_ methods to be a breaking change, but I could see how it would be if we assume clients may be maintaining their own implementations of `Manager`.

vincepri:
Makes sense, I missed that last call to mgr.Add in the `New` function. It'd be great to provide examples around `Configure` for new users.

alexeldeib:
Could you also add a test similar to the goroutine test for starting/stopping a controller using this code?

negz:
> Could you also add a test similar to the goroutine test for starting/stopping a controller using this code?

Happy to add a test - are you thinking of a test like the existing one that ensures goroutines are reaped when we stop a manager?

pkg/controller/controller.go
```diff
@@ -67,6 +67,17 @@ type Controller interface {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Add the controller as a Manager components
+	return c, mgr.Add(c)
+}
+
+// Configure a new controller without starting it or adding it to the manager.
+func Configure(name string, mgr manager.Manager, options Options) (Controller, error) {
```
devigned: Does it seem odd to have a required field in an options parameter (`options.Reconciler`)?

negz: @devigned It does to me, but this is how the existing `controller.New` function works. Given that I expect most folks will continue to use `controller.New` I'd prefer to mimic that function's API rather than try to fix it.

alexeldeib:
I was thinking of that test case, but doing that for one controller pretty much devolves to the existing manager and controller test cases. Nevermind 👍

negz:
> It'd be great to provide examples around Configure for new users.

@vincepri Sure. Where is the best place to add examples?

Also, I wonder how far such an example should go? I imagine that few folks will ever call `controller.Configure` unless they're doing some fairly advanced stuff (i.e. wanting to manager their own controller lifecycles). Something like https://github.com/negz/crossplane/commit/28e76ecd6fa285929370a8daa380d1fa27fbe9d8 illustrates this pattern best, but is quite verbose. Perhaps a good compromise would be something like:

```go
func ConfigureExample(m ctrl.Manager) error {
        // A reconciler that knows how to reconcile *v1alpha1.Example
        r := &ExampleReconciler{}

        // Configure creates a new controller but does not add it to the supplied manager.
	c, err := controller.Configure("example", m, controller.Options{Reconciler: r})
	if err != nil {
		return err
	}

        // Request that our configured controller watch for *v1alpha1.Example
	if err := c.Watch(&source.Kind{Type: &v1alpha1.Example{}}, &handler.EnqueueRequestForObject{}); err != nil {
		return err
	}

        // Create a stop channel for our controller. The controller will stop when this
        // channel is closed.
        stop := make(chan struct{})

        // Start our controller in a goroutine so that we do not block.
	go func() {
                // Block until our controller manager is elected leader. We presume our entire
                // process will terminate if we lose leadership, so we don't need to handle that.
		<-m.Elected()

                // Start our controller. This will block until the stop channel is closed, or the
                // controller returns an error.
		if err := c.Start(stop); err != nil {
			log.Error(err, "cannot run experiment controller")
		}
	}()

	return nil
}
```

djzager:
> Where is the best place to add examples?

I think, if we decide an example is appropriate, that [`pkg/controller/example_test.go`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/example_test.go) would be a great place for the kind of example you shared.

pkg/manager/manager.go
```diff
@@ -50,6 +50,10 @@ type Manager interface {
 	// non-leaderelection mode (always running) or leader election mode (managed by leader election if enabled).
 	Add(Runnable) error
 
+	// Elected is closed when this manager is elected the leader, or when no
+	// election is configured.
+	Elected() <-chan struct{}
```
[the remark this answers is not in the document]
alexeldeib: I don't personally consider new methods breaking changes -- there's no change required on upgrade for existing users.

alexeldeib:
Can you close the stop channel in the example just to make it super explicit you can individually stop this controller too? Pending the checked-in example, lgtm

pkg/manager/internal.go
```diff
@@ -423,6 +425,8 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
 			return err
 		}
 	} else {
+		// Treat not having an election the same as being elected.
```
vincepri:
```suggestion
		// Treat not having leader election enabled the same as being elected.
```

pkg/controller/controller.go
alvaroaleman: Not convinced by the name, we now have `Configure` func that returns a new `Controller`.  IMHO whatever the name, it should start with `New`, maybe `NewWithoutManager`?

Also, the godoc here should mention the usecase.

pkg/manager/internal.go
```diff
@@ -538,3 +543,7 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 	go l.Run(ctx)
 	return nil
 }
+
+func (cm *controllerManager) Elected() <-chan struct{} {
```
alvaroaleman: `IsLeader` or `Leading` would probably be better names, the election is the process, leading is the state we care about. Also, please add godocs.

negz:
> I would like to mention that this is not what we should strive for

@alvaroaleman That's fair. A more direct way for me to frame this would have been "my goal is to enable access to this functionality without blocking on redesigning graceful termination in controller-runtime". I had this in mind because the request for this functionality (#730) was rolled into the broader graceful termination redesign issue (#764).

pkg/manager/manager.go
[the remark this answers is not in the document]
negz: I agree that adding a new method to that types must include to satisfy an interface would typically be a breaking change in Go, presuming that folks using this library were building their own implementations of `manager.Manager`. That said, given the size and specificity of this interface, I highly doubt there are alternate implementations in the wild.

pkg/manager/internal.go
negz: `Leading` seems like an appropriate name, considering we may become "leader" without holding an election.

pkg/controller/controller.go
negz: Ideally I think this function would be called `New` and the function currently named `New` would be renamed to something more like `AddToManager`. I agree `Configure` is not the best name, but `NewWithoutManager` doesn't seem great either. What does "without manager" mean? It still accepts and uses a manager, it just doesn't add the controller to it.

Perhaps `NewStandalone` or something that aims to represent that the controller runs without being added to the manager?

alvaroaleman: `NewUnmanaged` maybe?

pkg/manager/manager.go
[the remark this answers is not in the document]
djzager: haha, I talked myself out of when I went looking through [`VERSIONING.md`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/VERSIONING.md#avoiding-breaking-changes) but ... :man_shrugging: ... I am not really sure what happened.

After reading through the versioning guide I don't believe this is a breaking change.

negz:
I believe all feedback has now been addressed!

alexeldeib:
I think this is useful in the public API. I think we should still simplify some of the internals -- @djzager has been laying a lot of the groundwork with the informer context work, but we need to do the same throughout the cache/informer/controllers innards so we can tackle the high level changes in the manager (I started trying to list these changes and would drop a comment in the relevant issue). IMO public changes would be incremental, and this signature could plausibly be kept with only implementation changes later.

I'm not hearing much in the way of objections -- Solly maybe you could take a look, since this adds to the public API?

/lgtm
/assign @DirectXMan12

negz:
How safe is it to assume based on the handful of LGTMs that this functionality is likely to be merged? We're blocked on it in Crossplane, so we're currently switching to a fork of controller-runtime that includes this PR. We'd prefer to return back to mainline controller-runtime ASAP.

DirectXMan12:
If it's from one of the reviewers, then it's pretty likely :-).

I'm just a bit behind in my approvals

pkg/manager/manager.go
```diff
@@ -50,6 +50,11 @@ type Manager interface {
 	// non-leaderelection mode (always running) or leader election mode (managed by leader election if enabled).
 	Add(Runnable) error
 
+	// Leading is closed when this manager becomes the leader of a group of
+	// managers, either because it won a leader election or because no leader
+	// election was configured.
+	Leading() <-chan struct{}
```
DirectXMan12, line 56: non-blocking nit: is there a better name for this?  "closed when leading" strikes me as hella confusing

DirectXMan12: Aside: the california slang has gotten to me :fearful:

DirectXMan12:
actually

/hold

can we get a test here?

DirectXMan12:
doesn't have to be super complicated

feel free to cancel the hold when there's a test in place

pkg/manager/manager.go
negz: Haha, I started with `Elected()`, but @alvaroaleman convinced me `Leading` was more appropriate in https://github.com/kubernetes-sigs/controller-runtime/pull/863#discussion_r398860725. Any preference between those two?

The main detractor for `Elected` is that an election doesn't actually always happen, but "closed when elected" does seem more in line with similar examples ([like `ctx.Done()`](https://golang.org/pkg/context/#Context)), in that it reads "the channel is closed when the thing (election) finishes", rather than "the channel is closed when the thing (leading) starts".

negz:
> can we get a test here?

`controller.NewUnmanaged` is now called by `controller.New`, so I was kind of hoping to cheat and use the fact that all the existing `controller.New` tests "test through" to `controller.NewUnmanaged`. Does this change your desire for a test at all @DirectXMan12?

DirectXMan12:
Mainly, I wanted something that showed/tested the behavior of `Elected` directly, if it's not too crazy

pkg/manager/manager.go
negz: > I wanted something that showed/tested the behavior of Elected directly

I'm hoping from the comment below that you prefer `Elected`, which is also my preference, so I've pushed a commit that re-renames it.

negz:
@DirectXMan12 I've added a tiny test for the `Elected` method, but it seems the new apidiff test (which is great - I'd love to set it up for Crossplane) doesn't agree with our judgement in https://github.com/kubernetes-sigs/controller-runtime/pull/863#discussion_r398805250 that growing the method set of the `Manager` interface should not be considered a breaking API change. 😬

I agree in principle that expanding the method set of a public interface is a breaking change, as API consumers might be defining types that would satisfy the interface before the change, but would not after. However in practice I'm skeptical that there are alternative implementations of the `manager.Manager` interface in the wild due to its size and specificity. How do you think we should proceed?

negz:
Thanks for helping us get this merged @alexeldeib!

shomron:
I'm excited about this change but wanted to point out some practicalities that we might want to document:

* An unmanaged controller still depends on the cache provided by the manager. The manager must be started or the controller will block waiting for the cache.
* As far as I know, informers do not currently support removal of event handlers. When stopping the controller in the example, I assume its work queue will continue to grow unbounded?

Also - would it make sense to bound an unmanaged controller's lifespan to that of the manager given that the manager shutting down would also shut down dependencies of the controller (e.g. the cache)?

negz:
https://pkg.go.dev/github.com/crossplane/crossplane-runtime/pkg/controller

```go
o := kcontroller.Options{Reconciler: r}
c.Start(requirement.ControllerName(p.GetName()), o,
	controller.For(rq, &handler.EnqueueRequestForObject{}),
	controller.For(cp, &EnqueueRequestForRequirement{}))
```

@shomron we wrote a little wrapper to help with this that couples cache lifecycles to controller lifecycles. I agree it's not ideal, but it does the trick. it seems like there's a broader refactor in the wings for controller lifecycle management which will hopefully improve on this.

Issue #740, opened 2019-12-20
Listing objects with a label selector that exceeds 63 characters returns non-matching objects and no error


flawedmatrix:
I was investigating a bug in a controller where the wrong objects were being deleted during reconciliation. Upon closer inspection, it looks like the List() method was returning objects that didn't match the label selector provided, and that this occurs when the label selector exceeds the 63 character limit.

While an error is returned from the API in this case, List() returns all objects of the type passed in.

I've managed to reproduce this behavior with a minimal example with v0.4.0 of controller-runtime (against a kind cluster):

Setup
```go
var dummyNamespace corev1.Namespace = corev1.Namespace{
	ObjectMeta: metav1.ObjectMeta{
		Name: "some-namespace-name",
		Labels: map[string]string{
			"namespaceUniqueIdentifier": strings.Repeat("a", 63),
		},
	},
}
err = k8sClient.Create(context.Background(), &dummyNamespace)
if err != nil {
	log.Fatalln("Error creating namespace", err)
}
```
Test function
```go
func listNamespacesWithLabel(k8sClient client.Client, testNo int, label string) {
	listOption := client.MatchingLabels{
		"namespaceUniqueIdentifier": label,
	}

	nsList := corev1.NamespaceList{}
	err := k8sClient.List(context.Background(), &nsList, listOption)
	if err != nil {
		log.Fatalln("Error listing namespaces", err)
	}

	fmt.Printf("\n#################### TEST NO %d\n", testNo)
	fmt.Printf("%d namespaces listed\n", len(nsList.Items))
	for _, ns := range nsList.Items {
		fmt.Printf("Name %s, Labels: %#v \n", ns.Name, ns.Labels)
	}
}
```

Test
```go
listNamespacesWithLabel(k8sClient, 1, strings.Repeat("a", 63))
/*
Expected 1 namespaces
Actual Result:
#################### TEST NO 1
1 namespaces listed
Name some-namespace-name, Labels: map[string]string{"namespaceUniqueIdentifier":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}
*/

listNamespacesWithLabel(k8sClient, 2, strings.Repeat("b", 63))
/*
Expected 0 namespaces
Actual Result:
#################### TEST NO 2
0 namespaces listed
*/

listNamespacesWithLabel(k8sClient, 3, strings.Repeat("b", 64))
[...]

PR #882, opened 2020-04-02
:bug: MatchingLabels: Make sure requests with invalid labels fail

alvaroaleman:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/740

Following up on an idea by [@lavalamp](https://github.com/kubernetes/kubernetes/pull/89747#discussion_r401959488) this PR makes `MatchingLabels` construct an invalid selector rather than a (valid) match all selector in case of invalid input, which means the server will reject the request.

/assign @alexeldeib @vincepri

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -381,7 +381,7 @@ type MatchingLabels map[string]string
 
 func (m MatchingLabels) ApplyToList(opts *ListOptions) {
 	// TODO(directxman12): can we avoid reserializing this over and over?
-	sel := labels.SelectorFromSet(map[string]string(m))
+	sel := labels.SelectorFromValidatedSet(map[string]string(m))
 	opts.LabelSelector = sel
 }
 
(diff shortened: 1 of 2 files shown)


alexeldeib:
Nice fix. Do you think it's worth adding an invalid call against a real server in client_test.go to catch regressions? I realize your unit test essentially is the server side behavior, so it's really a nit, but can't hurt? I'll defer to you.

/lgtm

alvaroaleman:
> Nice fix. Do you think it's worth adding an invalid call against a real server in client_test.go to catch regressions? I realize your unit test essentially is the server side behavior, so it's really a nit, but can't hurt? I'll defer to you.

I think its fine, the important part is that we verify that even on invalid input a selector is populated rather than getting an empty selector back. Even if this had an integration test, an actual apiserver could behave differently than the one used in the integration test.

vincepri:
code LGTM, should this be a breaking change though?

alvaroaleman:
> code LGTM, should this be a breaking change though?

No, the previous behavior of "Return empty label selector that matches everything on invalid selector" is a bug and very unexpected.

vincepri:
fair, we'll add some more details in the release notes for users

PR #884, opened 2020-04-02
✨ Add ListWatch Options to Manager

answer1991:
ListWatch Options Help manager to filter objects which manager does not cared about during set-up List/Watch request to API Server.

### ListWatch filter vs Informer Event Predicate filter

ListWatch filter and Informer Event Predicate filter are all used to filter Object which controller not cared about, however, using ListWatch filter will helps reduce the manager memory cost and improve controller performer with less event received.

Example to filter Pod(Manager only cared about Pods with `foo=bar` Label):
```go
log.Info("setting up manager")
	ListWatchOptions: map[runtime.Object]func(options *metav1.ListOptions){
		&corev1.Pod{}: func(options *metav1.ListOptions) {
			options.LabelSelector = "foo=bar"
		},
	},
})
```

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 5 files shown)


alvaroaleman:
Hi @answer1991 , thanks for your work and PR. We track the underlying issue for this in https://github.com/kubernetes-sigs/controller-runtime/issues/244 and haven't yet agreed on a design.

Just setting a labelSelector on the cache results in the problem that objects requested through the client are also filtered by this label selector, as the client is cache backed. This means that we will always return 404 for objects that do not match the label selector, which is not what one would expect.

/hold

Issue #501, opened 2019-06-25
re-enable golangci-lint's godoc comment checking


DirectXMan12:
golangci-lint disables the godoc comment checking because the author considers it annoying.  We should make sure it's on, because we've got a couple of issues with our godocs.

DirectXMan12:
#477 has the fix in the first commit

kensipe:
@DirectXMan12 @vincepri  I think this is resolved.   https://github.com/kubernetes-sigs/controller-runtime/pull/773 updated linters... included with that is `golint` which enables godoc checking.

vincepri:
I don't think golint is enough to have godoc comments. It appears that golangci-lint has default exclusions. To re-enable godoc errors we'll need to add:

```
issues:
  exclude-use-default: false
  exclude:
    # errcheck: Almost all programs ignore errors on these functions and in most cases it's ok
    - Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*printf?|os\.(Un)?Setenv). is not checked
    # golint: False positive when tests are defined in package 'test'
    - func name will be used as test\.Test.* by other packages, and that stutters; consider calling this
    # govet: Common false positives
    - (possible misuse of unsafe.Pointer|should have signature)
    # staticcheck: Developers tend to write in C-style with an explicit 'break' in a 'switch', so it's ok to ignore
    - ineffective break statement. Did you mean to break out of the outer loop
    # gosec: Too many false-positives on 'unsafe' usage
    - Use of unsafe calls should be audited
    # gosec: Too many false-positives for parametrized shell calls
    - Subprocess launch(ed with variable|ing should be audited)
    # gosec: Duplicated errcheck checks
    - G104
    # gosec: Too many issues in popular repos
    - (Expect directory permissions to be 0750 or less|Expect file permissions to be 0600 or less)
    # gosec: False positive is triggered by 'src, err := ioutil.ReadFile(filename)'
    - Potential file inclusion via variable
```

to `.golangci.yml`

kensipe:
perhaps I misunderstood based on my reading of https://github.com/golangci/golangci-lint/issues/456

I'll look into it

DirectXMan12:
My experience is that you have to do what @vincepri mentioned as well.  Should be easy to check though :-)

PR #523, opened 2019-07-17
🏃Re-enabling the golangci-lint

hpandeycodeit:
As per the Issue #501 , Re-enabling the golangci-lint.

PR #885, opened 2020-04-03
:book: re-enable golangci-lint's godoc comment checking

camilamacedo86:
**Description**
Now, the lint test will be checking the comments for the go-docs and then, it is called in the checks made in prow to allow or not merge the PR. 

Also, in this PR, we have the fix for the places where the comments were missing.

**Motivation**

Closes: #501

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,28 +1,36 @@
-linters:
-  disable-all: true
-  enable:
-    - misspell
-    - structcheck
-    - golint
-    - govet
-    - deadcode
-    - errcheck
-    - varcheck
-    - goconst
-    - unparam
-    - ineffassign
-    - nakedret
-    - interfacer
-    - gocyclo
-    - lll
-    - dupl
-    - goimports
-
+run:
+  deadline: 5m
 linters-settings:
   lll:
     line-length: 170
   dupl:
     threshold: 400
+issues:
+  # don't skip warning about doc comments
+  exclude-use-default: false
 
-run:
-  timeout: 5m
+  # restore some of the defaults
+  # (fill in the rest as needed)
+  exclude-rules:
+  - linters: [errcheck]
+    text: "Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*printf?|os\\.(Un)?Setenv). is not checked"
+linters:
+  disable-all: true
+  enable:
+  - misspell
+  - structcheck
+  - golint
+  - govet
+  - deadcode
+  - errcheck
+  - varcheck
+  - goconst
+  - unparam
+  - ineffassign
+  - nakedret
+  - interfacer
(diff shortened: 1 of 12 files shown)


.github/main.workflow
```diff
@@ -7,3 +7,14 @@ action "verify-emoji" {
     uses = "./hack/release"
     secrets = ["GITHUB_TOKEN"]
 }
+
+workflow "Linters and Test" {
+    on = "push"
+    resolves = ["lint"]
+}
+
+action "lint" {
```
alvaroaleman, line 16: Why is this done via github actions and not via  a prow presubmit like the other tests?

hack/lint2check/README.md
```diff
@@ -0,0 +1,4 @@
+# lint2check
```
alvaroaleman, line 1: So we have a tool that runs in a github action and talks to the github checks api to tell it about its own status? This feels super redundant.

camilamacedo86: I agree. I just follow what was defined here: https://github.com/kubernetes-sigs/controller-runtime/issues/501

However, IHMO would not be required the hack/lint2check dir at all. 

@DirectXMan12 could we remove `.github/main.workflow` and the `hack/lint2check` and just leave the lint be called to check the comments as it has been in the `make test` ?

.github/main.workflow
camilamacedo86: It is called by the prow already. See that if it is missing some comment the test will fail. 

Example: 

```
level=info msg="[runner] linters took 9.994443697s with stages: goanalysis_metalinter: 9.9921147s"
pkg/internal/testing/integration/internal/etcd.go:7:1: comment on exported var `EtcdDefaultArgs` should be of the form `EtcdDefaultArgs ...` (golint)
//todo(just to ensure that the pr will fails and the test is called in the prow): EtcdDefaultArgs allow tests to run offline, by preventing API server from attempting to
```

ref: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/885/pull-controller-runtime-test-master/1247116818334617600/build-log.txt

Test executed just to ensure that it will fail.

hack/lint2check/README.md
vincepri: +1 on not needing this, we just need to enable the extra checks in golanci-lint

ref: https://github.com/kubernetes-sigs/controller-runtime/pull/523#discussion_r325396374

camilamacedo86: Hi @vincepri, 

I understand that all that needs to be made is done here. I indeed push a test commit to seeing the CI failing. See: https://github.com/kubernetes-sigs/controller-runtime/pull/885#discussion_r404006333

So, did you see anything that is missing be done here?

vincepri: Current changes look good to me, I think it fixes the immediate issue

PR #886, opened 2020-04-03
:warning: Fakeclient: reject create/update if Name isn't provided

### LINKED ISSUE #838: fakeclient should reject create|update without meta.name
Re: https://github.com/kubernetes-sigs/controller-runtime/pull/832#discussion_r385186769

Adding some of the basic validation you would find in other clients (ie. [client-go](https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L136-L139)) would be nice to have.

GrigoriyMikhalkin:
Fixes #838

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/testing"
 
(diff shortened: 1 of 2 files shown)


alvaroaleman:
Thanks! Could you change the title prefix to `:warning:` so that ppl are aware that this is a breaking change, as it will break all tests that do not provide a name?

GrigoriyMikhalkin:
> Thanks! Could you change the title prefix to `:warning:` so that ppl are aware that this is a breaking change, as it will break all tests that do not provide a name?

Done

pkg/client/fake/client.go
```diff
@@ -85,6 +85,9 @@ func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Ob
 	if err != nil {
 		return err
 	}
+	if accessor.GetName() == "" {
```
alvaroaleman, line 88: This needs to be a `NewInvalid`, this is how that looks like from a real api server:
```
              ErrStatus: {
                  ListMeta: {
                      SelfLink: "",
                      ResourceVersion: "",
                      Continue: "",
                      RemainingItemCount: nil,
                  },
                  Status: "Failure",
                  Message: "ConfigMap \"\" is invalid: metadata.name: Required value: name or generateName is required",
                  Reason: "Invalid",
                  Details: {
                      Name: "",
                      Group: "",
                      Kind: "ConfigMap",
                      UID: "",
                      Causes: [
                          {
                              Type: "FieldValueRequired",
                              Message: "Required value: name or generateName is required",
                              Field: "metadata.name",
                          },
                      ],
                      RetryAfterSeconds: 0,
                  },
                  Code: 422,
              }
```

PR #892, opened 2020-04-07
:sparkles: add EnqueueRequestForAnnotation enqueues Requests based on the presence of an annotation to watch resources

camilamacedo86:
**Story**
As an operator developer, it would be nice to be able to watch dependent resources based on the presence of an annotation in the dependent resource.

**Description**

While owner references are the idiomatic way to manage dependent resources, they don't solve all use cases. Owner references have restrictions that limit which objects can own other objects.

An annotation-based watch handler does not have these restrictions. However, they don't provide the same feature set either.

The watch handler can be used to trigger reconciliation of a parent resource based on changes to dependent resources that include an annotation. The purpose of this handler is to support cross-scope ownership relationships that are not supported by native owner references.

Since this is a niche feature that should only be used in specific circumstances. Note that for example, owner references are used for garbage collection. If a resource's owner is deleted, the resource will also be deleted. Because of this, annotation references need to be paired with a finalizer so that the operator can manually garbage collect.

**Example**

```go
	// Watch ClusterRoles that has the following annotations:
	// ...
	// annotations:
	//    watch.kubebuilder.io/owner-namespaced-name:my-namespace/my-replicaset
	//    watch.kubebuilder.io/owner-type:apps.ReplicaSet
	// ...
	// It will enqueue a Request to the primary-resource-namespace when some change occurs in a ClusterRole resource with these
	// annotations
	if err := c.Watch(&source.Kind{
		// Watch cluster roles
		Type: &rbacv1.ClusterRole{}},

		// Enqueue ReplicaSet reconcile requests using the
		// namespacedName annotation value in the request.
		&handler.EnqueueRequestForAnnotation{schema.GroupKind{Group:"ReplicaSet", Kind:"apps"}}); err != nil {
		entryLog.Error(err, "unable to watch Cluster Role")
		os.Exit(1)
	}

```

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/888

### DIFF
--- a/pkg/handler/enqueue_annotation.go
+++ b/pkg/handler/enqueue_annotation.go
@@ -1,0 +1,202 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handler
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ EventHandler = &EnqueueRequestForAnnotation{}
+
+const (
+	// NamespacedNameAnnotation defines the annotation that will be used to get the primary resource namespaced name.
+	// The handler will use this value to build the types.NamespacedName object used to enqueue a Request when an event
+	// to update, to create or to delete the observed object is raised. Note that if only one value be informed without
+	// the "/"  then it will be set as the name of the primary resource.
+	NamespacedNameAnnotation = "watch.kubebuilder.io/owner-namespaced-name"
+
+	// TypeAnnotation define the annotation that will be used to verify that the primary resource is the primary resource
+	// to use. It should be the type schema.GroupKind. E.g watch.kubebuilder.io/owner-type:core.Pods
+	TypeAnnotation = "watch.kubebuilder.io/owner-type"
+)
+
+// EnqueueRequestForAnnotation enqueues Requests based on the presence of annotations that contain the type and
+// namespaced name of the primary resource. The purpose of this handler is to support cross-scope ownership
+// relationships that are not supported by native owner references.
+//
+// This handler should ALWAYS be paired with a finalizer on the primary resource. While the
+// annotation-based watch handler does not have the same scope restrictions that owner references
+// do, they also do not have the garbage collection guarantees that owner references do. Therefore,
+// if the reconciler of a primary resource creates a child resource across scopes not supported by
+// owner references, it is up to the reconciler to clean up that child resource.
+//
+// **NOTE** You might prefer to use the EnqueueRequestsFromMapFunc with predicates instead of it.
+//
+// **Examples:**
+//
+// The following code will enqueue a Request to the `primary-resource-namespace` when some change occurs in a Pod
+// resource:
+//
+// ...
+// annotation := schema.GroupKind{Group: "ReplicaSet", Kind: "apps"}
+// if err := c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForAnnotation{annotation}); err != nil {
+//    entryLog.Error(err, "unable to watch Pods")
+//    os.Exit(1)
+// }
+// ...
+//
+// With the annotations:
+//
+// ...
+// annotations:
+//    watch.kubebuilder.io/owner-namespaced-name:my-namespace/my-replicaset
+//    watch.kubebuilder.io/owner-type:apps.ReplicaSet
+// ...
+//
+// The following code will enqueue a Request to the `primary-resource-namespace`, ReplicaSet reconcile,
+// when some change occurs in a ClusterRole resource
+//
+// ...
+// if err := c.Watch(&source.Kind{
+//	  // Watch cluster roles
+//	  Type: &rbacv1.ClusterRole{}},
+//
+//	  // Enqueue ReplicaSet reconcile requests using the
+//	  // namespacedName annotation value in the request.
+//	  &handler.EnqueueRequestForAnnotation{schema.GroupKind{Group:"ReplicaSet", Kind:"apps"}}); err != nil {
+//	      entryLog.Error(err, "unable to watch ClusterRole")
+//	      os.Exit(1)
+//    }
+// }
+// ...
+// With the annotations:
+//
+// ...
+// annotations:
+//    watch.kubebuilder.io/owner-namespaced-name:my-namespace/my-replicaset
+//    watch.kubebuilder.io/owner-type:apps.ReplicaSet
+// ...
+//
+// **NOTE** Cluster-scoped resources will have the NamespacedNameAnnotation such as:
+// `watch.kubebuilder.io/owner-namespaced-name:my-replicaset
+type EnqueueRequestForAnnotation struct {
+	// It is used to verify that the primary resource is the primary resource to use.
+	// E.g watch.kubebuilder.io/owner-type:core.Pods
+	Type schema.GroupKind
+}
+
+// Create implements EventHandler
+func (e *EnqueueRequestForAnnotation) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.Meta); ok {
+		q.Add(req)
+	}
+}
+
+// Update implements EventHandler
+func (e *EnqueueRequestForAnnotation) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.MetaOld); ok {
+		q.Add(req)
+	} else if ok, req := e.getAnnotationRequests(evt.MetaNew); ok {
+		q.Add(req)
+	}
+}
+
+// Delete implements EventHandler
+func (e *EnqueueRequestForAnnotation) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.Meta); ok {
+		q.Add(req)
+	}
+}
+
+// Generic implements EventHandler
+func (e *EnqueueRequestForAnnotation) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.Meta); ok {
+		q.Add(req)
+	}
+}
+
+// getAnnotationRequests will check if the object has the annotations for the watch handler and requeue
+func (e *EnqueueRequestForAnnotation) getAnnotationRequests(object metav1.Object) (bool, reconcile.Request) {
+	if typeString, ok := object.GetAnnotations()[TypeAnnotation]; ok && typeString == e.Type.String() {
+		namespacedNameString, ok := object.GetAnnotations()[NamespacedNameAnnotation]
+		if !ok {
+			log.Info("Unable to find the annotation for handle watch annotation",
+				"resource", object, "annotation", NamespacedNameAnnotation)
+		}
+		if len(namespacedNameString) < 1 {
+			return false, reconcile.Request{}
+		}
+		return true, reconcile.Request{NamespacedName: parseNamespacedName(namespacedNameString)}
+	}
+	return false, reconcile.Request{}
+}
+
+// parseNamespacedName will parse the value informed in the NamespacedNameAnnotation and return types.NamespacedName
+// with. Note that if just one value is informed, then it will be set as the name.
+func parseNamespacedName(namespacedNameString string) types.NamespacedName {
+	values := strings.SplitN(namespacedNameString, "/", 2)
+	if len(values) == 1 {
+		return types.NamespacedName{
+			Name:      values[0],
+			Namespace: "",
+		}
+	}
+	if len(values) >= 2 {
+		return types.NamespacedName{
+			Name:      values[1],
+			Namespace: values[0],
+		}
+	}
+	return types.NamespacedName{}
+}
+
+// SetWatchOwnerAnnotation is a helper method to add the watch annotation-based with the owner NamespacedName and the
+// schema.GroupKind to the object provided. This allows you to declare the watch annotations of an owner to an object.
+// If a annotation to the same object already exists, it'll be overwritten with the newly provided version.
+func SetWatchOwnerAnnotation(owner, object metav1.Object, ownerGK schema.GroupKind) error {
+
+	if len(owner.GetName()) < 1 {
+		return fmt.Errorf("%T has not a name, cannot call SetWatchOwnerAnnotation", owner)
+	}
+
+	if len(ownerGK.Kind) < 1 {
+		return fmt.Errorf("Owner Kind is not found, cannot call SetWatchOwnerAnnotation")
+	}
+
+	if len(ownerGK.Group) < 1 {
+		return fmt.Errorf("Owner Group is not found, cannot call SetWatchOwnerAnnotation")
+	}
+
+	annotations := object.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[NamespacedNameAnnotation] = fmt.Sprintf("%v/%v", owner.GetNamespace(), owner.GetName())
+	annotations[TypeAnnotation] = ownerGK.String()
+	object.SetAnnotations(annotations)
+
+	return nil
+}
--- a/pkg/handler/eventhandler_test.go
+++ b/pkg/handler/eventhandler_test.go
@@ -17,6 +17,10 @@ limitations under the License.
 package handler_test
 
 import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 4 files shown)


vincepri:
Can't this be a predicate on the watch?

camilamacedo86:
Hi @vincepri, 

> Can't this be a predicate on the watch?

Thank you for your input. However, to do a predicate on the watch we still needing the ownership. Am I right?  Note that:

> The watch handler can be used to trigger reconciliation of a parent resource based on changes to dependent resources that include an annotation. The purpose of this handler is to support cross-scope ownership relationships that are not supported by native owner references.

I added more info in the first comment. Please, let me know if it makes more sense now and if you re ok with.

joelanford:
@camilamacedo86 The `Closes` link in the description points to the wrong repo. I think you meant to use https://github.com/kubernetes-sigs/kubebuilder/issues/888

vincepri:
> However, to do a predicate on the watch we still needing the ownership. Am I right?

If you get a `controller.Controller` back when you build your cluster, you can use `controller.Watch` directly without the need for ownership.

For example:
- In Cluster API, we have a Machine and a Cluster objects
- The machine controller watches Clusters https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/machine_controller.go#L94-L96 
- Which calls this function https://github.com/kubernetes-sigs/cluster-api/blob/cb8f8e40d3a013a9cf8328f7a3df5553501f7baf/util/util.go#L542-L554 that calls `controller.Watch` with predicates

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ EventHandler = &EnqueueRequestForAnnotation{}
+
+// EnqueueRequestForAnnotation enqueues Requests based on the presence of an annotation that contains the
```
joelanford, line 31:
```suggestion
// EnqueueRequestForAnnotation enqueues Requests based on the presence of annotations that contain the type and
```

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+// namespaced name of the primary resource. The purpose of this handler is to support cross-scope ownership
+// relationships that are not supported by native owner references and then, it should only be used in very
+// specific circumstances.
```
joelanford, line 34:
```suggestion
// relationships that are not supported by native owner references.
```

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// Note that an annotation-based watch handler does not have the same owner reference restrictions.
+// However, do not provide the same feature set either. Beware that owner references are used for garbage collection
+// and then, if a resource's owner is deleted, the resource will also be deleted. However, this scenario cannot be satisfied
+// with an annotation-based watch handler and to achieve the same would be required implement finalizers to allow
+// the operator manually garbage collect the objects.
```
joelanford, line 40:
```suggestion
// This handler should ALWAYS be paired with a finalizer on the primary resource. While the
// annotation-based watch handler does not have the same scope restrictions that owner references
// do, they also do not have the garbage collection guarantees that owner references do. Therefore,
// if the reconciler of a primary resource creates a child resource across scopes not supported by
// owner references, it is up to the reconciler to clean up that child resource.
```

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// The primary use case for this, is to have a controller enqueue requests for the following scenarios
+// 1. namespaced primary object and dependent cluster scoped resource
+// 2. cluster scoped primary object.
```
joelanford, line 44: Correct me if I'm wrong, but hasn't Kubernetes allowed cluster-scoped primary objects to own namespace-scoped child objects for a few versions now? If so, case number 2 should be removed.

pkg/handler/enqueue.go
```diff
@@ -33,6 +33,14 @@ var _ EventHandler = &EnqueueRequestForObject{}
...
 type EnqueueRequestForObject struct{}
 
+const (
+	// NamespacedNameAnnotation - annotation that will be used to get the primary resource namespaced name.
+	NamespacedNameAnnotation = "sigs.k8s.io/primary-resource"
+
+	// TypeAnnotation - annotation that will be used to verify that the primary resource is the primary resource to use.
+	TypeAnnotation = "sigs.k8s.io/primary-resource-type"
+)
+
```
joelanford, line 43: Move these to `enqueue_annotation.go` since that's the only place they're referenced.

joelanford:
@vincepri `EnqueueRequestForAnnotation` is basically a specific implementation of an `EnqueueRequestsFromMapFunc`. If I'm understanding correctly, a predicate could filter out events for objects that don't have the expected annotations, and the handler map function would convert the annotation to a request using the namespacedName annotation value.

So the answer is, yes, this _could_ be done with existing handler and predicate features.

The question we need to answer is whether the problem this handler sets out to solve is widespread enough that the controller-runtime community would find enough value in it to justify its addition here.

This solution isn't perfect though.
- It requires developers to code their reconcilers to add these annotations correctly on child resources.
- It requires developers to add a finalizer on the primary resource if they need to cleanup these child resources since garbage collection doesn't work in all cases.
- Barring some ugly hacking around of annotations, it only allows a resource to be "owned" by one other object (owner references are a list, annotations are a map)

examples/builtins/main.go
```diff
@@ -70,6 +71,22 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	// Watch Nodes and enqueue by resource annotation key
+	var nodes *corev1.Node
+	nodesAnnotationType := fmt.Sprintf("%v.%v", nodes.Kind, nodes.APIVersion)
+	if err := c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForAnnotation{nodesAnnotationType}); err != nil {
+		entryLog.Error(err, "unable to watch Nodes")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
camilamacedo86: The example was to show that it works for cluster-scoped one. 
Applied your suggestion but I will check if we can improve more this ones.

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+	NamespaceAnnotation = "sigs.k8s.io/primary-resource-namespace"
+
+	// NameAnnotation - annotation that will be used to get the primary resource namespaced name.
+	// E.g my-resource-name
+	NameAnnotation = "sigs.k8s.io/primary-resource-name"
+
+	// TypeAnnotation - annotation that will be used to verify that the primary resource is the primary resource to use.
+	// E.g Pods.core
+	TypeAnnotation = "sigs.k8s.io/primary-resource-type"
+)
```
camilamacedo86, line 41: Hi @joelanford, 

Also, I think is better we have 1 annotation for the namespace and other for the name. Otherwise, we need to verify if the "/" was used and make clear for the users that when it is cluster scoped the value should be "/name" which shows strange.  WDYT? Are you ok with?

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+// and then, if a resource's owner is deleted, the resource will also be deleted. However, this scenario cannot be satisfied
+// with an annotation-based watch handler and to achieve the same would be required implement finalizers to allow
+// the operator manually garbage collect the objects.
+//
+// The primary use case for this, is to have a controller enqueue requests for the following scenarios
+// 1. namespaced primary object and dependent cluster scoped resource
+// 2. cluster scoped primary object.
+// 3. namespaced primary object and dependent namespaced scoped but in a different namespace object.
+type EnqueueRequestForAnnotation struct {
+	Type string
```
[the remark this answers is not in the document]
camilamacedo86: Added the explanation. good spot. tks

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Update implements EventHandler
+func (e *EnqueueRequestForAnnotation) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.MetaOld); ok {
+		q.Add(req)
+	}
+	if ok, req := e.getAnnotationRequests(evt.MetaNew); ok {
+		q.Add(req)
+	}
+}
```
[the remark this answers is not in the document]
camilamacedo86: Applied suggestion.

pkg/handler/enqueue_annotation.go
camilamacedo86: IHMO: The primary use case was not really made easier the understatement. So, I removed it and added the examples.

pkg/handler/eventhandler_test.go
```diff
@@ -950,4 +964,205 @@ var _ = Describe("Eventhandler", func() {
...
+			}
+			instance.Delete(evt, q)
+			Expect(q.Len()).To(Equal(1))
+
+			i, _ := q.Get()
+			Expect(i).To(Equal(reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: pod.GetNamespace(), Name: "baz"}}))
+		})
+
+		It("should enqueue a Request with the Annotations  of the object in the UpdateEvent.", func() {
```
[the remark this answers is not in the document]
camilamacedo86: Done.

pkg/handler/enqueue_annotation.go
joelanford: I don't think that's necessary. A slash character is not allowed in a valid DNS-1233 label (which is what names and namespaces must validate as), so IMO, the existing function used to split the string is sufficient.

camilamacedo86: Hi @joelanford, 

It is fine . Done as you suggested.

camilamacedo86:
Hi @@vincepri, 

WDYT? Could we move forward with?

vincepri:
I'd like to hear from @DirectXMan12, I have to give a deeper look at the changes. My initial reaction is that this is something that could be a library for clients that want to use it, but I didn't have enough time this week to think about it more

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+var _ EventHandler = &EnqueueRequestForAnnotation{}
+
+const (
+	// NamespacedNameAnnotation defines the annotation that will be used to get the primary resource namespaced name.
+	// The handler will use this value to build the types.NamespacedName object used to enqueue a Request when an event
+	// to update, to create or to delete the observed object is raised. Note that if only one value be informed without
+	// the "/"  then it will be set as the name of the primary resource.
+	NamespacedNameAnnotation = "sigs.k8s.io/primary-resource-namespace"
```
mengqiy, line 36: > sigs.k8s.io/primary-resource-namespace

Probably better to use `watch.kubebuilder.io/owner-namespaced-name` or something similar.
Reasons: 1) using `kubebuilder.io` to better show the owner of this annotation. 2) It's misleading to call a `namespacedName` `namespace`.

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,173 @@
...
+//    entryLog.Error(err, "unable to watch Pods")
+//    os.Exit(1)
+// }
+// ...
+//
+// With the annotations:
+//
+// ...
+// annotations:
+//    sigs.k8s.io/primary-resource-name: "my-namespace/my-pod"
```
mengqiy, line 73: godoc doesn't match the annotation name.

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,173 @@
...
+func (e *EnqueueRequestForAnnotation) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	if ok, req := e.getAnnotationRequests(evt.Meta); ok {
+		q.Add(req)
+	}
+}
+
+// getAnnotationRequests will check if the object has the annotations for the watch handler and requeue
+func (e *EnqueueRequestForAnnotation) getAnnotationRequests(object metav1.Object) (bool, reconcile.Request) {
+	if typeString, ok := object.GetAnnotations()[TypeAnnotation]; ok && typeString == e.Type {
+		if typeString, ok := object.GetAnnotations()[TypeAnnotation]; ok && typeString == e.Type {
```
mengqiy, line 140: Is this line identical to the line above?

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,173 @@
...
+// ...
+// annotations:
+//    sigs.k8s.io/primary-resource-name: "my-namespace/my-replicaset"
+//    sigs.k8s.io/primary-resource-type: "ReplicaSet.apps"
+// ...
+type EnqueueRequestForAnnotation struct {
+	// It is used to verify that the primary resource is the primary resource to use. It can be any arbitrary value.
+	// However, it would be recommended to fill it with `<Kind>.<group>` from the primary resource to make it easy
+	// to tell which specific resource is interested in.
+	Type string
```
mengqiy, line 104: IMO this'd better to be a `schema.GroupKind` unless there are some use cases that requires it be to an arbitrary string.
It's easier for users to get this right if it's more structured and you also provide a few helper functions.
I imagine that if this is `GroupKind` then we can at least have one helper function to set namespacedName and GroupKind. e.g.`func SetOwnerAnnotation(owner, owned runtime.Object) error`

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,173 @@
...
+	values := strings.Split(namespacedNameString, "/")
+	if len(values) == 1 {
+		return types.NamespacedName{
+			Name:      values[0],
+			Namespace: "",
+		}
+	}
+	if len(values) >= 2 {
+		return types.NamespacedName{
+			Name:      values[1],
```
mengqiy, line 168: I don't remember if "/" is allowed in name, but a safer way to handle this is to use `strings.SplitN(namespacedNameString, "/", 2)`.

pkg/handler/enqueue_annotation.go
camilamacedo86: Ops. mistake. solved

examples/builtins/main.go
```diff
@@ -70,6 +72,39 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Watch Pods that has the following annotations:
+	// ...
+	// annotations:
+	//    sigs.k8s.io/primary-resource-name: "my-namespac/my-pod"
```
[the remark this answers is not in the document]
camilamacedo86: done

pkg/handler/enqueue_annotation.go
mengqiy: @DirectXMan12 @droot @joelanford @vincepri WDYT?

pkg/handler/enqueue_annotation.go
camilamacedo86: solved

pkg/handler/enqueue_annotation.go
camilamacedo86: solved

pkg/handler/enqueue_annotation.go
camilamacedo86: I agree with your suggestion. I am working on to address it.

examples/builtins/main.go
```diff
@@ -70,6 +72,39 @@ func main() {
...
+	//    sigs.k8s.io/primary-resource-type: "Pods.core"
+	// ...
+	// It will enqueue a Request to the primary-resource-namespace when some change occurs in a Pod resource with these
+	// annotations
+	podAnnotationType := fmt.Sprintf("%v.%v", "Pods", "core")
```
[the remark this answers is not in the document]
camilamacedo86: Changed to use the schema .. Let us know what do you think.

pkg/handler/enqueue_annotation.go
camilamacedo86: Done @mengqiy let us know wdyt..

pkg/handler/enqueue_annotation.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if len(ownerGK.Kind) < 1 {
+		return fmt.Errorf("Owner Kind is not found, cannot call SetWatchOwnerAnnotation")
+	}
+
+	if len(ownerGK.Group) < 1 {
+		return fmt.Errorf("Owner Group is not found, cannot call SetWatchOwnerAnnotation")
+	}
+
+	object.SetAnnotations(map[string]string{
```
mcristina422, line 193: I think this needs to be 

```golang
annotations := object.GetAnnotations()
annotations[NamespacedNameAnnotation] = fmt.Sprintf("%v/%v", owner.GetNamespace(), owner.GetName())
annotations[TypeAnnotation] = ownerGK.String()
object.SetAnnotations(annotations)
```

or else you'll drop any other set annotations on `object` when using `SetWatchOwnerAnnotation`

camilamacedo86: Oh .. really tks for the good catcher .. it was a silly mistake ..
fixed and added tests to ensure as well.

alvaroaleman:
I have a bit of a feeling that this change is a bit of a workaround for multiple problems:
a) Writing custom map functions is too hard and not very intuitive
b) OnwerRefs do not support cross-namespace referencing

For a), I think this is a totally solvable problem, it just seems that no one bothered to do that yet (We had a bunch of downstream code for this in a different project but didn't upstream it)

For b), I think we should first have a discussion with sig-apimachinery if this is something that just isn't possible today or something that will never be possible. If the latter, we should spend a bit of time on thinking about nice api that replaces ownerRefs for that case. One problem I see for example with the current approach is that at most one owner can be declared. Also, I am not a huge fun of putting the burden of remembering to set finalizers and cleaning up objects if their parents are gone (AKA: build a custom garbage collector controller) onto the users of c-r.

camilamacedo86:
Hi @alvaroaleman, 

So, if I understood you are ok with move with it but you also would like to see customization in the garbage collector controller in order to no be required work with finalizers. Am I right? Could we work to do this improvement in a follow-up? WDYT?

alvaroaleman:
> So, if I understood you are ok with move with 

Sorry I don't understand what you mean by that?

But no, I am not a fan of this right now because:
* It isn't clear if this is a problem that should be solved downstream or upstream
* There has been no design doc about this and the current design is lacking

camilamacedo86:
HI @alvaroaleman, 

> There has been no design doc about this and the current design is lacking

Could you please let me know where we do add a design doc about this?  Also, when you say that the current design is lacking are you speaking about the need to use the finalizer always? So, if we are able to delete the child resources by using annotations it would be solved?  Am I right?

alvaroaleman:
I think before we start implementing something that is only needed because upstream does not support a certain feature we should:
* Find out why upstream does not support it
* Check if that reason still applies to upstream/if it can/should be done upstream
* Check if that reason applies to downstream

The reason is apparently that:
```
liggitt  20 minutes ago
namespaces are the traditional acl boundary. extending ownership across that boundary was intentionally avoided in the current design

alvaroaleman  16 minutes ago
Or is the problem that this might be abused to get info about the existence of objects in other namespaces?

liggitt  15 minutes ago
to get info or to block owner deletion
```
Source: https://kubernetes.slack.com/archives/C0EG7JC6T/p1587388988464000

The get info part definitely applies to us, the block owner part might apply, depending on how ppl implement garbage collection (Which they shouldn't need to do in the first place).

Can you please start a discussion with upstream sig-apimachinery if there is anything that can be done to support an usecase like this in a safe manner?

/hold

camilamacedo86:
HI @alvaroaleman, 

The topic was raised in the ML as suggested.  https://kubernetes.slack.com/archives/C0EG7JC6T/p1587401211471400

camilamacedo86:
We are closing this one. Based on the feedback shows that it would not be generic enough to upstream.

mcristina422:
I was thinking about this PR when I needed to write a controller that reconciled on k8s `ServiceAccount` types and the secrets that back them.

The `Secret` it refers to has no `OwnerRef` but there are annotations that refer to the "owning" `ServiceAccount`

 `kubernetes.io/service-account.name` and `kubernetes.io/service-account.uid`

PR #894, opened 2020-04-10
Adding docs using utilruntime.Must for scheme registration

christopherhein:
In addition to adding scaffolded `utilruntime.Must` for scheme registration in `kubebuilder` (https://github.com/kubernetes-sigs/kubebuilder/pull/1462) this updates the docs for `controller-runtime` docs to include panicking when scheme registration fails. 

You will notice that internal to `localSchemeBuilder` in `client-go` (https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go#L133) and most schemes within kubernetes they cause panics if anything from conversions to defaults fail.

After discussing with @DirectXMan12 this being left out probably either an oversight to that this pattern was developed after these packages where made.

### DIFF
--- a/pkg/scheme/scheme.go
+++ b/pkg/scheme/scheme.go
@@ -36,13 +36,13 @@ limitations under the License.
 //  )
 //
 // This also true of the built-in Kubernetes types.  Then, in the entrypoint for
-// your manager, assemble the scheme containing exactly the types you need.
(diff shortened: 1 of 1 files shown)


mengqiy:
@DirectXMan12 What is best practice using `Must` for adding scheme?

christopherhein:
thanks @mengqiy!

Issue #594, opened 2019-09-10
Expose a union interface for metav1.Object and runtime.Object


alvaroaleman:
Most (all?) Kubernetes API types implement both metav1.Object and runtime.Object. Code that needs both usually accepts either, then type asserts for the other.

It would be nice to have a union interface for the two and re-use that throughout our code

munnerz:
Just my 2c - not all types *do* implement both, only those that are actually created as CRDs.

In cert-manager we have a few extensibility mechanisms that rely on webhooks, and so we use a type that implements TypeMeta but not ObjectMeta: https://github.com/jetstack/cert-manager/blob/master/pkg/acme/webhook/apis/acme/v1alpha1/types.go#L30

This means we implement runtime.Object, but not metav1.Object. This gives us conversion etc. It’s a pattern borrowed from apiextensions-apiserver itself with the AdmissionReview and ConversionReview types.

I think there are places where this sort of union interface could be useful, but we should use it sparingly where it’s actually helpful, and consider that we limit the number of things that can use such a method to those that also have a ‘metadata’ stanza.

DirectXMan12:
Right -- there are places where we actually want to say "we need object metadata", places where we say "we can do more stuff if we have object metadata", and places where all we need is the type.

Adirio:
As I said in #666, unifying the interface in cases when there is a type assertion (and by assertion I mean a type cast that returns an error if it doesn't meet it) would move some errors from run time to compile time and would also allow IDEs to catch this errors. `SetControllerRef` seems to be one of them. Should we make a list and categorize them as @DirectXMan12 suggests (both interfaces required vs extra interface helpful)?

DirectXMan12:
> Should we make a list and categorize them as @DirectXMan12 suggests (both interfaces required vs extra interface helpful)?

:+1: from me

ruromero:
FWIW we introduced the KubernetesResource to our [operator-utils](https://github.com/RHsyseng/operator-utils/blob/master/pkg/resource/types.go) library and found it useful in multiple situations.

As has been already mentioned, this interface can avoid runtime errors and avoid unnecessary type conversions and maybe create higher level abstractions for some methods.

alvaroaleman:
Heads-up for ppl watching this issue, #898 merged so we now have such an union interface in pkg/controller/controllerutil. We don't yet use it anywhere though, hence leaving this issue open until we do use it where it makes sense.

[...]

PR #898, opened 2020-04-14
✨ Introduce controllerutil.Object interface

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

Introduce an interface that combines `metav1.Object` and `runtime.Object` as discussed in https://github.com/kubernetes-sigs/controller-runtime/issues/594

The main purpose of this interface is to reduce type conversions between the two and provide generic method signatures that can show potential errors at compile time and help developers to write simpler code.

This can make reconsider many places in this library and dependent ones its usage each time a cast is done, usually like the following:
```
//metav1.Object -> runtime.Object
ro, ok := owner.(runtime.Object)

//runtime.Object -> metav1.Object
o, err := meta.Accessor(obj)
```
An example where this interface can be used is https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L95

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -22,7 +22,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -133,11 +132,4 @@ func IgnoreNotFound(err error) error {
 		return nil
 	}
 	return err
-}
-
-// KubernetesResource allows functions to work indistinctly with any resource that
-// implements both Object interfaces.
-type KubernetesResource interface {
-	metav1.Object
-	runtime.Object
 }
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -22,6 +22,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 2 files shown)


ruromero:
Added CLA

pkg/client/interfaces.go
```diff
@@ -133,3 +134,10 @@ func IgnoreNotFound(err error) error {
 	}
 	return err
 }
+
+// KubernetesResource allows functions to work indistinctly with any resource that
+// implements both Object interfaces.
+type KubernetesResource interface {
```
joelanford, line 140: Given that this is a union of two `*.Object` types would it be more intuitive for this to also include `Object` in the name?

Also, I wonder if the `client` package is the most appropriate place for this type. Nothing is really jumping out to me as an alternate though. Perhaps `pkg/controller/controllerutil`? :man_shrugging:

ruromero: Renamed and moved

DirectXMan12:
> decide whether we want to update those function signatures as well.

Yeah, I think we want to do that.  I *think* there's actually situations where some of those aren't a breaking change (I seem to recall adding additional interfaces to a compound interface casts transparently or something, but I'd have to double check).

vincepri:
Given that this lives under `controllerutil`, could we just call it `Object` instead of `KubernetesObject `?

ruromero:
renamed @vincepri

vincepri:
Should we backport this? I'd be useful for other projects as well

luxas:
Thanks for this change :100:!
I'd love to have this under core `k8s.io/apimachinery`.
Would someone be up for "backporting" this to there?

vincepri:
@luxas Happy to help, although need some pointers to where we should place this interface :)

luxas:
Without knowing exactly, maybe https://pkg.go.dev/k8s.io/apimachinery@v0.18.5/pkg/api/meta?tab=doc would be a good place?

PR #902, opened 2020-04-16
:bug: fix CRD filtering in envtest package

mengqiy:
There is a bug in how envtest package determine if a CRD should be considered as a duplicate and skipped:
If we specify more than one CRD files that live in the same directory, it always skip the everything but the first CRD.

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -278,9 +278,15 @@ func renderCRDs(options *CRDInstallOptions) ([]runtime.Object, error) {
 	var (
 		err   error
 		info  os.FileInfo
-		crds  []*unstructured.Unstructured
 		files []os.FileInfo
 	)
+
+	type GVKN struct {
+		GVK  schema.GroupVersionKind
+		Name string
+	}
+
+	crds := map[GVKN]*unstructured.Unstructured{}
 
 	for _, path := range options.Paths {
 		var filePath = path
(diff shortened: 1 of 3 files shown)


pkg/envtest/crd.go
```diff
@@ -308,10 +308,7 @@ func renderCRDs(options *CRDInstallOptions) ([]runtime.Object, error) {
 		}
 
 		// If CRD already in the list, skip it.
-		if existsUnstructured(crds, crdList) {
-			continue
-		}
-		crds = append(crds, crdList...)
+		crds = append(crds, findNotExistingUnstructured(crds, crdList)...)
```
droot, line 311: I think we can keep a map of CRDs key'd by GVKN for deduping that will keep the code a lot more simpler. WDYT ?
It took me sometime to parse the current logic :)

mengqiy: I initially uses 2 slices to keep some existing behavior: If there are more than one definition of the same CRD, the last one will be installed eventually.
But it seems we can still achieve it using a map by keeping the last seen definition.

After thinking about on this, I guess the right thing to do here probaby is to error out. 
Thoughts?

vincepri: Given that this is `envtest` I'm fine either way. We can potentially improve it later. Maybe adding a few comments explaining why we're doing this and how it might help readers.

mengqiy: Changed it to a map.

PR #903, opened 2020-04-16
⚠ make Start on Source interface cancellable

### LINKED ISSUE #879: Add `context.Context` to `Source` interface's `Start` method to make cancellable
Do we need to add `context.Context` to the `Source` interface's `Start` method to make this cancellable?

_Originally posted by @joelanford in https://github.com/kubernetes-sigs/controller-runtime/pull/830_

djzager:
Adding context.Context to the Source interface's Start method makes it
cancellable.

Fixes #879


pkg/internal/controller/controller.go
```diff
@@ -162,7 +163,7 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 		// caches.
 		for _, watch := range c.watches {
 			log.Info("Starting EventSource", "controller", c.Name, "source", watch.src)
-			if err := watch.src.Start(watch.handler, c.Queue, watch.predicates...); err != nil {
+			if err := watch.src.Start(context.TODO(), watch.handler, c.Queue, watch.predicates...); err != nil {
```
joelanford, line 166: Here (and in various other places), we have access to a stop channel but not a context. Rather than using `context.TODO()`, could we create an equivalent context from the channel?

For example:
```go
func contextWithStopChannel(parent context.Context, stop <-chan struct{}) (context.Context, context.CancelFunc) {
	ctx, cancel := context.WithCancel(parent)

	// If stop is already closed, cancel immediately synchronously.
	// Otherwise start a goroutine that asynchronously cancels when stop is closed.
	select {
	case <-stop:
		cancel()
	default:
		go func() {
			<-stop
			cancel()
		}()
	}
	return ctx, cancel
}
```
Using something like this, we may be able to make these sources cancellable while not having to fully migrate stop channels to contexts.

DirectXMan12: seems reasonable to me

djzager: > Here (and in various other places), we have access to a stop channel but not a context.

I don't doubt you, I simply wasn't able to find any inside the current scope of this PR. Should I create an issue to look into this (this being - find other scenarios where we have a stop channel but not a context)?

alexeldeib: I think you're not finding many because you started in a place that used the injection pattern and didn't already accept a context. For the cases where the Start function accepts a context, this might be useful to help use remove the stop channels from children without doing so from the parent right away.

pkg/source/source_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package source_test
 
 import (
+	"context"
```
joelanford, line 20: Do we need tests that check whether `Start` will return an error if the passed context is cancelled?

djzager: done

alexeldeib:
/test pull-controller-runtime-apidiff-master

(just testing)

alvaroaleman:
This was implemented as part of https://github.com/kubernetes-sigs/controller-runtime/pull/1116
/close

PR #906, opened 2020-04-16
:warning: log/zap options Level and StacktraceLevel type

### LINKED ISSUE #899: The log/zap options for Level and StacktraceLevel have wrong type
As per the title, they're currently a concrete type `*zapcore.AtomicLevel`.  As per the underlying mechanism, they should be `zapcore.LevelEnabler`, since the underlying calls accept that.

/kind bug
/help

djzager:
Update the options using the concrete type *zap.AtomicLevel to instead
use the zapcore.LevelEnabler interface.

Fixes #899

### DIFF
--- a/pkg/log/zap/flags.go
+++ b/pkg/log/zap/flags.go
@@ -81,7 +81,7 @@ func newConsoleEncoder() zapcore.Encoder {
 }
 
 type levelFlag struct {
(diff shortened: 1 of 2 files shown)


DirectXMan12:
this is breaking, marking as such

PR #908, opened 2020-04-17
:bug: Disable metrics and health servers during recorder test

mcristina422:
When running the tests locally the recorder test fails because the metrics server fails to bind to the port. Since the metrics or health server isnt used during this test this PR disables them during the test

```
STEP: Creating the Manager
2020-04-17T14:11:07.370-0500    INFO    controller-runtime.metrics      metrics server is starting to listen    {"addr": ":8080"}
2020-04-17T14:11:07.385-0500    ERROR   controller-runtime.metrics      metrics server failed to listen. You may want to disable the metrics server or use another port if it is due to conflicts       {"error": "error listening on :8080: listen tcp :8080: bind: address already in use"}

```

```
Summarizing 1 Failure:

[Fail] recorder recorder [It] should publish events 
/Users/mcristina444/go/src/github.com/kubernetes-sigs/controller-runtime/pkg/internal/recorder/recorder_integration_test.go:55

```

### DIFF
--- a/pkg/internal/recorder/recorder_integration_test.go
+++ b/pkg/internal/recorder/recorder_integration_test.go
@@ -51,7 +51,10 @@ var _ = Describe("recorder", func() {
 	Describe("recorder", func() {
 		It("should publish events", func(done Done) {
 			By("Creating the Manager")
-			cm, err := manager.New(cfg, manager.Options{})
+			cm, err := manager.New(cfg, manager.Options{
+				HealthProbeBindAddress: "0",
+				MetricsBindAddress:     "0",
+			})
 			Expect(err).NotTo(HaveOccurred())
 
 			By("Creating the Controller")


mcristina422:
Also found as a failure in the output of https://github.com/kubernetes-sigs/controller-runtime/issues/907

alexeldeib:
The tests don't bring up anything multiple things on 8080, do they? I'm fine with disabling regardless, but I wanted to double check the error was from a local process.

/lgtm

thanks!

mcristina422:
There are multiple integration tests that start managers with open ports. I'm not sure why this one is causing issues, as none of the others ever fail to start this way. I was able to reproduce this a handful of time. This fix definitely worked on my machine though

alexeldeib:
That sounds potentially race-y, we may want to make that more deterministic (or more random I guess)

PR #909, opened 2020-04-17
:bug: Ensure KUBECONFIG is unset when testing GetConfigWithContext

mcristina422:
Where `~/kube/foo.yaml` is a valid yaml, when $KUBECONFIG is set the tests fail. 
```
KUBECONFIG=~/kube/foo.yaml TRACE=1 ./hack/check-everything.sh
```

```

Expected <snip> to be nil

Summarizing 1 Failure:

[Fail] Config GetConfigWithContext when kubeconfig files don't exist [It] should fail 
/Users/mcristi1/go/src/github.com/kubernetes-sigs/controller-runtime/pkg/client/config/config_test.go:85


```

This ensures the env var is unset prior to testing the failure case of `GetConfigWithContext("")`

PR #910, opened 2020-04-17
🏃 don't use apidiff on dirty commits

alexeldeib:
ref https://github.com/kubernetes-sigs/controller-runtime/pull/909#issuecomment-615505119

### DIFF
--- a/hack/apidiff.sh
+++ b/hack/apidiff.sh
@@ -25,8 +25,11 @@ cd "${REPO_ROOT}"
 
 APIDIFF="hack/tools/bin/go-apidiff"
 
-header_text "fetching tools"
+header_text "fetching and building go-apidiff"
 make "${APIDIFF}"
 
+git status
+
 header_text "verifying api diff"
+header_text "invoking: '${APIDIFF} ${PULL_BASE_SHA} --print-compatible'"
 "${APIDIFF}" "${PULL_BASE_SHA}" --print-compatible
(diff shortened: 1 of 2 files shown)


vincepri:
Is this because it's running during `make test`?

alexeldeib:
Yeah, exactly. I didn’t double check if it was the apidiff binary itself making the tree dirty or something else, but I figure this is fine either way.

alexeldeib:
I thought it might be because of make test, but they're in separate containers on prow so that can't be it. My other two thoughts are either it's because the binary is built under the git path which makes it dirty or it's something due to how prow checks it out (I looked at that code and nothing stood out immediately).

Either way, I figure we can add logging and try to force it through. We can at least ensure it's running against the right commits and figure out why it fails in the future.

alexeldeib:
Would've helped to test on the right branch -_- was wondering why prow wasn't retesting for me


The issue was with go.sum. I guess go changed a crypto library out under us? I left the `git status` in case this happens again. We can always revisit and clean up this script further.

alexeldeib:
fixes #681

joelanford:
@alexeldeib @vincepri FYI, not sure this is related, but I have noticed some cases where go-apidiff will leave the git tree in a dirty state when it is finished. I haven't had the time to dig into why that happens, but if you notice that, let me know and we can dig deeper.

Issue #912, opened 2020-04-20
Add log option and flag for zap time format


bharathi-tenneti:
Description: 

Add log option and flag for `zap time format` [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go#L127).
Follow up from https://github.com/kubernetes-sigs/controller-runtime/pull/767#discussion_r386226102

bharathi-tenneti:
The linked PR is being worked on to include Encoder Options filed instead of specific `--zap-time-encoder` flag, as added in the earlier stages of PR.

PR #915, opened 2020-04-22
 ✨ Added Encoder Config Options field.

bharathi-tenneti:
Description: This PR adds field in Options struct, for adding Encoder configuration options programmatically.

Motivation: This will make it easier for operator developers to specify different Encoder options.

Fixes: #912

### DIFF
--- a/pkg/log/zap/flags.go
+++ b/pkg/log/zap/flags.go
@@ -19,11 +19,11 @@ limitations under the License.
 package zap
 
 import (
+	"flag"
 	"fmt"
 	"strconv"
 	"strings"
 
-	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
@@ -42,11 +42,11 @@ var stackLevelStrings = map[string]zapcore.Level{
 }
 
 type encoderFlag struct {
-	setFunc func(zapcore.Encoder)
+	setFunc func(NewEncoderFunc)
 	value   string
 }
 
-var _ pflag.Value = &encoderFlag{}
+var _ flag.Value = &encoderFlag{}
 
 func (ev *encoderFlag) String() string {
 	return ev.value
@@ -60,9 +60,9 @@ func (ev *encoderFlag) Set(flagValue string) error {
 	val := strings.ToLower(flagValue)
 	switch val {
 	case "json":
-		ev.setFunc(newJSONEncoder())
+		ev.setFunc(newJSONEncoder)
 	case "console":
-		ev.setFunc(newConsoleEncoder())
+		ev.setFunc(newConsoleEncoder)
 	default:
 		return fmt.Errorf("invalid encoder value \"%s\"", flagValue)
 	}
@@ -70,22 +70,12 @@ func (ev *encoderFlag) Set(flagValue string) error {
 	return nil
 }
 
-func newJSONEncoder() zapcore.Encoder {
-	encoderConfig := zap.NewProductionEncoderConfig()
-	return zapcore.NewJSONEncoder(encoderConfig)
-}
-
-func newConsoleEncoder() zapcore.Encoder {
-	encoderConfig := zap.NewDevelopmentEncoderConfig()
-	return zapcore.NewConsoleEncoder(encoderConfig)
-}
-
 type levelFlag struct {
 	setFunc func(zapcore.LevelEnabler)
 	value   string
 }
 
-var _ pflag.Value = &levelFlag{}
+var _ flag.Value = &levelFlag{}
 
 func (ev *levelFlag) Set(flagValue string) error {
 	level, validLevel := levelStrings[strings.ToLower(flagValue)]
@@ -120,7 +110,7 @@ type stackTraceFlag struct {
 	value   string
 }
 
-var _ pflag.Value = &stackTraceFlag{}
+var _ flag.Value = &stackTraceFlag{}
 
 func (ev *stackTraceFlag) Set(flagValue string) error {
 	level, validLevel := stackLevelStrings[strings.ToLower(flagValue)]
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -30,6 +30,12 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// EncoderConfigOption is a function that can modify a `zapcore.EncoderConfig`.
+type EncoderConfigOption func(*zapcore.EncoderConfig)
+
+// NewEncoderFunc is a function that creates an Encoder using the provided EncoderConfigOptions.
+type NewEncoderFunc func(...EncoderConfigOption) zapcore.Encoder
+
 // New returns a brand new Logger configured with Opts. It
 // uses KubeAwareEncoder which adds Type information and
 // Namespace/Name to the log.
@@ -63,6 +69,22 @@ func Encoder(encoder zapcore.Encoder) func(o *Options) {
 	return func(o *Options) {
 		o.Encoder = encoder
 	}
+}
+
+func newJSONEncoder(opts ...EncoderConfigOption) zapcore.Encoder {
+	encoderConfig := zap.NewProductionEncoderConfig()
+	for _, opt := range opts {
+		opt(&encoderConfig)
+	}
+	return zapcore.NewJSONEncoder(encoderConfig)
+}
+
+func newConsoleEncoder(opts ...EncoderConfigOption) zapcore.Encoder {
+	encoderConfig := zap.NewDevelopmentEncoderConfig()
+	for _, opt := range opts {
+		opt(&encoderConfig)
+	}
+	return zapcore.NewConsoleEncoder(encoderConfig)
 }
 
 // Level sets the the minimum enabled logging level e.g Debug, Info
@@ -99,6 +121,14 @@ type Options struct {
 	// Encoder configures how Zap will encode the output.  Defaults to
 	// console when Development is true and JSON otherwise
 	Encoder zapcore.Encoder
+	// EncoderConfigOptions can modify the EncoderConfig needed to initialize an Encoder.
+	// See https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig for the list of options
+	// that can be configured.
+	// Note that the EncoderConfigOptions are not applied when the Encoder option is already set.
+	EncoderConfigOptions []EncoderConfigOption
+	// NewEncoder configures Encoder using the provided EncoderConfigOptions.
+	// Note that the NewEncoder function is not used when the Encoder option is already set.
+	NewEncoder NewEncoderFunc
 	// DestWritter controls the destination of the log output.  Defaults to
 	// os.Stderr.
 	DestWritter io.Writer
@@ -121,9 +151,8 @@ func (o *Options) addDefaults() {
 	}
 
 	if o.Development {
-		if o.Encoder == nil {
-			encCfg := zap.NewDevelopmentEncoderConfig()
-			o.Encoder = zapcore.NewConsoleEncoder(encCfg)
+		if o.NewEncoder == nil {
+			o.NewEncoder = newConsoleEncoder
 		}
 		if o.Level == nil {
 			lvl := zap.NewAtomicLevelAt(zap.DebugLevel)
@@ -136,9 +165,8 @@ func (o *Options) addDefaults() {
 		o.ZapOpts = append(o.ZapOpts, zap.Development())
 
 	} else {
-		if o.Encoder == nil {
-			encCfg := zap.NewProductionEncoderConfig()
-			o.Encoder = zapcore.NewJSONEncoder(encCfg)
+		if o.NewEncoder == nil {
+			o.NewEncoder = newJSONEncoder
 		}
 		if o.Level == nil {
 			lvl := zap.NewAtomicLevelAt(zap.InfoLevel)
@@ -156,6 +184,9 @@ func (o *Options) addDefaults() {
 					return zapcore.NewSampler(core, time.Second, 100, 100)
 				}))
 		}
+	}
+	if o.Encoder == nil {
+		o.Encoder = o.NewEncoder(o.EncoderConfigOptions...)
 	}
 	o.ZapOpts = append(o.ZapOpts, zap.AddStacktrace(o.StacktraceLevel))
 }
@@ -182,7 +213,7 @@ func NewRaw(opts ...Opts) *zap.Logger {
 // BindFlags will parse the given flagset for zap option flags and set the log options accordingly
 //  zap-devel: Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn)
 //			  Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
-//  zap-encoder: Zap log encoding ('json' or 'console')
+//  zap-encoder: Zap log encoding (one of 'json' or 'console')
 //  zap-log-level:  Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error',
 //			       or any integer value > 0 which corresponds to custom debug levels of increasing verbosity")
 //  zap-stacktrace-level: Zap Level at and above which stacktraces are captured (one of 'info' or 'error')
@@ -195,10 +226,10 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 
 	// Set Encoder value
 	var encVal encoderFlag
-	encVal.setFunc = func(fromFlag zapcore.Encoder) {
-		o.Encoder = fromFlag
+	encVal.setFunc = func(fromFlag NewEncoderFunc) {
+		o.NewEncoder = fromFlag
 	}
-	fs.Var(&encVal, "zap-encoder", "Zap log encoding ('json' or 'console')")
+	fs.Var(&encVal, "zap-encoder", "Zap log encoding (one of 'json' or 'console')")
 
 	// Set the Log Level
 	var levelVal levelFlag
@@ -221,10 +252,10 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 // UseFlagOptions configures the logger to use the Options set by parsing zap option flags from the CLI.
 //  opts := zap.Options{}
 //  opts.BindFlags(flag.CommandLine)
+//  flag.Parse()
 //  log := zap.New(zap.UseFlagOptions(&opts))
 func UseFlagOptions(in *Options) Opts {
 	return func(o *Options) {
 		*o = *in
-		o.addDefaults()
 	}
 }
--- a/pkg/log/zap/zap_test.go
+++ b/pkg/log/zap/zap_test.go
@@ -425,4 +425,84 @@ var _ = Describe("Zap log level flag options setup", func() {
 
 	})
 
+	Context("with only -zap-devel flag provided", func() {
+		It("Should set dev=true.", func() {
+			args := []string{"--zap-devel=true"}
+			fromFlags.BindFlags(&fs)
+			if err := fs.Parse(args); err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
+			out := Options{}
+			UseFlagOptions(&fromFlags)(&out)
+
+			Expect(out.Development).To(BeTrue())
+			Expect(out.Encoder).To(BeNil())
+			Expect(out.Level).To(BeNil())
+			Expect(out.StacktraceLevel).To(BeNil())
+			Expect(out.EncoderConfigOptions).To(BeNil())
+		})
+		It("Should set dev=false", func() {
+			args := []string{"--zap-devel=false"}
+			fromFlags.BindFlags(&fs)
+			if err := fs.Parse(args); err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
+			out := Options{}
+			UseFlagOptions(&fromFlags)(&out)
+
+			Expect(out.Development).To(BeFalse())
+			Expect(out.Encoder).To(BeNil())
+			Expect(out.Level).To(BeNil())
+			Expect(out.StacktraceLevel).To(BeNil())
+			Expect(out.EncoderConfigOptions).To(BeNil())
+
+		})
+	})
+
+	Context("with encoder options provided programmatically.", func() {
+
+		It("Should set Console Encoder, with given Nanos TimeEncoder option.", func() {
+			logOut := new(bytes.Buffer)
+			f := func(ec *zapcore.EncoderConfig) {
+				if err := ec.EncodeTime.UnmarshalText([]byte("nanos")); err != nil {
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
+			opts := func(o *Options) {
+				o.EncoderConfigOptions = append(o.EncoderConfigOptions, f)
+			}
+			log := New(UseDevMode(true), WriteTo(logOut), opts)
+			log.Info("This is a test message")
+			outRaw := logOut.Bytes()
+			// Assert for Console Encoder
+			res := map[string]interface{}{}
+			Expect(json.Unmarshal(outRaw, &res)).ToNot(Succeed())
+			// Assert for Epoch Nanos TimeEncoder
+			Expect(string(outRaw)).ShouldNot(ContainSubstring("."))
+
+		})
+		It("Should set JSON Encoder, with given Millis TimeEncoder option, and MessageKey", func() {
+			logOut := new(bytes.Buffer)
+			f := func(ec *zapcore.EncoderConfig) {
+				ec.MessageKey = "MillisTimeFormat"
+				if err := ec.EncodeTime.UnmarshalText([]byte("millis")); err != nil {
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
+			opts := func(o *Options) {
+				o.EncoderConfigOptions = append(o.EncoderConfigOptions, f)
+			}
+			log := New(UseDevMode(false), WriteTo(logOut), opts)
+			log.Info("This is a test message")
+			outRaw := logOut.Bytes()
+			// Assert for JSON Encoder
+			res := map[string]interface{}{}
+			Expect(json.Unmarshal(outRaw, &res)).To(Succeed())
+			// Assert for Epoch Nanos TimeEncoder
+			Expect(string(outRaw)).Should(ContainSubstring("."))
+			// Assert for MessageKey
+			Expect(string(outRaw)).Should(ContainSubstring("MillisTimeFormat"))
+		})
+
+	})
 })
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/onsi/gomega v1.10.1
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
-	github.com/spf13/pflag v1.0.5
 	go.uber.org/zap v1.15.0
 	golang.org/x/time v0.0.0-20200630173020-3af7569d3a1e
 	gomodules.xyz/jsonpatch/v2 v2.1.0
--- a/go.sum
+++ b/go.sum
@@ -35,6 +35,7 @@ github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbt
 github.com/PuerkitoBio/urlesc v0.0.0-20160726150825-5bd2802263f2/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
+github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc h1:cAKDfWh5VpdgMhJosfJnn5/FoN2SRZ4p7fJNX58YPaU=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
@@ -233,6 +234,7 @@ github.com/julienschmidt/httprouter v1.2.0/go.mod h1:SYymIcj16QtmaHHD7aYtjjsJG7V
 github.com/kisielk/errcheck v1.1.0/go.mod h1:EZBBE59ingxPouuu3KfxchcWSUPOHkagtvWXihfKN4Q=
 github.com/kisielk/errcheck v1.2.0/go.mod h1:/BMXB+zMLi60iA8Vv6Ksmxu/1UDYcXs4uQLJ+jE2L00=
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
+github.com/konsorten/go-windows-terminal-sequences v1.0.1 h1:mweAR1A6xJ3oS2pRaGiHgQ4OO8tzTaLawm8vnODuwDk=
 github.com/konsorten/go-windows-terminal-sequences v1.0.1/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/konsorten/go-windows-terminal-sequences v1.0.3/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/kr/logfmt v0.0.0-20140226030751-b84e30acd515/go.mod h1:+0opPa2QZZtGFBFZlji/RkVcI2GknAs/DXo4wKdlNEc=


bharathi-tenneti:
@hasbro17

pkg/log/zap/zap.go
```diff
@@ -132,6 +134,8 @@ type Options struct {
 	// Encoder configures how Zap will encode the output.  Defaults to
 	// console when Development is true and JSON otherwise
 	Encoder zapcore.Encoder
+	// TimeEncoder configures the zap time format. Defaults to epoch.
+	TimeEncoder zapcore.TimeEncoder
```
hasbro17, line 138: @bharathi-tenneti How would I configure this option with `zap.New()` without using any command line flags?

We have functional options for all the other fields that can be configured without flags:
```Go
log := NewRaw(zap.UseDevMode(true), zap.WriteTo(destWriter), zap.RawZapOpts(zapOpts...))
```

However if we just add the functional option helper for TimeEncoder to set the option as is:
```Go
func TimeEncoder(timeEncoder zapcore.TimeEncoder) func(o *Options) {
	return func(o *Options) {
		o.TimeEncoder = timeEncoder
	}
}
```
It could be a little cumbersome to set with `zap.New()`:
```Go
var te zapcore.TimeEncoder
if err := te.UnmarshalText([]byte("millis")); err != nil {
	return fmt.Errorf("not able to unmarshal time encoder value \"%s\"", val)
}

log := zap.New(zap.UseDevMode(true), zap.TimeEncoder(te))

```

So in the interest of usability I'm wondering if we can't change this field to be `TimeEncoder string`, so that the option is easier to set:
```Go
func TimeEncoder(timeEncoder string) func(o *Options) {
	return func(o *Options) {
		o.TimeEncoder = timeEncoder
	}
}

log := zap.New(zap.UseDevMode(true), zap.TimeEncoder("millis"))
```
And `addDefaults()` is the place where the TimeEncoder is actually initialized based on the option string:
```Go
func (o *Options) addDefaults() {
...
	if o.Development {
		if o.Encoder == nil {
			if o.TimeEncoder != "" {
				ecfs = append(ecfs, withTimeEncoding(o.TimeEncoder))
			}
			o.Encoder = newConsoleEncoder(ecfs)
		}
...
}

func withTimeEncoding(timeEncoderFormat string) encoderConfigFunc {
	return func(ec *zapcore.EncoderConfig) {
		var te zapcore.TimeEncoder
		if err := te.UnmarshalText([]byte(timeEncoderFormat)); err != nil {
			// As of zap v1.9.1 TimeEncoder's `UnmarshalText()` does not return an error and defaults to epoch
			// but if it ever does, default to epoch.
			te.UnmarshalText([]byte("epoch"))
		}
		ec.EncodeTime = te
	}
}
```

`BindFlags()` would need to be updated slightly but it should still be doable I think.

joelanford: @hasbro17 @bharathi-tenneti I think `zapcore.Encoder` is probably okay.

It isn't necessary to use `UnmarshalText`. The zapcore package contains a bunch of named and exported TimeEncoder implementations (e.g. [`zapcore.EpochMillisTimeEncoder`](https://godoc.org/go.uber.org/zap/zapcore#EpochMillisTimeEncoder))

So if I'm understanding correctly, a user could do this:
```go
te := zapcore.EpochMillisTimeEncoder
log := zap.New(zap.UseDevMode(true), zap.TimeEncoder(te))
```

pkg/log/zap/zap.go
```diff
@@ -98,6 +100,14 @@ func Encoder(encoder zapcore.Encoder) func(o *Options) {
 	}
 }
 
+// TimeEncoder configures how the logger will encode time format.
+// See Options.TimeEncoder
+func TimeEncoder(timeEncoder string) func(o *Options) {
+	return func(o *Options) {
+		o.TimeEncoder = newTimeEncoder(timeEncoder)
```
hasbro17, line 107: @bharathi-tenneti Can you test and update this per Joe's suggestion below to use `zapcore.TimeEncoder` as the input arg type.
```suggestion
func TimeEncoder(timeEncoder zapcore.TimeEncoder) func(o *Options) {
	return func(o *Options) {
		o.TimeEncoder = timeEncoder
```
If `zap.New(zap.TimeEncoder(zapcore.EpochMillisTimeEncoder))` works then that's better than parsing a string.

bharathi-tenneti: It is working as expected. 
Using helper func:
main.go
```go
logger := zap.New(zap.TimeEncoder(zapcore.EpochMillisTimeEncoder), zap.UseDevMode(true))
```
output
------
```shell
$ operator-sdk run --local
INFO[0000] Running the operator locally in namespace default. 
1.5880158036435872e+12	INFO	cmd	Operator Version: 0.0.1
1.588015803643628e+12	INFO	cmd	Go Version: go1.14.2
1.5880158036436328e+12	INFO	cmd	Go OS/Arch: darwin/amd64
1.588015803643636e+12	INFO	cmd	Version of operator-sdk: v0.17.0+git
1.588015803644689e+12	INFO	leader	Trying to become the leader.
1.5880158036447e+12	INFO	leader	Skipping leader election; not running in a cluster.
1.58801580396404e+12	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
1.588015803964294e+12	INFO	cmd	Registering Components.
1.588015803964316e+12	INFO	cmd	Starting the Cmd.
1.588015803964658e+12	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}

```
Using flags:
-----------
```go
logger := zap.New(zap.UseFlagOptions(&opts))
```
```shell
 $ operator-sdk run --local --operator-flags="--zap-encoder=console --zap-time-encoder=nanos"
INFO[0000] Running the operator locally in namespace default. 
1588015875500174000	INFO	cmd	Operator Version: 0.0.1
1588015875500221000	INFO	cmd	Go Version: go1.14.2
1588015875500228000	INFO	cmd	Go OS/Arch: darwin/amd64
1588015875500235000	INFO	cmd	Version of operator-sdk: v0.17.0+git
1588015875501658000	INFO	leader	Trying to become the leader.
1588015875501674000	INFO	leader	Skipping leader election; not running in a cluster.
1588015875812863000	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
1588015875813021000	INFO	cmd	Registering Components.
1588015875813041000	INFO	cmd	Starting the Cmd.
1588015875813202000	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}

```

pkg/log/zap/flags.go
```diff
@@ -60,26 +64,73 @@ func (ev *encoderFlag) Set(flagValue string) error {
...
+func newTimeEncoder(flagValue string) zapcore.TimeEncoder {
+	var encoder zapcore.TimeEncoder
+	val := strings.ToLower(flagValue)
+	switch val {
+	case "iso8601", "rfc3339", "rfc3339nano", "millis", "nanos", "epoch":
+		if err := encoder.UnmarshalText([]byte(val)); err != nil {
+			fmt.Printf("not able to unmarshal time encoder value \"%s\"", val)
+		}
+	default:
+		fmt.Printf("invalid time encoder value \"%s\"", val)
```
hasbro17, line 123: If we pass an invalid string `--zap-time-encoder=foobar` then this prints out an error but returns a nil encoder right? Shouldn't we be returning an error here and above?

bharathi-tenneti: True, Can we set default to "epoch" time encoder, and print out message saying the same?
```go
	switch val {
	case "iso8601", "rfc3339", "rfc3339nano", "millis", "nanos", "epoch":
		if err := encoder.UnmarshalText([]byte(val)); err != nil {
			fmt.Printf("not able to unmarshal time encoder value \"%s\"", val)
		}
	default:
		encoder = zapcore.TimeEncoder(zapcore.EpochTimeEncoder)
		fmt.Printf("invalid value for time encoder, \"%s\". Setting default to epoch. \n", val)
	}
	return encoder
```

hasbro17: I would lean towards returning the error since that's consistent with the behavior for other logging flags.

pkg/log/zap/zap.go
```diff
@@ -255,6 +277,14 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
...
 	return func(o *Options) {
 		*o = *in
+		if o.Encoder != nil && o.TimeEncoder != nil {
+			ecfs = append(ecfs, withTimeEncoding(o.TimeEncoder))
+			if encVal.value == "json" {
+				o.Encoder = newJSONEncoder(ecfs)
+			} else {
+				o.Encoder = newConsoleEncoder(ecfs)
+			}
+		}
```
joelanford, line 287: Is this block of code handled by the added logic in the `addDefaults()` function?

pkg/log/zap/zap.go
```diff
@@ -132,6 +142,8 @@ type Options struct {
 	// Encoder configures how Zap will encode the output.  Defaults to
 	// console when Development is true and JSON otherwise
 	Encoder zapcore.Encoder
+	// TimeEncoder configures the zap time format. Defaults to epoch.
+	TimeEncoder zapcore.TimeEncoder
```
joelanford, line 146: I envision a future feature request for other `zapcore.EncoderConfig` fields to be exposed as flags. Instead of adding new fields to the `Options` struct for each one, I wonder if it makes sense to have a more generic `Options` field, like `EncoderConfig *zapcore.EncoderConfig` or `EncoderConfigOptions []func(*zapcore.EncoderConfig)`?

Also, one thing I'm wondering is how obvious it is that the `TimeEncoder` value is only used when the `Encoder` field is nil.  As far as I can tell, there's no way to inject `TimeEncoder` (or any other `EncoderConfig` field) into an Encoder after it has been created. This would be an issue no matter what we put in the `Options` struct related to configuring the `EncoderConfig`.

pkg/log/zap/zap.go
bharathi-tenneti: @joelanford as I understand `addDefaults()`, will add the timeEncoder to the default Encoder value. Above block of code is for when `Encoder` is being set from flags.

pkg/log/zap/zap.go
bharathi-tenneti: @hasbro17 Please share your thoughts as well.

pkg/log/zap/flags.go
```diff
@@ -41,6 +41,10 @@ var levelStrings = map[string]zapcore.Level{
 	"fatal":  zap.FatalLevel,
 }
 
+type encoderConfigFunc func(*zapcore.EncoderConfig)
+
+var ecfs []encoderConfigFunc
```
estroz, line 46: Is this global necessary? If so, can you add an explanatory comment as to why one is needed?

pkg/log/zap/flags.go
```diff
@@ -41,6 +41,10 @@ var levelStrings = map[string]zapcore.Level{
 	"fatal":  zap.FatalLevel,
 }
 
+type encoderConfigFunc func(*zapcore.EncoderConfig)
```
estroz, line 44: Type usage comment.

pkg/log/zap/zap.go
estroz: An error should be returned if a user tries to set a `TimeEncoder` with a non-nil `Encoder`.

pkg/log/zap/zap.go
```diff
@@ -223,8 +239,14 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		"Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). "+
 			"Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)")
 
+	// Set TimeEncoder value
+	var timeEncVal timeEncoderFlag
+	timeEncVal.setFunc = func(fromFlag zapcore.TimeEncoder) {
+		o.TimeEncoder = fromFlag
+	}
+	fs.Var(&timeEncVal, "zap-time-encoder", "Zap time encoding format('rfc3339', 'rfc339nanos', 'epoch', 'millis', 'nano', or 'iso8601')")
```
estroz, line 247:
```suggestion
	fs.Var(&timeEncVal, "zap-time-encoder", "Zap time encoding format: ('rfc3339', 'rfc339nanos', 'epoch', 'millis', 'nano', or 'iso8601')")
```

pkg/log/zap/zap.go
```diff
@@ -255,6 +277,14 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 func UseFlagOptions(in *Options) Opts {
 	return func(o *Options) {
 		*o = *in
+		if o.Encoder != nil && o.TimeEncoder != nil {
+			ecfs = append(ecfs, withTimeEncoding(o.TimeEncoder))
+			if encVal.value == "json" {
+				o.Encoder = newJSONEncoder(ecfs)
+			} else {
+				o.Encoder = newConsoleEncoder(ecfs)
```
estroz, line 285: Must `o.Encoder` always be set here? If so are we assuming the default is "console"?

pkg/log/zap/zap.go
hasbro17: > I'm wondering is how obvious it is that the TimeEncoder value is only used when the Encoder field is nil

@joe I think the Encoder option and any EncoderConfig options should be exclusive. The Encoder option field encompasses all EncoderConfig options. So it takes precedence.

So if the user passes an Encoder with some preconfigured EncoderConfig then I'm not sure why it makes sense to further configure the EncoderConfig. Maybe we can document that better but I think the behavior makes sense.
E.g:
```Go
encCfg:= zapcore.EncoderConfig{
        MessageKey: "message",
        LevelKey:    "level",
        EncodeLevel: zapcore.CapitalLevelEncoder,
        TimeKey:    "time",
        EncodeTime: zapcore.ISO8601TimeEncoder,
        CallerKey:    "caller",
        EncodeCaller: zapcore.ShortCallerEncoder,
}
encoder := zapcore.NewConsoleEncoder(encCfg)
log := zap.New(zap.Encoder(encoder))

// why pass this option if you control the entire encoder config already?
newTimeEncoder := zapcore.EpochNanosTimeEncoder
log := zap.New(zap.Encoder(encoder), zap.TimeEncoder(newTimeEncoder))

```

estroz: @hasbro17 having a no-op option, regardless of what a user _should_ be doing, is weird UX. We should at least add docs.

hasbro17: But I agree we probably should make this into an `EncoderConfigOptions` field to allow configuring all EncoderConfig fields.
So maybe something like:
```Go
// Options contains all possible settings
type Options struct {
...
	EncoderConfigOptions []func(*zapcore.EncoderConfig)
...
}
```
Where the helpers can help tweak EncoderConfig before it's passed to initialize the Encoder.
```Go
// TimeEncoder configures how the logger will encode time format.
// See Options.TimeEncoder
func TimeEncoder(timeEncoder zapcore.TimeEncoder) func(o *Options) {
	return func(o *Options) {
		f := func(ec *zapcore.EncoderConfig) {
			ec.EncodeTime = timeEncoder
		}
		o.EncoderConfigOption = append(o.EncoderConfigOption, f)
	}
}

// LevelEncoder configures how the logger will encode level format.
// See Options.TimeEncoder
func LevelEncoder(levelEncoder zapcore.LevelEncoder) func(o *Options) {
	return func(o *Options) {
		f := func(ec *zapcore.EncoderConfig) {
			ec.EncodeLevel = levelEncoder
		}
		o.EncoderConfigOption = append(o.EncoderConfigOption, f)
	}
}
```
And that's used when setting the defaults:
```Go
// addDefaults adds defaults to the Options
func (o *Options) addDefaults() {
...
	if o.Development {
		if o.Encoder == nil {
			o.Encoder = newConsoleEncoder(o.EncoderConfigOptions)
		}
...
```
And we already have:
```Go
type encoderConfigFunc func(*zapcore.EncoderConfig)

func newConsoleEncoder(ecfs []encoderConfigFunc) zapcore.Encoder {
	encoderConfig := zap.NewDevelopmentEncoderConfig()
	for _, f := range ecfs {
		f(&encoderConfig)
	}
	return zapcore.NewConsoleEncoder(encoderConfig)
}
```

@bharathi-tenneti WDYT? I think the BindFlags should be able to configure the `EncoderConfigOptions` field based on the individual flags like `zap-time-format`.

pkg/log/zap/zap.go
bharathi-tenneti: Yes, the possible values are `json` or `console`, we error out for anything else.
And the reason, we are setting here, is to set any EncoderConfigOptions that are passed using commandLine Flags, before we call `addDefaults`.

pkg/log/zap/flags.go
bharathi-tenneti: Moved `var ecfs []encoderConfigFunc`

pkg/log/zap/zap.go
```diff
@@ -142,8 +145,8 @@ type Options struct {
 	// Encoder configures how Zap will encode the output.  Defaults to
 	// console when Development is true and JSON otherwise
 	Encoder zapcore.Encoder
-	// TimeEncoder configures the zap time format. Defaults to epoch.
-	TimeEncoder zapcore.TimeEncoder
+	// EncoderConfigOptions configures options for Encoder.
```
hasbro17, line 148: We should probably add a little more detail in the godocs for this option. 
And point out the caveat that @joelanford mentioned that this does not get applied when `Encoder != nil`.
```suggestion
	// EncoderConfigOptions can modify the EncoderConfig needed to initialize an Encoder.
	// See https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig for the list of options 
	// that can be configured.
	// Note that the EncoderConfigOptions are not applied when the Encoder option is already set.
```

pkg/log/zap/zap.go
```diff
@@ -98,6 +100,17 @@ func Encoder(encoder zapcore.Encoder) func(o *Options) {
 	}
 }
 
+// TimeEncoder configures how the logger will encode time format.
+// See Options.EncoderConfigOptions
+func TimeEncoder(timeEncoder zapcore.TimeEncoder) func(o *Options) {
```
hasbro17, line 105: @joelanford Now that all [EncoderConfig options](https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig) are configurable I'm thinking we could add helpers for all the other fields while we're here. 
Should be straightforward enough now, e.g:
```Go
func LevelEncoder(levelEncoder zapcore.LevelEncoder) func(o *Options) {
	return func(o *Options) {
		f := func(ec *zapcore.EncoderConfig) {
			ec.EncodeLevel = levelEncoder
		}
		o.EncoderConfigOption = append(o.EncoderConfigOption, f)
	}
}
```
Or wait until someone actually needs them, like the TimeEncoder?

bharathi-tenneti: As discussed, We would not be adding helpers as of yet.

pkg/log/zap/zap.go
```diff
@@ -255,6 +282,14 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
...
 	return func(o *Options) {
 		*o = *in
+		// Set the encoder to use EncoderConfigOptions if exists.
+		if o.Encoder != nil && o.EncoderConfigOptions != nil {
+			if encVal.value == "json" {
+				o.Encoder = newJSONEncoder(o.EncoderConfigOptions)
+			} else {
+				o.Encoder = newConsoleEncoder(o.EncoderConfigOptions)
+			}
+		}
```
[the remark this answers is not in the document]
bharathi-tenneti: As discussed with @joelanford , We may want to think about re-design. Meanwhile added unit test cases in latest commit, to help testing log options easier.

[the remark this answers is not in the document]
bharathi-tenneti: Below is the proposed design change to capture `timeEncoder` and any other `encoderConfigOptions` in the future to be set, as discussed with @joelanford 
```func UseFlags(fs flag.FlagSet) Opts {
    flags := structContainingFlagValueTypes{}
    flags.bindFlags(fs)
    return func(opts *Options) {
        encoderConfigOptions := []EncoderConfigOption{}
        if flags.timeEncoderVal.set {
            encoderConfigOptions = append(encoderConfigOptions, flags.timeEncoderVal.value)
        }
        // future encoder config flags would be configured here
        if flags.encoderVal.set {
            opts.Encoder = flags.encoderVal.newEncoderFunc(encoderConfigOptions)
        } else {
            opts.EncoderConfigOptions = append(opts.EncoderConfigOptions, encoderConfigOptions...)
        }
        if flags.levelVal.set {
            opts.Level = flags.levelVal.value
        }
        if flags.stacktraceLevelVal.set {
            opts.StacktraceLevel = flags.stacktraceLevelVal.value
        }
    }
}
```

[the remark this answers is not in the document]
bharathi-tenneti: As dicussed, commit has been made with below design approach:
-- Add `o.NewEncoder` field to Options struct, of type `type NewEncoderFunc func([]EncoderConfigOption) zapcore.Encoder`

pkg/log/zap/flags.go
```diff
@@ -60,26 +60,63 @@ func (ev *encoderFlag) Set(flagValue string) error {
...
-func newConsoleEncoder() zapcore.Encoder {
+func newConsoleEncoder(ecfs []EncoderConfigOption) zapcore.Encoder {
 	encoderConfig := zap.NewDevelopmentEncoderConfig()
+	for _, f := range ecfs {
+		f(&encoderConfig)
+	}
 	return zapcore.NewConsoleEncoder(encoderConfig)
 }
 
+type timeEncoderFlag struct {
```
estroz, line 89: Unless I'm missing something, this flag seems to be doing something that could be accomplished by an `encoderFlag` since a `zapcore.TimeEncoder` is a field on `zapcore.EncoderConfig` right?

pkg/log/zap/flags.go
```diff
@@ -60,26 +60,63 @@ func (ev *encoderFlag) Set(flagValue string) error {
...
+		ev.setFunc(newConsoleEncoder)
 	default:
 		return fmt.Errorf("invalid encoder value \"%s\"", flagValue)
 	}
 	ev.value = flagValue
 	return nil
 }
 
-func newJSONEncoder() zapcore.Encoder {
+func newJSONEncoder(ecfs []EncoderConfigOption) zapcore.Encoder {
```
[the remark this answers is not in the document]
bharathi-tenneti: When defined  as suggested , I am note able to call the `newJSONEncoder` func by Identifier.
```Go
	switch val {
	case "json":
		ev.setFunc(newJSONEncoder)
	case "console":
		ev.setFunc(newConsoleEncoder)
```

pkg/log/zap/flags.go
```diff
@@ -60,26 +60,63 @@ func (ev *encoderFlag) Set(flagValue string) error {
...
 	encoderConfig := zap.NewProductionEncoderConfig()
+	for _, f := range ecfs {
+		f(&encoderConfig)
+	}
 	return zapcore.NewJSONEncoder(encoderConfig)
 }
 
-func newConsoleEncoder() zapcore.Encoder {
+func newConsoleEncoder(ecfs []EncoderConfigOption) zapcore.Encoder {
```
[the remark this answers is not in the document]
bharathi-tenneti: https://github.com/kubernetes-sigs/controller-runtime/pull/915#discussion_r428811678

pkg/log/zap/flags.go
bharathi-tenneti: As far as i understand, What you said is true for providing `zapcore.TImeEncoder` programmatically. `timeEncoderFlag` parses from `--zap-time-encoder=millis` flag directly, and applies to the `--zap-encoder=console` flag and then create the encoder.

pkg/log/zap/zap.go
```diff
@@ -223,10 +255,21 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
...
+	// Set TimeEncoder value
+	var timeEncVal timeEncoderFlag
+	timeEncVal.setFunc = func(fromFlag zapcore.TimeEncoder) {
+		// Set o.EncoderConfigOptions with zap-time-encoder
+		f := func(ec *zapcore.EncoderConfig) {
+			ec.EncodeTime = fromFlag
+		}
+		o.EncoderConfigOptions = append(o.EncoderConfigOptions, f)
+	}
+	fs.Var(&timeEncVal, "zap-time-encoder", "Zap time encoding format: ('rfc3339', 'rfc339nanos', 'epoch', 'millis', 'nano', or 'iso8601')")
```
joelanford, line 267: Ditto: re: not hard-coding allowed encoder strings.

pkg/log/zap/zap.go
```diff
@@ -251,6 +294,7 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 // UseFlagOptions configures the logger to use the Options set by parsing zap option flags from the CLI.
 //  opts := zap.Options{}
 //  opts.BindFlags(flag.CommandLine)
+//  flag.Parse()
 //  log := zap.New(zap.UseFlagOptions(&opts))
 func UseFlagOptions(in *Options) Opts {
 	return func(o *Options) {
```
joelanford, line 300: Just noticed this now, but I don't think we should be calling `o.addDefaults()` in this function. That gets called later by `zap.New()` after ALL of the option functions have been called.

pkg/log/zap/flags.go
```diff
@@ -60,26 +60,63 @@ func (ev *encoderFlag) Set(flagValue string) error {
...
+	var encoder zapcore.TimeEncoder
+	val := strings.ToLower(flagValue)
+	switch val {
+	case "iso8601", "rfc3339", "rfc3339nano", "millis", "nanos", "epoch":
+		if err := encoder.UnmarshalText([]byte(val)); err != nil {
+			return fmt.Errorf("unable to unmarshal, \"%s\"", val)
+		}
+	default:
+		return fmt.Errorf("invalid value for time encoder, \"%s\"", val)
+	}
```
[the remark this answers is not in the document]
bharathi-tenneti: However, we would not be able to return error on `invalid` values explicitly, as the default is `EpochTimeEncoder` returned by `UnmarshalText`
https://github.com/uber-go/zap/blob/c9ca99951c74a6fcaf507962e772861e2218ce2e/zapcore/encoder.go#L174
I  believe that is how we wanted the design initially.  @hasbro17 WDYT?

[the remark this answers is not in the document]
estroz: That makes sense. I'm surprised an unmarshaler doesn't return an error for an invalid value.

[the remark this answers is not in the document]
bharathi-tenneti: True, But that still relies assuming `Unmarshaltext` would always return `epoch` as default in future as well. Either ways there is still dependency.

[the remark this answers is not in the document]
joelanford: I agree that it's weird that `UnmarshalText` never returns an error, but that seems to be an explicit choice of the zap maintainers. I guess my point of view is that we're basically plumbing the flag value through to the library and having it return what it thinks is the best thing to return.

So maybe it is reasonable to silently default to `epoch` when someone passes `--zap-time-encoder=foobar` because that lines up with what zap itself is doing.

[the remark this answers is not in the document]
hasbro17: I'm fine with defaulting to whatever UnmarshalText gives us.
And maybe we can mention that invalid values will result in the default zap time encoder in the flag help text.

pkg/log/zap/zap.go
```diff
@@ -213,6 +241,7 @@ func NewRaw(opts ...Opts) *zap.Logger {
 //  zap-devel: Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn)
 //			  Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
 //  zap-encoder: Zap log encoding ('json' or 'console')
+//  zap-time-encoder: Zap time encoding format, defaukts to 'epoch'.
```
estroz, line 244:
```suggestion
//  zap-time-encoder: Zap time encoding format, defaults to 'epoch'.
```

pkg/log/zap/zap.go
```diff
@@ -223,10 +252,18 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		"Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). "+
 			"Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)")
 
+	// Set TimeEncoder value
+	var timeEncVal timeEncoderFlag
+	timeEncVal.setFunc = func(fromFlag zapcore.TimeEncoder) {
+		// Set o.EncoderConfigOptions with zap-time-encoder
+		TimeEncoder(fromFlag)(o)
+	}
+	fs.Var(&timeEncVal, "zap-time-encoder", "Zap time encoding format, defaults to 'epoch'")
```
estroz, line 261: `--zap-time-encoder` and `--zap-encoder` are mutually exclusive right? Can you update the descriptions to include this info?

bharathi-tenneti: They are not designed to be mutually exclusive, `--zap-encoder` takes either console/json, and if `--zap-time-encoder` is given, then will set the Encoder with given time format.
Example, "--zap-encoder=json --zap-time-encoder=millis" will result in JSON encoder with `EpochMillisEncoder` set.

pkg/log/zap/zap.go
```diff
@@ -154,10 +180,10 @@ func (o *Options) addDefaults() {
 	}
 
 	if o.Development {
-		if o.Encoder == nil {
-			encCfg := zap.NewDevelopmentEncoderConfig()
-			o.Encoder = zapcore.NewConsoleEncoder(encCfg)
+		if o.NewEncoder == nil {
+			o.Encoder = newConsoleEncoder(o.EncoderConfigOptions...)
```
joelanford, line 184: I think this needs to set the `NewEncoder`, not `Encoder`, because if `Encoder` is already set, we'll overwrite it here.

```suggestion
			o.NewEncoder = newConsoleEncoder
```

pkg/log/zap/zap.go
```diff
@@ -169,10 +195,10 @@ func (o *Options) addDefaults() {
 		o.ZapOpts = append(o.ZapOpts, zap.Development())
 
 	} else {
-		if o.Encoder == nil {
-			encCfg := zap.NewProductionEncoderConfig()
-			o.Encoder = zapcore.NewJSONEncoder(encCfg)
+		if o.NewEncoder == nil {
+			o.Encoder = newJSONEncoder(o.EncoderConfigOptions...)
```
joelanford, line 199: Ditto here:
```suggestion
			o.NewEncoder = newJSONEncoder
```

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +300,128 @@ var _ = Describe("Zap logger setup", func() {
 		})
 	})
 })
+
+var _ = Describe("Zap flag options setup", func() {
+	fromFlags := Options{}
+
+	BeforeEach(func() {
+		flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError) //flags are now reset
```
joelanford, line 308: Nit: Rather than using `flag.CommandLine`, would it work to assign `flag.NewFlagSet()` to a separately declared `flag.FlagSet`, and then use the `.Parse()` method on that struct? If so, that seems like it could simplify things.

```go
var _ = Describe("Zap flag options setup", func() {
  var (
    fromFlags Options
    fs flag.FlagSet
  )

  BeforeEach(func() {
    fromFlags = Options{}
    fs = flag.NewFlagSet(...)
  })

  Context("with -zap-devel options provided", func() {
    It("Should set dev=true with relative options.", func() {
      args := []string{"-zap-devel=true"}
      fromFlags.BindFlags(fs)
      fs.Parse(args)
      ...
    })
  })
```

If you go this route, note that the arguments passed to the `.Parse()` method on the FlagSet should not contain the command name.

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +300,128 @@ var _ = Describe("Zap logger setup", func() {
...
+	})
+
+	Context("with -zap-devel options provided", func() {
+		It("Should set dev=true with relative options.", func() {
+			os.Args = []string{"cmd", "-zap-devel=true"}
+			fromFlags.BindFlags(flag.CommandLine)
+			flag.Parse()
+			out := Options{}
+			UseFlagOptions(&fromFlags)(&out)
+			out.addDefaults()
```
joelanford, line 318: Non-blocking nit: this unit test is testing both `UseFlagOptions` and `addDefaults()`. Could we skip the `out.addDefaults` call here, and then below, only check that `out.Development` is `true` since the other checks are checking what `addDefaults` does?

And then add a couple of tests for `addDefaults` where we're setting up the `Options` struct manually instead of parsing flags. That way we separate the concerns a bit.

Ditto in other tests that are testing just the flag parsing into options.

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +300,128 @@ var _ = Describe("Zap logger setup", func() {
...
+			Expect(out.Development).To(BeFalse())
+			Expect(out.Level.Enabled(zapcore.InfoLevel)).To(BeTrue())
+			Expect(out.StacktraceLevel.Enabled(zapcore.ErrorLevel)).To(BeTrue())
+			cfg := zap.NewProductionEncoderConfig()
+			Expect(out.Encoder).Should(BeAssignableToTypeOf(zapcore.NewJSONEncoder(cfg)))
+			Expect(out.EncoderConfigOptions).To(BeNil())
+		})
+	})
+
+	Context("with -zap-Encoder and -zap-Time-Encoder options provided.", func() {
```
joelanford, line 344: Nit: capitalization
```suggestion
	Context("with -zap-encoder and -zap-time-encoder options provided.", func() {
```

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +300,128 @@ var _ = Describe("Zap logger setup", func() {
...
+			fromFlags.BindFlags(flag.CommandLine)
+			flag.Parse()
+			out := Options{}
+			UseFlagOptions(&fromFlags)(&out)
+			out.addDefaults()
+
+			// Set logger with custom io.writer
+			logOut := new(bytes.Buffer)
+			sink := zapcore.AddSync(logOut)
+			log := zap.New(zapcore.NewCore(&KubeAwareEncoder{Encoder: out.Encoder, Verbose: out.Development}, sink, out.Level))
```
joelanford, line 357: Nit: instead of the above, would the following work?

```suggestion
			logOut := new(bytes.Buffer)
			log := New(UseFlagOptions(&fromFlags), WriteTo(logOut))
```

In this cases, I don't think we would even need the `out` variable.

Ditto elsewhere.

joelanford:
Also, based on @DirectXMan12's [comment here][0], I think we should add the `Type()` functions back in (leaving the interface assertion assignments out) so that the flags still implement `pflag.Value` as a convenience.

[0]: https://github.com/kubernetes-sigs/controller-runtime/pull/972/files#diff-b742ebc263492d7922597d3d5972d5bfR30-R31

bharathi-tenneti:
> Also, based on @DirectXMan12's [comment here](https://github.com/kubernetes-sigs/controller-runtime/pull/972/files#diff-b742ebc263492d7922597d3d5972d5bfR30-R31), I think we should add the `Type()` functions back in (leaving the interface assertion assignments out) so that the flags still implement `pflag.Value` as a convenience.

Does this mean, adding Something similar as below?
```Go
var _ pflag.Value = &levelFlag{}

func (ev *levelFlag) Type() string {
	return "level"
}
```

pkg/log/zap/zap.go
bharathi-tenneti: Same as above concern https://github.com/kubernetes-sigs/controller-runtime/pull/915#discussion_r435360010

pkg/log/zap/zap.go
```diff
@@ -154,9 +180,8 @@ func (o *Options) addDefaults() {
 	}
 
 	if o.Development {
-		if o.Encoder == nil {
-			encCfg := zap.NewDevelopmentEncoderConfig()
-			o.Encoder = zapcore.NewConsoleEncoder(encCfg)
+		if o.NewEncoder == nil {
+			o.NewEncoder = newConsoleEncoder
```
joelanford, line 184: Totally minor/non-blocking nit:

Move the `newConsoleEncoder` and `newJSONEncoder` functions into this file now that we're using them outside the context of flag parsing.

DirectXMan12:
/hold

can you elaborate on this a bit -- there's not much in the related bug, there's nothing in the PR description, and it's not quite clear why this is necessary as an always-on flag.

PR descriptions should *always* stand on their own as an explanation, regardless of corresponding issue.

bharathi-tenneti:
> /hold
> 
> can you elaborate on this a bit -- there's not much in the related bug, there's nothing in the PR description, and it's not quite clear why this is necessary as an always-on flag.
> 
> PR descriptions should _always_ stand on their own as an explanation, regardless of corresponding issue.

Sure, This PR is to enable existing encoder option with a way to encode time as well, using  `--zap-time-encoder` flag. This is an optional flag that developers can provide either using command line or programmatically. Provided a value, logger is set with the given time encoding options, which can be found [here](https://github.com/uber-go/zap/blob/master/zapcore/encoder.go#L161).

DirectXMan12:
Thank you for updating the PR.  I think my main question here is *why* do we need this flag?  Is customizing the timestamp format common enough to merit a flag?

joelanford:
@DirectXMan12 Here's the PR (authored by an operator developer) where we added this feature in Operator SDK. It looks like it was also used in another operator here: https://github.com/kiegroup/kogito-cloud-operator/pull/60

I guess the question is whether operator developers want to be able to customize (but still hardcode) the time encoding, or if they want to give users of their operator the ability to customize the time encoding.

bharathi-tenneti:
@DirectXMan12 

> @DirectXMan12 Here's the [PR](https://github.com/operator-framework/operator-sdk/pull/1529) (authored by an operator developer) where we added this feature in Operator SDK. It looks like it was also used in another operator here: [kiegroup/kogito-cloud-operator#60](https://github.com/kiegroup/kogito-cloud-operator/pull/60)
> 
> I guess the question is whether operator developers want to be able to customize (but still hardcode) the time encoding, or if they want to give users of their operator the ability to customize the time encoding.

@DirectXMan12  Do you have any other concerns?

bharathi-tenneti:
This PR is being re-worked to add Encoder Config Options , rather than specific time encoder flag, seen in the earlier implementations.

pkg/log/zap/zap_test.go
```diff
@@ -459,4 +459,81 @@ var _ = Describe("Zap log level flag options setup", func() {
...
+			UseFlagOptions(&fromFlags)(&out)
+
+			Expect(out.Development).To(BeTrue())
+			Expect(out.Encoder).To(BeNil())
+			Expect(out.Level).To(BeNil())
+			Expect(out.StacktraceLevel).To(BeNil())
+			Expect(out.EncoderConfigOptions).To(BeNil())
+		})
+		It("Should set dev=false with log options.", func() {
+			args := []string{"--zap-log-level=2"}
```
joelanford, line 479: This seems to be setting `args` without `-zap-devel`, but the context it is in says `with only -zap-devel flag provided`. That doesn't seem quite right.

pkg/log/zap/zap_test.go
```diff
@@ -459,4 +459,81 @@ var _ = Describe("Zap log level flag options setup", func() {
...
+
+			Expect(out.Development).To(BeFalse())
+			Expect(out.Level.Enabled(zapcore.ErrorLevel)).To(BeTrue())
+
+		})
+	})
+
+	Context("with encoder options provided programmatically.", func() {
+
+		It("Should set Console Encoder, with given Nanos TimEncoder option.", func() {
```
joelanford, line 495:
```suggestion
		It("Should set Console Encoder, with given Nanos TimeEncoder option.", func() {
```

pkg/log/zap/zap_test.go
```diff
@@ -459,4 +459,81 @@ var _ = Describe("Zap log level flag options setup", func() {
...
+			log.Info("This is a test message")
+			outRaw := logOut.Bytes()
+			// Assert for Console Encoder
+			res := map[string]interface{}{}
+			Expect(json.Unmarshal(outRaw, &res)).ToNot(Succeed())
+			// Assert for Epoch Nanos TimeEncoder
+			Expect(string(outRaw)).ShouldNot(ContainSubstring("."))
+
+		})
+		It("Should set JSON Encoder, with given Millis TimEncoder option, and MessageKey", func() {
```
joelanford, line 515:
```suggestion
		It("Should set JSON Encoder, with given Millis TimeEncoder option, and MessageKey", func() {
```

joelanford:
This looks pretty good to me now based on our convo in the community meeting last time.

@DirectXMan12 Does this align with what you were thinking now?

alvaroaleman:
/hold cancel
based on Joes approval and sollys lgtm (the /hold was originally added by Solly)

Issue #918, opened 2020-04-22
Fake client Create call changes the object even if an error returned


muvaf:
`Create` call normally does not change the local object if the result is error. However, fake client [does set its `resourceVersion` to `1`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L98) in all cases.

PR #919, opened 2020-04-22
🐛 Reset resource version if fake client Create call failed

muvaf:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/918

This breaks the code that relies on first checking if an object already exists via the error returned from `Create` and then run `Patch`. Since `Patch` call of fake client does an `Update` operation, which checks resource version match, it returns `object was changed` error in these cases because the local object now has resource version 1 but what storage has is resource version 0 since the client was initialized with that object.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -96,7 +96,11 @@ func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Ob
 		return apierrors.NewBadRequest("resourceVersion can not be set for Create requests")
 	}
 	accessor.SetResourceVersion("1")
-	return t.ObjectTracker.Create(gvr, obj, ns)
+	if err := t.ObjectTracker.Create(gvr, obj, ns); err != nil {
+		accessor.SetResourceVersion("")
+		return err
+	}
+	return nil
 }
 
 func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Object, ns string) error {
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -186,6 +186,14 @@ var _ = Describe("Fake client", func() {
 			Expect(apierrors.IsBadRequest(err)).To(BeTrue())
 		})
 
+		It("should not change the submitted object if Create failed", func() {
+			By("Trying to create an existing configmap")
+			submitted := cm.DeepCopy()
+			err := cl.Create(context.Background(), submitted)
+			Expect(apierrors.IsAlreadyExists(err)).To(BeTrue())
+			Expect(submitted).To(Equal(cm))
+		})
+
 		It("should error on Create with empty Name", func() {
 			By("Creating a new configmap")
 			newcm := &corev1.ConfigMap{


djzager:
As I understand the scenario being described:

- user calls `NewFakeClient([]runtime.Object{myObject})`, it's unlikely that the user would set the resource version here so the **stored** version is `""`
- user calls `Create(gvr, myObject, myObject.Namespace)`, this returns an error because `myObject` already exists
- user calls `Update(gvr, myObject, myObject.Namespace)`, this fails because https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L121-L123

Assuming that this change addresses this scenario, what happens if you started the client with an object, made a series of `Update(...)` calls, called `Create(...)` which would fail because the object already exists. This change would not address this scenario as the stored `objectMeta.ResourceVersion != ""`.

My personal feeling is that there is not a guarantee to be made in the case that you call `Create(...)` that a subsequent call to `Update(...)` with the same object should work. Namely, I believe if you wish to call `Update(...)` you should first `Get(...)`, modify that object, and then call `Update(...)`. WDYT?

muvaf:
> Assuming that this change addresses this scenario, what happens if you started the client with an object, made a series of Update(...) calls, called Create(...) which would fail because the object already exists? This change would not fix a subsequent call to Update(...) in this case as the stored objectMeta.ResourceVersion != "".

Each `Update` increases the resource version and updates the local object you submitted, so `Create` will fail saying `resource version cannot be set in create` and `Update`s should work since the local object is same as what's stored.

> My personal feeling is that there is not a guarantee to be made in the case that you call Create(...) that a subsequent call to Update(...) with the same object should work. Namely, I believe if you wish to call Update(...) you should first Get(...), modify that object, and then call Update(...). WDYT?

Actually, the problem is not about `Update`. In my case, `Create` is called and if the resource exists I call `Patch`, which normally doesn't require you to know resource version at all. However, fake client uses `Update` under the hood when `Patch` is called. What it does is bascially applying the patch locally and then call `Update` and in that process `resourceVersion` suddenly becomes important. I think this process is problematic as well but this PR aims to fix only `Create` to match the existing real client behavior, i.e. object should not be modified if the call fails.

djzager:
Thank you for the explanation.

> Actually, the problem is not about `Update`. In my case, `Create` is called and if the resource exists I call `Patch`, which normally doesn't require you to know resource version at all. However, fake client uses `Update` under the hood when `Patch` is called.

Looking at the current implementation of `Patch` I see nothing to indicate that `Update` is being used "under the hood". Is it possible that you are using an older version of controller runtime?

> I think this process is problematic as well but this PR aims to fix only `Create` to match the existing real client behavior, i.e. object should not be modified if the call fails.

I am a little lost here, my apologies, but this PR is modifying the object when `Create` fails (ie. `accessor.SetResourceVersion("")`).

muvaf:
No problem!

> Looking at the current implementation of Patch I see nothing to indicate that Update is being used "under the hood". Is it possible that you are using an older version of controller runtime?

[Here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L349) `testing.ObjectReaction` is called and `testing.NewPatchAction` is supplied in the next call. When you look at the implementation of [`testing.ObjectReaction`](https://github.com/kubernetes/client-go/blob/release-1.18/testing/fixture.go#L77) (client-go v0.18.0 is used by controller-runtime master.) you'll see that [`case PatchActionImpl`](https://github.com/kubernetes/client-go/blob/release-1.18/testing/fixture.go#L133) does the patch operation via json data and call [`Update`](https://github.com/kubernetes/client-go/blob/release-1.18/testing/fixture.go#L184).

> I am a little lost here, my apologies, but this PR is modifying the object when Create fails (ie. accessor.SetResourceVersion("")).

If you look at a few lines before the actual `Create` call, [it requires `resourceVersion` to be empty](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/client/fake#L89) even to make it there. So, we know that it was `""`, then `Create` changes it to `"1"`. If call fails, it stays `"1"`. What this PR does is to revert it back to `""` since the call failed.

djzager:
Now I got it. Thanks.

/lgtm

muvaf:
It seems to be a flake.

/retest

muvaf:
@shawn-hurley thanks for the test! Seems like this is ready to merge.

djzager:
@muvaf One more thing. I think it would be appropriate for you to add a test to `pkg/client/fake/client_test.go` that verifies the resourceVersion is unchanged after failed Create.

muvaf:
> I think it would be appropriate for you to add a test to pkg/client/fake/client_test.go that verifies the resourceVersion is unchanged after failed Create.

@djzager @shawn-hurley  Added a test that verifies the submitted object is not changed after a failed `Create` call.

djzager:
Thanks for the fix @muvaf

PR #921, opened 2020-04-23
🏃 Add a WithoutLeaderElection option to controllers

qinqon:
It will run the controllers with option `WithoutLeaderElection` set to `true` like the webhooks ignoring leader election in case manager is configured with LeaderElection. 

`WithoutLeaderElection` is choosen instead of `WithLeaderElection` to use the default bool value ot false so we don't need to discover if it's set by the user or not.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -41,6 +41,10 @@ type Options struct {
 	// Defaults to MaxOfRateLimiter which has both overall and per-item rate limiting.
 	// The overall is a token bucket and the per-item is exponential.
 	RateLimiter ratelimiter.RateLimiter
+
+	// WithoutLeaderElection is used to bypass manager leader election in case
+	// it's activated, by default leader election is activated for controllers.
+	WithoutLeaderElection bool
 }
 
 // Controller implements a Kubernetes API.  A Controller manages a work queue fed reconcile.Requests
@@ -109,6 +113,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
+		WithoutLeaderElection:   options.WithoutLeaderElection,
 	}
 
 	return c, nil
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -71,6 +71,9 @@ type Controller struct {
 	// Started is true if the Controller has been Started
 	Started bool
 
+	// Flag to indicate if this controller need leader election or not
+	WithoutLeaderElection bool
+
 	// TODO(community): Consider initializing a logger with the Controller Name as the tag
 
 	// watches maintains a list of sources, handlers, and predicates to start when the controller is started.
@@ -266,6 +269,11 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 func (c *Controller) InjectFunc(f inject.Func) error {
 	c.SetFields = f
 	return nil
+}
+
+// NeedLeaderElection implements LeaderElectionRunnable
+func (c *Controller) NeedLeaderElection() bool {
+	return !c.WithoutLeaderElection
 }
 
 // updateMetrics updates prometheus metrics within the controller
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -74,6 +74,20 @@ var _ = Describe("controller", func() {
 		close(stop)
 	})
 
+	Describe("NeedLeaderElection", func() {
+		It("should return true by default", func() {
+			Expect(ctrl.NeedLeaderElection()).To(BeTrue())
+		})
+		It("should be true WithoutLeaderElection is false", func() {
+			ctrl.WithoutLeaderElection = false
+			Expect(ctrl.NeedLeaderElection()).To(BeTrue())
+		})
+		It("should be false WithoutLeaderElection is true", func() {
+			ctrl.WithoutLeaderElection = true
+			Expect(ctrl.NeedLeaderElection()).To(BeFalse())
+		})
+	})
+
 	Describe("Reconciler", func() {
 		It("should call the Reconciler function", func() {
 			ctrl.Do = reconcile.Func(func(reconcile.Request) (reconcile.Result, error) {


alvaroaleman:
@qinqon do you have a sample usecase for running the manager with LeaderElection but some of the controllers without? Controllers specifically are usually build with the assumption that they are not run concurrently.

alvaroaleman:
> from the kubevirt team we have a pair of projects that use controller-runtime without leader election activated, but we are going to implement a cert-manager as a new controller and that needs to run with leader election.

I see. Can you please add tests that verify that
* Controller defaults to `NeedLeaderElection`
* Switching the bool changes the return of `NeedLeaderElection` ?

/ok-to-test

qinqon:
Will do thanks

qinqon:
@alvaroaleman how to I handle incompatible api changes ? 
```
sigs.k8s.io/controller-runtime/pkg/controller
  Incompatible changes:
  - Controller.NeedLeaderElection: added
  Compatible changes:
  - Options.WithoutLeaderElection: added
```

vincepri:
I'm going to defer to @DirectXMan12 on this one, I'm not 100% sure about having controllers behaving differently w.r.t. leader election in the same manager. 

If leader election is lost for example, the whole manager pod will crash and go in backoff. In production workloads I'd expect controller not needing leader election to be living in a different deployment/manager.

alvaroaleman:
> If leader election is lost for example, the whole manager pod will crash and go in backoff. In production workloads I'd expect controller not needing leader election to be living in a different deployment/manager.

That's a good point. My thinking here was that if the given controller does not need leaderElection, it will run in all replicas, even the ones that do not have acquired a leader lease (Which implies that the controller-manager is run with replicas > 1).

alvaroaleman:
The only thing I am not 100% convinced about is running a controller without leader election ever. The concept of "no more than routine working on a given object" is very much baked into the underlying machinery by doing per-key locking.

qinqon:
Thanks for reviewing! 
Just as a note we are using succefully controller-runtime without leader elections at https/GitHub.com/nmstate/kubernetes-nmstate, I don't see the issue with mixing leader election and not leader election aware controllers.

DirectXMan12:
I can see circumstances where you'd want this -- for example, if you had some way to pre-shard your dataset, but most of them in my head would work slightly better as "no leader election for the entire controller".  Just so we make sure we're thinking about this correctly, can you elaborate a bit on your use case?  In general, I'm inclined to let this in, just want to make sure we can document it well, etc.

fengye87:
I'm facing a case like this right now, so l thought I might give an example here.

In my case, I have 2 controllers in one manager. One controller need to operate something on each node locally, so I'm deploying the manager as a DaemonSet. However, at the same time, the other controller is a global one, it operates at the cluster level and thus needs leader election to avoid conflicts.

Without the ability to turn leader election on/off for each single controller (which I think is the case right now), I would have to separate these 2 controllers into two parties, one being a Deployment and the other still being a DaemonSet. Although I could make them share the same binary & Docker image, but they could not share cache, and it involves more moving parts.

@DirectXMan12

qinqon:
Our scenario is same as @fengye87 we have this project github.com/kubernetes-nmstate it uses controller-runtime to run the manager as DaemonSet to be able to apply networking changes at all nodes but in a distribution fashion so we can still rollback per node in case of bad networking config, it also has some custom small cert-manager for webhooks and that cert-manager has to run with leader election.

qinqon:
@DirectXMan12 the scenarios feel good to you ?

fengye87:
@DirectXMan12 It has been over a week since your last comment, anything I could help?

qinqon:
Rebased @alvaroaleman can you add LGTM again ?

fengye87:
The two binary approach surely works, but is not ideal. For my use case, most logics lie in the DaemonSet part, with only a few global operations that I need them to have a leader to do that. I can certainly separate them with two binaries, one being solely DaemonSet and the other being solely leader thing. But that would be two moving parts for me, which might be a burden to carry considering my global operations are very light and occasional. Besides, the two binary could not share cache, which would also introduce more memory usage.

And to your "because you don't need to find the one DaemonSet pod that has the leader lock" point, I think it applies to Deployment too if its replicas need leader elections?

@alvaroaleman

qinqon:
@alvaroaleman our use case is similar to @fengye87 the main logic is at a leaderless manager withing DaemonSet and we have a small webhook that do some optional validations (it can even be disable) and we generate the TLS certs/keys at a LeaderElectionRunnable that's why we need to mix non leader-aware controllers with a leader aware manager.

At the end we have splitted into DaemonSet/Deployment (for other reasons not related to this) but using the same binary, just passing a env var to decide the role of the binary, but still I see that users need to be the ones deciding on splitting or not.

fengye87:
@qinqon Looks like there's conflicting files.

@alvaroaleman I'm still waiting for this feature, anything I can do to help?

qinqon:
@fengye87 done

alvaroaleman:
> And to your "because you don't need to find the one DaemonSet pod that has the leader lock" point, I think it applies to Deployment too if its replicas need leader elections?

Usually you don't run such deployments with multiple replicas, it just has multiple replicas during upgrade or in some other corner cases.

Well tbh I am not completely convinced. If we merge this, we basically make the approach blessed but I am not so convinced its a good idea.

vincepri:
/hold

Going to hold the PR to discuss it at the next community meeting.

qinqon:
@vincepri did we have news on this ?

vincepri:
@qinqon I wasn't able to attend last week's meeting, I'm not sure what was the outcome

alvaroaleman:
@qinqon this wasn't discussed because no one brought it up. To expand a little of the above points why I am not convinced this is a good idea:
* All prior art I am aware of in this space uses a combination of a Deployment and a DaemonSet (e.G. [calico](https://docs.projectcalico.org/v3.14/manifests/calico.yaml), [cilium](https://github.com/cilium/cilium/blob/master/install/kubernetes/quick-install.yaml), [gcp csi](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/tree/master/deploy/kubernetes/base) )
* Coupling this in one binary means you can not set appropriate requests because whats needed will differ depending on if that binary got the leader lease or not
* Rolling out a controller updates on larger clusters might take long (or short, depending on how lucky you are)
* Finding the one DaemonSet pod that got the leader lease will be a huge PITA

If you find none of this convincing, a downstream wrapper to get the functionality proposed here is only a couple of lines code:
```go
type nonLeaderLeaseNeedingController struct {
  controller.Controller
}

func (nonLeaderLeaseNeedingController) NeedLeaderElection() bool {
  return false
}
```

Adding functionality to controller-runtime sends a message that using that functionality is a good idea. That is IMHO not the case for the proposed change here. Combined with the fact that its trivial to get this via a downstream wrapper if you really want, I am  opposed to merging this.

qinqon:
@alvaroaleman thanks for the final feedback, since the wrapper is easy and also this PR is not very difficult to re-create in case of someone else is needed I am going to close this.

Thanks a lot for all the reviews.

PR #922, opened 2020-04-23
:sparkles: Add ContainsFinalizer helper to the controllerutil

### LINKED ISSUE #920: Add ContainsFinalizer helper to the controllerutil
It would be nice to have ContainsFinalizer helper in the controllerutil in advance to the RemoveFinalizer and AddFinalizer.
It prevents to write you own code or use third-party libraries for this.

dmvolod:
This is ContainsFinalizer helper in the controllerutil package which checks that provided finalizer exists in the Object meta

Fixes #920

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -282,6 +282,17 @@ func RemoveFinalizerWithError(o runtime.Object, finalizer string) error {
 	return nil
 }
 
+// ContainsFinalizer checks a metav1 object that the provided finalizer is present.
+func ContainsFinalizer(o Object, finalizer string) bool {
+	f := o.GetFinalizers()
+	for _, e := range f {
+		if e == finalizer {
+			return true
+		}
+	}
+	return false
+}
+
 // Object allows functions to work indistinctly with any resource that
 // implements both Object interfaces.
 type Object interface {
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -452,6 +452,18 @@ var _ = Describe("Controllerutil", func() {
 				Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{}))
 			})
 		})
+
+		Describe("ContainsFinalizer", func() {
+			It("should check that finalizer is present", func() {
+				controllerutil.AddFinalizer(deploy, testFinalizer)
+				Expect(controllerutil.ContainsFinalizer(deploy, testFinalizer)).To(Equal(true))
+			})
+
+			It("should check that finalizer is not present after RemoveFinalizer call", func() {
+				controllerutil.RemoveFinalizer(deploy, testFinalizer)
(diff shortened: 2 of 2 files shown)


dmvolod:
@joelanford, @alexeldeib  could you please to have a look at this changes.

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -446,6 +446,18 @@ var _ = Describe("Controllerutil", func() {
 				Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{}))
 			})
 		})
+
```
joelanford, line 449: Can you also add a test for `ContainsFinalizerWithError`, similar to the ones in lines 413-423?

dmvolod: @joelanford , thanks, fixed

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -281,3 +281,24 @@ func RemoveFinalizerWithError(o runtime.Object, finalizer string) error {
...
+		if e == finalizer {
+			return true
+		}
+	}
+	return false
+}
+
+// ContainsFinalizerWithError checks a metav1 object that the provided finalizer is present.
+// It returns an error if the provided object cannot provide an accessor.
+func ContainsFinalizerWithError(o runtime.Object, finalizer string) (bool, error) {
```
alvaroaleman, line 298: Why is this needed? All concrete api types, be they built-in or CRDs, have am emedded `metav1.ObjectMeta` and thus fulfill the `metav1.Object` interface. If you really end up writing generic code that acts on a `runtime.Object` you can just assert it as `metav1.Object` prior to trying to find out if it has the given finalizer.

The only type of `runtime.Object` where this will not work is list types, but that makes sense as they do not implement the `metav1.Object` interface. IMHO it would be better to tell ppl at compile time that this isn't possible for Lists rather than at runtime when `meta.Accessor` errors.

joelanford: I asked for this to be added to align with the other existing `*WithError` functions in this package.

I agree, though, that compile time type checking is better than runtime type assertions. #898 may help with this.

alvaroaleman: #898 merged so maybe we should remove `ContainsFinalizerWithError` again and update `ContainsFinalizer` to take an `controllerutil.Object` as first agument?

dmvolod: @alvaroaleman , thanks, I will do. Do we need also delete other *WithError methods and replace with controllerutil.Object methods which are not handling errors?

alvaroaleman: For now, please leave the existing code as it is. Removing the `WithError` funcs that we already have is a breaking change, so we should do it in a dedicated PR to make sure it gets its own release note and only gets merged when we are ready to take breaking changes.

dmvolod: Fixed

alvaroaleman: Thank you!
/lgtm

PR #923, opened 2020-04-23
typos

AdheipSingh:
📖 
comments

### DIFF
--- a/examples/builtins/validatingwebhook.go
+++ b/examples/builtins/validatingwebhook.go
@@ -34,7 +34,7 @@ type podValidator struct {
 	decoder *admission.Decoder
 }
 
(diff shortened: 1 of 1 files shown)


examples/builtins/validatingwebhook.go
```diff
@@ -34,7 +34,7 @@ type podValidator struct {
 	decoder *admission.Decoder
 }
 
-// podValidator admits a pod iff a specific annotation exists.
+// podValidator admits a pod if a specific annotation exists.
```
rajathagasthya, line 37: I don't know if this is a typo. "iff" is an abbreviation for "if and only if". Looks intended in this context.

AdheipSingh: oh i was not aware of this :)

Issue #935, opened 2020-05-02
Allow removing individual informers from the Cache


shomron:
## Overview

In a somewhat similar use-case to that described in #863, the [Gatekeeper](https://github.com/open-policy-agent/gatekeeper) project uses one set of custom resources to describe (and dynamically register) another set of CRDs. Once subordinate CRDs are created, we then use informers to register watches against resources of those types. The set of subordinate resource types change over time. 

The existing APIs in controller-runtime do not allow removal of informers once they have been created. To work around this limitation, we have [forked](https://github.com/open-policy-agent/gatekeeper/pull/490) the controller-runtime cache.

This issue tracks the effort to upstream our changes in a way that is aligned with the design goals of controller-runtime.

## Goals:

* Enable removal of individual informers from a Cache
* Support more granular lifecycle management for informers to allow them to be stopped during removal

## Non-goals

* Reference counting of informers

## Open questions

* Should existing interfaces be expanded (`Cache`, `Informers`) in a breaking manner, or should new interfaces be introduced?

camilamacedo86:
Hi @shomron, 

Could you please let us know if it still a need? Shows that yes, 

You said that you forked controller-runtime and applied your needs on it. Why you did not push a PR against master with your suggestions? Would it be possible? Maybe it can also clarify better your need and required changes that you are asking here. 

If it is too significant changes in its design, wdyt about push a PR with a doc design proposal? See https://github.com/kubernetes-sigs/controller-runtime/tree/master/designs. 

And then, could you also provide the use cases that this RFE would address?

PR #936, opened 2020-05-02
✨ Allow removing individual informers from the cache (#935)

shomron:
This change adds support for removing individual informers from the
cache using the new Remove() method. This is allowed before or after the
cache has been started.  Informers are stopped at the time of removal -
once stopped, they will no longer deliver events to registered event
handlers, and registered watched will be aborted.

Also adds non-blocking API for getting informers without waiting for their
cache to sync - GetInformerNonBlocking().

Currently marked as :sparkles: but depending on whether we decide to extend existing interfaces, may become breaking ⚠️ . 

This PR is introduced without comprehensive tests to give an opportunity to iron out the API with the maintainers.

Signed-off-by: Oren Shomron <shomron@gmail.com>

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -126,7 +126,7 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 	// Construct a new Mapper if unset
 	if opts.Mapper == nil {
 		var err error
-		opts.Mapper, err = apiutil.NewDiscoveryRESTMapper(config)
+		opts.Mapper, err = apiutil.NewDynamicRESTMapper(config)
 		if err != nil {
 			log.WithName("setup").Error(err, "Failed to get API Group-Resources")
 			return opts, fmt.Errorf("could not create RESTMapper from config")
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -70,6 +70,9 @@ type MapEntry struct {
 
 	// CacheReader wraps Informer and implements the CacheReader interface for a single type
 	Reader CacheReader
+
+	// Stop can be used to stop this individual informer without stopping the entire specificInformersMap.
+	stop chan struct{}
 }
 
 // specificInformersMap create and caches Informers for (runtime.Object, schema.GroupVersionKind) pairs.
@@ -121,6 +124,17 @@ type specificInformersMap struct {
 	namespace string
 }
 
+// Start starts the informer managed by a MapEntry.
+// Blocks until the informer stops. The informer can be stopped
+// either individually (via the entry's stop channel) or globally
+// via the provided stop argument.
+func (e *MapEntry) Start(stop <-chan struct{}) {
+	// Stop on either the whole map stopping or just this informer being removed.
+	internalStop, cancel := anyOf(stop, e.stop)
+	defer cancel()
+	e.Informer.Run(internalStop)
+}
+
 // Start calls Run on each of the informers and sets started to true.  Blocks on the stop channel.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
 func (ip *specificInformersMap) Start(stop <-chan struct{}) {
@@ -132,8 +146,8 @@ func (ip *specificInformersMap) Start(stop <-chan struct{}) {
 		ip.stop = stop
 
 		// Start each informer
-		for _, informer := range ip.informersByGVK {
-			go informer.Informer.Run(stop)
+		for _, entry := range ip.informersByGVK {
+			go entry.Start(stop)
 		}
 
 		// Set started to true so we immediately start any informers added later.
@@ -183,8 +197,12 @@ func (ip *specificInformersMap) Get(ctx context.Context, gvk schema.GroupVersion
 
 	if started && !i.Informer.HasSynced() {
 		// Wait for it to sync before returning the Informer so that folks don't read from a stale cache.
-		if !cache.WaitForCacheSync(ctx.Done(), i.Informer.HasSynced) {
-			return started, nil, apierrors.NewTimeoutError(fmt.Sprintf("failed waiting for %T Informer to sync", obj), 0)
+		// Cancel for context, informer stopping, or entire map stopping.
+		syncStop, cancel := mergeChan(ctx.Done(), i.stop, ip.stop)
+		defer cancel()
+		if !cache.WaitForCacheSync(syncStop, i.Informer.HasSynced) {
+			// Return entry even on timeout - caller may have intended a non-blocking fetch.
+			return started, i, apierrors.NewTimeoutError(fmt.Sprintf("failed waiting for %T Informer to sync", obj), 0)
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


shomron:
@vincepri what's the best way to get some 👀 on this issue and decide how to proceed?

shawn-hurley:
/ok-to-test

Sorry if I am missing some context, do we have an issue that describes the need for this? Sorry if I am out of the loop a little, just trying to link through when this would be used

shomron:
@shawn-hurley yes, the linked issue is #935 and the use case is described there. Let me know if you have any questions, I'd be happy to elaborate further.

pkg/cache/cache.go
```diff
@@ -126,7 +126,7 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 	// Construct a new Mapper if unset
 	if opts.Mapper == nil {
 		var err error
-		opts.Mapper, err = apiutil.NewDiscoveryRESTMapper(config)
+		opts.Mapper, err = apiutil.NewDynamicRESTMapper(config)
```
alvaroaleman, line 129: This is a good change but should be its own PR as it doesn't really have anything to do with this PR

alvaroaleman:
@shomron this means that controllers that use the given informer for their watch will silently degrade, correct?

shomron:
> this means that controllers that use the given informer for their watch will silently degrade, correct?

@alvaroaleman yes that is correct. In our use case, we have a primary controller multiplexing watch events from dynamically watched resources and forwarding them on to interested parties using a stable `source.Channel`-based watch. That is also the point where we do things like reference counting for interested parties.

pkg/cache/cache.go
shomron: Fair enough, I will move this into a separate PR. 👍🏻

alvaroaleman:
>@alvaroaleman yes that is correct. In our use case, we have a primary controller multiplexing watch events from dynamically watched resources and forwarding them on to interested parties using a stable source.Channel-based watch. That is also the point where we do things like reference counting for interested parties.

Okay. So from what I can see on a quick look, it is currently not possible to find out if an Informer has event handlers registered (correct me if I missed something there). This means that there is no way for us to prevent removal of informers that are being used in a watch by a controller and that in turn means that the implementation in this PR introduces a huge footgun.

That being said, it would be great to support your usecase, we've had something similiar as well in the past. What prevents you from creating a cache per watch?

cc @thetechnick

shomron:
@alvaroaleman thank you for looking at this! Let me try to address your points:

Regarding checking whether the informer has registered event handlers - in our case we know the informer has registered event handlers, and explicitly want to stop watching. I think the issue is not introspection, but that event handlers cannot be removed from informers. Tangentially, as I called out in https://github.com/kubernetes-sigs/controller-runtime/pull/863#issuecomment-620696147 - it is now possible to stop an unmanaged controller with registered event handlers - which I believe would cause its work queue to grow unbounded.

> What prevents you from creating a cache per watch?

This seems less than ideal. The role of the cache is to manage the lifecycle of informers over disparate resources and provide reader/writer abstractions on top. If we created a cache-per-resource, we would end up implementing a cache of caches, where each cache manages a single informer. I feel we could achieve a more concise design.

Thoughts?

alvaroaleman:
> Regarding checking whether the informer has registered event handlers - in our case we know the informer has registered event handlers, and explicitly want to stop watching. I think the issue is not introspection, but that event handlers cannot be removed from informers. Tangentially, as I called out in #863 (comment) - it is now possible to stop an unmanaged controller with registered event handlers - which I believe would cause its work queue to grow unbounded.

True, deregistering is the better solution. My point here is that this introduces a silent and hard to debug failure mode for everyone in order to fix a in the grand scheme of things rather exotic use case. And that is in my opinion not a good trade-off.

Maybe we can add refcounting ourselves on top of the underlying informer and refuse removing them if they have registered eventhandlers? Potentially with the option to say "I know this has N eventhandlers but this is okay because I am also stopping the controller?"

This is far from an elegant solution but from what I can see the best we can do without upstream changes. Maybe it would be a good idea to also start a discussion with upstream sig-apimachinery folks if they would consider adding an option to remove event handlers?

Tangentially, do you know if its safe to stop individual informers or if that might leak goroutines and/or memory?

> This seems less than ideal. The role of the cache is to manage the lifecycle of informers over disparate resources and provide reader/writer abstractions on top. If we created a cache-per-resource, we would end up implementing a cache of caches, where each cache manages a single informer. I feel we could achieve a more concise design.

I completely agree that this isn't an ideal solution, its a workaround. The trade-off that is in this PR right now (big footgun for everyone to support a somewhat uncommon usecase) is in my opinion not a good one. But I'd love for others to weight in on that.

/cc @DirectXMan12 @mengqiy @alexeldeib

alexeldeib:
I'm torn on this. I think the use case totally makes sense. I also agree with everything Alvaro said about it being a footgun and possible solutions. 

I'm curious what other API machinery folks think about (de)registering eventhandlers, as that's definitely the most pure solution albeit probably very non-trivial. I'd probably prefer doc'ing a cache of caches as a short term workaround for this case and fixing the eventhandlers long term to enable this scenario if I had full choice.

third opinions?

alexeldeib:
I ran into https://docs.google.com/document/d/1Wi3LM3sG6Qgfzm--bWb6R0SEKCkQCCt-ene6cO62FlM/edit?usp=sharing which touches on this use case

vincepri:
Having read the whole discussion, it might be worth to take a step back and do a small proposal following the examples we already have in docs/proposals.

In general, the use case is valid. That said, we need to make sure we don't open a whole set of potential issues that weren't possible before. I'd like to see a way for us to support this feature while leaning on the side of safety and explicit user configuration/intent.

pkg/cache/internal/informers_map.go
```diff
@@ -121,6 +124,17 @@ type specificInformersMap struct {
 	namespace string
 }
 
+// Start starts the informer managed by a MapEntry.
+// Blocks until the informer stops. The informer can be stopped
+// either individually (via the entry's stop channel) or globally
+// via the provided stop argument.
+func (e *MapEntry) Start(stop <-chan struct{}) {
+	// Stop on either the whole map stopping or just this informer being removed.
+	internalStop, cancel := anyOf(stop, e.stop)
```
DirectXMan12, line 133: #1116 should help here

DirectXMan12 (commented):
I've been discussing a usecase with some folks around this.  The general gist is that that a given CRD may or may not exist on a cluster, and you want to gracefully degrade when it's not present, but you want to run a controller while it is.  In various cases you don't actually want to restart the manager, so it's useful to be able to say "run this controller and equivalent informers while a resource is available, otherwise don't".

All that being said, a decent first step for this might be to keep it internal, use it for that use case, and work on refcounting or something in parallel.

kevindelgado:
Ahead of tomorrow's community meeting, I've been working on a more formal design doc for this as requested above. See #1192

DirectXMan12:
gonna close this for the moment, we'll roll it into whatever we decide to do with the stoppable controllers work and give attribution.

PR #938, opened 2020-05-06
:running: Remove unsed Config attribute from controller

alvaroaleman:
The controller contains the unused Config attribute. Since its exported,
linters will not complain about it, but since its in an internal
package, it is not part of our public api.


alvaroaleman:
/assign @vincepri 
apidiff is actually smart enough to notice that this is not an external api, nice.

PR #940, opened 2020-05-09
:running: Remove unused attributes from controller

alvaroaleman:
This removes a set of unused attributes from the controller.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -102,10 +102,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 
 	// Create controller with dependencies set
 	c := &controller.Controller{
-		Do:       options.Reconciler,
-		Scheme:   mgr.GetScheme(),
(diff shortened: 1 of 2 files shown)


vincepri:
@DirectXMan12 Any reason why these fields were included in the first place?

DirectXMan12:
IIRC, it has to do with injection into the reconciler -- at least Scheme was commonly injected as opposed to directly being passed in, at least in the early days.  We'll prob want to figure something out before 1.0.0 wrt that.

Right now, we prob don't want to remove those without double-checking that the corresponding injection still works

alvaroaleman:
@DirectXMan12 they do work, otherwise this PR wouldn't compile as the corresponding attributes are gone.

The injection is done by just storing the `SetFields` from the manager and re-using it: https://github.com/kubernetes-sigs/controller-runtime/pull/940/files#diff-3c862eb54a8e0e161b534e0c67e5379eL113

PR #941, opened 2020-05-10
✨ Controller: Let sources sync even if they use a different cluster

alvaroaleman:
Currently, the controller gets a cache injected and waits for that cache
to sync before starting workers. This model works well for single
cluster usecases, but doesn't work at all for multi-cluster usecases, as
the source may be using a completely different cache.

This PR adds a new source.SyncingSource interface and makes the
controller wait for all sources that implement it. Apart from solving
the issue mentioned in the title, this also allows to implement custom
sources that need syncing.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -103,7 +103,6 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 	// Create controller with dependencies set
 	c := &controller.Controller{
 		Do:       options.Reconciler,
-		Cache:    mgr.GetCache(),
 		Scheme:   mgr.GetScheme(),
 		Client:   mgr.GetClient(),
 		Recorder: mgr.GetEventRecorderFor(name),
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
-	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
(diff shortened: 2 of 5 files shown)


alvaroaleman:
/retest
The tests in `pkg/internal/testing/integration/internal/process_test.go` seem to be pretty flaky.

alvaroaleman:
I don't really understand why the apidiff presubmit errors, all changes in this PR are backwards-compatible.
/assign @vincepri @alexeldeib

vincepri:
🤔 

```
  Incompatible changes:
  - Func: no longer implements Source
  - Source: changed from Source to Source
  - SyncingSource.WaitForSync: added
```

alvaroaleman:
@vincepri @alexeldeib what do you think about the actual changes in the PR if we go by the assumption that its not breaking the API?

alexeldeib:
Oh 😄 Seems reasonable to me. After seeing https://github.com/kubernetes-sigs/controller-runtime/pull/936 I'd like to ponder multicluster failure modes a bit before signing off, but generally 👍

alvaroaleman:
#936 has nothing to do with multicluster. Multicluster works today already, but the startup isn't done properly because:
1. You need to register a second manager on the first manager
2. Once your first manager got its leader lock, it will start the second manager
3. Now the second manager will start the cache for the other cluster

However, the `Controller` gets started in parallel with the seconds manager and doesn't wait for the cache sync so:
* You might see one or two `ErrCacheNotStarted`
* The reconcile duration metric get skewed, because reconciliation is started before the cache is synced

alexeldeib:
I'm aware #936 doesn't have anything to do with multicluster, I meant the potential for degradation/footguns. Like I said the code looks fine to me, I'm just not very familiar with this scenario. I appreciate the explanation. If I understand correctly, this PR basically makes the manager behave like a proper runnable (which is how you'd use it in multicluster)?

alvaroaleman:
So how multicluster with c-r works today is:
1*) You create one manager per cluster and disable metrics and leader election on all except one
2) You add all managers to this one manager, this works because manager itself is a `Runnable`
3) For the watches in a different cluster, you use `source.NewKindWithCache` with the correct cache

From a pure correctness POV of the resulting controller, the only thing wrong is that workers may get started before all caches are synced and/or started. This PR fixes that.

From a more practical/UX POV its super weird to construct multiple managers. I'll be posting a prototype that extracts all cluster-specifics out of the manager and allows using that part without constructing a manager. That change is a lot more involved than this, this here is just a low-hanging fruit I discovered along the way.

\* In theory you don't need multiple managers and you could just create a cache and client per cluster yourself, but the `defaultNewClient` wasn't exported until very recently, so that is why its rather annoying.

alexeldeib:
Cool, that makes sense. Thanks for clarifying. Looking forward to see what you’ve got up your sleeve for multicluster :)

/lgtm

PR #944, opened 2020-05-12
:sparkles: Add webhook total and in-flight metrics

boekkooi-fresh:
Add a gauge for requests currently being handled and a counter for the total amount of requests processed.

I used [prometheus/promhttp](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp?tab=doc) package which provides nice decorators for `Http.Handler` to handle the metrics instead of the custom code.

### DIFF
--- a/pkg/webhook/internal/metrics/metrics.go
+++ b/pkg/webhook/internal/metrics/metrics.go
@@ -30,7 +30,7 @@ var (
 			Name: "controller_runtime_webhook_latency_seconds",
 			Help: "Histogram of the latency of processing admission requests",
 		},
-		[]string{"webhook", "code"},
+		[]string{"webhook"},
 	)
 
 	// RequestTotal is a prometheus metric which is a counter of the total processed admission requests.
--- a/pkg/webhook/internal/metrics/metrics.go
+++ b/pkg/webhook/internal/metrics/metrics.go
@@ -30,11 +30,32 @@ var (
 			Name: "controller_runtime_webhook_latency_seconds",
 			Help: "Histogram of the latency of processing admission requests",
 		},
-		[]string{"webhook"},
+		[]string{"webhook", "code"},
 	)
+
+	// RequestTotal is a prometheus metric which is a counter of the total processed admission requests.
+	RequestTotal = func() *prometheus.CounterVec {
+		return prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "controller_runtime_webhook_requests_total",
+				Help: "Total number of admission requests by HTTP status code.",
+			},
+			[]string{"webhook", "code"},
+		)
+	}()
+
+	// RequestInFlight is a prometheus metric which is a gauge of the in-flight admission requests.
+	RequestInFlight = func() *prometheus.GaugeVec {
+		return prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "controller_runtime_webhook_requests_in_flight",
+				Help: "Current number of admission requests being served.",
+			},
+			[]string{"webhook"},
+		)
+	}()
 )
 
 func init() {
-	metrics.Registry.MustRegister(
-		RequestLatency)
+	metrics.Registry.MustRegister(RequestLatency, RequestTotal, RequestInFlight)
 }
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -130,6 +130,12 @@ func instrumentedHook(path string, hookRaw http.Handler) http.Handler {
 	cnt := metrics.RequestTotal.MustCurryWith(lbl)
 	gge := metrics.RequestInFlight.With(lbl)
 
+	// Initialize the most likely HTTP status codes.
+	lat.WithLabelValues("200")
+	lat.WithLabelValues("500")
+	cnt.WithLabelValues("200")
+	cnt.WithLabelValues("500")
+
 	return promhttp.InstrumentHandlerDuration(
 		lat,
 		promhttp.InstrumentHandlerCounter(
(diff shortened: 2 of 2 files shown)


pkg/webhook/internal/metrics/metrics.go
```diff
@@ -30,11 +30,32 @@ var (
 			Name: "controller_runtime_webhook_latency_seconds",
 			Help: "Histogram of the latency of processing admission requests",
 		},
-		[]string{"webhook"},
+		[]string{"webhook", "code"},
```
boekkooi-fresh, line 33: Hi @directxman12 Maybe you can advise if this is the way to go since I add this because of [pkg/webhook/server.go:131](https://github.com/kubernetes-sigs/controller-runtime/pull/944/files#diff-fbc18bb07cdd05391b7081acc1dfe170L131) but from my point of view this is not optimal since it multiplies the number of metrics by status code.

boekkooi-fresh: Removed this in https://github.com/kubernetes-sigs/controller-runtime/pull/944/commits/374972fdbad05f4b8afa945d13aee1674dfa70c4 because it caused a BC break in one of my dashboards

pkg/webhook/internal/metrics/metrics.go
```diff
@@ -32,9 +32,30 @@ var (
...
 	)
+
+	// RequestTotal is a prometheus metric which is a counter of the total processed admission requests.
+	RequestTotal = func() *prometheus.CounterVec {
+		return prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "controller_runtime_webhook_requests_total",
+				Help: "Total number of admission requests by HTTP status code.",
+			},
+			[]string{"webhook", "code"},
```
alvaroaleman, line 43: I think `path` would be the more sensible name than `webhook`.

Also, please use `const` for the label keys.

pkg/webhook/server.go
```diff
@@ -123,13 +124,23 @@ func (s *Server) Register(path string, hook http.Handler) {
...
-		// TODO(directxman12): add back in metric about total requests broken down by result?
-	})
+	lbl := prometheus.Labels{"webhook": path}
+
+	lat := metrics.RequestLatency.MustCurryWith(lbl)
+	cnt := metrics.RequestTotal.MustCurryWith(lbl)
+	gge := metrics.RequestInFlight.With(lbl)
+
+	// Initialize the most likely HTTP status codes.
+	cnt.WithLabelValues("200")
```
alvaroaleman, line 134: What value does initialization here add?

pkg/webhook/internal/metrics/metrics.go
```diff
@@ -32,9 +32,30 @@ var (
...
+		)
+	}()
+
+	// RequestInFlight is a prometheus metric which is a gauge of the in-flight admission requests.
+	RequestInFlight = func() *prometheus.GaugeVec {
+		return prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "controller_runtime_webhook_requests_in_flight",
```
alvaroaleman, line 51: technically a conversion hook is also a `webhook`, would it maybe make sense to add the `admission` somewhere in the name to avoid name clashes?

pkg/webhook/internal/metrics/metrics.go
boekkooi-fresh: > I think `path` would be the more sensible name than `webhook`.

This maybe a better name but I preferred to sick with the naming convention introduced by [`RequestLatency`](https://github.com/kubernetes-sigs/controller-runtime/blob/1fb86b6/pkg/webhook/internal/metrics/metrics.go#L33).
In my opinion have the labels consistent across metrics and avoiding a BC break was more important.

> 
> Also, please use `const` for the label keys.

Using `const` here could be done but would not fit within the current code based. See for example [pkg/webhook/internal/metrics/metrics.go](https://github.com/kubernetes-sigs/controller-runtime/blob/1fb86b6/pkg/webhook/internal/metrics/metrics.go#L33), [pkg/metrics/client_go_adapter.go](https://github.com/boekkooi-fresh/controller-runtime/blob/754026bd851020d5a94aa420e3a7613dbde76c65/pkg/metrics/client_go_adapter.go#L54), [pkg/metrics/workqueue.go](https://github.com/boekkooi-fresh/controller-runtime/blob/daa6f4d9e7c191732b38f310338bafc8e398ddc5/pkg/metrics/workqueue.go#L45) etc.
From my point of view I prefer to sick with the current code style.
Would you like me to open a separate PR to ensure const's are used for all metrics within `controller-runtime`?

pkg/webhook/internal/metrics/metrics.go
boekkooi-fresh: I'm honestly not sure what the best name here but the `controller_runtime_webhook_` prefixed used I got from [`RequestLatency`](https://github.com/kubernetes-sigs/controller-runtime/blob/1fb86b6/pkg/webhook/internal/metrics/metrics.go#L33) since these metrics apply for the same piece of code.

It maybe better to change these names in a follow up PR to avoid a BC break within this PR?

pkg/webhook/server.go
boekkooi-fresh: It adds the metric to the `/metrics` endpoint with the count set to `0`.
In my experience this allows the user of the metrics just this little more insight about which data they can work with, (used [prometheus/client_golang/prometheus/promhttp/http.go](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/http.go#L240) as my example here)

pkg/webhook/internal/metrics/metrics.go
alvaroaleman: Yeah, makes sense

Issue #945, opened 2020-05-13
Builder's `For`, `Owns` and `Watches` methods overwrite multiple options


jaypipes:
The `Builder` struct's `For`, `Owns` and `Watches` methods all follow the same pattern of having a variable number of modifying `Option` structs as the final parameter to the method. Here is the code for the `For` method:

```go
// For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
// update events by *reconciling the object*.
// This is the equivalent of calling
// Watches(&source.Kind{Type: apiType}, &handler.EnqueueRequestForObject{})
func (blder *Builder) For(object runtime.Object, opts ...ForOption) *Builder {
	input := ForInput{object: object}
	for _, opt := range opts {
		opt.ApplyToFor(&input)
	}

	blder.forInput = input
	return blder
}
```

The problem with the above is if you call `For` with more than one `ForOption`, the builder's `forInput` member ends up containing only the predicates contained in the last `ForOption`. The reason is because the `ForOption.ApplyToFor` *assigns* the `ForOption.predicates` member to the supplied `ForInput.predicates` member (note: `ForOption` is an interface that is implemented by the `Predicates` struct):

```go
// ApplyToFor applies this configuration to the given ForInput options.
func (w Predicates) ApplyToFor(opts *ForInput) {
	opts.predicates = w.predicates
}
```

The solution is to append `w.predicates` to the supplied `opt.predicates` instead of assigning/overwriting the value of `opts.predicates`:

```go
// ApplyToFor applies this configuration to the given ForInput options.
func (w Predicates) ApplyToFor(opts *ForInput) {
	opts.predicates = append(opts.predicates, w.predicates...)
}
```

troy0820:
@vincepri @sbueringer @alvaroaleman Is the intent to only have 1 For option but make it not required when doing this?  If we need to adjust to allow multiple, I can have a PR that appends the values to allow multiple `ForOptions` to be applied to the `forInput`

PR #946, opened 2020-05-13
:bug: ensure multiple Option values append predicates

jaypipes:
The `Builder` struct's `For`, `Owns` and `Watches` methods all follow
the same pattern of having a variable number of modifying `Option`
structs as the final parameter to the method. Here is the code for the
`For` method:

```go
// For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
// update events by *reconciling the object*.
// Watches(&source.Kind{Type: apiType}, &handler.EnqueueRequestForObject{})
func (blder *Builder) For(object runtime.Object, opts ...ForOption) *Builder {
	input := ForInput{object: object}
	for _, opt := range opts {
		opt.ApplyToFor(&input)
	}

	blder.forInput = input
	return blder
}
```

The problem with the above is if you call `For` with more than one
`ForOption`, the builder's `forInput` member ends up containing only the
predicates contained in the last `ForOption`. The reason is because the
`ForOption.ApplyToFor` *assigns* the `ForOption.predicates` member to
the supplied `ForInput.predicates` member (note: `ForOption` is an
interface that is implemented by the `Predicates` struct):

```go
// ApplyToFor applies this configuration to the given ForInput options.
func (w Predicates) ApplyToFor(opts *ForInput) {
	opts.predicates = w.predicates
}
```

The solution is to append `w.predicates` to the supplied
`opt.predicates` instead of assigning/overwriting the value of
`opts.predicates`:

```go
// ApplyToFor applies this configuration to the given ForInput options.
func (w Predicates) ApplyToFor(opts *ForInput) {
	opts.predicates = append(opts.predicates, w.predicates...)
}
```

Fixes Issue #945

### DIFF
--- a/pkg/builder/options.go
+++ b/pkg/builder/options.go
@@ -58,17 +58,17 @@ type Predicates struct {
 
 // ApplyToFor applies this configuration to the given ForInput options.
 func (w Predicates) ApplyToFor(opts *ForInput) {
-	opts.predicates = w.predicates
+	opts.predicates = append(opts.predicates, w.predicates...)
 }
 
 // ApplyToOwns applies this configuration to the given OwnsInput options.
 func (w Predicates) ApplyToOwns(opts *OwnsInput) {
-	opts.predicates = w.predicates
+	opts.predicates = append(opts.predicates, w.predicates...)
 }
 
 // ApplyToWatches applies this configuration to the given WatchesInput options.
 func (w Predicates) ApplyToWatches(opts *WatchesInput) {
-	opts.predicates = w.predicates
+	opts.predicates = append(opts.predicates, w.predicates...)
 }
 
 var _ ForOption = &Predicates{}
--- a/pkg/builder/options_test.go
+++ b/pkg/builder/options_test.go
@@ -1,0 +1,62 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


pkg/builder/options.go
```diff
@@ -58,17 +58,17 @@ type Predicates struct {
 
 // ApplyToFor applies this configuration to the given ForInput options.
 func (w Predicates) ApplyToFor(opts *ForInput) {
-	opts.predicates = w.predicates
+	opts.predicates = append(opts.predicates, w.predicates...)
```
alvaroaleman, line 61: This is an api breach. It is debatable if the previous approach or the current one is "correct" but since we have the previous one already, I think it makes sense to stick with it.

This approach also has the drawback that you can not reset the `predicates` via `opts`.

alvaroaleman: What we probably should do instead is just nilcheck the predicates before setting them

jaypipes: > This is an api breach. It is debatable if the previous approach or the current one is "correct" but since we have the previous one already, I think it makes sense to stick with it.

Well, having a variable number of Option parameters for the `For()` method definitely implies that each of those Option things will be "applied", no? It's a surprising user experience to have all but the last supplied Option do nothing at all.

> This approach also has the drawback that you can not reset the `predicates` via `opts`.

There should probably be a `SetOptions` method then for that?

alvaroaleman: > Well, having a variable number of Option parameters for the For() method definitely implies that each of those Option things will be "applied", no? It's a surprising user experience to have all but the last supplied Option do nothing at all.

Yeah unless you explicitly set it to an empty list. Nilchecking solves the problem you have while still allowing to explicitly reset them by using a non-nil but empty `[]Predicate`

vincepri: If we want to proceed with this change, we can include it in the v0.7.x breaking release

Issue #745, opened 2019-12-30
Support for controllers spanning across clusters


loewenstein:
We have a use case for a controller responsible for a (custom) resource `A` in one cluster that kind of owns another (custom) resource `B` in another cluster. I say "kind of owns" because there are of course no owner references that span clusters, but it can be imitated with labels and a mapping event handler rather easily.

The problem is that there is no easy way to configure the controller to watch for `B`s in that there client. 
* A 'Kind' source would look for `B`s in the wrong cluster. 
* I suppose it could work with a separate manager and controller with no reconciler who's only purpose is to setup a watch and provide an Informer for an Informer source.
Not sure if that would really work though, we would have a controller that prepends to reconcile `B`s and we would need a dummy reconciler. Feels like a hack.
* we could probably use a Channel source, create our own client using client-go and feed events. However, this sounds like a lot of work/custom code, I am not sure how well it would integrate with the rest of controller-runtime

Could we instead have a manager that can have n clients and way that kind watches can specify which client to use?

alvaroaleman:
This is already possible by using multiple managers and injecting the correct cache into `source.Kind`, although this doesn't have tests, hence it's not officially supported, see #650

loewenstein:
I was made aware of #650 as well (after opening this one).

It describes what is probably the best option currently available. However, as managers drive the lifecycle of controllers (and other parts of the controller-runtime), I would hesitate to call it a solution.

Do you think a multi-cluster setup as first-class citizen of the controller-runtime is a viable and feasible option?

akutz:
A project with which I am involved is doing this. I will share the design here once I figure out what I am allowed to share.

alvaroaleman:
xref POC on how this could look like: https://github.com/kubernetes-sigs/controller-runtime/pull/950
/remove-lifecycle stale

adrienjt:
JFYI: I forked controller-runtime back in November 2018 and built this: https://github.com/admiraltyio/multicluster-controller

Merging it back would be a breaking change to controller-runtime's API though.

The folks at Orange use admiraltyio/multicluster-controller for their multi-cluster Cassandra operator: https://github.com/Orange-OpenSource/casskop/blob/master/multi-casskop/Readme.md

[...]

PR #950, opened 2020-05-15
POC: Move cluster-specific code out of the manager

alvaroaleman:
This PR is a POC that aims to improve the multi cluster story with controller-runtime. While multi-cluster controllers are already possible today, they are awkward, because they require to either:
* Construct one manager.Manager per cluster and add all of them except one to the main manager.Manager
* Add a container type that holds client and cache for each additional cluster and basically copies the initialization logic from the manager for those, then make that container type a `Runnable` and add it to the manager

Additionally, startup is incorrect as we don't start secondary caches ahead of time.

So the goal here is basically:
* Don't change anything for the single-cluster controller usecase
* Do not encourage more than one manager for the multi-cluster usecase
* Make sure that startup works correctly and caches get started before anything else

The commits are:
1. Move cluster-specific code from `pkg/manager` to `pkg/clusterconnector` (Better name ideas welcome!)
2. Make sure the manager always starts caches before anything else
3. Add an example multicluster controller

The PR is huge because there was so much code to move. The most important parts to look at are:
* `pkg/manager.Manager` that now embedds the cluster-specifics
* `pkg/clusterconnector.Clusterconnector` which contains the cluster-specifics
* `examples/multicluster/reconciler.Add` to see how such a multi cluster controller looks like

If we can agree on the idea, I will create separate PRs for each of the commits.

Would love to hear opinions on this!

/assign @estroz @DirectXMan12 @vincepri @alexeldeib

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -31,6 +31,15 @@ This example implements a *new* Kubernetes resource, ChaosPod, and creates a cus
     5. Adds ChaosPod webhooks to manager
     6. Starts the manager
 
+### multicluster
+
+This example implements a simplistic controller that watches pods in one cluster (`referencecluster`) and creates
+an identical pod for each pod observed in a different cluster (`mirrorcluster`).
+
+* `main.go`: Initialization code
+* `main_test.go`: Tests that verify the reconciliation
+* `reconciler/reconciler.go`: The actual reconciliation logic
+
 ## Deploying and Running
 
 To install and run the provided examples, see the Kubebuilder [Quick Start](https://book.kubebuilder.io/quick-start.html).
--- a/examples/multicluster/main.go
+++ b/examples/multicluster/main.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


kramerul:
We integrated your POC into our controller and found it quite helpful to have support for additional cluster connections without creating additional managers.

* The required changes were quite small (about 30 lines of code)
* Your POC would be even better if the API would be designed in a way that would reflect the real world. In your example, the controller has connections to 2 Clusters. Therefore, I would expect to have 2 `clusterconnector.ClusterConnector`. But there is one `ctrl.Manager` and one `clusterconnector.ClusterConnector`. Perhaps it would be possible to introduce some kind of a `ctrl.BaseManager` without a connection. This `ctrl.BaseManager` could have functions to open new connections to cluster.

image

mengqiy:
After seeing more and more use cases similar to https://github.com/kubernetes-sigs/controller-runtime/issues/745#issue-543912539, IMO this proposal is useful and promising. I haven't really dived into the implementation details, but I'd like to see this get finalized and merged.

alvaroaleman:
@mengqiy so what do you think is the best way forward for this, should I write a design doc that outlines the changes to make it more digestible?

vincepri:
+1 to have a design doc, it'd also be helpful to visualize and document the structure of the project when we merge this in

mengqiy:
You may want to write a design doc in the [designs](https://github.com/kubernetes-sigs/controller-runtime/tree/master/designs) folder.

@DirectXMan12 @joelanford @estroz  Does anyone have concerns or objection?

estroz:
@mengqiy sgtm, will take time to review the doc once its up. Great stuff so far.

alvaroaleman:
Created a PR for a design doc in https://github.com/kubernetes-sigs/controller-runtime/pull/1075

alvaroaleman:
Closing as the design doc got merged and describes what we want to do here

PR #955, opened 2020-05-20
Backports for v0.5.3

vincepri:


alexeldeib:
Out of scope for this, but seems like we're generating `admissionregistration.k8s.io/v1beta1
` and not `admissionregistration.k8s.io/v1` in master of controller-tools? Might want to update that if I'm reading correctly? Not one of the biggest changes, but nice hygiene

https://github.com/kubernetes-sigs/controller-tools/blob/820a4a27ea84247e0da004cb4033ae9dc3839849/pkg/webhook/parser.go#L29

PR #958, opened 2020-05-20
Backport release-0.6: Fix scheme used by `DeleteAllOf` in the fake client.

Random-Liu:


vincepri:
Thanks for the PR, release-0.6 is the same as master today, we'll probably fast-forward when we'll cut a new v0.6.x release.

/close

PR #962, opened 2020-05-23
:sparkles: Migrate controllerutil AddFinalizer and RemoveFinalizer to controllerutil.Object and deprecate *WithError functions

### LINKED ISSUE #959: Migrate controllerutil AddFinalizer and RemoveFinalizer controllerutil.Object and deprecate *WithError functions
As #898 introduces controllerutil.Object interface it would be nice to migrate controllerutil AddFinalizer and RemoveFinalizer to this interface and remove *WithError functions as they are not needed anymore as checks are performing on the compile time.

dmvolod:
As #898 introduces controllerutil.Object interface this fix migrates controllerutil AddFinalizer and RemoveFinalizer to the controllerutil.Object interface and remove *WithError functions as they are not needed anymore as checks are performing on the compile time.

Fixes #959

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -237,8 +237,8 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
 // MutateFn is a function which mutates the existing object into it's desired state.
 type MutateFn func() error
 
-// AddFinalizer accepts a metav1 object and adds the provided finalizer if not present.
-func AddFinalizer(o metav1.Object, finalizer string) {
+// AddFinalizer accepts an Object and adds the provided finalizer if not present.
+func AddFinalizer(o Object, finalizer string) {
 	f := o.GetFinalizers()
 	for _, e := range f {
 		if e == finalizer {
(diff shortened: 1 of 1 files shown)


dmvolod:
@alvaroaleman , as you suggested, I've opened a separate issue and PR which is migrating to the controllerutil.Object in RemoveFinalizer and AddFinalizer. Also removed *WithError as not needed anymore. Please have a look at this changes. Thank you.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -248,19 +247,8 @@ func AddFinalizer(o metav1.Object, finalizer string) {
 	o.SetFinalizers(append(f, finalizer))
 }
 
-// AddFinalizerWithError tries to convert a runtime object to a metav1 object and add the provided finalizer.
```
alvaroaleman, line 251: hi @dmvolod I think it would be better for now to leave the `WithError` funcs and just add a comment that they are deprecated and shouldn't be used. We can remove them when we also merge other braking changes. AFAIK right now we do not have any breaking changes planned.

/ok-to-test

dmvolod: Thanks for feedback, @alvaroaleman 
I've changed just common functions and marked *WithError as deprecated

PR #967, opened 2020-05-26
:sparkles: Implement graceful shutdown

### LINKED ISSUE #644: Add support for ControllerExpectations
There are certain use cases when controllerexpectations https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L142 makes sense for controller logic (even though I agree they could be misused).

To hook this in properly into controller it would need to be part of the `processNextItem` so it makes sense from my point of view to be integrated into the runtime.

WDYT?

alvaroaleman:
This PR implements graceful shutdown.

It is a rebased version of https://github.com/kubernetes-sigs/controller-runtime/pull/664 plus a set of "Verschlimmbesserungen":

* There was a deadlock in case the cache start failed, because we required the lock to `close(cm.internalStopper)` but the cache kept it while waiting for sync, so if we shutdown before sync completed, this is a deadlock
* Godocs: https://github.com/kubernetes-sigs/controller-runtime/pull/664/files#r355773233
* Make the things we start `Runnables` as well so they are part of the graceful shutdown: https://github.com/kubernetes-sigs/controller-runtime/pull/664/files#r355783939
* Use a defer to do the graceful shutdown to make sure it always happens
* Return both the graceful shutdown and the run error, rather than only logging the former
* Rename `runnableTearDownTimeout` to `gracefulShutdownTimeout`, IMHO the latter makes the idea clearer as its not using terms that are bound to concepts that only exist within c-r
* `exit 1` on leader election loss. This is important, because otherwise controllers might continue to run after the leader lease was given up. Maybe we should only cancel the leader election `context` after graceful shutdown was finished?

Supersedes and thereby closes #664 
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/764

/assign @DirectXMan12 @alexeldeib

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -18,6 +18,7 @@ package manager
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -27,6 +28,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
@@ -44,9 +46,10 @@ import (
 
 const (
 	// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
-	defaultLeaseDuration = 15 * time.Second
-	defaultRenewDeadline = 10 * time.Second
-	defaultRetryPeriod   = 2 * time.Second
+	defaultLeaseDuration          = 15 * time.Second
+	defaultRenewDeadline          = 10 * time.Second
+	defaultRetryPeriod            = 2 * time.Second
+	defaultGracefulShutdownPeriod = 30 * time.Second
 
 	defaultReadinessEndpoint = "/readyz"
 	defaultLivenessEndpoint  = "/healthz"
@@ -118,11 +121,7 @@ type controllerManager struct {
 	started        bool
 	startedLeader  bool
 	healthzStarted bool
-
-	// NB(directxman12): we don't just use an error channel here to avoid the situation where the
-	// error channel is too small and we end up blocking some goroutines waiting to report their errors.
-	// errSignal lets us track when we should stop because an error occurred
-	errSignal *errSignaler
+	errChan        chan error
 
 	// internalStop is the stop channel *actually* used by everything involved
 	// with the manager as a stop channel, so that we can pass a stop channel
@@ -133,6 +132,14 @@ type controllerManager struct {
 	// internalStopper is the write side of the internal stop channel, allowing us to close it.
 	// It and `internalStop` should point to the same channel.
 	internalStopper chan<- struct{}
+
+	// leaderElectionCancel is used to cancel the leader election. It is distinct from internalStopper,
+	// because for safety reasons we need to os.Exit() when we lose the leader election, meaning that
+	// it must be deferred until after gracefulShutdown is done.
+	leaderElectionCancel context.CancelFunc
+
+	// stop procedure engaged. In other words, we should not add anything else to the manager
+	stopProcedureEngaged bool
 
 	// elected is closed when this manager becomes the leader of a group of
 	// managers, either because it won a leader election or because no leader
@@ -161,57 +168,32 @@ type controllerManager struct {
 	// retryPeriod is the duration the LeaderElector clients should wait
 	// between tries of actions.
 	retryPeriod time.Duration
-}
 
-type errSignaler struct {
-	// errSignal indicates that an error occurred, when closed.  It shouldn't
-	// be written to.
-	errSignal chan struct{}
+	// waitForRunnable is holding the number of runnables currently running so that
+	// we can wait for them to exit before quitting the manager
+	waitForRunnable sync.WaitGroup
 
-	// err is the received error
-	err error
+	// gracefulShutdownTimeout is the duration given to runnable to stop
+	// before the manager actually returns on stop.
+	gracefulShutdownTimeout time.Duration
 
-	mu sync.Mutex
-}
+	// onStoppedLeading is callled when the leader election lease is lost.
+	// It can be overridden for tests.
+	onStoppedLeading func()
 
-func (r *errSignaler) SignalError(err error) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	if err == nil {
-		// non-error, ignore
-		log.Error(nil, "SignalError called without an (with a nil) error, which should never happen, ignoring")
-		return
-	}
-
-	if r.err != nil {
-		// we already have an error, don't try again
-		return
-	}
-
-	// save the error and report it
-	r.err = err
-	close(r.errSignal)
-}
-
-func (r *errSignaler) Error() error {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	return r.err
-}
-
-func (r *errSignaler) GotError() chan struct{} {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	return r.errSignal
+	// shutdownCtx is the context that can be used during shutdown. It will be cancelled
+	// after the gracefulShutdownTimeout ended. It must not be accessed before internalStop
+	// is closed because it will be nil.
+	shutdownCtx context.Context
 }
 
 // Add sets dependencies on i, and adds it to the list of Runnables to start.
 func (cm *controllerManager) Add(r Runnable) error {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+	if cm.stopProcedureEngaged {
+		return errors.New("can't accept new runnable as stop procedure is already engaged")
+	}
 
 	// Set dependencies on the object
 	if err := cm.SetFields(r); err != nil {
@@ -231,11 +213,7 @@ func (cm *controllerManager) Add(r Runnable) error {
 
 	if shouldStart {
 		// If already started, start the controller
-		go func() {
-			if err := r.Start(cm.internalStop); err != nil {
-				cm.errSignal.SignalError(err)
-			}
-		}()
+		cm.startRunnable(r)
 	}
 
 	return nil
@@ -293,6 +271,10 @@ func (cm *controllerManager) AddHealthzCheck(name string, check healthz.Checker)
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
 
+	if cm.stopProcedureEngaged {
+		return errors.New("can't accept new healthCheck as stop procedure is already engaged")
+	}
+
 	if cm.healthzStarted {
 		return fmt.Errorf("unable to add new checker because healthz endpoint has already been created")
 	}
@@ -309,6 +291,10 @@ func (cm *controllerManager) AddHealthzCheck(name string, check healthz.Checker)
 func (cm *controllerManager) AddReadyzCheck(name string, check healthz.Checker) error {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+
+	if cm.stopProcedureEngaged {
+		return errors.New("can't accept new ready check as stop procedure is already engaged")
+	}
 
 	if cm.healthzStarted {
 		return fmt.Errorf("unable to add new checker because readyz endpoint has already been created")
@@ -389,17 +375,18 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
 		Handler: mux,
 	}
 	// Run the server
-	go func() {
+	cm.startRunnable(RunnableFunc(func(stop <-chan struct{}) error {
 		log.Info("starting metrics server", "path", defaultMetricsEndpoint)
 		if err := server.Serve(cm.metricsListener); err != nil && err != http.ErrServerClosed {
-			cm.errSignal.SignalError(err)
+			return err
 		}
-	}()
+		return nil
+	}))
 
 	// Shutdown the server when stop is closed
 	<-stop
-	if err := server.Shutdown(context.Background()); err != nil {
-		cm.errSignal.SignalError(err)
+	if err := server.Shutdown(cm.shutdownCtx); err != nil {
+		cm.errChan <- err
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/manager/internal.go
```diff
@@ -420,27 +397,39 @@ func (cm *controllerManager) serveHealthProbes(stop <-chan struct{}) {
...
 }
 
-func (cm *controllerManager) Start(stop <-chan struct{}) error {
-	// join the passed-in stop channel as an upstream feeding into cm.internalStopper
-	defer close(cm.internalStopper)
+func (cm *controllerManager) Start(stop <-chan struct{}) (err error) {
+	// This chan indicates that stop is complete, in other words all runnables have returned or timeout on stop request
+	stopComplete := make(chan struct{})
+	defer close(stopComplete)
+	defer func() {
```
vincepri, line 420: Do we want to add a comment to make sure this function should always be deferred after the stop given the order of executions of defers in Go?

pkg/manager/internal.go
```diff
@@ -389,17 +365,18 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
...
-			cm.errSignal.SignalError(err)
+			return err
 		}
-	}()
+		return nil
+	}))
 
 	// Shutdown the server when stop is closed
 	<-stop
 	if err := server.Shutdown(context.Background()); err != nil {
```
vincepri, line 378: Do we still want to use `context.Background` here or give it a context with a timeout?

pkg/manager/internal.go
```diff
@@ -420,27 +397,39 @@ func (cm *controllerManager) serveHealthProbes(stop <-chan struct{}) {
...
 		}
-	}()
+		return nil
+	}))
 	cm.healthzStarted = true
 	cm.mu.Unlock()
 
 	// Shutdown the server when stop is closed
 	<-stop
 	if err := server.Shutdown(context.Background()); err != nil {
```
vincepri, line 411: Same as above

DirectXMan12:
> "Verschlimmbesserungen"

wooo! compound German words for things are are weird to express in English!

Sorry for the delay in reviewing this, gonna take a peek today/tomorrow

vincepri:
Last changes look good to me

pkg/manager/internal.go
alvaroaleman: So the reason I left this with `context.Background` is that we use channels directly and not a `context.Context` to handle the graceful shutdown.

I guess it would be possible to use a `context.WithTimeout` and propagate it but then we have to select on its `Done()` and afterwards return an error only if its `Error() == context.DeadlineExceeded()`. WDYT?

vincepri: sgtm, probably worth a comment for posterity

pkg/manager/internal.go
```diff
@@ -389,17 +370,18 @@ func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
 		Handler: mux,
 	}
 	// Run the server
-	go func() {
+	cm.startRunnable(RunnableFunc(func(stop <-chan struct{}) error {
```
DirectXMan12, line 373: plumb this as context instead of stop channel?  Or do we want to save that for a followup?

pkg/manager/internal.go
DirectXMan12: :+1: on contexts, but can maybe do that overhaul in a follow-up?  Whatever you feel better about

pkg/manager/internal.go
```diff
@@ -420,27 +402,40 @@ func (cm *controllerManager) serveHealthProbes(stop <-chan struct{}) {
...
-func (cm *controllerManager) Start(stop <-chan struct{}) error {
-	// join the passed-in stop channel as an upstream feeding into cm.internalStopper
-	defer close(cm.internalStopper)
+func (cm *controllerManager) Start(stop <-chan struct{}) (err error) {
+	// This chan indicates that stop is complete, in other words all runnables have returned or timeout on stop request
+	stopComplete := make(chan struct{})
+	defer close(stopComplete)
+	// This must be deferred after closing stopComplete, otherwise we deadlock
+	defer func() {
+		stopErr := cm.engageStopProcedure(stopComplete)
```
DirectXMan12, line 427: very important: link to a picture of picard

pkg/manager/internal.go
```diff
@@ -161,57 +168,27 @@ type controllerManager struct {
...
+	// waitForRunnable is holding the number of runnables currently running so that
+	// we can wait for them to exit before quitting the manager
+	waitForRunnable sync.WaitGroup
 
-func (r *errSignaler) Error() error {
-	r.mu.Lock()
-	defer r.mu.Unlock()
+	// gracefulShutdownTimeout is the duration given to runnable to stop
+	// before the manager actually returns on stop.
+	gracefulShutdownTimeout time.Duration
```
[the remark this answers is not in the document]
DirectXMan12: alternatively, we should have a good way to detect dealine exceeded errors, the same was context does

pkg/manager/internal.go
```diff
@@ -471,9 +466,57 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
...
+	go func() {
+		cm.waitForRunnable.Wait()
+		close(allStopped)
+	}()
+
+	select {
+	case <-allStopped:
+		return nil
+	case <-gracefulShutdownTimer.C:
+		return fmt.Errorf("not all runnables have stopped within the grace period of %s", cm.gracefulShutdownTimeout.String())
```
DirectXMan12, line 519: make this a typed error (see above note about detected deadline-exceeded errors)

DirectXMan12: (also, no need for the call to `.String` if the thing is already stringable)

pkg/manager/internal.go
```diff
@@ -532,19 +559,35 @@ func (cm *controllerManager) waitForCache() {
...
+	cm.leaderElectionCancel = cancel
+	cm.mu.Unlock()
+
+	if cm.onStoppedLeading == nil {
+		cm.onStoppedLeading = func() {
+			// We have to exit here, otherwise the graceful shutdown or anything
+			// else that keeps the binary running might allow controllers that
+			// need leader election to run after we lost the lease.
+			log.Info("leader lease lost, exiting")
+			os.Exit(0)
```
DirectXMan12, line 588: raise a panic in case we've got defers and other stuff folks want to finish?  Nice for cleanup of temporary files and such, esp when testing

pkg/manager/internal.go
```diff
@@ -584,3 +613,13 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 func (cm *controllerManager) Elected() <-chan struct{} {
 	return cm.elected
 }
+
+func (cm *controllerManager) startRunnable(r Runnable) {
+	cm.waitForRunnable.Add(1)
+	go func() {
+		defer cm.waitForRunnable.Done()
+		if err := r.Start(cm.internalStop); err != nil {
+			cm.errChan <- err
```
DirectXMan12, line 622: aren't we gonna have the same issue we had before, where if multiple things error out at the same time, we'll have some block forever since the error channel isn't drained?  That's why we had the weird error-signaler construct before

pkg/manager/internal.go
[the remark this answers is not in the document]
alvaroaleman: > is there a way to say "no timeout"?

I don't think so, except for maybe setting a multi year timeout

> alternatively, we should have a good way to detect dealine exceeded errors, the same was context does

I have changed the shutdown to use a `context` rather than a timer and propagate its error back + extended the tests to verify that we can find it via `errors.Is`

pkg/manager/internal.go
alvaroaleman: That breaks the api so must go into a 0.7.0, whereas this PR does not I believe, hence I would prefer to do it in a different PR

pkg/manager/internal.go
alvaroaleman: hmpf. I usually use references to (non scifi,fantasy) literature characters where possible to falsify the cliche of "all it ppl are somehow in Star Trek, Star Wars or LOTR". But I don't know of a commander off-hand

pkg/manager/internal.go
alvaroaleman: Done

pkg/manager/internal.go
alvaroaleman: I don't think so because we do read from it in `engageStopProcedure` (but only log them rather than returning).

I have added a comment clarifying everything that writes into the channel must not block the `Start` func to make sure we get to `engageStopProcedure` and changed the latter to do the error reading before acquiring the lock to avoid deadlocks.

pkg/manager/internal.go
alvaroaleman: (That approach is also very helpful in quickly identifying which code I wrote vs which code others wrote :upside_down_face: )

pkg/manager/internal.go
alvaroaleman: The problem is that entering this is part of a normal shutdown of a `Manager` that has leader election enabled, so panicing here would mean that all controller-managers that have LE enabled will panic on shutdown. For tests, the whole func can be overriden.

I am not really sure what best to do here. The `os.Exit` is not nice and we also always exit with the same code, because there are different scenarios this has to cover for but which we can not distinguish:
* We lost the leader lock unexpectedly
* We just ended normally without any error
* We had an error during graceful shutdown (we will just swallow that because its returned from the `manager` but by the time that happens we will already have exited the binary here)

alvaroaleman: Maybe push the problem down by documenting "You must exit the binary immediately after the manager ended when leader election is enabled" (and removing the os.Exit())?

pkg/manager/internal.go
DirectXMan12: ah, yeah, missed the "for" in `engageStopProceedure`

pkg/manager/internal.go
DirectXMan12: Alternatively provide a way for external users to configure this, and default it to `os.Exit`?  The immediate `Exit` is just annoying when dealing with stuff like defers.

pkg/manager/internal.go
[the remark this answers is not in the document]
michaelgugino: time.Duration supports negative numbers.  You could detect a negative and use if/else to set the shutdownCtx appropriately.

pkg/manager/internal.go
michaelgugino: +1 plumbing context as a follow-on and keeping API as-is.

pkg/manager/internal.go
alvaroaleman: Okay so after thinking about this some more, I still think what we do today is not correct and unsafe, but its not trivial to improve this so I would like to have a dedicated discussion and PR for that.

I have updated this PR to disable gracefulShutdown when the leader election is lost.

pkg/manager/internal.go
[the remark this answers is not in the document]
alvaroaleman: Great idea, done

alvaroaleman:
@DirectXMan12 @michaelgugino @vincepri updated, PTAL

pkg/manager/internal.go
```diff
@@ -456,27 +464,88 @@ func (cm *controllerManager) Start(stop <-chan struct{}) error {
...
+		// Error starting or running a runnable
+		return err
 	}
 }
 
+// engageStopProcedure signals all runnables to stop, reads potential errors
+// from the errChan and waits for them to end. It must not be called more than once.
+func (cm *controllerManager) engageStopProcedure(stopComplete chan struct{}) error {
+	var cancel context.CancelFunc
+	if cm.gracefulShutdownTimeout > 0 {
```
michaelgugino, line 494: This probably should be >= 0 to be in line with the comments.

alvaroaleman: This doesn't really matter, we return if its zero and then close the context via the `defer cancel()`

alvaroaleman:
> However, if I set graceperiod to short amount (one second or less, I have tested), my tests hang indefinitely.

I tried to reproduce this here: https://github.com/alvaroaleman/machine-api-operator/commit/b8ba348835fc8a6d3b55bc13cb06952f36f79121 However tests pass. Furthermore in the stacktrace I can not find any reference to controller-runtime. Can you provide a repro for this?
In the PR here are multiple tests that use a sub-second `gracefulShutdownPeriod` and they pass and don't hang.

> Also, I'm not sure I want an option that stops waiting (GracePeriod), I want an option that will wait for X seconds, and then force-terminate the child processes. Ignoring them is fruitless for a parent process that does not terminate (such as the unit tests I'm working with). I want to hard-close the webhook processes, I don't care if they finish or not.

Yup I agree with that sentiment, but to my knowledge its not possible to terminate goroutines.

> Anyway, even if we decide to use GracePeriod as we're doing in this patch, we should be more clear about goroutines left running in the background,

How would you make this more clear? Extend the godocs?

>  and I think we should wait indefinitely by default (least surprise, I expect the manager to block until it's actually finished unless I specify otherwise).

I disagree with this, because the result in case of a `Runnable` that doesn't end is that the `manager` will just hang indefinitely without doing anything which is a hard to debug situation. Returning an error stating `It didn't finish` shows where the problem is.

michaelgugino:
@alvaroaleman here's the reproducer: https://github.com/openshift/machine-api-operator/pull/620/files

Make sure you have kubebuilder installed
```
$ ls /usr/local/kubebuilder/bin/
etcd  kube-apiserver  kubebuilder  kubectl
```
and you can run the tests with 
```
go test ./pkg/apis/machine/v1beta1/
```

So, what I think is the essence of the problem here is that the kubeapi server is holding connections open to the webhook goroutine for one reason or another.  There's not really any fundamental differences between my patch and yours, other than I'm running additional webhook handlers.

> Yup I agree with that sentiment, but to my knowledge its not possible to terminate goroutines.

I was thinking, there must be some way to signal the go http.Server to stop immediately via a context or some method, or possibly some configurable on http.Server or one of the related components.

> How would you make this more clear? Extend the godocs?

Yeah, in the documentation, be explicit that the goroutines maybe (will be if we hit the timeout?) left running in the background.  It's kind of not really what I think most would expect by the name of the option.  Maybe something like 'orphanGoRoutinesAfter' or something like that to indicate that we're doing something bad.  Personally, I would prefer to block forever than leak goroutines.  GracePeriod implies that we're going to wait for a connection to finish/cleanup, but if that time expires, we close the connection immediately.

I'll see if I can create another reproducer to try to eliminate my changes specifically.

alvaroaleman:
@michaelgugino those are just bugs in your tests, you are waiting for channels that are never closed when the `manager` errors: https://github.com/alvaroaleman/machine-api-operator/commit/4da76a6db899ac0b9fd503c3e2eca0295b80c6f5

> I was thinking, there must be some way to signal the go http.Server to stop immediately via a context or some method, or possibly some configurable on http.Server or one of the related components.

That might be possible for the http server but is not possible for other `Runnables`.

> Personally, I would prefer to block forever than leak goroutines. GracePeriod implies that we're going to wait for a connection to finish/cleanup, but if that time expires, we close the connection immediately.

Rename it to `WaitPeriod` maybe? As seen in the isssues in your PR, blocking forever just makes debugging harder. 

Generally it is expected that the binary will end if the `manager` ends and doing so is a mandatory requirement when `LeaderElection` is enabled. Hence returning means the binary ends which means all go routines stop.

michaelgugino:
> @michaelgugino those are just bugs in your tests, you are waiting for channels that are never closed when the manager errors: alvaroaleman/machine-api-operator@4da76a6

You're right.  We weren't hitting this before as the manager never produced an error, and we never expected it to, but this new patch set returns an error.

I'm kind of on the fence as to what the right thing is.  On the one hand, the manager never really returned an error after the stop channel was closed, and the manager ended quickly, pretty much instantly.  For webhooks, this is probably okay, as we probably just want to tear it down as quickly as possible, nobody is going to be too upset if a request goes unfulfilled.  On the other hand, the existing behavior is definitely a bug, we should wait for the webhooks to stop.

Let's consider the case where we wait indefinitely by default.  The manager will block forever, and that might be unexpected and hard to diagnose.  This might impact things that are otherwise working today, and a simple rebase onto this patch set is not going to signal any change to the end user as we're not breaking the interface.  However, it would be pretty straight forward to see that the manager is blocking, though it might be hard to determine what it's waiting on.  The flip side to this coin is, if anyone is eagerly restarting a manager without terminating a container (eg, they have the manager in some while loop, and they're catching a signal or something to break that loop), we're going to leak goroutines.  I can't think of a great reason to do this in reality, but some users might be doing that.  IMO, leaking goroutines is definitely worse than blocking forever as it's going to happen silently and eventually your containers probably going to crash.

My preference is to block indefinitely by default, make the timeout opt-in, and specify in some reasonable way that the timeout is going to orphan goroutines, and those goroutines could potentially stick around forever unless the process is stopped.

Personally, since I'm aware of this patch, I can live with either scenario.  I think we should just consider the impact to other users that might not be aware of the change potentially.  I'm not super firm on this position, but I do want to highlight it for consideration.

Otherwise, this patch works well for me.

alvaroaleman:
> I'm kind of on the fence as to what the right thing is

For the record, leaking goroutines is what we already do today and not something this PR introduces. This PR just makes that visible.

In my opinion, if a runnable doesn't stop within 30 seconds its unlikely to stop later on. We can not kill goroutines and just hanging is guaranteed to result in a lot of bug reports and questions, since it looks like a bug in c-r. Returning an error OTOH gives users a clue as to how to debug problems we can not solve for them (get all goroutines to stop) .

PR #968, opened 2020-05-27
🐛 fix(dependency) Use github.com/fsnotify/fsnotify

holyhope:
Use github.com/fsnotify/fsnotify instead of deprecated gopkg.in/fsnotify.v1
See https://github.com/fsnotify/fsnotify/blob/master/go.mod#L1

Signed-off-by: Pierre Péronnet <pierre.peronnet@ovhcloud.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/evanphx/json-patch v4.5.0+incompatible
+	github.com/fsnotify/fsnotify v1.4.7
 	github.com/go-logr/logr v0.1.0
 	github.com/go-logr/zapr v0.1.0
 	github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef // indirect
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@holyhope where do you infer from that the gopkg.in import is deprecated?

alvaroaleman:
Hm, seems the github import is mostly used: https://grep.app/search?q=github.com/fsnotify/fsnotify&filter[path.pattern][0]=go.mod&filter[repo][0]=kubernetes/kubernetes
/lgtm

holyhope:
I had issue importing `controller-runtime` and some other dependencies (I think it was `github.com/uber/jaeger-client-go v2.23.1+incompatible`), and found it was deprecated (see [go.mod](https://github.com/fsnotify/fsnotify/blob/master/go.mod#L1)).
Other reference: https://github.com/fsnotify/fsnotify/issues/328

PR #969, opened 2020-05-27
:sparkles: Add Patch MergeFrom optimistic locking option

vincepri:

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -1210,6 +1210,42 @@ var _ = Describe("Client", func() {
 				close(done)
 			})
 
+			It("should patch an existing object from a go struct, using optimistic locking", func(done Done) {
+				cl, err := client.New(cfg, client.Options{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(cl).NotTo(BeNil())
+
+				By("initially creating a Deployment")
+				dep, err := clientset.AppsV1().Deployments(ns).Create(ctx, dep, metav1.CreateOptions{})
+				Expect(err).NotTo(HaveOccurred())
+
+				By("creating a patch from with optimistic lock")
+				patch := client.MergeFromWithOptions(dep.DeepCopy(), client.MergeFromWithOptimisticLock{})
+
+				By("adding a new annotation")
+				dep.Annotations = map[string]string{
+					"foo": "bar",
+				}
+
+				By("patching the Deployment")
+				err = cl.Patch(context.TODO(), dep, patch)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("validating patched Deployment has new annotation")
+				actual, err := clientset.AppsV1().Deployments(ns).Get(ctx, dep.Name, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(actual).NotTo(BeNil())
+				Expect(actual.Annotations["foo"]).To(Equal("bar"))
+
+				By("validating that a patch should fail with conflict, when it has an outdated resource version")
+				dep.Annotations["should"] = "conflict"
+				err = cl.Patch(context.TODO(), dep, patch)
+				Expect(err).To(HaveOccurred())
+				Expect(apierrors.IsConflict(err)).To(BeTrue())
+
+				close(done)
+			})
+
 			It("should patch and preserve type information", func(done Done) {
 				cl, err := client.New(cfg, client.Options{})
 				Expect(err).NotTo(HaveOccurred())
@@ -2656,8 +2692,9 @@ var _ = Describe("Patch", func() {
 		BeforeEach(func() {
 			cm = &corev1.ConfigMap{
 				ObjectMeta: metav1.ObjectMeta{
-					Namespace: metav1.NamespaceDefault,
-					Name:      "cm",
+					Namespace:       metav1.NamespaceDefault,
+					Name:            "cm",
+					ResourceVersion: "10",
 				},
 			}
 		})
@@ -2685,6 +2722,31 @@ var _ = Describe("Patch", func() {
 
 			By("returning a patch with data only containing the annotation change")
 			Expect(data).To(Equal([]byte(fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`, annotationKey, annotationValue))))
+		})
+
+		It("creates a merge patch with the modifications applied during the mutation, using optimistic locking", func() {
+			const (
+				annotationKey   = "test"
+				annotationValue = "foo"
+			)
+
+			By("creating a merge patch")
+			patch := client.MergeFromWithOptions(cm.DeepCopy(), client.MergeFromWithOptimisticLock{})
+
+			By("returning a patch with type MergePatch")
+			Expect(patch.Type()).To(Equal(types.MergePatchType))
+
+			By("retrieving modifying the config map")
+			metav1.SetMetaDataAnnotation(&cm.ObjectMeta, annotationKey, annotationValue)
+
+			By("computing the patch data")
+			data, err := patch.Data(cm)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning a patch with data containing the annotation change and the resourceVersion change")
+			Expect(data).To(Equal([]byte(fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"},"resourceVersion":"%s"}}`, annotationKey, annotationValue, cm.ResourceVersion))))
 		})
 	})
 })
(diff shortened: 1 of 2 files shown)


pkg/client/client_test.go
```diff
@@ -2685,6 +2722,31 @@ var _ = Describe("Patch", func() {
...
+			By("retrieving modifying the config map")
+			metav1.SetMetaDataAnnotation(&cm.ObjectMeta, annotationKey, annotationValue)
+
+			By("computing the patch data")
+			data, err := patch.Data(cm)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning a patch with data only containing the annotation change")
```
fabriziopandini, line 2747:
```suggestion
			By("returning a patch with data containing the annotation change and the resourceVersion change")
```

pkg/client/client_test.go
```diff
@@ -1209,6 +1209,42 @@ var _ = Describe("Client", func() {
 				close(done)
 			})
 
+			It("should patch an existing object from a go struct, using optimistic locking", func(done Done) {
```
benmoss, line 1212: Maybe out of the scope of this PR but why are all these tests async? I thought this was only useful if you were testing asynchronous code

pkg/client/patch.go
```diff
@@ -59,8 +63,29 @@ func ConstantPatch(patchType types.PatchType, data []byte) Patch {
...
+
+func (m MergeFromWithOptimisticLock) ApplyToMergeFrom(in *MergeFromOptions) {
+	in.OptimisticLock = true
+}
+
+// MergeFromOption is some configuration that modifies options for a merge-from patch data.
+type MergeFromOption interface {
+	// ApplyToMergeFrom applies this configuration to the given patch options.
+	ApplyToMergeFrom(*MergeFromOptions)
+}
```
benmoss, line 76: Could this just be a function type?

vincepri: This is following the same pattern you find in options.go

pkg/client/client_test.go
vincepri: I'm not sure, what do you mean?

benmoss: That they accept the `done Done` parameter, it's an optional feature of Ginkgo `It`: https://onsi.github.io/ginkgo/#asynchronous-tests

pkg/client/patch.go
benmoss: Cool, I didn't see any other Options types here and wasn't familiar with that

pkg/client/client_test.go
vincepri: Not 100% sure, I guess they can all run in parallel

vincepri:
cc @munnerz

alvaroaleman:
@vincepri Can you change the PR to target `master`? We can still cherry-pick it after that but we shouldn't do the initial merge into an older release branch

vincepri:
@alvaroaleman Yeah I can rework the branch, unfortunately when I worked on the patch I was already on release-0.5 and forgot to change the base branch :/

pkg/client/patch.go
```diff
@@ -80,12 +105,47 @@ func (s *mergeFromPatch) Data(obj runtime.Object) ([]byte, error) {
...
+	if s.opts.OptimisticLock {
+		dataMap := map[string]interface{}{}
+		if err := json.Unmarshal(data, &dataMap); err != nil {
+			return nil, err
+		}
+		fromMeta, ok := s.from.(metav1.Object)
+		if !ok {
+			return nil, fmt.Errorf("cannot use OptimisticLock, from object %q is not a valid metav1.Object", s.from)
+		}
+		resourceVersion := fromMeta.GetResourceVersion()
```
alvaroaleman, line 122: How does patch with a RV precondition differ from just doing `Update`?

pkg/client/patch.go
```diff
@@ -59,8 +63,29 @@ func ConstantPatch(patchType types.PatchType, data []byte) Patch {
 	return RawPatch(patchType, data)
 }
 
+type MergeFromWithOptimisticLock struct{}
```
alvaroaleman, line 66: I had to read the code to understand what you meant by `OptimisticLock`. This definitely needs godocs.

I also don't understand how that differs from `Update` except that you put less data on the wire and trade it for a local computation of the patch.

pkg/client/patch.go
vincepri: That's a good question, I think when it comes down to patch is easier to merge only the fields a client has updated. It does keep the scope down to only things that have changes, so it's still possible to retry to patch only things necessary.

This particular use case came up from Cluster API conditions work, where we need to make sure that when patching an object we aren't overwriting other controllers. Adding resource version on patch keeps the scope down to the patch and we can still try to do a best-effort merge.

pkg/client/patch.go
vincepri: +1 I'll add a godoc

pkg/client/patch.go
alvaroaleman: But with this precondition, if anything on the object was changed its RV gets increased and this patch fails. Hence, it has the exact same semantics as `UPDATE` unless I am missing something.

It seems to me a bit that what you actually want is a precondition on the conditions and not the RV?

vincepri: Yeah, it's the same update semantic, although in our case it'll allow us to only focus on a single field and try merging that field without bothering about the rest

alvaroaleman:
Yeah but how is this different from just doing `Update(newObject)`?

I guess I don't properly understand the use case. Do you have an example
somewhere where this does what you want and Update does not?

vincepri:
It's not different in terms of functionality, apart from the need to set the whole spec/status vs only a single field, although we're trying to integrate this logic in https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go#L84 and it'd be a little strange to issue Update within a Patch helper

vincepri:
Do you see this as something that you wouldn't want exposed in controller runtime? I guess we can bring this in Cluster API itself if it wouldn't be useful for others as well. 

Although, given that the functionality of using resource version with Patch is there, and kubectl also allows it, I thought it'd be good to expose it for users that want to go down this path.

alvaroaleman:
> Do you see this as something that you wouldn't want exposed in controller runtime? I guess we can bring this in Cluster API itself if it wouldn't be useful for others as well.

 > Although, given that the functionality of using resource version with Patch is there, and kubectl also allows it, I thought it'd be good to expose it for users that want to go down this path.

Well to my understanding the semantic difference between `Update` and `Patch` is that the latter can be applied even if someone else modified the object. With this precondition, that is not true anymore and the `Patch` operation is just a different way of doing an `Update`.

My objection here is that its not clear to me how this differs from an `Update` operation. So sooner or later someone will ask which of the two to use. And unless we have an answer to that that is better than `Its implemented a bit differently but it behaves identically` I don't think we should add this to our API because it will just result in confusion but not really help much.

ncdc:
There is a difference between update and patch that I believe justifies this PR. Let's say you add a new field to an object without changing the API version. And then you populate that field. If you have a client that is using an older copy of the code that doesn't have this new field, when it sends an update (e.g. maybe you were setting an annotation), it won't have the new field included. Because update is a wholesale replace, the contents of that new field are now lost.

If you use patch, you can target just the changes you want to make, and it won't result in fields that aren't present in the older copy of the go structs being dropped. Adding support for optional optimistic currency control in this patch seems reasonable to me because of this short-coming that update has.

alvaroaleman:
okay, that makes sense. Can we add this information to the godoc ("like update but avoids dropping fields the apiserver has but that aren't present in your go types")?

vincepri:
@alvaroaleman @ncdc added godoc

pkg/client/patch.go
```diff
@@ -59,8 +63,35 @@ func ConstantPatch(patchType types.PatchType, data []byte) Patch {
 	return RawPatch(patchType, data)
 }
 
+// MergeFromWithOptimisticLock can be used if clients want to make sure a patch
+// is being applied to the latest resource version of an object.
+//
+// The behavior is similar to what an Update would do, without the need to send the
+// whole object. Usually this method is useful if you might have multiple clients
+// acting on the same object with different API versions.
```
ncdc, line 71:
```suggestion
// acting on the same object and the same API version, but with different versions of the Go structs. For example,
// an "older" copy of a Widget that has fields A and B, and a "newer" copy with A, B, and C.
// Sending an update using the older struct definition results in C being dropped, whereas using a patch does not.
```

ish 😄

alvaroaleman:
> @alvaroaleman good to merge?

Oh yeah sorry, I am very used to "github appove adds lgtm and approve"

/lgtm
/approve

Issue #965, opened 2020-05-25
default health check not working


lazybetrayer:
if no `AddHealthzCheck` is called `cm.readyzHandler` is nil, the health check endpoint will not be added.

https://github.com/kubernetes-sigs/controller-runtime/blob/a457e2791293540eb7f522b918f9e0892de3bc5f/pkg/manager/internal.go#L415-L417

But according to
https://github.com/kubernetes-sigs/controller-runtime/blob/a457e2791293540eb7f522b918f9e0892de3bc5f/pkg/healthz/healthz.go#L153-L156

there should be a default check.

lazybetrayer:
Actually with the default endpoint name `/healthz`, only the exact match is handled, paths like `/healthz/ping` is never handled.

GrigoriyMikhalkin:
IMO, we need to delete this `if len(h.Checks) == 0 && reqPath[1:] == "ping"` condition.

Regarding default endpoint `/healthz` -- i think it would be ok to add trailing slash and add wrapper for `mux` that would remove trailing slash.

GrigoriyMikhalkin:
There's is default check if you create handler, but don't add any checks to it. But still, you need to create handler. By default controller doesn't have any health endpoints, i.e. no handlers is initialized.

Regarding second comment. Default endpoint i think it would be ok to add trailing slash to default endpoints and add wrapper for mux that would remove trailing slash.

lazybetrayer:
@GrigoriyMikhalkin How to create health check handler without any custom checks?

GrigoriyMikhalkin:
Ah, you can't do that for provided manager implementation. Beckause handler is created when you add check via `Add...Check` methods. But, i guess, that's not that big of a issue, to address it. If you only need `ping` check, you can just add it manually, like `m.Add...Check("ping", healthz.Ping)`.

vincepri:
/milestone Next
/help

This needs some investigation, I'm not too familiar with this part of the codebase.

GrigoriyMikhalkin:
@lazybetrayer Looks like solution to this problem was already merged https://github.com/kubernetes-sigs/controller-runtime/commit/d6fcdd62cf2db5e2b208fb7c32d05cbecfa83999

PR #971, opened 2020-05-27
CR-965: added subpaths for health checks

GrigoriyMikhalkin:
Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/965

I just added paths with trailing slash. Shouldn't break anyone's code -- don't think anyone changed default endpoints names to names with slashes(in that case he would need to make separate call for each check, which is impractical).


GrigoriyMikhalkin:
Identical code was already merged: https://github.com/kubernetes-sigs/controller-runtime/pull/1134.
Closing this PR.

PR #975, opened 2020-05-29
🐛 Always shutdown etcd when stopping envtest

akutz:
This patch fixes an issue that occurs when shutting down an `envtest.Environment`. If the API server fails to shutdown, then the etcd process is orphaned. This can result in dozens of running etcd processes when debugging failed tests.

Please see this [Slack discussion](https://kubernetes.slack.com/archives/CAR30FCJZ/p1590779385006000) for more information.

cc @DirectXMan12 @fstrudel

### DIFF
--- a/pkg/internal/testing/integration/control_plane.go
+++ b/pkg/internal/testing/integration/control_plane.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


pkg/internal/testing/integration/control_plane.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/url"
 
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/errors"
```
alvaroaleman, line 8: mind aliasing the import as `utilerrors`? Thats how its done almost everywhere and the code becomes confusing to read if a given package name refers to different imports in different packages

akutz: Done, thanks!

PR #982, opened 2020-06-05
🐛 Fake client should set resource version on init objects

vincepri:

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -70,8 +70,14 @@ func NewFakeClient(initObjs ...runtime.Object) client.Client {
 func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.Object) client.Client {
 	tracker := testing.NewObjectTracker(clientScheme, scheme.Codecs.UniversalDecoder())
 	for _, obj := range initObjs {
-		err := tracker.Add(obj)
+		accessor, err := meta.Accessor(obj)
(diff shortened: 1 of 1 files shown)


alvaroaleman:
This might be a breaking change for some ppl :grimacing:

vincepri:
Even if it only gets set in case it's empty?

vincepri:
Oh well, it seems it is indeed a breaking change. I can't blame @DirectXMan12 for wanting to deprecate the fake client as-is

DirectXMan12:
hmm... I *think* this should be non-breaking -- the real apiserver sets this, and explicitly guarantees it to not be anything in particular beyond "monotonically increasing on any given object between changes".

@alvaroaleman what was your reasoning behind it being breaking -- does my reasoning above seem sound?

alvaroaleman:
> @alvaroaleman what was your reasoning behind it being breaking -- does my reasoning above seem sound?

Its sound wrt to what the apiserver does, but there are lots of tests out there that create some objects in the fakeclient, then retrieve them and do a `DeepEqual` with a reference. All of them will break after this change :upside_down_face:

vincepri:
I was thinking to make a deep copy and store that one with the resource version, so you could get it after. 

I’ve also noticed that patch doesn’t change resource version if there are changes, but that’s a different problem / PR

vincepri:
Closing this for now, if anyone would like to do this change definitely welcome to

/close

PR #985, opened 2020-06-08
🏃 Update json-patch dependency to e83c0a1c26c8

vincepri:
See https://github.com/kubernetes/kubernetes/pull/91622

> JSON patch implemented by https://github.com/evanphx/json-patch now conforms to RFC6902 replacement semantics and does not accept "replace" patch operations if the specified path does not exist.


alvaroaleman:
/hold
Looks like Prow missed an event

PR #986, opened 2020-06-08
✨ KubeAwareEncoder -> Log kind and apiversion from schema if not set explicitly

nmiculinic:
This PR adds support for deducing GVK for `runtime.Object` from the passed in schema inside the KubeAwareEncoder in the zap helper package. 

It fixes a TODO note which hasn't been implemented and I'd love seeing this functionality.

### DIFF
--- a/pkg/log/zap/kube_helpers.go
+++ b/pkg/log/zap/kube_helpers.go
@@ -23,7 +23,10 @@ import (
 	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // KubeAwareEncoder is a Kubernetes-aware Zap Encoder.
@@ -38,6 +41,9 @@ type KubeAwareEncoder struct {
 	// If false, only name, namespace, api version, and kind are printed.
 	// Otherwise, the full object is logged.
 	Verbose bool
+
+	// Scheme provides the necessary GVK data for arbitrary objects
+	Scheme *runtime.Scheme
 }
 
 // namespacedNameWrapper is a zapcore.ObjectMarshaler for Kubernetes NamespacedName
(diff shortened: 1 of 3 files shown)


pkg/log/zap/zap_test.go
```diff
@@ -285,6 +287,32 @@ var _ = Describe("Zap logger setup", func() {
 			defineTests()
 
 		})
+		Context("with logger created using New with scheme", func() {
```
nmiculinic, line 290: not sure if this is the best testing approach, feedback appreciated.

pkg/log/zap/kube_helpers.go
```diff
@@ -90,6 +102,8 @@ func (w kubeObjectWrapper) MarshalLogObject(enc zapcore.ObjectEncoder) error {
 func (k *KubeAwareEncoder) Clone() zapcore.Encoder {
 	return &KubeAwareEncoder{
 		Encoder: k.Encoder.Clone(),
+		Verbose: k.Verbose,
```
nmiculinic, line 105: not sure why the verbose flag wasn't being cloned.

nmiculinic:
CC @DirectXMan12  since this is implementing TODO assigned to you.

pkg/log/zap/kube_helpers.go
```diff
@@ -57,14 +62,21 @@ func (w namespacedNameWrapper) MarshalLogObject(enc zapcore.ObjectEncoder) error
...
 // MarshalLogObject implements zapcore.ObjectMarshaler
 func (w kubeObjectWrapper) MarshalLogObject(enc zapcore.ObjectEncoder) error {
-	// TODO(directxman12): log kind and apiversion if not set explicitly (common case)
-	// -- needs an a scheme to convert to the GVK.
 	gvk := w.obj.GetObjectKind().GroupVersionKind()
+	if gvk.Version == "" && w.scheme != nil {
+		var err error
+		gvk, err = apiutil.GVKForObject(w.obj, w.scheme)
+		if err != nil {
+			return fmt.Errorf("cannot deduce GVK for object: %v", w.obj)
```
DirectXMan12, line 76: is it more useful to do this, or to just continue on with an error or some sort of sentinel value (e.g. `__unknown` or somesuch)?

I think the latter, but I'm willing to hear counter-arguments.

pkg/log/zap/zap.go
```diff
@@ -145,6 +146,10 @@ type Options struct {
 	// ZapOpts allows passing arbitrary zap.Options to configure on the
 	// underlying Zap logger.
 	ZapOpts []zap.Option
+
+	// Scheme for deducing GVK for arbitrary runtime.Object if not
+	// explicitly specified
+	Scheme *runtime.Scheme
```
alvaroaleman, line 152: It would be nice to default this in `addDefaults`

pkg/log/zap/kube_helpers.go
nmiculinic: Since this is logging, I'd say it's more useful giving any information over crashing with an error. In the common use-case, the schema should have all types registered. 

I'd argue the most useful value would be 

```go
gvk.Kind = fmt.Sprintf("%T", w.obj)
gvk.Version = "__VersionInternal" // IIRC there's some constant for this when I worker with extension API server, though I'm not 100% in which package it is. 
gvk.Group= "__unknown; object not registered in logger schema"
```   

What I'd like presenting to the user is both some useful information (the object's type), and how can he improve the logging information (by adding the type in the schema). 

What do you think?

pkg/log/zap/zap.go
nmiculinic: Something like:
```go
import "	"k8s.io/client-go/kubernetes/scheme"

...
	if o.Scheme == nil {
		o.Scheme = scheme.Scheme
	}
```

Or did you mean something else?

alvaroaleman: Exactly that

pkg/log/zap/kube_helpers.go
DirectXMan12: that seems reasonable, although I'd avoid the spaces in Group and just go with `Kind: fmt.Sprintf("%T", w.obj), gvk.Version: "__unknown", gvk.Group: "__unknown"}`

pkg/log/zap/zap.go
nmiculinic: done

pkg/log/zap/kube_helpers.go
nmiculinic: done

nmiculinic:
@vincepri  sure, when I find time to do it. Since last time it stood there for 13 days with lgtm flag and nothing really happened, until rebase was required.

DirectXMan12:
Looks like it doesn't have an `approved` label yet.  Sorry for the confusion.

DirectXMan12:
/approve

If we get this rebased, we can merge it.

PR #989, opened 2020-06-10
🐛 Avoid marshalling error in debug log line

databus23:
Currently every webhook requests ended up logging this in debug:
```
2020-06-10T16:04:20.883+0200	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/validate-some-resource", "UID": "e5d00608-7f47-4d00-bd91-b1a8955dc5a7", "allowed": true, "result": {}, "resultError": "got runtime.Object without object metadata: &Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:,Message:,Reason:,Details:nil,Code:200,}"}
```

The problem is that a `Status` Object is a runtime.Object with no metadata which is not well received by the `zap.KubeAwareEncoder`.

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -99,6 +99,11 @@ func (wh *Webhook) writeResponse(w io.Writer, response Response) {
 		wh.writeResponse(w, Errored(http.StatusInternalServerError, err))
 	} else {
 		res := responseAdmissionReview.Response
-		wh.log.V(1).Info("wrote response", "UID", res.UID, "allowed", res.Allowed, "result", res.Result)
+		if log := wh.log; log.V(1).Enabled() {
+			if res.Result != nil {
+				log = log.WithValues("code", res.Result.Code, "reason", res.Result.Reason)
+			}
+			log.V(1).Info("wrote response", "UID", res.UID, "allowed", res.Allowed)
+		}
 	}
 }


pkg/webhook/admission/http.go
```diff
@@ -99,6 +99,6 @@ func (wh *Webhook) writeResponse(w io.Writer, response Response) {
 		wh.writeResponse(w, Errored(http.StatusInternalServerError, err))
 	} else {
 		res := responseAdmissionReview.Response
-		wh.log.V(1).Info("wrote response", "UID", res.UID, "allowed", res.Allowed, "result", res.Result)
+		wh.log.V(1).Info("wrote response", "UID", res.UID, "allowed", res.Allowed, "code", res.Result.Code, "reason", res.Result.Reason)
```
alvaroaleman, line 102: `res.Result` might be nil if the request gets approved, if we just unconditionally access it like this we might panic

vincepri: /hold

databus23: Ups. Good catch!

databus23:
@vincepri I addressed the nil pointer access. Not sure about the failing tests now. I think its unrelated timeouts happening somewhere else.

vassilvk:
@databus23, did this fix make it into 0.6.2?
I upgraded from 0.5.0 to 0.6.2, but I'm still getting the `got runtime.Object without object metadata` error.

vassilvk:
@databus23 - actually - never mind :)
I was looking at the wrong binary.
It works as expected with 0.6.2 - thank you!!!

vincepri:
Needs a new release, probably going to do one tomorrow if time permits

Issue #990, opened 2020-06-10
zap loglevels incorrectly returns log level when used from command line.


bharathi-tenneti:
/kind bug

In current implementation, zap log levels are not being set as expected, and are not in line with `logr` levels. 

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/flags.go#L90-L106

PR #991, opened 2020-06-10
 🐛 Fix allowed levels for --zap-log-levels flag to align with logr levels

bharathi-tenneti:
--Current implementation in `flags.go` has a bug, which is leading to incorrect zap log level to be set. 
--This PR also focuses on aligning `--zap-log-level` with logr levels.
Example: `--zap-log-level=2` should map to using increased verbosity `logr.V(2).Info()`.
-- This PR also disables Sampling for debug levels with increased verbosity,  as they will cause index out of bounds errors in sampling code.

Closes: #990

### DIFF
--- a/pkg/log/zap/flags.go
+++ b/pkg/log/zap/flags.go
@@ -29,16 +29,16 @@ import (
 )
 
 var levelStrings = map[string]zapcore.Level{
-	"debug":  zap.DebugLevel,
-	"-1":     zap.DebugLevel,
-	"info":   zap.InfoLevel,
-	"0":      zap.InfoLevel,
-	"error":  zap.ErrorLevel,
-	"2":      zap.ErrorLevel,
-	"dpanic": zap.DPanicLevel,
-	"panic":  zap.PanicLevel,
-	"warn":   zap.WarnLevel,
-	"fatal":  zap.FatalLevel,
+	"debug": zap.DebugLevel,
+	"info":  zap.InfoLevel,
+	"error": zap.ErrorLevel,
+}
+
+// TODO Add level to disable stacktraces.
+// https://github.com/kubernetes-sigs/controller-runtime/issues/1035
+var stackLevelStrings = map[string]zapcore.Level{
+	"info":  zap.InfoLevel,
+	"error": zap.ErrorLevel,
 }
 
 type encoderFlag struct {
@@ -100,8 +100,9 @@ func (ev *levelFlag) Set(flagValue string) error {
 		} else {
 			return fmt.Errorf("invalid log level \"%s\"", flagValue)
 		}
+	} else {
+		ev.setFunc(zap.NewAtomicLevelAt(level))
 	}
-	ev.setFunc(zap.NewAtomicLevelAt(level))
 	ev.value = flagValue
 	return nil
 }
@@ -122,7 +123,7 @@ type stackTraceFlag struct {
 var _ pflag.Value = &stackTraceFlag{}
 
 func (ev *stackTraceFlag) Set(flagValue string) error {
-	level, validLevel := levelStrings[strings.ToLower(flagValue)]
+	level, validLevel := stackLevelStrings[strings.ToLower(flagValue)]
 	if !validLevel {
 		return fmt.Errorf("invalid stacktrace level \"%s\"", flagValue)
 	}
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -181,11 +181,9 @@ func (o *Options) addDefaults() {
 			lvl := zap.NewAtomicLevelAt(zap.ErrorLevel)
 			o.StacktraceLevel = &lvl
 		}
-		// Disable sampling when we are in debug mode. Otherwise, this will
+		// Disable sampling for increased Debug levels. Otherwise, this will
 		// cause index out of bounds errors in the sampling code.
-		if o.Level.Enabled(zapcore.DebugLevel) {
-			o.ZapOpts = append(o.ZapOpts)
-		} else {
+		if !o.Level.Enabled(zapcore.Level(-2)) {
 			o.ZapOpts = append(o.ZapOpts,
 				zap.WrapCore(func(core zapcore.Core) zapcore.Core {
 					return zapcore.NewSampler(core, time.Second, 100, 100)
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -181,12 +181,17 @@ func (o *Options) addDefaults() {
 			lvl := zap.NewAtomicLevelAt(zap.ErrorLevel)
 			o.StacktraceLevel = &lvl
 		}
-		o.ZapOpts = append(o.ZapOpts,
-			zap.WrapCore(func(core zapcore.Core) zapcore.Core {
-				return zapcore.NewSampler(core, time.Second, 100, 100)
-			}))
+		// Disable sampling when we are in debug mode. Otherwise, this will
+		// cause index out of bounds errors in the sampling code.
+		if o.Level.Enabled(zapcore.DebugLevel) {
+			o.ZapOpts = append(o.ZapOpts)
+		} else {
+			o.ZapOpts = append(o.ZapOpts,
+				zap.WrapCore(func(core zapcore.Core) zapcore.Core {
+					return zapcore.NewSampler(core, time.Second, 100, 100)
+				}))
+		}
 	}
-
 	o.ZapOpts = append(o.ZapOpts, zap.AddStacktrace(o.StacktraceLevel))
 }
 
@@ -215,7 +220,7 @@ func NewRaw(opts ...Opts) *zap.Logger {
 //  zap-encoder: Zap log encoding ('json' or 'console')
 //  zap-log-level:  Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error',
 //			       or any integer value > 0 which corresponds to custom debug levels of increasing verbosity")
-//  zap-stacktrace-level: Zap Level at and above which stacktraces are captured (one of 'warn' or 'error')
+//  zap-stacktrace-level: Zap Level at and above which stacktraces are captured (one of 'info' or 'error')
 func (o *Options) BindFlags(fs *flag.FlagSet) {
 
 	// Set Development mode value
@@ -245,7 +250,7 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		o.StacktraceLevel = fromFlag
 	}
 	fs.Var(&stackVal, "zap-stacktrace-level",
-		"Zap Level at and above which stacktraces are captured (one of 'warn' or 'error')")
+		"Zap Level at and above which stacktraces are captured (one of 'info', 'error').")
 }
 
 // UseFlagOptions configures the logger to use the Options set by parsing zap option flags from the CLI.
--- a/pkg/log/zap/zap_test.go
+++ b/pkg/log/zap/zap_test.go
@@ -19,11 +19,14 @@ package zap
 import (
 	"bytes"
 	"encoding/json"
+	"flag"
 	"io/ioutil"
+	"os"
 
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"go.uber.org/zap/zapcore"
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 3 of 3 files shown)


bharathi-tenneti:
This bug may be tested using below snippet of code .
```Go
	opts := zap.Options{}
	opts.BindFlags(flag.CommandLine)
	flag.Parse()

	logger := zap.New(zap.UseFlagOptions(&opts))

	logf.SetLogger(logger)
	logf.Log.V(0).Info("info text")
	logf.Log.V(1).Info("debug 1 text")
	logf.Log.V(2).Info("debug 2 text")
	logf.Log.V(3).Info("debug 3 text")
```

pkg/log/zap/flags.go
```diff
@@ -29,16 +29,15 @@ import (
...
-	"panic":  zap.PanicLevel,
-	"warn":   zap.WarnLevel,
-	"fatal":  zap.FatalLevel,
+	"debug": zap.DebugLevel,
+	"info":  zap.InfoLevel,
+	"error": zap.ErrorLevel,
+}
+
+var stackLevelStrings = map[string]zapcore.Level{
+	"disabled": zap.FatalLevel,
```
hasbro17, line 38: If we're allowing all stacktraces to be completely disabled, we need to mention what `disabled` means in the `--zap-stacktrace-level` flag help text since it's not a zap level.
https://github.com/kubernetes-sigs/controller-runtime/blob/3156051da1ede28679182e1918b226c6f3fa8a74/pkg/log/zap/zap.go#L247-L248

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+		fs = *flag.NewFlagSet(os.Args[0], flag.ExitOnError) //flags are now reset
+	})
+
+	Context("with  zap-log-level options provided", func() {
+		It("Should set info/debug zap-log-level flags.", func() {
+			args := []string{"--zap-log-level=debug"}
+			fromFlags.BindFlags(&fs)
+			if err := fs.Parse(args); err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
```
hasbro17, line 320: Can't this just be an assertion on the error without checking `err!=nil`?
```suggestion
			err := fs.Parse(args)
			Expect(err).ToNot(HaveOccurred())
```

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+			logOut := new(bytes.Buffer)
+
+			logger := New(UseFlagOptions(&fromFlags), WriteTo(logOut))
+			logger.V(0).Info("info text")
+			logger.V(1).Info("debug 1 text")
+
+			outRaw := logOut.Bytes()
+
+			Expect(string(outRaw)).Should(ContainSubstring("info text"))
+			Expect(string(outRaw)).Should(ContainSubstring("debug 1 text"))
```
hasbro17, line 330: Use vars for `info text` and `debug 1 text` to avoid repeating them here and above.

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+
+		})
+
+		It("Should set error zap-log-level flags.", func() {
+			args := []string{"--zap-log-level=error"}
+			fromFlags.BindFlags(&fs)
+			if err := fs.Parse(args); err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
```
hasbro17, line 339:
```suggestion
```suggestion
			err := fs.Parse(args)
			Expect(err).ToNot(HaveOccurred())
```
```

File: pkg/log/zap/zap_test.go [hunk 5]
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+
+			Expect(outRaw).To(BeEmpty())
+		})
+
+		It("Should set level 1 debug zap-log-level flags.", func() {
+			args := []string{"--zap-log-level=1"}
+			fromFlags.BindFlags(&fs)
+			if err := fs.Parse(args); err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
```
hasbro17, line 356:
```suggestion
```suggestion
			err := fs.Parse(args)
			Expect(err).ToNot(HaveOccurred())
```
```

File: pkg/log/zap/zap_test.go [hunk 6]
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+			logger := New(UseFlagOptions(&fromFlags), WriteTo(logOut))
+			logger.V(0).Info("info text")
+			logger.V(1).Info("debug 1 text")
+
+			outRaw := logOut.Bytes()
+
+			Expect(outRaw).To(BeEmpty())
+		})
+
+		It("Should set level 1 debug zap-log-level flags.", func() {
```
hasbro17, line 351:
```suggestion
		It("Should set integer debug level 1 for zap-log-level flag.", func() {
```

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,135 @@ var _ = Describe("Zap logger setup", func() {
...
+
+			outRaw := logOut.Bytes()
+
+			Expect(string(outRaw)).Should(ContainSubstring("info text"))
+			Expect(string(outRaw)).Should(ContainSubstring("debug 1 text"))
+		})
+	})
+
+	Context("with  zap-log-level  with increased verbosity.", func() {
+		It("Should set Level 2 increased verbosity for zap-log-level flags.", func() {
```
hasbro17, line 371: Why is this in a separate Context? This test is the same as `Should set level 1 debug zap-log-level flags` but for level 2.

pkg/log/zap/flags.go
```diff
@@ -29,16 +29,15 @@ import (
...
-	"0":      zap.InfoLevel,
-	"error":  zap.ErrorLevel,
-	"2":      zap.ErrorLevel,
-	"dpanic": zap.DPanicLevel,
-	"panic":  zap.PanicLevel,
-	"warn":   zap.WarnLevel,
-	"fatal":  zap.FatalLevel,
+	"debug": zap.DebugLevel,
+	"info":  zap.InfoLevel,
+	"error": zap.ErrorLevel,
```
camilamacedo86, line 34: Should we not keep the others levels? (`warn`, `panic`, `fatal`)

pkg/log/zap/flags.go
camilamacedo86, line 38: why disabled is `FatalLevel`?

pkg/log/zap/flags.go
joelanford: @camilamacedo86 Those are not valid `logr` levels, so they can't be used when `zapr` is being used to implement the `logr` interface for zap loggers.

pkg/log/zap/flags.go
joelanford: I tend to agree that this could be confusing.

@hasbro17 @bharathi-tenneti and I had an offline conversation where we discussed that since `Fatal` is not a valid logging level with `logr`, we didn't want to include it in the set of log level values handled by `--zap-stacktrace-level`. But there's still a desire to be able to disable stacktraces entirely. The best way to do that would be to NOT add `zap.AddStacktrace()` to the set of `ZapOpts` [here](https://github.com/kubernetes-sigs/controller-runtime/blob/add0b6444f5091f56824b070d28f930090d8deab/pkg/log/zap/zap.go#L190).

But given the existing Options fields and defaults, it isn't immediately clear how to do that.

 See https://github.com/kubernetes-sigs/controller-runtime/pull/991#discussion_r447704765

pkg/log/zap/zap.go
```diff
@@ -181,12 +181,17 @@ func (o *Options) addDefaults() {
...
+		// Disable sampling when we are in debug mode. Otherwise, this will
+		// cause index out of bounds errors in the sampling code.
+		if o.Level.Enabled(zapcore.DebugLevel) {
+			o.ZapOpts = append(o.ZapOpts, zap.Development())
+		} else {
+			o.ZapOpts = append(o.ZapOpts,
+				zap.WrapCore(func(core zapcore.Core) zapcore.Core {
+					return zapcore.NewSampler(core, time.Second, 100, 100)
+				}))
+		}
```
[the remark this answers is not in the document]
bharathi-tenneti: Wouldn't this only cover `Level(-2)` ,We still need to handle `Level(-3)` or below.

pkg/log/zap/zap.go
```diff
@@ -245,7 +250,8 @@ func (o *Options) BindFlags(fs *flag.FlagSet) {
 		o.StacktraceLevel = fromFlag
 	}
 	fs.Var(&stackVal, "zap-stacktrace-level",
-		"Zap Level at and above which stacktraces are captured (one of 'warn' or 'error')")
+		"Zap Level at and above which stacktraces are captured (one of 'info', 'error', or 'disabled')."+
+			"--zap-stacktrace-level=disabled will set level to be 'fatal'.")
```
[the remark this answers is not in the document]
camilamacedo86: wdyt about the warn level? Should we not keep this one?

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,170 @@ var _ = Describe("Zap logger setup", func() {
...
 })
+
+var _ = Describe("Zap log level flag options setup", func() {
+	var (
+		fromFlags Options
+		fs        flag.FlagSet
+		info      = "info text"
+		debug1    = "debug 1 text"
+		debug2    = "debug 2 text"
+		debug3    = "debug 3 text"
```
camilamacedo86, line 310:
```suggestion
		logLevel1<Type>    = "output to test log level 1 (type)"
		logLevel2   = "output to test log level 2 (type)"
		logLevel3   = "output to test log level 3 (type)"
```
WDYT? would not make more clear what it has been testing? 
Could we add the type (info/debug and etc) in each level to make clear?

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,170 @@ var _ = Describe("Zap logger setup", func() {
...
+			logger := New(UseFlagOptions(&fromFlags), WriteTo(logOut))
+			logger.V(0).Info(info)
+			logger.V(1).Info(debug1)
+			logger.V(3).Info(debug3)
+
+			outRaw := logOut.Bytes()
+
+			Expect(string(outRaw)).Should(ContainSubstring(info))
+			Expect(string(outRaw)).Should(ContainSubstring(debug1))
+			Expect(string(outRaw)).Should(ContainSubstring(debug3))
```
camilamacedo86, line 426: Shows that the test is duplicated and this one is testing the previous scenarios as well. in this way, could we not keep this one only and then add tests to ensure that the logs do not appear when the level is not set?

pkg/log/zap/zap.go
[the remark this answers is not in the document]
bharathi-tenneti: Agree , We can remove the developement mode, and it can be as shown below.
```Go
	// Disable sampling when we are in debug mode. Otherwise, this will
		// cause index out of bounds errors in the sampling code.
		if o.Level.Enabled(zapcore.DebugLevel) {
			o.ZapOpts = append(o.ZapOpts)
		} else {
			o.ZapOpts = append(o.ZapOpts,
				zap.WrapCore(func(core zapcore.Core) zapcore.Core {
					return zapcore.NewSampler(core, time.Second, 100, 100)
				}))
		}
```

pkg/log/zap/zap_test.go
bharathi-tenneti: Previous test case checks for `args := []string{"--zap-log-level=2", "--zap-devel=true"}`, where as above test case checks for `args := []string{"--zap-log-level=3", "--zap-devel=false"}`. They are different test cases.

pkg/log/zap/zap_test.go
bharathi-tenneti: I am not sue I understood the concern in this statement. 
"Also, we have in levelStrings just 3 levels mapped, however, here we have info + 3 others levels. is not info a level as well? What does mean info?"

pkg/log/zap/flags.go
bharathi-tenneti: We can add a condition to check for `zapcore.FatalLevel` and apply `zap.AddStackTrace()` accordingly. However, its still a question, that we need to provide some level if not for `fatal` to identify `disabled`.

pkg/log/zap/zap.go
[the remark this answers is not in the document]
hasbro17: @joelanford @bharathi-tenneti Agreed, Let's defer on the changes to `--zap-stacktrace-level=disabled`  and follow up in a separate PR.

It's more important that we fix the `--zap-log-levels` bug without going back and forth on the semantics of how to disable stacktraces for now. We can circle back to this.

@bharathi-tenneti We can just keep the levels `info` and `error` for disabling stacktraces for now.

@camilamacedo86 No point in keeping `warn`. 
With logr, users can log at `info`/`logr.V(0).Info()` or `error`/`logr.Error()`. There's no level for `warn`.

`--zap-stacktrace-level=warn` would capture stacktraces for `warn, error, dpanic, ...` and not for `debug, info`.
And `--zap-stacktrace-level=error` would capture stacktraces for `error, dpanic, ...` and not for `debug, info, warn`.

So `error` should suffice.

pkg/log/zap/flags.go
hasbro17: Responded in https://github.com/kubernetes-sigs/controller-runtime/pull/991#discussion_r451194763
Let's handle this in a follow up PR.

pkg/log/zap/zap_test.go
camilamacedo86: What means debug1 or 2 or 3? What is the level(error/info) related to each one? It is not clear, and because of this, I'd suggest you add the type(error, info) in the variable name in order to be easier its understanding. 

Also, we have only info and the error level why we have info, debug1, debug2, debug3?

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,170 @@ var _ = Describe("Zap logger setup", func() {
...
+		debug3    = "debug 3 text"
+	)
+
+	BeforeEach(func() {
+		fromFlags = Options{}
+		fs = *flag.NewFlagSet(os.Args[0], flag.ExitOnError) //flags are now reset
+	})
+
+	Context("with  zap-log-level options provided", func() {
+		It("Should set info/debug zap-log-level flags.", func() {
```
camilamacedo86, line 319: The tests cases here are not clear. What should occur when `set info/debug zap-log-level flags`?

See how it is in the master: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap_test.go

```
var _ = Describe("Zap logger setup", func() {
	Context("with the default output", func() {
		It("shouldn't fail when setting up production", func() {
....

	Context("when logging kubernetes objects", func() {
		var logOut *bytes.Buffer
		var logger logr.Logger

		defineTests := func() {
			It("should log a standard namespaced Kubernetes object name and namespace", func() {
				pod := &kapi.Pod{}
				pod.Name = "some-pod"

```

We should have similar tests and the descriptions should make easier understand what is testing and what should be the expected result. 

Also, should we not keep the same tests and just adapt them when required and add the new use cases that should be covered instead of remove all?

pkg/log/zap/zap_test.go
camilamacedo86: We should make clear the test cases. E.g `should not output errors logs when the zap level informed in info`, should output debug logs when the zaplevel informed is info and the development mode is enable `. 

Also, what means level 3 if we have just error and info? (0,1)? 
What should be output when the debug mode `--zap-devel=true` is used or not in each case?

pkg/log/zap/zap.go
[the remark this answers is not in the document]
camilamacedo86: Hi @hasbro17, 

It is fine and I agree that it can be a pr to fix and then a pr to add the flag `"--zap-stacktrace-level=disabled`. But then we need to : 
- Not  add "--zap-stacktrace-level=disabled` in this PR
- Add a //todo comment describing what should be done in a follow-up and track an issue for that in this repo.  

c/c @bharathi-tenneti

pkg/log/zap/zap_test.go
```diff
@@ -296,3 +299,170 @@ var _ = Describe("Zap logger setup", func() {
...
+			logger := New(UseFlagOptions(&fromFlags), WriteTo(logOut))
+			logger.V(0).Info(info)
+			logger.V(1).Info(debug1)
+			logger.V(2).Info(debug2)
+
+			outRaw := logOut.Bytes()
+
+			Expect(string(outRaw)).Should(ContainSubstring(info))
+			Expect(string(outRaw)).Should(ContainSubstring(debug1))
+			Expect(string(outRaw)).Should(ContainSubstring(debug2))
```
camilamacedo86, line 407:
```suggestion
			Expect(string(outRaw)).Should(ContainSubstring(infoLevelText))
			Expect(string(outRaw)).Should(ContainSubstring(errorLevelText))
			Expect(string(outRaw)).Should(ContainSubstring(debug2)) // what is debug 2? 
```

pkg/log/zap/zap_test.go
bharathi-tenneti: We do have `debug` for `--zap-log-level` , please see below:
```var levelStrings = map[string]zapcore.Level{
	"debug": zap.DebugLevel,
	"info":  zap.InfoLevel,
	"error": zap.ErrorLevel,
}
```

Issue #680, opened 2019-11-13
Ability to obtain RESTClient for given type


chemikadze:
While trying to use controller-runtime client (outside of controller context), I have faced the need to create watches on objects. Doing that via utilities provided in client-go requires RESTClient to be acquired, which is not exposed by controller-runtime client, and handling creation of RESTClient just for single API call feels a bit redundant. Also, exposing RESTClient would unlock things like #452 in cases, when typed client is available.

DirectXMan12:
https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/apiutil#RESTClientForGVK is what you're looking for.

I'm not super-keen on exposing it via the client interface, especially since not all CR clients are api clients based on RESTClient (some are cache clients or fake clients)

DirectXMan12:
(closing for bookkeeping, but generally willing to hear counter arguments -- not trying to be rude :-) )

chemikadze:
> especially since not all CR clients are api clients based on RESTClient

Oh I see, makes sense.

> counter arguments

Use of RESTClientForGVK somewhat suffers from same issues -- scheme, mapper, codecs should be lined up before it can be used, while controller-runtime already takes care of these things. For example, if client.New would return superset of client.Client and some new RESTClientFactory interface, it would help avoid inconsistencies between manually created and controller-runtime created clients.

DirectXMan12:
Even with `client.New`, I'm not super-keen on tying that to RESTClient directly, since I mostly view RESTClient as an implementation detail there.

That being said, having a helper that pulled that stuff out of the manager might be helpful in making it simpler, since the correct series of calls is

```go
gvk, err := apiutil.GVKForObject(obj, mgr.GetScheme())
if err != nil { /* give yourself RSI from typing this so many times */ }
restClient, err := apiutil.RESTClientForGVK(gvk, mgr.GetConfig(), serializer.NewCodecFactory(mgr.GetScheme()))
```

which is definitely unwieldy.

chemikadze:
Yeah, helper would be good enough if rest clients obtained in this way are more or less same as used by client.

DirectXMan12:
/good-first-issue

a helper for this should be a relatively simple PR.  The code is above.

PR #996, opened 2020-06-15
(:warning:) Add helper to get RESTClient from Manager

varshaprasad96:
This PR adds a helper to Manager, which can be used to create
REST clients to access resources of the same GVK as the
specified runtime object.

This PR is created to address the issue: [#680 ](https://github.com/kubernetes-sigs/controller-runtime/issues/680)

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -28,6 +28,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
@@ -36,6 +37,7 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -355,6 +357,18 @@ func (cm *controllerManager) GetWebhookServer() *webhook.Server {
 		}
 	}
 	return cm.webhookServer
+}
+
+func (cm *controllerManager) GetRESTClientFor(obj runtime.Object) (rest.Interface, error) {
+	gvk, err := apiutil.GVKForObject(obj, cm.GetScheme())
+	if err != nil {
+		return nil, err
+	}
+	restClient, err := apiutil.RESTClientForGVK(gvk, cm.GetConfig(), serializer.NewCodecFactory(cm.GetScheme()))
+	if err != nil {
+		return nil, err
+	}
+	return restClient, nil
 }
 
 func (cm *controllerManager) serveMetrics(stop <-chan struct{}) {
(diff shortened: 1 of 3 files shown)


alvaroaleman:
@varshaprasad96 can you clarify a bit the use-case?

varshaprasad96:
This is to address the issue raised here: https://github.com/kubernetes-sigs/controller-runtime/issues/680, where the user wants a rest client while creating watches on objects through client-go utilities. The wrapper will pull in details from manager, and provide an interface to access the specified resources.

pkg/manager/internal.go
```diff
@@ -368,6 +370,18 @@ func (cm *controllerManager) GetWebhookServer() *webhook.Server {
 	return cm.webhookServer
 }
 
+func (cm *controllerManager) GetRESTClient(obj runtime.Object) (rest.Interface, error) {
```
joelanford, line 373: Nit: Should we suffix the method name with `For` so that it is more clear that this is for a specific object?

```suggestion
func (cm *controllerManager) GetRESTClientFor(obj runtime.Object) (rest.Interface, error) {
```

pkg/manager/manager_test.go
```diff
@@ -896,6 +897,12 @@ var _ = Describe("manger.Manager", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(m.GetAPIReader()).NotTo(BeNil())
 	})
+	It("should provide a client to access runtime object", func() {
```
joelanford, line 900: Nit: just to be consistent with other tests
```suggestion
	It("should provide a function to get a REST client for a runtime object", func() {
```

pkg/manager/internal.go
hasbro17: Not to be pedantic but wouldn't this be `GetRESTClientForGVK()`?
The client is for the GVK, and not just the object.
Where the interface would be `RESTClientForGVK(gvk schema.GroupVersionKind)`. 

Slightly more verbose to actually use though a bit clearer imo:
```Go
gvk, err := apiutil.GVKForObject(obj, mgr.GetScheme())
...
rc, err := mgr.RESTClientForGVK(gvk)
```

joelanford: I don't disagree that it could be confusing, but good go doc could go a long way. There is also some precedent already for using a `runtime.Object` to mean all objects of this type. See [`Builder.For()`](https://github.com/kubernetes-sigs/controller-runtime/blob/53626115bd889e6c9039aaaad9b482a713de5bd5/pkg/builder/controller.go#L72-L84).

varshaprasad96:
@hasbro17 @joelanford Made the suggested changes here. Is this good to go?

pkg/manager/internal.go
```diff
@@ -357,6 +359,18 @@ func (cm *controllerManager) GetWebhookServer() *webhook.Server {
...
+	gvk, err := apiutil.GVKForObject(obj, cm.GetScheme())
+	if err != nil {
+		return nil, err
+	}
+	restClient, err := apiutil.RESTClientForGVK(gvk, cm.GetConfig(), serializer.NewCodecFactory(cm.GetScheme()))
+	if err != nil {
+		return nil, err
+	}
+	return restClient, nil
+}
```
vincepri, line 372: Should this be here, or rather have a function in `controllerutil` where you could pass the Scheme and object in?

vincepri:
@varshaprasad96 @alvaroaleman Should we get this in v0.7.0?

pkg/manager/internal.go
```diff
@@ -357,6 +359,18 @@ func (cm *controllerManager) GetWebhookServer() *webhook.Server {
 	return cm.webhookServer
 }
 
+func (cm *controllerManager) GetRESTClientFor(obj runtime.Object) (rest.Interface, error) {
```
alvaroaleman, line 362: Wouldn't this make more sense of the client? Ultimately, the use case for this is subresources that require a typed client, right? And can we even instantiate a typed client from a `rest.Client`?

PR #999, opened 2020-06-16
[WIP] :sparkles:  conversion webhook support in envtest

### LINKED ISSUE #998: Conversion webhook support in envtest
When we landed webhook support in envtest, we appear to only have added support for mutating & validating webhooks.  We'll need to extend this to conversion webhooks as well.

DirectXMan12:
This adds support for CRD conversion webhooks in envtest.  It functions
similarly to how other webhook support works in envtest, and shares the
underlying CA with any other webhooks that envtest knows about.

I'm not particularly happy with the structure of the "sharing CA between conversion and mutating/defaulting webhooks", but I didn't really want to expose the underlying internal CA machinery.  I think probably we could make do with a few fewer pointers.

This is largely a prototype, and I'd be happy to pass this off to someone else to hack on.

Fixes #998

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+	"fmt"
 
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


hasbro17:
Currently trying to test this out locally. If I can make sense of everything I can try to move this further along.

We'll need a test for conversion webhooks in `pkg/envtest/webhook_test.go` as well I think.

Issue #797, opened 2020-02-13
Reconciler panics should not crash the manager


ekuefler:
Currently, an unhandled panic in a reconciler will not be recovered from, and will likely cause the manager binary to crash. This is a problem, since a panic might be triggered by a single resource in an unexpected state, so that one bad resource could prevent all other resources from being processed. Since Kubernetes is likely to restart the manager pod after a crash, this can also cause the manager to DOS the Kubernetes API server as it continually restarts.

In my project, I wrote this utility function:

```go
func MakeSafe(r reconcile.Reconciler) reconcile.Reconciler {
	return safeReconciler{impl: r}
}

type safeReconciler struct {
	impl reconcile.Reconciler
}

func (r safeReconciler) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
	defer func() {
		if r := recover(); r != nil {
			result = reconcile.Result{}
			err = fmt.Errorf("panic: %v [recovered]\n\n%s", r, debug.Stack())
		}
	}()
	return r.impl.Reconcile(request)
}
```

Every time I pass a reconciler to `Complete`, I wrap it with this. It ensures that any panics raised by the reconciler are converted to normal errors.

alvaroaleman:
Yeah, this would be great to have :)

rajathagasthya:
Something like apimachinery's `HandleCrash` function would also work if we plug it in the right place: https://github.com/kubernetes/apimachinery/blob/3253b0a30d67e7e362b8615e463156bac729c82f/pkg/util/runtime/runtime.go#L45

vincepri:
We can revisit the milestone if the design doesn't have breaking changes. Folks might be relying on panics to detect failures today.

/priority important-soon
/help
/kind design

FillZpp:
@varshaprasad96 Hi, are you still pursuing this? Mind if I take a stab at it :) ?

varshaprasad96:
@FillZpp sure, please feel free to create a PR for this.

PR #1001, opened 2020-06-17
:sparkles: Add helper to handle panics in reconciler

varshaprasad96:
This PR adds a helper function to handle panics in reconciler.
The function converts panics to normal reconciler errors. Custom handlers can also be provided to handle the go routine.

This PR addresses the issue : https://github.com/kubernetes-sigs/controller-runtime/issues/797

### DIFF
--- a/pkg/reconcile/reconcile.go
+++ b/pkg/reconcile/reconcile.go
@@ -17,6 +17,8 @@ limitations under the License.
 package reconcile
 
 import (
+	"fmt"
+	"runtime/debug"
 	"time"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -91,3 +93,31 @@ var _ Reconciler = Func(nil)
 
 // Reconcile implements Reconciler.
 func (r Func) Reconcile(o Request) (Result, error) { return r(o) }
+
+// PanicHandler can be called with defer in order to recover from panic in reconcile logic.
+// Additional handlers can also be passed to manage the go routines. The error passed to the
+// PanicHandler is set to the value of panic if not nil.
+// If any resource triggers a panic, the reconciler will recover and modify the raised panic
+// to the normal reconciler error.
+// Example usage:
+/*
+
+import (
+	...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	)
+
+func (a *ReplicaSetReconciler) Reconcile(req reconcile.Request) (result reconcile.Result, err error) {
+	defer PanicHandler(&err)
+	// reconciler logic here
+
+	...
+*/
+func PanicHandler(err *error, panicHandlers ...func(interface{})) {
+	if r := recover(); r != nil {
+		for _, fn := range panicHandlers {
+			fn(r)
+		}
+		*err = fmt.Errorf("panic: %v [recovered]\n\n%s", r, debug.Stack())
+	}
+}
--- a/pkg/reconcile/reconcile_test.go
+++ b/pkg/reconcile/reconcile_test.go
@@ -66,4 +66,30 @@ var _ = Describe("reconcile", func() {
 			Expect(actualErr).To(Equal(err))
 		})
 	})
+	Describe("Test panic handler", func() {
+		It("should panic and return error", func() {
+			var err error
+			func() {
+				defer reconcile.PanicHandler(&err)
+				panic("Test Panic")
+			}()
+			Expect(err).NotTo(BeNil())
+		})
+		It("should panic and return error as directed by custom handler", func() {
+			var res interface{}
+			var err error
+			handlers := []func(interface{}){
+				func(r interface{}) {
+					res = r
+				},
+			}
+			func() {
+				defer reconcile.PanicHandler(&err, handlers...)
+				panic("Test Panic")
+			}()
+			Expect(err).NotTo(BeNil())
+			Expect(res).To(Equal("Test Panic"))
+		})
+
+	})
 })


pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,26 @@ var _ Reconciler = Func(nil)
...
+// Example usage:
+/*
+func (r *ReconcileMemcached) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
+	defer HandleCrashReconciler(&err)
+	// reconciler logic here
+
+	...
+	return reconcile.Result{}, nil
+*/
+func PanicHandler(err *error, panicHandlers ...func(interface{})) {
```
varshaprasad96, line 111: `PanicHandler` is not tied with the implementation of `Renconciler` and can be an independent helper function, having it here since it was targeted to use with reconciler logic.

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,25 @@ var _ Reconciler = Func(nil)
...
+
+// PanicHandler can be called with defer in order to recover from panic in reconcile logic.
+// Additional handlers can also be passed to manage the go routines. The error passed to the
+// PanicHandler is set to the value of panic if not nil.
+// If any resource triggers a panic, the reconciler will recover and modify the raised panic
+// to the normal reconciler error.
+// Example usage:
+/*
+func (r *ReconcileMemcached) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
+	defer HandleCrashReconciler(&err)
```
hasbro17, line 105: What's `HandleCrashReconciler()` over here?
Is this supposed to be `defer reconcile.PanicHandler(&err)`?

varshaprasad96: Ops, Im sorry, that's a mistake. Yes, that's supposed to be `reconcile.PanicHandler(&err)`. Will change that.

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,25 @@ var _ Reconciler = Func(nil)
...
 func (r Func) Reconcile(o Request) (Result, error) { return r(o) }
+
+// PanicHandler can be called with defer in order to recover from panic in reconcile logic.
+// Additional handlers can also be passed to manage the go routines. The error passed to the
+// PanicHandler is set to the value of panic if not nil.
+// If any resource triggers a panic, the reconciler will recover and modify the raised panic
+// to the normal reconciler error.
+// Example usage:
+/*
+func (r *ReconcileMemcached) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
```
hasbro17, line 104: Very minor nit, but not sure about the name of the reconciler struct here `ReconcileMemcached`.
It might make sense in the context of Kubebuilder/Operator SDK scaffolds but maybe we can use the same example reconciler that's consistent with the existing example in the godocs for this pkg. 

https://github.com/kubernetes-sigs/controller-runtime/blob/c4893e0b992fb1e6e7da29acd692a0cdf9d7aec9/pkg/reconcile/example_test.go#L30-L34

Or I guess we have another one here:

https://github.com/kubernetes-sigs/controller-runtime/blob/c4893e0b992fb1e6e7da29acd692a0cdf9d7aec9/example_test.go#L119

varshaprasad96: Agreed! Modified it to be the `ReplicaSetReconciler` as in `controller-runtime/example_test.go`

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,31 @@ var _ Reconciler = Func(nil)
...
+// Example usage:
+/*
+
+import (
+	...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	)
+
+func (a *ReplicaSetReconciler) Reconcile(req reconcile.Request) (result reconcile.Result, err error) {
+	defer PanicHandler(&err)
```
hasbro17, line 111:
```suggestion
	defer reconcile.PanicHandler(&err)
```

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,31 @@ var _ Reconciler = Func(nil)
...
+	// reconciler logic here
+
+	...
+*/
+func PanicHandler(err *error, panicHandlers ...func(interface{})) {
```
hasbro17, line 116: Nit: The names for the helper `PanicHandler()` and the additional handlers that are passed in `panicHandlers` are a bit similar.

How about making the function name more actionable e.g `HandlePanic()` or `HandleCrash()`, and the extra handlers as `additionalHandlers`.
```suggestion
func HandlePanic(err *error, additionalHandlers ...func(interface{})) {
```

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,31 @@ var _ Reconciler = Func(nil)
 
 // Reconcile implements Reconciler.
 func (r Func) Reconcile(o Request) (Result, error) { return r(o) }
+
+// PanicHandler can be called with defer in order to recover from panic in reconcile logic.
+// Additional handlers can also be passed to manage the go routines. The error passed to the
+// PanicHandler is set to the value of panic if not nil.
+// If any resource triggers a panic, the reconciler will recover and modify the raised panic
+// to the normal reconciler error.
```
hasbro17, line 101: This must be called with `defer`.
Also some suggestions on clarifying how and why the error is set.
```suggestion
// PanicHandler is meant to be called with defer in the Reconcile function in order to recover from a panic. 
// Additional handlers can be passed for more custom logic to handle the panic e.g shutting down go routines. 
// PanicHandler will set the error, err, as the panic message and the stack trace.
// This can be used to set the Reconciler's return error to safely retry when recovering from a panic. 
```

alvaroaleman:
Wouldn't it be better to make this part of the `controller.Controller`? Then ppl do not have to call the `defer`. Also, it would be easier to discover.

pkg/reconcile/reconcile.go
camilamacedo86: My suggestion for it would be : 
```suggestion
func HandlePanicWith(err *error, handleOptions ...func(interface{})) {
```

pkg/reconcile/reconcile.go
```diff
@@ -91,3 +93,31 @@ var _ Reconciler = Func(nil)
...
+	// reconciler logic here
+
+	...
+*/
+func PanicHandler(err *error, panicHandlers ...func(interface{})) {
+	if r := recover(); r != nil {
+		for _, fn := range panicHandlers {
+			fn(r)
+		}
+		*err = fmt.Errorf("panic: %v [recovered]\n\n%s", r, debug.Stack())
```
camilamacedo86, line 121: we are not throwing a panic. So, why the panic in the error msg?  Should it be a fatal error?

PR #1003, opened 2020-06-18
 🏃 Cleanup controller internal logger

vincepri:

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
+	"sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/ratelimiter"
@@ -109,6 +110,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
+		Log:                     log.RuntimeLog.WithName("controller").WithValues("controller", name),
 	}
 
 	return c, nil
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -21,19 +21,17 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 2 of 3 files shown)


pkg/internal/controller/controller.go
```diff
@@ -215,23 +218,26 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 		c.updateMetrics(time.Since(reconcileStartTS))
 	}()
 
-	var req reconcile.Request
-	var ok bool
-	if req, ok = obj.(reconcile.Request); !ok {
+	// Declare a local log using this controller's name.
+	log := log.WithName(c.Name)
```
alvaroaleman, line 222: Why do we have to initialize it twice, can we store it on the controller instead?

vincepri: Yeah that's a good point, changed it up.

pkg/controller/controller.go
```diff
@@ -109,6 +110,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
+		Log:                     log.RuntimeLog.WithName("controller").WithName(name),
```
alvaroaleman, line 113: I think it would be better to keep the fields the same as they are today by doing
```
log.RuntimeLog.WithName("controller").WithValues("controller", name)
```

Changing this might break ppls queries

vincepri: Makes sense, although I've also changed the other log lines to explicitly call `name` and `namespace`; do we consider logs to be an API within this project?

alvaroaleman: > Makes sense, although I've also changed the other log lines to explicitly call name and namespace;

That `name` and `namespace` is from the `reconcile.Request` and not the name of the controller or WDYM?

> do we consider logs to be an API within this project?

Probably not the same as we consider other apis but I think we should try to avoid breaking it unless we have a good reason to.

vincepri: That's fair, I don't have strong objections on the name, I've changed it because it kinda makes sense to have the name as the full path like `controller-runtime/controller/object-reconciler` instead of having it separately.

> That name and namespace is from the reconcile.Request and not the name of the controller or WDYM?
Right, what I meant is that I changed the request from being logged like `"request", req` to a more explicit `req.Name, "namespace", req.Namespace` (line 258)

alvaroaleman: > That's fair, I don't have strong objections on the name, I've changed it because it kinda makes sense to have the name as the full path like controller-runtime/controller/object-reconciler instead of having it separately.

I disagree, filtering logs by controller is a very common usecase and that is easier if there is a dedicated field for that rather than needing to know an arcane prefix.

vincepri: Fixed

PR #1004, opened 2020-06-18
Injectable service object to handle defaulting and validating.

berlin-ab:
Backstory: 

We want to be able to validate that an incoming parameter is valid, and to check if it is valid, we'd like to be able to use the kubernetes client to look up information from the cluster.

Currently the defaulter and validators are methods implemented on the CR runtime object. This limits validations and defaulting to only what metadata a CR has.

This PR suggests a way that we can inject a service object which could contain other dependencies like a client or logger. The service object can then perform Defaulting or Validating.

PR:

I copied the webhook_test.go into webhook_registration_test.go and implemented the feature based on the existing tests.

```
err = WebhookManagedBy(m).
    For(&SomeCustomResource{}).
    WithDefaulter(&SomeCustomResourceDefaulter{}).
    WithValidator(&SomeCustomResourceValidator{}).
    Complete()

type ValidatorService interface {
	ValidateCreate(new runtime.Object) error
	ValidateUpdate(new runtime.Object, old runtime.Object) error
	ValidateDelete(existing runtime.Object) error
}

type DefaulterService interface {
	Default(runtime.Object)
}


```
---

Notes:

- This is similar to the approach for registering a reconciler, and I was initially surprised that the Validator and Defaulter did not work in the same way.
- I couldn't think of better names for the interface. I'm open to ideas.
- The existing registration of validators and defaulters is untouched - although we might consider deprecating them if this feature goes forward.
- There might be need for additional unit tests at lower layers. I drove this from the perspective of the WebhookBuilder.

### DIFF
--- a/pkg/builder/webhook_registration_test.go
+++ b/pkg/builder/webhook_registration_test.go
@@ -1,0 +1,676 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package builder
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var _ = Describe("application", func() {
+	var stop chan struct{}
+
+	BeforeEach(func() {
+		stop = make(chan struct{})
+		newController = controller.New
+	})
+
+	AfterEach(func() {
+		close(stop)
+	})
+
+	Describe("New with registration", func() {
+		It("should scaffold a defaulting webhook if the registered type implements the Defaulter interface", func() {
+			By("creating a controller manager")
+			m, err := manager.New(cfg, manager.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			By("registering the type in the Scheme")
+			builder := scheme.Builder{GroupVersion: testRegisteredDefaulterGVK.GroupVersion()}
+			builder.Register(&TestRegisteredDefaulter{}, &TestRegisteredDefaulterList{})
+			err = builder.AddToScheme(m.GetScheme())
+			Expect(err).NotTo(HaveOccurred())
+
+			err = WebhookManagedBy(m).
+				For(&TestRegisteredDefaulter{}).
+				WithDefaulter(&testDefaulterService{}).
+				Complete()
+
+			Expect(err).NotTo(HaveOccurred())
+			svr := m.GetWebhookServer()
+			Expect(svr).NotTo(BeNil())
+
+			reader := strings.NewReader(`{
+  "kind":"AdmissionReview",
+  "apiVersion":"admission.k8s.io/v1beta1",
+  "request":{
+    "uid":"07e52e8d-4513-11e9-a716-42010a800270",
+    "kind":{
+      "group":"",
+      "version":"v1",
+      "kind":"TestRegisteredDefaulter"
+    },
+    "resource":{
+      "group":"",
+      "version":"v1",
+      "resource":"testdefaulter"
+    },
+    "namespace":"default",
+    "operation":"CREATE",
+    "object":{
+      "replica":1
+    },
+    "oldObject":null
+  }
+}`)
+
+			stopCh := make(chan struct{})
+			close(stopCh)
+			// TODO: we may want to improve it to make it be able to inject dependencies,
+			// but not always try to load certs and return not found error.
+			err = svr.Start(stopCh)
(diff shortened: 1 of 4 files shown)


berlin-ab:
I signed it

berlin-ab:
The "Multi-Namespace Informer Cache" tests fail locally for me. Is this a known problem, or is there some local configuration I need to do?

berlin-ab:
@droot Any chance you'd be able to look at this soon?

berlin-ab:
@gerred @shawn-hurley Are either of you able to take a look at this?

vincepri:
From a quick glance the changes look good, it might be worth adding some documentation in kubebuilder after this merges. 

One thing to point out is that now we have two ways to achieve the same functionality, the PR description says that additional method would allow us to use a client, although I don't see that we're injecting or using a client in the example, is this something each user will have to do separately?

Is the primary use case to use the manager's client? If so, should we look into expanding the current interfaces to accept a client? Using a remote client should also come with a big ⚠️, given that transient errors might incur in data loss or loops if the chain of object isn't well specifid.

I'd like to hear more from others as well @DirectXMan12 @alexeldeib @alvaroaleman

berlin-ab:
> From a quick glance the changes look good, it might be worth adding some documentation in kubebuilder after this merges.

> One thing to point out is that now we have two ways to achieve the same functionality, the PR description says that additional method would allow us to use a client, although I don't see that we're injecting or using a client in the example, is this something each user will have to do separately?

I imagine a future where we deprecate the current solution in place of this solution. I intentionally did not modify the existing solution to allow for them to live side-by-side until the original is deprecated, and the generated code points newcomers to use the builder to inject their validators.

> 
> Is the primary use case to use the manager's client? If so, should we look into expanding the current interfaces to accept a client? Using a remote client should also come with a big ⚠️, given that transient errors might incur in data loss or loops if the chain of object isn't well specifid.

In our usecase, we'd use the manager's client. The same client we hand over to our reconciler. In the tests that I added, I tested that any arbitrary dependency can be used within the defaulters and validators. I didn't think testing specifically for a client was necessary.

berlin-ab:
@DirectXMan12 @alexeldeib @alvaroaleman Do you have time to take a look at this?

@vincepri any comments on the above?

nikhita:
/assign @alvaroaleman 
:wave: can you take a look at this when you get a chance?

pkg/builder/webhook_registration_test.go
```diff
@@ -0,0 +1,676 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
ncdc, line 2: 2020?

alvaroaleman:
Its on my list, sorry for the huge delay, I am a bit swamped.Will get to it till next week.  https://github.com/kubernetes/test-infra should fix the presubmit.

alvaroaleman:
/retest
Supposively the presubmit works now

alvaroaleman:
It just fails for no reason before it actually tests anything?
/retest

alvaroaleman:
> Currently the defaulter and validators are methods implemented on the CR runtime object. This limits validations and defaulting to only what metadata a CR has.

Yeah, this is weird, I am not sure why it exists. What I find a bit weird about the whole `WebhookBuilder` is that it essentially requires to know `generateMutatePath` and `generateValidatePath` by heart, otherwise its not possible to actually configure the webhook. It seems this existed before and was delibaretly removed in c7969dc2f7b3dad6c5115450caeafabf68d8af17 in favor of just not using the `WebhookBuilder` at all, see e.G. `examples/builtins/mutatingwebhook.go`.  In the same commit `pkg/webhook/admission/builder` was completely removed. Maybe whats currently in `pkg/builder` was supposed to be removed as well and was missed back then?

I think I would like to hear @DirectXMan12 voice on what was intended to happen here. Using something along the lines of whats in `examples/builtin/{validating,mutating}webhook.go` should unblock you for now.

/assign @DirectXMan12

james-w:
Hi,

I had a very similar idea to this before I found this PR.

In my case I would like access to the user info in the request, but the current interface
doesn't allow for that.

My implementation approach was very slightly different as well, it is here if you are interested:

https://github.com/james-w/controller-runtime/commit/1d39362a493bea4fa89e18b08a23ccc4484db63d

It does look like https://book.kubebuilder.io/reference/webhook-for-core-types.html might be the better
way to go for more than the basic use case allowed by the `Validator` interface, but then unifying
on something that can evolve as needs become more complex would be good.

alvaroaleman:
@james-w Using `examples/builtin/{validating,mutating}webhook.go` you do get the full `admissionv1beta1.AdmissionRequest` and hence the user info

berlin-ab:
Thanks for the tip on the `examples/builtin` @alvaroaleman. We're coming to controller-runtime via Kubebuilder, so I would still like to move forward with some design that Kubebuilder can take advantage of that abstracts away the request handler (parsing and response logic). We definitely could take advantage of raw controller-runtime for now though.

DirectXMan12:
Whoops, realized I never followed up here after the meeting.  It's not clear to me what you mean by "something that Kubebuilder can take advantage of".  I think the main issue here is that I'm not sure there's enough additional complexity in just switching to the "lower-level" (https://book.kubebuilder.io/reference/webhook-for-core-types.html) webhooks to justify adding this in.  In retrospect, adding in the "high-level" interface may have actually been a mistake; it was supposed to enable semi-idiomatic custom validation & pseudo-defaulting, but since we have better declarative validation support & declarative defaulting now, it kinda just seems like it's adding confusion.

berlin-ab:
I still find value in having a high-level interface for defaulting and validating through code on top of the declarative configuration. I also like that our operator is not concerned with the request/response handling. What we're missing is the ability to configure our validators and defaulters.

vincepri:
> What we're missing is the ability to configure our validators and defaulters.

Is there a concrete example and some use cases? Do your validators or defaulter need external dependencies? From the conversations above, it seems like the current proposed implementation is not something we might want to merge in as-is, because it creates yet-another-way to have _the same functionality_ in a different way. 

Taking a step back, let's come up with clear use cases and examples where the current offering fails to meet the expectations and then arrive to an implementation together.

While this is not going to make it for v0.7.x, we can discuss it and try to come to a design/agreement for a future version.

Thoughts?

phil9909:
> Is there a concrete example and some use cases?

We would be using it to get a client during the validation. Our spec contains a template for another CRD and we would delegate the validation of the template change to the validator of that CRD.

PR #1012, opened 2020-06-21
 🏃 Run tests with race detector enabled

alvaroaleman:
Changes the test script to run with `--race` and fixes the ~three~ five races found (all of which are in test code)

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -21,7 +21,9 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	rt "runtime"
+	"runtime/pprof"
 
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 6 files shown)


pkg/manager/manager_test.go
```diff
@@ -845,11 +850,13 @@ var _ = Describe("manger.Manager", func() {
 		m, err := New(cfg, Options{})
 		Expect(err).NotTo(HaveOccurred())
 		startGoruntime := rt.NumGoroutine()
+		_ = pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
```
alvaroaleman, line 853: The test here is apparently flaky with `-race` set. I am not sure if we can fix this, but having the dump here will at least give some clues.

I am unfortunately not able to repro the issue on my computer, that's why I left this here.

alvaroaleman:
Okay, that failure was presumably cpu starvation:
```
"timeout waiting for process kube-apiserver to start
```

Generally running the tests with `-race` means they need much more cpu so we might have to bump the requests of the job (But I think using the race detector provides a lot of value)
/retest

vincepri:
We could limit the amount of parallel tests with `-p`

alvaroaleman:
> We could limit the amount of parallel tests with -p

Do you know how that interacts with ginkgo?

vincepri:
Ah, forgot about we were using ginkgo — I'll need to look into that, we can set `GINKGO_NODES`, which IIRC defaults to 1

alvaroaleman:
@vincepri wdyt about merging this? If we end up getting flakes we could for example use build flags to run a subset of tests without race detector

PR #1014, opened 2020-06-22
Prepare for v0.5.7

vincepri:
This PR backports some fixes and updates from the current release to release-0.5.


vincepri:
@JoelSpeed This backports the log line fix we discussed last week

detiber:
/lgtm
/hold
hold to allow for @alvaroaleman a chance to take a look

alvaroaleman:
Thanks!
/lgtm
/hold cancel

Sidenote, right now there is a set of features/bugfixes that only exist in a 0.5.X release but not in a 0.6.X release, I think we should do a release of the latter as well so ppl dont lose fixes by upgrading :grimacing:

Issue #1015, opened 2020-06-23
Help with validating webhook example


msiebuhr:
The example at <https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/example_test.go> does not seem to provide an actual, working webhook to go from:

<strike>First, running without an initialized manager (`var mgr ctrl.Manager`) causes the code to segfault on startup. The correct approach from the controller examples seem to be `mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})`. At least it will then start on my machine/microk8s-cluster.</strike> Fixed in #1016 

Second, when calling any of the webhooks (I've mostly tried the validating one), I get the following response (newlines added for readability):

```sh
>kubectl apply -f deployment/invalid-deployment.yaml --dry-run=server
Error from server (InternalError): error when creating "deployment/invalid-deployment.yaml":
    Internal error occurred: failed calling webhook "admission-webhook.example.com":
    expected response.uid="8aa5d1e8-bff4-41d0-aed9-9555a2ba7c3e", got ""
```
From intercepting the response from the validating webhook (wrapper around http.ResponseWriter), I can see it complains about some kind of missing registration:
```json
{
  "response": {
    "uid": "",
    "allowed": false,
    "status": {
      "metadata": {},
      "message": "no kind \"AdmissionReview\" is registered for version \"admission.k8s.io/v1\" in scheme \"sigs.k8s.io/controller-runtime/pkg/webhook/admission/http.go:34\"",
      "code": 400
    }
  }
}
```
The request quite clearly makes it to the server, but (based on liberal use of `log.Printf()`) doesn't make it to any code I control before running into this error.

Some questions:

 1. How do I register versions/schemes?
 2. <strike>There's a lot of `klog.V(x)...`, which looks quite helpful. How do I turn them on?</strike> See comment below.

In the likely case the webhook configuration is important, here it is:

```yaml
apiVersion: admissionregistration.k8s.io/v1 # Most examples in the wild use "v1beta1", but it makes no difference for me
kind: ValidatingWebhookConfiguration
metadata:
  name: validating-webhook-example-cfg
webhooks:
  - name: admission-webhook.example.com
    admissionReviewVersions: ["v1"]
    sideEffects: None
    clientConfig:
      service:
        name: webhook
        namespace: admission-webhook
        path: "/validating"
      caBundle: |
         ....
    rules:
      - operations: [ "CREATE" ]
[...]

PR #1016, opened 2020-06-23
:bug: Fix webhook example

msiebuhr:
Trying to use a nil/zero'ed Manager results the example code to immediate panic on startup, so create a proper Manger instead.

See #1015


msiebuhr:
I signed it

PR #1031, opened 2020-07-06
add a known limitations section to fake Client

jpatel-pivotal:
This PR adds more information to let users know of the currently known limitations as they exist in `fake.Client`.

We need this to:

* share this list with potential users of the fake client to help with expectations around TDD
* also will help with discussion around which item on the list may be fixed via its own issue. 

This PR also removes the deprecation warning as per issue #768 and is the first step to see if we can improve TDD experience for unit testing by using the fake Client in various testing use cases and scenarios.

### DIFF
--- a/pkg/client/fake/doc.go
+++ b/pkg/client/fake/doc.go
@@ -27,7 +27,7 @@ You can invoke the methods defined in the Client interface.
 When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
 
-Current Limitations / Known Issues with the fake Client:
+WARNING: ⚠️ Current Limitations / Known Issues with the fake Client ⚠️
 - This client does not have a way to inject specific errors to test handled vs. unhandled errors.
 - There is some support for sub resources which can cause issues with tests if you're trying to update
   e.g. metadata and status in the same reconcile.
--- a/pkg/client/fake/doc.go
+++ b/pkg/client/fake/doc.go
@@ -28,8 +28,7 @@ When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
 
 Current Limitations / Known Issues with the fake Client:
-- this client does not use reactors so it can not be setup with detailed responses for testing
-- possible locking issues when using fake client and not the actual client
+- this client does not have a way to inject specific errors to test handled vs. unhandled errors
 - there is some support for sub resources which can cause issues with tests if you're trying to update
   e.g. metadata and status in the same reconcile
 - No OpeanAPI validation is performed when creating or updating objects.
--- a/pkg/client/fake/doc.go
+++ b/pkg/client/fake/doc.go
@@ -28,9 +28,9 @@ When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
 
 Current Limitations / Known Issues with the fake Client:
-- this client does not have a way to inject specific errors to test handled vs. unhandled errors
-- there is some support for sub resources which can cause issues with tests if you're trying to update
-  e.g. metadata and status in the same reconcile
+- This client does not have a way to inject specific errors to test handled vs. unhandled errors.
+- There is some support for sub resources which can cause issues with tests if you're trying to update
+  e.g. metadata and status in the same reconcile.
 - No OpeanAPI validation is performed when creating or updating objects.
 - ObjectMeta's `Generation` and `ResourceVersion` don't behave properly, Patch or Update
 operations that rely on these fields will fail, or give false positives.
--- a/pkg/client/fake/doc.go
+++ b/pkg/client/fake/doc.go
@@ -17,10 +17,10 @@ limitations under the License.
 /*
 Package fake provides a fake client for testing.
 
-fake client is backed by its simple object store indexed by GroupVersionResource.
+A fake client is backed by its simple object store indexed by GroupVersionResource.
 You can create a fake client with optional objects.
 
-	client := NewFakeClient(initObjs...) // initObjs is a slice of runtime.Object
+	client := NewFakeClientWithScheme(scheme, initObjs...) // initObjs is a slice of runtime.Object
 
 You can invoke the methods defined in the Client interface.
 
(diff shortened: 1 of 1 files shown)


jpatel-pivotal:
I have signed the cla /check-cla

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
 	client := NewFakeClient(initObjs...) // initObjs is a slice of runtime.Object
 
 You can invoke the methods defined in the Client interface.
 
-When it doubt, it's almost always better not to use this package and instead use
+When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
+
+Current Limitations / Known Issues with the fake Client:
+- this client does not use reactors so it can not be setup with detailed responses for testing
```
vincepri, line 31:
```suggestion
```

We can probably remove this line, given that it's clear it's a different implementation.

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
 /*
 Package fake provides a fake client for testing.
 
-Deprecated: please use pkg/envtest for testing. This package will be dropped
-before the v1.0.0 release.
-
-An fake client is backed by its simple object store indexed by GroupVersionResource.
+fake client is backed by its simple object store indexed by GroupVersionResource.
```
vincepri, line 20:
```suggestion
A fake client is backed by its simple object store indexed by GroupVersionResource.
```

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
 /*
 Package fake provides a fake client for testing.
 
-Deprecated: please use pkg/envtest for testing. This package will be dropped
-before the v1.0.0 release.
```
vincepri, line 21: cc @DirectXMan12

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
 
 You can invoke the methods defined in the Client interface.
 
-When it doubt, it's almost always better not to use this package and instead use
+When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
+
+Current Limitations / Known Issues with the fake Client:
+- this client does not use reactors so it can not be setup with detailed responses for testing
+- possible locking issues when using fake client and not the actual client
```
vincepri, line 32: Which locking issues are we referring to? It'd be great to have more details

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
+- by default you are using the wrong scheme
```
vincepri, line 33: What do you mean by this? The function that creates the fake client without a scheme has been deprecated, and will be removed in the next version.

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
+- there is no support for sub resources which can cause issues with tests if you're trying to update
+  e.g. metadata and status in the same reconcile
```
vincepri, line 35: There is some support for the `status` subresource

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
+- There is also no OpenAPI validation
```
vincepri, line 36:
```suggestion
- No OpeanAPI validation is performed when creating or updating objects.
```

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,24 @@ limitations under the License.
...
+- It does not bump `Generation` or `ResourceVersion` so Patch/Update will not behave properly
```
vincepri, line 37:
```suggestion
- ObjectMeta's `Generation` and `ResourceVersion` don't behave properly, Patch or Update operations that rely on these fields will fail, or give false positives.
```

pkg/client/fake/doc.go
jpatel-pivotal: I noted this down as we were discussing in the sig meeting so you are right we need more detail on this item. Are there others that can provide more detail here?

pkg/client/fake/doc.go
jpatel-pivotal: The use case here is if you would like to test error cases in your controller's reconcile logic, we need a way to inject a specific error. For e.g. to test the psuedo code below:

pkg/client/fake/doc.go
jpatel-pivotal: Fixed

pkg/client/fake/doc.go
jpatel-pivotal: Changed the example in the doc on how to create a new fake Client with a scheme

pkg/client/fake/doc.go
jpatel-pivotal: fixed wording

jpatel-pivotal:
I signed it

pkg/client/fake/doc.go
vincepri: Sounds good to me

pkg/client/fake/doc.go
vincepri: I'd double check with the recording, but IIRC this is related to locking with update/patch when using resourceVersion, which is covered below

jpatel-pivotal: You are right. Changed it and resolving this.

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,23 @@ limitations under the License.
...
 You can invoke the methods defined in the Client interface.
 
-When it doubt, it's almost always better not to use this package and instead use
+When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
+
+Current Limitations / Known Issues with the fake Client:
+- this client does not have a way to inject specific errors to test handled vs. unhandled errors
+- there is some support for sub resources which can cause issues with tests if you're trying to update
+  e.g. metadata and status in the same reconcile
```
vincepri, line 33:
```suggestion
- This client does not have a way to inject specific errors to test handled vs. unhandled errors.
- There is some support for sub resources which can cause issues with tests if you're trying to update
  e.g. metadata and status in the same reconcile.
```

pkg/client/fake/doc.go
```diff
@@ -17,17 +17,23 @@ limitations under the License.
...
-	client := NewFakeClient(initObjs...) // initObjs is a slice of runtime.Object
+	client := NewFakeClientWithScheme(scheme, initObjs...) // initObjs is a slice of runtime.Object
 
 You can invoke the methods defined in the Client interface.
 
-When it doubt, it's almost always better not to use this package and instead use
+When in doubt, it's almost always better not to use this package and instead use
 envtest.Environment with a real client and API server.
+
+Current Limitations / Known Issues with the fake Client:
```
vincepri, line 30:
```suggestion
WARNING: ⚠️ Current Limitations / Known Issues with the fake Client ⚠️
```

vincepri:
@jpatel-pivotal A few other suggestions, otherwise the change should be good to go, @alvaroaleman wdyt

Issue #1028, opened 2020-07-02
Remove "master" word from our codebase


vincepri:
- https://github.com/kubernetes-sigs/controller-runtime/blob/add0b6444f5091f56824b070d28f930090d8deab/pkg/client/config/config.go#L43
- https://github.com/kubernetes-sigs/controller-runtime/blob/add0b6444f5091f56824b070d28f930090d8deab/pkg/manager/manager.go#L147
- https://github.com/kubernetes-sigs/controller-runtime/blob/add0b6444f5091f56824b070d28f930090d8deab/pkg/manager/internal.go#L158

/milestone Next
/wg naming

fabriziopandini:
I checked the codebase and it seems everything in this issue is already addressed. Is there something else I should take care of?

vincepri:
/close

This can be closed, the only thing left to do is #1026

PR #1038, opened 2020-07-09
feat:Remove master word from our codebase

tanjunchen:
Fixes:https://github.com/kubernetes-sigs/controller-runtime/issues/1028
/cc @vincepri

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -155,7 +155,7 @@ type controllerManager struct {
 	// leaseDuration is the duration that non-leader candidates will
 	// wait to force acquire leadership.
 	leaseDuration time.Duration
-	// renewDeadline is the duration that the acting master will retry
+	// renewDeadline is the duration that the acting controlplane will retry
 	// refreshing leadership before giving up.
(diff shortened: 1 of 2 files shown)


pkg/client/config/config.go
```diff
@@ -40,7 +40,7 @@ func init() {
 		"Paths to a kubeconfig. Only required if out-of-cluster.")
 
 	// This flag is deprecated, it'll be removed in a future iteration, please switch to --kubeconfig.
-	flag.StringVar(&apiServerURL, "master", "",
+	flag.StringVar(&apiServerURL, "apiServerURL", "",
```
vincepri, line 43: This requires a ⚠️ and it's breaking-release. We should just removed the flag

vincepri: Let's do separate these in two PRs, one for this change that removes the flag entirely, and one for the comments

tanjunchen: ok thanks

tanjunchen: done

PR #1043, opened 2020-07-10
✨ pkg/predicate.go: add And and Or predicates

joelanford:
This PR adds two new boolean predicate functions (`predicate.And()` and `predicate.Or()`) that take a variadic predicates argument.

When creating a watch, multiple predicates can be used, but they are always logically ANDed together. If these boolean predicates existed, it would be easier for users to reuse simple predicates to build complex predicate filters for their watches without having to resort to writing their own monolithic predicate from scratch.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -41,6 +41,8 @@ type Predicate interface {
 var _ Predicate = Funcs{}
 var _ Predicate = ResourceVersionChangedPredicate{}
 var _ Predicate = GenerationChangedPredicate{}
+var _ Predicate = or{}
+var _ Predicate = and{}
 
 // Funcs is a function that implements Predicate.
 type Funcs struct {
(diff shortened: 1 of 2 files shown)


joelanford:
/hold
to get an approval other than mine.

alvaroaleman:
> When creating a watch, multiple predicates can be used, but they are always logically ANDed together. If these boolean predicates existed, it would be easier for users to reuse simple predicates to build complex predicate filters for their watches without having to resort to writing their own monolithic predicate from scratch.

If and is already the default, why add an explicit wrapper to replicate that?

joelanford:
> If and is already the default, why add an explicit wrapper to replicate that?

It would still be needed for arbitrary boolean syntax: e.g. `(A || (B && C))`

joelanford:
/hold cancel

This was discussed at the biweekly developer meeting and there was consensus to merge it.

joelanford:
^^^ @vincepri can we get this into the next v0.6.x release?

vincepri:
/milestone v0.6.x

@joelanford Yes, it'll definitely be 👍

PR #1053, opened 2020-07-17
 🏃 Generate junit in CI

alvaroaleman:

### DIFF
--- a/hack/test-all.sh
+++ b/hack/test-all.sh
@@ -23,3 +23,7 @@ setup_envs
 header_text "running go test"
 
 go test -race ${MOD_OPT} ./...
+
+if [[ -n ${ARTIFACTS:-} ]]; then
+  if grep -Rin '<failure type="Failure">' ${ARTIFACTS}/*; then exit 1; fi
+fi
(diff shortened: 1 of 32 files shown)


alvaroaleman:
/hold
This makes a test run with failing tests pass https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1053/pull-controller-runtime-test-master/1284892162755923969 ?

alvaroaleman:
@vincepri I've updated the test script to check in CI if the junit xmls have failures as a workaround to the ginkgo not properly failing on failures ¯\_(ツ)_/¯

vincepri:
Any reason we're skipping 30 tests?

vincepri:
Looking at the code, I see the warning on the reporter, I'm hoping that folks don't use this blindly :/

alvaroaleman:
> Any reason we're skipping 30 tests?

No idea but not because of this PR, we just see that now which I guess is a good thing.

> Looking at the code, I see the warning on the reporter, I'm hoping that folks don't use this blindly :/

Yeah, me too :grimacing:

alvaroaleman:
@vincepri are you comfortable with how the change looks like currently?

My subjective impression is that our tests are still flaking quite a bit and it would be great if we could use testgrid to see which tests exactly are flaking how much rather than just seeing "some runs pass, some fail"

vincepri:
Yeah that makes sense, let's get this merged!

/lgtm

PR #1054, opened 2020-07-18
⚠ Add a context w/ logger to Reconciler interface

vincepri:

### DIFF
--- a/examples/builtins/controller.go
+++ b/examples/builtins/controller.go
@@ -20,11 +20,10 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/go-logr/logr"
-
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -32,15 +31,14 @@ import (
 type reconcileReplicaSet struct {
 	// client can be used to retrieve objects from the APIServer.
 	client client.Client
-	log    logr.Logger
 }
 
 // Implement reconcile.Reconciler so the controller can reconcile objects
 var _ reconcile.Reconciler = &reconcileReplicaSet{}
 
-func (r *reconcileReplicaSet) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *reconcileReplicaSet) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	// set up a convenient log object so we don't have to type request over and over again
-	log := r.log.WithValues("request", request)
+	log := log.FromContext(ctx)
 
 	// Fetch the ReplicaSet from the cache
 	rs := &appsv1.ReplicaSet{}
--- a/examples/builtins/main.go
+++ b/examples/builtins/main.go
@@ -25,7 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
@@ -33,11 +33,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
-var log = logf.Log.WithName("example-controller")
+func init() {
+	log.SetLogger(zap.New())
+}
 
 func main() {
-	logf.SetLogger(zap.New())
-	entryLog := log.WithName("entrypoint")
+	entryLog := log.Log.WithName("entrypoint")
 
 	// Setup a Manager
 	entryLog.Info("setting up manager")
@@ -50,7 +51,7 @@ func main() {
 	// Setup a new controller to reconcile ReplicaSets
 	entryLog.Info("Setting up controller")
 	c, err := controller.New("foo-controller", mgr, controller.Options{
-		Reconciler: &reconcileReplicaSet{client: mgr.GetClient(), log: log.WithName("reconciler")},
+		Reconciler: &reconcileReplicaSet{client: mgr.GetClient()},
 	})
 	if err != nil {
 		entryLog.Error(err, "unable to set up individual controller")
--- a/examples/crd/main.go
+++ b/examples/crd/main.go
@@ -29,12 +29,12 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	api "sigs.k8s.io/controller-runtime/examples/crd/pkg"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 var (
 	setupLog = ctrl.Log.WithName("setup")
-	recLog   = ctrl.Log.WithName("reconciler")
 )
 
 type reconciler struct {
@@ -42,10 +42,9 @@ type reconciler struct {
 	scheme *runtime.Scheme
 }
 
-func (r *reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	log := recLog.WithValues("chaospod", req.NamespacedName)
+func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx).WithValues("chaospod", req.NamespacedName)
 	log.V(1).Info("reconciling chaos pod")
-	ctx := context.Background()
 
 	var chaospod api.ChaosPod
 	if err := r.Get(ctx, req.NamespacedName, &chaospod); err != nil {
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -130,6 +130,12 @@ func (blder *Builder) WithOptions(options controller.Options) *Builder {
 	return blder
 }
 
+// WithLogger overrides the controller options's logger used.
+func (blder *Builder) WithLogger(log logr.Logger) *Builder {
+	blder.log = log
+	return blder
+}
+
 // Named sets the name of the controller to the given name.  The name shows up
 // in metrics, among other things, and thus should be a prometheus compatible name
 // (underscores and alphanumeric characters only).
@@ -137,12 +143,6 @@ func (blder *Builder) WithOptions(options controller.Options) *Builder {
 // By default, controllers are named using the lowercase version of their kind.
 func (blder *Builder) Named(name string) *Builder {
 	blder.name = name
-	return blder
-}
-
-// WithLogger overrides the controller options's logger used.
-func (blder *Builder) WithLogger(log logr.Logger) *Builder {
-	blder.log = log
 	return blder
 }
 
--- a/pkg/builder/controller_test.go
+++ b/pkg/builder/controller_test.go
@@ -44,7 +44,7 @@ import (
 
 type typedNoop struct{}
 
-func (typedNoop) Reconcile(reconcile.Request) (reconcile.Result, error) {
+func (typedNoop) Reconcile(context.Context, reconcile.Request) (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
@@ -60,7 +60,9 @@ var _ = Describe("application", func() {
 		close(stop)
 	})
 
-	noop := reconcile.Func(func(req reconcile.Request) (reconcile.Result, error) { return reconcile.Result{}, nil })
+	noop := reconcile.Func(func(context.Context, reconcile.Request) (reconcile.Result, error) {
+		return reconcile.Result{}, nil
+	})
 
 	Describe("New", func() {
 		It("should return success if given valid objects", func() {
@@ -302,7 +304,7 @@ func doReconcileTest(nameSuffix string, stop chan struct{}, blder *Builder, mgr 
 
 	By("Creating the application")
 	ch := make(chan reconcile.Request)
-	fn := reconcile.Func(func(req reconcile.Request) (reconcile.Result, error) {
+	fn := reconcile.Func(func(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
 		defer GinkgoRecover()
 		if !strings.HasSuffix(req.Name, nameSuffix) {
 			// From different test, ignore this request.  Etcd is shared across tests.
--- a/pkg/builder/example_test.go
+++ b/pkg/builder/example_test.go
@@ -79,25 +79,24 @@ type ReplicaSetReconciler struct {
 // * Read the ReplicaSet
 // * Read the Pods
 // * Set a Label on the ReplicaSet with the Pod count
-func (a *ReplicaSetReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+func (a *ReplicaSetReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	// Read the ReplicaSet
 	rs := &appsv1.ReplicaSet{}
-	err := a.Get(context.TODO(), req.NamespacedName, rs)
+	err := a.Get(ctx, req.NamespacedName, rs)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
 	// List the Pods matching the PodTemplate Labels
 	pods := &corev1.PodList{}
-	err = a.List(context.TODO(), pods, client.InNamespace(req.Namespace),
-		client.MatchingLabels(rs.Spec.Template.Labels))
+	err = a.List(ctx, pods, client.InNamespace(req.Namespace), client.MatchingLabels(rs.Spec.Template.Labels))
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
 	// Update the ReplicaSet
 	rs.Labels["pod-count"] = fmt.Sprintf("%v", len(pods.Items))
-	err = a.Update(context.TODO(), rs)
+	err = a.Update(ctx, rs)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -43,7 +43,8 @@ type Options struct {
 	// The overall is a token bucket and the per-item is exponential.
 	RateLimiter ratelimiter.RateLimiter
 
-	// Log is the logger used for this controller.
+	// Log is the logger used for this controller and passed to each reconciliation
+	// request via the context field.
 	Log logr.Logger
 }
 
@@ -91,6 +92,10 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		return nil, fmt.Errorf("must specify Name for Controller")
 	}
 
+	if options.Log == nil {
+		options.Log = mgr.GetLogger()
+	}
+
 	if options.MaxConcurrentReconciles <= 0 {
 		options.MaxConcurrentReconciles = 1
 	}
@@ -117,6 +122,6 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
-		Log:                     options.Log.WithName("controller").WithValues("controller", name),
+		Log:                     options.Log.WithName("controller").WithName(name),
 	}, nil
 }
--- a/pkg/controller/controller_integration_test.go
+++ b/pkg/controller/controller_integration_test.go
@@ -62,7 +62,7 @@ var _ = Describe("controller", func() {
 			By("Creating the Controller")
 			instance, err := controller.New("foo-controller", cm, controller.Options{
 				Reconciler: reconcile.Func(
-					func(request reconcile.Request) (reconcile.Result, error) {
+					func(_ context.Context, request reconcile.Request) (reconcile.Result, error) {
 						reconciled <- request
 						return reconcile.Result{}, nil
 					}),
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"time"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
@@ -86,8 +88,10 @@ type watchDescription struct {
 }
 
 // Reconcile implements reconcile.Reconciler
-func (c *Controller) Reconcile(r reconcile.Request) (reconcile.Result, error) {
-	return c.Do.Reconcile(r)
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	ctx = logf.IntoContext(ctx, log)
+	return c.Do.Reconcile(ctx, req)
 }
 
 // Watch implements controller.Controller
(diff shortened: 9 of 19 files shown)


vincepri:
This is mostly a concept, and would like some feedback on the approach

vincepri:
/cc @alvaroaleman @christopherhein @devigned @alexeldeib 

for feedback :)

pkg/reconcile/reconcile.go
```diff
@@ -81,13 +83,20 @@ type Reconciler interface {
...
+type Func func(Context, Request) (Result, error)
 
 var _ Reconciler = Func(nil)
 
 // Reconcile implements Reconciler.
-func (r Func) Reconcile(o Request) (Result, error) { return r(o) }
+func (r Func) Reconcile(ctx Context, o Request) (Result, error) { return r(ctx, o) }
+
+// Context is the context type passed to each Reconcile function, it MUST satisfy the context.Context interface.
+type Context struct {
```
luxas, line 98: Why not an interface embedding `context.Context` and exposing `Logger() logr.Logger` as a function?
That would allow way more flexibility in passing arbitrary implementation structs which can be richer in functionality once casted to the right type, or more detailed interface.

pkg/builder/example_test.go
```diff
@@ -79,7 +79,7 @@ type ReplicaSetReconciler struct {
 // * Read the ReplicaSet
 // * Read the Pods
 // * Set a Label on the ReplicaSet with the Pod count
-func (a *ReplicaSetReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+func (a *ReplicaSetReconciler) Reconcile(ctx reconcile.Context, req reconcile.Request) (reconcile.Result, error) {
 	// Read the ReplicaSet
 	rs := &appsv1.ReplicaSet{}
 	err := a.Get(context.TODO(), req.NamespacedName, rs)
```
luxas, line 85: Swap out these to ctx?

pkg/controller/controller_integration_test.go
```diff
@@ -62,7 +62,7 @@ var _ = Describe("controller", func() {
 			By("Creating the Controller")
 			instance, err := controller.New("foo-controller", cm, controller.Options{
 				Reconciler: reconcile.Func(
-					func(request reconcile.Request) (reconcile.Result, error) {
+					func(ctx reconcile.Context, request reconcile.Request) (reconcile.Result, error) {
```
luxas, line 65: mark as `_` as unused?

pkg/builder/controller_test.go
```diff
@@ -275,7 +277,7 @@ func doReconcileTest(nameSuffix string, stop chan struct{}, blder *Builder, mgr
 
 	By("Creating the application")
 	ch := make(chan reconcile.Request)
-	fn := reconcile.Func(func(req reconcile.Request) (reconcile.Result, error) {
+	fn := reconcile.Func(func(ctx reconcile.Context, req reconcile.Request) (reconcile.Result, error) {
```
luxas, line 280: replace context.TODO below?

pkg/builder/controller_test.go
```diff
@@ -54,7 +54,9 @@ var _ = Describe("application", func() {
 		close(stop)
 	})
 
-	noop := reconcile.Func(func(req reconcile.Request) (reconcile.Result, error) { return reconcile.Result{}, nil })
+	noop := reconcile.Func(func(ctx reconcile.Context, req reconcile.Request) (reconcile.Result, error) {
```
luxas, line 57: unused ctx & req -> `_`

pkg/internal/controller/controller.go
```diff
@@ -228,11 +229,16 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 		return true
 	}
 
+	ctx := reconcile.Context{
+		Context: context.Background(),
+		Log:     c.Log.WithValues("name", req.Name, "namespace", req.Namespace),
```
luxas, line 234: also add gvk information? and/or reconciler name?

alvaroaleman:
Generally I really like it. One thing I am wondering though is how we expect ppl to use this, because they will rightfully want to use the standard `With{Cancel,Deadline,Timeout}` however if they do that, they get a `context.Context` and not a `reconcile.Context`. Maybe it would make sense to provide a `context` package that exposes those?

vincepri:
@luxas The main reason to prefer a struct was to avoid `ctx.Logger().Info()` everywhere, it's mostly for readability but I can see that an interface could be nice.

@alvaroaleman That's a good point, we can have our context utilities that would follow upstream and return our own representation instead

pkg/reconcile/reconcile.go
vincepri: The main reason to prefer a struct was to avoid `ctx.Logger().Info()` everywhere, it's mostly for readability but I can see that an interface could be nice.

vincepri: Actually, if we go with what @alvaroaleman suggested below, creating a `context` package an interface would be great.

We could have `context.Background()` or similar things that would use the global logger, and other functions like `context.WithLogger()` that would take a context and change the internal logger implementation.

pkg/context/context.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// A Context carries a deadline, a cancellation signal, and other values across
+// API boundaries.
+//
+// Context's methods may be called by multiple goroutines simultaneously.
+type Context interface {
```
alvaroaleman, line 31: cc @munnerz I think you did something similiar for cert-manager?

pkg/reconcile/reconcile.go
luxas: Yes :tada:!

luxas: > The main reason to prefer a struct was to avoid ctx.Logger().Info() everywhere, it's mostly for readability but I can see that an interface could be nice.

I don't personally think that's too much of an issue, as the consumer can always do `ctxLog := ctx.Logger()` if they get tired of calling the logger function all the time.

pkg/context/context.go
munnerz: Yes, although we specifically avoided doing what's been done here and adding a new Context type, instead using accessors that accept regular contexts in order to 'read' or 'write' the appropriate logger etc: https://github.com/jetstack/cert-manager/blob/master/pkg/logs/logs.go#L132

I've not had a chance to properly review this PR, but I'd be strongly -1 to creating a new Context type as it'll make interoperability with other packages harder, as the regular 'ctx' being passed into a caller would not be able to just hand this off to a controller-runtime function.

By keeping the logger passing and the context handling isolated from each other (by just making use of context keys to store the logger), it's easy to integrate existing code with cert-manager packages. If that context does not contain a Logger, the 'default' is instead used (either a NullLogger or the 'global' default one).

I'll try and take a proper look at this PR later this week, but just a few thoughts based on our experience 😄

munnerz: IMO a context isn't being used for what it is designed for if we are embedding it into another interface. I'm no expert but it feels like a bit of a code smell to me.

If this is truly the approach we want, in order to ensure that the Log is always explicitly provided by the caller (so no defaulting behaviour happens), it seems more explicit to just pass in the Logger as an argument where required.

In practice though I would lean towards making use of context keys 🙂 then at the start of methods that need it, you'd call something like `log := logf.FromContext(ctx)`.

The only frustrating part here (which is also true for this design too) is the need to re-wrap loggers with contexts if a 'child' logger is made. We have a `NewContext` function to make that easier: https://github.com/jetstack/cert-manager/blob/master/pkg/logs/logs.go#L144

vincepri: Thanks for the context around context @munnerz 😂, we'll have to think about it a little more and so far we're experimenting and see how we'd like to proceed.

> as it'll make interoperability with other packages harder

Can you provide an example for this? 

@luxas @alvaroaleman what do you all think?

pkg/internal/controller/controller.go
vincepri: The reconciler/controller name is already within the c.Log logger, +1 adding gvk info

pkg/context/context.go
alvaroaleman: Well I pinged james after I randomly had a discussion about this with someone yesterday who said mostly the same, the idea of the context is to hold all context and if we now build our own context, we basically break eco-system compatibility because everything else uses the `context.Context` (and that can not be used to retrieve the logger anymore). So I'd vote for not having our own context as well.

vincepri: I'm fine moving the logger somewhere else, @alvaroaleman and I discussed to potentially move it under the reconcile.Request struct if that makes sense and use stdlib's context.Context. The other option (shown here), if we want to go down the path of having our own context library that's a drop-in replacement for the stdlib's one, but be explicit that this should only be used for the current request.

Personally, I'd like to be as explicit as possible and probably not have the drop-in package. If the logger is bundled with the Request, it's probably going to be very clear that that logger is associated with the object that came in. 

Let's hear from more folks though, I'd say I feel a little meh about both approaches 😂

devigned: +1 to using a func with `context.Context` as a param to access extended functionality `to.Logger(ctx).Info()` where `ctx` is a `context.Context`.

You lose a little bit of convenience by not being able to say `ctx.Logger()`, but having a vanilla `context.Context` will be natural to anyone familiar to the stdlib.

mrIncompetent: +1 to using the vanilla `context.Context`. 

The grpc-middleware project has a nice example on how it can be done:
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/zap/ctxzap/context.go#L35

vincepri:
Thank you all for the great feedback. Pushed some changes to use the stdlib's context.Context and have some helper methods to retrieve and set the logger.

Let me know what do you all think, ready for another round 😊

pkg/log/log.go
```diff
@@ -46,3 +52,22 @@ func SetLogger(l logr.Logger) {
...
+		lv := ctx.Value(contextKey)
+		log = lv.(logr.Logger)
+	}
+	log.WithValues(keysAndValues...)
+	return log
+}
+
+// SetContext takes a context and sets the logger as one of its keys.
+// Use FromContext function to retrieve the logger.
+func SetContext(ctx context.Context, log logr.Logger) context.Context {
```
detiber, line 71: Seeing this in use without knowing in advance that it is just adding the logger to a given context is a bit confusing. At first I thought this was setting package level variable the same way that `SetLogger` does:
```suggestion
func ContextWithLogger(ctx context.Context, log logr.Logger) context.Context {
```

pkg/internal/controller/controller.go
```diff
@@ -228,11 +230,14 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
...
+	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	ctx := logf.SetContext(context.Background(), log)
+
 	// RunInformersAndControllers the syncHandler, passing it the namespace/Name string of the
 	// resource to be synced.
-	if result, err := c.Do.Reconcile(req); err != nil {
+	if result, err := c.Do.Reconcile(ctx, req); err != nil {
 		c.Queue.AddRateLimited(req)
-		c.Log.Error(err, "Reconciler error", "name", req.Name, "namespace", req.Namespace)
+		log.Error(err, "Reconciler error")
```
detiber, line 240: Do we need to potentially worry about a user replacing the logger completely in the context? If so, then it might be good to do a `log = logf.FromContext(ctx)` here (and other codepaths where log is used after this call).

vincepri, line 240: Thoughts about removing this? I'd prefer that we don't log for users if they don't want to, or if verbosity isn't high enough

pkg/internal/controller/controller.go
```diff
@@ -256,7 +261,7 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 	c.Queue.Forget(obj)
 
 	// TODO(directxman12): What does 1 mean?  Do we want level constants?  Do we want levels at all?
-	c.Log.V(1).Info("Successfully Reconciled", "name", req.Name, "namespace", req.Namespace)
+	log.V(1).Info("Successfully Reconciled", "name", req.Name, "namespace", req.Namespace)
```
vincepri, line 264: Should we remove this as well, or increase verbosity to 4 or something?

pkg/internal/controller/controller.go
vincepri: I want to remove it if possible, we're logging for users and potentially losing information or polluting the logs

pkg/internal/controller/controller.go
detiber: I think higher verbosity makes sense, I can see it coming in handy for debugging purposes, but generally it's just noise.

pkg/internal/controller/controller.go
mrIncompetent: Even though i don't like this particular log, removing it looks like a breaking change.
Users probably rely on this log message & don't have a own logging of errors.

detiber: +1 on increased verbosity from me by default, I think it would help in debugging

detiber: Similar to others, I think it would be helpful to have this at a higher verbosity by default so that it could help with debugging type purposes, but by +1 to not outputting it by default

ncdc, line 240: One thing that has long been a problem with this log line, at least if you're using klogr, is that the output shows only the basename of **this** file, `controller.go`, as the source of the log message. It does not include any provenance about the error, so you essentially see something like this:

```
controller.go:248] controller-runtime/controller "msg"="Reconciler error" "error"="oops" "controller"="kubeadmcontrolplane" "name"="bmo-control-plane" "namespace"="default"
```

There is no way to tell where the error message actually came from, as controller.go:248 refers to this file.

It would be really nice if:
- We logged the full path of the file
- We logged the full path and line where the error came from (assuming we have stack information for it)

This may not be possible in a generic way.

vincepri: Another good reason to completely remove this log line :)

vincepri:
@detiber Regarding passing context/logger to the EventHandlers, that might mean that we'd need to have an abstraction on top of it. As far as I can tell, today we pass those directly to client-go

pkg/internal/controller/controller.go
ncdc: If we are not able to log the provenance of the error, 👍 to removing it.

ncdc: Or make it V=5 or 10 in case you really want to know that it's recognizing the error in this spot (for the AddRateLimited call)

pkg/log/log.go
alvaroaleman: Since at the end the function call will be prefixed with the package name, what about `IntoContext`? That would mean the call is `log.IntoContext(ctx, log)`

detiber:
> @detiber Regarding passing context/logger to the EventHandlers, that might mean that we'd need to have an abstraction on top of it. As far as I can tell, today we pass those directly to client-go

I was thinking more specifically for the helpers that call user-defined code, which might be limited to `handler.EnqueueRequestsFromMapFunc`

benmoss:
What do you think about adding a UUID to these request contexts by default? This seems like a useful feature to then use with your logger to have a way to do request-scoped logging.

alvaroaleman:
@benmoss time+namespacedName is your UUID here, because the underlying machinery does object-level locking, i.E. only one reconciliation for a given key can happen at a time within one controller

ncdc:
@alvaroaleman time changes, though. If you have dozens or hundreds of log lines that are all part of a single reconcile iteration, they presumably won't all occur at the exact same time. Having a single, unchanging value you can use as a search term is immensely helpful.

pkg/log/log.go
vincepri: +1 either way

alvaroaleman:
> @alvaroaleman time changes, though. If you have dozens or hundreds of log lines that are all part of a single reconcile iteration, they presumably won't all occur at the exact same time. Having a single, unchanging value you can use as a search term is immensely helpful.

Fair point. But then the better approach would probably be to set it as a field on the logger?

ncdc:
> But then the better approach would probably be to set it as a field on the logger?
👍

pkg/builder/controller.go
```diff
@@ -228,26 +238,33 @@ func (blder *Builder) loadRestConfig() {
...
+	gvk, err := getGvk(blder.forInput.object, blder.mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+
+	// Setup the logger.
+	if ctrlOptions.Log == nil {
+		ctrlOptions.Log = logf.Log
+	}
+	ctrlOptions.Log = ctrlOptions.Log.WithValues("groupVersionKind", gvk.String())
```
vincepri, line 265: Do we want to split this into multiple values, or is `groupVersionKind` enough here?

pkg/builder/controller.go
```diff
@@ -145,6 +149,12 @@ func (blder *Builder) WithOptions(options controller.Options) *Builder {
 	return blder
 }
 
+// WithLogger overrides the controller options's logger used. Defaults to nil logger.
```
detiber, line 152: It might be nice to extend this same type of Options helper for setting the logger to `manager.New()` as well, I'm not sure it would completely eliminate the need to initialize a logger with `ctrl.SetLogger`, but might be a good start down that path.

It would also allow a way for controllers built with `ctrl.NewControllerManagedBy(mgr)` to inherit the same base logger without having to configure it in multiple places or rely on `ctrl.SetLogger()` for the behavior.

pkg/builder/controller.go
```diff
@@ -228,26 +238,33 @@ func (blder *Builder) loadRestConfig() {
...
+	// Retrieve the GVK from the object we're reconciling
+	// to prepopulate logger information, and to optionally generate a default name.
+	gvk, err := getGvk(blder.forInput.object, blder.mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+
+	// Setup the logger.
+	if ctrlOptions.Log == nil {
+		ctrlOptions.Log = logf.Log
```
detiber, line 263: Continuing from above, if the manager has a logger, then we could inherit the loger from there rather than defaulting to `logf.Log` directly here.

pkg/builder/controller.go
detiber: I think gvk is probably good here, I do kind of wish we could abbreviate it to `gvk` for the key for brevity, but it's probably better to keep it spelled out. The length is also pretty minimal compared to a full gvk anyway :joy:

pkg/controller/controller.go
```diff
@@ -88,6 +93,10 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		return nil, fmt.Errorf("must specify Name for Controller")
 	}
 
+	if options.Log == nil {
+		options.Log = log.NullLogger{}
```
detiber, line 97: It feels a bit odd to be defaulting to `log.NullLogger` here rather than `logf.Log` as in other places. That said, ideally I think direct inheritance from the manager through the builder for the happy path would be a better happy path for the main use cases and use of the null logger when a logger is not provided for other use cases would be better than the use of a package level variable if we can.

vincepri: Yeah that makes sense, we can default to logf.Log here as well, in case folks want to disable the logging the value can be set to NullLogger

vincepri: Updated, ptal

pkg/builder/controller.go
vincepri: Updated!

pkg/builder/controller.go
vincepri: Done, ptal

pkg/builder/controller.go
vincepri: Seems gvk.String() is `return gvk.Group + "/" + gvk.Version + ", Kind=" + gvk.Kind`, that might be a little weird to parse for folks. I'm also ok to keep it this way and spell it as `gvk`, given that `groupVersionKind` is kinda large in log lines.

Is the group and version information actually useful? Or do we wanto just output the kind?

detiber: I kind of wish we could have different behavior based on log level :joy: I think Kind is probably sufficient in most cases, but I also know that we've wanted to output version information in the past as well (mainly around upgrade scenarios).

alvaroaleman:
The changes look good but IMHO it would be better to put the logger defaulting and enriching with GVK into a distinct PR so we get a distinct changelog and reviewing is easier

pkg/builder/controller.go
alvaroaleman: For `Version` I am not sure if its useful because the different versions are supposively just different views on the same data.

For `Group` and `Kind` we definitely should have both, some clusters have multiple identically named kinds in different api groups (`Cluster` is a common example for this :upside_down_face: )

pkg/internal/controller/controller.go
```diff
@@ -86,8 +88,8 @@ type watchDescription struct {
 }
 
 // Reconcile implements reconcile.Reconciler
-func (c *Controller) Reconcile(r reconcile.Request) (reconcile.Result, error) {
-	return c.Do.Reconcile(r)
+func (c *Controller) Reconcile(ctx context.Context, r reconcile.Request) (reconcile.Result, error) {
```
detiber, line 91: It's not overly clear to me when this `Reconcile()` method may be called as opposed to the logic in `reconcileHandler()`, should it also do similar enrichment of the log values and inject the log into the context here as well?

pkg/builder/controller.go
detiber: lgtm, thanks!

pkg/builder/controller.go
vincepri: I'll split them up!

pkg/builder/controller.go
```diff
@@ -148,6 +148,12 @@ func (blder *Builder) WithOptions(options controller.Options) *Builder {
 	return blder
 }
 
+// WithLogger overrides the controller options's logger used.
+func (blder *Builder) WithLogger(log logr.Logger) *Builder {
+	blder.log = log
+	return blder
+}
+
```
detiber, line 156: Any particular reason to move this method, or just an artifact of the rebase?

vincepri: Both 😂, moved it up at the bottom of the other `With..` methods

vincepri:
This should be good to go as well, if there are no other comments

vincepri:
@alvaroaleman This should also be good to go

examples/crd/main.go
```diff
@@ -29,23 +29,22 @@ import (
...
 
 type reconciler struct {
 	client.Client
 	scheme *runtime.Scheme
 }
 
-func (r *reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	log := recLog.WithValues("chaospod", req.NamespacedName)
+func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx).WithValues("chaospod", req.NamespacedName)
```
detiber, line 46: The use of .WithValues shouldn't be needed here, since you can pass in the key/value pairs to `log.FromContext`
```suggestion
	log := log.FromContext(ctx, "chaospod", req.NamespacedName)
```

examples/builtins/controller.go
```diff
@@ -20,27 +20,25 @@ import (
...
 var _ reconcile.Reconciler = &reconcileReplicaSet{}
 
-func (r *reconcileReplicaSet) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *reconcileReplicaSet) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	// set up a convenient log object so we don't have to type request over and over again
-	log := r.log.WithValues("request", request)
+	log := log.FromContext(ctx)
 
 	// Fetch the ReplicaSet from the cache
 	rs := &appsv1.ReplicaSet{}
```
detiber, line 44: Should the error cases below be updated to use the logger from the context to output error info rather than returning wrapped errors?

examples/crd/main.go
```diff
@@ -29,23 +29,22 @@ import (
...
 	log.V(1).Info("reconciling chaos pod")
-	ctx := context.Background()
 
 	var chaospod api.ChaosPod
 	if err := r.Get(ctx, req.NamespacedName, &chaospod); err != nil {
```
detiber, line 50: Should this example also show plumbing the logger through the manager in `main()` below?

pkg/internal/controller/controller.go
```diff
@@ -229,12 +233,15 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 	}
 
 	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	ctx := logf.IntoContext(context.Background(), log)
```
detiber, line 236: Would `context.TODO()` work a bit better here? It seems like eventually a `WithCancel` context can be plumbed into here from `Start()` that could eventually be used to more cleanly stop in-flight operations rather than just forcefully killing the workers.

vincepri:
@detiber I'll follow-up with another PR for your comments / respond soon

examples/builtins/main.go
```diff
@@ -25,19 +25,20 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
-var log = logf.Log.WithName("example-controller")
+func init() {
```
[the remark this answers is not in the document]
DirectXMan12: I know I'm a bit late to the party here, but it's not clear why we switched to using init here

PR #1058, opened 2020-07-20
⚠ Expose Client runtime.Scheme

vincepri:

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -224,6 +224,10 @@ func (c *fakeClient) List(ctx context.Context, obj runtime.Object, opts ...clien
 	return nil
 }
 
+func (c *fakeClient) Scheme() *runtime.Scheme {
+	return c.scheme
+}
+
 func (c *fakeClient) Create(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
 	createOptions := &client.CreateOptions{}
 	createOptions.ApplyOptions(opts)
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -85,6 +85,7 @@ func New(config *rest.Config, options Options) (Client, error) {
 			cache:      clientcache,
 			paramCodec: noConversionParamCodec{},
 		},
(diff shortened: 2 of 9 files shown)


alvaroaleman:
@vincepri I am all in for the restmapper change, but lets please make it a distinct PR so it gets its own release note :)

vincepri:
I separated the commits because I knew you'd say that :P, wanted to get some eyes on it, given all the PRs all blocked for now

pkg/client/fake/client.go
```diff
@@ -224,6 +224,14 @@ func (c *fakeClient) List(ctx context.Context, obj runtime.Object, opts ...clien
 	return nil
 }
 
+func (c *fakeClient) Scheme() *runtime.Scheme {
+	return c.scheme
+}
+
+func (c *fakeClient) RESTMapper() meta.RESTMapper {
+	return nil
```
vincepri, line 232: Is there a way to have a fake RESTMapper?

alvaroaleman: from a quick glance it seems there is no exported fake for this, only a private one: https://github.com/kubernetes/kubernetes/blob/0051d65f9f30db724dfb88256f70c23e37f6b257/staging/src/k8s.io/client-go/restmapper/shortcut_test.go#L197

At the end of the day we will need to hand-maintain a list of cluster-scoped objects to get this working for unit tests though. I think that should be okay though, because kube doesn't have that many cluster-scoped objects. We can also allow ppl to add additional types and their scope information so this can be used with cluster-scoped crds.

vincepri: Would you be ok with having it nil for now and ask for some contributor to fill this in? I don't find the fake client particularly useful in many cases, but that's just me probably :D

alvaroaleman: yeah sgtm

vincepri: Will do! I'll split this commit in another PR first :)

vincepri:
@alvaroaleman Should be good to go / be merged

PR #1059, opened 2020-07-21
:sparkles: Add IsZero method to reconcile.Result

vincepri:

/assign @alvaroaleman


coderanger:
Might want to use a pointer receiver to avoid a heap copy? Guessing this intended to be fast :)

vincepri:
@coderanger Having it as a non-pointer makes it possible to use it on a nil variable as well, which was the intent behind it

vincepri:
@coderanger I stand corrected, we do want to have it as a pointer receiver to check if it's nil

PR #1061, opened 2020-07-22
Prepare for v0.5.9

vincepri:
/assign @alvaroaleman


alvaroaleman:
@vincepri the graceful shutdown PR means that we now block when we didn't. Would it make sense to default the waitPeriod to 0 seconds for reelase-0.5 to keep the behavior consistent while still allowing to opt into the feature?

vincepri:
@alvaroaleman Removed the graceful shutdown commits, given that it was built on top of other commits that can't be backported (metrics server, leader election changes)

vincepri:
Yay finally passed 🎉

PR #1063, opened 2020-07-22
:seedling: Give more time for prow / goroutine leak test

vincepri:

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -19,6 +19,7 @@ package controller_test
 import (
 	"fmt"
 	rt "runtime"
+	"time"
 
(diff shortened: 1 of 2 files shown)


vincepri:
Seems with these changes we can make more progress

/assign @alvaroaleman

alvaroaleman:
Failed on second try :( I think we should just put it into a dedicated file, give it its own build tag and do a dedicated `go run` invocation to prevent other tests from interfering

vincepri:
It got closer 😂 !

vincepri:
@alvaroaleman Let's see if this + the updated CPU requests work out better

vincepri:
Closing in favor of #1070

/close

PR #1066, opened 2020-07-23
✨ Implement NewPredicateFuncs to create predicate functions from a filter function

### LINKED ISSUE #1052: Make writing predicates easy
Right now, writing predicates involves creating four functions to either directly implement `./pkg/predicate.Predicate` or `./pkg/predicate.Funcs`. This should be easier, it should be possible to define just one predicate function that is then re-used for all kind of event types. Something like this: https://github.com/kubermatic/kubermatic/blob/8243232500ae34839b1e7f6f821e876baf080835/pkg/controller/util/predicate/predicate.go#L30

/good-first-issue

jiachengxu:
Fixes #1052

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -17,6 +17,9 @@ limitations under the License.
 package predicate
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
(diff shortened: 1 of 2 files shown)


pkg/predicate/predicate.go
```diff
@@ -91,6 +94,26 @@ func (p Funcs) Generic(e event.GenericEvent) bool {
...
+// NewPredicateFuncs returns a predicate funcs that applies the given filter function
+// on CREATE, UPDATE, DELETE and GENERIC events. For UPDATE events, the filter is applied
+// to both the old and new object and OR's the result.
+func NewPredicateFuncs(filter func(meta metav1.Object, object runtime.Object) bool) Funcs {
+	return Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return filter(e.Meta, e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return filter(e.MetaOld, e.ObjectOld) || filter(e.MetaNew, e.ObjectNew)
```
alvaroaleman, line 106: @jiachengxu IMHO it makes more sense to not OR but filter only on the new object

jiachengxu: Thanks for the review, I also think it makes more sense. Fixed.

PR #1070, opened 2020-07-23
Mark leak goroutine tests as pending

vincepri:

This PR disables the leak goroutine tests that are causing a lot of flakes in CI. After a conversation on Slack with @alvaroaleman, we agreed that these tests are very error-prone and should be revisited/rewritten later on.

/milestone v0.6.x


alvaroaleman:
Created https://github.com/kubernetes-sigs/controller-runtime/issues/1071 for the data race not sure why it only appears sometimes

PR #1072, opened 2020-07-24
🐛  Enable Logger Injection for Controller Manager

### LINKED ISSUE #1056: Why is a Logger not injected by the Controller Manager?
Hi there,

Along with the other injector-style interfaces, there is `InjectLogger`, https://github.com/kubernetes-sigs/controller-runtime/blob/57ed2a3ccc943a6a3a07300771326a7e930a9671/pkg/runtime/inject/inject.go#L149-L153

However, the only place this seems to be used is [with respect to webhooks](https://github.com/kubernetes-sigs/controller-runtime/search?q=InjectLogger&unscoped_q=InjectLogger).

Why is `InjectLogger` not part of the `SetFields` function in the Controller Manager so loggers are added to components added to the manager at https://github.com/kubernetes-sigs/controller-runtime/blob/ea795f6400dc064dce43c0aa6a065f47e0633077/pkg/manager/internal.go#L244-L270

jiachengxu:
This PR fixes #1056 
Now Logger can be injected by the controller manager to the Object which implements the `inject.Logger` interface.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -244,6 +244,9 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	if _, err := inject.MapperInto(cm.mapper, i); err != nil {
 		return err
 	}
+	if _, err := inject.LoggerInto(log, i); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/lgtm
/approve
@vincepri can we add this to the v0.6 milestone?

PR #1075, opened 2020-07-26
:running: Proposal to extract cluster-specifics out of the Manager

alvaroaleman:
A proposal based on https://github.com/kubernetes-sigs/controller-runtime/pull/950
/assign @mengqiy @vincepri @estroz

### DIFF
--- a/designs/move-cluster-specific-code-out-of-manager.md
+++ b/designs/move-cluster-specific-code-out-of-manager.md
@@ -1,0 +1,228 @@
+Move cluster-specific code out of the manager
+===================
+
+## Motivation
+
+Today, it is already possible to use controller-runtime to build controllers that act on
+more than one cluster. However, this is undocumented and not straight-forward, requiring
+users to look into the implementation details to figure out how to make this work.
+
+## Goals
+
+* Provide an easy-to-discover way to build controllers that act on multiple clusters
+* Decouple the management of `Runnables` from the construction of "things that require a kubeconfig"
+* Do not introduce changes for users that build controllers that act on one cluster only
+
+## Non-Goals
+
+## Proposal
+
+Currently, the `./pkg/manager.Manager` has two purposes:
+
+* Handle running controllers/other runnables and managing their lifecycle
+* Setting up various things to interact with the Kubernetes cluster,
+  for example a `Client` and a `Cache`
+
+This works very well when building controllers that talk to a single cluster,
+however some use-cases require controllers that interact with more than
+one cluster. This multi-cluster usecase is very awkward today, because it
+requires to construct one manager per cluster and adding all subsequent
+managers to the first one.
+
+This document proposes to move all cluster-specific code out of the manager
+and into a new package and interface, that then gets embedded into the manager.
+This allows to keep the usage for single-cluster cases the same and introduce
+this change in a backwards-compatible manner.
+
+Furthermore, the manager gets extended to start all caches before any other
+`runnables` are started.
+
+
+The new `Cluster` interface will look like this:
+
+```go
+type Cluster interface {
+	// SetFields will set cluster-specific dependencies on an object for which the object has implemented the inject
+	// interface, specifically inject.Client, inject.Cache, inject.Scheme, inject.Config and inject.APIReader
+	SetFields(interface{}) error
+
+	// GetConfig returns an initialized Config
+	GetConfig() *rest.Config
+
+	// GetClient returns a client configured with the Config. This client may
+	// not be a fully "direct" client -- it may read from a cache, for
+	// instance.  See Options.NewClient for more information on how the default
+	// implementation works.
+	GetClient() client.Client
+
+	// GetFieldIndexer returns a client.FieldIndexer configured with the client
+	GetFieldIndexer() client.FieldIndexer
+
+	// GetCache returns a cache.Cache
+	GetCache() cache.Cache
+
+	// GetEventRecorderFor returns a new EventRecorder for the provided name
+	GetEventRecorderFor(name string) record.EventRecorder
+
+	// GetRESTMapper returns a RESTMapper
+	GetRESTMapper() meta.RESTMapper
+
+	// GetAPIReader returns a reader that will be configured to use the API server.
+	// This should be used sparingly and only when the client does not fit your
+	// use case.
+	GetAPIReader() client.Reader
+
+	// GetScheme returns an initialized Scheme
+	GetScheme() *runtime.Scheme
+
+	// Start starts the connection tothe Cluster
+	Start(<-chan struct{}) error
+}
+```
+
+And the current `Manager` interface will change to look like this:
+
+```go
+type Manager interface {
+	// Cluster holds objects to connect to a cluster
+	cluser.Cluster
+
+	// Add will set requested dependencies on the component, and cause the component to be
+	// started when Start is called.  Add will inject any dependencies for which the argument
+	// implements the inject interface - e.g. inject.Client.
+	// Depending on if a Runnable implements LeaderElectionRunnable interface, a Runnable can be run in either
+	// non-leaderelection mode (always running) or leader election mode (managed by leader election if enabled).
+	Add(Runnable) error
+
+	// Elected is closed when this manager is elected leader of a group of
+	// managers, either because it won a leader election or because no leader
+	// election was configured.
+	Elected() <-chan struct{}
+
+	// SetFields will set any dependencies on an object for which the object has implemented the inject
+	// interface - e.g. inject.Client.
+	SetFields(interface{}) error
+
+	// AddMetricsExtraHandler adds an extra handler served on path to the http server that serves metrics.
+	// Might be useful to register some diagnostic endpoints e.g. pprof. Note that these endpoints meant to be
+	// sensitive and shouldn't be exposed publicly.
+	// If the simple path -> handler mapping offered here is not enough, a new http server/listener should be added as
+	// Runnable to the manager via Add method.
+	AddMetricsExtraHandler(path string, handler http.Handler) error
+
+	// AddHealthzCheck allows you to add Healthz checker
+	AddHealthzCheck(name string, check healthz.Checker) error
+
+	// AddReadyzCheck allows you to add Readyz checker
+	AddReadyzCheck(name string, check healthz.Checker) error
+
+	// Start starts all registered Controllers and blocks until the Stop channel is closed.
+	// Returns an error if there is an error starting any controller.
+	// If LeaderElection is used, the binary must be exited immediately after this returns,
+	// otherwise components that need leader election might continue to run after the leader
+	// lock was lost.
+	Start(<-chan struct{}) error
+
+	// GetWebhookServer returns a webhook.Server
+	GetWebhookServer() *webhook.Server
+}
+```
+
+Furthermore, during startup, the `Manager` will use type assertion to find `Cluster`s
+to be able to start their caches before anything else:
+
+```go
+type HasCaches interface {
+  GetCache()
+}
+if getter, hasCaches := runnable.(HasCaches); hasCaches {
+	m.caches = append(m.caches, getter())
+}
+```
+
+```go
+for idx := range cm.caches {
+	go func(idx int) {cm.caches[idx].Start(cm.internalStop)}
+}
+
+for _, cache := range cm.caches {
+	cache.WaitForCacheSync(cm.internalStop)
+}
+
+// Start all other runnables
+```
+
+## Example
+
+Below is a sample `reconciler` that will create a secret in a `mirrorCluster` for each
+secret found in `referenceCluster` if none of that name already exists. To keep the sample
+short, it won't compare the contents of the secrets.
+
+```go
+type secretMirrorReconciler struct {
+	referenceClusterClient, mirrorClusterClient client.Client
+}
+
+func (r *secretMirrorReconciler) Reconcile(r reconcile.Request)(reconcile.Result, error){
+	s := &corev1.Secret{}
+	if err := r.referenceClusterClient.Get(context.TODO(), r.NamespacedName, s); err != nil {
+		if kerrors.IsNotFound{ return reconcile.Result{}, nil }
+		return reconcile.Result, err
+	}
+
+	if err := r.mirrorClusterClient.Get(context.TODO(), r.NamespacedName, &corev1.Secret); err != nil {
+		if !kerrors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+
+		mirrorSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{Namespace: s.Namespace, Name: s.Name},
+			Data: s.Data,
+		}
+		return reconcile.Result{}, r.mirrorClusterClient.Create(context.TODO(), mirrorSecret)
+	}
+
+	return nil
+}
+
+func NewSecretMirrorReconciler(mgr manager.Manager, mirrorCluster cluster.Cluster) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		// Watch Secrets in the reference cluster
+		For(&corev1.Secret{}).
+		// Watch Secrets in the mirror cluster
+		Watches(
+			source.NewKindWithCache(&corev1.Secret{}, mirrorCluster.GetCache()),
+			&handler.EnqueueRequestForObject{},
+		).
+		Complete(&secretMirrorReconciler{
+			referenceClusterClient: mgr.GetClient(),
+			mirrorClusterClient:    mirrorCluster.GetClient(),
+		})
+	}
+}
+
+func main(){
+
+	mgr, err := manager.New(cfg1, manager.Options{})
+	if err != nil {
+		panic(err)
+	}
+
+	mirrorCluster, err := cluster.New(cfg2)
+	if err != nil {
+		panic(err)
+	}
+
+	if err := mgr.Add(mirrorCluster); err != nil {
+		panic(err)
+	}
+
+	if err := NewSecretMirrorReconciler(mgr, mirrorCluster); err != nil {
+		panic(err)
+	}
+
+	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
+		panic(err)
+	}
+}
+```


designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
...
+This document proposes to move all cluster-specific code out of the manager
+and into a new package and interface, that then gets embedded into the manager.
+This allows to keep the usage for single-cluster cases the same and introduce
+this change in a backwards-compatible manner.
+
+Furthermore, the manager gets extended to start all caches before any other
+`runnables` are started.
+
+
+The new `ClusterConnector` interface will look like this:
```
alvaroaleman, line 25: Ideas for a better name are very welcome :)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
...
+	GetWebhookServer() *webhook.Server
+}
+```
+
+Furthermore, during startup, the `Manager` will use type assertion to find `ClusterConnector`s
+to be able to start their caches before anything else:
+
+```go
+if cc, isClusterConnector:= runnable.(clusterconnector.ClusterConnector); isClusterConnector {
+	m.caches = append(m.caches, cc.GetCache())
```
alvaroaleman, line 120: This makes the whole thing a bit weird, because we will end up starting both the `ClusterConnectors` Cache and the `ClusterConnector` separately. The alternate would be to require ppl to do a `mgr.Add(clusterConnector.GetCache)` which I would like to avoid because its unintuitive.

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
...
+	// If LeaderElection is used, the binary must be exited immediately after this returns,
+	// otherwise components that need leader election might continue to run after the leader
+	// lock was lost.
+	Start(<-chan struct{}) error
+
+	// GetWebhookServer returns a webhook.Server
+	GetWebhookServer() *webhook.Server
+}
+```
+
```
kramerul, line 114: Would it be possible to separate this into 3 interfaces?

```go
type cluserconnector.ClusterConnector{
    // same as above
...
}}
type MultiClusterManager interface {
    // Create a new connection to a cluster. Creates a new instance of ClusterConnector and adds it to this manager
    Connect(config *rest.Config, name string, opts ...Option) cluserconnector.ClusterConnector
    // same as Manager from proposal except cluserconnector.ClusterConnector
    Add(Runnable) error
    Elected() <-chan struct{}
    SetFields(interface{}) error
    AddMetricsExtraHandler(path string, handler http.Handler) error
    AddHealthzCheck(name string, check healthz.Checker) error
    Start(<-chan struct{}) error
    GetWebhookServer() *webhook.Server
}

type Manager interface {
     cluserconnector.ClusterConnector
     MultiClusterManager
}
```

Introducing a factory method `Connect` inside `MultiClusterManager` would also solve the problems a having to use type assertions.

designs/move-cluster-specific-code-out-of-manager.md
jiachengxu: Or maybe`PeerCluster` or `ClusterPeer`? :)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
...
+		Watches(
+			source.NewKindWithCache(&corev1.Secret{}, mirrorCluster.GetCache()),
+			&handler.EnqueueRequestForObject{},
+		).
+		Complete(&secretMirrorReconciler{
+			referenceClusterClient: mgr.GetClient(),
+			mirrorClusterClient:    mirrorCluster.GetClient(),
+		})
+	}
+}
```
jiachengxu, line 183: @alvaroaleman I like this approach, currently, we can do something like the following to achieve the same feature with only one manager:
```go
func  (r *secretMirrorReconciler) SetupWithManager(mgr manager.Manager, mirrorClusterCache cache.Cache) error {
	return ctrl.NewControllerManagedBy(mgr).
		// Watch Secrets in the reference cluster
		For(&corev1.Secret{}).
		// Watch pods in the mirror cluster
		Watches(
			source.NewKindWithCache(&corev1.Secret{}, mirrorClusterCache),
			&handler.EnqueueRequestForObject{},
		).
		Complete(r)
	}
}
```
in `main.go` we can do:
```go
func main(){
	mgr, err := manager.New(cfg1, manager.Options{})
	if err != nil {
		panic(err)
	}
	mirrorMapper, err := apiutil.NewDiscoveryRESTMapper(mirrorCfg)
	if err != nil {
		panic(err)
	}
	mirrorClient, err := client.New(mirrorCfg, client.Options{
		Scheme: mirrorScheme,
		Mapper: mirrorMapper,
	})
	if err != nil {
		panic(err)
	}
	mirrorCache, err := cache.New(mirrorCfg, cache.Options{
		Scheme: mirrorScheme,
		Mapper: mirrorMapper,
	})
	if err = mgr.Add(mirrorCache); err != nil {
		panic(err)
	}
	if err := mgr.Add(mirrorConnector); err != nil {
		panic(err)
	}
	if err := (&SecretMirrorReconciler{
	        ReferenceClusterClient: mgr.GetClient(),
		MirrorClusterClient:    mirrorCluster.GetClient(),
	}).SetupWithManager(mgr, mirrorCache); err != nil {
		panic(err)
	}
	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
		panic(err)
	}
}
```
Which requires a lot of work & setup, if we have the connector implemented, should be much easier to do cross-cluster stuff :)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
+Move cluster-specific code out of the manager
+===================
```
vincepri, line 2: Can we add these sections to clearly define the scope and the problem statement:

- Motivation
- Goal / non-goals for this proposal
- Requirements

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: @kramerul Generally I like the idea (although I would probably embedd the `Manager` ito the `MultiClusterManager` and not the other way round) but it opens up an interesting question: What config will we use for leader election?

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: @jiachengxu Yeah I am aware that this is possible, but it is very inconvenient to use and until recently it required rebuilding the `DefaultNewClient` because it wasn't exported (your example uses a api-backed `Reader` for the mirror cluster even though you have a cache)

designs/move-cluster-specific-code-out-of-manager.md
kramerul: @alvaroaleman, I haven't thought about this aspect. But it's absolutely correct, that you need to have a primary connection for leader election. In this case I would agree to have only one `Manager` interface.

The `Connect` method could ease the usage of the API.

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,209 @@
...
+	Add(Runnable) error
+
+	// Elected is closed when this manager is elected leader of a group of
+	// managers, either because it won a leader election or because no leader
+	// election was configured.
+	Elected() <-chan struct{}
+
+	// SetFields will set any dependencies on an object for which the object has implemented the inject
+	// interface - e.g. inject.Client.
+	SetFields(interface{}) error
```
mengqiy, line 88: Not sure what will happen when the embeded `ClusterConnector` also has `SetFields` method

designs/move-cluster-specific-code-out-of-manager.md
mengqiy: I like the idea of having an explicit interface for MultiCluster manager.

> I would probably embedd the Manager ito the MultiClusterManager and not the other way round)

+1

A possible alternative is:

```
type ClusterConnector{
    // same as above
}

// For single cluster use-case only
type Manager interface {
     // same as above
}

type MultiClusterManager interface {
    Manager

    // Create a new connection to a secondary cluster. Creates a new instance of ClusterConnector and adds it to this manager. Each ClusterConnector should have its unique name.
    ConnectSecondaryCluster(config *rest.Config, name string, opts ...Option) (cluserconnector.ClusterConnector, error)
    // Look up the ClusterConnector by its unique name.
    GetClusterConnectorFor(name string) (cluserconnector.ClusterConnector, error)
}
```

designs/move-cluster-specific-code-out-of-manager.md
mengqiy: It's probably fine to do type assertion here, since this's internal implementation details.

> The alternate would be to require ppl to do a mgr.Add(clusterConnector.GetCache) which I would like to avoid because its unintuitive.

+1

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: Regarding naming: What about just `Cluster`? Its pretty much what this represents

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: AFAIK this is fine as long as they have the same signature

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: I've added it like this:
```
type MultiClusterManager interface {
       Manager
       // Add another named cluster to the MultiClusterManager. The name
       // must be unique. The MultiClusterManager will wait for all clusters
       // caches to be started before starting anytything else.
       AddCluster(config *rest.Config, name string) error
       // GetClusters returns all Clusters this MultiClusterManager knows about.
       // The cluster used to construct the MultiClusterManager is named `primary`
       // and will be used for LeaderElection, if enabled.
       GetClusters() map[string]clusterconnector.ClusterConnector
}
```

In all the multi-cluster controllers I've built so far, the actual controller just wants to watch in all clusters but doesn't necessarily know the number or names of them ahead of time which is why a method that returns a map of all clusters is IMHO more useful. If there is one that gets a special treatment, this is still possible.

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+	// GetAPIReader returns a reader that will be configured to use the API server.
+	// This should be used sparingly and only when the client does not fit your
+	// use case.
+	GetAPIReader() client.Reader
+
+	// GetScheme returns an initialized Scheme
+	GetScheme() *runtime.Scheme
+
+	// Start starts the ClusterConnector
+	Start(<-chan struct{}) error
```
DirectXMan12, line 79: are we going to integrate the context work here?

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: A number of the patterns I've seen for multicluster end up having clusters come & go over time.  Even if we're not going to tackle that now, we should keep that in mind while designing.

I'm also not certain about the whole "primary" / "secondary" cluster thing.  Perhaps modeling this as `there's a "leader election" cluster and "functional" clusters (one of which may be the leader election cluster)`?

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: can't we just do:

```
type HasCaches interface {
  GetCache() ...
}

if getter, hasCaches := runnable.(hasCaches); hasCaches {
  ...
}
```

(could make a additional method to avoid accidentally conflicting too)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func (r *secretMirrorReconciler) Reconcile(r reconcile.Request)(reconcile.Result, error){
+	s := &corev1.Secret{}
+	if err := r.referenceClusterClient.Get(context.TODO(), r.NamespacedName, s); err != nil {
+		if kerrors.IsNotFound{ return reconcile.Result{}, nil }
+		return reconcile.Result, err
+	}
+
+	if err := r.mirrorClusterClient.Get(context.TODO(), r.NamespacedName, &corev1.Secret); err != nil {
```
DirectXMan12, line 187: Tangent: Not that we need to tackle it here, but we could even create a multicluster client that (ab)used the unused `ClusterName` field (maybe dangerous), or used a client option, context field, etc to avoid needing multiple clients.

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+		if !kerrors.IsNotFound(err) {
```
DirectXMan12, line 188: Nit: `client.IgnoreNotFound(err)` exists ;-)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+
+func NewSecretMirrorReconciler(mgr manager.Manager, mirrorConnector clusterconnector.ClusterConnector) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		// Watch Secrets in the reference cluster
+		For(&corev1.Secret{}).
+		// Watch pods in the mirror cluster
+		Watches(
+			source.NewKindWithCache(&corev1.Secret{}, mirrorCluster.GetCache()),
+			&handler.EnqueueRequestForObject{},
+		).
```
DirectXMan12, line 210: Shouldn't we also be watching the original secret?  How does that look under this pattern? (probably like the "owner reference but via annotation" pattern)

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func main(){
+
+	mgr, err := manager.New(cfg1, manager.Options{})
+	if err != nil {
+		panic(err)
+	}
+
+	mirrorClusterConnector, err := clusterconnector.New(cfg2)
```
DirectXMan12, line 225: should we leave the possibility for options here?

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,242 @@
...
+`runnables` are started.
+
+
+The new `ClusterConnector` interface will look like this:
+
+```go
+type ClusterConnector interface {
+	// SetFields will set cluster-specific dependencies on an object for which the object has implemented the inject
+	// interface, specifically inject.Client, inject.Cache, inject.Scheme, inject.Config and inject.APIReader
+	SetFields(interface{}) error
```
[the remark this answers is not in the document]
alvaroaleman: Its an unrelated topic but IMHO we should try to get rid of all of it, because it makes changes to it runtime errors and not compile-time errors which is very bad

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: Yeah, this proposal is orthogonal to the context work, the `MultiClusterManager` will make use of the same `Runnable` interface as the normal `Manager`, whatever that is at time of implementation

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: That sounds like a great idea (but I would prefer to keep that as an orthogonal work item)

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: I don't understand how that is of use, we have to return on NotFound so its handled differently from `no error`

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: See above:
```
		// Watch Secrets in the reference cluster
￼		For(&corev1.Secret{}).
```

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: IMHO we should use functional opts once we have a need for that

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: > A number of the patterns I've seen for multicluster end up having clusters come & go over time. Even if we're not going to tackle that now, we should keep that in mind while designing.

IMHO this proposal should do nothing to prevent that but it does not need to do anything yet to simplify that. Once we want such a functionality, we would probably add a `RemoveCluster` to the interface.

> I'm also not certain about the whole "primary" / "secondary" cluster thing. Perhaps modeling this as there's a "leader election" cluster and "functional" clusters (one of which may be the leader election cluster)?

I guess we could probably change `AddCluster` to have a `UseForLeaderElection` bool opt if leader election is enabled, use that cluster (and error out when starting and note exactly one cluster is configured to be used for leader election). WDYT?

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: Updated

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: On second thought, after trying to update the sample, I would like to not make a `MultiClusterManager` part of this proposal, because it opens up a set of questions I would like to not answer just yet:

* The `Builder` needs to be extended to have something like `WatchesFromCluster(source, "cluster-name"`, handler)`, otherwise we need to blindly dereference elements from the `map[clusterName]cluster.Cluster`, potentially getting NPDs
* The `Builder` needs to be extended to have something like `WatchesFromAllClustersExcept(source, "exepcted-cluster-name", handler) 
* Probably more use cases for watches I didn't think about
* There should be a nice way to pass on all clients for all clusters to a controller that doesn't involve writing a loop (probably the multi cluster client abstraction that Solly suggested?)
* When adding official support for building a controller that watches an arbitraty number of clusters, we also need an official way of encoding the cluster name into the `reconcile.Request`

And probably more.

IMHO, extracting the clusterspecifics out of the manager won't block any of the work mentioned above but allows us to start working on the topic without requiring us to already anticipate all use cases.

alvaroaleman:
Everyone, I think I have responded to all feedback, PTAL

designs/move-cluster-specific-code-out-of-manager.md
alvaroaleman: Cluster it is now

designs/move-cluster-specific-code-out-of-manager.md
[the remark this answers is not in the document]
DirectXMan12: agreed.  I'd like to see a proposal for how to cleanly do that (this is not to sound dismissive or snarky -- I genuinely would like to see a proposal :-) )

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: ok, fine with doing that in a follow-up

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: oh, the pattern just becomes `if client.IgnoreNotFound(err) != nil`.  Mainly just avoids remembering another k8s package to import.

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+Furthermore, during startup, the `Manager` will use type assertion to find `Cluster`s
+to be able to start their caches before anything else:
+
+```go
+type HasCaches interface {
+  GetCache()
+}
+if getter, hasCaches := runnable.(hasCaches); hasCaches {
```
DirectXMan12, line 138: nit:

```suggestion
if getter, hasCaches := runnable.(HasCaches); hasCaches {
```

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,228 @@
...
+			ObjectMeta: metav1.ObjectMeta{Namespace: s.Namespace, Name: s.Name},
+			Data: s.Data,
+		}
+		return reconcile.Result{}, r.mirrorClusterClient.Create(context.TODO(), mirrorSecret)
+	}
+
+	return nil
+}
+
+func NewSecretMirrorReconciler(mgr manager.Manager, mirrorConnector clusterconnector.ClusterConnector) error {
```
DirectXMan12, line 188: nit: are we using the name `Cluster` or `ClusterConnector`?  Above you mention `Cluster`

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: :woman_facepalming: nevermind :-P

designs/move-cluster-specific-code-out-of-manager.md
```diff
@@ -0,0 +1,228 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		// Watch Secrets in the reference cluster
+		For(&corev1.Secret{}).
+		// Watch pods in the mirror cluster
```
DirectXMan12, line 192: nit:
```suggestion
		// Watch secrets in the mirror cluster
```

designs/move-cluster-specific-code-out-of-manager.md
DirectXMan12: ack, need to specify that in the signature so it's not breaking.

We should probably do a review before v1 and figure out spots where we should put functional options in (e.g. `manager.New` probably wants functional options).

DirectXMan12 (approved):
very minor nits, otherwise

/approve

(will add the LGTM once nits are addressed, since it'll be removed anyway.  Really would love "lgtm with nits" in k8s :-/)

alvaroaleman:
@DirectXMan12 updated to incorporate your suggestions

DirectXMan12:
test failure is https://github.com/kubernetes-sigs/controller-runtime/issues/1171

ashishvya:
@alvaroaleman Is this proposal got implemented? If not do you have any tracking ticket for it.

alvaroaleman:
@ashishvya it is implemented

alvaroaleman:
@taragu the implementation of this is in pkg/cluster, you can use the git history to find the relevant PRs

PR #1076, opened 2020-07-27
⚠️ Change default webhook port to 9443

### LINKED ISSUE #1018: Default webhook port (443) is problematic for operators running as non-root
If left unset, the manager [defaults to running the webhook server on port 443](https://github.com/kubernetes-sigs/controller-runtime/blob/add0b6444f5091f56824b070d28f930090d8deab/pkg/webhook/server.go#L39-L40).

Since it is a security best practice to run pods as a non-root user, this default can cause unexpected problems for developers using controller-runtime directly (nonroot users typically can't open listeners on low-numbered ports).

In Kubebuilder, the default `main.go` scaffold hardcodes this value to `9443`, which solves the problem for users of Kubebuilder that scaffolded their project somewhat recently, but all other users are left figuring this out on their own.

I'm wondering if we should change the default value to `9443` to match Kubebuilder's current scaffold configuration? 

**This _would_ be a breaking change.**

jiachengxu:
Fixes #1018 
This PR changes the default port to serve webhook from 443 to 9443 since 443 causes some issues for no-root users.

### DIFF
--- a/pkg/builder/builder_suite_test.go
+++ b/pkg/builder/builder_suite_test.go
@@ -73,7 +73,7 @@ var _ = AfterSuite(func() {
 	metrics.DefaultBindAddress = ":8080"
 
 	// Change the webhook.DefaultPort back to the original default.
-	webhook.DefaultPort = 443
+	webhook.DefaultPort = 9443
 })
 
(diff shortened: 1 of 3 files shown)


alvaroaleman:
Ideally we would have a "You really must read this" section like kube, since this change will break users without failing compilation :grimacing: 

/lgtm

PR #1077, opened 2020-07-27
🐛 Test can inspect stdout & stderr: Give some time to exit

alvaroaleman:
Hopefully will fix our top failing test. I believe it is failing because we expect it to be exited immediately after it started. However it might need a bit more time after printing the successful start message to be exited.

### DIFF
--- a/pkg/internal/testing/integration/internal/process_test.go
+++ b/pkg/internal/testing/integration/internal/process_test.go
@@ -216,7 +216,7 @@ var _ = Describe("Start method", func() {
 			processState.StartTimeout = 1 * time.Second
 
 			Expect(processState.Start(stdout, stderr)).To(Succeed())
-			Expect(processState.Session).Should(gexec.Exit())
+			Eventually(processState.Session).Should(gexec.Exit())
 
 			Expect(stdout.String()).To(Equal("that is stdout\n"))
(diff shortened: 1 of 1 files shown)


alvaroaleman:
The job failed one of five times and that because of a different error so I think this PR definitely improves the situation: https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=controller-runtime&pr=1077

/assign @vincepri

vincepri:
Should we increase the eventually global timeout, to something like 10-20 seconds?

vincepri:
```
SetDefaultEventuallyPollingInterval(100 * time.Millisecond)
SetDefaultEventuallyTimeout(30 * time.Second)
```
This is what we use in Cluster API

alvaroaleman:
@vincepri hmhm, so if possible I would prefer to only do that on individual testcases when we know they need that (does ginkgo allow that?)

vincepri:
Yeah, that would be the second argument of Eventually

alvaroaleman:
@vincepri yeah then I would be in favor of just doing that when we see a need rather than globally

alvaroaleman:
@vincepri can we merge this or is there something unaddressed left?

PR #1080, opened 2020-07-28
⚠️ Add support to read webhook configurations from files for WebhookInstallOptions

### LINKED ISSUE #1019: webhookinstalloptions should support individual files
We should support specifying paths to individual files in webhookinstalloptions, like we do for CRDs.

/kind feature
/priority important-longterm

jiachengxu:
Fixes #1019 
This PR adds support to read webhook configurations from files for `WebhookInstallOptions`

### DIFF
--- a/pkg/envtest/webhook.go
+++ b/pkg/envtest/webhook.go
@@ -40,8 +40,8 @@ import (
 
 // WebhookInstallOptions are the options for installing mutating or validating webhooks
 type WebhookInstallOptions struct {
-	// Paths is a list of paths to the directories containing the mutating or validating webhooks yaml or json configs.
-	DirectoryPaths []string
+	// Paths is a list of paths to the directories or files containing the mutating or validating webhooks yaml or json configs.
+	Paths []string
 
 	// MutatingWebhooks is a list of MutatingWebhookConfigurations to install
 	MutatingWebhooks []runtime.Object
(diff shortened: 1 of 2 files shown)


pkg/envtest/webhook.go
```diff
@@ -40,8 +40,8 @@ import (
 
 // WebhookInstallOptions are the options for installing mutating or validating webhooks
 type WebhookInstallOptions struct {
-	// Paths is a list of paths to the directories containing the mutating or validating webhooks yaml or json configs.
-	DirectoryPaths []string
+	// Paths is a list of paths to the directories or files containing the mutating or validating webhooks yaml or json configs.
+	Paths []string
```
hasbro17, line 44: Renaming `DirectoryPaths ` to `Paths` would be a breaking change right?
If so, we should probably mark this commit with a :warning: prefix.

Unless we want to avoid that and add a separate field e.g `FilePaths`.

jiachengxu: yeah, you are right, it is a breaking change.

I prefer to not add `FilePaths` since I would like to keep it consistent with CRDInstallOptions.

WDYT? @vincepri @DirectXMan12

vincepri: +1 for consistency, we can delay this change until v0.7

jiachengxu:
@alvaroaleman @vincepri How can I make the `pull-controller-runtime-apidiff-master ` happy? This PR does introduce breaking changes :).

alvaroaleman:
You can't, but having it pass is not mandatory for merges, its more of an informational thing.
/approve
/assign @vincepri

PR #1087, opened 2020-07-30
🐛 Use metav1.Object instead of Object

boekkooi-fresh:
In order to allow for a broader adoption of ContainsFinalizer, RemoveFinalizer and AddFinalizer and remove a BC break introduced to 0.5.x I switched the method back to the `metav1.Object`.

After reading https://github.com/kubernetes-sigs/controller-runtime/issues/959 and https://github.com/kubernetes-sigs/controller-runtime/pull/962 I'm unable to understand the benefit of requiring `runtime.Object` so I propose to revert the BC break and possible panic on line 285.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -238,7 +238,7 @@ func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
 type MutateFn func() error
 
 // AddFinalizer accepts an Object and adds the provided finalizer if not present.
-func AddFinalizer(o Object, finalizer string) {
+func AddFinalizer(o metav1.Object, finalizer string) {
 	f := o.GetFinalizers()
 	for _, e := range f {
 		if e == finalizer {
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@vincepri I think this would be a good candidate for 0.7
/ok-to-test

vincepri:
I'm not sure if we want to go down this path, we want to make sure objects are "Kubernetes Objects" going forward, or that was the original intent.

alvaroaleman:
Yeah I apparantly didn't read the change properly. I think what we want for all finalizer-related funcs is to work on an `Object` instead of a `metav1.Object` or `runtime.Object`

vincepri:
I think so, I also wouldn't be opposed to move the interface under the main package, maybe under the alias.go file?

boekkooi-fresh:
I have to agree with @vincepri here that moving this under the main package maybe better for the `Object`.

Regarding my main issue is that between `v0.5.0` and `v0.6.1` I had to change my controller code from using `metav1.Object` to `controllerutil.Object` to keep using the `controllerutil.RemoveFinalizer` etc.  This also caused a bunch of tests to break and as far as I can see none of the now required `runtime.Object` methods are ever used by `controllerutil.*Finalizer` making me wonder why this change was made. 
Since I was unable to determine why this change was made based on the related PR's I assume this was an oversight and opened this PR.

vincepri:
@boekkooi-fresh Apologies this caused a breaking change for you, we're trying to move to this object interface internally in controller runtime to avoid casting at runtime, instead we want to require that objects are always runtime.Object and metav1.Object

boekkooi-fresh:
Hey @vincepri, 
It happens I'm already very happy that I can use controller-runtime for my project. I managed to refactor my code to use the new `controllerutil.Object` and the code actually looks better for it which is nice.
I'm fine with closing this PR but maybe next time if at all possible a BC break can be done in a minor bump?

Thanks again for the nice library!

vincepri:
@boekkooi-fresh That's our goal! This was an oversight, apologies again for that.

/close

Issue #1073, opened 2020-07-24
Namespace enforcing wrapper for client.Client


alvaroaleman:
Currently when using the `client.Client`, all functions that `Get/Create/List/DeleteCollection` need to know the namespace when interacting with namespaced objects. It would be useful to have a wrapper for the `client.Client` that always sets the namespace so it doesn't need to be passed around everywhere.

Should probably be constructed like this:

```go
func NewNamespaceWrapper(upstream Client, namespace string) Client {
```

/kind feature
/help

alvaroaleman:
Fixed in #1088

PR #1088, opened 2020-07-30
:sparkles: Add namespace enforcing wrapper for client.Client

varshaprasad96:
This PR adds a namespace enforcing wrapper for client.Client.
This helps while dealing with namespace-scoped objects, where the
namespace value need not be specified in every operation.

This PR addresses the issue: #1073

### DIFF
--- a/pkg/client/namespaced_client.go
+++ b/pkg/client/namespaced_client.go
@@ -21,18 +21,18 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // NewNamespacedClient wraps an existing client enforcing the namespace value.
 // All functions using this client will have the same namespace declared here.
-func NewNamespacedClient(c Client, ns string, rmp meta.RESTMapper, sch *runtime.Scheme) Client {
+func NewNamespacedClient(c Client, ns string) Client {
 	return &namespacedClient{
-		client:     c,
-		namespace:  ns,
-		restmapper: rmp,
-		scheme:     *sch,
+		client:    c,
+		namespace: ns,
 	}
 }
 
@@ -40,39 +40,54 @@ var _ Client = &namespacedClient{}
 
 // namespacedClient is a Client that wraps another Client in order to enforce the specified namespace value.
 type namespacedClient struct {
-	namespace  string
-	client     Client
-	restmapper meta.RESTMapper
-	scheme     runtime.Scheme
+	namespace string
+	client    Client
 }
 
-func getNamespace(restmapper meta.RESTMapper, obj runtime.Object, sch *runtime.Scheme) (bool, error) {
-	// var sch = runtime.NewScheme()
-	// // appsv1.AddToScheme(sch)
-	// rbacv1.AddToScheme(sch)
-	gvk, err := apiutil.GVKForObject(obj, sch)
-	if err != nil {
-		return false, err
-	}
-	if restmapper == nil {
-		return false, err
+// Scheme returns the scheme this client is using.
+func (n *namespacedClient) Scheme() *runtime.Scheme {
+	return n.client.Scheme()
+}
+
+// RESTMapper returns the scheme this client is using.
+func (n *namespacedClient) RESTMapper() meta.RESTMapper {
+	return n.client.RESTMapper()
+}
+
+// isNamespaced returns true if the object is namespace scoped.
+// For unstructured objects the gvk is found from the object itself.
+func isNamespaced(c Client, obj runtime.Object) (bool, error) {
+	var gvk schema.GroupVersionKind
+	var err error
+
+	_, isUnstructured := obj.(*unstructured.Unstructured)
+	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
+
+	isUnstructured = isUnstructured || isUnstructuredList
+	if isUnstructured {
+		gvk = obj.GetObjectKind().GroupVersionKind()
+	} else {
+		gvk, err = apiutil.GVKForObject(obj, c.Scheme())
+		if err != nil {
+			return false, err
+		}
 	}
 
-	// gvk := schema.GroupKind{
-	// 	Group: obj.GetObjectKind().GroupVersionKind().Group,
-	// 	Kind:  obj.GetObjectKind().GroupVersionKind().Kind,
-	// }
-	restmapping, err := restmapper.RESTMapping(gvk.GroupKind())
+	gk := schema.GroupKind{
+		Group: gvk.Group,
+		Kind:  gvk.Kind,
+	}
+	restmapping, err := c.RESTMapper().RESTMapping(gk)
 	if err != nil {
-		return false, fmt.Errorf("error here restmapping %v", obj)
+		return false, fmt.Errorf("failed to get restmapping: %w", err)
 	}
 	scope := restmapping.Scope.Name()
 
 	if scope == "" {
-		return false, nil
+		return false, fmt.Errorf("Scope cannot be identified. Empty scope returned")
 	}
 
-	if scope != meta.RESTScopeNameNamespace {
+	if scope != meta.RESTScopeNameRoot {
 		return true, nil
 	}
 	return false, nil
@@ -85,9 +100,9 @@ func (n *namespacedClient) Create(ctx context.Context, obj runtime.Object, opts 
 		return err
 	}
 
-	isNamespaceScoped, err := getNamespace(n.restmapper, obj, &n.scheme)
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
 	if err != nil {
-		return fmt.Errorf("erroring Here, %v", err)
+		return fmt.Errorf("error finding the scope of the object %v", err)
 	}
 	if isNamespaceScoped {
 		metaObj.SetNamespace(n.namespace)
@@ -102,7 +117,12 @@ func (n *namespacedClient) Update(ctx context.Context, obj runtime.Object, opts 
 		return err
 	}
 
-	if n.namespace != "" && metaObj.GetNamespace() != n.namespace {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+
+	if isNamespaceScoped && metaObj.GetNamespace() != n.namespace {
 		metaObj.SetNamespace(n.namespace)
 	}
 	return n.client.Update(ctx, obj, opts...)
@@ -115,7 +135,12 @@ func (n *namespacedClient) Delete(ctx context.Context, obj runtime.Object, opts 
 		return err
 	}
 
-	if n.namespace != "" && metaObj.GetNamespace() != n.namespace {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+
+	if isNamespaceScoped && metaObj.GetNamespace() != n.namespace {
 		metaObj.SetNamespace(n.namespace)
 	}
 	return n.client.Delete(ctx, obj, opts...)
@@ -123,7 +148,12 @@ func (n *namespacedClient) Delete(ctx context.Context, obj runtime.Object, opts 
 
 // DeleteAllOf implements client.Client
 func (n *namespacedClient) DeleteAllOf(ctx context.Context, obj runtime.Object, opts ...DeleteAllOfOption) error {
-	if n.namespace != "" {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+
+	if isNamespaceScoped {
 		opts = append(opts, InNamespace(n.namespace))
 	}
 	return n.client.DeleteAllOf(ctx, obj, opts...)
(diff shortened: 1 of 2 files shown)


pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// NewNamespacedClient wraps an existing client enforcing the namespace value.
+// All functions using this client will have the same namespace declared here.
+func NewNamespacedClient(ns string, c Client) Client {
```
alvaroaleman, line 28: Nit: I think its a bit mor logical to make the `Client` the first argument and the `ns` the second

vincepri:
What about having a method on the main `Client` interface that returns the namespaced-client implementation underneath?

For example:

```
client.WithNamespace().Get()

or

client.ForNamespace().Get()

etc.

```

alvaroaleman:
@vincepri one drawback of that is that we break existing wrappers

vincepri:
That would be a breaking change yes

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Update implements client.Client
+func (n *namespacedClient) Update(ctx context.Context, obj runtime.Object, opts ...UpdateOption) error {
+	metaObj, err := meta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+
+	if n.namespace != "" && metaObj.GetNamespace() != n.namespace {
```
alvaroaleman, line 63: I think rather than trying to infer from the object if its namespaced by checking if any namespace is set, it would be more logical to take a [`RESTMapper`](pkg/client/apiutil/dynamicrestmapper_test.go) in the constructor of this wrapper and use its `RESTMapping` method to figure out if a given object is namespaced or not (You can use [`GVKForObject`](https://github.com/alvaroaleman/controller-runtime/blob/accbce34c7c0a807ad2beca5e7cf9672aa914a18/pkg/client/apiutil/apimachinery.go#L50) to get the gvk)

varshaprasad96: Thanks for the review @alvaroaleman. Had a doubt regarding this. `GVKForObject` requires `runtime.Scheme`, which is useful in `RestMapping` method too. Currently we don't have a method to obtain the `scheme` from client Options, when it is [generated](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client?tab=doc#New). If we go ahead with creating a new scheme in the implementation, we would have to register the object Kind manually using [`AddToScheme`](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime?tab=doc#SchemeBuilder.AddToScheme). Is there any another way to go about this ?

vincepri: Related #1058

alvaroaleman: I think for now the `constructor` needs to take both a `RESTMapper` and a `Scheme`. Ideally we could just obtain both of those from the Client, but that is not the case today. Once #1058 is in, we can at least obtain the Scheme from the Client.

varshaprasad96: Thanks @vincepri and @alvaroaleman . Will update the PR.

varshaprasad96: I see that in the implementation of [`client.New(..)`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client?tab=doc#New), `runtime.Scheme` will not be used for unstructured objects. Hence, even when we pass a `scheme` with the `client`, GVK is not registered - similar to test case present [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client_test.go#L525). If this is the case, then wouldn't `RESTMapping` not be useful for unstructured objects, since for them GVK is extracted from Object itself ?

vincepri: With unstructured objects, the user is required to set the GVK before passing in to a client, so you should be able to use `GetObjectKind()` on those.

vincepri: That said, exposing a RESTMapper from a Client is probably a good thing. We can follow the approach used in #1058 to work out something similar

vincepri:
Re-reviewed this PR, and I'd like to propose to go down the `client.ForNamespace()` (or similar) route, which is flexible in a lot of ways and transparent to users.

alvaroaleman:
> Re-reviewed this PR, and I'd like to propose to go down the client.ForNamespace() (or similar) route, which is flexible in a lot of ways and transparent to users.

Well, since we have at least three implementations (delegating, normal and fake) and how this works is identical for all of them, this needs to be implemented as a wrapper anyways. We can probably debate in a follow-up if we want an interface change.

pkg/client/namespaced_client.go
vincepri: We've merged https://github.com/kubernetes-sigs/controller-runtime/pull/1109/files, which brings in support to get a RESTMapper from a client

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,232 @@
...
+// RESTMapper returns the scheme this client is using.
+func (n *namespacedClient) RESTMapper() meta.RESTMapper {
+	return n.client.RESTMapper()
+}
+
+// isNamespaced returns true if the object is namespace scoped.
+// For unstructured objects the gvk is found from the object itself.
+func isNamespaced(c Client, obj runtime.Object) (bool, error) {
+	gvk, err := apiutil.GVKForObject(obj, c.Scheme())
+	if err != nil || gvk.GroupKind().Empty() {
```
alvaroaleman, line 60: rather than just tolerating any error here, can we extend GVKForObject to cope with `unstructured` (by doing an assertion if the obj is an `*unstructured.Unstructed` or `unstructured.UnstructuredList) and if thats the case, get the GVK from the object itself?

Just ignoring any errors is generally never good and the logic we built here belongs into `GVKForObject` IMHO

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,232 @@
...
+		gvk = obj.GetObjectKind().GroupVersionKind()
+	}
+
+	gk := schema.GroupKind{
+		Group: gvk.Group,
+		Kind:  gvk.Kind,
+	}
+	restmapping, err := c.RESTMapper().RESTMapping(gk)
+	if err != nil {
+		return false, fmt.Errorf("error here restmapping %v", err)
```
alvaroaleman, line 70: `failed to get restmapping: %w` would IMHO be a more appropriate error

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	scope := restmapping.Scope.Name()
+
+	if scope != "" && scope != meta.RESTScopeNameRoot {
```
alvaroaleman, line 74: Shouldn't we return an error if the `scope` is an empty string rather than assuming it means `namespaced`?

pkg/client/namespaced_client.go
varshaprasad96: Agreed. Sorry, had missed this.

coderanger:
It seems a little weird to silently overwrite a namespace on an object. I could see defaulting it (set only if == ""), or erroring if set to the wrong thing, but this flavor seems more likely to create bugs by redirecting calls to the wrong object.

alvaroaleman:
> It seems a little weird to silently overwrite a namespace on an object. I could see defaulting it (set only if == ""), or erroring if set to the wrong thing, but this flavor seems more likely to create bugs by redirecting calls to the wrong object.

Yeah agree on that, if its set to a different NS we should error out instead of overwriting it

varshaprasad96:
Addressed review comments.

coderanger:
This now checks if the namespace matches the client twice, basically two strcmps where I think you only need one?

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	if isNamespaceScoped {
+		metaObj.SetNamespace(n.namespace)
+	}
+	return n.client.Create(ctx, obj, opts...)
+}
+
+// Update implements client.Client
+func (n *namespacedClient) Update(ctx context.Context, obj Object, opts ...UpdateOption) error {
+	metaObj, err := meta.Accessor(obj)
```
alvaroaleman, line 115: Nit: We don't need all the `meta.Accessor` calls anymore, as `Object` is a metav1.Object already

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,273 @@
...
+		Kind:  gvk.Kind,
+	}
+	restmapping, err := c.RESTMapper().RESTMapping(gk)
+	if err != nil {
+		return false, fmt.Errorf("failed to get restmapping: %w", err)
+	}
+	scope := restmapping.Scope.Name()
+
+	if scope == "" {
+		return false, fmt.Errorf("Scope cannot be identified. Empty scope returned")
```
alvaroaleman, line 87: Nit: Use `errors.New` when there is no format string

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,273 @@
...
+	metaObj, err := meta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+	if isNamespaceScoped {
```
alvaroaleman, line 107: This lacks a check if the object has a pre-existing namespace

alvaroaleman (commented):
> This now checks if the namespace matches the client twice, basically two strcmps where I think you only need one?

@coderanger What exactly do you mean by that, could you comment on the code where you see this?

coderanger:
> > This now checks if the namespace matches the client twice, basically two strcmps where I think you only need one?
> 
> @coderanger What exactly do you mean by that, could you comment on the code where you see this?

I think I understand the flow now, it looks like if the object has `Namespace == ""` it will silently apply the requested namespace. If the object has any other value (other than the bound namespace itself) it will error.

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,250 @@
...
+	return n.client.Patch(ctx, obj, patch, opts...)
+}
+
+// Get implements client.Client
+func (n *namespacedClient) Get(ctx context.Context, key ObjectKey, obj Object) error {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+	if isNamespaceScoped {
```
alvaroaleman, line 188: This also needs to error if the `namespace` in the key doesn't match the namespace of the `client`

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,250 @@
...
+// Patch implements client.Client
+func (n *namespacedClient) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
+	}
+
+	objectNamespace := obj.GetNamespace()
+	if objectNamespace != n.namespace && objectNamespace != "" {
+		return fmt.Errorf("Namespace %s of the object %s does not match %s", objectNamespace, obj.GetName(), n.namespace)
```
alvaroaleman, line 173: Please clarify the issue here (and in all the other identical errors), something like `namespace %s on object does not match the namespace %s on the client`

pkg/client/namespaced_client.go
```diff
@@ -0,0 +1,250 @@
...
+		opts = append(opts, InNamespace(n.namespace))
+	}
+	return n.client.DeleteAllOf(ctx, obj, opts...)
+}
+
+// Patch implements client.Client
+func (n *namespacedClient) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error {
+	isNamespaceScoped, err := isNamespaced(n.client, obj)
+	if err != nil {
+		return fmt.Errorf("error finding the scope of the object %v", err)
```
alvaroaleman, line 168: please delimit the error with a colon: `error finding the scope of the object: %v`

pkg/client/namespaced_client_test.go
```diff
@@ -0,0 +1,597 @@
...
+		})
+
+		It("should not create object if the namespace of the object is different", func() {
+			By("creating the object initially")
+			dep.SetNamespace("non-default")
+			err := getClient().Create(ctx, dep)
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should create an object in the namespace specified with the client", func() {
```
alvaroaleman, line 161: This test is identical to `It("should successfully create object in the right namespace"` or not?

pkg/client/namespaced_client_test.go
```diff
@@ -0,0 +1,597 @@
...
+
+			// Delete the clusterRole Resource
+			deleteClusterRole(ctx, cr)
+		})
+	})
+
+	Describe("Update", func() {
+		var err error
+		BeforeEach(func() {
+			dep.Annotations = map[string]string{"foo": "bar"}
```
alvaroaleman, line 203: I think you meant to add that after the creation, all the `Update` calls in this test are no-ops

pkg/client/namespaced_client_test.go
```diff
@@ -0,0 +1,597 @@
...
+			By("validating patched Deployment has new annotations")
+			actual, err := clientset.AppsV1().Deployments(ns).Get(ctx, dep.Name, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(actual.Annotations["foo"]).To(Equal("bar"))
+			Expect(actual.GetNamespace()).To(Equal(ns))
+		})
+
+		It("should not modify the object when namespace of the object is different", func() {
+			dep.SetNamespace("non-default")
+			By("Applying Patch")
```
alvaroaleman, line 354: I think the `By` only makes sense when you have more than one of it to split your test in logical parts

alvaroaleman (commented):
Some comments, after that it should be good to go. Thanks for your work on this!

alvaroaleman (approved):
/retitle :sparkles: Add namespace enforcing wrapper for client.Client
/lgtm
/approve

Thanks for your work!

PR #1089, opened 2020-07-31
Try to avoid event handling leaks

### LINKED ISSUE #637: Clean up event handling code goroutine leaks
Technically, we can leak goroutines when you create a manager, since the event initialization code used in CR creates new goroutines.

There's two places -- the first is when we create a new broadcaster, it launches some goroutines.  As of https://github.com/kubernetes/kubernetes/pull/83489, that can be shut down, but we probably need to move that to start.

The second is in the actual event provider setup code -- that launches some goroutines when we set up the sinks.  If we grab the results of the setup functions, we can call their stop methods to clean them up.  Those should probably also be moved to start.

Actually unwinding this will probably require some changes to the event setup flow, though, so if anyone comes across this, it might be tricky.

Not the highest priority since it only occurs if you repeated create managers -- the normal "1 manager per process" model doesn't actually care about this issue.

/kind bug
/priority important-longterm

DirectXMan12:
Since we now have the ability shut down the event broadcaster, we can
write mostly goroutine-leak-free event handling setup.  This changes the
default event handling setup to defer the broadcaster initialization the
first time it's used, and then to shut it down once the manager shuts
down.

In the case where a broadcaster is manually specified, it's the
specifier's job to shut down the broadcaster instead.

We'll probably still want to overhaul the whole event system at some
point in the future though.

This also re-enables the tests for leaks, switching them to an
eventually to avoid flakes & reducing the threshold to zero.

Closes #637

/kind bug

### DIFF
--- a/pkg/internal/recorder/recorder.go
+++ b/pkg/internal/recorder/recorder.go
@@ -17,7 +17,9 @@ limitations under the License.
 package recorder
 
 import (
+	"context"
 	"fmt"
+	"sync"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


DirectXMan12:
/assign @vincepri 

@vincepri IIRC, you've got context on the flakes on CI, so I'd appreciate a look from you on this one.

pkg/internal/recorder/recorder.go
```diff
@@ -26,35 +28,129 @@ import (
...
+func (p *Provider) getBroadcaster() record.EventBroadcaster {
+	// NB(directxman12): this can technically still leak if something calls
+	// "getBroadcaster" (i.e. Emits an Event) but never calls Start, but if we
+	// create the broadcaster in start, we could race with other things that
+	// are started at the same time & want to emit events.  The alternative is
+	// silently swallowing events and more locking, but that seems suboptimal.
+
+	p.broadcasterOnce.Do(func() {
+		broadcaster, stop := p.makeBroadcaster()
+		broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: p.evtClient})
```
mindprince, line 95: Does the `watch.Interface` returned by `StartRecordingToSink` and `StartEventWatcher` need to be `Stop()`ed or is it taken care of by the above `broadcaster.Shutdown()`?

DirectXMan12: ah, good catch, let me double-check

DirectXMan12: I *think* the broadcaster handles it: that watcher is returned by apimachinery's watch.Broadcaster, which doesn't actually start a goroutine -- it just holds a channel.  `Shutdown` on the broadcaster calls the underlying `watch.Broadcaster`'s `Shutdown`, which closes `Broadcaster.incoming`, which breaks the loop in `Broadcaster.loop`, which calls `m.closeAll` after the loop is done, which calls `close` on all those open channels.

mindprince: As long as https://github.com/kubernetes/kubernetes/blob/0051d65f9f30db724dfb88256f70c23e37f6b257/staging/src/k8s.io/client-go/tools/record/event.go#L299 exits when `brodcaster.Shutdown()` is called. I am fine.

vincepri:
@DirectXMan12 Seems we're still getting some flakes

DirectXMan12:
Ack, I'll add some reporting in, see if we can figure out why, and then turn that test back off if it's not obvious.

/hold

DirectXMan12:
:crossed_fingers: this should fix that flake, or at the very least give us better info.  I *think* the flake may have been due to keep-alive in the http package, which the `goleak` library showed a bit more clearly.

DirectXMan12:
gonna run this a couple more times to check for flakes

DirectXMan12:

I think this looks good.  Can't repro the flake locally any more, can't repro here, assuming this last run is fine

mindprince:
Can we merge this?

DirectXMan12:
@vincepri can you re-lgtm this?

vincepri:
Seems we have a conflict, need to rebase

DirectXMan12:
rebased

PR #1093, opened 2020-08-03
✨ Add ability to set/get logger in/from a context.Context

vincepri:

This we might be able to merge now, given that's not a breaking change

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -34,7 +34,13 @@ limitations under the License.
 package log
 
 import (
+	"context"
+
 	"github.com/go-logr/logr"
+)
+
+var (
+	contextKey = &struct{}{}
 )
 
 // SetLogger sets a concrete logging implementation for all deferred Loggers.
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -34,9 +34,15 @@ limitations under the License.
 package log
 
 import (
+	"context"
+
 	"github.com/go-logr/logr"
 )
 
+var (
+	contextKey = &struct{}{}
```
alvaroaleman, line 43: err...Is this common, can we just claim a key like this?

vincepri: Yes, this is the approach pointed out in https://github.com/kubernetes-sigs/controller-runtime/pull/1054#discussion_r457908599

It's actually a pretty nice approach in general, it won't use a string which might cause issues and the pointer is used to determine equality, which users can't overwrite

alvaroaleman: Do we need the equality? I have to admit I am not an expert on how a context is supposed to be used but since it ends up being shared among code owned by different parties I intuitively thought it would make sense to somehow scope context keys to the project (e.G. via a string prefix) to avoid collisions that will result in unexpected behavior

vincepri: We can always change it later, unless we don't want to merge this for some reason, I do like that for now this is private and not allowed to be overwritten, I'd rather have this go in given that it has been proven in another project to work well and iterate on it as we go

detiber: From that comment thread, there are at least 2 different active projects linked that use the same approach (cert-manager and go-grpc-middleware), so +1 to this approach

detiber:
lgtm, but will defer to @alvaroaleman for tagging based on the discussion thread.

pkg/log/log.go
alvaroaleman: I actually didn't realize the context key is the pointer address and was concerned this might conflict with someone else also using an empty struct as identifier.
/lgtm

piepmatz: I think this breaks when you have multiple of such context keys. They are likely to be identical as they can have the same address: https://play.golang.org/p/4uqlsUv5Xuf

The Go ref (https://golang.org/ref/spec#Size_and_alignment_guarantees) says: 
> Two distinct zero-size variables may have the same address in memory.

PR #1095, opened 2020-08-03
:sparkles: Add options to configure a logger for manager and controllers

vincepri:

Made these changes also backward compatible to be able to go in v0.6.2

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -45,6 +47,7 @@ type Builder struct {
 	config           *rest.Config
 	ctrl             controller.Controller
 	ctrlOptions      controller.Options
+	log              logr.Logger
 	name             string
 }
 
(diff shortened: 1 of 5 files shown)


pkg/builder/controller.go
```diff
@@ -228,26 +237,33 @@ func (blder *Builder) loadRestConfig() {
...
+	gvk, err := getGvk(blder.forInput.object, blder.mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+
+	// Setup the logger.
+	if ctrlOptions.Log == nil {
+		ctrlOptions.Log = blder.mgr.GetLogger()
+	}
+	ctrlOptions.Log = ctrlOptions.Log.WithValues("groupVersionKind", gvk.String())
```
alvaroaleman, line 264: How do we properly distinguish this from what we do in the encoder https://github.com/kubernetes-sigs/controller-runtime/blob/6633846423ba4853b1c4466f28aa6b66d6804930/pkg/log/zap/kube_helpers.go#L69 ?

I think the log key should maybe be `reoncilerGroupVersionKind` or something like that?

pkg/internal/controller/controller.go
```diff
@@ -256,7 +258,7 @@ func (c *Controller) reconcileHandler(obj interface{}) bool {
 	c.Queue.Forget(obj)
 
 	// TODO(directxman12): What does 1 mean?  Do we want level constants?  Do we want levels at all?
-	c.Log.V(1).Info("Successfully Reconciled", "name", req.Name, "namespace", req.Namespace)
+	log.V(1).Info("Successfully Reconciled", "name", req.Name, "namespace", req.Namespace)
```
detiber, line 261:
```suggestion
	log.V(1).Info("Successfully Reconciled")
```

pkg/builder/controller.go
detiber: It seems like the logging here should remove the need to enrich the log entries in the zap helper. Would you expect any delta between the gvk defined here and the resource passed to reconcile?

alvaroaleman: Its totally possible that during the reconciliation of an object type a an object type b has to be updated and that fails so I think the two are complementary.

vincepri: Fixed, used `reconcilerGroup` and `reconcilerKind`

PR #1092, opened 2020-08-03
🏃 setup-envtest.sh: standalone script for setting up envtest

joelanford:
`envtest` is a test framework provided by `controller-runtime` that simplifies running integration tests for controllers.

Currently, it is not straightforward to setup envtest without also setting up kubebuilder. However, Kubebuilder is not a prerequisite to use `envtest`.

This PR splits out the envtest setup code from controller-runtime's CI scripts such that users could easily download and use the envtest setup code in their own CI scripts.

It would work like this:
```console
source <(curl -sfL https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh)

envtest_assets_dir=$(pwd)/test/assets
fetch_envtest_tools "$envtest_assets_dir"
setup_envtest_env "$envtest_assets_dir"
```

PR #1097, opened 2020-08-03
🏃 hack/setup-envtest.sh: follow-up from #1092

joelanford:
This follows up #1092 to address some of the comments and suggestions.

### DIFF
--- a/hack/setup-envtest.sh
+++ b/hack/setup-envtest.sh
@@ -14,32 +14,60 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-set -e
+set -o errexit
+set -o pipefail
+
+# Turn colors in this script off by setting the NO_COLOR variable in your
+# environment to any value:
+#
+# $ NO_COLOR=1 test.sh
+NO_COLOR=${NO_COLOR:-""}
+if [ -z "$NO_COLOR" ]; then
+  header=$'\e[1;33m'
+  reset=$'\e[0m'
+else
+  header=''
+  reset=''
+fi
+
+function header_text {
+  echo "$header$*$reset"
+}
 
 function setup_envtest_env {
+  header_text "setting up env vars"
+
   # Setup env vars
+  KUBEBUILDER_ASSETS=${KUBEBUILDER_ASSETS:-""}
   if [[ -z "${KUBEBUILDER_ASSETS}" ]]; then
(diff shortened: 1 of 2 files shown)


joelanford:
/cc @camilamacedo86 @vincepri @alvaroaleman @DirectXMan12 

@vincepri Just saw your follow-up about copying the header text function. I'll update this PR now.

vincepri:
Thanks for following up! The changes look great, will tag it once the header text is in

joelanford:
@vincepri WDYT about removing `set -o nounset` from `setup-envtest.sh`?

I was testing it out by running `source ./hack/setup-envtest.sh; echo $KUBEBUILDER_ASSETS` and expecting to see it return an empty string before calling `setup_envtest_env`.

Instead, my terminal exited because `$KUBEBUILDER_ASSETS` was not set yet.

We could add this to `hack/check-everything.sh` since we don't expect people to be `source`-ing that script.

vincepri:
Got it, I don't have time to try today, but I didn't get the same error in a similar cluster api script

hack/setup-envtest.sh
```diff
@@ -14,32 +14,60 @@
...
+
+function header_text {
+  echo "$header$*$reset"
+}
 
 function setup_envtest_env {
+  header_text "setting up env vars"
+
   # Setup env vars
+  KUBEBUILDER_ASSETS=${KUBEBUILDER_ASSETS:-""}
```
vincepri, line 41: Can probably remove this?

hack/setup-envtest.sh
```diff
@@ -14,32 +14,60 @@
...
+# Skip fetching and untaring the tools by setting the SKIP_FETCH_TOOLS variable
+# in your environment to any value:
+#
+# $ SKIP_FETCH_TOOLS=1 ./check-everything.sh
+#
+# If you skip fetching tools, this script will use the tools already on your
+# machine.
 function fetch_envtest_tools {
+  SKIP_FETCH_TOOLS=${SKIP_FETCH_TOOLS:-""}
+  if [ -n "$SKIP_FETCH_TOOLS" ]; then
```
vincepri, line 58: Can probably remove this?

hack/setup-envtest.sh
joelanford: Actually had to keep these if we want to use `set -o nounset` in `./hack/check-everything.sh`

PR #1099, opened 2020-08-03
Update to Kuberentes v1.17.9

vincepri:


vincepri:
@alvaroaleman Need another look, I had to update the go.sum file

PR #1100, opened 2020-08-04
:sparkles: Make individual readiness and liveness checks accessible

jimmidyson:
When a named readiness/liveness check is registered it should be available under /readyz/<name>.
This PR enables this by adding a trailing slash to the default readyz and healthz endpoints, configuring
the server to handle existing path (`/readyz`) and subpaths (`/readyz/my-check`).


jimmidyson:
@alenkacz That's right. Also, someone could add an endpoint without trailing slash that would override the root path. It's a bit funky but I think this is right default behaviour.

PR #1101, opened 2020-08-04
:seedling:  Un-deprecate NewFakeClient

vincepri:

This PR removes the long deprecated function from the fake client, and replaces it with the one that requires a scheme

/milestone v0.7.x

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -58,8 +58,6 @@ const (
 
 // NewFakeClient creates a new fake client for testing.
 // You can choose to initialize it with a slice of runtime.Object.
-// Deprecated: use NewFakeClientWithScheme.  You should always be
-// passing an explicit Scheme.
 func NewFakeClient(initObjs ...runtime.Object) client.Client {
 	return NewFakeClientWithScheme(scheme.Scheme, initObjs...)
 }


pkg/client/fake/client.go
```diff
@@ -56,18 +56,11 @@ const (
...
-	return NewFakeClientWithScheme(scheme.Scheme, initObjs...)
-}
-
-// NewFakeClientWithScheme creates a new fake client with the given scheme
+// NewFakeClient creates a new fake client with the given scheme
 // for testing.
+//
 // You can choose to initialize it with a slice of runtime.Object.
-func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...runtime.Object) client.Client {
+func NewFakeClient(clientScheme *runtime.Scheme, initObjs ...runtime.Object) client.Client {
```
vincepri, line 63: Should we remove `initObjs` support as well here? It creates object without a resourceVersion which is technically incorrect

alvaroaleman:
I don't really understand why we need to enforce passing in a scheme. IMHO, using the kubernetes one by default and allowing an override is totally reasonable.

alvaroaleman:
That is probably more of a question for @DirectXMan12 as he was driving the whole "enforce scheme everywhere"

pkg/client/fake/client.go
alvaroaleman: I would like to keep it, its super convenient. If anything, we can extend this to set the resourceVersion

vincepri:
It might be error-prone, folks can always pass scheme.Scheme explicitly this way

pkg/client/fake/client.go
vincepri: Yeah, there was another PR out to do that, although that breaks lots of tests

alvaroaleman:
> It might be error-prone, folks can always pass scheme.Scheme explicitly this way

When is this error prone? I've literally never used the `WithScheme` and the only thing I will do different if we merge this (and I presume most ppl that use the fake client) is construct a downstream helper that passes in the kubernetes scheme as the first arg. Can we maybe discuss this at the next meeting to clarify what risks exactly this is preventing, because thats not clear to me?

vincepri:
We have a lot of tests using custom schemes (in cluster api), because that's how they were setup. I'm not opposed to say that folks should use `scheme.Scheme`, although might be good to keep both and remove the deprecation notice if we'd like to keep the other constructor

alvaroaleman:
I want to discuss this at the next meeting. IMHO we should undeprecate it. It makes total sense to allow passing a Scheme but it does not make sense to me to enforce this. I've a long history of building controllers and I've never used anything other than the default client-go scheme. To me its not clear what the problem with that is.

alvaroaleman:
We discussed this in the meeting on Aug 27th and decided that we can undeprecate it. The main reason for deprecating it was that ppl used it with CRDs without adding them to their scheme, resulting in errors. We can instead better handle this error and return an error that clarifies what needs to be done to add the types to the scheme.

vincepri:
Gonna rework the PR later this week

vincepri:
@alvaroaleman Reworked to un-deprecate

alvaroaleman (approved):
/lgtm
/hold cancel
Thanks!

PR #1116, opened 2020-08-05
⚠️ Propagate context.Context throughout the codebase

vincepri:

/assign @detiber @alvaroaleman 
/milestone v0.7.x

If anyone has better ideas on how to propagate this context, please let me know 😄

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -83,16 +83,17 @@ var _ = Describe("Multi-Namespace Informer Cache", func() {
 func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (cache.Cache, error)) {
 	Describe("Cache test", func() {
 		var (
-			informerCache cache.Cache
-			stop          chan struct{}
-			knownPod1     runtime.Object
-			knownPod2     runtime.Object
-			knownPod3     runtime.Object
-			knownPod4     runtime.Object
+			informerCache       cache.Cache
+			informerCacheCtx    context.Context
+			informerCacheCancel context.CancelFunc
+			knownPod1           runtime.Object
+			knownPod2           runtime.Object
+			knownPod3           runtime.Object
+			knownPod4           runtime.Object
 		)
 
 		BeforeEach(func() {
-			stop = make(chan struct{})
+			informerCacheCtx, informerCacheCancel = context.WithCancel(context.Background())
 			Expect(cfg).NotTo(BeNil())
 
 			By("creating three pods")
@@ -123,11 +124,11 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			Expect(err).NotTo(HaveOccurred())
 			By("running the cache and waiting for it to sync")
 			// pass as an arg so that we don't race between close and re-assign
-			go func(stopCh chan struct{}) {
+			go func(ctx context.Context) {
 				defer GinkgoRecover()
-				Expect(informerCache.Start(stopCh)).To(Succeed())
-			}(stop)
-			Expect(informerCache.WaitForCacheSync(stop)).To(BeTrue())
+				Expect(informerCache.Start(ctx)).To(Succeed())
+			}(informerCacheCtx)
+			Expect(informerCache.WaitForCacheSync(informerCacheCtx)).To(BeTrue())
 		})
 
 		AfterEach(func() {
@@ -137,7 +138,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			deletePod(knownPod3)
 			deletePod(knownPod4)
 
-			close(stop)
+			informerCacheCancel()
 		})
 
 		Describe("as a Reader", func() {
@@ -270,13 +271,12 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 				})
 
 				It("should return an error when context is cancelled", func() {
-					By("creating a context and cancelling it")
-					ctx, cancel := context.WithCancel(context.Background())
-					cancel()
+					By("cancelling the context")
+					informerCacheCancel()
 
 					By("listing pods in test-namespace-1 with a cancelled context")
 					listObj := &kcorev1.PodList{}
-					err := informerCache.List(ctx, listObj, client.InNamespace(testNamespaceOne))
+					err := informerCache.List(informerCacheCtx, listObj, client.InNamespace(testNamespaceOne))
 
 					By("verifying that an error is returned")
 					Expect(err).To(HaveOccurred())
@@ -396,9 +396,9 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					By("running the cache and waiting for it to sync")
 					go func() {
 						defer GinkgoRecover()
-						Expect(namespacedCache.Start(stop)).To(Succeed())
+						Expect(namespacedCache.Start(informerCacheCtx)).To(Succeed())
 					}()
-					Expect(namespacedCache.WaitForCacheSync(stop)).NotTo(BeFalse())
+					Expect(namespacedCache.WaitForCacheSync(informerCacheCtx)).NotTo(BeFalse())
 
 					By("listing pods in all namespaces")
 					out := &unstructured.UnstructuredList{}
@@ -574,9 +574,9 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					By("running the cache and waiting for it to sync")
 					go func() {
 						defer GinkgoRecover()
-						Expect(informer.Start(stop)).To(Succeed())
+						Expect(informer.Start(informerCacheCtx)).To(Succeed())
 					}()
-					Expect(informer.WaitForCacheSync(stop)).NotTo(BeFalse())
+					Expect(informer.WaitForCacheSync(informerCacheCtx)).NotTo(BeFalse())
 
 					By("listing Pods with restartPolicyOnFailure")
 					listObj := &kcorev1.PodList{}
@@ -591,8 +591,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 
 				It("should allow for get informer to be cancelled", func() {
 					By("creating a context and cancelling it")
-					ctx, cancel := context.WithCancel(context.Background())
-					cancel()
+					informerCacheCancel()
 
 					By("getting a shared index informer for a pod with a cancelled context")
 					pod := &kcorev1.Pod{
@@ -609,7 +608,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 							},
 						},
 					}
-					sii, err := informerCache.GetInformer(ctx, pod)
+					sii, err := informerCache.GetInformer(informerCacheCtx, pod)
 					Expect(err).To(HaveOccurred())
 					Expect(sii).To(BeNil())
 					Expect(errors.IsTimeout(err)).To(BeTrue())
@@ -617,12 +616,11 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 
 				It("should allow getting an informer by group/version/kind to be cancelled", func() {
 					By("creating a context and cancelling it")
-					ctx, cancel := context.WithCancel(context.Background())
-					cancel()
+					informerCacheCancel()
 
 					By("getting an shared index informer for gvk = core/v1/pod with a cancelled context")
 					gvk := schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"}
-					sii, err := informerCache.GetInformerForKind(ctx, gvk)
+					sii, err := informerCache.GetInformerForKind(informerCacheCtx, gvk)
 					Expect(err).To(HaveOccurred())
 					Expect(sii).To(BeNil())
 					Expect(errors.IsTimeout(err)).To(BeTrue())
@@ -636,7 +634,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 						Object: map[string]interface{}{
 							"spec": map[string]interface{}{
 								"containers": []map[string]interface{}{
-									map[string]interface{}{
+									{
 										"name":  "nginx",
 										"image": "nginx",
 									},
@@ -702,9 +700,9 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					By("running the cache and waiting for it to sync")
 					go func() {
 						defer GinkgoRecover()
-						Expect(informer.Start(stop)).To(Succeed())
+						Expect(informer.Start(informerCacheCtx)).To(Succeed())
 					}()
-					Expect(informer.WaitForCacheSync(stop)).NotTo(BeFalse())
+					Expect(informer.WaitForCacheSync(informerCacheCtx)).NotTo(BeFalse())
 
 					By("listing Pods with restartPolicyOnFailure")
 					listObj := &unstructured.UnstructuredList{}
@@ -725,9 +723,8 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 				}, 3)
 
 				It("should allow for get informer to be cancelled", func() {
-					By("creating a context and cancelling it")
-					ctx, cancel := context.WithCancel(context.Background())
-					cancel()
+					By("cancelling the context")
+					informerCacheCancel()
 
 					By("getting a shared index informer for a pod with a cancelled context")
 					pod := &unstructured.Unstructured{}
@@ -738,7 +735,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 						Version: "v1",
 						Kind:    "Pod",
 					})
-					sii, err := informerCache.GetInformer(ctx, pod)
+					sii, err := informerCache.GetInformer(informerCacheCtx, pod)
 					Expect(err).To(HaveOccurred())
 					Expect(sii).To(BeNil())
 					Expect(errors.IsTimeout(err)).To(BeTrue())
--- a/pkg/cache/informertest/fake_cache.go
+++ b/pkg/cache/informertest/fake_cache.go
@@ -80,7 +80,7 @@ func (c *FakeInformers) GetInformer(ctx context.Context, obj runtime.Object) (ca
 }
 
 // WaitForCacheSync implements Informers
-func (c *FakeInformers) WaitForCacheSync(stop <-chan struct{}) bool {
+func (c *FakeInformers) WaitForCacheSync(ctx context.Context) bool {
 	if c.Synced == nil {
 		return true
 	}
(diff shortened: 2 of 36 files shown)


pkg/internal/controller/controller.go
```diff
@@ -130,6 +130,9 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 	c.Queue = c.MakeQueue()
 	defer c.Queue.ShutDown() // needs to be outside the iife so that we shutdown after the stop channel is closed
 
+	ctx, cancel := context.WithCancel(context.Background())
```
alvaroaleman, line 133: I think `TODO` is better suited since we want to eventually change the `Runnable` signature to take a context and then use that context here

pkg/internal/controller/controller.go
```diff
@@ -170,8 +174,12 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 		// Launch workers to process resources
 		c.Log.Info("Starting workers", "worker count", c.MaxConcurrentReconciles)
 		for i := 0; i < c.MaxConcurrentReconciles; i++ {
-			// Process work items
-			go wait.Until(c.worker, c.JitterPeriod, stop)
+			// Run a worker thread that just dequeues items, processes them, and marks them done.
+			// It enforces that the reconcileHandler is never invoked concurrently with the same object.
+			go wait.Until(func() {
```
detiber, line 179: This can likely use `wait.UntilWithContext()` now

vincepri: Done!

detiber: :grimacing: maybe that will need to wait until after context is plumbed through Runnable, looks like test is not happy with the ordering of stopping leader election compared to the workers because of this.

pkg/manager/internal.go
```diff
@@ -124,16 +124,6 @@ type controllerManager struct {
...
 
-	// internalStop is the stop channel *actually* used by everything involved
-	// with the manager as a stop channel, so that we can pass a stop channel
-	// to things that need it off the bat (like the Channel source).  It can
-	// be closed via `internalStopper` (by being the same underlying channel).
-	internalStop <-chan struct{}
-
-	// internalStopper is the write side of the internal stop channel, allowing us to close it.
-	// It and `internalStop` should point to the same channel.
-	internalStopper chan<- struct{}
```
vincepri, line 135: These have been removed in favor of `internalCtx/internalCancel`

pkg/manager/internal.go
```diff
@@ -190,6 +180,9 @@ type controllerManager struct {
 	// after the gracefulShutdownTimeout ended. It must not be accessed before internalStop
 	// is closed because it will be nil.
 	shutdownCtx context.Context
+
+	internalCtx    context.Context
+	internalCancel context.CancelFunc
```
vincepri, line 185: Going to add some documentation on what these are used for internally once we agree this is what we want / have some feedback

pkg/manager/internal.go
```diff
@@ -531,14 +525,12 @@ func (cm *controllerManager) engageStopProcedure(stopComplete chan struct{}) err
...
 	defer cm.mu.Unlock()
 	cm.stopProcedureEngaged = true
-	return cm.waitForRunnableToEnd(cm.shutdownCtx, cancel)
+	return cm.waitForRunnableToEnd(shutdownCancel)
 }
 
 // waitForRunnableToEnd blocks until all runnables ended or the
 // tearDownTimeout was reached. In the latter case, an error is returned.
-func (cm *controllerManager) waitForRunnableToEnd(ctx context.Context, cancel context.CancelFunc) error {
-	defer cancel()
```
vincepri, line 540: The caller already defers this cancel, removed it from here

pkg/internal/controller/controller.go
```diff
@@ -115,14 +115,14 @@ func (c *Controller) Watch(src source.Source, evthdler handler.EventHandler, prc
 	c.watches = append(c.watches, watchDescription{src: src, handler: evthdler, predicates: prct})
 	if c.Started {
 		c.Log.Info("Starting EventSource", "source", src)
-		return src.Start(evthdler, c.Queue, prct...)
+		return src.Start(context.TODO(), evthdler, c.Queue, prct...)
```
vincepri, line 118: Should we modify `Watch` to also accept a context, or save it as part of `Start` and pass it in here?

vincepri: Propagated the context with the one passed into Start, given that seems the most consistent approach

pkg/manager/internal.go
vincepri: `Add` is where I hit a blocker, if we pass a context in there, we might not be able to cancel all running routines depending on which context is passed in. 

I had to go with an internal context and force it on every `Add` call to make sure that we cancel that context before exiting.

The other issue I found with having a tree of context structs is that when you do `WithCancel`, the cancel is propagated and we don't really want that for this use case.

I'm happy to iterate on it in multiple small steps, this was the one that was the least confusing 😄

pkg/controller/controller_integration_test.go
```diff
@@ -84,9 +79,11 @@ var _ = Describe("controller", func() {
 			Expect(err).To(Equal(&cache.ErrCacheNotStarted{}))
 
 			By("Starting the Manager")
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
```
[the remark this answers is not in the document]
vincepri: Removed the one on line 51, it was a leftover.

pkg/internal/controller/controller.go
```diff
@@ -71,7 +71,8 @@ type Controller struct {
 	// Started is true if the Controller has been Started
 	Started bool
 
-	// TODO(community): Consider initializing a logger with the Controller Name as the tag
+	// ctx is the context that was passed to Start() and used when starting watches.
+	ctx context.Context
```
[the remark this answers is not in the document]
vincepri: Not that I could think about, the code today isn't really structured to use contexts how they're supposed to work, and I didn't want to break too much of the APIs yet, until we get more familiar with the internals.

`Watch()` is called both internally and externally, if we change the signature to pass a context as first parameter, we get into a situation where some watches might be using a context and some other are using another, which isn't ideal

pkg/manager/signals/signal.go
```diff
@@ -26,18 +27,19 @@ var onlyOneSignalHandler = make(chan struct{})
 // SetupSignalHandler registers for SIGTERM and SIGINT. A stop channel is returned
 // which is closed on one of these signals. If a second signal is caught, the program
 // is terminated with exit code 1.
-func SetupSignalHandler() (stopCh <-chan struct{}) {
+func SetupSignalHandler() (ctx context.Context) {
```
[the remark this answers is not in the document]
vincepri: We pass this directly to `Start` in multiple places, which now accepts a context

pkg/manager/internal.go
alvaroaleman: Yeah I think this won't be possible without storing it, unless we want to create a new context on every `Add` and then store a slice of cancel funcs that are all called on stop, which seems a lot worse

pkg/manager/internal.go
```diff
@@ -243,9 +236,6 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	if _, err := inject.InjectorInto(cm.SetFields, i); err != nil {
 		return err
 	}
-	if _, err := inject.StopChannelInto(cm.internalStop, i); err != nil {
```
alvaroaleman, line 246: Shoulnd't this get replaced with an `inject.ContextInto`?

pkg/manager/internal.go
```diff
@@ -654,11 +646,11 @@ func (cm *controllerManager) Elected() <-chan struct{} {
 	return cm.elected
 }
 
-func (cm *controllerManager) startRunnable(r Runnable) {
+func (cm *controllerManager) startRunnable(ctx context.Context, r Runnable) {
```
alvaroaleman, line 649: If I didn't miss anything the `ctx` here is always `cm.internalCtx` so wouldn't it be easier to access that from the `controllerManager` rather than adding a new arg?

pkg/manager/signals/signal.go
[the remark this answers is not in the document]
alvaroaleman: In a way jason has a good point though, I think the reason the context package has the `don't store it in a struct` comment is because a context is not only for canceling but also to transport request-scoped information. But we don't do that here, we only care about the canceling aspect, so the channel might actually a better fit than the context?

pkg/manager/internal.go
vincepri: Yeah good point

pkg/manager/internal.go
vincepri: It depends, given that we should really pass context as the first parameter, instead of injecting it into structs

pkg/manager/signals/signal.go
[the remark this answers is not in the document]
vincepri: Yeah, wasn't objecting to it, just wanted to point out that we're accepting a context now. In a way, it works like stdlib behaves in a lot of places (http server is an example) so I copied that approach

vincepri:
@DirectXMan12 Do you have some time to go over these changes? I'm happy to drop this PR if it's too many changes together and continue with smaller changes over time

pkg/manager/internal.go
alvaroaleman: Well the problem is that due to this terrible inject pattern the code will continue to compile but not work as before (I would assume it will just panic now for most ppl that relied on this). Can we maybe keep this for now, inject `cm.internalCtx.Done` and print a big fat deprecation warning in `inject.StopChannelInto` if this happens?

How is kubebuilder the CLI going to migrate projects?

pkg/manager/internal.go
```diff
@@ -423,7 +413,7 @@ func (cm *controllerManager) serveHealthProbes(stop <-chan struct{}) {
 		Handler: mux,
 	}
 	// Run server
-	cm.startRunnable(RunnableFunc(func(stop <-chan struct{}) error {
+	cm.startRunnable(RunnableFunc(func(ctx context.Context) error {
```
alvaroaleman, line 416: could we change this to `func(_ context.Context` to make it clearer, that the context we pass here is not used?

pkg/manager/manager_test.go
```diff
@@ -52,16 +53,6 @@ import (
 )
 
 var _ = Describe("manger.Manager", func() {
-	var stop chan struct{}
-
-	BeforeEach(func() {
-		stop = make(chan struct{})
```
alvaroaleman, line 58: Rather than creating and canceling a context in each testcase, can we just create the ctx and cancel in the BeforeEach and `cancel` in the `AfterEach`?

pkg/runtime/inject/inject.go
```diff
@@ -100,21 +100,6 @@ func SchemeInto(scheme *runtime.Scheme, i interface{}) (bool, error) {
 	return false, nil
 }
 
-// Stoppable is used by the ControllerManager to inject stop channel into Sources,
-// EventHandlers, Predicates, and Reconciles.
-type Stoppable interface {
-	InjectStopChannel(<-chan struct{}) error
```
alvaroaleman, line 106: As mentioned elsewhere, I don't think we should just remove it, because that will result in hard to debug panics down the line.

pkg/manager/manager_test.go
vincepri: The race detector will complain about using the context concurrently, so I changed it that way

pkg/runtime/inject/inject.go
vincepri: I'd rather have the context properly propagated within reconciler calls, rather than having depedency injection like these around, which would result in the context to be stored as a struct field

pkg/manager/internal.go
vincepri: I see, would we pass the ctx.Done() channel here then?

pkg/runtime/inject/inject.go
vincepri: (See reply above, we can inject ctx.Done())

pkg/manager/internal.go
joelanford: > Can we maybe keep this for now, inject cm.internalCtx.Done and print a big fat deprecation warning in inject.StopChannelInto if this happens?

:+1: 

> How is kubebuilder the CLI going to migrate projects?

Sounds like we'll keep this interface and injection in for now, so no changes necessary on the kubebuilder scaffolding side related to this particular aspect of the PR?

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
vincepri: Added a comment around the use of a context in struct

pkg/manager/signals/signal.go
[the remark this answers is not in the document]
vincepri: Fixed, thank you all

pkg/manager/internal.go
vincepri: @alvaroaleman Mind reviewing the last changes related to `StopChannelInto`, I'm a little confused about this interface and the whole injection code left me with some questions, but we can fix those later. For the purpose of this PR, can you double check that the logic here is restored as you'd expect?

pkg/manager/internal.go
vincepri: Done

pkg/controller/controller_integration_test.go
```diff
@@ -38,23 +38,18 @@ import (
...
-	AfterEach(func() {
-		close(stop)
-	})
-
 	Describe("controller", func() {
 		// TODO(directxman12): write a whole suite of controller-client interaction tests
 
 		It("should reconcile", func(done Done) {
+			ctx := context.Background()
+
```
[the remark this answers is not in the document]
vincepri: It should be used now, thank you!

pkg/controller/controller.go
```diff
@@ -66,7 +67,7 @@ type Controller interface {
 
 	// Start starts the controller.  Start blocks until stop is closed or a
```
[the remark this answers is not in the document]
vincepri: Done, thank you!

pkg/cache/informertest/fake_cache.go
```diff
@@ -121,7 +121,7 @@ func (c *FakeInformers) informerFor(gvk schema.GroupVersionKind, _ runtime.Objec
 }
 
 // Start implements Informers
-func (c *FakeInformers) Start(stopCh <-chan struct{}) error {
+func (c *FakeInformers) Start(ctx context.Context) error {
 	return c.Error
```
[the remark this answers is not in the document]
vincepri: Do you want to do this in tests as well?

pkg/cache/multi_namespace_cache.go
```diff
@@ -94,23 +94,23 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
 	return &multiNamespaceInformer{namespaceToInformer: informers}, nil
 }
 
-func (c *multiNamespaceCache) Start(stopCh <-chan struct{}) error {
+func (c *multiNamespaceCache) Start(ctx context.Context) error {
```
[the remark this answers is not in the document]
vincepri: Did a quick scan throughout our codebase and fixed all the comments I could find that were outdated (e.g. referenced a stop channel)

pkg/envtest/webhook_test.go
```diff
@@ -31,9 +31,10 @@ var _ = Describe("Test", func() {
 			server := m.GetWebhookServer()
 			server.Register("/failing", &webhook.Admission{Handler: &rejectingValidator{}})
 
-			stopCh := make(chan struct{})
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
```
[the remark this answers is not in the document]
vincepri: Done!

pkg/cache/internal/deleg_map.go
```diff
@@ -57,25 +57,25 @@ func NewInformersMap(config *rest.Config,
...
 // Start calls Run on each of the informers and sets started to true.  Blocks on the stop channel.
-func (m *InformersMap) Start(stop <-chan struct{}) error {
-	go m.structured.Start(stop)
-	go m.unstructured.Start(stop)
-	<-stop
+func (m *InformersMap) Start(ctx context.Context) error {
+	go m.structured.Start(ctx)
+	go m.unstructured.Start(ctx)
+	<-ctx.Done()
 	return nil
```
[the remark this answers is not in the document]
vincepri: Actually, just reverted this to still return nil, I'll open a different issue to tackle this in a different PR, but all the tests are expecting Start to return nil, and it doesn't take into account `context.Cancelled` error when the context is cancelled at the end of a test, does that sound good?

pkg/manager/internal.go
alvaroaleman: I don't see anything related to that, it still seems to be gone, am I completely missing it?

pkg/cache/internal/deleg_map.go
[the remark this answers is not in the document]
alvaroaleman: I think that makes more sense, because usually the manager is started with something like
```
if err := mgr.Start(ctx); err != nil {
    log.Fatal(err, "manager failed")
}
```

even thought a canceled context is an expected condition.

pkg/manager/internal.go
vincepri: Done, must have forgotten to push the changes

alvaroaleman: I still only see the removed code?

vincepri: * actually pushed now *

pkg/cache/cache_test.go
```diff
@@ -394,11 +395,13 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					Expect(err).NotTo(HaveOccurred())
 
 					By("running the cache and waiting for it to sync")
+					ctx, cancel := context.WithCancel(context.Background())
```
alvaroaleman, line 398: Why do we need a new `context` here and in the two tests below?

pkg/manager/manager_test.go
```diff
@@ -1244,20 +1250,14 @@ var _ = Describe("manger.Manager", func() {
 				},
 			})
 			Expect(err).To(Equal(expected))
-			err = m.SetFields(&injectable{
```
alvaroaleman, line 1247: I think we should keep that or not?

alvaroaleman (commented):
Two questions, other than that lgtm. It would be great if we could fine someone else to give this a pass though in case I missed something due to the sheer size of the PR.

pkg/cache/cache_test.go
vincepri: Fixed

pkg/manager/manager_test.go
vincepri: Fixed

vincepri:
@DirectXMan12 @ncdc @detiber If you have some time, an extra set of 👀 would be greatly appreciated

alvaroaleman (approved):
/lgtm
/hold
so others can have a look

ncdc:
I think this LGTM. There's an open question still for FakeInformers.Start() - ?

vincepri:
Thanks! I'm ok having that come in later, it might break tests if we return ctx.Err(), which seems we've decided not to do in https://github.com/kubernetes-sigs/controller-runtime/pull/1116#discussion_r493008378

vincepri:
/hold cancel

Checked with @alvaroaleman before removing the hold, let's get this merged so we can unblock other work :)

PR #1117, opened 2020-08-05
:sparkles: Add controllerutil.IsOwner functions to check that object has existing reference

### LINKED ISSUE #1114: Add controllerutil.IsOwner functions to check that object has existing reference
It would be nice to have a controllerutil.IsOwner functions to check that `object` has `owner` reference based on current `indexOwnerRef` implementation.
If `object` has another than `owner` reference the `AlreadyOwnedError` is thrown, but it's not good to throw this error for the current `owner` behavior

dmvolod:
IsOwner is a helper method to validate if existing object has provided owner as reference
It prevents object from unnecessary updates in case of the owner has not changed

Fixes #1114

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -119,6 +119,36 @@ func SetOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) erro
 
 }
 
+// IsOwner is a helper method to validate if existing object has provided owner as reference
+// It prevent objects from unnecessary updates in case of the owner has not changed
+func IsOwner(owner, object Object, scheme *runtime.Scheme) (bool, error) {
+	// Validate the owner.
+	ro, ok := owner.(runtime.Object)
+	if !ok {
+		return false, fmt.Errorf("%T is not a runtime.Object, cannot call SetOwnerReference", owner)
+	}
+	if err := validateOwner(owner, object); err != nil {
+		return false, err
+	}
+	// Create a new owner ref.
+	gvk, err := apiutil.GVKForObject(ro, scheme)
+	if err != nil {
+		return false, err
+	}
+	ref := metav1.OwnerReference{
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@dmvolod whats the problem with doing a no-op update call?

dmvolod:
@alvaroaleman thanks for response
if it validated by controller and don't really make an update every time, there is no problem. Is it true or it will send an update call every time?

alvaroaleman:
> if it validated by controller and don't really make an update every time, there is no problem. Is it true or it will send an update call every time?

Sorry, I don't understand what you are saying. Just doing an `UPDATE` without changes should work.

dmvolod:
> > if it validated by controller and don't really make an update every time, there is no problem. Is it true or it will send an update call every time?
> 
> Sorry, I don't understand what you are saying. Just doing an `UPDATE` without changes should work.

Yes, if OwnerReference object is not changed the update on the API server will not execute or it always calls for validation?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -119,6 +119,36 @@ func SetOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) erro
 
 }
 
+// HasExistingReference is a helper method to validate if existing object has provided owner as reference
+// It prevent objects from unnecessary updates in case of the owner has not changed
+func HasExistingReference(owner, object metav1.Object, scheme *runtime.Scheme) (bool, error) {
```
vincepri, line 124:
```suggestion
// IsOwner is a helper method to validate if existing object has provided owner as reference
// It prevent objects from unnecessary updates in case of the owner has not changed
func IsOwner(owner, object Object, scheme *runtime.Scheme) (bool, error) {
```

dmvolod: Fixed, thanks @vincepri

PR #1118, opened 2020-08-05
:warning: Use controllerutil.Object for event handling

alvaroaleman:
This change makes it significantly easier to interact with the various functions in the handling stack by using `controllerutil.Object` for events rather than `runtime.Object + metav1.Object` in dedicated fields.


pkg/event/event.go
```diff
@@ -17,44 +17,31 @@ limitations under the License.
...
 
 // CreateEvent is an event where a Kubernetes object was created.  CreateEvent should be generated
 // by a source.Source and transformed into a reconcile.Request by an handler.EventHandler.
 type CreateEvent struct {
-	// Meta is the ObjectMeta of the Kubernetes Type that was created
-	Meta metav1.Object
-
 	// Object is the object from the event
-	Object runtime.Object
+	Object controllerutil.Object
```
vincepri, line 27: Sidenote: Should we move this Object interface somewhere else?

vincepri: What do you think having it under `reconcile.Object`?

alvaroaleman: I think pretty much all places we have are subpar and this should be in upstream somewhere. We can still move it around freely later and then alias it.

If possible I would like to keep that discussion out of this PR :)

vincepri: Well, while I agree that there needs to be a better place to have this interface, it feels weird that we're now using an `util` package throughout our codebase. We might also get into circular dependencies if at any point controllerutil needs to use a symbol from one of the packages that now uses this interface

alvaroaleman: Yeah but solving this now is premature optimization. If we actually get that problem, we can still move the definition elsewhere and make `controllerutil.Object` an alias.

I guess  my point here is mostly that I think that discussion is outside of the scope of this PR.

vincepri: > I guess my point here is mostly that I think that discussion is outside of the scope of this PR.

+1, was brainstorming to follow-up with another PR later, mostly wanted to agree on _something_ and change it before we cut v0.7.0

PR #1119, opened 2020-08-06
:sparkles: Simplify usage of EnqueueRequestsFromMapFunc

alvaroaleman:
Makes creation of custom mappers a lot simpler

/assign @vincepri

### DIFF
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -24,8 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 )
 
-var _ EventHandler = &EnqueueRequestsFromMapFunc{}
-
 // EnqueueRequestsFromMapFunc enqueues Requests by running a transformation function that outputs a collection
 // of reconcile.Requests on each Event.  The reconcile.Requests may be for an arbitrary set of objects
 // defined by some user specified transformation of the source Event.  (e.g. trigger Reconciler for a set of objects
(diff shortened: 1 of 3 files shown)


pkg/handler/enqueue_mapped.go
```diff
@@ -25,6 +25,22 @@ import (
...
+//
+// For UpdateEvents which contain both a new and old object, the transformation function is run on both
+// objects and both sets of Requests are enqueue.
+func FromMapFunc(mapFN func(MapObject) []reconcile.Request) EventHandler {
+	return &EnqueueRequestsFromMapFunc{
+		ToRequests: ToRequestsFunc(mapFN),
+	}
+}
+
 var _ EventHandler = &EnqueueRequestsFromMapFunc{}
```
alvaroaleman, line 44: Alternatively, we can completely unexport `EnqueueRequestsFromMapFunc`. It doesn't provide any advantage over using `FromMapFunc` and is a lot more annoying to construct

vincepri: +1 to that, better have a single way to use it than two and confuse users

vincepri: Let's make that a ⚠️

alvaroaleman: Updated, PTAL

pkg/handler/enqueue_mapped.go
```diff
@@ -36,33 +34,41 @@ var _ EventHandler = &EnqueueRequestsFromMapFunc{}
 //
 // For UpdateEvents which contain both a new and old object, the transformation function is run on both
 // objects and both sets of Requests are enqueue.
-type EnqueueRequestsFromMapFunc struct {
+func EnqueueRequestsFromMapFunc(mapFN func(MapObject) []reconcile.Request) EventHandler {
```
vincepri, line 37:
```suggestion
func MapFunc(mapFN func(MapObject) []reconcile.Request) EventHandler {
```

Do we want to go for something a little simpler? wdyt?

alvaroaleman: All existing mappers are called `EnqueueRequestSomething` I thought it would be nice to keep that pattern

alvaroaleman: It also makes this change easier for existing users

DirectXMan12:
FYI, this is breaking

PR #1121, opened 2020-08-07
✨ Added the LabelSelectorPredicate function for filtering events

VenkatRamaraju:
The LabelSelectorPredicate function implements predicate functions to only reconcile events that contain certain labels.

This allows a user to pre-determine which events will undergo reconciliation via a Selector object that will compare against an event's labels, thereby "filtering" events.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -17,6 +17,8 @@ limitations under the License.
 package predicate
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
@@ -267,4 +269,16 @@ func (o or) Generic(e event.GenericEvent) bool {
 		}
 	}
 	return false
+}
+
+// LabelSelectorPredicate constructs a Predicate from a LabelSelector.
+// Only objects matching the LabelSelector will be admitted.
+func LabelSelectorPredicate(s metav1.LabelSelector) (Predicate, error) {
+	selector, err := metav1.LabelSelectorAsSelector(&s)
+	if err != nil {
+		return Funcs{}, err
+	}
+	return NewPredicateFuncs(func(o controllerutil.Object) bool {
+		return selector.Matches(labels.Set(o.GetLabels()))
+	}), nil
 }
--- a/pkg/predicate/predicate_test.go
+++ b/pkg/predicate/predicate_test.go
@@ -516,4 +516,35 @@ var _ = Describe("Predicate", func() {
 			})
 		})
 	})
+
+	Describe("When checking a LabelSelectorPredicate", func() {
+		instance, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}})
+		if err != nil {
+			Fail("Improper Label Selector passed during predicate instantiation.")
+		}
+
+		Context("When the Selector does not match the event labels", func() {
+			It("should return false", func() {
+				failMatch := &corev1.Pod{}
+				Expect(instance.Create(event.CreateEvent{Object: failMatch})).To(BeFalse())
+				Expect(instance.Delete(event.DeleteEvent{Object: failMatch})).To(BeFalse())
+				Expect(instance.Generic(event.GenericEvent{Object: failMatch})).To(BeFalse())
+				Expect(instance.Update(event.UpdateEvent{ObjectNew: failMatch})).To(BeFalse())
+			})
+		})
+
+		Context("When the Selector matches the event labels", func() {
+			It("should return true", func() {
+				successMatch := &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{"foo": "bar"},
+					},
+				}
+				Expect(instance.Create(event.CreateEvent{Object: successMatch})).To(BeTrue())
+				Expect(instance.Delete(event.DeleteEvent{Object: successMatch})).To(BeTrue())
+				Expect(instance.Generic(event.GenericEvent{Object: successMatch})).To(BeTrue())
+				Expect(instance.Update(event.UpdateEvent{ObjectNew: successMatch})).To(BeTrue())
+			})
+		})
+	})
 })


pkg/predicate/predicate.go
```diff
@@ -268,3 +271,48 @@ func (o or) Generic(e event.GenericEvent) bool {
...
+	Selector labels.Selector
+}
+
+// eventFilter skips reconciliation of events that have labels matching selectors
+func (r LabelSelectorPredicate) eventFilter(eventLabels map[string]string) bool {
+	return r.Selector.Matches(labels.Set(eventLabels))
+}
+
+// NewLabelSelectorPredicate instantiates a new LabelSelectorPredicate with the LabelSelector specified through parameters.
+func NewLabelSelectorPredicate(s metav1.LabelSelector) (Predicate, error) {
```
alvaroaleman, line 294: I think the implementation of this can be drastically simplified by making it use `NewPredicateFuncs`

pkg/predicate/predicate.go
```diff
@@ -268,3 +271,48 @@ func (o or) Generic(e event.GenericEvent) bool {
...
+// LabelSelectorPredicate implements predicate functions and requires a selector that will filter an event by its labels.
+//
+// This predicate will skip reconciliation on all events that have labels that do not match the selector of the predicate.
+// A metav1.LabelSelector is passed as an input parameter while instantiating an LabelSelectorPredicate, and this is (explicitly)
+// converted to a labels.Selector object, and stored within the struct. When a predicate function is evoked, the event labels are checked
+// for a match with the predicate's selector to decide whether an event undergoes reconciliation or not.
+// If the labels match, the event will undergo reconciliation. If the labels do not match, it will skip reconciliation for that particular event.
+
+type LabelSelectorPredicate struct {
+	Funcs
```
alvaroaleman, line 284: Why the embedded `Funcs`? Its unused or not? And why is the type exported?

VenkatRamaraju: I realized that the Funcs are available by default. I will remove them from the struct and un-export the struct too.

pkg/predicate/predicate.go
VenkatRamaraju: `NewLabelSelectorPredicate ` is used to initialize the struct. If there is a way to have all types of predicate functions (Create, Delete, Update, Generic) pass through a "general" function with a "general object" as its parameter, we can simplify the implementation by doing the `Matches` within that.

I'm a little confused about the implementation you suggested, could you explain a little more how that would work?

pkg/predicate/predicate.go
alvaroaleman:
```
func NewLabelSelectorPredicate(s metav1.LabelSelector)(Predicate, error){
    selector, err := metav1.LabelSelectorAsSelector(&s)
    if err != nil {return nil, err}
    return NewPredicateFuncs(func (o controllerutil.Object) bool {
         selector.Matches(labels.Set(o.GetLabels())
    }
}

File: pkg/predicate/predicate.go (hunk 1 above)
Comment (alvaroaleman) [reply pkg/predicate/predicate.go:294]:
```
func NewLabelSelectorPredicate(s metav1.LabelSelector)(Predicate, error){
    selector, err := metav1.LabelSelectorAsSelector(&s)
    if err != nil {return nil, err}
    return NewPredicateFuncs(func (o controllerutil.Object) bool {
         selector.Matches(labels.Set(o.GetLabels())
    }
}

pkg/predicate/predicate.go
```diff
@@ -270,3 +270,16 @@ func (o or) Generic(e event.GenericEvent) bool {
 	}
 	return false
 }
+
+// EventReconcileFilter takes in a LabelSelector as a parameter and returns predicate functions. The labels of
+// events that are contained as parameters in these predicate functions are compared against the instantiated
+// Selector object and reconciliation of the event will only take place if the labels match.
+func EventReconcileFilter(s metav1.LabelSelector) (Funcs, error) {
```
alvaroaleman, line 277: Please rename to `LabelSelectorPredicate`, no one will find this by its current name. Same goes for the PR title.
Also, the PR has merge conflicts and needs to be squashed so there is only one commit.

VenkatRamaraju: Will do.

**Quick question:**
The conflict was caused because a lot of the event struct's fields have changed (since I forked this repo), now taking in a controllerutil object. The events require a `controllerutil.Object` type (interface), which I believe can be implemented as `metav1.Object`. However, the type that supports labels as metadata is the `metav1.ObjectMeta`. How would I initialize a `controllerutil.Object` -typed object that contains labels in it? That would help me write some tests.

Thanks!

alvaroaleman: All kube objects that are not lists fullfil the `controllerutil.Object` interface, for example `*corev1.Pod`

pkg/predicate/predicate.go
```diff
@@ -268,3 +270,16 @@ func (o or) Generic(e event.GenericEvent) bool {
 	}
 	return false
 }
+
+// LabelSelectorPredicate takes in a LabelSelector as a parameter and returns predicate functions. The labels of
```
alvaroaleman, line 274:
```suggestion
// LabelSelectorPredicate constructs a Predicate from a LabelSelector. Only objects matching the LabelSelector will be admitted.
```

pkg/predicate/predicate.go
```diff
@@ -268,3 +270,16 @@ func (o or) Generic(e event.GenericEvent) bool {
...
+// events that are contained as parameters in these predicate functions are compared against the instantiated
+// Selector object and reconciliation of the event will only take place if the labels match.
+func LabelSelectorPredicate(s metav1.LabelSelector) (Funcs, error) {
```
alvaroaleman, line 277: Please change this to return `(Predicate, error)`. The `Funcs` are an implementation detail ppl should not rely on.

pkg/predicate/predicate_test.go
```diff
@@ -516,4 +516,39 @@ var _ = Describe("Predicate", func() {
 			})
 		})
 	})
+
+	Describe("When checking a LabelSelectorPredicate", func() {
+		instance, err := predicate.LabelSelectorPredicate(*(&metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}}))
```
alvaroaleman, line 521: Why do you first construct a pointer just to instantly dereference it?

pkg/predicate/predicate_test.go
```diff
@@ -516,4 +516,39 @@ var _ = Describe("Predicate", func() {
...
+		if err != nil {
+			Fail("Improper Label Selector passed during predicate instantiation.")
+		}
+
+		Context("When the Selector does not match the event labels", func() {
+			It("should return false", func() {
+				failMatch := &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{"bar": "foo"},
```
alvaroaleman, line 530: Nit: Just using `&corev1.Pod{}` would be enough, no need to give ti a different set of labels

alvaroaleman:
Thanks @VenkatRamaraju !

Issue #723, opened 2019-12-09
Add default readiness check for webhook server?


ncdc:
In the Cluster API project, we're trying to automate the following in our CI testing:

1. Create a kind cluster
1. Deploy cert-manager
1. Wait for cert-manager to be ready
1. Deploy Cluster API (CRDs, RBAC, Deployments, etc)
1. Wait for Cluster API pods to be ready (so our webhooks are ready)
1. Create a `Cluster`, `Machine`, etc.

When we're waiting for the Cluster API pods to be ready, we've noticed some test flakes because it appears the pods are ready but their webhook servers are not fully online and ready to respond to requests yet.

I've thought about adding a readiness probe to our Manager, but I believe there is still a chance that it could return a false positive; i.e., the check returns ready, but because the webhook server is technically a different server than the readiness server, the webhook server might not be ready yet.

Would it make sense to add a default readiness check for the webhook server?

DirectXMan12:
yeah, that makes sense

ncdc:
@DirectXMan12 I can work on this. How much flexibility would you like for the check? Should we add `AddReadyzCheck()` to `webhook.Server`? Or start simple and hard-code a simple ping response?

ncdc:
@vincepri I'm still interested in working on this. Do you have thoughts on my question above?

vincepri:
I'd start simple and hardcode the Readyz checks for now, we can provide a flag later. This should also be backward compatible

ncdc:
Define backward compat here?

vincepri:
As in, adding new endpoints for readiness is additive and doesn't impact the current behaviors of the webhook server

phoracek:
Just curious, what is this supposed to on top of #419 ? Just to use the probe by default? It seems fairly simple to add a readiness probe in the current state of the controller-runtime.

ncdc:
This would just be exposing checks that people could configure in their pods. Not sure there's any compat issues or impacts.

@phoracek please see the description above for the rationale.

alenkacz:
@ncdc are you still working on this? I've just spent many hours trying to figure out a nice way how to create such readiness check from outside that I am starting to think it would be easier to implement it within controller-runtime

ncdc:
@alenkacz not at the moment, no

PR #1124, opened 2020-08-09
✨ (:sparkles, minor) Make it possible to monitor readiness for webhook

alenkacz:
This is a very minimal implementation of #723

With this in place, one can define a readiness check for webhook like this:
```
err = mgr.AddReadyzCheck("ready", func(_ *http.Request) error {
		if mgr.GetWebhookServer().Started {
			return nil
		}
		return errors.New("Webhook server not yet started")
	})
```

there's still a possibility for a very small race because technically all the paths for webhook are bound when `srv.Serve` is done. But the possibility of hitting it is very small considering that the readiness check has to pick it up, return ready and someone has to make a request over network to that webhook - that will be definitely slower than the processing inside `srv.Serve`.

If anyone has better idea, I am all ears.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -76,6 +76,8 @@ type Server struct {
 
 	// defaultingOnce ensures that the default fields are only ever set once.
 	defaultingOnce sync.Once
+
+	Started bool
 }
 
 // setDefaults does defaulting for the Server.
@@ -226,6 +228,7 @@ func (s *Server) Start(stop <-chan struct{}) error {
 		close(idleConnsClosed)
 	}()
 
+	s.Started = true
 	err = srv.Serve(listener)
 	if err != nil && err != http.ErrServerClosed {
 		return err


alenkacz:
ping @ncdc @vincepri

ncdc:
@alenkacz thanks for this! @vincepri & I were both on vacation last week, so we'll take a look as soon as we've dug out from our email backlogs 😄

pkg/webhook/server.go
```diff
@@ -76,6 +76,8 @@ type Server struct {
 
 	// defaultingOnce ensures that the default fields are only ever set once.
 	defaultingOnce sync.Once
+
+	Started bool
```
vincepri, line 80: Let's document this field, I'm wondering though if we should expose it as a method rather than a field on the struct, given that this field isn't safe to use concurrently

pkg/webhook/server.go
```diff
@@ -226,6 +228,7 @@ func (s *Server) Start(stop <-chan struct{}) error {
 		close(idleConnsClosed)
 	}()
 
+	s.Started = true
 	err = srv.Serve(listener)
 	if err != nil && err != http.ErrServerClosed {
```
databus23, line 233: Should this maybe set back to false if the `Serve` method returns?

pkg/webhook/server.go
DirectXMan12: yeah, let's put that behind a method so that it's safe to call concurrently.  See also #1155

ncdc:
@alenkacz hi, are you still interested in getting this merged?

alenkacz:
I'll refresh this

sbueringer:
@alenkacz I can take over this issue, if it's okay for you.

sbueringer:
Has been implemented in #1588 
/lgtm

PR #1125, opened 2020-08-11
:sparkles: Add metrics for total workers and active workers

alvaroaleman:
Helps debugging issues around reconciliation duration. Right now its impossible to find out the number of workers from metrics and its also impossible to find out the number of active workers. The only thing we have is `workqueue_unfinished_work_seconds` which if big is probably not good but its not clearly indicating if we have one worker that takes longer, if all workers are blocked etc.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -169,6 +169,7 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 
 		// Launch workers to process resources
 		c.Log.Info("Starting workers", "worker count", c.MaxConcurrentReconciles)
+		ctrlmetrics.WorkerCount.WithLabelValues(c.Name).Set(float64(c.MaxConcurrentReconciles))
 		for i := 0; i < c.MaxConcurrentReconciles; i++ {
 			// Process work items
 			go wait.Until(c.worker, c.JitterPeriod, stop)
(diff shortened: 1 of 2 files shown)


pkg/internal/controller/controller.go
```diff
@@ -169,6 +169,7 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 
 		// Launch workers to process resources
 		c.Log.Info("Starting workers", "worker count", c.MaxConcurrentReconciles)
+		ctrlmetrics.WorkerCount.WithLabelValues(c.Name).Set(float64(c.MaxConcurrentReconciles))
```
estroz, line 172: Can you explain the logic behind collecting this metric?

alvaroaleman: * It allows correlating other metrics to the number of workers, for example workqueue_depth
* It allows calculating the `available_workers`

alvaroaleman:
@estroz PTAL

estroz:
Without knowing too much about how metrics are collected this looks fine to me. However there are now 7 metrics collected for a controller. Will an increase in the number of metrics impact cluster resources significantly?

/lgtm

estroz:
Cancel whenever

/hold

alvaroaleman:
> Without knowing too much about how metrics are collected this looks fine to me. However there are now 7 metrics collected for a controller. Will an increase in the number of metrics impact cluster resources significantly?

No, the way this works is that only the current value of a metric is stored and Prometheus is expected to collect it. The thing to keep in mind is to not expose too many metrics which usually happens when using labels that have a yet-to-be-determined set of values, because every metric+label combinaation actually ends up being an unique metric in Prometheus: https://www.robustperception.io/cardinality-is-key

/hold cancel

PR #1129, opened 2020-08-18
:warning: admission responses with raw Status

Shpectator:
If a custom webhook validator returns a metav1.Status object as the error, the validation handler will build an admission response with this Status object as is, instead of building a new Status object based on err.Error()

This might be considered as a breaking change since the behaviour of the
validation handler is going to be slightly different from what the users
might expect based on the previous version of the code

### DIFF
--- a/pkg/webhook/admission/validator.go
+++ b/pkg/webhook/admission/validator.go
@@ -20,9 +20,10 @@ import (
 	"context"
 	"net/http"
 
+	goerrors "errors"
+
 	"k8s.io/api/admission/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
@@ -69,12 +70,10 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 		}
 
 		err = obj.ValidateCreate()
-
 		if err != nil {
-
-			isStatusError, status := isStatusError(&err)
-			if isStatusError {
-				return ValidationResponseFromStatus(false, status)
+			var apiStatus errors.APIStatus
+			if goerrors.As(err, &apiStatus) {
+				return validationResponseFromStatus(false, apiStatus.Status())
 			}
 			return Denied(err.Error())
 		}
@@ -93,11 +92,10 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 		}
 
 		err = obj.ValidateUpdate(oldObj)
-
 		if err != nil {
-			isStatusError, status := isStatusError(&err)
-			if isStatusError {
-				return ValidationResponseFromStatus(false, status)
+			var apiStatus errors.APIStatus
+			if goerrors.As(err, &apiStatus) {
+				return validationResponseFromStatus(false, apiStatus.Status())
 			}
 			return Denied(err.Error())
 		}
@@ -113,22 +111,13 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 
 		err = obj.ValidateDelete()
 		if err != nil {
-			isStatusError, status := isStatusError(&err)
-			if isStatusError {
-				return ValidationResponseFromStatus(false, status)
+			var apiStatus errors.APIStatus
+			if goerrors.As(err, &apiStatus) {
+				return validationResponseFromStatus(false, apiStatus.Status())
 			}
 			return Denied(err.Error())
 		}
 	}
 
 	return Allowed("")
-}
-
-func isStatusError(err *error) (bool, *metav1.Status) {
-	statusError, isStatusError := (*err).(*errors.StatusError)
-	if isStatusError {
-		return true, &statusError.ErrStatus
-	}
-
-	return false, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/webhook/admission/validator.go
```diff
@@ -67,7 +69,13 @@ func (h *validatingHandler) Handle(ctx context.Context, req Request) Response {
 		}
 
 		err = obj.ValidateCreate()
+
 		if err != nil {
+
+			isStatusError, status := isStatusError(&err)
+			if isStatusError {
+				return ValidationResponseFromStatus(false, status)
```
DirectXMan12, line 77:
```suggestion
var statusError errors.StatusError
if goerrors.As(&statusError) {
  return ...
}
```

DirectXMan12: (the other nice thing about this is that if the status error is wrapped in such a way that it's intended to be unwrapped, we'll still pull the status)

DirectXMan12: is there any corresponding interface for status-returning errors in the k8s api errors package that we should be asserting on?  (basically, is there a pattern for "my custom error returns status information)

Shpectator:
I signed it

Shpectator:
ping cla bot

Shpectator:
I signed it

Shpectator:
@directxman12 I've addressed all your comments by modifying the existing commit (no new commits added)

DirectXMan12:
looks good, just need tests :-)

pkg/webhook/admission/validator_test.go
```diff
@@ -0,0 +1,280 @@
...
+						Raw:    []byte("{}"),
+						Object: handler.validator,
+					},
+				},
+			})
+			Expect(response.Allowed).Should(BeTrue())
+			Expect(response.Result.Code).Should(Equal(int32(http.StatusOK)))
+		})
+
+		It("should return 400 in response when update fails on decoding new object", func() {
```
DirectXMan12, line 104: these should be marked `PIt` to mark them as "pending" (not-yet-written" tests)

Also, don't leave TODOs without a name (yours) or bug number without context attached.

pkg/webhook/admission/validator_test.go
```diff
@@ -0,0 +1,280 @@
...
+func (v *fakeValidator) GroupVersionKind() schema.GroupVersionKind {
+	return fakeValidatorVK
+}
+
+func (v *fakeValidator) SetGroupVersionKind(gvk schema.GroupVersionKind) {}
+
+func createSucceedingValidatingHandler() *validatingHandler {
+	decoder, _ := NewDecoder(scheme.Scheme)
+	f := &fakeValidator{ErrorToReturn: nil}
+	return &validatingHandler{f, decoder}
```
DirectXMan12, line 260: use field names please

pkg/webhook/admission/validator_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"k8s.io/api/admission/v1beta1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+var _ = Describe("validatingHandler", func() {
+	Describe("Handle", func() {
```
DirectXMan12, line 20: nit: don't need an extra level of describe here

pkg/webhook/admission/validator_test.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (v *fakeValidator) GroupVersionKind() schema.GroupVersionKind {
+	return fakeValidatorVK
+}
+
+func (v *fakeValidator) SetGroupVersionKind(gvk schema.GroupVersionKind) {}
+
+func createSucceedingValidatingHandler() *validatingHandler {
+	decoder, _ := NewDecoder(scheme.Scheme)
```
DirectXMan12, line 258: if you just make decoder a global, you can inline much of this, which might make improve readability.

Alternatively, something like this:

```go
Describe("validating handler", func() {
   Context("when dealing with status errors", func() {
       handler := &fakeHandler{...}
       It("should propagate the status from ValidateDelete's return value to the HTTP response", func() { ... })
       // etc
   })
   Context("when dealing with non-status errors", func() { ... })
   Context("when dealing with successful results", func() { ... })
})
```

pkg/webhook/admission/validator_test.go
Shpectator: Fixed

pkg/webhook/admission/validator_test.go
Shpectator: Fixed

pkg/webhook/admission/validator_test.go
Shpectator: Fixed

pkg/webhook/admission/validator_test.go
Shpectator: Fixed

DirectXMan12:
awesome, looks great now

DirectXMan12:
hehe, this branch doesn't have an owners file.  I'll fix that quickly.

DirectXMan12:
o_O I'll file a bug about this flake.

/retest for the moment

PR #1141, opened 2020-08-26
:bug: Handle empty contexts in log.FromContext

### LINKED ISSUE #1140: log.FromContext doesn't add provided tags and doesn't handle empty contexts
I came across a couple of bugs in the `log.FromContext` function:
- If the passed in context doesn't contain a logger, it attempts to cast a nil value as `logr.Logger` -- which causes a panic.
- The `keysAndValues` argument has no effect because the logger returned from the `WithValues` method is ignored in the return value.

I have a patch that should address these here: https://github.com/kubernetes-sigs/controller-runtime/compare/master...charithe:fix/logging-ctx

charithe:
This fixes a couple of bugs in the `log.FromContext` function:
- If the passed in context doesn't contain a logger, it attempts to cast a nil value as `logr.Logger` -- which causes a panic.
- The `keysAndValues` argument has no effect because the logger returned from the `WithValues` method is ignored in the return value.

Fixes #1140

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -55,15 +55,15 @@ var Log = NewDelegatingLogger(NullLogger{})
 
 // FromContext returns a logger with predefined values from a context.Context.
 func FromContext(ctx context.Context, keysAndValues ...interface{}) logr.Logger {
-	var log logr.Logger
-	if ctx == nil {
-		log = Log
-	} else {
-		lv := ctx.Value(contextKey)
-		log = lv.(logr.Logger)
+	var log logr.Logger = Log
+
+	if ctx != nil {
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@charithe the commit needs an author email that is associated to your github user (You can use `charithe@@users.noreply.github.com` if you don't want to hand out a real email address)

charithe:
/check-cla

My email address is associated with GitHub. Not sure why that's not picked up.

charithe:
Seems to be an unrelated test failure.

/retest
/check-cla

charithe:
Another unrelated test failure.

/retest

PR #1144, opened 2020-08-29
⚠ Change leaderlock from ConfigMap to ConfigMapsLeasesResourceLock

alvaroaleman:
This commit changes the leaderlock to ConfigMapsLeasesResourceLock,
which is intended to transition ppl from the ConfigMap to the more
modern Leases. It is breaking because additional RBAC is required.

Related issue is https://github.com/kubernetes-sigs/controller-runtime/issues/460

PRing this now because this change is breaking, so must happen before we cut 0.7.0. Adding an option to override the default is non-breaking, so not as important to happen ASAP.

/assign @vincepri

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -302,6 +302,24 @@ var _ = Describe("manger.Manager", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring("unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace"))
 			})
+
+			// We must keep this default until we are sure all controller-runtime users have upgraded from the original default
+			// ConfigMap lock to a controller-runtime version that has this new default. Many users of controller-runtime skip
+			// versions, so we should be extremely conservative here.
+			It("should default to ConfigMapsLeasesResourceLock", func() {
+				m, err := New(cfg, Options{LeaderElection: true, LeaderElectionID: "controller-runtime", LeaderElectionNamespace: "my-ns"})
+				Expect(m).ToNot(BeNil())
+				Expect(err).ToNot(HaveOccurred())
+				cm, ok := m.(*controllerManager)
+				Expect(ok).To(BeTrue())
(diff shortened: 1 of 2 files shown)


vincepri:
Do we need a related change in controller-tools for RBAC?

pkg/manager/manager_test.go
```diff
@@ -302,6 +302,24 @@ var _ = Describe("manger.Manager", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring("unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace"))
 			})
+
+			// We must keep this default until we are sure all controller-runtime users have upgraded from the original default
+			// ConfigMap lock to a controller-runtime version that has this new default. Many users of controller-runtime skip
```
vincepri, line 307: How are we going to track this for all our users? We could maybe say that v0.7.x is going to be the last release to default to ConfigMap locks, and going forward there is going to be a breaking change that user need to take into account. 

The change can be part of upgrading and should be very explicit in release notes.

alvaroaleman: The `MultiLock` we set up requires both a `ConfigMap` and a `Lease`. It is intended for transitions. I would like to start using this ASAP (so in v0.7, not later), because we will have to stick to this for a very long time, until we are sure everyone had a c-r release that had the `MultiLock`. Othewise we can break users in a subtile way if their controller suddenly used a `Lease` but not a `ConfigMap` anymore, resulting in two leaders existing at the same time.

We also want to allow ppl to explicitly configure a different strategy if they know that all their users had a release with the `MultiLock`. However the default here has to be like this for a long time (certainly more than release, I would say at least a year after the first c-r release that has this change).

alvaroaleman:
> Do we need a related change in controller-tools for RBAC?

I don't know? I've never used it to generate rbac. Do you use it for that?

vincepri:
@alvaroaleman I'll report back on the RBAC changes once we start testing it in Cluster API, I don't think there is any _immediate_ change, but if there is some, we'll have to propagate in controller-tools and kubebuilder later

PR #1147, opened 2020-09-01
✨ Make leader election resourcelock configurable

### LINKED ISSUE #460: use leader election API if available

timebertt:
This PR makes the leader election resource lock configurable via `manager.Options.LeaderElectionResourceLock`.
It can be set to one of the lock types supported by client-go ([ref](https://github.com/kubernetes/client-go/blob/release-1.19/tools/leaderelection/resourcelock/interface.go#L31-L35)):
- `endpoints`
- `configmaps`
- `leases `
- `endpointsleases`
- `configmapsleases`

It keeps the default of `configmapsleases` introduced by #1144, but allows folks to explicitly configure other LE strategies.

Fixes #460

/assign @alvaroaleman

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -142,11 +142,26 @@ type Options struct {
 	// starting the manager.
 	LeaderElection bool
 
+	// LeaderElectionResourceLock determines which resource lock to use for leader election,
+	// defaults to "configmapsleases". Change this value only if you know what you are doing.
+	// Otherwise, users of your controller might end up with multiple running instances that
+	// each acquired leadership through different resource locks during upgrades and thus
+	// act on the same resources concurrently.
+	// If you want to migrate to the "leases" resource lock, you might do so by migrating to the
+	// respective multilock first ("configmapsleases" or "endpointsleases"), which will acquire a
+	// leader lock on both resources. After all your users have migrated to the multilock, you can
+	// go ahead and migrate to "leases". Please also keep in mind, that users might skip versions
+	// of your controller.
+	//
+	// Note: before controller-runtime version v0.7, the resource lock was set to "configmaps".
+	// Please keep this in mind, when planning a proper migration path for your controller.
+	LeaderElectionResourceLock string
+
 	// LeaderElectionNamespace determines the namespace in which the leader
-	// election configmap will be created.
+	// election resource will be created.
 	LeaderElectionNamespace string
 
-	// LeaderElectionID determines the name of the configmap that leader election
+	// LeaderElectionID determines the name of the resource that leader election
 	// will use for holding the leader lock.
 	LeaderElectionID string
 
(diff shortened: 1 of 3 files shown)


pkg/manager/manager_test.go
```diff
@@ -296,6 +296,15 @@ var _ = Describe("manger.Manager", func() {
 				<-m2done
 			})
 
+			It("should return an error if it can't create a ResourceLock", func() {
```
alvaroaleman, line 299: Please add a testcase where you call `manager.New` with any non-default resourcelock and verify that you end up getting that. You can use the existing `should default to ConfigMapsLeasesResourceLock` testcase to get some inspiration

timebertt: Makes sense, done. Extended the existing case to test all resource lock types. PTAL :)

pkg/manager/manager_test.go
```diff
@@ -296,30 +297,48 @@ var _ = Describe("manger.Manager", func() {
...
+					Expect(multiLock.Primary).To(BeAssignableToTypeOf(expectedLocks[0]))
+					Expect(multiLock.Secondary).To(BeAssignableToTypeOf(expectedLocks[1]))
+				} else {
+					Expect(cm.resourceLock).To(BeAssignableToTypeOf(expectedLocks[0]))
+				}
+			},
+				// We must keep this default until we are sure all controller-runtime users have upgraded from the original default
+				// ConfigMap lock to a controller-runtime version that has this new default. Many users of controller-runtime skip
+				// versions, so we should be extremely conservative here.
+				Entry("not specified (should default to ConfigMapsLeasesResourceLock)", "", true, &resourcelock.ConfigMapLock{}, &resourcelock.LeaseLock{}),
```
alvaroaleman, line 335: I don't think there is value in testing more than `use default` and `use non-default` and these testcases are super annoying to read. Please just do the stupid thing, leave the existing test case how it was and add a copy of it where you explicitly configure a different Lock and verify that gets passed through

timebertt: Fair enough, done. PTAL.

pkg/manager/manager.go
```diff
@@ -142,11 +142,15 @@ type Options struct {
 	// starting the manager.
 	LeaderElection bool
 
+	// LeaderElectionResourceLock determines which resource lock to use for leader election,
+	// defaults to "configmapsleases".
+	LeaderElectionResourceLock string
```
alvaroaleman, line 147: Sorry, I forgot to press submit on my comment yesterday. Please mention here that ppl should only change this if they know what they are doing and that they must make sure all their users are transitiion to the new lock via a multilock to the new lock before they change this, keeping in mind that users might skip versions of the controller.

timebertt: Done.

PR #1149, opened 2020-09-02
:warning: Use application/vnd.kubernetes.protobuf as content-type if possible

FillZpp:
Signed-off-by: Siyu Wang <FillZpp.pub@gmail.com>

Use `application/vnd.kubernetes.protobuf` as content-type if it is possible, currently for those native resources defined in client-go.

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -21,6 +21,7 @@ package apiutil
 
 import (
 	"fmt"
+	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,9 +29,32 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/discovery"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 )
+
+var (
+	protobufScheme     = runtime.NewScheme()
+	protobufSchemeLock sync.RWMutex
+)
+
+func init() {
+	// Currently only enabled for built-in resources which are guaranteed to implement Protocol Buffers.
+	// For custom resources, CRDs can not support Protocol Buffers but Aggregated API can.
+	// See doc: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility
+	if err := clientgoscheme.AddToScheme(protobufScheme); err != nil {
+		panic(err)
+	}
+}
+
+// AddToProtobufScheme add the given SchemeBuilder into protobufScheme, which should
+// be additional types that do support protobuf.
+func AddToProtobufScheme(builder runtime.SchemeBuilder) error {
+	protobufSchemeLock.Lock()
+	defer protobufSchemeLock.Unlock()
+	return builder.AddToScheme(protobufScheme)
+}
 
 // NewDiscoveryRESTMapper constructs a new RESTMapper based on discovery
 // information fetched by a new client with the given config.
@@ -93,8 +117,8 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 // RESTClientForGVK constructs a new rest.Interface capable of accessing the resource associated
 // with the given GroupVersionKind. The REST client will be configured to use the negotiated serializer from
 // baseConfig, if set, otherwise a default serializer will be set.
-func RESTClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config, codecs serializer.CodecFactory) (rest.Interface, error) {
-	cfg := createRestConfig(gvk, baseConfig)
+func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory) (rest.Interface, error) {
+	cfg := createRestConfig(gvk, isUnstructured, baseConfig)
 	if cfg.NegotiatedSerializer == nil {
 		cfg.NegotiatedSerializer = serializer.WithoutConversionCodecFactory{CodecFactory: codecs}
 	}
@@ -102,7 +126,7 @@ func RESTClientForGVK(gvk schema.GroupVersionKind, baseConfig *rest.Config, code
 }
 
 //createRestConfig copies the base config and updates needed fields for a new rest config
-func createRestConfig(gvk schema.GroupVersionKind, baseConfig *rest.Config) *rest.Config {
+func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config) *rest.Config {
 	gv := gvk.GroupVersion()
 
 	cfg := rest.CopyConfig(baseConfig)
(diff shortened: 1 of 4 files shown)


pkg/client/apiutil/apimachinery.go
```diff
@@ -93,5 +102,8 @@ func createRestConfig(gvk schema.GroupVersionKind, baseConfig *rest.Config) *res
 	if cfg.UserAgent == "" {
 		cfg.UserAgent = rest.DefaultKubernetesUserAgent()
 	}
+	if cfg.ContentType == "" && localScheme.Recognizes(gvk) {
```
alvaroaleman, line 105: So you copy the client-go scheme so this is only used for built-in types? Are you sure protobuf works for all built-in types on all kube server versions? Does upstream have any docs on this?

alvaroaleman: Do you know what happens if we request protobuf for a type that doesn't support it, is there maybe some way to negotiate?

FillZpp: > So you copy the client-go scheme so this is only used for built-in types?

 Yes, this is only used for built-in resources. For custom resources, `CRDs` can not support Protocol Buffers but `Aggregated API` can. You can see it in [this table](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility).

> Are you sure protobuf works for all built-in types on all kube server versions? Does upstream have any docs on this?

All built-in resources are guaranteed to implement Protocal Buffers:

- This [design doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/protobuf.md#schema)  for contributors says protobuf schema for each API group and version in k8s.io/api will be generated from the objects in that API group and version.
- [hack/update-generated-protobuf.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/update-generated-protobuf.sh) in Kubernetes repo will find all packages contains `+k8s:protobuf-gen=package` and generate protobuf schema for them. And all API definition package must contain a doc.go which has these comments:

```golang
// +k8s:deepcopy-gen=package
// +k8s:protobuf-gen=packag
```

pkg/client/apiutil/apimachinery.go
```diff
@@ -93,5 +102,8 @@ func createRestConfig(gvk schema.GroupVersionKind, baseConfig *rest.Config) *res
...
+		cfg.ContentType = "application/vnd.kubernetes.protobuf"
```
alvaroaleman, line 106: Nit: Can you use `runtime.ContentTypeProtobuf` here?

alvaroaleman (approved):
/approve
/hold
/assign @DirectXMan12

pkg/client/apiutil/apimachinery.go
FillZpp: Done

pkg/client/apiutil/apimachinery.go
```diff
@@ -27,10 +27,19 @@ import (
...
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 )
 
+var (
+	localScheme = runtime.NewScheme()
+)
+
+func init() {
+	_ = clientgoscheme.AddToScheme(localScheme)
```
DirectXMan12, line 40: this should not be using some sort of magic private scheme -- we should be using the user's passed-in scheme.  That way, if the user also defines protobuf support for their types (e.g. aggregated api servers, if we get proto support in the future for CRDs, etc) it'll just work.

DirectXMan12 (changes requested):

pkg/client/apiutil/apimachinery.go
DirectXMan12: missed this comment.  I *think* this should probably be a bit less magic, since this ignores aggregated api servers or any future improvements to CRDs.  This is probably fine in the short run, but in the long run I think we want to check discovery or something to make sure that this is actually true.

pkg/client/client_cache.go
```diff
@@ -74,11 +77,19 @@ func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
 	if err != nil {
 		return nil, err
 	}
+	_, isUnstructured := obj.(*unstructured.Unstructured)
+	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
+	isUnstructured = isUnstructured || isUnstructuredList
+
+	resourceByType := c.structuredResourceByType
```
DirectXMan12, line 84: move this so that it's under the lock

DirectXMan12 (changes requested):

pkg/client/apiutil/apimachinery.go
FillZpp: Understand, but currently there is no way to check if a resource type supports protobuf or not. I have added a `TODO` to describe this.

pkg/client/apiutil/apimachinery.go
FillZpp: Done

pkg/client/client_cache.go
FillZpp: Done

alvaroaleman (commented):
Some nits, other than that it looks good. Also this change is breaking due to the signature change of `RESTClientForGVK`.

pkg/client/apiutil/apimachinery.go
```diff
@@ -27,10 +27,22 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/discovery"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 )
 
+var (
+	localScheme = runtime.NewScheme()
```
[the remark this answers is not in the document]
FillZpp: All done

alvaroaleman (approved):
/lgtm
leaving hold cancel to @DirectXMan12

alvaroaleman:
and thanks for your work, this is a great change!

vincepri (commented):
LGTM
/milestone v0.7.x

FillZpp:
PTAL @DirectXMan12 when you have time

alvaroaleman:
Retitling because this PR includes a breaking change in RESTClientForGVK
/retitle :warning: Use application/vnd.kubernetes.protobuf as content-type if possible

vincepri:
Are we ok unholding this change? @alvaroaleman @DirectXMan12

pkg/client/apiutil/apimachinery.go
```diff
@@ -27,10 +27,24 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/discovery"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 )
 
+var (
+	protobufScheme = runtime.NewScheme()
```
alvaroaleman, line 36: Actually one nit: Can we add a setter for this so projects that have additional apis that do support protobuf (metrics-server, all the openshift apis that are served via its aggregated apiserver) can add themselves here?

alvaroaleman: I am also fine doing a quick follow-up for that and merging this PR as-is

FillZpp: > Actually one nit: Can we add a setter for this so projects that have additional apis that do support protobuf (metrics-server, all the openshift apis that are served via its aggregated apiserver) can add themselves here?

Sure. How about provide a function in `pkg/client/apiutil/apimachinery.go`?

```golang
func AdditionalProtobufScheme(builder runtime.SchemeBuilder) error {
    return builder.AddToScheme(protobufScheme)
}
```

alvaroaleman: Sounds good to me. Could you also add a small godoc comment explaining it?

vincepri: How about `AddToProtobufScheme` to make the experience familiar with the other AddToScheme methods?

FillZpp: I have add this func and a RWMutex to avoid race modification of protobufScheme.
 PTAL @alvaroaleman @vincepri

alvaroaleman (approved):
/lgtm

vincepri (approved):
/approve

@alvaroaleman good to unhold?

PR #1151, opened 2020-09-07
⚠️ DynamicRestMapper: return NoMatchError when resource doesn't exist

### LINKED ISSUE #1120: Dynamic RESTMapper returns ErrRateLimited instead of expected NoKindMatchError
/kind bug

In our code we have some `client.Delete` calls that should be ignored, when the respective resources are not available on the API server (e.g. additional CRDs not deployed, because some feature gate is disabled).
Therefore we use `meta.IsNoMatchError` and ignore the error, if it returns true.

So basically something like this:
```go
if err := client.Delete(ctx, obj); client.IgnoreNotFound(err) != nil {
	if meta.IsNoMatchError(err) {
		return nil
	}
	return err
}
```

When under high load, we have a lot of these calls in parallel, which might end up with `NoKindMatchError`s.

After switching to the dynamic RESTMapper in our clients, `meta.IsNoMatchError` can no longer catch this, as the dynamic RESTMapper returns `ErrRateLimited` instead of `NoKindMatchError`, which is (implicitly) expected from implementations of the `meta.RESTMapper` interface.

I went back and forth about how this can be improved and gave these options a thought:
1. returning the `NoKindMatchError` directly
    - is not backwards-compatible but would implement the expected behaviour
    - how would folks then detect that they are hitting the rate limit?
1. wrapping either the underlying `NoKindMatchError` or the `ErrRateLimited` with `fmt.Errorf("%w...` or `errors.Wrap`
    - you could then check the error with `errors.As(err, &NoKindMatchError)`
    - still kind of differing from the expected error and you will not be able to use `meta.IsNoMatchError` as well

I can do it in a PR, if we agree on one approach.

WDYT?
@estroz @DirectXMan12

timebertt:
Fixes #1120 

This PR fixes the DynamicRESTMapper's behaviour when it is rate limited:
Instead of returning `ErrRateLimited`, it now directly returns the `meta.NoKindMatchError` / `meta.NoResourceMatchError` given by the underlying mapper. This way, existing controller code using `meta.IsNoMatchError` works as expected also in the case when the RESTMapper is rate limited.

**This is breaking ⚠️  because:**
- the `DynamicRESTMapper` does now return a `NoKindMatchError`/`NoResourceMatchError` instead of `ErrRateLimited` in case the resource doesn't exist and it was rate limited
- `ErrRateLimited` and `DelayIfRateLimited` are removed, as they aren't needed anymore

**Additional notes:**
The given rate limit can now be seen as an agreement on how often the mapper should refresh its cache (at max). So if the limit is reached, it basically means, we have refreshed often enough (met the contract) and we can safely rely on the cache and directly return the error.

I improved the tests a bit along the way, to ensure all cases are properly tested and made them a bit faster.

### DIFF
--- a/pkg/client/apiutil/dynamicrestmapper.go
+++ b/pkg/client/apiutil/dynamicrestmapper.go
@@ -19,7 +19,6 @@ package apiutil
 import (
 	"errors"
 	"sync"
-	"time"
 
 	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -29,34 +28,12 @@ import (
 	"k8s.io/client-go/restmapper"
 )
 
-// ErrRateLimited is returned by a RESTMapper method if the number of API
-// calls has exceeded a limit within a certain time period.
-type ErrRateLimited struct {
-	// Duration to wait until the next API call can be made.
-	Delay time.Duration
-}
-
-func (e ErrRateLimited) Error() string {
-	return "too many API calls to the RESTMapper within a timeframe"
-}
-
-// DelayIfRateLimited returns the delay time until the next API call is
-// allowed and true if err is of type ErrRateLimited. The zero
-// time.Duration value and false are returned if err is not a ErrRateLimited.
-func DelayIfRateLimited(err error) (time.Duration, bool) {
-	var rlerr ErrRateLimited
-	if errors.As(err, &rlerr) {
-		return rlerr.Delay, true
-	}
-	return 0, false
-}
-
 // dynamicRESTMapper is a RESTMapper that dynamically discovers resource
 // types at runtime.
 type dynamicRESTMapper struct {
 	mu           sync.RWMutex // protects the following fields
 	staticMapper meta.RESTMapper
-	limiter      *dynamicLimiter
+	limiter      *rate.Limiter
 	newMapper    func() (meta.RESTMapper, error)
 
 	lazy bool
@@ -70,7 +47,7 @@ type DynamicRESTMapperOption func(*dynamicRESTMapper) error
 // WithLimiter sets the RESTMapper's underlying limiter to lim.
 func WithLimiter(lim *rate.Limiter) DynamicRESTMapperOption {
 	return func(drm *dynamicRESTMapper) error {
-		drm.limiter = &dynamicLimiter{lim}
+		drm.limiter = lim
 		return nil
 	}
 }
@@ -103,9 +80,7 @@ func NewDynamicRESTMapper(cfg *rest.Config, opts ...DynamicRESTMapperOption) (me
 		return nil, err
 	}
 	drm := &dynamicRESTMapper{
-		limiter: &dynamicLimiter{
-			rate.NewLimiter(rate.Limit(defaultRefillRate), defaultLimitSize),
-		},
+		limiter: rate.NewLimiter(rate.Limit(defaultRefillRate), defaultLimitSize),
 		newMapper: func() (meta.RESTMapper, error) {
 			groupResources, err := restmapper.GetAPIGroupResources(client)
 			if err != nil {
@@ -161,12 +136,13 @@ func (drm *dynamicRESTMapper) init() (err error) {
 // checkAndReload attempts to call the given callback, which is assumed to be dependent
 // on the data in the restmapper.
 //
-// If the callback returns a NoKindMatchError, it will attempt to reload
+// If the callback returns an error that matches the given error, it will attempt to reload
 // the RESTMapper's data and re-call the callback once that's occurred.
 // If the callback returns any other error, the function will return immediately regardless.
 //
-// It will take care
-// ensuring that reloads are rate-limitted and that extraneous calls aren't made.
+// It will take care of ensuring that reloads are rate-limited and that extraneous calls
+// aren't made. If a reload would exceed the limiters rate, it returns the error return by
+// the callback.
 // It's thread-safe, and worries about thread-safety for the callback (so the callback does
 // not need to attempt to lock the restmapper).
 func (drm *dynamicRESTMapper) checkAndReload(needsReloadErr error, checkNeedsReload func() error) error {
@@ -199,7 +175,9 @@ func (drm *dynamicRESTMapper) checkAndReload(needsReloadErr error, checkNeedsRel
 	}
 
 	// we're still stale, so grab a rate-limit token if we can...
-	if err := drm.limiter.checkRate(); err != nil {
+	if !drm.limiter.Allow() {
+		// return error from static mapper here, we have refreshed often enough (exceeding rate of provided limiter)
+		// so that client's can handle this the same way as a "normal" NoResourceMatchError / NoKindMatchError
 		return err
 	}
 
@@ -304,20 +282,4 @@ func (drm *dynamicRESTMapper) ResourceSingularizer(resource string) (string, err
 		return err
 	})
 	return singular, err
-}
-
-// dynamicLimiter holds a rate limiter used to throttle chatty RESTMapper users.
-type dynamicLimiter struct {
-	*rate.Limiter
-}
-
-// checkRate returns an ErrRateLimited if too many API calls have been made
-// within the set limit.
-func (b *dynamicLimiter) checkRate() error {
-	res := b.Reserve()
-	if res.Delay() == 0 {
-		return nil
-	}
-	res.Cancel()
-	return ErrRateLimited{res.Delay()}
 }
--- a/pkg/client/apiutil/dynamicrestmapper_test.go
+++ b/pkg/client/apiutil/dynamicrestmapper_test.go
@@ -1,11 +1,12 @@
 package apiutil_test
 
 import (
-	"errors"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
 	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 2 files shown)


pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -29,34 +28,12 @@ import (
...
-}
-
-func (e ErrRateLimited) Error() string {
-	return "too many API calls to the RESTMapper within a timeframe"
-}
-
-// DelayIfRateLimited returns the delay time until the next API call is
-// allowed and true if err is of type ErrRateLimited. The zero
-// time.Duration value and false are returned if err is not a ErrRateLimited.
-func DelayIfRateLimited(err error) (time.Duration, bool) {
```
timebertt, line 46: We can also think about deprecating `DelayIfRateLimited` (and always returning `0, false`) instead of removing it.
This would make this PR 🐛 instead of ⚠️ FMPOV, so we could cherry-pick it to `release-0.x`.
WDYT?

DirectXMan12: could also do a wrapper error that wraps both errors if it's a nokindmatcherror that's *caused* by rate limitting.  This has the downside of not working with `meta.IsNoMatchError` (but that function is actually not really correct in Go 1.13+) until they're updated to work with Go's 1.13 standard `errors` package, but has the advantage that it works with idiomatic Go:

```go
if errors.As(err, &meta.NoKindMatchError{}) {
  // no kind matched
}
```

as well as still allowing you to have `DelayIfRateLimited` work.

I'd imagine the implementation would look like:

```go
type ErrorNoMatch struct {
  Delay time.Duration
  matchError error
}

func (e *ErrNoMatch) Unwrap()  error {
  return e.matchError
}
// implement Error() as necessary

// alias for backwards compat

type ErrRateLimited = ErrNoMatch
```

WDYT?

DirectXMan12 (commented):
like the general concept.  Left a comment re: your question.

pkg/client/apiutil/dynamicrestmapper.go
timebertt: Thanks @DirectXMan12 for your suggestion. I like it wrt using go errors idiomatically and also keeping backwards compatability.

Although I am not voting totally against it, I want to note something:
TBH I don't see a real benefit of annotating the `NoKindMatchError` with a delay. I can't think of a use case where a controller/consumer would actually want to distinguish between a "normal" nokindmatch and a nokindmatch that might have been returned by a slightly outdated cache, and then wait until it can retry.
If you added a new CRD for example and you want to add a CR, then you will have to expect a slight delay until the resource comes available in the discovery API anyways, right?
Also, IISIC the code actually doesn't guarantee the caller, that they won't be rate limited again after waiting the returned delay (reservation is cancelled and other callers trying to rediscover might sneak in).

I would think of the specified rate limit as some kind of a SLA between the controller/consumer and the RESTMapper, which says "Please make sure, my discovery cache is not outdated by more than 1s" (if you specify a rate limiter with 1rps).
And if you're call was rate limited because some other rediscoveries exhausted the rate limited only a few milliseconds before yours, than your cache is very likely not to be outdated, although the rate limiter didn't allow the request.

It quite unfortunate, that upstream has not adapted go1.13 errors, yet. But I would expect that an implementation of the `meta.RESTMapper` interface is also working as expected with the provided helper functions in the `meta` package, like `meta.IsNoMatchError`.


As said, I'm not against implementing your approach, I just wanted to give some more background on the motivation why I started with the proposed implementation.

timebertt:
@DirectXMan12 WDYT about my comment?

alvaroaleman:
@timebertt can we retitle the PR to clarify the problem its solving rather than implementation details (something like "DynamicRestMapper: Return NoMatchError when resource type doesn't exist") ?

timebertt:
/title DynamicRestMapper: return NoMatchError when resource doesn't exist

Any thought on the compatability story, @alvaroaleman?

timebertt:
Pinging @alvaroaleman @DirectXMan12 
Any additional feedback on this PR?

timebertt:
I made this PR backwards-compatible by deprecating `ErrRateLimited` and associated funcs instead of just removing it.

Can we now proceed with the PR?
/assign @DirectXMan12 @alvaroaleman

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -31,6 +31,9 @@ import (
 
 // ErrRateLimited is returned by a RESTMapper method if the number of API
 // calls has exceeded a limit within a certain time period.
+// Deprecated: as the dynamic RESTMapper now returns NoMatch errors directly
```
vincepri, line 34:
```suggestion
// calls has exceeded a limit within a certain time period.
//
// DEPRECATED: As the dynamic RESTMapper now returns NoMatch errors directly
```

Just to give more emphasis :)

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -43,6 +46,9 @@ func (e ErrRateLimited) Error() string {
 // DelayIfRateLimited returns the delay time until the next API call is
 // allowed and true if err is of type ErrRateLimited. The zero
 // time.Duration value and false are returned if err is not a ErrRateLimited.
+// Deprecated: as the dynamic RESTMapper now returns NoMatch errors directly
```
vincepri, line 49:
```suggestion
// time.Duration value and false are returned if err is not a ErrRateLimited.
//
// DEPRECATED: As the dynamic RESTMapper now returns NoMatch errors directly
```

pkg/client/apiutil/dynamicrestmapper.go
vincepri: Given that we're working on v0.7.x on the main branch, can we actually just remove this to avoid confusion? 

cc @alvaroaleman @DirectXMan12

vincepri (commented):
/ok-to-test

pkg/client/apiutil/dynamicrestmapper.go
timebertt: > Just to give more emphasis :)

`Deprecated:` is actually convention and some tools test for this explicitly (ref https://github.com/golang/go/wiki/Deprecated), so I would keep it, if you don't mind :)

timebertt: > Given that we're working on v0.7.x on the main branch, can we actually just remove this to avoid confusion?

Sure, I'm open for it. You decide, what better fits you :)

vincepri: > Deprecated: is actually convention and some tools test for this explicitly (ref https://github.com/golang/go/wiki/Deprecated), so I would keep it, if you don't mind :)

SGTM, let's add a space though between the deprecation notice and the original comment, if we're going down this path.

> Sure, I'm open for it. You decide, what better fits you :)

I'm +1 for it, although I'd like to wait to hear from other folks

timebertt: One additional option would be to keep this PR backwards-compatible as it is, so we can cherry-pick it to `0.6` and `0.5`, and remove the deprecated types/funcs in a separate PR only targeting `>= 0.7`.

vincepri: Hm, not sure if we want to backport these changes, given that's a breaking behavior change (check one error vs another) and could impact some users

timebertt: Yeah, fair enough...

pkg/client/apiutil/dynamicrestmapper_test.go
```diff
@@ -57,53 +58,49 @@ var _ = Describe("Dynamic REST Mapper", func() {
...
+			count := 0
 			addToMapper = func(baseMapper *meta.DefaultRESTMapper) {
+				count++
 				baseMapper.Add(targetGVK, meta.RESTScopeNamespace)
-				baseMapper.Add(secondGVK, meta.RESTScopeNamespace)
 			}
-			Expect(callWithOther()).To(Succeed())
-			Expect(callWithTarget()).To(Succeed())
-		})
+			Expect(callWithOther()).To(beNoMatchError())
```
alvaroaleman, line 70: Why does this fail on the first attempt now and only works on the second?

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -31,6 +31,10 @@ import (
 
 // ErrRateLimited is returned by a RESTMapper method if the number of API
 // calls has exceeded a limit within a certain time period.
+//
+// Deprecated: as the dynamic RESTMapper now returns NoMatch errors directly
+// instead of RateLimited errors, it is meaningless and only kept for the sake
+// of backwards-compatibility.
 type ErrRateLimited struct {
```
alvaroaleman, line 38: Does it really make sense to keep this and the `DelayIfRateLimited` if we don't use it anymore? The only use case I can see for those is to handle ratelimit responses from the mapper but that won't work anymore, so removing them and make ppls code fail to compile if they use it seems the better option to me here. WDYT?

vincepri: See https://github.com/kubernetes-sigs/controller-runtime/pull/1151#discussion_r495996649 :)

pkg/client/apiutil/dynamicrestmapper.go
alvaroaleman: This PR is never going to be backwards compatible even if we keep the error type because that error type is never returned (which is btw much more subtile and annoying to debug than a compilation error). So I am very much in favor of just removing it.

pkg/client/apiutil/dynamicrestmapper_test.go
timebertt: It was also failing on the first attempt before (see [L62](https://github.com/kubernetes-sigs/controller-runtime/pull/1151/files/93a54481b9d2acb5231e6fd33663b1f9acf1ad4b#diff-515d07e6da6fc55b48a1766c7948ba61L62)) as the `secondGVK` is not present by default (see [L33](https://github.com/kubernetes-sigs/controller-runtime/pull/1151/files/93a54481b9d2acb5231e6fd33663b1f9acf1ad4b#diff-515d07e6da6fc55b48a1766c7948ba61L33))
I just moved down the first `callWithOther` and added `count` to ensure, that the RESTMapper really tries to rediscover for the call with the unknown resource.

timebertt:
@alvaroaleman @vincepri Thanks for your feedback. Makes sense to me.
I removed `ErrRateLimited` and `DelayIfRateLimited`, marked the PR/commit as ⚠️ and added a note to the PR description.
PTAL :)

timebertt:
How can I make `pull-controller-runtime-apidiff-master` happy?

alvaroaleman:
> How can I make pull-controller-runtime-apidiff-master happy?

You can't, because its a breaking change. The job is optional though.

/lgtm
/approve

Thanks for your work!

timebertt:
> > How can I make pull-controller-runtime-apidiff-master happy?
> 
> You can't, because its a breaking change. The job is optional though.

Ah, got it.
Thanks :)

PR #1153, opened 2020-09-08
:book: Fix typo

gogolok:
Reading the documentation for the leaderelection pkg I got confused whether the package provides one constructor or multiple constructors. Currently, I can only see one constructur, so hopefully the patch makes sense. Otherwise, please correct me and we can fix it in the other direction :-)


vincepri (approved):
/approve
/lgtm
/ok-to-test

PR #1163, opened 2020-09-15
🐛 Controller.Watch() should not store watches if already started

vincepri:
The controller internal struct holds a list of watches
(as []watchDescription) when someone calls .Watch() to then start the
watches and informers once we're ready to call Start().

This behavior caused a memory leak in the case Watch was called after
a controller has already been started and if the source.Kind's cache was
either stopped or not available any longer. The leak was caused by the
watches internal slice holding on to all references to each watch ever
issued (and their respective caches).

/assign @alvaroaleman @DirectXMan12 @pwittrock 

I'll backport the fix to release-0.5 and release-0.6 if we all agree that this is the fix we want. For consistency, I also nil'd the c.watches slice after Start() has been called and the controller has started.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -74,8 +74,8 @@ type Controller struct {
 
 	// TODO(community): Consider initializing a logger with the Controller Name as the tag
 
-	// watches maintains a list of sources, handlers, and predicates to start when the controller is started.
-	watches []watchDescription
+	// startWatches maintains a list of sources, handlers, and predicates to start when the controller is started.
+	startWatches []watchDescription
 
 	// Log is used to log messages to users during reconciliation, or for example when a watch is started.
 	Log logr.Logger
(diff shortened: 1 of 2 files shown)


pkg/internal/controller/controller.go
```diff
@@ -167,6 +168,9 @@ func (c *Controller) Start(stop <-chan struct{}) error {
 			}
 		}
 
+		// All the watches have been started, we can reset the local slice.
```
ncdc, line 171: I'd add more details here as to why it's bad to hold on to the watchDescriptions

pkg/internal/controller/controller.go
```diff
@@ -113,13 +113,14 @@ func (c *Controller) Watch(src source.Source, evthdler handler.EventHandler, prc
 		}
 	}
 
-	c.watches = append(c.watches, watchDescription{src: src, handler: evthdler, predicates: prct})
-	if c.Started {
-		c.Log.Info("Starting EventSource", "source", src)
-		return src.Start(evthdler, c.Queue, prct...)
+	// Controller hasn't started yet, store the watches locally and return.
```
ncdc, line 116: Clarify that we're storing the watches so they can all be started when the manager starts this controller

pkg/internal/controller/controller.go
vincepri: Done!

pkg/internal/controller/controller.go
vincepri: Done!

ncdc:
LGTM

alvaroaleman:
@vincepri are you sure this properly solves the problem? Even when the Controller doesn't hold the Source anymore, the underlying informer will have an eventhandler attached if was ever in use, preventing it from being garbage collected - or am I missing something?

/lgtm
/hold

ncdc:
@alvaroaleman we are dynamically creating a Cache and closing its stop channel when we're done with it, which GC's the informers.

vincepri:
What Andy said, if the stop channel is called, the informer properly stops, although the controller holds a reference to the KindSource which has a reference to the underlying cache, which will never be garbage collected

alvaroaleman:
okay, thanks for clarifying
/hold cancel

PR #1164, opened 2020-09-15
[0.5] :bug: Controller.Watch() should not store watches if already started

vincepri:
The controller internal struct holds a list of watches
(as []watchDescription) when someone calls .Watch() to then start the
watches and informers once we're ready to call Start().

This behavior caused a memory leak in the case Watch was called after
a controller has already been started and if the source.Kind's cache was
either stopped or not available any longer. The leak was caused by the
watches internal slice holding on to all references to each watch ever
issued (and their respective caches).

/assign @alvaroaleman 

/milestone v0.5.x


alvaroaleman (approved):
/lgtm

PR #1166, opened 2020-09-16
🐛 Fix builders WithLogger method

databus23:
The logger in the `Builder` struct is not used anywhere. Instead we need to set it in the controller options.

### DIFF
--- a/pkg/builder/controller_test.go
+++ b/pkg/builder/controller_test.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync/atomic"
 
+	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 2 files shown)


vincepri (approved):
/approve
/assign @alvaroaleman

alvaroaleman (commented):
@databus23 Could you add a test for this?

databus23:
@alvaroaleman Unfortunately I don't really have time to attend to this. I just noticed that `WithLogger` is doing nothing and am use `WithOptions` for now to set the logger.

databus23:
@alvaroaleman Looking at the tests, it should be fairly easy to add a test. Let me try to add a test for this.

pkg/builder/controller_test.go
```diff
@@ -160,6 +161,29 @@ var _ = Describe("application", func() {
...
+			m, err := manager.New(cfg, manager.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			instance, err := ControllerManagedBy(m).
+				For(&appsv1.ReplicaSet{}).
+				Owns(&appsv1.ReplicaSet{}).
+				WithLogger(logger).
+				Build(noop)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(instance).NotTo(BeNil())
```
alvaroaleman, line 184: That test will also succeed without your code changes - Please test that you get the same logger as you inserted, not that you get any logger

databus23: No it does not. I tested it. Without my changes it fails as i get back a `*zap.Logger` In that case.

databus23: While the test was working I made it more explicit by using a minimal custom `testLogger`.
Without my changes the test now fails with:
```
logger expected *builder.testLogger but found *zapr.zapLogger
```

alvaroaleman (approved):
/lgtm
/approve

PR #1167, opened 2020-09-16
fix gomodules.xyz/jsonpatch/v2 not found

cjhpaul:
404 on https://gomodules.xyz/jsonpatch/v2?go-get=1
Sample output below:
go: sigs.k8s.io/controller-runtime@v0.6.3 requires
        gomodules.xyz/jsonpatch/v2@v2.0.1: unrecognized import path "gomodules.xyz/jsonpatch/v2": reading https://gomodules.xyz/jsonpatch/v2?go-get=1: 404 Not Found
        server response: 404 page not found

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/googleapis/gnostic v0.5.1 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.10 // indirect
+	github.com/mattbaird/jsonpatch v0.0.0-20200820163806-098863c1fc24
 	github.com/onsi/ginkgo v1.14.0
 	github.com/onsi/gomega v1.10.1
 	github.com/prometheus/client_golang v1.7.1
@@ -18,7 +19,6 @@ require (
 	go.uber.org/goleak v1.1.10
 	go.uber.org/zap v1.15.0
 	golang.org/x/time v0.0.0-20200630173020-3af7569d3a1e
-	gomodules.xyz/jsonpatch/v2 v2.1.0
 	google.golang.org/appengine v1.6.6 // indirect
 	k8s.io/api v0.19.0
 	k8s.io/apiextensions-apiserver v0.19.0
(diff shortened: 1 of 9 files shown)


vincepri:
Why is this needed? I ran `make modules` yesterday and was able to download and run all the modules in this repository

cjhpaul:
> Why is this needed? I ran `make modules` yesterday and was able to download and run all the modules in this repository

Yeah, I ran yesterday with no issue. I just noticed it just now. Hopefully it's transient but I don't see jsonpatch listed under https://gomodules.xyz/ anymore.

jsing157:
`export GOPROXY=https://proxy.golang.org` should fix it if you add this to your go env

vincepri:
Maybe @tamalsaha can help here?

cjhpaul:
> `export GOPROXY=https://proxy.golang.org` should fix it if you add this to your go env

I feel like it's a workaround.

vincepri:
It seems the package is now called `kmodules.xyz/jsonpath`

alvaroaleman:
Also its not clear if this is the same code. There is no 2.1.0 tag on mattbairds repo

tamalsaha:
https://github.com/gomodules/jsonpatch/tree/release-2.0

jsing157:
@cjhpaul Can you explain why it is a workaround? I do not understand. Thanks!

cjhpaul:
> @cjhpaul Can you explain why it is a workaround? I do not understand. Thanks!

I think using goproxy is a good option to consider. But I think ideally it would have to build fine without using the proxy, which made me think we would have to either revive the broken path (gomodules.xyz/jsonpatch/v2) or find an alternative package. Of course, there could be a better option.
For me, I can probably live with goproxy along with goprivate since I also have private modules.
Hope that answers your question.

Also, it turned out it's working fine as of now. It looks like an intermittent issue, so I will close this.
Thanks all for taking time on this matter.

mnencia:
> Hi,
> The code is where was https://github.com/gomodules/jsonpatch/tree/release-2.0
> 
> We have not changed anything.
> 
> Also the package is available here: https://pkg.go.dev/mod/gomodules.xyz/jsonpatch/v2?tab=versions
> 
> This seems like a issue at OP's network, I guess?

The problem is that gomodules.xyz reply with the content of kmodules.xyz
This is probably due to some misconfiguration of the serving host.

tamalsaha:
I have reployed the app engine config. It should be fixed now. Sorry about the mix up.

PR #1174, opened 2020-09-24
:sparkles: metadata-only watches

### LINKED ISSUE #1159: Metadata-only list/watch support
The kube apiserver supports listing and watching *just* the metadata for resources.  We should expose support for this both in our watch machinery and in our client.

This is particularly useful for performance when watching large numbers of different resources, or when you're watching a particularly numerous resource but need only the metadata for most/all instances of the resource.

/kind feature
/priority important-longterm

DirectXMan12:
This introduces metadata-only support (PartialObjectMetadata) to the client, cache, and controller builder.  It works much like unstructured does, with an end-use of:

```go
.Owns(&corev1.Pod{}, builder.OnlyMetadata)
// or
.Owns(&metav1.PartialObjectMetadata{/* GVK for corev1.Pod filled out here */})

// ...

var pods metav1.PartialObjectMetadataList
pods.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("PodList"))
cl.List(ctx, &pods, client.MatchingLabels{...})
```

Note that like unstructured, the caches are *completely* separate -- asking for an unstructured object or concrete object will end up constructing a separate cache.

Fixes #1159

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -21,8 +21,10 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -36,6 +38,17 @@ import (
 // Supporting mocking out functions for testing
 var newController = controller.New
 var getGvk = apiutil.GVKForObject
+
+// project represents other forms that the we can use to
+// send/receive a given resource (metadata-only, unstructured, etc)
+type objectProjection int
+
+const (
+	// projectAsNormal doesn't change the object from the form given
+	projectAsNormal objectProjection = iota
+	// projectAsMetadata turns this into an metadata-only watch
+	projectAsMetadata
+)
 
 // Builder builds a Controller.
 type Builder struct {
@@ -57,9 +70,10 @@ func ControllerManagedBy(m manager.Manager) *Builder {
 
 // ForInput represents the information set by For method.
 type ForInput struct {
-	object     client.Object
-	predicates []predicate.Predicate
-	err        error
+	object           client.Object
+	predicates       []predicate.Predicate
+	objectProjection objectProjection
+	err              error
 }
 
 // For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
@@ -82,8 +96,9 @@ func (blder *Builder) For(object client.Object, opts ...ForOption) *Builder {
 
 // OwnsInput represents the information set by Owns method.
 type OwnsInput struct {
-	object     client.Object
-	predicates []predicate.Predicate
+	object           client.Object
+	predicates       []predicate.Predicate
+	objectProjection objectProjection
 }
 
 // Owns defines types of Objects being *generated* by the ControllerManagedBy, and configures the ControllerManagedBy to respond to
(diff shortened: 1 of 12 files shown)


DirectXMan12:
/assign @vincepri 

since he expressed interest

vincepri:
/milestone v0.7.x

Will take a look tomorrow :)

pkg/builder/controller.go
```diff
@@ -177,19 +192,44 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
...
+	typeForSrc, err := blder.project(blder.forInput.object, blder.forInput.objectProjection)
+	if err != nil {
+		return err
+	}
+	src := &source.Kind{Type: typeForSrc}
 	hdler := &handler.EnqueueRequestForObject{}
 	allPredicates := append(blder.globalPredicates, blder.forInput.predicates...)
-	err := blder.ctrl.Watch(src, hdler, allPredicates...)
+	err = blder.ctrl.Watch(src, hdler, allPredicates...)
 	if err != nil {
```
vincepri, line 222: Nit
```suggestion
	if err := blder.ctrl.Watch(src, hdler, allPredicates...); err != nil {
```

pkg/client/client.go
```diff
@@ -204,19 +238,25 @@ var _ StatusWriter = &statusWriter{}
...
 // Patch implements client.Client
 func (sw *statusWriter) Patch(ctx context.Context, obj runtime.Object, patch Patch, opts ...PatchOption) error {
 	defer sw.client.resetGroupVersionKind(obj, obj.GetObjectKind().GroupVersionKind())
-	_, ok := obj.(*unstructured.Unstructured)
-	if ok {
+	switch obj.(type) {
+	case *unstructured.Unstructured:
 		return sw.client.unstructuredClient.PatchStatus(ctx, obj, patch, opts...)
+	case *metav1.PartialObjectMetadata:
+		return sw.client.metadataClient.PatchStatus(ctx, obj, patch, opts...)
```
vincepri, line 258: How would patching the status on a partial metadata object work in this case? 🤔

pkg/client/metadata_client.go
```diff
@@ -0,0 +1,194 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/metadata"
+)
+
+// TODO(directxman12): we could rewrite this on top of the low-level REST
+// client to avoid the extra shallow copy at the end, but I'm not sure it's
+// worth it -- the metadata client deals with falling back to loading the whole
+// object on older API servers, etc, and we'd have to reproduce that.
```
vincepri, line 34: :+1:

pkg/client/client.go
DirectXMan12: metadata is modifiable on status...

... except:

| Kind | Fields |
|--------|---------|
| Deployments | Labels |
| CRD | Labels, Annotations, OwnerReferences, Generation, SelfLink |
| APIService | Labels, Annotations, Finalizers, OwnerReferences |

which is uuuhh not the most internally consistent.

DirectXMan12: (FWIW, the client-go client allows it, which is mainly why I added it here)

vincepri: Should we consider not allowing patching status fields on partial objects to avoid confusion?

vincepri:
@DirectXMan12 Are you able to rebase so we can get this merged in v0.7.x?

DirectXMan12:
/retest

shutdown data race, still working on a fix

DirectXMan12:
@vincepri should be good to merge

DirectXMan12:
huh, that's actually a different test race this time.  Will produce a fix, just a test race.

pkg/builder/controller.go
```diff
@@ -21,9 +21,11 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
```
vincepri, line 28: Remove this?

pkg/client/client.go
```diff
@@ -21,12 +21,15 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/rest"
+
```
vincepri, line 32: Remove this too?

pkg/client/client.go
vincepri: Do we still want to do this?

pkg/builder/controller.go
DirectXMan12: whoops, yeah, bad merge

pkg/client/client.go
DirectXMan12: that's actually correct -- imports sections are

```
import (
  <stdlib>

  <other-libs>

  <other-things-in-this-lib>
)
```

PR #1176, opened 2020-09-24
:sparkles: Add error check for multiple apiTypes as reconciliation object

### LINKED ISSUE #1173: Calling Builder.For() multiple times loses previous apiTypes
At https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L74 , For latches the apiType of the For() call.

If a controller-runtime consumer calls Builder.For(foo).For(bar), then foo will be silently lost in favour of bar.

This contrasts with Owns' behaviour, where Builder.Owns(foo).Owns(bar) will append both foo and bar to an internal slice.

Judging by the code comment on For (`For defines the type of Object being *reconciled*`), it may not be intended for For to be called multiple times on a Builder.  Nonetheless, the current behaviour can be confusing.

As a workaround, a consumer who wants to watch multiple apiTypes can use Watches(foo).Watches(bar) instead.

IMO it would be nice to make the behaviour here clearer, either:
* by raising an error if For() is called and bldr.apiType is already set, or
* by making apiTypes a slice and allowing For() to be called multiple times, as Owns() or Watches() is.

prafull01:
Added error checks for the multiple calls of For() on the apiType. Currently, it silently ignores the first apiType and take the second apiType as reconciliation object. This should throw an error that multiple apiType are not allowed for reconciliation.

Fixes #1173

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -59,6 +59,7 @@ func ControllerManagedBy(m manager.Manager) *Builder {
 type ForInput struct {
 	object     runtime.Object
 	predicates []predicate.Predicate
+	err        error
 }
 
 // For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
@@ -66,6 +67,10 @@ type ForInput struct {
 // This is the equivalent of calling
 // Watches(&source.Kind{Type: apiType}, &handler.EnqueueRequestForObject{})
 func (blder *Builder) For(object runtime.Object, opts ...ForOption) *Builder {
+	if blder.forInput.object != nil {
+		blder.forInput.err = fmt.Errorf("For(...) should only be called once, could not assign multiple objects for reconciliation")
+		return blder
+	}
 	input := ForInput{object: object}
 	for _, opt := range opts {
 		opt.ApplyToFor(&input)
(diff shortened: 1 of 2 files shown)


pkg/builder/controller.go
```diff
@@ -59,13 +59,18 @@ func ControllerManagedBy(m manager.Manager) *Builder {
...
+	err        error
 }
 
 // For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete /
 // update events by *reconciling the object*.
 // This is the equivalent of calling
 // Watches(&source.Kind{Type: apiType}, &handler.EnqueueRequestForObject{})
 func (blder *Builder) For(object runtime.Object, opts ...ForOption) *Builder {
+	if blder.forInput.object != nil {
+		blder.forInput.err = fmt.Errorf("could not assign multiple apiType for reconciliation")
```
vincepri, line 71:
```suggestion
		blder.forInput.err = fmt.Errorf("For(...) should only be called once, could not assign multiple objects for reconciliation")
```

vincepri (commented):
One minor suggestion, the rest looks good to me

/approve

pkg/builder/controller.go
prafull01: Done

vincepri (approved):
/lgtm

PR #1180, opened 2020-09-25
[WIP] :seedling: Controller Lifecycle Management

### LINKED ISSUE #1179: Support running controller managers for CRDs not installed
### Summary
Enable controller managers to successfully operate when the CRD the controller is configured to watch does not exist.

Successful operation of a controller manager includes:
Starts and runs without error when a CRD the controller watches does not exist.
Begins watching the CRD once it is installed
Unregisters (stop watching) once a CRD is uninstalled.

### Background
Usually there is a 1:1 relationship between controller and resource and a 1:1 relationship between controller and k8s cluster. When this is the case it is fine to assume that for a controller to run successfully, the resource it controls must be installed on the cluster that the controller is watching.

We are now seeing use cases where a single controller is responsible for multiple resources across multiple clusters. This creates situations where controllers need to successfully run even when the resource is not installed, and need to proceed successfully as it begins/terminates its watch on a resource when the resource is installed/uninstalled.

kevindelgado:
This is a proof of concept PR for the Controller Lifecycle Management design doc at #1192.

It is:
1. A rebase of shomron’s PR at https://github.com/kubernetes-sigs/controller-runtime/pull/936 to enable removal of individual informers that is pre-requisite for the other goals.
2. A minimal set of hooks needed to enable a user to conditionally start/stop/restart controllers on their own.
3. An implementation of EventHandler reference counting on the Informer.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -126,7 +126,7 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 	// Construct a new Mapper if unset
 	if opts.Mapper == nil {
 		var err error
-		opts.Mapper, err = apiutil.NewDiscoveryRESTMapper(config)
+		opts.Mapper, err = apiutil.NewDynamicRESTMapper(config)
 		if err != nil {
 			log.WithName("setup").Error(err, "Failed to get API Group-Resources")
 			return opts, fmt.Errorf("could not create RESTMapper from config")
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -58,6 +58,9 @@ type Informers interface {
 	// of the underlying object.
 	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind) (Informer, error)
 
+	// Remove the informer for the given object.
+	Remove(ctx context.Context, obj runtime.Object) error
+
 	// Start runs all the informers known to this cache until the context is closed.
 	// It blocks.
 	Start(ctx context.Context) error
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -87,6 +87,14 @@ type Informer interface {
 	AddIndexers(indexers toolscache.Indexers) error
 	//HasSynced return true if the informers underlying store has synced
 	HasSynced() bool
+
+	// RemoveEventHandler currently just decrements a the count of event handlers
+	// The goals it to have SharedInformer support RemoveEventHandler (and actually remove
+	// the handler instead of just decrementing a count).
+	RemoveEventHandler(id int) error
+
+	// CountEventHandlers returns the number of event handlers added to an informer.
+	CountEventHandlers() int
 }
 
 // Options are the optional arguments for creating a new InformersMap object
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -243,6 +243,5 @@ func (ip *informerCache) Remove(ctx context.Context, obj runtime.Object) error {
 		return err
 	}
 
-	ip.InformersMap.Remove(gvk, obj)
-	return nil
+	return ip.InformersMap.Remove(gvk, obj)
 }
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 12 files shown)


kevindelgado:
Adding WIP, recent change to prevent controllers from starting more than once (#1139) breaks this. Will fix.

pkg/manager/internal.go
```diff
@@ -582,12 +606,84 @@ func (cm *controllerManager) startLeaderElectionRunnables() {
...
 		// Controllers block, but we want to return an error if any have an error starting.
 		// Write any Start errors to a channel so we can return them
-		cm.startRunnable(c)
+		cm.startRunnable(c, cm.internalStop)
 	}
 
 	cm.startedLeader = true
 }
 
+func (cm *controllerManager) startConditionalRunnables() {
```
alvaroaleman, line 615: hi @kevindelgado , thanks for this PR. It seems to be its a bit chaotic and involves a set of semi-related changes, like the removal of informers which is IIRC not something we have agreed on and definitely something for a distinct PR.

As for the `conditionanally runnables` idea, a much simpler approach would be to just check whatever you want to check before adding them to the `manager`:
```
if clusterHasMyCRD(){
    controller.New()
}
```

This also has the advantage that you can check anything, not whatever happens to be implemented for this in the `manager`.

kevindelgado: Thanks for the feedback! 

We’re definitely open to better ways of doing this, but we are currently doing what you suggest. Outside of controller-runtime we conditionally start the manager based on the existence of the CRD. A couple issues we see are:

1. Increasing complexity. Specifically, when multiple controllers are abstracted to appear as a single controller or when the manager is running multiple controllers that watch a mixture of CRDs that will always be installed on the cluster, but also CRDs that are only sometimes installed. 
2. This approach still isn’t great for fluidly handling CRD install/uninstall/reinstall (rather than just waiting for install before starting), especially when latency becomes important immediately upon CRD installation.

/cc @DirectXMan12  in case they can better explain our desire to push this functionality further down the stack into c-r.

kevindelgado:
Fixed the rebasing issues caused by not being able to restart controllers.

pkg/manager/internal.go
alvaroaleman: This sounds to me like you should probably build a metacontroller that watches crds and then starts and stops the actual controllers. 

If you continue to think this requires changes in controller-runtime, the first step would be a design doc that describes the use-case, why this is problematic with how controller-runtime works currently, what changes would be required and possible alternatives so we can properly discuss this. We have the /designs folder for this.

kevindelgado: Sure thing, we’re working to create a more compelling story for this and will have a design doc up if/when that is the case. 

Sorry for the lack of context here, when learning the codebase it was easier to express ideas through concrete code than in prose initially. 

While the code has been helpful to steer discussions and now make it easier to formalize into a design doc, I definitely understand that it appears chaotic and disorganized externally. Feel free to close if you want to reduce clutter in the repo, I can always submit a new one if necessary.

pkg/manager/internal.go
```diff
@@ -611,12 +636,91 @@ func (cm *controllerManager) startLeaderElectionRunnables() {
...
+						close(presentStop)
+						if err := cm.cache.Remove(obj); err != nil {
+							log.Error(err, "could not remove object from cache")
+						}
+
+						controller, ok := c.(*intctrl.Controller)
+						if !ok {
+							log.Error(errors.New("runnable is not a controller"), "")
+						}
+						controller.Started = false
```
DirectXMan12, line 701: this is not a safe operation, IIRC

pkg/manager/internal.go
DirectXMan12: I think we can probably move a lot of this into a separate helper runnable that manages its own sub-runnables and managed the start/stop of things.  We'd still need a way to start/stop caches though, and possible some other minor tweaks.

I think the value of having this *somewhere* in controller-runtime is that if folks want to do something like this, they don't have to re-implement from scratch, since there are some corner cases.  I've seen it pop up a few times from various users on our side, and at the very least I think it'd be useful to have some basic hooks for better being able to start/stop controllers in CR, especially for multi-cluster usecases, where you don't necessarily want to/can't degrade performance for all clusters just to deal with a change in one. It also provides a motivating minimal usecase for starting/stopping caches.

DirectXMan12: I think one of the things to keep in mind here is that if we go with the approach of "oh, you should just build it as a meta-controller", we need to make sure the hooks are there to do it cleanly, and we need to make sure those hooks are documented as such so that they don't get removed later on down the road.

alvaroaleman: > At the very least I think it'd be useful to have some basic hooks for better being able to start/stop controllers in CR, especially for multi-cluster usecases, where you don't necessarily want to/can't degrade performance for all clusters just to deal with a change in one. It also provides a motivating minimal usecase for starting/stopping caches

Sure thing, but today we don't have any official story at all for multi-cluster _without the assumption that clusters might come and go_. I am not objecting to the basic idea (although I still think that if this is about being fast as mentioned above, a metacontroller would be the best), I just think that this requires a lot of groundwork first.

kevindelgado:
Here’s a major overhaul of the previous approach that hopefully is much less invasive to the existing code that goes along with the design doc at #1192. Gonna bring this up for discussion at the next kubebuilder meeting. 

If you can take a look at each commit individually, hopefully it comes across as less chaotic.

The second one is pretty minimal and is the only change strictly necessary to solving our problem (the second and third are optional and less fleshed out, the first is just a direct rebase and copy of #936).

The hope is that we can find some part of this small enough to move forward with, without first having a great story around multi-cluster.

@alvaroaleman @DirectXMan12

kevindelgado:
I've distilled this PR into just the minimal changes to C-R necessary to enable better controller lifecycle management. 

The POC for conditional controllers is on this branch https://github.com/kevindelgado/controller-runtime/tree/experimental/conditional-runnables, but it was clear that there's plenty to discuss just on allowing informer removal and the minimal hooks needed use informer removal to start/stop/restart controllers cleanly.

The design doc at #1192 has also been updated to reflect this.

kevindelgado:
Updated to reflect updates to the doc #1192, PTAL anyone interested

PR #1182, opened 2020-09-26
:bug: add error check when no apiType is provided for reconciliation

### LINKED ISSUE #1181: Builder causes panics when For isn't called
Currently the causes builder panics when `For` isn't called, because it constructs a `Kind` source with a nil `Type` field.  We should detect this in the builder's `doWatch` method and return an actual error instead.

/kind bug
/priority important-longterm
/good-first-issue

prafull01:
When no reconciliation type is provided in the For(...) function to the Builder, it used to panic. Added the error handling if no apiType is provided in the For(...) function.

Fixes #1181

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -167,6 +167,10 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
 	if blder.forInput.err != nil {
 		return nil, blder.forInput.err
 	}
+	// Checking the reconcile type exist or not
+	if blder.forInput.object == nil {
+		return nil, fmt.Errorf("must provide an object for reconciliation")
+	}
 
 	// Set the Config
 	blder.loadRestConfig()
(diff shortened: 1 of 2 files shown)


pkg/builder/controller.go
```diff
@@ -167,6 +167,10 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
 	if blder.forInput.err != nil {
 		return nil, blder.forInput.err
 	}
+	// Checking the reconcile type exist or not
+	if blder.forInput.object == nil {
+		return nil, fmt.Errorf("must provide an object for reconciliation")
```
alvaroaleman, line 172: please use `errors.New` for error messages without a format string

prafull01: @alvaroaleman In the same func, fmt.Errorf is used for other errors which do not have a format string. IMHO this needs to be fixed horizontally for all such error messages.

alvaroaleman (approved):
/lgtm
/approve

PR #1187, opened 2020-09-28
:seedling: Update Go mod version to 1.15

### LINKED ISSUE #825: Upgrade to Go 1.15?
As title says, Go 1.14 is out with great changes, we might want to consider the path forward to potentially upgrade

vincepri:

Fixes #825
/milestone v0.7.x
/assign @alvaroaleman

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.13
+go 1.15
 
 require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime/hack/tools
 
-go 1.13
+go 1.15
 
 require (
 	github.com/golangci/golangci-lint v1.23.6


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.13
+go 1.15
```
alvaroaleman, line 3: The presubmits still run with go 1.13, so we should probably change that first: https://github.com/kubernetes/test-infra/blob/f6e9dbfd16c179a89c84f98481326884ca69bcc4/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-master.yaml#L12

and why does specifying a newer go version than used in the tests not make the tests fail?

vincepri: From https://golang.org/doc/go1.12#modules:

> The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.

vincepri: I'll open a PR in test-infra to update the version

vincepri: https://github.com/kubernetes/test-infra/pull/19383

alvaroaleman (approved):
/lgtm

vincepri:
/retest

Not sure why now the tests are failing, hope it's not related, I've been testing locally with 1.15 for a long while

PR #1189, opened 2020-09-29
:bug: Fix the unserved version CRD install timeout in envtest

prafull01:
Add fix in the waitForCRD function not to add the default version for polling (wait for it to become ready) if the multiple versions are present. If there are no versions field in CRD then add the default version else check the serve condition before adding each version from versions field.


pkg/envtest/envtest_test.go
```diff
@@ -81,6 +81,44 @@ var _ = Describe("Test", func() {
...
+								Storage: false,
+								Served:  false,
+							},
+						}},
+				},
+			},
+				CRDInstallOptions{MaxTime: 50 * time.Millisecond, PollInterval: 15 * time.Millisecond},
+			)
+			Expect(err).NotTo(HaveOccurred())
+		})
```
vincepri, line 121: Seems this should be gofmt'd?

prafull01: @vincepri Thanks for the review. The file is formatted properly. The issue was the fix is using `runtime.Object` which is replace by `client.Object`. I have rebase the branch and updated the fix with `client.Object`.

pkg/envtest/crd.go
```diff
@@ -127,14 +127,17 @@ func WaitForCRDs(config *rest.Config, crds []client.Object, options CRDInstallOp
...
-
-		versions, _, err := unstructured.NestedSlice(crd.Object, "spec", "versions")
+		versions, found, err := unstructured.NestedSlice(crd.Object, "spec", "versions")
 		if err != nil {
 			return err
 		}
+
+		// gvs should be added here only if single version is found. If multiple version is found we will add those version
+		// based on the version is served or not.
+		if crdVersion != "" && !found {
```
DirectXMan12, line 137: nit: (don't consider this blocking, but if for some reason you have to do another push, please fix): make `found` more descriptive, since it lasts more than a line.  For example, `isMultiVersion`

DirectXMan12 (changes requested):
EDIT: nevermind, I'm reviewing too fast apparently, sorry about that.  Looks good

DirectXMan12 (approved):

PR #1192, opened 2020-10-01
:seedling: Initial ConditionalControllers design doc

kevindelgado:
Design doc for conditional controllers

### DIFF
--- a/designs/conditional-controllers.md
+++ b/designs/conditional-controllers.md
@@ -1,0 +1,204 @@
+Conditionally Runnable Controllers
+==========================
+
+## Summary
+
+Enable controller managers to successfully operate when the CRD the controller
+is configured to watch does not exist.
+
+Successful operation of a controller manager includes:
+* Starts and runs without error when a CRD the controller watches does not exist.
+* Begins watching the CRD once it is installed.
+* Unregisters (stops watching) once a CRD is uninstalled.
+
+## Background/Motivation
+
+Usually there is a 1:1 relationship between controller and resource and a 1:1
+relationship between controller and k8s cluster. When this is the case it is
+fine to assume that for a controller to run successfully, the resource it
+controls must be installed on the cluster that the controller is watching.
+
+We are now seeing use cases where a single controller is responsible for
+multiple resources across multiple clusters. This creates situations where
+controllers need to successfully run even when the resource is not installed,
+and need to proceed successfully as it begins/terminates its watch on a resource
+when the resource is installed/uninstalled.
+
+The current approach is to check the discovery for a CRDs existence prior to
+adding the controller to the manager. This has its limitations, as complexity
+increases greatly for users who need to manage multiple controllers for a
+mixture of CRDs that might not always be installed on the cluster or are
+installed in different order.
+
+While there is a lot of groundwork needed to fully support multi-cluster
+controllers, this proposal offers incremental steps to supporting one specific use case
+(conditionally runnable controllers), of which hopefully some minimal form can be agreed
+upon before needing a comple multi-cluster story.
+
+## Goals
+
+(In order from most necessary to least)
+1. A mechanism for starting/stopping/restarting controllers and their caches.
+2. A solution for running a controller conditionally, such that it automatically
+starts/stops/restarts upon installation/uninstallation/reinstallation of its
+respective CRD in the cluster.
+3. An easy-to-use mechanism for solving goal #2 without the end user needing to
+understand too much.
+
+## Non-Goals
+
+TODO
+
+## Proposal
+
+The following proposal attempts to address the three goals in four separate
+steps.
+
+The first goal of enabling the possibility of starting/stopping/restarting
+controllers is addressed in two parts, first where a mechanism is built to
+enable the removal of informers, and second where we expose this mechanism and
+enable the restarting of controllers
+
+The second goal of a solution for running controllers conditionally is addressed
+by creating a wrapper around a controller called a ConditionalController that
+within its Start() function, it polls the discovery doc for the existence of the
+watched object and starts/stops/restarts the underlying controller as necessary 
+
+The third goal of an ergonomic mechanism to use ConditionalControllers is a
+small modification to the controller builder to enable running a controller as a
+ConditionalController.
+
+###### Proof of concept
+A proof of concept PR exists at
+[#1180](https://github.com/kubernetes-sigs/controller-runtime/pull/1180)
+
+Each commit maps to one step in the proposal and can loosely be considered going
+from most necessary to least.
+
+### Informer Removal [Pre-req]
+
+This proposal assumes a mechanism exists for removing individual informers. We
+are agnostic to how this is done, but the current proposal is built on top of
+Shomron’s proposed implementation of individual informer removal
+[#936](https://github.com/kubernetes-sigs/controller-runtime/pull/936).
+
+
+Risks/mitigations and alternatives are discussed in the linked PR as well as the
+corresponding issue
+[#935](https://github.com/kubernetes-sigs/controller-runtime/issues/935). 
+
+We are currently looking into whether that PR can be distilled further into a more
+minimal solution.
+
+### Minimal hooks
+
+Given that a mechanism exists to remove individual informers, the next step is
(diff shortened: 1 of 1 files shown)


designs/conditional-controllers.md
```diff
@@ -0,0 +1,204 @@
...
+unclear what effect explicitly enabling multiple starts in the case of
+conditional controllers will hae on the number of workers and workqueues
+relative to expectations and metrics.
+* [#1163](https://github.com/kubernetes-sigs/controller-runtime/pull/1163) discusses the
+memory leak caused by no clearing out the watches internal slice. A possible
+mitigation is to clear out the slices upon ConditionalController shutdown.
+
+#### Alternatives
+
+* Instead of exposing ResetStart and SaveWatches on the internal controller struct
```
alvaroaleman, line 119: Can we mention the option of a CRD-controller here that starts/stops other controllers based on the presence/absence of crds?

kevindelgado: done, lemme if I've captured the pros/cons accurately

designs/conditional-controllers.md
```diff
@@ -0,0 +1,209 @@
...
+conditional controllers will hae on the number of workers and workqueues
+relative to expectations and metrics.
+* [#1163](https://github.com/kubernetes-sigs/controller-runtime/pull/1163) discusses the
+memory leak caused by no clearing out the watches internal slice. A possible
+mitigation is to clear out the slices upon ConditionalController shutdown.
+
+#### Alternatives
+
+* A metacontroller or CRD controller could start and stop controllers based on
+the existence of their corresponding CRDs. This requires no changes to made to
```
DirectXMan12, line 120: I think the "meta controller" strategy still requires minimal hooks.

designs/conditional-controllers.md
```diff
@@ -0,0 +1,174 @@
...
+* [#1163](https://github.com/kubernetes-sigs/controller-runtime/pull/1163) discusses the
+memory leak caused by no clearing out the watches internal slice. A possible
+mitigation is to clear out the slices upon ConditionalController shutdown.
+
+#### Alternatives
+
+* A metacontroller or CRD controller could start and stop controllers based on
+the existence of their corresponding CRDs. This puts the complexity of designing such a controller
+onto the end user, but there are potentially ways to provide end users with
+default, pluggable CRD controllers. More importantly, this probably is not even
```
alvaroaleman, line 143: I think the informer removal doesn't really have anything to do with metacontroller or not, its a requirement regardless of how the add/removal functionality is implemented.

The reason I personally like the metacontroller is that it is event-based, rather than doing polling.

designs/conditional-controllers.md
```diff
@@ -0,0 +1,174 @@
+Informer Removal / Controller Lifecycle Management
+==========================
+
+## Summary
+
+Enable fine-grained control over the lifecycle of a controller, including the
+ability to start/stop/restart controllers and their caches by exposing a way to
+remove individual informers from the cache and working around restrictions that
+currently prevent controllers from starting multiple times.
```
alvaroaleman, line 9: Is supporting this in multi-cluster scenarios a goal?

designs/conditional-controllers.md
```diff
@@ -0,0 +1,174 @@
...
+
+Greater detail is given in the "Future Work / Use Cases" section, but a summary
+of the motivating use-cases for the proposal is anytime a controller's watched resource
+can be installed/uninstalled unexpectedly or whenever the administrator of a
+controller is different from the administrator of the cluster (and thus has no
+control over which resources are installed).
+
+## Goals
+
+An implementation of the minimally viable hooks needed in controller-runtime to
```
alvaroaleman, line 30: Is an actual implementation of those hooks also a goal or just to provide a hook?

designs/conditional-controllers.md
```diff
@@ -0,0 +1,174 @@
...
+enable controller adminstrators to start, stop, and restart controllers and their caches.
```
alvaroaleman, line 31: non-goals says that you don't want to add support to start/stop on arbitrary conditions. Please explicitly mention the conditions this is supposed to support

designs/conditional-controllers.md
```diff
@@ -0,0 +1,174 @@
...
+   stopped.
+
+This proposal focuses on solutions that are entirely contained in
+controller-runtime. In the alternatives section, we discuss potential ways that
+changes can be added to api-machinery code in core kubernetes to enable a
+possibly cleaner interface of accomplishing our goals.
+
+### Informer Removal
+
+The starting point for this proposal is Shomron’s proposed implementation of
```
alvaroaleman, line 53: Please describe this a bit, rahter than just saying "This PR explains it", because the PR might change

designs/conditional-controllers.md
kevindelgado: Just providing a hook for now.

alvaroaleman: But that then achieves your non-goal of supporting start/stop on arbitrary conditions, right?

designs/conditional-controllers.md
kevindelgado: Sorry, I don't think the non-goals are clear, basically I was just trying to say that the goals are just to provide a hook, non-goals are an actual implementation of the hooks.

designs/conditional-controllers.md
kevindelgado: _Supporting_ start/stop on arbitrary conditions is the goal (which I believe is accomplished just by implementing a mechanism to remove informers and rerun controllers).
 
_Implementing_ a solution (the actual polling mechanism that does the start/stopping such as a metacontroller or conditional controller) is not the goal.

designs/conditional-controllers.md
kevindelgado: That's fair, I think I was a bit confused on this point. I agree that metacontroller is not an alternative and will update the doc to reflect that.

A discussion of the metacontroller probably belongs in the "Future works / use cases" section rather than in the proposal which is more focused on providing the capability to externally build the metacontroller outside of c-r.

designs/conditional-controllers.md
alvaroaleman: okay perfect, that wasn't clear to me before. And agreed with https://github.com/kubernetes-sigs/controller-runtime/pull/1192#discussion_r501871400 that metacontroller is not an alternative but more of a follow-up

designs/conditional-controllers.md
kevindelgado: My understanding is that there's isn't anything explicitly preventing a single-cluster solution from working for multiple clusters out of the box. It's sounds like you have specific concerns around this approach not working multi-cluster.

It's just that because controller-runtime historically has not attempted to support multi-cluster, attempting to use any solution that comes from this effort in a multi-cluster manner would be harder and undocumented.

I think the short answer is no, we aren't looking to fully flesh out a multi-cluster solution here but @DirectXMan12 can chime in if I'm misspeaking

alvaroaleman: I guess if this is a generic question, it doesn't really matter what cluster a given controller talks to (But I can't mark this convo as resolved)

kevindelgado:
I took a pass at reorganizing, expanding, and addressing feedback on this design.

Main updates:
1. Added tracking of event handlers to the proposal to address the feedback about silently degrading (https://github.com/kubernetes-sigs/controller-runtime/pull/936#issuecomment-625232121)
2. Added a discussion of what we could potentially ask api-machinery/client-go for in terms of updates to the SharedInformer interface. I'm working on a proposal and POC for that now, not sure if it'll be ready to present at the api-machinery meeting this wednesday, but that's my goal right now.

proof-of-concept updated at #1180 

I think I've addressed all your feedback from last time @alvaroaleman, let me know if I've missed anything.

PTAL
cc @DirectXMan12 @shomron
cc

designs/conditional-controllers.md
kevindelgado: done

kevindelgado:
Dusting the cobwebs off this with a new WIP implementation at https://github.com/kubernetes-sigs/controller-runtime/pull/1527

PR #1194, opened 2020-10-01
🌱 Expose handler.MapFunc to allow implementers to test mappers

vincepri:
This is a follow-up from a previous change to the handler package that
deprecated the old handler.ToRequestFunc in favor of
EnqueueRequestsFromMapFunc.

While we still want to prefer using EnqueueRequestsFromMapFunc, this
change also exposes the type of MapFunc signature expected. Users can
use this type to test their own mappers in unit tests.

### DIFF
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -24,6 +24,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 )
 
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm
/approve

aoxn:
@vincepri @alvaroaleman  Hi, I noticed that the mapper struct has been removed.  I want to known how could i inject client into my mapper function use `EnqueueRequestsFromMapFunc` or `MapFunc`?

PR #1195, opened 2020-10-01
⚠ Introduce and use client.Object and client.ObjectList

vincepri:
This change introduces two interfaces: client.Object and
client.ObjectList.

These interfaces are now used throughout the codebase, more specifically
in the client.

/milestone v0.7.x

### DIFF
--- a/pkg/cache/informer_cache_unit_test.go
+++ b/pkg/cache/informer_cache_unit_test.go
@@ -25,12 +25,6 @@ var _ = Describe("ip.objectTypeForListObject", func() {
 		InformersMap: &internal.InformersMap{Scheme: scheme.Scheme},
 	}
 
-	It("should error on non-list types", func() {
-		_, _, err := ip.objectTypeForListObject(&corev1.Pod{})
-		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(Equal(`non-list type *v1.Pod (kind "/v1, Kind=Pod") passed as output`))
-	})
-
 	It("should find the object type for unstructured lists", func() {
 		unstructuredList := &unstructured.UnstructuredList{}
 		unstructuredList.SetAPIVersion("v1")
--- a/pkg/client/typed_client.go
+++ b/pkg/client/typed_client.go
@@ -22,6 +22,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var _ Reader = &typedClient{}
+var _ Writer = &typedClient{}
+var _ StatusWriter = &typedClient{}
+
 // client is a client.Client that reads and writes directly from/to an API server.  It lazily initializes
 // new clients at the time they are used, and caches the client.
 type typedClient struct {
@@ -30,7 +34,7 @@ type typedClient struct {
 }
 
 // Create implements client.Client
-func (c *typedClient) Create(ctx context.Context, obj runtime.Object, opts ...CreateOption) error {
+func (c *typedClient) Create(ctx context.Context, obj Object, opts ...CreateOption) error {
 	o, err := c.cache.getObjMeta(obj)
 	if err != nil {
 		return err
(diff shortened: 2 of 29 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -377,10 +377,3 @@ func ContainsFinalizer(o Object, finalizer string) bool {
 	}
 	return false
 }
-
-// Object allows functions to work indistinctly with any resource that
-// implements both Object interfaces.
-type Object interface {
```
coderanger, line 383: Maybe throw in a `type Object = client.Object` to ease the transition?

vincepri: Maybe? It could be a good opportunity to finally move this interface somewhere else :)

pkg/cache/informer_cache_unit_test.go
```diff
@@ -25,11 +25,11 @@ var _ = Describe("ip.objectTypeForListObject", func() {
 		InformersMap: &internal.InformersMap{Scheme: scheme.Scheme},
 	}
 
-	It("should error on non-list types", func() {
-		_, _, err := ip.objectTypeForListObject(&corev1.Pod{})
-		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(Equal(`non-list type *v1.Pod (kind "/v1, Kind=Pod") passed as output`))
-	})
+	// It("should error on non-list types", func() {
+	// 	_, _, err := ip.objectTypeForListObject(&corev1.Pod{})
```
vincepri, line 29: going to remove this probably, it now fails at compilation time :)

pkg/controller/controllerutil/controllerutil.go
coderanger: It's a breaking change without that, seems unnecessary.

alvaroaleman: I tend to agree, because its not obvious without reading release notes where this went. What about keeping the alias for a minor release add adding a `Deprecated: Use from client package instead` comment?

vincepri: Yeah, there are tons of breaking changes in this release. I'm happy to hear more opinions from others as well, I can definitely add the alias, although the function signature will change regardless, so 🤷‍♂️

alvaroaleman: Well the definition of this interface stays the same. This is mostly relevant for ppl who use this interface in their code, for example as the type for a function argument.

vincepri: Done, ptal

DirectXMan12: yeah, that's generally a good idea.

Can mark the alias as deprecated though.

alvaroaleman (approved):
/lgtm
/hold

Looks good, but lets give folks a couple of days to comment

DirectXMan12 (changes requested):
this should modify typed_client too, right?

DirectXMan12 (approved):

pkg/client/typed_client.go
```diff
@@ -22,6 +22,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var _ Reader = &typedClient{}
```
alvaroaleman, line 25: Wouldn't it be easier to verify that `&typedClient{}` is a `Client` rather than that it is everything a `Client` is composed of?

DirectXMan12:
(this is gonna fail to compile, doing a quick push to test something with the new PR verifier)

pkg/client/typed_client.go
vincepri: tried that, although it doens't actually satisfy that interface, because it's missing Status() to get a status client, and RESTMapper() and Scheme()

I could backfill those if we want, but not sure if it's worth it

alvaroaleman: Ah okay, no its fine like that then

PR #1196, opened 2020-10-01
🐛 Controller reconcileHandler shouldn't stop a worker on error

vincepri:
During a code walkthrough noticed that oddly the reconcileHandler was
returning false and stopping the worker entirely when an error is
returned from the reconciler. This change removes the return boolean
from reconcileHandler altogether and a worker is now only stopped when
the queue is shutdown.

/milestone v0.7.x
/assign @alvaroaleman @ncdc @DirectXMan12


alvaroaleman (approved):
/lgtm
/approve

PR #1198, opened 2020-10-01
:book: Update versioning to reference common guidelines

DirectXMan12:
Update versioning to reference the common guidelines in the release
tools repo.  The guidelines were patterned on the CR guidelines, so we
don't have any specific overrides.

Dependent on kubernetes-sigs/kubebuilder-release-tools#4


vincepri (approved):
/lgtm

PR #1199, opened 2020-10-01
:book: Be more specific about compat in VERSIONING

DirectXMan12:
This adds a few CR-specific notes around versioning and compatibility,
mainly to emphasize the answer to the "what client-go version can I use"
question.


vincepri (approved):
💯 

/lgtm

PR #1200, opened 2020-10-02
:seedling: Use client.Object interface in envtest

vincepri:
This is a follow-up from an earlier PR that introduced the
client.Object interface. While working in Cluster API downstream,
noticed that this was missing from the envtest codebase.

/milestone v0.7.x


alvaroaleman (approved):
/lgtm

PR #1201, opened 2020-10-02
:warning: Propagate context.Context through source and predicates

vincepri:
This PR propagates a context on sources and predicates. The reasoning behind this is to allow for MapFunc, or predicate functions to get a context and pass it through for use cases like when using a client, or to get a logger.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	go.uber.org/goleak v1.1.10
 	go.uber.org/zap v1.15.0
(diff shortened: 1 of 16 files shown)


alvaroaleman:
IMHO this is taking it too far:
* Its confusing to have context part of things that can not be canceled
* Allowing to insert a logger is a nice idea, but that should remain optional
* Logger, client and any other dependency should be put into the EventHandler/Predicate at construction time, there is no need to get it as method argument

vincepri:
That's fair, I wasn't sure either after making the changes if this was something we'd want to pursue, happy to close this out

PR #1202, opened 2020-10-02
🌱 Add ctrl.{LoggerFrom, LoggerInto}

vincepri:
Renames the current (unreleased at the time of writing)
LoggerFromContext to LoggerFrom, and adds a new alias named LoggerInto.

From a user standpoint these new aliases are really clear and
straightfoward, example:
```go
func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) {
  log := ctrl.LoggerFrom(ctx)
```

This reads much better than ctrl.LoggerFromContext(ctx) which repeats
the context.

Another use case that this change solves is to enrich the logger with
more values, example:

```go
ctx = ctrl.LoggerInto(ctx, log.WithValues(...))
```

allows the ctx to still be the de-facto context, and the logger for the
inner reconciler function to still only accept ctx, without adding any
more parameters.

/milestone v0.7.x
/assign @DirectXMan12 @alvaroaleman

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -125,10 +125,15 @@ var (
 	// get any actual logging.
 	Log = log.Log
 
-	// LoggerFromContext returns a logger with predefined values from a context.Context.
+	// LoggerFrom returns a logger with predefined values from a context.Context.
 	//
 	// This is meant to be used with the context supplied in a struct that satisfies the Reconciler interface.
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm
/approve

PR #1204, opened 2020-10-02
:seedling: Add controller.GetLogger

vincepri:
Adds a new GetLogger method required on the controller interface. For
folks using controller without managers it's a good way to get a logger
pre-filled with information about the controller itself and avoid
duplication.

/assign @alvaroaleman


alvaroaleman (approved):
/lgtm

PR #1205, opened 2020-10-02
:warning: Propagate context on Manager.Start(...)

vincepri:
This change reshuffles how the manager accepts and operates on
a context. With this change, the user experience is greatly improved,
users can now use `ctrl.SetupSignalHandler()` to create a context, enrich
it if they want to, and pass it to `manager.Start`.

In addition, this PR changes how the context and stop channel are
handled internally to ensure proper cancellation.

/assign @alvaroaleman 
/milestone v0.7.x

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,6 @@ require (
 	k8s.io/apiextensions-apiserver v0.19.2
 	k8s.io/apimachinery v0.19.2
 	k8s.io/client-go v0.19.2
-	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
+	k8s.io/utils v0.0.0-20200912215256-4140de9c8800
 	sigs.k8s.io/yaml v1.2.0
 )
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -57,6 +57,7 @@ const (
 	defaultMetricsEndpoint   = "/metrics"
 )
 
+var _ Runnable = &controllerManager{}
 var log = logf.RuntimeLog.WithName("manager")
 
 type controllerManager struct {
@@ -187,6 +188,10 @@ type controllerManager struct {
 
 	internalCtx    context.Context
 	internalCancel context.CancelFunc
+
+	// internalProceduresStop channel is used internally to the manager when coordinating
+	// the proper shutdown of servers. This channel is also used for dependency injection.
+	internalProceduresStop chan struct{}
 }
 
 // Add sets dependencies on i, and adds it to the list of Runnables to start.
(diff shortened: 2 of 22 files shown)


hack/check-everything.sh
```diff
@@ -31,7 +31,7 @@ fetch_envtest_tools "$kb_root_dir"
 fetch_envtest_tools "${hack_dir}/../pkg/internal/testing/integration/assets"
 setup_envtest_env "$kb_root_dir"
 
-${hack_dir}/verify.sh
+# ${hack_dir}/verify.sh
```
alvaroaleman, line 34: Why is this commented?

pkg/manager/internal.go
```diff
@@ -185,8 +186,9 @@ type controllerManager struct {
 	// is closed because it will be nil.
 	shutdownCtx context.Context
 
-	internalCtx    context.Context
-	internalCancel context.CancelFunc
+	internalCtx            context.Context
+	internalCancel         context.CancelFunc
+	internalProceduresStop chan struct{}
```
alvaroaleman, line 191: Could you add a comment mentioning that this is only required for the `StopChannelInto`?

go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/controller-runtime
 go 1.15
 
 require (
+	github.com/davecgh/go-spew v1.1.1
```
alvaroaleman, line 6: Why is this now required, I can not see it referenced anywhere?

hack/check-everything.sh
vincepri: Leftover :)

pkg/manager/internal.go
vincepri: Done ptal

go.mod
vincepri: leftover

alvaroaleman (approved):
/lgtm

EronWright:
Would like to point out that this PR may have a secondary positive effect on test suites which utilize the controller test framework.  By propagating the manager's cancelable `context` to the event sources, and then to the cache during its initial sync, we avoid a possible deadlock due to the cache sync occurring after the manager (and its cache) has been stopped.  In other words, folks should see fewer spurious test failures.

vincepri:
@EronWright Thanks for pointing that out! I actually noticed the same on some flaky test within the Cluster API project, the other PR that merged a few days ago https://github.com/kubernetes-sigs/controller-runtime/pull/1354 also helps if you need to wait for the cache to sync during tests.

PR #1206, opened 2020-10-02
:sparkles: Zap: Add JSONEncoder and ConsoleEncoder opts

alvaroaleman:
Currently, setting the encoding is pretty annoying and requires to
import two more packages apart from log/zap:

```
encoderConfig := uberzap.NewProductionEncoderConfig()
zap.New(zap.Encoder(zapcore.NewJSONEncoder(encoderConfig)))
```

This PR adds a JSONEncoder and ConsoleEncoder opt to simplify this


vincepri (approved):
/lgtm

PR #1207, opened 2020-10-02
:warning: Handler: Remove MapObject type and use client.Object directly

alvaroaleman:
We updated the handlers MapObject type to not include a metav1.Object
and a runtime.Object representation of the same object but instead a
client.Object, which is a union of the two above.

The only reason for the MapObject type to exist in the first place
though was to hold these two fields. Now that there is only the
client.Object left, we can use that directly and get rid of the
MapObject type.

### DIFF
--- a/pkg/handler/example_test.go
+++ b/pkg/handler/example_test.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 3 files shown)


pkg/handler/example_test.go
```diff
@@ -65,15 +66,15 @@ func ExampleEnqueueRequestsFromMapFunc() {
 	// controller is a controller.controller
 	err := c.Watch(
 		&source.Kind{Type: &appsv1.Deployment{}},
-		handler.EnqueueRequestsFromMapFunc(func(a handler.MapObject) []reconcile.Request {
+		handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
```
alvaroaleman, line 69: This example basically shows how this makes things easier

vincepri (approved):
/approve
/lgtm
/milestone v0.7.x

PR #1211, opened 2020-10-09
🌱 Add distributed tracing framework

bboreham:
Created in support of https://github.com/kubernetes-sigs/cluster-api/issues/3760
[This doc](http://docs.google.com/document/d/1XeQKYWTk3dmVL4A8LxaNzR5TOHO8Kicw3NKPD7Qdqu4/) goes into more detail and is probably the best place to discuss.

What is in this PR:
 * A wrapper for `client.Client` which will log all Kubernetes API calls to the current trace.
 * A wrapper for `logr.Logger` which will add all log messages to a trace.
 * Support for passing trace span context from one program to another via an Annotation (as described in [this KEP](https://github.com/kubernetes/enhancements/blob/c7d895dbec5304345cb797113fdefd7be5761720/keps/sig-instrumentation/0034-distributed-tracing-kep.md))

This PR does not create any tracing spans.  You would use it in a controller like this (error handling elided):
```
func main() {
	...
		...
		NewClient: tracing.WrapRuntimeClient(util.ManagerDelegatingClientFunc),
	...
}

func (r *MyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, rerr error) {
	myObject := &MyObject{}
	r.Client.Get(ctx, req.NamespacedName, myObject)

	ctx, sp, log := tracing.FromObject(ctx, "Reconcile", myObject)
	if sp != nil {
		defer sp.Finish()
	}
	...
}
```

The code is written using the OpenTracing API and Jaeger; I will look at porting it to OpenTelemetry.


yue9944882:
@bboreham thank you for proposing/bringing open-telemetry integration to this library! i was looking for this for a period of time and also had a few PoC works before (in some of our internal projects) which is very close/similar to the current state of this pull. i think overall the proposed change is fairly intuitive but based on my humble experience from our practice, i think there're yet a few things not addressed in this pull:

- __the trace never closes__: the controller can be syncing resources by processing all items periodically even if there's no change on any resources --- that will create unbounded # of child spans which even makes it hard to visualize if there's multiple controllers working under the same trace. so the trace need to be enclosed properly if the resource successfully reach its desired state. and another tricky part is that if we're closing the trace by removing or editing the annotation, that will trigger a new reconciliation upon the resource, will that count into the trace?

- __sorting/dedup'ing context propagation__: consider cases where we're tracing cascading controller instances (as the following image roughly shows a A->B->C controller topology). i think the spans need to properly de-dup'd otherwise the visualization (on jaeger) can be confusing due to extra verbosity.

Screen Shot 2020-10-09 at 8 47 01 PM



and as far as i know this [KEP](https://github.com/kubernetes/enhancements/blob/c7d895dbec5304345cb797113fdefd7be5761720/keps/sig-instrumentation/0034-distributed-tracing-kep.md) was completely refactored to merely do the tracing for incoming requests at server side instead of controller frameworks, is it? the proposed change from this pull seems to be conflict w/ the latest KEP b/c having tracing at both server and client side can be be redundant i suppose.. cc @lavalamp

vincepri:
cc @alvaroaleman @DirectXMan12

bboreham:
Hi @yue9944882, lot of great points there.  I'll respond in pieces:

> and as far as i know this KEP was completely refactored to merely do the tracing for incoming requests at server side 

Yes, I noticed that today ([change](https://github.com/kubernetes/enhancements/pull/1458) was made 8 days ago).

> instead of controller frameworks, is it?

Nit: it specified kube-apiserver, kube-scheduler, kube-controller-manager and kubelet, (most of?) which don't use the controller-runtime framework.  But your point is a good one: I would now like to resurrect much of the text that was removed.

> the proposed change from this pull seems to be conflict w/ the latest KEP

I think they are complementary.

bboreham:
> the trace never closes

Yes, this is called out in the linked [Google doc](http://docs.google.com/document/d/1XeQKYWTk3dmVL4A8LxaNzR5TOHO8Kicw3NKPD7Qdqu4/).

>  the trace need to be enclosed properly if the resource successfully reach its desired state

Great thought.  I think for many specific cases this can be identified by a single controller.

> if we're closing the trace by removing or editing the annotation, that will trigger a new reconciliation upon the resource

Also true.  I wonder if it can be fixed by adding something to Status?  There isn't any handy bag-of-values field like Annotations, but maybe can be achieved as a Condition?

Further, suppose the observed state changes; we'd expect a reconcillation which (given the above) would start a new span.

bboreham:
> sorting/dedup'ing context propagation

Sorry I don't get this one at all, can you please explain more?

The text on your image is too small to see; here's a grab of the graph I get for a Cluster-API trace:
image

dashpole:
cc @DirectXMan12 
We discussed this a _looooong_ time ago

DirectXMan12:
(minor point of order: this would be a feature PR, not infra)

pkg/tracing/logger.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (t tracingLogger) Info(msg string, keysAndValues ...interface{}) {
+	t.Logger.Info(msg, keysAndValues...)
+	t.Span.LogKV(append([]interface{}{"info", msg}, keysAndValues...)...)
+}
+
+func (t tracingLogger) Error(err error, msg string, keysAndValues ...interface{}) {
+	t.Logger.Error(err, msg, keysAndValues...)
+	t.Span.LogKV(append([]interface{}{"error", err, "message", msg}, keysAndValues...)...)
```
DirectXMan12, line 25: is there a way to avoid needing an extra allocation here?

DirectXMan12: also, is there a semantic difference between `WithTags` and the key-value pairs specified here?

DirectXMan12 (commented):
Neat so far.

One of the things I've also played around with in the past is hooking the event emitter into this pipeline as well, so we get k8s events showing up too.

bboreham:
> hooking the event emitter into this pipeline

I want to do that, but it looks awkward, since the existing code takes it from the controller, and there is no Context parameter to the event recorder interface. Unless you know a way?

pkg/tracing/logger.go
bboreham: That `LogKV()` call is through an interface so the memory will escape; it's unlikely to be possible to avoid an allocation. This particular line is only hit under error conditions so probably not performance-sensitive, but there are similar ones.

Tags apply to the whole span so it wouldn't be right to put details from one error as a tag, however the next line sets an `error` tag on the span so you can pick it out as one that had at least one error.

bboreham:
Code in this PR is now using OpenTelemetry instead of OpenTracing.

One point of detail: span contexts inside objects now look like this:
```
  annotations:
    trace.kubernetes.io/traceparent: 00-99a904aa21ed09e78201d8b1e5ed7d44-a3901ad9eae5035c-01
```
which is a divergence from the (now-deleted) KEP I followed previously.

mcristina422:
Is there still a plan to carry this forward, with or without an updated KEP?

bboreham:
I'd still _like_ this to go forward, but I don't really have a plan.  I have been exploring related topics, particularly Events, but it took rather longer than I expected.

I believe a new KEP is needed, all relevant parts having been edited out of the one I originally followed.

Any advice welcome on how to take it forward.

bboreham:
Since I last updated, a new KEP has been opened: https://github.com/kubernetes/enhancements/pull/2312

It is still under active discussion; I have not changed my code to match the KEP.

bboreham:
Hi, I haven’t done much on this for a long time; do you need me to update it?

dims:
@bboreham yes please!

pkg/tracing/jaeger.go
```diff
@@ -0,0 +1,39 @@
...
+	"go.opentelemetry.io/otel/exporters/trace/jaeger"
+	"go.opentelemetry.io/otel/propagators"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// SetupJaeger sets up Jaeger with some defaults
+func SetupJaeger(serviceName string) (io.Closer, error) {
+	// Create and install Jaeger export pipeline
+	flush, err := jaeger.InstallNewPipeline(
+		jaeger.WithCollectorEndpoint("http://jaeger-agent.default:14268/api/traces"), // FIXME name?
```
fstrudel, line 16: Should we allow to specify the endpoint? (parameter/env variable/other?)

akselleirv:
Is there a possibility that this PR will be merged? It appears to be the closest effort to adding tracing support in controller-runtime.

Mershab99:
It would be awesome to see distributed tracing be available to operators!

@bboreham is there anything left to do to push this over the goal line?

Issue #1212, opened 2020-10-09
Fakeclient shold honor AllowUnconditionalUpdate and AllowCreateOnUpdate for resources without TypeMeta


fabriziopandini:
If I call `fake.Client.Update` passing a ConfigMap object without TypeMeta field set (`Kind=""`), I get error `"object was modified"`

This is probably because `AllowUnconditionalUpdate` retrives gvk from the object 

https://github.com/kubernetes-sigs/controller-runtime/blob/c000ea850121b53dd769f624f3bfa74a4fcf100f/pkg/client/fake/client.go#L433

(and similarly `AllowCreateOnUpdate`) 

In order to prevent this, in case `kind` is empty, `fake.Clien.Upgrade` should get the gkv from the registered scheme and pass it to `AllowUnconditionalUpdate` and `AllowCreateOnUpdate`

PR #1213, opened 2020-10-09
🐛 Tracker upgrade should use gvk from scheme

fabriziopandini:
This PR makes sure `tracker.Upgrade` uses the gvk from scheme for `AllowUnconditionalUpdate` and `AllowCreateOnUpdate` in case `TypeMeta` is not defined on the obj

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/1212

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -41,6 +41,7 @@ import (
 
 type versionedTracker struct {
 	testing.ObjectTracker
+	scheme *runtime.Scheme
 }
 
 type fakeClient struct {
@@ -74,7 +75,7 @@ func NewFakeClientWithScheme(clientScheme *runtime.Scheme, initObjs ...client.Ob
 		}
 	}
 	return &fakeClient{
-		tracker: versionedTracker{tracker},
+		tracker: versionedTracker{ObjectTracker: tracker, scheme: clientScheme},
 		scheme:  clientScheme,
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/client/fake/client.go
```diff
@@ -114,11 +115,20 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 			field.ErrorList{field.Required(field.NewPath("metadata.name"), "name is required")})
 	}
 
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	emptyGvk := schema.GroupVersionKind{}
+	if gvk == emptyGvk {
```
vincepri, line 120:
```suggestion
	if gvk == schema.GroupVersionKind{} {
```

detiber:
lgtm pending Vince's suggestion and fixing the PR title to appease the checks.

pkg/client/fake/client.go
fabriziopandini: @vincepri I have tested this syntax first, but it is failing to compile with "expected declaration, found oldObject"

vincepri: Oh I think you can just do `gvk.Empty()`

vincepri (approved):
/approve

vincepri:
@fabriziopandini Please fix the PR title

cc @DirectXMan12 Not sure why the check isn't running anymore 🤔

PR #1214, opened 2020-10-10
:sparkles: Allow passing the assets path bin via the env config

camilamacedo86:
**Description**
- Allow passing the assets path bin via the env config

```go
        By("bootstrapping test environment")
		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
                BinaryAssetsDirectory: "filepath.Join("..", "bin")" // See here
	}
```

The priority here is: 

0. the specific envvars for each binary: `TEST_ASSET_KUBE_APISERVER`, `TEST_ASSET_ETCD`, `TEST_ASSET_KUBECTL`
1. the path for all env var: `KUBEBUILDER_ASSETS `
2. the value set via the structure in this case `Environment{}` 
3. the default value which is currently `/usr/local/kubebuilder/bin`

**Motivation**

Allow passing the bin path to simplify the required implementation to config the EnvTest when the bin is not in the default path. Currently, if a developer would like to try to get the bins from a specific dir if they exist would be required an implementation such as: 

```go

	By("bootstrapping test environment")
		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
	}

	By("configuring env test binaries")
	configEnvTestBinaries()

	var err error
	cfg, err = testEnv.Start()
	Expect(err).ToNot(HaveOccurred())
	Expect(cfg).ToNot(BeNil())

	err = crewv1.AddToScheme(scheme.Scheme)

	// +kubebuilder:scaffold:scheme

	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
	Expect(err).ToNot(HaveOccurred())
	Expect(k8sClient).ToNot(BeNil())

	close(done)
}, 60)

var _ = AfterSuite(func() {
	By("tearing down the test environment")
	err := testEnv.Stop()
	Expect(err).ToNot(HaveOccurred())
})

// configEnvTestBinaries will check if the required binaries are in the bin directory
// of the project and configure to use them.
func configEnvTestBinaries() {
	binPath := filepath.Join("..", "..", "bin")
	if hasEnvTestBinaries(binPath) {
		os.Setenv("KUBEBUILDER_ASSETS", binPath)
	}
}

// hasEnvTestBinaries will return true when the kubebuilder tools binaries are found
// in the path informed
func hasEnvTestBinaries(path string) bool {
	if _, err := os.Stat(path); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "etcd")); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "kube-apiserver")); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "kubectl")); os.IsNotExist(err) {
		return false
	}
	return true
}
```

With this change we can just:

```go

	By("bootstrapping test environment")
		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
                BinaryAssetsDirectory: "filepath.Join("..", "bin")" // See here
	}

	var err error
	cfg, err = testEnv.Start()
	Expect(err).ToNot(HaveOccurred())
	Expect(cfg).ToNot(BeNil())

	err = crewv1.AddToScheme(scheme.Scheme)

	// +kubebuilder:scaffold:scheme

	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
	Expect(err).ToNot(HaveOccurred())
	Expect(k8sClient).ToNot(BeNil())

	close(done)
}, 60)

var _ = AfterSuite(func() {
	By("tearing down the test environment")
	err := testEnv.Stop()
	Expect(err).ToNot(HaveOccurred())
})
```

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -62,14 +62,22 @@ const (
 	defaultKubebuilderControlPlaneStopTimeout  = 20 * time.Second
 )
 
-// Default binary path for test framework
-func defaultAssetPath(binary string) string {
-	assetPath := os.Getenv(envKubebuilderPath)
-	if assetPath == "" {
-		assetPath = defaultKubebuilderPath
+// getBinAssetPath returns a path for binary from the following list of locations,
+// ordered by precedence:
+// 0. KUBEBUILDER_ASSETS
+// 1. Environment.BinaryAssetsDirectory
+// 2. The default path, "/usr/local/kubebuilder/bin"
+func (te *Environment) getBinAssetPath(binary string) string {
+	valueFromEnvVar := os.Getenv(envKubebuilderPath)
+	if valueFromEnvVar != "" {
+		return filepath.Join(valueFromEnvVar, binary)
 	}
-	return filepath.Join(assetPath, binary)
 
+	if te.BinaryAssetsDirectory != "" {
+		return filepath.Join(te.BinaryAssetsDirectory, binary)
+	}
+
+	return filepath.Join(defaultKubebuilderPath, binary)
 }
 
 // ControlPlane is the re-exported ControlPlane type from the internal integration package
@@ -112,6 +120,10 @@ type Environment struct {
 	// If both this field and Paths field in CRDInstallOptions are specified, the
 	// values are merged.
 	CRDDirectoryPaths []string
+
+	// BinaryAssetsDirectory is the path where the binaries required for the envtest are
+	// located in the local environment. This field can be overridden by setting KUBEBUILDER_ASSETS.
+	BinaryAssetsDirectory string
 
 	// UseExisting indicates that this environments should use an
 	// existing kubeconfig, instead of trying to stand up a new control plane.
@@ -217,14 +229,14 @@ func (te *Environment) Start() (*rest.Config, error) {
 		}
 
 		if os.Getenv(envKubeAPIServerBin) == "" {
-			te.ControlPlane.APIServer.Path = defaultAssetPath("kube-apiserver")
+			te.ControlPlane.APIServer.Path = te.getBinAssetPath("kube-apiserver")
 		}
 		if os.Getenv(envEtcdBin) == "" {
-			te.ControlPlane.Etcd.Path = defaultAssetPath("etcd")
+			te.ControlPlane.Etcd.Path = te.getBinAssetPath("etcd")
 		}
 		if os.Getenv(envKubectlBin) == "" {
 			// we can't just set the path manually (it's behind a function), so set the environment variable instead
-			if err := os.Setenv(envKubectlBin, defaultAssetPath("kubectl")); err != nil {
+			if err := os.Setenv(envKubectlBin, te.getBinAssetPath("kubectl")); err != nil {
 				return nil, err
 			}
 		}


pkg/envtest/server.go
```diff
@@ -43,33 +43,56 @@ It's possible to override some defaults, by setting the following environment va
...
-	StopTimeout            = 60
-
+	envUseExistingClusterEnvVar = "USE_EXISTING_CLUSTER"
+	envKubeAPIServerBinEnvVar   = "TEST_ASSET_KUBE_APISERVER"
+	envEtcdBinEnvVar            = "TEST_ASSET_ETCD"
+	envKubectlBinEnvVar         = "TEST_ASSET_KUBECTL"
+	envKubebuilderPathEnvVar    = "KUBEBUILDER_ASSETS"
+	envStartTimeoutEnvVar       = "KUBEBUILDER_CONTROLPLANE_START_TIMEOUT"
+	envStopTimeoutEnvVar        = "KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT"
+	envAttachOutputEnvVar       = "KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT"
```
camilamacedo86, line 55: I just renamed the env vars const to make clear what is EnvVar and what is default values

pkg/envtest/server.go
```diff
@@ -62,14 +62,17 @@ const (
...
-		assetPath = defaultKubebuilderPath
+// getBinAssetPath will return the path for the binary informed
+func (te *Environment) getBinAssetPath(binary string) string {
+	if hasBinary(binary, te.BinaryDirectoryPath) {
+		return filepath.Join(te.BinaryDirectoryPath, binary)
 	}
-	return filepath.Join(assetPath, binary)
 
+	valueFromEnvVar := os.Getenv(envKubebuilderPath)
+	if hasBinary(binary, valueFromEnvVar) {
```
coderanger, line 72: I would probably throw in a `valueFromEnvVar != "" &&` on this, just to avoid silently using `/kube-apiserver`, though I admit it's unlikely that would exist :)

coderanger: (and same above)

coderanger: Alternatively could do `if path == "" { return false }` down in hasBinary.

camilamacedo86: done in the hasBinary func 👍

coderanger (approved):

pkg/envtest/server.go
```diff
@@ -62,14 +62,17 @@ const (
...
 )
 
-// Default binary path for test framework
-func defaultAssetPath(binary string) string {
-	assetPath := os.Getenv(envKubebuilderPath)
-	if assetPath == "" {
-		assetPath = defaultKubebuilderPath
+// getBinAssetPath will return the path for the binary informed
+func (te *Environment) getBinAssetPath(binary string) string {
+	if hasBinary(binary, te.BinaryDirectoryPath) {
```
estroz, line 67: The `hasBinary` pattern doesn't work here, because we want to inform the envtest caller that the binary does not exist at the path they specified and not at the default path, which is what this method currently does. Instead, just check if each data variable is set, and if so use it as the parent dir.
```suggestion
	if te.BinaryDirectoryPath != "" {
```

pkg/envtest/server.go
```diff
@@ -62,14 +62,17 @@ const (
...
+		return filepath.Join(te.BinaryDirectoryPath, binary)
 	}
-	return filepath.Join(assetPath, binary)
 
+	valueFromEnvVar := os.Getenv(envKubebuilderPath)
+	if hasBinary(binary, valueFromEnvVar) {
+		return filepath.Join(valueFromEnvVar, binary)
+	}
```
estroz, line 74: Typically, environment variables override hard-coded configuration, so this should be
```
func (te *Environment) getBinAssetPath(binary string) string {
	if valueFromEnvVar := os.Getenv(envKubebuilderPath); valueFromEnvVar != "" {
		return filepath.Join(valueFromEnvVar, binary)
	}
	if te.BinaryDirectoryPath != "" {
		return filepath.Join(te.BinaryDirectoryPath, binary)
	}
	return filepath.Join(defaultKubebuilderPath, binary)
}

File: pkg/envtest/server.go [hunk 5]
```diff
@@ -113,6 +116,10 @@ type Environment struct {
 	// values are merged.
 	CRDDirectoryPaths []string
 
+	// BinaryDirectoryPath is the path where the binaries required for the envtest are
+	// locate in the environment
+	BinaryDirectoryPath string
```
estroz, line 121: Nit: if envtest requires an absolute path, this is the variable name to use
```suggestion
	AbsBinaryDirectory string
```

coderanger: I don't think anything in this requires it be an absolute path? Though definitely would be playing with fire to do otherwise :)

pkg/envtest/server.go
estroz: Interestingly the k8s integration package comment says the order is [reversed](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.7.0-alpha.6/pkg/internal/testing/integration/doc.go#L51-L66), which I find strange because environment variables are typically used as overrides. Any thoughts on this?

pkg/envtest/server.go
estroz: Just checked, it isn't. It definitely should be because theoretically you could run the test from anywhere in the project, which would probably mess with how envtest processes are spawned.

pkg/envtest/server.go
camilamacedo86: The priority here is: 

0. the value set via the structure in this case `Environment{}` 
1. the specific envvars for each binary: `TEST_ASSET_KUBE_APISERVER`, `TEST_ASSET_ETCD`, `TEST_ASSET_KUBECTL`
2. the path for all env var: `KUBEBUILDER_ASSETS `
3. the default value which is currently `/usr/local/kubebuilder/bin`

which matches with other cases scenarios in place as you linked. So, because of this, I do not agree with the suggestion on https://github.com/kubernetes-sigs/controller-runtime/pull/1214#discussion_r520916615

pkg/envtest/server.go
camilamacedo86: I'd thought that would be nice if envtest was able to check if the bin exists or not and then try to check the follow-up the options according to the priority. However, I can agree that it is out of the scope over `Allow passing the assets path bin via the env config` so I changed it as you suggested. 👍

pkg/envtest/server.go
camilamacedo86: Changed as suggested.

pkg/envtest/server.go
```diff
@@ -113,6 +116,10 @@ type Environment struct {
 	// values are merged.
 	CRDDirectoryPaths []string
 
+	// AbsBinaryDirectory is the path where the binaries required for the envtest are
+	// locate in the environment
+	AbsBinaryDirectory string
```
DirectXMan12, line 121: This name is a bit unclear.  Can we make this `TestBinariesDirectory` or `BinaryAssetsDirectory` or something?

DirectXMan12 (changes requested):
minor comment on naming, otherwise lgtm

pkg/envtest/server.go
camilamacedo86: done

pkg/envtest/server.go
```diff
@@ -62,14 +62,18 @@ const (
 	defaultKubebuilderControlPlaneStopTimeout  = 20 * time.Second
 )
 
-// Default binary path for test framework
-func defaultAssetPath(binary string) string {
-	assetPath := os.Getenv(envKubebuilderPath)
-	if assetPath == "" {
-		assetPath = defaultKubebuilderPath
+// getBinAssetPath will return the path for the binary informed
```
estroz, line 65:
```suggestion
// getBinAssetPath returns a path for binary from the following list of locations,
// ordered by precedence:
// 1. KUBEBUILDER_ASSETS
// 2. Environment.BinaryAssetsDirectory
// 3. The default path, "/usr/local/kubebuilder/bin"
```

pkg/envtest/server.go
```diff
@@ -113,6 +117,10 @@ type Environment struct {
 	// values are merged.
 	CRDDirectoryPaths []string
 
+	// BinaryAssetsDirectory is the path where the binaries required for the envtest are
+	// locate in the environment
```
estroz, line 121:
```suggestion
	// located in the environment. This field can be overridden by setting KUBEBUILDER_ASSETS.
```

camilamacedo86:
Hi @estroz, 

Thank you for the review. All done 👍

estroz (approved):
/lgtm

PR #1224, opened 2020-10-22
:warning: client.ObjectKeyFromObject now uses client.Client

vincepri:
This change removes the need to get an accessor from the runtime object,
because the object interface already has all the information needed.

It also removes the need to return an error, which makes it a little
easier to use.

/assign @alvaroaleman 
/milestone v0.7.x


alvaroaleman (approved):
/lgtm
/approve

alvaroaleman (approved):
/lgtm

PR #1231, opened 2020-10-28
:book: Improve docs for client.Object

DirectXMan12:
It was brought to my attention that the relationship between
client.Object and runtime.Object isn't particularly clear to folks who
are not used to working in the guts of client-go & apimachinery, so this
adds some docs to helpfully clarify and provide examples.

### DIFF
--- a/pkg/client/object.go
+++ b/pkg/client/object.go
@@ -21,15 +21,56 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cc @cflewis

cflewis (approved):
LGTM, thanks for this!

alvaroaleman (approved):
/lgtm
Thanks!

alvaroaleman:
@DirectXMan12 sidenote, do you think we should try to uptstream these two interfaces to make them a bit more official?

PR #1232, opened 2020-10-29
:book: Improve docs for manager.Options.SyncPeriod

dlipovetsky:
The comment tries to capture more of the thoughts exchanged in #88.

It explains why a period sync happens. It also addresses two (common?) reasons for making the period shorter, and recommends an alternative.

The current comment includes this warning: "Change this value only if you know what you are doing." I found this unsatisfying, and wanted to add more context to the comment. I'm not an authority on the behavior/history of `SyncPeriod`, and I welcome feedback :slightly_smiling_face:.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -140,6 +140,8 @@ type Options struct {
 	// there will a 10 percent jitter between the SyncPeriod of all controllers
 	// so that all controllers will not send list requests simultaneously.
 	//
+	// This applies to all controllers.
+	//
 	// A period sync happens for two reasons:
 	// 1. To insure against a bug in the controller that causes an object to not
 	// be requeued, when it otherwise should be requeued.
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -145,7 +145,7 @@ type Options struct {
 	// A period sync happens for two reasons:
 	// 1. To insure against a bug in the controller that causes an object to not
 	// be requeued, when it otherwise should be requeued.
-	// 2. To insure against a bug in controller-runtime, or its dependencies,
+	// 2. To insure against an unknown bug in controller-runtime, or its dependencies,
 	// that causes an object to not be requeued, when it otherwise should be
 	// requeued, or to be removed from the queue, when it otherwise should not
 	// be removed.
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -139,6 +139,23 @@ type Options struct {
 	// value only if you know what you are doing. Defaults to 10 hours if unset.
 	// there will a 10 percent jitter between the SyncPeriod of all controllers
 	// so that all controllers will not send list requests simultaneously.
+	//
+	// A period sync happens for two reasons:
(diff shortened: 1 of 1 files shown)


dlipovetsky:
(Updated for clarity and force-pushed)

pkg/manager/manager.go
```diff
@@ -139,6 +139,23 @@ type Options struct {
...
+	// 1. To insure against a bug in the controller that causes an object to not
+	// be requeued, when it otherwise should be requeued.
+	// 2. To insure against a bug in controller-runtime, or its dependencies,
+	// that causes an object to not be requeued, when it otherwise should be
+	// requeued, or to be removed from the queue, when it otherwise should not
+	// be removed.
+	//
+	// If you want
+	// 1. to insure against missed watch events, or
+	// 2. to poll services that cannot be watched,
```
coderanger, line 153: Depending on the use case, this can be better served by a channel watch and a background goroutine, but that's definitely the more advanced version :)

joejulian (approved):

pkg/manager/manager.go
```diff
@@ -139,6 +139,23 @@ type Options struct {
 	// value only if you know what you are doing. Defaults to 10 hours if unset.
 	// there will a 10 percent jitter between the SyncPeriod of all controllers
 	// so that all controllers will not send list requests simultaneously.
+	//
```
alvaroaleman, line 142: Nit: Could you mention that this setting applies to all controllers (which is why I think using it is bad but we don't have a nice `Source` that would make this easy on controller-level) and that it works by creating an artificial ADD event?

pkg/manager/manager.go
```diff
@@ -139,6 +139,23 @@ type Options struct {
...
+	// A period sync happens for two reasons:
+	// 1. To insure against a bug in the controller that causes an object to not
+	// be requeued, when it otherwise should be requeued.
+	// 2. To insure against a bug in controller-runtime, or its dependencies,
```
alvaroaleman, line 146: This sounds like there was a known bug. I think it is very unlikely that there is a bug in that in either controller-runtime or the underlying kube api machinery

joejulian:
@apelisse @hoegaarden Can this get an approval please?

alvaroaleman:
@joejulian https://github.com/kubernetes-sigs/controller-runtime/pull/1232/files#r515084286 is still open

joejulian:
Ah, I see. I'm not used to PR's being held up by comments marked as nits. My mistake.

pkg/manager/manager.go
joejulian:
```suggestion
	//
	// This applies to all controllers.
	//
```

pkg/manager/manager.go
joejulian:
```suggestion
	// 2. To insure against an unknown bug in controller-runtime, or its dependencies,
```

dlipovetsky:
Thanks addressing the feedback @joejulian!
 
/retest

alvaroaleman (approved):
/lgtm
/approve

thank you!

Issue #1161, opened 2020-09-14
admissionv1.AddToScheme needs to be called for v1.AdmissionReview


ellistarn:
Controller-gen 0.4.0 now uses v1 types. This needs to be included here: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/http.go#L28

```
2020-09-14T19:47:34.638Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/mutate", "UID": "", "allowed": false, "result": {}, "resultError": "got runtime.Object without object metadata: &Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:,Message:no kind \"AdmissionReview\" is registered for version \"admission.k8s.io/v1\" in scheme \"pkg/runtime/scheme.go:101\",Reason:,Details:nil,Code:400,}"}
```

ellistarn:
@DirectXMan12 any word here?

ellistarn:
@DirectXMan12 any thoughts?

vincepri:
Do we want to support only admission request v1, or both?

ellistarn:
For backwards compatibility, I think both need to be supported.

ellistarn:
For any kubebuilder users, I'm temporarily fixing this with
```
perl -pi -e 's/^  - v1$$//g' config/webhook/manifests.yaml
```

bsctl:
Hello team, any plan on this?

camilamacedo86:
HI @ellistarn and @bsctl, 

Shows that it is done in the master with https://github.com/kubernetes-sigs/controller-runtime/pull/1293. COuld you please check and close this one if it is the case?

ellistarn:
Will check it out! Thanks.

prometherion:
Hey @camilamacedo86, thanks for sharing these: everything works fine with the latest versions, thanks a lot! 🙌🏻

PR #1233, opened 2020-10-30
⚠ pkg/webhook: Add support for k8s.io/api/admission/v1

### LINKED ISSUE #1123: pkg/webhook: k8s.io/api/admission/v1beta1 is deprecated in v1.19
See https://github.com/kubernetes/kubernetes/pull/90300. Bumping to v1 is a breaking change.

/cc @vincepri @DirectXMan12

jiachengxu:
This PR adds support of `k8s.io/api/admission/v1` in webhook package and moves support of `v1beta1` to legacy.
This is because  `k8s.io/api/admission.k8s.io/v1beta1` is deprecated in kubernetes v1.16 in favor of `admission.k8s.io/v1`.
controller-tools v0.4.0 can already generate WebhookConfigurations in `v1`, and more and more folks are trying to upgrade to that, but then they figured out that controller-runtime is not compatible with it. This PR tackles that.
Fixes #1123 #1161
Ref: https://github.com/kubernetes-sigs/controller-tools/pull/469

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -62,6 +62,17 @@ func (blder *WebhookBuilder) Complete() error {
 	return blder.registerWebhooks()
 }
 
+// CompleteLegacy builds the webhook to support `k8s.io/api/admission.k8s.io/v1beta1` api.
+// Since `k8s.io/api/admission.k8s.io/v1beta1` is deprecated in kubernetes v1.16 in favor of admission.k8s.io/v1, please consider
+// to upgrade if possible.
+func (blder *WebhookBuilder) CompleteLegacy() error {
+	// Set the Config
+	blder.loadRestConfig()
+
+	// Set the Webhook if needed
+	return blder.registerWebhooksLegacy()
+}
+
 func (blder *WebhookBuilder) loadRestConfig() {
 	if blder.config == nil {
 		blder.config = blder.mgr.GetConfig()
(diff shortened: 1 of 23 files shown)


jiachengxu:
cc @vincepri @alvaroaleman @estroz

pkg/builder/webhook.go
```diff
@@ -62,6 +62,17 @@ func (blder *WebhookBuilder) Complete() error {
 	return blder.registerWebhooks()
 }
 
+// CompleteLegacy builds the webhook to support `k8s.io/api/admission.k8s.io/v1beta1` api.
+// Since `k8s.io/api/admission.k8s.io/v1beta1` is deprecated in kubernetes v1.16 in favor of admission.k8s.io/v1, please consider
+// to upgrade if possible.
+func (blder *WebhookBuilder) CompleteLegacy() error {
```
alvaroaleman, line 68: So this PR now basically duplicates the whole webhook code. I see two problems with that:
a): We duplicate a lot of code
b) Controller authors might now know what webhook version is available to their end users (managed kube is _very_ slow)

What about we build a v1beta1 <-> v1 round tripping function, internally only handle one of the two and always expose two handlers, one for v1beta1, one for v1, one of which will convert the request and the result?

/cc @DirectXMan12

alvaroaleman: (I see that I agreed to the current approach in https://github.com/kubernetes-sigs/controller-runtime/issues/1123, I am very sorry about that. I didn't realize that this means we have duplicate so much code)

jiachengxu: Sorry that I was pretty busy with my work in the last few days.
> We duplicate a lot of code

Yeah, that's true. However, I think duplicating code makes it easier to drop the support in the future. In our case, we are going to drop support of `v1beta1` in the future instead of supporting both `v1` and `v1beta1`, we can just remove the files and functions with the `Legacy` suffix, and it will work.
Also, building a fancy round-tripping function may be a good choice, but it may need more interface designing & discussion. Since more and more folks found that controller-runtime cannot work with controller-tool v0.4.0 for webhook `v1` version, I wanted to implement something fast and workable for them.
Finally, if we still want to have more discussion and designing around this, how about we just release a new version of controller-tool with this PR: https://github.com/kubernetes-sigs/controller-tools/pull/474, so at least user can specify the `AdmissionReviewVersion` to `v1beta1` and migrate to webhook `v1`. Then we figure out how we want to support `AdmissionReviewVersion` `v1` in controller-runtime. (/cc @vincepri)

>  Controller authors might now know what webhook version is available to their end users (managed kube is very slow)

If I got this correctly, you meant controller authors might need to know which version of `AdmissionReviewVersion` that the end users' clusters support. I think they need to know this anyway if they want to introduce break changes and migrate from legacy `v1beta1` to `v1`? Because I think with two `handlers` `v1` and `v1beta1` if `v1beta1` of `AdmissionReview` API is deprecated in the cluster, controller authors might still need to migrate handlers to `v1`? Please correct if I got this wrong.  

Finally, please don't feel sorry about our previous discussion :) We are all trying to make the community better!

alvaroaleman: > Also, building a fancy round-tripping function may be a good choice, but it may need more interface designing & discussion.

I don't think it's fancy, its just copying all fields of one type into the equivalent fields of the other type and write a fuzz test that roundtripps using something like https://github.com/google/gofuzz. Since this is not going to be part of an external interface, I don't think there will be much designing and discussion involved.

> I think they need to know this anyway if they want to introduce break changes and migrate from legacy v1beta1 to v1? Because I think with two handlers v1 and v1beta1 if v1beta1 of AdmissionReview API is deprecated in the cluster, controller authors might still need to migrate handlers to v1? Please correct if I got this wrong.

From what I can see they are actually identical, so its irrelevant for authors: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/admission/v1/types.go#L40 https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/admission/v1beta1/types.go#L45

DirectXMan12: At the very least, for the builder, none of the underlying webhook payload is surfaced to the user, so the builder should handle both cases automatically for the user.

The "lower-level" variants may be a different story.

DirectXMan12: Oooh, if they're currently identical, we can just deserialize into v1 safely.

DirectXMan12 (changes requested):
At the very least, the high-level builder doesn't surface the payload to the user, so we should be able to paper over things.

That being said: currently, `v1` and `v1beta1` are identical (there's a change from `should` to `must` in the comments, but it doesn't change the actual field type), which means that we can safely just deserialize v1beta1 into a v1 struct and call it a day, without the duplication.  Since you can't (modulo deprecated fields in the future, but by that time v1beta1 will prob be removed) remove fields from a stable API, we should be good on this into the future if we just do that.

DirectXMan12:
NB: v1 changed "should copy the UID" to "must copy the UID" in the response, so we need to make sure we're doing that

vincepri:
Closing in favor of #1284

/close

PR #1238, opened 2020-11-02
🐛 Fix use of `UnstructuredList` with unregistered typed

ironcladlou:
Before this change, passing an `unstructured.UnstructuredList` to
`client.List()` would erroneously delegate to the typed client, causing an error
if the target type isn't registered with the client's scheme, violating the
purpose of using `unstructured`.

This patch fixes the client so that the `unstructured.UnstructuredList` is
delegated to the internal unstructured client.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -213,7 +213,7 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object) error {
 // List implements client.Client
 func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) error {
 	switch obj.(type) {
-	case *unstructured.Unstructured:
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm
/approve

ironcladlou:
Here's an example of what the failure looks like:

```go
list := &unstructured.UnstructuredList{}
list.SetGroupVersionKind(schema.GroupVersionKind{
  Group:   "group.example.com",
  Version: "v1",
  Kind:    "FooList",
})
client.List(context.TODO(), list)
```

```
v1.ListOptions is not suitable for converting to "group.example.com/v1" in scheme "example.go:xx"
```

alvaroaleman:
The heck.
/retest

PR #1241, opened 2020-11-03
 🌱 Use lowercase space separated keys

enxebre:
Change built-in logging keys to be consistent with https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md#logging-structured-values-key-value-pairs


alvaroaleman (commented):
/cc @vincepri 
/lgtm
/approve

PR #1242, opened 2020-11-03
:sparkles: Allow to use builder.OnlyMetadata option with Watches

vincepri:
This change allows builders to use builder.OnlyMetadata when creating
extra watches with .Watches(...). The code inspects the passed
source.Source, and if it's of type *source.Kind, it calls the internal
project method that allows to use metav1.PartialObjectMetadata in
mapping functions.

/milestone v0.7.x

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -116,9 +116,10 @@ func (blder *Builder) Owns(object client.Object, opts ...OwnsOption) *Builder {
 
 // WatchesInput represents the information set by Watches method.
 type WatchesInput struct {
-	src          source.Source
-	eventhandler handler.EventHandler
-	predicates   []predicate.Predicate
+	src              source.Source
+	eventhandler     handler.EventHandler
+	predicates       []predicate.Predicate
+	objectProjection objectProjection
 }
 
 // Watches exposes the lower-level ControllerManagedBy Watches functions through the builder.  Consider using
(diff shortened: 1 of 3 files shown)


pkg/builder/options.go
```diff
@@ -94,6 +94,11 @@ func (p projectAs) ApplyToOwns(opts *OwnsInput) {
 	opts.objectProjection = objectProjection(p)
 }
 
+// ApplyToOwns applies this configuration to the given OwnsInput options.
+func (p projectAs) ApplyToWatches(opts *WatchesInput) {
```
alvaroaleman, line 98: Godoc is wrong

pkg/builder/controller.go
```diff
@@ -255,10 +256,20 @@ func (blder *Builder) doWatch() error {
...
-			return err
+		watchSource := w.src
+
+		// If the source of this watch is of type *source.Kind, project it.
+		if srckind, ok := watchSource.(*source.Kind); ok {
+			typeForSrc, err := blder.project(srckind.Type, w.objectProjection)
+			if err != nil {
+				return err
+			}
+			watchSource = &source.Kind{Type: typeForSrc}
```
alvaroaleman, line 267: Please re-use the existing object and only overwrite the `Type`, right now we drop for example the cache and in theory it could get more attributes in the future

vincepri: Done

alvaroaleman (approved):
/lgtm
/approve

Issue #1219, opened 2020-10-17
Allow configuring a cache sync timeout


alvaroaleman:
Currently when a cache can not be synced (for example due to misconfigured rbac) the `Controller` will just block indefinitely/until the binary gets stopped: https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/internal/controller/controller.go#L172

As an author of controllers I want to configure a timeout for this and want my `Controller` to error out if its unable to sync its caches within this timeout.

varshaprasad96:
Hi @alvaroaleman , I would like to work on this issue. Would the implementation be to:
- Wrap [`WaitForSync`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/source/source.go#L135-L141) to pass a `Context` with a timeout so that the ctx is cancelled when [`WaitForCacheSync`](https://www.sohamkamani.com/golang/2018-06-17-golang-using-context-cancellation/) is called.
- include a `Timeout` variable in the [controller](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L42) which falls back to a default value if not set by the user.

Is this the right approach for the implementation?

alvaroaleman:
@varshaprasad96 yes, that sounds exactly right

PR #1247, opened 2020-11-05
:sparkles: Allow configuring cache sync timeouts

varshaprasad96:
This PR allows users to configure timeout for cache syncs
while starting the controller.

Closes: #1219

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	"k8s.io/client-go/util/workqueue"
@@ -47,6 +48,10 @@ type Options struct {
 	// Log is the logger used for this controller and passed to each reconciliation
 	// request via the context field.
 	Log logr.Logger
+
+	// CacheSyncTimeout refers to the time limit set to wait for syncing caches.
+	// Defaults to 2 minutes if not set.
+	CacheSyncTimeout time.Duration
 }
 
 // Controller implements a Kubernetes API.  A Controller manages a work queue fed reconcile.Requests
@@ -104,6 +109,10 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		options.MaxConcurrentReconciles = 1
 	}
 
+	if options.CacheSyncTimeout == 0 {
+		options.CacheSyncTimeout = 2 * time.Minute
+	}
+
 	if options.RateLimiter == nil {
 		options.RateLimiter = workqueue.DefaultControllerRateLimiter()
 	}
@@ -120,6 +129,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 			return workqueue.NewNamedRateLimitingQueue(options.RateLimiter, name)
 		},
 		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
+		CacheSyncTimeout:        options.CacheSyncTimeout,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
 		Log:                     options.Log.WithName("controller").WithName(name),
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -79,6 +79,10 @@ type Controller struct {
 	// undergo a major refactoring and redesign to allow for context to not be stored in a struct.
 	ctx context.Context
 
+	// CacheSyncTimeout refers to the time limit set on waiting cache to sync
+	// Defaults to 2 minutes if not set.
+	CacheSyncTimeout time.Duration
+
 	// startWatches maintains a list of sources, handlers, and predicates to start when the controller is started.
 	startWatches []watchDescription
 
@@ -142,6 +146,10 @@ func (c *Controller) Start(ctx context.Context) error {
 	// Set the internal context.
 	c.ctx = ctx
 
+	// use a context with timeout for launching sources and syncing caches.
+	sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
+	defer cancel()
+
 	c.Queue = c.MakeQueue()
 	defer c.Queue.ShutDown() // needs to be outside the iife so that we shutdown after the stop channel is closed
 
@@ -156,7 +164,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		// caches.
 		for _, watch := range c.startWatches {
 			c.Log.Info("Starting EventSource", "source", watch.src)
-			if err := watch.src.Start(ctx, watch.handler, c.Queue, watch.predicates...); err != nil {
+			if err := watch.src.Start(sourceStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
 				return err
 			}
 		}
@@ -169,7 +177,7 @@ func (c *Controller) Start(ctx context.Context) error {
 			if !ok {
 				continue
 			}
-			if err := syncingSource.WaitForSync(ctx); err != nil {
+			if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
 				// This code is unreachable in case of kube watches since WaitForCacheSync will never return an error
 				// Leaving it here because that could happen in the future
 				err := fmt.Errorf("failed to wait for %s caches to sync: %w", c.Name, err)
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -79,7 +79,7 @@ type Controller struct {
 	// undergo a major refactoring and redesign to allow for context to not be stored in a struct.
 	ctx context.Context
 
-	// CacheSyncTimeout refers to the time limit set on waiting cache to sync
+	// CacheSyncTimeout refers to the time limit set on waiting for cache to sync
 	// Defaults to 2 minutes if not set.
 	CacheSyncTimeout time.Duration
 
@@ -146,10 +146,6 @@ func (c *Controller) Start(ctx context.Context) error {
 	// Set the internal context.
 	c.ctx = ctx
 
-	// use a context with timeout for launching sources and syncing caches.
-	sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
-	defer cancel()
-
 	c.Queue = c.MakeQueue()
 	defer c.Queue.ShutDown() // needs to be outside the iife so that we shutdown after the stop channel is closed
 
@@ -164,7 +160,10 @@ func (c *Controller) Start(ctx context.Context) error {
 		// caches.
 		for _, watch := range c.startWatches {
 			c.Log.Info("Starting EventSource", "source", watch.src)
-			if err := watch.src.Start(sourceStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
+
+			watchStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
+			defer cancel()
+			if err := watch.src.Start(watchStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
 				return err
 			}
 		}
@@ -177,9 +176,14 @@ func (c *Controller) Start(ctx context.Context) error {
 			if !ok {
 				continue
 			}
+
+			// use a context with timeout for launching sources and syncing caches.
+			sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
+			defer cancel()
+
+			// WaitForSync waits for a definitive timeout, and returns if there
+			// is an error or a timeout
 			if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
-				// This code is unreachable in case of kube watches since WaitForCacheSync will never return an error
-				// Leaving it here because that could happen in the future
 				err := fmt.Errorf("failed to wait for %s caches to sync: %w", c.Name, err)
 				c.Log.Error(err, "Could not wait for Cache to sync")
 				return err
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -122,6 +122,60 @@ var _ = Describe("controller", func() {
 			close(done)
 		})
 
+		It("should error when cache sync timeout occurs", func(done Done) {
+			ctrl.CacheSyncTimeout = 10 * time.Nanosecond
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			_, err = c.GetInformer(context.TODO(), &appsv1.Deployment{})
+			Expect(err).NotTo(HaveOccurred())
+			sync := false
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, &informertest.FakeInformers{
+					Synced: &sync,
+				}),
+			}}
+
+			err = ctrl.Start(context.TODO())
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("cache did not sync"))
+
+			close(done)
+		})
+
+		It("should not error when cache sync time out is of reasonable value", func(done Done) {
+			ctrl.CacheSyncTimeout = 1 * time.Second
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, c),
+			}}
+
+			By("running the cache and waiting for it to sync")
+			go func() {
+				defer GinkgoRecover()
+				Expect(c.Start(context.TODO())).To(Succeed())
+			}()
+			close(done)
+		})
+
+		It("should error when timeout is set to a very low value such that cache cannot sync", func(done Done) {
+			ctrl.CacheSyncTimeout = 1 * time.Nanosecond
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, c),
+			}}
+
+			By("running the cache and waiting for it to sync")
+			go func() {
+				defer GinkgoRecover()
+				err = ctrl.Start(context.TODO())
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("cache did not sync"))
+			}()
+			close(done)
+		})
+
 		It("should call Start on sources with the appropriate EventHandler, Queue, and Predicates", func() {
 			pr1 := &predicate.Funcs{}
 			pr2 := &predicate.Funcs{}
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -124,15 +124,12 @@ var _ = Describe("controller", func() {
 
 		It("should error when cache sync timeout occurs", func(done Done) {
 			ctrl.CacheSyncTimeout = 10 * time.Nanosecond
+
 			c, err := cache.New(cfg, cache.Options{})
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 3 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -112,6 +117,10 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		options.Log = mgr.GetLogger()
 	}
 
+	if options.CacheSyncTimeout == 0 {
+		options.CacheSyncTimeout = 10 * time.Second
+	}
```
jmrodri, line 122: I would add the comment about defaulting to 10 seconds above this code as well. 
```golang
// Default timeout to 10 seconds
if options.CacheSyncTimeout == 0 {
...
```

pkg/internal/controller/controller.go
```diff
@@ -79,6 +79,10 @@ type Controller struct {
 	// undergo a major refactoring and redesign to allow for context to not be stored in a struct.
 	ctx context.Context
 
+	// CacheSyncTimeout refers to the time limit set on waiting cache to sync
+	// Defaults to 10 seconds if not set.
+	CacheSyncTimeout time.Duration
```
jmrodri, line 84: Where does this get defaulted to the 10 seconds? I only saw the one above.

pkg/source/source_test.go
```diff
@@ -211,7 +212,7 @@ var _ = Describe("Source", func() {
 			instance := source.Kind{}
 			f := false
 			Expect(instance.InjectCache(&informertest.FakeInformers{Synced: &f})).To(Succeed())
-			err := instance.WaitForSync(nil)
+			err := instance.WaitForSync(context.Background(), time.Duration(10)*time.Second)
```
jmrodri, line 215: Do we need to pass in a different duration? We default to 10, what happens if this is > 10?

pkg/source/source_test.go
```diff
@@ -247,13 +248,24 @@ var _ = Describe("Source", func() {
...
 		})
+
+		It("should return an error if nil context is passed", func(done Done) {
+			f := false
+			instance := source.NewKindWithCache(nil, &informertest.FakeInformers{Synced: &f})
+			err := instance.WaitForSync(nil, time.Duration(10)*time.Second)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal("cannot pass a nil context"))
+
+			close(done)
```
jmrodri, line 266: :+1:

jmrodri (changes requested):
a couple questions and a suggestion to add another comment.

alvaroaleman (commented):
/ok-to-test

pkg/controller/controller.go
```diff
@@ -112,6 +117,10 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		options.Log = mgr.GetLogger()
 	}
 
+	if options.CacheSyncTimeout == 0 {
+		options.CacheSyncTimeout = 10 * time.Second
```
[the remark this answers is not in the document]
vincepri: Should we leave it as is? Indefinite so we don't break lots of potential controllerS?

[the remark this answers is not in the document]
alvaroaleman: IMHO the current behavior of just locking without any feedback is pretty bad. Having a high default sounds good to me if you want that? Maybe accompany it with an explanation of what could cause this (too many objects, rbac, resource doesn't exist on server).

The underlying annoyance is that the apimachinery doesn't expose this info at all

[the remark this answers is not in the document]
vincepri: Sure, I'm ok having it a high default and make it visible

pkg/source/source.go
```diff
@@ -132,8 +133,13 @@ func (ks *Kind) String() string {
 
 // WaitForSync implements SyncingSource to allow controllers to wait with starting
 // workers until the cache is synced.
-func (ks *Kind) WaitForSync(ctx context.Context) error {
-	if !ks.cache.WaitForCacheSync(ctx) {
+func (ks *Kind) WaitForSync(ctx context.Context, timeout time.Duration) error {
```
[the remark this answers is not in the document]
joelanford: If I understand @alvaroaleman's suggestion correctly, I think we can call `context.WithTimeout()` in `controller.Start()` and pass that new context to `syncingSource.WaitForSync()`.

That way, we don't have to change function signatures for `WaitForSync`.

pkg/controller/controller.go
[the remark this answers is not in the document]
alvaroaleman: @vincepri do you have a number in your head for `high default`? 60s maybe?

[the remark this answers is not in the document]
vincepri: Let's start maybe with 10 minutes? Or is that too much?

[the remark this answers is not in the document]
alvaroaleman: 10 minutes? We have a controller that currently has about 9k objects it watches for its resource and that one sync its cache fine in 30 seconds. Have you ever seen a cache sync duration that even came close to the 1 minute?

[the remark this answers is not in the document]
alvaroaleman: The slowest sync I've personally ever seen was from an openshift api that at that point in time had about 3 gigabytes of data, when running the corresponding controller locally it took about 2 min to sync. But that amount of data is extremely unusual IMHO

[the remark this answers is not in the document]
vincepri: That's good to hear, let's start with 2 mins then?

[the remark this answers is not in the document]
alvaroaleman: sounds good to me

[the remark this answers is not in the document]
varshaprasad96: Updated to use 2 mins as default value.

pkg/source/source.go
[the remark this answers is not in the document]
varshaprasad96: Modified to pass a context with deadline to the entire `Start()` function. This sets a timeout for starting the sources as well as for syncing the caches.

pkg/source/source_test.go
```diff
@@ -247,13 +248,24 @@ var _ = Describe("Source", func() {
 		It("should return an error if syncing fails", func(done Done) {
 			f := false
 			instance := source.NewKindWithCache(nil, &informertest.FakeInformers{Synced: &f})
-			err := instance.WaitForSync(nil)
+			err := instance.WaitForSync(context.Background(), time.Duration(10)*time.Second)
```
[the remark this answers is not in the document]
varshaprasad96: Have added a test case - where the caches do not sync and because of the timeout which is set there is an error (this verifies that we don't go into an infinite loop waiting for caches to sync). However, verifying the case with a reasonable timeout requires us to pass a context to `controller.Start(..)` and wait for the channel to close. Currently, the [`fakeinformer` ](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/informertest/fake_cache.go#L83-L88) does not use the context and instead only mocks whether caches are synced or not to ensure that `Start()` is not blocked. I suppose that's why in all the other test cases we are passing a cancelled context. Modifying the fake informer to accept the context would break the other test cases. Is there a workaround for this, or is the current test sufficient?

pkg/internal/controller/controller_test.go
```diff
@@ -122,6 +122,26 @@ var _ = Describe("controller", func() {
...
 
+		It("should error when cache sync timeout occurs", func(done Done) {
+			ctrl.CacheSyncTimeout = 10 * time.Nanosecond
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			_, err = c.GetInformer(context.TODO(), &appsv1.Deployment{})
+			Expect(err).NotTo(HaveOccurred())
+			sync := false
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, &informertest.FakeInformers{
```
alvaroaleman, line 133: Instead of using the `FakeInformer`, please use the cache you construct in line 127 (The testcase you copied this from seems to be bugged, it constructs a cache and then never uses it).

To test the positive case (we correctly sync when there is a reasonable timeout), you could do the following:
* Start off similiar like this test, construct a cache and set it up in `startWatches` 
* Build a wrapper struct for the `Source` that closes a channel after `WaitForSync` was called (below)
* Start the cache in a distinct goroutine (as that is blocking)
* Start the controller in a distinct goroutine (as that is blocking, too)
* Wait for the channel in the source or a timeout via `select`
* Close the context used to start the cache and controller

wrapper would be something like this:
```
type signalingSourceWrapper struct {
  cacheSyncDone chan struct{}
  source.Source
}

func (s *singalingSourceWrapper) WaitForSync(ctx context.Context) error {
  defer func(){close(s.cacheSyncDone} ()
  return s.Source.WaitForSync(ctx)
}
```

Let me know if you have further questions on this

pkg/source/source_test.go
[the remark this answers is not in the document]
alvaroaleman: See my other comment, writing an integration test that uses a real cache and not the fakeinformer should be doable here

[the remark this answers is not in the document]
varshaprasad96: Got it.. modified to include the test cases

pkg/internal/controller/controller_test.go
varshaprasad96: ops, sorry, I had missed this. Does the current implementation look fine or shall I modify it ?

pkg/internal/controller/controller_test.go
```diff
@@ -122,6 +122,60 @@ var _ = Describe("controller", func() {
...
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, c),
+			}}
+
+			By("running the cache and waiting for it to sync")
+			go func() {
+				defer GinkgoRecover()
+				err = ctrl.Start(context.TODO())
```
varshaprasad96, line 172: `Start` seems to return a closed channel, so I haven't cancelled the context later. Can I cancel the context after `Start` fails or succeeds, or would it be better to use a wrapper that closes the channel as you have mentioned previously.

pkg/internal/controller/controller_test.go
```diff
@@ -122,6 +122,60 @@ var _ = Describe("controller", func() {
...
+			c, err := cache.New(cfg, cache.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			ctrl.startWatches = []watchDescription{{
+				src: source.NewKindWithCache(&appsv1.Deployment{}, c),
+			}}
+
+			By("running the cache and waiting for it to sync")
+			go func() {
+				defer GinkgoRecover()
+				Expect(c.Start(context.TODO())).To(Succeed())
```
alvaroaleman, line 156: We never start the controller here so this test doesn't test what it claims

pkg/internal/controller/controller_test.go
alvaroaleman: This will pass with any timeout, because the cache is never started which short-circuits the actual `WaitForCacheSync`: https://github.com/kubernetes-sigs/controller-runtime/blob/33f1d80ebeaa33c9e3d86c86c713ac5847dbe859/pkg/cache/internal/deleg_map.go#L77-L86

Please write two identical tests that only differ in the timeout.

> Start seems to close the channel, so I haven't cancelled the context later. Can I cancel the context after Start fails or succeeds, or would it be better to use a wrapper that closes the channel as you have mentioned previously.

What channel are you referring to that gets closed by `Start`?

Right now in most cases the `go func` will execute after the test is ended via the `close(done`) in line 176, you would need to block before that via a channel that gets closed at the end of the `go func`. But its unclear to my why this is even in a distinct goroutine.

pkg/internal/controller/controller_test.go
varshaprasad96: I tried to do the following:
1. Created a wrapper around `SyncingSource` as that in turn wraps Source:
```
type SingnallingSourceWrapper struct {
	cacheSyncDone chan struct{}
	source.SyncingSource
}

func NewSignallingSource(cacheChannel chan struct{}, src source.SyncingSource) source.SyncingSource {
	return &SingnallingSourceWrapper{cacheSyncDone: cacheChannel, SyncingSource: src}
}

func (s *SingnallingSourceWrapper) WaitForSync(ctx context.Context) error {
	defer func() {
		close(s.cacheSyncDone)
	}()
	return s.SyncingSource.WaitForSync(ctx)
}
```

In the test, I tried the following as you have suggested:

```
// Configure a new cache:
c, err := cache.New(cfg, cache.Options{})
ctx := context.TODO()


// Set it up with start watches. Create a new `channel` for SignallingSource:
src := source.NewKindWithCache(&appsv1.Deployment{}, c)
ctrl.startWatches = []watchDescription{{
	src: NewSignallingSource(channel, src),
}}

// Start cache and controller in a separate go routine:
go c.Start(ctx)
go ctrl.Start(ctx)

// Wait for channel to close:
select {
	case <-channel:
		// channel has closed, cache has synced.
	default:
		// there is a timeout. cache did not sync
}

ctx.Done()
```

When I check for the condition where cache is synced, there is always a timeout. And I am also facing the error:
```
watch of *v1.Deployment ended with: very short watch: pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156: Unexpected watch close - watch lasted less than a second and no items received
```

I'm not able to figure out the reason for timing out always since the context which is being sent for starting caches and controller is closed in the end. Also the `channel` seems to still be open. Is there something which Im missing?

alvaroaleman: Can you check the error return of the cache and controller?

alvaroaleman: Also, the `default` case needs to not be `default` but something that actually times out, i.E.:

```
timeoutCtx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
defer cancel()
select {
	case <-channel:
		// channel has closed, cache has synced.
	case <-timeoutCtx.Done():
		// there is a timeout. cache did not sync
}
```

varshaprasad96: The controller does return an error:
```
ErrStatus: {
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "Timeout: failed waiting for *v1.Deployment Informer to sync here",
                Reason: "Timeout",
                Details: {Name: "", Group: "", Kind: "", UID: "", Causes: nil, RetryAfterSeconds: 0},
                Code: 504,
            }
```

It seems to error while waiting for informer to sync (ie) [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L187-L189)

alvaroaleman: Ah. Well, that simplifies the whole thing, we can then simply start the controller synchronously and verify we get that error

varshaprasad96: Yeah, the same error does occur when controller is started synchronously without any timeout set. But isn't informer not getting synced also mean that caches are not synced? if so, how can we verify the case that everything works fine and caches are synced when no timeout is set.

pkg/internal/controller/controller.go
```diff
@@ -156,7 +164,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		// caches.
 		for _, watch := range c.startWatches {
 			c.Log.Info("Starting EventSource", "source", watch.src)
-			if err := watch.src.Start(ctx, watch.handler, c.Queue, watch.predicates...); err != nil {
+			if err := watch.src.Start(sourceStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
```
estroz, line 167: This context should be set up for each `Start` call. Otherwise there could be an erroneous timeout if you have a bunch of watches.
```suggestion
			// use a context with timeout for launching sources and syncing caches.
			watchStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
			defer cancel()
			if err := watch.src.Start(watchStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
```

pkg/internal/controller/controller.go
```diff
@@ -142,6 +146,10 @@ func (c *Controller) Start(ctx context.Context) error {
 	// Set the internal context.
 	c.ctx = ctx
 
+	// use a context with timeout for launching sources and syncing caches.
+	sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
+	defer cancel()
+
```
estroz, line 152:
```suggestion
```

pkg/internal/controller/controller.go
```diff
@@ -169,7 +177,7 @@ func (c *Controller) Start(ctx context.Context) error {
 			if !ok {
 				continue
 			}
-			if err := syncingSource.WaitForSync(ctx); err != nil {
+			if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
```
estroz, line 180:
```suggestion
			// use a context with timeout for launching sources and syncing caches.
			sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
			defer cancel()
			if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
```

estroz (changes requested):

pkg/internal/controller/controller_test.go
alvaroaleman: So the reason this happens is that the cache isn't started properly by the time we start the controller. Since `cache.Start()` is blocking, it can not be used well to figure out when it finished starting. You could probably use `Eventually` with `cache.List()` for something that is not a deployment to verify that it finished starting before starting the controller (its important that its a different object than whatever the controller has, as the `List` creates and syncs an informer for that object type)

varshaprasad96: When I start the cache in a separate go routine, wait for it to start, then start a controller in a separate go routine - I get the error `pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156: watch of *v1.ReplicaSet ended with: very short watch: pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156: Unexpected watch close - watch lasted less than a second and no items received`

Why is it required that when we set a watch, there should be a change in the state of the resource ?

alvaroaleman: Not sure why you are seeing that error, it is not required to have any change in the state of the resources for this to work. What works for me is this:
```
    It("should error when cache sync timeout occurs", func(done Done) {
      c, err := cache.New(cfg, cache.Options{})
      ctrl.CacheSyncTimeout = 10 * time.Millisecond
      ctrl.startWatches = []watchDescription{{src: source.NewKindWithCache(&appsv1.Deployment{}, c)}}

      err = ctrl.Start(context.TODO())
      Expect(err.Error()).To(ContainSubstring("cache did not sync"))

      close(done)
    })

    It("should not error when cache sync time out is of reasonable value", func(done Done) {                                                                                                                                                                                                                                   
      ctx, cancel := context.WithCancel(context.Background())
      defer cancel()

      ctrl.CacheSyncTimeout = 1 * time.Second
      c, err := cache.New(cfg, cache.Options{})

      sourceSynced := make(chan struct{})
      ctrl.startWatches = []watchDescription{{
        src: &syncingSourceWrapper{
          SyncingSource: source.NewKindWithCache(&appsv1.Deployment{}, c), 
          synced:        sourceSynced,
        },
      }}  

      // Start the cache and controller asynchronously and wait for the source to sync
      go func() {
        defer GinkgoRecover()
        Expect(c.Start(ctx)).To(Succeed())
      }() 
      go func() {
        defer GinkgoRecover()
        Expect(ctrl.Start(ctx)).To(Succeed())
      }() 
      <-sourceSynced

      close(done)
    }, 10.0)
```

Using this wrapper:
```
type syncingSourceWrapper struct {
  source.SyncingSource
  synced chan struct{}
}

func (ssw *syncingSourceWrapper) WaitForSync(ctx context.Context) error {
  if err := ssw.SyncingSource.WaitForSync(ctx); err != nil {
    return err
  }
  close(ssw.synced)
  return nil
} 
```

alvaroaleman (approved):
/lgtm
/approve

Thank you for your work on this :)

Issue #1222, opened 2020-10-20
Support APIReader/NoCache option for ControllerUtil.CreateOr(Update|Patch) functions to avoid OOM errors


akutz:
# Use Case(s)

As a Developer,
I would like to use [`controllerutil.CreateOrUpdate`](https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/controller/controllerutil/controllerutil.go#L190-L229) and/or [`controllerutil.CreateOrPatch`](https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/controller/controllerutil/controllerutil.go#L238-L331) without the `Get` calls in both functions resulting in the creation of an informer/watch for the specified object,
Because the implicit watch behavior creates scaling issues and OOM errors on pods that may not ever explicitly watch a `ConfigMap`, but _do_ create one, on a cluster with thousands of `ConfigMap` resources.

# Details

By default, upon creation, the controller-runtime Manager [initializes a `DelegatingClient`](https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/manager/manager.go#L514-L517), which reads from an internal cache and writes directly to the API server. This client uses a backing cache implementation that automatically sets up informers (aka watches) on resource types during [`Get`](https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/cache/informer_cache.go#L53-L69) and [`List`](https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/pkg/cache/informer_cache.go#L71-L89) calls.

This is by design, and is normally fine, but contributes to issues at scale. Imagine a controller that creates `ConfigMap` resources, but never explicitly watches them. That controller may be running in a pod with a 30MiB memory limit. The pod will be terminated with an out-of-memory (OOM) error on a cluster with 100-120 `ConfigMap` resources at ~50 KiB each -- resources that have **nothing at all** to do with this controller.

[...]

PR #1249, opened 2020-11-05
⚠Add ability for the delegating client to avoid caching objects

vincepri:
A long standing issue in controller runtime, the delegating client is
the default client that the manager creates. Whenever a user calls
Get/List on a typed object, the internal cache spins up an informer and
start watching *all* objects for that group-version-kind.

This change introduces the ability for the delegating client to take
a list of objects that should always hit the live api-server, and bypass
the cache.

We also offer the ability to build a manager.NewClientFunc with a new
builder that exposes the ability to mark which objects we want to be
uncached.

Marking this as ⚠ because it changes the signature of NewDelegatingClient, which can now return an error.

/milestone v0.7.x

Related to #1222

### DIFF
--- a/pkg/client/split.go
+++ b/pkg/client/split.go
@@ -22,12 +22,15 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // NewDelegatingClientInput encapsulates the input parameters to create a new delegating client.
 type NewDelegatingClientInput struct {
-	CacheReader Reader
-	Client      Client
+	CacheReader     Reader
+	Client          Client
+	UncachedObjects []Object
 }
 
 // NewDelegatingClient creates a new delegating client.
@@ -35,17 +38,28 @@ type NewDelegatingClientInput struct {
 // A delegating client forms a Client by composing separate reader, writer and
 // statusclient interfaces.  This way, you can have an Client that reads from a
 // cache and writes to the API server.
-func NewDelegatingClient(in NewDelegatingClientInput) Client {
+func NewDelegatingClient(in NewDelegatingClientInput) (Client, error) {
+	uncachedGVKs := map[schema.GroupVersionKind]struct{}{}
+	for _, obj := range in.UncachedObjects {
+		gvk, err := apiutil.GVKForObject(obj, in.Client.Scheme())
+		if err != nil {
+			return nil, err
+		}
+		uncachedGVKs[gvk] = struct{}{}
+	}
+
 	return &delegatingClient{
 		scheme: in.Client.Scheme(),
 		mapper: in.Client.RESTMapper(),
 		Reader: &delegatingReader{
 			CacheReader:  in.CacheReader,
 			ClientReader: in.Client,
+			scheme:       in.Client.Scheme(),
+			uncachedGVKs: uncachedGVKs,
 		},
 		Writer:       in.Client,
 		StatusClient: in.Client,
-	}
+	}, nil
 }
 
 type delegatingClient struct {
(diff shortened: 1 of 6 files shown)


pkg/manager/client.go
```diff
@@ -0,0 +1,66 @@
...
+package manager
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DefaultNewClient creates the default caching client
```
alvaroaleman, line 26: This is still super inconvenient to use, can we just make the `WithUncached` an opt in the managers `Options` struct?

vincepri: What's inconvenient to use?

alvaroaleman: Well, in order to set this, the whole call will look something like:
```
mgr, err := manager.New(cfg, manager.Optionss{
   NewClientFunc: manager.NewClientFuncBuilder().WithUncached(&SomeType{}),
}
```

Something like:
```
    DisableCacheFor: []runtime.Object{&SomeType{}},
}
```

is IMHO both easier to discover and easier to use. WDYT?

vincepri: I see, we could do both maybe? The only thing that throws me off with the second one is that the cache bypass is technically used with the client, not with a manager. What would happen for example if one sets `DisableCacheFor` list, and supply a `NewClientFunc` that doesn't support it?

alvaroaleman: > The only thing that throws me off with the second one is that the cache bypass is technically used with the client, not with a manager.

Maybe call it `DisableClientCacheFor`?

> What would happen for example if one sets DisableCacheFor list, and supply a NewClientFunc that doesn't support it?

Well, we can add a comment to mention that, but generally I would expect ppl that set a NewClientFunc to realize that it will overwrite all client-related settings. AFAIK we also have a `DryRunClient` option that would get invalidated just the same.

pkg/client/split.go
```diff
@@ -75,21 +89,37 @@ func (d *delegatingClient) RESTMapper() meta.RESTMapper {
...
+
+func (d *delegatingReader) isUncached(obj runtime.Object) (bool, error) {
+	gvk, err := apiutil.GVKForObject(obj, d.scheme)
+	if err != nil {
+		return false, err
+	}
+	_, isUncached := d.uncachedGVKs[gvk]
+	_, isUnstructured := obj.(*unstructured.Unstructured)
+	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
+	return isUncached || isUnstructured || isUnstructuredList, nil
```
ncdc, line 105: I would probably limit this strictly to checking if it's uncached, and keep the Unstructured check in Get and the UnstructuredList check in List.

alvaroaleman: What I was wondering is if it would make sense to re-use the `isUncached` mechanism for this. The main issue is that ppl that set types as uncached probably do not want to overwrite the existing logic for unstructured with that.

vincepri: I moved it all here because unstructured and list are all uncached by default, although if that's preferable I can move it back

pkg/manager/client.go
vincepri: Oh! We can change the NewClientFunc to be the builder interface, which defaults to the internal builder, and it can call the methods it knows

vincepri: Does that sound good to you @alvaroaleman? I might also backport it in a non-breaking way

alvaroaleman: But what does that change, we would still have to pass a `NewClientFunc` rather than just configuring the option we care about? IMHO this is really hard to discover

vincepri: @alvaroaleman Updated with what I was thinking, it's a middle ground between both, if you still don't like it happy to just remove the builder

pkg/client/split.go
alvaroaleman: So the current approach means that we allow configuring this except for `unstructured`. IMHO if we want to make this configurable, then we should also allow to disable it for `unstructured`, the current hardcoded assumption that someone who uses this type doesn't want it to get cached might or might not not be true (I've had cases where it wasn't). That in turns means that we need to split the configuration for this into a `SetUncached` and `AddUncached` (its completely fine if you only implement the latter here, but I don't want another interface change down the line, `WithUncached` is neither of the two).

This also has the advantage that we can better document the behavior on the constructors, right now there is no godoc mentioning the unstructured behavior

alvaroaleman (commented):
> @alvaroaleman Updated with what I was thinking, it's a middle ground between both, if you still don't like it happy to just remove the builder

For some reason github doesn't allow me to respond in the thread. The builder now is fine, but please see my other comment ragarding `unstructured`

akutz:
I just wanted to say thank you @vincepri for working on this.

alvaroaleman (approved):
/lgtm

PR #1253, opened 2020-11-08
✨ rename GetControllerManagerConfiguration to Complete

### LINKED ISSUE #1252: Rename GetControllerManagerConfiguration to Complete for ComponentConfig
I'd like to make a change to `pkg/config/config.go` which will register as a breaking change, when I was implementing the ComponentConfig `DeferredFileLoader` I setup a func called `GetControllerManagerConfiguration()` while this name is factually correct it's quite long and couple be replaced by a simpler sounding name like `Complete()`. 

Links: 
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/config/config.go#L36
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L419

Changing this would make the calls:

```golang
newObj, err := loader.Complete()
// error handling
```

If I were to implement this the PR Checks specifically the `go-apidiff` will fail since it's a breaking change. Is there a path around this without adding a new `func Complete()` and supporting both given that this is an alpha feature and it's fully implemented?

Ref #895

christopherhein:
This is a breaking change for an unreleased feature that will be rolled out in `0.7.0`. This renames `loader.GetControllerManagerConfiguration()` to `loader.Complete()`.

Fixes #1252

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,8 +32,8 @@ import (
 type ControllerManagerConfiguration interface {
 	runtime.Object
 
-	// GetControllerManagerConfiguration returns the versioned configuration
-	GetControllerManagerConfiguration() (v1alpha1.ControllerManagerConfigurationSpec, error)
+	// Complete returns the versioned configuration
+	Complete() (v1alpha1.ControllerManagerConfigurationSpec, error)
 }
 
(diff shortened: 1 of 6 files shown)


vincepri (approved):
/approve
/lgtm

PR #1254, opened 2020-11-09
✨ Add predicate for annotations change on update event

zcahana:
This PR introduces an `AnnotationChangedPredicate` that skips update events that have no change in the object's annotation.

This predicate complements the `GenerationChangedPredicate` introduced in #553, and is intended to be used in conjunction with it, as in the following example:

```go
Controller.Watch(
	&source.Kind{Type: v1.MyCustomKind},
	&handler.EnqueueRequestForObject{},
	predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}))
```

This is mostly useful for controllers that need to trigger both when the resource's generation is incremented
(i.e., when the resource' `.spec` changes), or an annotation changes (e.g., for a staging/alpha API).

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -17,6 +17,8 @@ limitations under the License.
 package predicate
 
 import (
+	"reflect"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
lgtm
/approve

Thanks!

PR #1256, opened 2020-11-10
✨ Backport release-0.5: Support metadata-only watches 

vincepri:
This PR backports the metadata only watches to the release-0.5 branch in a non breaking change. The API is a bit different and I had to have some workarounds to make it work, but it seems that it should fit nicely with the rest of it.

In particular, given the lack of Options for the builder For, Watches, and Owns, we support metadata-only types by encapsulating them in a custom internal type, which then gets projected to the actual type when the manager gets built.

/milestone v0.5.x
/assign @alvaroaleman @DirectXMan12 @ncdc

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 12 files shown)


alvaroaleman (approved):

DirectXMan12:
uuhhh, this retroactively makes the original PR and the other backport breaking, right?

DirectXMan12:
chatted offline briefly, I think this is fine as long as we remember to also mark the corresponding changes in v0.6.4 and v0.7.0 as breaking instead (with a bit of an explanation as to what happened), since neither of those have been released yet.

PR #1258, opened 2020-11-10
:sparkles: hack/setup-envtest.sh: install binaries and print export variable command

estroz:
The current setup-envtest.sh script is a little clunky, since it contains a combination of global settings and functions, and your calling shell is expected to be bash. Instead, the script should only download the envtest binaries, then print what to do next (`export KUBEBUILDER_ASSETS=bin`). You can automatically do the next thing with

```sh
eval $(./setup-envtest.sh | tail -1)
```

Binaries are now installed into the argument directory directly instead of under `$1/bin`, which is more intuitive; the default is `$PWD/bin`.

### DIFF
--- a/hack/check-everything.sh
+++ b/hack/check-everything.sh
@@ -20,16 +20,12 @@ set -o pipefail
 
 hack_dir=$(dirname ${BASH_SOURCE})
 source ${hack_dir}/common.sh
-source ${hack_dir}/setup-envtest.sh
 
-tmp_root=/tmp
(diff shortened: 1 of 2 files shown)


estroz:
This should help fix problems like https://github.com/operator-framework/operator-sdk/issues/4203

estroz:
Doesn't seem to be getting attention, and I think there was affirmative discussion of providing a library to download and/or look up envtest binaries rather than rely on the wrapping shell env.

PR #1259, opened 2020-11-10
:warning: Fakeclient: Allow to pass ObjectLists to constructors

alvaroaleman:
When changing the fakeclients constructors to take a `client.Object` we missed that its possible to initialize it with `ObjectLists` which do not fulfill that interface. Hence, reverting it back to `runtime.Object`.

/assign @vincepri

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -541,7 +541,7 @@ var _ = Describe("Fake client", func() {
 			Expect(corev1.AddToScheme(scheme)).To(Succeed())
 			Expect(appsv1.AddToScheme(scheme)).To(Succeed())
 			Expect(coordinationv1.AddToScheme(scheme)).To(Succeed())
-			cl = NewFakeClientWithScheme(scheme, dep, dep2, cm)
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client_test.go
```diff
@@ -541,7 +541,7 @@ var _ = Describe("Fake client", func() {
 			Expect(corev1.AddToScheme(scheme)).To(Succeed())
 			Expect(appsv1.AddToScheme(scheme)).To(Succeed())
 			Expect(coordinationv1.AddToScheme(scheme)).To(Succeed())
-			cl = NewFakeClientWithScheme(scheme, dep, dep2, cm)
+			cl = NewFakeClientWithScheme(scheme, &appsv1.DeploymentList{Items: []appsv1.Deployment{*dep, *dep2}}, cm)
```
alvaroaleman, line 544: The test is a tad implicit, but we run all tests once for each of the two constructor, so this was the easiest way to add a check

vincepri (approved):
/approve
/lgtm

PR #1265, opened 2020-11-16
✨ Support for using both an external control plane and automatic webhooks together

coderanger:
This requires pre-filling LocalServingHost and possibly LocalServingExternalName with the right listen interface and external name so that the external control plane can reach the webhook server launched by envtest.

Using this successfully for local integration tests against kind and k3s. It does change some APIs but the only compat change is on addr.Suggest and that's explicitly an internal API so I think that's okay?

### DIFF
--- a/pkg/envtest/webhook.go
+++ b/pkg/envtest/webhook.go
@@ -67,6 +67,9 @@ type WebhookInstallOptions struct {
 	// CAData is the CA that can be used to trust the serving certificates in LocalServingCertDir.
 	LocalServingCAData []byte
 
+	// LocalServingHostExternalName is the hostname to use to reach the webhook server.
+	LocalServingHostExternalName string
+
 	// MaxTime is the max time to wait
 	MaxTime time.Duration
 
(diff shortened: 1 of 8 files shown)


pkg/envtest/webhook.go
```diff
@@ -137,13 +140,19 @@ func modifyWebhook(webhook map[string]interface{}, caData []byte, hostPort strin
...
-	port, host, err := addr.Suggest()
-	if err != nil {
-		return "", fmt.Errorf("unable to grab random port for serving webhooks on: %v", err)
+	if o.LocalServingPort == 0 {
+		port, host, err := addr.Suggest(o.LocalServingHost)
+		if err != nil {
+			return "", fmt.Errorf("unable to grab random port for serving webhooks on: %v", err)
+		}
+		o.LocalServingPort = port
+		o.LocalServingHost = host
```
alenkacz, line 149: What do you think about reusing this field instead? It would be generated IF not provided.

alenkacz (commented):
Also - is there a way to test this on a higher level than unit test? 🤔

pkg/envtest/webhook.go
coderanger: You mean re: adding the new struct field? LocalServingHost has to be a name or IP for the webhook server bind to, in most cases I would expect it to be set to `0.0.0.0` when doing this style of test. The ExternalName field is what ends up written into the webhook definition and needs to be the externally visible name for the remote kube-apiserver to reach us on.

pkg/envtest/webhook.go
```diff
@@ -137,13 +140,19 @@ func modifyWebhook(webhook map[string]interface{}, caData []byte, hostPort strin
 }
 
 func (o *WebhookInstallOptions) generateHostPort() (string, error) {
-	port, host, err := addr.Suggest()
-	if err != nil {
-		return "", fmt.Errorf("unable to grab random port for serving webhooks on: %v", err)
+	if o.LocalServingPort == 0 {
```
pwittrock, line 143: Is there any case where it is sensible to have the port be 0 and use and external name?

coderanger: Like to actually set it to 0 in the OS? I don't think so, the reason for this check is the reverse, allowing passing in a fixed port if you are willing to take ownership of any collisions (ex. running just one suite in a CI job) while leaving the default still be "pick me a random port" like the OS would do, but without reuse collisions.

Issue #601, opened 2019-09-12
controller-runtime versioning


joelanford:
At the community meeting on 9/11/2019, there was a discussion about how to version controller-runtime to handle scenarios around dependency changes:
* How to handle breaking changes in controller-runtime APIs introduced by controller-runtime contributors or by breaking changes in dependencies (e.g. k8s.io/client-go)
* How to handle no breaking changes in controller-runtime APIs but possible user-facing breaking changes due to upgrades of controller-runtime's transitive dependencies
* Can we backport fixes to versions of controller-runtime that support older versions of Kubernetes?

My understanding of semver is that we need to only consider the first scenario when choosing a version number after dependency changes.

My two cents - if the controller-runtime API changes are backwards-compatible, I'm not sure it makes sense to try to ALSO consider compatibility of all transitive dependencies -- that's what the version numbers of the transitive dependencies are for.

If we want to be helpful, we could add a statement to the release notes when we update dependencies to let users know that they _might_ require code changes if they're using those dependencies directly.

With these scenarios in mind, I'd propose a slight change to the versioning scheme while we're pre-1.0:
* Bump the minor version when we add features, make breaking changes, and/or update Kubernetes versions to a new minor version (e.g. go from 1.14 to 1.15)
* Bump the patch version only for patches, backports, etc.

post-1.0, I would propose:
* Bump the major version when breaking changes are introduced in controller-runtime APIs
* Bump the minor version when we add backwards-compatible changes or make backwards compatible dependency updates, where we consider only controller-runtime API compatibility.
* Bump the patch version only for patches, backports, etc.

Thoughts?

/cc @DirectXMan12 @shawn-hurley @justinsb @droot

shawn-hurley:
> How to handle breaking changes in controller-runtime APIs introduced by controller-runtime contributors or by breaking changes in dependencies (e.g. k8s.io/client-go)


There was the potential talk of a generic library that does not depend on k8s.io/api or k8s.io/apimachinery in the apimachinery sig meeting. I would like to explore if we could either help with this work or do this work? I think this would be very helpful for everyone. @DirectXMan12 Have you seen anything more about this?

This would allow us to not have to follow kubernetes release IIUC and allow this library and others to have their own release's. We may still want a compatibility matrix if something drastically changes in the underlying library but I think this might be the place to spend our cycles?

What do folks think?

DirectXMan12:
[...]

PR #1266, opened 2020-11-17
[:warning:] bump k8s dependencies to 1.19 in 0.6.x release

varshaprasad96:
This PR bumps k8s dependencies in 0.6.x release of controller-runtime to 1.19.4. 

Reason:
The current v2 plugin in Kubebuilder uses controller-runtime 0.6.x. If its possible to make patch releases of controller-runtime having upper versions of Kubernetes, it would be helpful to upgrade v2 plugin to use latest releases of k8s without any breaking changes which controller-runtime v0.7.0 may cause. This would also be helpful to update [Operator SDK ](https://github.com/operator-framework/operator-sdk) to support latest releases of Kubernetes, without bumping controller-runtime to 0.7.0.

cc: @joelanford


varshaprasad96:
@joelanford @alvaroaleman @vincepri - Can you please have a look on whether we can upgrade k8s dependencies for minor version releases of controller-runtime.

coderanger:
This would be a compat break, so I don't think we should do it.

vincepri:
/hold

1.19 is going to come in v0.7.x

vincepri:
@varshaprasad96 We usually wait for the next minor release before updating, the current mainline branch has v0.19.x, but we usually don't backport minor kubernetes version bumps because it could cause breaking changes to clients using controller runtime

joelanford:
SDK maintainers were discussing various options for getting to 1.19 and then 1.20 shortly after it GAs. This was one option that we wanted to explore, even if just for completeness sake. I turned up two potential breaking API changes in my quick perusal of the v1.19 notes (there could be more):

> The unused series.state field, deprecated since v1.14, is removed from the events.k8s.io/v1beta1 and v1 Event types. (#90449, @wojtek-t) [SIG Apps]

> The alpha DynamicAuditing feature gate and auditregistration.k8s.io/v1alpha1 API have been removed and are no longer supported. (#91502, @deads2k) [SIG API Machinery, Auth and Testing]

Also just to be transparent if we _did_ go down this path, the intention would also be to bump v0.6.x to Kubernetes 1.20 provided there are no API breaks in that release either. The benefit with this approach would be that users could use newer versions of Kubernetes without having to migrate their projects to handle breaking changes in controller-runtime. Downside of course is the aforementioned potential for API breaks.

---

Another option would be what's been suggested here and what we've traditionally done: minor versions of controller-runtime that have breaking changes and/or Kubernetes version bumps. I think the implication is that there could be up to 4 new kubebuilder go plugin versions every year to handle breaking changes in controller-runtime if we always allow breaking changes on `master` (do we have a policy on that?).

The v0.7.0 (for 1.19) and then subsequent v0.8.0 (for 1.20) option requires a new version of the Kubebuilder go plugin because of the breaking changes in controller runtime.

Right now that would be the v3-alpha plugin. IMO, before we can make that the default in Kubebuilder (and Operator SDK) and start migrating users, we need to make that plugin stable so users can migrate and trust that it won't break.

There are some outstanding items in the [kubebuilder v3 plugin milestone](https://github.com/kubernetes-sigs/kubebuilder/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+milestone%3A%22v3%2B+plugin%22) and possibly other breaking changes we need to make to project version `3-alpha` depending on whether we want to get [phase 2 plugins](https://github.com/kubernetes-sigs/kubebuilder/pull/1801) in before v3 is stable. But perhaps that's held until project version 4 or we find a backwards compatible way to introduce it to project version 3 🤷‍♂️ 

**TL;DR:**
Anyway, this is a very longwinded way of saying that we don't have a great story around getting releases out of controller-runtime and kubebuilder to quickly add support for Kubernetes releases when they GA, and [I think we need one](https://github.com/kubernetes-sigs/controller-runtime/issues/601). 🙂

joelanford:
Ah. Looks like there are also some breaking changes in controller-runtime APIs due to the logr change.

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1266/pull-controller-runtime-apidiff-release-0-6/1328679793574547456#1:build-log.txt%3A71

Even if we were good with all of the other changes that bumping v0.6 to 1.19 would bring, I think it would be a non-starter to directly break a controller-runtime API in a patch release.

@varshaprasad96 I think this means we need to close this PR and make a push to:
1. get `v0.7.0` released with v1.19 support
2. bump `master` to k8s v1.20 (IMO, we can do an initial PR that bumps to a 1.20 pre-release, so that its a quick PR to bump to 1.20 GA as soon as it drops)
3. release `v0.8.0` with v1.20 support
4. update kubebuilder v3-alpha plugin to use C-R v0.8.0
5. finalize kubebuilder v3-alpha plugin and rename to v3
6. change kubebuilder default plugin to v3
7. release kubebuilder v3

Thoughts?

varshaprasad96:
Closing this PR in favor of #1268 which bumps master to k8s 1.20.

PR #1274, opened 2020-11-25
✨ Add UnsafeDisableDeepCopy into cache option to avoid deep copy during get/list

### LINKED ISSUE #1235: CacheReader: DeepCopyObject on every Get/List method calls
We are actively using `CacheReader` in our project and we faced a memory issue caused by `DeepCopyObject` inside `Get` and `List` methods. 

CacheReader's `List` calls Indexer's `List` and then calls `DeepCopyObject` for every object:
```go 
func (c *CacheReader) List(_ context.Context, out runtime.Object, opts ...client.ListOption) error {
...
    objs = c.indexer.List()
    ...
    runtimeObjs := make([]runtime.Object, 0, len(objs))
    for _, item := range objs {
    ...
        outObj := obj.DeepCopyObject()
        outObj.GetObjectKind().SetGroupVersionKind(c.groupVersionKind)
        runtimeObjs = append(runtimeObjs, outObj)
    }
...
}
```

But `Indexer` implemented by `cache` has the following doc-comment for method `List`:
```go
// List returns a list of all the items.
// List is completely threadsafe as long as you treat all items as immutable.
func (c *cache) List() []interface{} {
	return c.cacheStorage.List()
}
```

And that's exactly our scenario. We call `CacheReader.List` from different goroutines and everywhere treat the result items as immutable. 

Is there any way to avoid `DeepCopyObject` and have this behavior parametrized?

FillZpp:
Signed-off-by: Siyu Wang <FillZpp.pub@gmail.com>

Add UnsafeDisableCacheDeepCopy into cache option to avoid deep copy during list.

fixes #1235 

@vincepri @alvaroaleman  I fully understand why the CacheReader calls `DeepCopyObject` for each object by default, the directly modification to underlying cache object is terrible. However, we can also provide an option for those people who really need to avoid deep copy.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -113,6 +113,12 @@ type Options struct {
 	// [1] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selector
 	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
 	SelectorsByObject SelectorsByObject
+
+	// UnsafeDisableDeepCopyByObject indicates not to deep copy objects during get or
+	// list objects per GVK at the specified object.
+	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
+	// otherwise you will mutate the object in the cache.
+	UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
 }
 
 var defaultResyncTime = 10 * time.Hour
@@ -127,7 +133,11 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace, selectorsByGVK)
+	disableDeepCopyByGVK, err := convertToDisableDeepCopyByGVK(opts.UnsafeDisableDeepCopyByObject, opts.Scheme)
+	if err != nil {
+		return nil, err
+	}
+	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace, selectorsByGVK, disableDeepCopyByGVK)
 	return &informerCache{InformersMap: im}, nil
 }
 
@@ -136,6 +146,8 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 // SelectorsByObject
 // WARNING: if SelectorsByObject is specified. filtered out resources are not
 //          returned.
+// WARNING: if UnsafeDisableDeepCopy is enabled, you must DeepCopy any object
+//          returned from cache get/list before mutating it.
 func BuilderWithOptions(options Options) NewCacheFunc {
 	return func(config *rest.Config, opts Options) (Cache, error) {
 		if opts.Scheme == nil {
@@ -151,6 +163,7 @@ func BuilderWithOptions(options Options) NewCacheFunc {
 			opts.Namespace = options.Namespace
 		}
 		opts.SelectorsByObject = options.SelectorsByObject
+		opts.UnsafeDisableDeepCopyByObject = options.UnsafeDisableDeepCopyByObject
 		return New(config, opts)
 	}
 }
@@ -188,4 +201,31 @@ func convertToSelectorsByGVK(selectorsByObject SelectorsByObject, scheme *runtim
 		selectorsByGVK[gvk] = selector
 	}
 	return selectorsByGVK, nil
+}
+
+// DisableDeepCopyByObject associate a client.Object's GVK to disable DeepCopy during get or list from cache.
+type DisableDeepCopyByObject map[client.Object]bool
+
+var _ client.Object = &ObjectAll{}
+
+// ObjectAll is the argument to represent all objects' types.
+type ObjectAll struct {
+	client.Object
+}
+
+func convertToDisableDeepCopyByGVK(disableDeepCopyByObject DisableDeepCopyByObject, scheme *runtime.Scheme) (internal.DisableDeepCopyByGVK, error) {
+	disableDeepCopyByGVK := internal.DisableDeepCopyByGVK{}
+	for obj, disable := range disableDeepCopyByObject {
+		switch obj.(type) {
+		case ObjectAll, *ObjectAll:
+			disableDeepCopyByGVK[internal.GroupVersionKindAll] = disable
+		default:
+			gvk, err := apiutil.GVKForObject(obj, scheme)
+			if err != nil {
+				return nil, err
+			}
+			disableDeepCopyByGVK[gvk] = disable
+		}
+	}
+	return disableDeepCopyByGVK, nil
 }
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -19,6 +19,8 @@ package cache_test
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"sort"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -83,13 +85,16 @@ func deletePod(pod client.Object) {
 }
 
 var _ = Describe("Informer Cache", func() {
-	CacheTest(cache.New)
+	CacheTest(cache.New, cache.Options{})
 })
 var _ = Describe("Multi-Namespace Informer Cache", func() {
-	CacheTest(cache.MultiNamespacedCacheBuilder([]string{testNamespaceOne, testNamespaceTwo, "default"}))
+	CacheTest(cache.MultiNamespacedCacheBuilder([]string{testNamespaceOne, testNamespaceTwo, "default"}), cache.Options{})
+})
+var _ = Describe("Informer Cache without DeepCopy", func() {
+	CacheTest(cache.New, cache.Options{UnsafeDisableDeepCopyByObject: cache.DisableDeepCopyByObject{cache.ObjectAll{}: true}})
 })
 
-func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (cache.Cache, error)) {
+func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (cache.Cache, error), opts cache.Options) {
 	Describe("Cache test", func() {
 		var (
 			informerCache       cache.Cache
@@ -139,7 +144,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			knownPod6.GetObjectKind().SetGroupVersionKind(podGVK)
 
 			By("creating the informer cache")
-			informerCache, err = createCacheFunc(cfg, cache.Options{})
+			informerCache, err = createCacheFunc(cfg, opts)
 			Expect(err).NotTo(HaveOccurred())
 			By("running the cache and waiting for it to sync")
 			// pass as an arg so that we don't race between close and re-assign
@@ -228,18 +233,20 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					}
 				})
 
-				It("should be able to list objects with GVK populated", func() {
-					By("listing pods")
-					out := &corev1.PodList{}
-					Expect(informerCache.List(context.Background(), out)).To(Succeed())
+				if !isPodDisableDeepCopy(opts) {
+					It("should be able to list objects with GVK populated", func() {
+						By("listing pods")
+						out := &corev1.PodList{}
+						Expect(informerCache.List(context.Background(), out)).To(Succeed())
 
-					By("verifying that the returned pods have GVK populated")
-					Expect(out.Items).NotTo(BeEmpty())
-					Expect(out.Items).Should(SatisfyAny(HaveLen(5), HaveLen(6)))
-					for _, p := range out.Items {
-						Expect(p.GroupVersionKind()).To(Equal(corev1.SchemeGroupVersion.WithKind("Pod")))
-					}
-				})
+						By("verifying that the returned pods have GVK populated")
+						Expect(out.Items).NotTo(BeEmpty())
+						Expect(out.Items).Should(SatisfyAny(HaveLen(5), HaveLen(6)))
+						for _, p := range out.Items {
+							Expect(p.GroupVersionKind()).To(Equal(corev1.SchemeGroupVersion.WithKind("Pod")))
+						}
+					})
+				}
 
 				It("should be able to list objects by namespace", func() {
 					By("listing pods in test-namespace-1")
@@ -255,21 +262,53 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					}
 				})
 
-				It("should deep copy the object unless told otherwise", func() {
-					By("retrieving a specific pod from the cache")
-					out := &corev1.Pod{}
-					podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
-					Expect(informerCache.Get(context.Background(), podKey, out)).To(Succeed())
+				if !isPodDisableDeepCopy(opts) {
+					It("should deep copy the object unless told otherwise", func() {
+						By("retrieving a specific pod from the cache")
+						out := &corev1.Pod{}
+						podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
+						Expect(informerCache.Get(context.Background(), podKey, out)).To(Succeed())
 
-					By("verifying the retrieved pod is equal to a known pod")
-					Expect(out).To(Equal(knownPod2))
+						By("verifying the retrieved pod is equal to a known pod")
+						Expect(out).To(Equal(knownPod2))
 
-					By("altering a field in the retrieved pod")
-					*out.Spec.ActiveDeadlineSeconds = 4
+						By("altering a field in the retrieved pod")
+						*out.Spec.ActiveDeadlineSeconds = 4
 
-					By("verifying the pods are no longer equal")
-					Expect(out).NotTo(Equal(knownPod2))
-				})
+						By("verifying the pods are no longer equal")
+						Expect(out).NotTo(Equal(knownPod2))
+					})
+				} else {
+					It("should not deep copy the object if UnsafeDisableDeepCopy is enabled", func() {
+						By("getting a specific pod from the cache twice")
+						podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
+						out1 := &corev1.Pod{}
+						Expect(informerCache.Get(context.Background(), podKey, out1)).To(Succeed())
+						out2 := &corev1.Pod{}
+						Expect(informerCache.Get(context.Background(), podKey, out2)).To(Succeed())
+
+						By("verifying the pointer fields in pod have the same addresses")
+						Expect(out1).To(Equal(out2))
+						Expect(reflect.ValueOf(out1.Labels).Pointer()).To(BeIdenticalTo(reflect.ValueOf(out2.Labels).Pointer()))
+
+						By("listing pods from the cache twice")
+						outList1 := &corev1.PodList{}
+						Expect(informerCache.List(context.Background(), outList1, client.InNamespace(testNamespaceOne))).To(Succeed())
+						outList2 := &corev1.PodList{}
+						Expect(informerCache.List(context.Background(), outList2, client.InNamespace(testNamespaceOne))).To(Succeed())
+
+						By("verifying the pointer fields in pod have the same addresses")
+						Expect(len(outList1.Items)).To(Equal(len(outList2.Items)))
+						sort.SliceStable(outList1.Items, func(i, j int) bool { return outList1.Items[i].Name <= outList1.Items[j].Name })
+						sort.SliceStable(outList2.Items, func(i, j int) bool { return outList2.Items[i].Name <= outList2.Items[j].Name })
+						for i := range outList1.Items {
+							a := &outList1.Items[i]
+							b := &outList2.Items[i]
+							Expect(a).To(Equal(b))
+							Expect(reflect.ValueOf(a.Labels).Pointer()).To(BeIdenticalTo(reflect.ValueOf(b.Labels).Pointer()))
+						}
+					})
+				}
 
 				It("should return an error if the object is not found", func() {
 					By("getting a service that does not exists")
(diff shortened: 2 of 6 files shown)


alvaroaleman:
@FillZpp what is the usecase here that requires listing large numbers of objects and can not be covered via an Index as described e.G. [here](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup)?

FillZpp:
> @FillZpp what is the usecase here that requires listing large numbers of objects and can not be covered via an Index as described e.G. [here](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup)?

In Alibaba, we have lots of K8s clusters and some of them have even more than 200k Pods in a single cluster.
In such cluster, operators like [OpenKruise](https://github.com/openkruise/kruise) call `ListWatch` for resources including Pods. So there may be multiple controllers and reconcile workers in one operator process that list their Pods simultaneously, which brings huge pressure for cpu/memory and sometimes leads to OOM.

The default `DeepCopy` to avoid cache pollution is good for those people who are unfamiliar with controller/informer, but also prejudicial to performance in large-scale clusters. We should provide an option for users that really need to disable this behavior. @alvaroaleman

alvaroaleman:
> In Alibaba, we have lots of K8s clusters and some of them have even more than 200k Pods in a single cluster.
In such cluster, operators like OpenKruise call ListWatch for resources including Pods.

This change wont reduce the memory usage of the ListWatch

> So there may be multiple controllers and reconcile workers in one operator process that list their Pods simultaneously, which brings huge pressure for cpu/memory and sometimes leads to OOM.

I would expect those lists to filter by either using a LabelSelector or an Index

> The default DeepCopy to avoid cache pollution is good for those people who are unfamiliar with controller/informer, but also prejudicial to performance in large-scale clusters. We should provide an option for users that really need to disable this behavior.

I disagree, prejudical to performance for large-scale clusters is it to always filter lists via labels or indexers. I wouldn't trust myself to always correctly do the deepcopy. so IMHO this is a huge footgun, even for ppl that are very familiar with controller-runtime.

FillZpp:
> would expect those lists to filter by either using a LabelSelector or an Index

Yes, those lists do have label or field selectors. Let me explain it.

Take OpenKruise for example, currently there are seven different controllers in OpenKruise and each of them have multiple workers. In these workload controllers, I take `CloneSet` and `SidecarSet` for example:

- `CloneSet` is a workload for stateless application, just like Deployment/ReplicaSet.  When reconcile a CloneSet, the worker  should list all Pods belongs to this CloneSet using ownerReference field selector. For multiple workers, they may list their Pods simultaneously and usually there are thousands of Pods in a single CloneSet.
- `SidecarSet` is a workload that injects sidecar into Pods during creating. Our mesh and ops sidecars injected into all application Pods are all managed by SidecarSet. So each reconcile worker should list all Pods that have sidecar injected by it using label selector, and also multiple workers may list their Pods simultaneously.

These examples are so mush, I can't describe all of them. In one word, using label/field selectors is recommended, but it can not avoid the pressure caused by DeepCopy.

> I disagree, prejudical to performance for large-scale clusters is it to always filter lists via labels or indexers. I wouldn't trust myself to always correctly do the deepcopy. so IMHO this is a huge footgun, even for ppl that are very familiar with controller-runtime.

I disagree. If so, why doesn't the informer or lister in client-go call DeepCopy for each object? The kube-controller-manager/kube-scheduler and other controllers based on client-go all list objects without DeepCopy. If someone adds DeepCopy into client-go lister and these controllers, it will degrade the performance of K8s.

I agree with the default DeepCopy, but users should have the right to choose not to use it.
@alvaroaleman

alvaroaleman:
> The kube-controller-manager/kube-scheduler and other controllers based on client-go all list objects without DeepCopy. If someone adds DeepCopy into client-go lister and these controllers, it will degrade the performance of K8s.

Upstream kube is also very careful about using the cache mutation detector in integration/e2e tests.
Do you have any data on how much memory exactly this saves you?

Generally speaking, I still think that this is likely to cause subtile bugs ppl will not detect and that the memory gain is too little to be worth introducing this. I would assume that if you can afford to run a cluster with 200k pods, you can afford to give your CM 2 gigs more ram.

If you want to discuss this, could you join the biweekly meeting this Thursday?

vincepri:
If we do allow the cache to expose internal objects, we should opt for the option to be something like `UnsafeDisableCacheDeepCopy` or something similar.

A similar use case like this we encountered in Cluster API has driven us to use metadata only watches, and use a live reader (with something like https://github.com/kubernetes-sigs/controller-runtime/pull/1249) when querying for the objects

FillZpp:
@alvaroaleman  To hide sensitive data, I can only show part of the heep profile, which was dumped from one of our extend controller in the large-scale cluster. And you can find the DeepCopy in CacheReader costs huge heap memory.

image

FillZpp:
> If we do allow the cache to expose internal objects, we should opt for the option to be something like UnsafeDisableCacheDeepCopy or something similar.

Good idea. I have updated it.

FillZpp:
Hi, we don't meant to close this, right? I'm still looking for someone to review it.
WDYT?
/cc @vincepri @alvaroaleman

coderanger:
My vote is that this is handled well enough by other trapdoors (indexed lists, non-caching clients) that we don't need to add it.

pkg/client/options.go
```diff
@@ -521,6 +524,18 @@ func (c Continue) ApplyToList(opts *ListOptions) {
 	opts.Continue = string(c)
 }
 
+// UnsafeDisableCacheDeepCopy indicates not to deep copy objects during list objects in cache reader.
+type UnsafeDisableCacheDeepCopy struct {
+	// if IgnoreGVK is true, cache reader will always list without deep copy.
+	// Otherwise, it will only avoid deep copy for objects with correct groupVersionKind.
+	IgnoreGVK bool
```
alvaroaleman, line 531: I don't understand this option, the cache is gvk indexed anyways?

pkg/client/options.go
```diff
@@ -521,6 +524,18 @@ func (c Continue) ApplyToList(opts *ListOptions) {
 	opts.Continue = string(c)
 }
 
+// UnsafeDisableCacheDeepCopy indicates not to deep copy objects during list objects in cache reader.
```
alvaroaleman, line 527:
```suggestion
// UnsafeDisableCacheDeepCopy indicates not to deep copy objects during list objects in cache reader.
// Be very careful with this, when enabled you must DeepCopy any object before mutating it, otherwise you
// will mutate the object in the cache.
```

alvaroaleman (commented):
> My vote is that this is handled well enough by other trapdoors (indexed lists, non-caching clients) that we don't need to add it.

I think it can make a difference when there are lots of objects that also have lots of changes. The option has `Unsafe` in the name so that should sufficiently warn ppl IMHO.

@vincepri @joelanford WDYT?

joelanford:
In general, I think an option with `Unsafe` in the name to avoid DeepCopy is a reasonable compromise.

However one thing that's a minor concern for me is that we're adding a field to `ListOptions` in the `client` package, which at best leaks an implementation detail of the client and at worst is a no-op if the client in use does not have a backing cache.

Would it be possible (and @FillZpp would it meet your use case) to keep this scoped to a specific cache implementation such that all cache reads would skip `DeepCopy()` calls? This option would probably require something like the `GVKCaches` composite cache proposed in #1591, which allows separate cache implementations per GVK.

If this option works, we could avoid a leaky option on a client call, and it would hide this unsafety in a cache implementation that a developer would have to specifically opt into.

pkg/client/options.go
FillZpp: done

pkg/client/options.go
FillZpp: The `IgnoreGVK` is used [here](https://github.com/kubernetes-sigs/controller-runtime/pull/1274/files#diff-7e7ff90d22397cbc25370857a4c26baa1062866c5c9581e440d1e37603e45011R169). For I see the objects will be `SetGroupVersionKind` in [previous logic](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L154), it seems some objects have no GVK when they are listed from informer.
So if this `IgnoreGVK` is true, the objects will still be deepcopy if they have no GVK when listed from informer, because they need to be `SetGroupVersionKind`.
I don’t know if I explain it clearly...

FillZpp:
> However one thing that's a minor concern for me is that we're adding a field to ListOptions in the client package, which at best leaks an implementation detail of the client and at worst is a no-op if the client in use does not have a backing cache.

@joelanford Thanks! I understand your concern and I think it makes sense :)  
However, the `disable deep copy` is used in `cache reader` instead of `informer cache`. So I don't think it is a good idea to implement a new cache type.

How about add it as an optional field in `cache.Options` and finally transmit into `CacheReader`? So developers can use `cache.BuilderWithOptions` function to set this field when initialize.

@alvaroaleman @joelanford @vincepri WDYT?

joelanford:
> How about add it as an optional field in cache.Options and finally transmit into CacheReader? So developers can use cache.BuilderWithOptions function to set this field when initialize.

Yeah, that's definitely more along the lines of what I had in mind.

pkg/client/options.go
alvaroaleman: This is a bug in the deserialization: https://github.com/kubernetes/kubernetes/issues/80609

We work around this in some places by manually setting GVK.

IMHO this option should not exist, as it is not controllable by users if TypeMeta is filled. We should just ignore TypeMeta intead.

FillZpp:
1. I have move `UnsafeDisableCacheDeepCopy` into `cache.Options`, which is set by developers using `cache.BuilderWithOptions`. The `Unsafe` in the name should sufficiently warn ppl.
2. The internal fields are `disableDeepCopy` as developers could not set them and I suppose to simplify the name.
3. Added some unit tests to verify the addresses of objects got from cache should be equal.

PTAL @alvaroaleman @joelanford

FillZpp:
Please take a look when you have time, thanks. @alvaroaleman @vincepri @joelanford

pkg/cache/internal/informers_map.go
```diff
@@ -65,6 +66,7 @@ func newSpecificInformersMap(config *rest.Config,
 		createListWatcher: createListWatcher,
 		namespace:         namespace,
 		selectors:         selectors,
+		disableDeepCopy:   disableDeepCopy,
```
alvaroaleman, line 69: Generally this looks good, thank you for doing this change.

Only thing I was wondering is if it would make sense to key this by object type (maybe with a `*` wildcard to make it global). WDYT @vincepri @joelanford ?

pkg/cache/cache_test.go
```diff
@@ -271,6 +280,35 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+					}
+					By("getting a specific pod from the cache twice")
+					podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
+					out1 := &corev1.Pod{}
+					Expect(informerCache.Get(context.Background(), podKey, out1)).To(Succeed())
+					out2 := &corev1.Pod{}
+					Expect(informerCache.Get(context.Background(), podKey, out2)).To(Succeed())
+
+					By("verifying the pods have the same pointer addresses")
+					Expect(fmt.Sprintf("%p", out1)).NotTo(Equal(fmt.Sprintf("%p", out2)))
```
joelanford, line 295: Nit: here and elsewhere use [`BeIdenticalTo`](https://onsi.github.io/gomega/#beidenticaltoexpected-interface) instead of string comparisons of pointer values.

pkg/cache/cache_test.go
```diff
@@ -229,6 +232,9 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 				})
 
 				It("should be able to list objects with GVK populated", func() {
+					if opts.UnsafeDisableCacheDeepCopy {
+						Skip("skip for UnsafeDisableCacheDeepCopy enabled")
+					}
```
joelanford, line 237: I'm not sure what's better. This use of `Skip` or wrapping the entire `It` in an if statement to prevent it from being declared at all in these cases.

Based on ginkgo docs:
> By default, Ginkgo will print out a description for each skipped spec. You can suppress this by setting the --noisySkippings=false flag.

That makes me lean toward an outer if statement around the `It`. Since we'll get test output will _always_ contain these intentionally skipped tests, and we likely _do_ want to keep noisy skippings enabled to signal.

Anyone else have thoughts here about what's more idiomatic or preferred?

pkg/cache/internal/informers_map.go
FillZpp: I'm ok with that. The type has to be defined in `pkg/client` for both `pkg/cache` and `pkg/cache/internal` need to use it.
WDYT?

FillZpp: Updated. I agree with that, for we might need to add some optional policies for it in future.

pkg/cache/cache_test.go
FillZpp: done

pkg/cache/cache_test.go
FillZpp: Yep, I have moved the `It` into an if statement and there is no more `skip` in ginkgo test. I think this is better.

pkg/cache/cache.go
```diff
@@ -113,6 +116,11 @@ type Options struct {
 	// [1] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selector
 	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
 	SelectorsByObject SelectorsByObject
+
+	// UnsafeDisableCacheDeepCopy indicates not to deep copy objects during get or list objects.
+	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
+	// otherwise you will mutate the object in the cache.
+	UnsafeDisableCacheDeepCopy *DisableCacheDeepCopy
```
joelanford, line 123: Can this just be a `bool` now?

Also name nit: since this is in the cache package on the `cache.Options` struct, I don't think we need the word `Cache` in the name.

```suggestion
	// UnsafeDisableDeepCopy indicates not to deep copy objects during get or list objects.
	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
	// otherwise you will mutate the object in the cache.
	UnsafeDisableDeepCopy bool
```

pkg/cache/internal/informers_map.go
joelanford: @alvaroaleman I also considered the need for this to be GVK-specific. My [comment earlier](https://github.com/kubernetes-sigs/controller-runtime/pull/1274#issuecomment-887974926) touches on that:
> This option would probably require something like the GVKCaches composite cache proposed in #1591, which allows separate cache implementations per GVK.

It's probably do-able to make this option work in the informersMap on a per-GVK basis, but I feel like there are going to be other use cases around GVK-specific cache semantics (e.g. watch GVKA in namespace `foo`, GVKB in namespace `bar` and GVKC cluster-wide) that it may be better to keep this minimal.

That said, I see we already have selectors keyed by GVK, so maybe its best to do the same here and re-evaluate where GVK-specifics should live in a separate feature?

pkg/cache/cache_test.go
```diff
@@ -721,34 +796,71 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+						out1.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"})
+						Expect(informerCache.Get(context.Background(), podKey, out1)).To(Succeed())
+						out2 := &metav1.PartialObjectMetadata{}
+						out2.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"})
+						Expect(informerCache.Get(context.Background(), podKey, out2)).To(Succeed())
+
+						By("verifying the pods have the same pointer addresses")
+						By("verifying the pointer fields in pod have the same addresses")
+						Expect(out1).To(Equal(out2))
+						Expect(reflect.ValueOf(out1.Labels).Pointer()).To(BeIdenticalTo(reflect.ValueOf(out2.Labels).Pointer()))
```
joelanford, line 842: non-blocking nit: pretty sure there's no need to actually fetch the pointer for `BeIdenticalTo`. This could be simplified to:

```suggestion
						Expect(out1.Labels).To(BeIdenticalTo(out2.Labels))
```

pkg/cache/internal/cache_reader.go
```diff
@@ -76,9 +84,14 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out client.Ob
...
 
-	// deep copy to avoid mutating cache
-	// TODO(directxman12): revisit the decision to always deepcopy
-	obj = obj.(runtime.Object).DeepCopyObject()
+	if c.disableDeepCopy != nil {
+		// skip deep copy which might be unsafe
+		// you must DeepCopy any object before mutating it outside
+	} else {
+		// deep copy to avoid mutating cache
+		// TODO(directxman12): revisit the decision to always deepcopy
```
joelanford, line 92: Can probably remove this comment now.
```suggestion
```

pkg/cache/internal/cache_reader.go
```diff
@@ -150,8 +165,15 @@ func (c *CacheReader) List(_ context.Context, out client.ObjectList, opts ...cli
 			}
 		}
 
-		outObj := obj.DeepCopyObject()
-		outObj.GetObjectKind().SetGroupVersionKind(c.groupVersionKind)
+		var outObj runtime.Object
+		if c.disableDeepCopy != nil {
+			// skip deep copy which might be unsafe
+			// you must DeepCopy any object before mutating it outside
+			outObj = obj
```
joelanford, line 172: Do we need to set the GVK here?

pkg/cache/cache_test.go
```diff
@@ -255,21 +262,53 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+						podKey := client.ObjectKey{Name: "test-pod-2", Namespace: testNamespaceTwo}
+						out1 := &corev1.Pod{}
+						Expect(informerCache.Get(context.Background(), podKey, out1)).To(Succeed())
+						out2 := &corev1.Pod{}
+						Expect(informerCache.Get(context.Background(), podKey, out2)).To(Succeed())
+
+						By("verifying the pointer fields in pod have the same addresses")
+						Expect(out1).To(Equal(out2))
+						Expect(reflect.ValueOf(out1.Labels).Pointer()).To(BeIdenticalTo(reflect.ValueOf(out2.Labels).Pointer()))
+
```
joelanford, line 293: Can we add another expectation that verifies the GVK is set on the returned objects after `Get()` and `List()` when deep copying is disabled?

joelanford (commented):
This is looking pretty good to me, just a few more comments. Thanks for working on this!

pkg/cache/cache_test.go
FillZpp: I tried `Expect(out1.Labels).To(BeIdenticalTo(out2.Labels))` before, but it failed. I'm not sure why this happened, the `out1.Labels` and `out2.Labels` do have the same pointer addresses.

pkg/cache/internal/cache_reader.go
FillZpp: Set the GVK here might modify the object in Informer. We may just ignore it according to the [discussion earlier](https://github.com/kubernetes-sigs/controller-runtime/pull/1274#discussion_r678615164).

pkg/cache/cache.go
FillZpp: > Can this just be a `bool` now?

I changed the type from `bool` to `struct` in case we might need to add some optional policies for it in future.

> Also name nit: since this is in the cache package on the cache.Options struct, I don't think we need the word Cache in the name.

Have removed `Cache` in the name.

pkg/cache/internal/cache_reader.go
FillZpp: done

pkg/cache/cache_test.go
FillZpp: When deep copying is disabled, we could not ensure the GVK is set on the returned objects.

pkg/cache/internal/informers_map.go
alvaroaleman: > That said, I see we already have selectors keyed by GVK, so maybe its best to do the same here and re-evaluate where GVK-specifics should live in a separate feature?

Yeah, that was what I thought, because otherwise we end up in a situation where ppl have to set different options on different levels and that will be pretty confusing. Discussing what option we want to be configurable by GVK how and where in a separate issue is a good idea :+1:

pkg/cache/cache_test.go
joelanford: Hmm, alright 🤷‍♂️. Thanks for checking that out!

pkg/cache/cache.go
joelanford: Not sure if @vincepri @alvaroaleman and @varshaprasad96 have opinions, but I think I'd prefer to make this a `bool` since we have no options right now to deal with (and may never).

If we _do_ end up needing to configure other aspects of this in the future, we could introduce an `UnsafeDisableDeepCopyOptions` struct alongside this `bool` if we wanted to avoid a breaking change.

pkg/cache/internal/cache_reader.go
joelanford: Ah got it! That earlier discussion makes more sense to me now :)

pkg/cache/cache.go
FillZpp: Alright, I have changed it back to `bool` 🤔

FillZpp:
@alvaroaleman @joelanford Thanks for all your comments. I have done all of them. PTAL again 😆

joelanford:
Looks good to me. I think the only outstanding question (in [this](https://github.com/kubernetes-sigs/controller-runtime/pull/1274#discussion_r680979294) thread) is whether we should have:
```go
UnsafeDisableDeepCopy bool
```

or 

```go
UnsafeDisableDeepCopy map[schema.GroupVersionKind]bool
```

which would align this with the `selectors SelectorsByGVK` field.

FillZpp:
> which would align this with the selectors SelectorsByGVK field.

So... How about two fields in `cache.Options`?
```
type Options {
    // ......

    // UnsafeDisableDeepCopy indicates not to deep copy objects during get or list objects.
    UnsafeDisableDeepCopy bool

    // UnsafeDisableDeepCopyByObject  indicates not to deep copy objects during
    // get or list objects per GVK at the specified object.
    UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
}

// DisableDeepCopyByObject associate a client.Object's GVK to whether it should disable deep copy.
DisableDeepCopyByObject map[client.Object]bool
```

Then developers can choose to disable all with some GVK specified to enable, or enable all with some GVK specified to disable.

WDYT @joelanford

joelanford:
It might be simpler to just define a constant:
```go
package cache
const GroupVersionKindAll = ""
```

And if it is present in the map, we would skip deep copy for all GVKs. I think this is what @alvaroaleman had in mind, and there's precendence for this in other areas of controller-runtime (e.g. [`cache.Options.Namespace`](https://github.com/kubernetes-sigs/controller-runtime/blob/c0a5babc6854d42ab37eb5323ff024eb0dc24b9d/pkg/cache/cache.go#L106-L108)) and core kube (e.g. [`corev1.NamespaceAll`](https://github.com/kubernetes/api/blob/7ee5548f30139395482912e3a26f661c21507462/core/v1/types.go#L30))

FillZpp:
Defining its type as `map[schema.GroupVersionKind]bool` requires developers to set the option with GVK, which might be a little complicated, comparing to the `SelectorsByObject` option.

However if its type is `map[client.Object]bool`, how can we define the object for all object? Do you have any idea? @joelanford

FillZpp:
@joelanford I have changed the option type to `DisableDeepCopyByObject` and `DisableDeepCopyByGVK` internal.

FillZpp:
PTAL @alvaroaleman @joelanford when you have time, thanks.

joelanford:
/approve

This looks great @FillZpp! Thanks for sticking with this!

FillZpp:
Thanks for reviewing this @joelanford .Now we need lgtm, PTAL @alvaroaleman

alvaroaleman (approved):
thank you!

PR #1279, opened 2020-11-30
:running: AddToProtobufScheme: Take AddToScheme, not Schemebulilder

alvaroaleman:
Depending on the API, the SchemeBuilder might not be exported. Since we
only care about its AddToScheme, take that as argument instead.

/assign @vincepri 
/cc @FillZpp


vincepri (approved):
/approve
/lgtm
/milestone v0.7.x

PR #1281, opened 2020-11-30
:sparkles: Add fake.NewClientBuilder to build new fake clients

vincepri:
This change allows us to iterate on the fake client without relying on
a small set of arguments passed to
NewFakeClient/NewFakeClientWithScheme.

The new builder accepts a variety of options to initialize the fake
client, including the scheme.

Because the builder seems more of a long term solution, I've marked the
older functions as deprecated in favor of NewClientBuilder.

Note: I dropped the "Fake" prefix given that it stutters with the
package name.

/assign @alvaroaleman


alvaroaleman (approved):
/lgtm

PR #1284, opened 2020-12-03
:warning: pkg/webhook/admission: upgrade v1beta1 admission types to v1

estroz:
As per [this review](https://github.com/kubernetes-sigs/controller-runtime/pull/1233#pullrequestreview-534898408) on #1233:

>currently, v1 and v1beta1 are identical [...], which means that we can safely just deserialize v1beta1 into a v1 struct and call it a day, without the duplication

This PR captures the essence of that comment.


vincepri (approved):
/lgtm

PR #1288, opened 2020-12-03
🌱 Add test that checks List works on UnstructuredList not in Scheme

vincepri:

After adding the metadata only watch and client support, the List object
was type checking on Unstructured instead of UnstructuredList.

Now! "Why were the tests passing??" one might ask, it seems that even
though our tests had some using UnstructuredLists and querying against
List, the objects we were asking (appsv1/Deployments) are core objects
that are usually included in the default client-go scheme.Scheme.

For the above reason, the internal typedClient was able to make the
request to the API server and convert the list of object back to
UnstructuredList.

/milestone v0.7.x

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -1686,6 +1686,36 @@ var _ = Describe("Client", func() {
 				close(done)
 			}, serverSideTimeoutSeconds)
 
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm

PR #1289, opened 2020-12-04
🐛 Inject manager's logger instead of internal one

timebertt:
Before this PR, the manager injected the controller-runtime-internal logger into controllers, predicates and so on, that implemented the `inject.Logger` interface. This means, controllers and so on were provided with a logger which name is set to `controller-runtime.manager`.

I found this behaviour to be confusing, especially if you provided the manager with a custom logger via `manager.Options.Logger` (e.g. with a different top-level name).

With this PR, the manager injects its own logger or the default one, if none is specified.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -38,7 +38,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
-	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	intrec "sigs.k8s.io/controller-runtime/pkg/internal/recorder"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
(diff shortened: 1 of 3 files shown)


pkg/manager/internal.go
```diff
@@ -251,7 +251,7 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 	if _, err := inject.MapperInto(cm.mapper, i); err != nil {
 		return err
 	}
-	if _, err := inject.LoggerInto(log, i); err != nil {
+	if _, err := inject.LoggerInto(cm.logger, i); err != nil {
```
alvaroaleman, line 254: What a mess this is, so the `log` variable is:


*  `logf.RuntimeLog.WithName("manager")` where
  -> `logf.RuntimeLog` is `log.Log.WithName("controller-runtime")` where
    -> `log.Log` is `NewDelegatingLogger(NullLogger{})`

And the managers logger is defaulted to `logf.Log`

So I think we should have the `manager`  default `options.Logger` to what is currently the `log` variable instead, right?

ALso, can you check if the `log` variable is used by anything else and if not, remove it and move the construction of its logger into the managers log defaulting?

/ok-to-test

timebertt: Yes, valid point.
Changed the defaulting of the manager's logger to `logf.RuntimeLog.WithName("manager")` and switched the remaining usages of the `log` variable now.
Hope I got you right, PTAL

alvaroaleman (approved):
/lgtm
/approve
thanks!

PR #1293, opened 2020-12-09
:bug: pkg/webhook/admission: set GVK on AdmissionReview responses

estroz:
Specifying multiple `admissionReviewVersions` in a webhook config with `v1` as the first version causes untyped (no GVK) `AdmissionReview` responses to fail with an error like
```
failed calling webhook "mfoo.kb.io": expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind=
```
This error results from the API server [checking response type](https://github.com/kubernetes/apiserver/blob/v0.19.4/pkg/admission/plugin/webhook/request/admissionreview.go#L58-L61) for v1 but not v1beta1. By setting the `AdmissionReview` object's GVK to the request's type if it exists or defaulting to v1, the webhook handler can patch over this problem.

/kind bug

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -27,6 +27,7 @@ import (
 	v1 "k8s.io/api/admission/v1"
 	"k8s.io/api/admission/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


vincepri (approved):
/lgtm
/assign @alvaroaleman

PR #1297, opened 2020-12-15
:bug: Cache bypass should take into account List types

vincepri:
Currently, when we create a new delegating client and we force some
types to be uncached, we can only specify client.Object(s).

When these objects are passed to the delegating client, and their GVK is
matched against the internal map, the check doesn't take into account
that the client might be issuing a List call with a ObjectList instead
of a metav1.Object.

The proposed solution is taken from the discovery rest mapper, which
does the same ~hack. While it's not perfect, probably close to all
generated objects are going to have a list that has the "List" suffix,
that when removed gives back the original GVK.

If there are other solutions, happy to change this hack.

/assign @alvaroaleman

### DIFF
--- a/pkg/client/split.go
+++ b/pkg/client/split.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	"context"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm

PR #1299, opened 2020-12-16
🐛 Do not set PatchType if patch is empty

### LINKED ISSUE #1295: v0.7.0: Handling empty patch in PatchResponseFromRaw
In admission v1, api server requires that patch and patchType are both provided or none are provided:

https://github.com/kubernetes/apiserver/blob/6d407d18a284aae4db693a8a50ca3647cc3ad429/pkg/admission/plugin/webhook/request/admissionreview.go#L67-L73

admission v1beta1 does not have this kind of requirement.

In controller-runtime, PatchResponseFromRaw sets patchType regardless of existence of patch:

https://github.com/kubernetes-sigs/controller-runtime/blob/5720da0ae6bbe744888a5f2fd246773e17c52016/pkg/webhook/admission/response.go#L94

So if patch is empty, a response contains only patchType and api server does not like it.  Webhook call fails.

To make transition from admission v1beta1 to v1 smooth, it would be nice to set patchType only when patch is not empty, like so:

```go
--- a/pkg/webhook/admission/response.go
+++ b/pkg/webhook/admission/response.go
@@ -91,7 +91,13 @@ func PatchResponseFromRaw(original, current []byte) Response {
                Patches: patches,
                AdmissionResponse: admissionv1.AdmissionResponse{
                        Allowed:   true,
-                       PatchType: func() *admissionv1.PatchType { pt := admissionv1.PatchTypeJSONPatch; return &pt }(),
+                       PatchType: func() *admissionv1.PatchType {
+                               if len(patches) == 0 {
+                                       return nil
+                               }
+                               pt := admissionv1.PatchTypeJSONPatch;
+                               return &pt
+                       }(),
                },
        }
 }
```

tatsuhiro-t:
In admission v1, API server requires that Patch and PatchType are both
provided or none are provided.  Meanwhile, admission v1beta1 does not
have this kind of requirement.

In controller-runtime, PatchResponseFromRaw sets PatchType regardless
of the existence of patch.  If patch is empty, a response contains
only PatchType and API server does not like it.  Webhook call fails.

This change fixes this issue by not setting PatchType if patch is
empty.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1295

### DIFF
--- a/pkg/webhook/admission/response.go
+++ b/pkg/webhook/admission/response.go
@@ -90,8 +90,14 @@ func PatchResponseFromRaw(original, current []byte) Response {
 	return Response{
 		Patches: patches,
 		AdmissionResponse: admissionv1.AdmissionResponse{
-			Allowed:   true,
-			PatchType: func() *admissionv1.PatchType { pt := admissionv1.PatchTypeJSONPatch; return &pt }(),
+			Allowed: true,
+			PatchType: func() *admissionv1.PatchType {
+				if len(patches) == 0 {
+					return nil
+				}
+				pt := admissionv1.PatchTypeJSONPatch
+				return &pt
+			}(),
 		},
 	}
 }


alvaroaleman (approved):
/lgtm
/approve

Thanks!

PR #1303, opened 2020-12-19
🌱 Add coderanger as an approver

coderanger:
So I can help out a bit more with PR review and general contribution.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -19,6 +19,7 @@ aliases:
   - shawn-hurley
   - joelanford
   - alvaroaleman
+  - coderanger
 
   # folks who can review and LGTM any PRs in the repo (doesn't
   # include approvers & admins -- those count too via the OWNERS


alvaroaleman:
There is no need to be an approver in order to review PRs, everyone is welcome to do so :)

In the kubebuilder repo we have a doc on how to become a reviewer/approver: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/CONTRIBUTING-ROLES.md#reviewers

coderanger:
@alvaroaleman Was requested by @pwittrock :)

pwittrock:
@alvaroaleman 

I suggested this when I saw @coderanger had been blocked on a review for a relatively simple PR.

The approver role suggestion was based on them having shown sustained contributions and domain expertise in the slack channel for the past 2+ years -- ~3K message since Aug 2018 (for comparison I'm at 342 messages).  For this reason I was comfortable with adding them as an approver.

However, I incorrectly assumed they had also made significant direct code contributions, and come close to meeting the requirements defined in the contributor ladder, so it was probably a premature suggestion.  :P

@coderanger sorry for leading you down the wrong path here.  WDYT of going through the formal process outlined above?

coderanger:
Sure, that works too. I keep most of my code changes out of controller-runtime since they are usually fairly weird, though now that https://github.com/coderanger/controller-utils is getting to be a bit more solid, I should probably figure out if there is stuff that should be moved upstream.

Issue #1122, opened 2020-08-07
DelegatingLogger seems to be a memory hog


alvaroaleman:
We have a controllermanager that doesn't use the built-in logging, which means the logger construction in c-r falls back to the default in `pkg/log/log.go` which is `NewDelegatingLogger(NullLogger{})`.

Analyzing the memory usage of this controller I noticed that logging seems to have the second highest number of heap allocations:

```
  584.03MB 17.25% 17.25%  1370.90MB 40.49%  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
  551.55MB 16.29% 33.54%   551.55MB 16.29%  sigs.k8s.io/controller-runtime/pkg/log.(*loggerPromise).WithValues
  542.76MB 16.03% 49.57%   542.76MB 16.03%  k8s.io/apimachinery/pkg/runtime.(*RawExtension).UnmarshalJSON
  399.21MB 11.79% 61.36%   413.97MB 12.23%  github.com/json-iterator/go.(*Iterator).ReadString
  297.66MB  8.79% 70.15%   297.66MB  8.79%  reflect.unsafe_NewArray
  243.51MB  7.19% 77.34%   243.51MB  7.19%  fmt.Sprintf
  189.01MB  5.58% 82.92%   740.56MB 21.87%  sigs.k8s.io/controller-runtime/pkg/log.(*DelegatingLogger).WithValues
  171.58MB  5.07% 87.99%   171.58MB  5.07%  reflect.mapassign
  100.15MB  2.96% 90.94%   100.15MB  2.96%  encoding/base64.(*Encoding).DecodeString
   60.01MB  1.77% 92.72%    60.01MB  1.77%  github.com/modern-go/reflect2.(*unsafeType).UnsafeNew
```

IMHO we should be _a lot_ more efficient in doing nothing.

/kind bug

vincepri:
This might be worth fixing

vincepri:
Reading the code a bit, I think we're expecting the caller to call SetLogger with NullLogger{} if they actually want a null logger. Seems the delegating logger's only purpose is to keep log lines around until SetLogger is called, which might never be called if someone never sets it.

I'm not sure if this is an actual problem or not, but I'm in favor of completely removing the `DelegatingLogger` and default to a plain `NullLogger{}` for all logs, or we can also default to `klogr` if we prefer, which now defaults to stderr as well.

PR #1309, opened 2020-12-24
🐛 pkg/log: Set Log to NullLogger after 30 seconds

alvaroaleman:
The delegating Logger will cause a high number of memory allocations
    when no actual Logger is set. This change makes us set a NullLogger
    after 30 seconds if none was set yet to avoid that.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1122

/assign @DirectXMan12 @vincepri

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -35,18 +35,50 @@ package log
 
 import (
 	"context"
+	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 )
 
 // SetLogger sets a concrete logging implementation for all deferred Loggers.
 func SetLogger(l logr.Logger) {
+	loggerWasSetLock.Lock()
+	defer loggerWasSetLock.Unlock()
+
+	loggerWasSet = true
 	Log.Fulfill(l)
 }
 
+// It is safe to assume that if this wasn't set within the first 30 seconds of a binaries
+// lifetime, it will never get set. The DelegatingLogger causes a high number of memory
+// allocations when not given an actual Logger, so we set a NullLogger to avoid that.
+//
+// We need to keep the DelegatingLogger because we have various inits() that get a logger from
+// here. They will always get executed before any code that imports controller-runtime
+// has a chance to run and hence to set an actual logger.
+func init() {
+	// Init is blocking, so start a new goroutine
+	go func() {
+		time.Sleep(30 * time.Second)
+		loggerWasSetLock.Lock()
+		defer loggerWasSetLock.Unlock()
+		if !loggerWasSet {
+			Log.Fulfill(NullLogger{})
+		}
+	}()
+}
+
+var (
+	loggerWasSetLock sync.Mutex
+	loggerWasSet     bool
+)
+
 // Log is the base logger used by kubebuilder.  It delegates
-// to another logr.Logger.  You *must* call SetLogger to
-// get any actual logging.
+// to another logr.Logger. You *must* call SetLogger to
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -35,18 +35,50 @@ package log
...
+		defer loggerWasSetLock.Unlock()
+		if !loggerWasSet {
+			Log.Fulfill(NullLogger{})
+		}
+	}()
+}
+
+var (
+	loggerWasSetLock sync.Mutex
+	loggerWasSet     bool
```
vincepri, line 74: We could also use an atomic bool?

pkg/log/log.go
```diff
@@ -35,18 +35,50 @@ package log
...
+// here. They will always get executed before any code that imports controller-runtime
+// has a chance to run and hence to set an actual logger.
+func init() {
+	// Init is blocking, so start a new goroutine
+	go func() {
+		time.Sleep(30 * time.Second)
+		loggerWasSetLock.Lock()
+		defer loggerWasSetLock.Unlock()
+		if !loggerWasSet {
+			Log.Fulfill(NullLogger{})
```
vincepri, line 67: Do we want to print out a warning or error of some sort before transitioning the logger?

It seems that the delegating logger default might not be the best choice if it could potentially fill up memory if used for a long period of time.

We could also change direction here and default to a well known logger instead?

pkg/log/log.go
alvaroaleman: atomic bool?

pkg/log/log.go
alvaroaleman: > Do we want to print out a warning or error of some sort before transitioning the logger?

Hm good idea, but what logger should we use? Generally, this doesn't change anything unless someone changed the logger after more than 30 seconds before which I think is unlikely

> It seems that the delegating logger default might not be the best choice if it could potentially fill up memory if used for a long period of time.

See the description in the PR, we need it because various packages construct a logger via `init` which always happens before any non-controller-runtime code is executed. Without the delegating logger it is impossible to set the logger for those. It is only inefficient if no logger is set.

> We could also change direction here and default to a well known logger instead?

Sure

pkg/log/log.go
vincepri: Meant something like https://golang.org/pkg/sync/atomic/#CompareAndSwapInt64, this is really low priority though

pkg/log/log.go
vincepri: Yeah that makes sense, I was thinking when we're transitioning we could print an error to stderr via the stdlib log package

vincepri: The alternative is to just default to `klogr` and let folks use a different logger if they want to, we could do the init of the default logger early when the package loads up, which should solve the problem described above

alvaroaleman: I don't think it properly solves the problem, because right now you can set the global base logger for everything via `SetLogger` (which IMHO is convenient and useful). What you describe would only allow to override the default logger within the components.

Also, c-r doesn't use klogr anywhere that I am aware, it uses `zap` with a `logr.Logger` wrapper.

So right now the whole thing is documented as not logging anything if `SetLogger` is not called, so changing that would make this PR break compatibility, hence I would prefer to stick with the NullLogger so this can be backported. WDYT?

vincepri: That sounds good for now yeah, I was more brainstorming for later / long temr

PR #1313, opened 2020-12-31
✨ Use anonymous function for locking when serving health probes

hasheddan:
Updates serveHealthProbes to use the same anonymous function pattern
that serveMetrics does for performing operations that necessitate
acquiring the controller manager's lock.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


alvaroaleman (approved):
/lgtm
/approve

PR #1315, opened 2021-01-04
✨ introduce LabelChangedPredicate to trigger event on label change

mars1024:
Signed-off-by: Bruce Ma <brucema19901024@gmail.com>

This is a non-breaking feature :sparkles: .

Just as the title, a `LabelChangedPredicate` is introduced to trigger event on label change, if some extra specification information is carried through object's label, this predicate function will be useful.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -200,6 +200,36 @@ func (AnnotationChangedPredicate) Update(e event.UpdateEvent) bool {
 	return !reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations())
 }
 
+// LabelChangedPredicate implements a default update predicate function on label change.
+//
+// This predicate will skip update events that have no change in the object's label.
+// It is intended to be used in conjunction with the GenerationChangedPredicate, as in the following example:
+//
+// Controller.Watch(
+//		&source.Kind{Type: v1.MyCustomKind},
+// 		&handler.EnqueueRequestForObject{},
+//		predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))
+//
+// This will be helpful when object's labels is carrying some extra specification information beyond object's spec,
+// and the controller will be triggered if any valid spec change (not only in spec, but also in labels) happens.
+type LabelChangedPredicate struct {
+	Funcs
+}
+
+// Update implements default UpdateEvent filter for checking label change
+func (LabelChangedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil {
+		log.Error(nil, "Update event has no old object to update", "event", e)
+		return false
+	}
+	if e.ObjectNew == nil {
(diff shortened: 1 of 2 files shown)


pkg/predicate/predicate.go
```diff
@@ -200,6 +200,36 @@ func (AnnotationChangedPredicate) Update(e event.UpdateEvent) bool {
...
+//
+// Controller.Watch(
+//		&source.Kind{Type: v1.MyCustomKind},
+// 		&handler.EnqueueRequestForObject{},
+//		predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))
+//
+// This will be helpful when object's labels is carrying some extra specification information beyond object's spec,
+// and the controller will be triggered if any valid spec change (not only in spec, but also in labels) happens.
+type LabelChangedPredicate struct {
+	Funcs
```
alvaroaleman, line 216: Can you use  `Funcs.UpdateFunc` rather than overwriting it?

mars1024: Sorry, maybe I didn't catch your point, I'm just doing the same way like [ResourceVersionChangedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/5564be7542e5d99c77118f2faf3fa9664dc965de/pkg/predicate/predicate.go#L121) and [GenerationChangedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/5564be7542e5d99c77118f2faf3fa9664dc965de/pkg/predicate/predicate.go#L155), overriding the `Update` method and inherit the others. 

If we want to assign `Funcs.UpdateFunc`, another constructing function will be introduced for this, and I think maybe it's not easy for using.

alvaroaleman: Ah that's how you came up with it. From a user perspective it is just calling a constructor rather than using a type :)

The reason I think it's slightly better is that it might be confusing to have an embedded `UpdateFunc` described as being called for Update events that does not actually get called for Update events :)

pkg/predicate/predicate_test.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//nolint:dupl
```
alvaroaleman, line 17: Why is this needed? Can we add an explanation comment for posterity and set it on a more granular level than on the whole package?

pkg/predicate/predicate.go
mars1024: :) Got your point, but your suggestion is keeping same as the other predicates or change all the predicates to a new way (use a constructor instead of stating a type), but I think the second way maybe a break change for some "legacy" users.
Or maybe we could just use the **constructing** way for the new `LabelChangedPredicate`?

pkg/predicate/predicate_test.go
mars1024: comments is added and level is set to single function from whole package

mars1024:
friendly ping @alvaroaleman , could you help to promote this PR? I think all comments have been addressed.

alvaroaleman (approved):
/lgtm
/approve

PR #1322, opened 2021-01-07
:sparkles: Deprecate Inject interface

varshaprasad96:
Deprecate `inject` interface

cc: @estroz

### DIFF
--- a/pkg/runtime/inject/inject.go
+++ b/pkg/runtime/inject/inject.go
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package inject is used by a Manager to inject types into Sources, EventHandlers, Predicates, and Reconciles.
+// Deprecated: Use manager.Options fields directly. This package will be removed in v0.10.
 package inject
 
 import (
@@ -24,7 +26,16 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
+
+// log is specifically to add a warning message for injectors.
+var log = logf.RuntimeLog.WithName("injectors-warning")
+
+// logWarningMsg logs a warning message if inject is used
+func logWarningMsg() {
+	log.Info("Injectors are deprecated, and will be removed in v0.10.x")
+}
 
 // Cache is used by the ControllerManager to inject Cache into Sources, EventHandlers, Predicates, and
 // Reconciles
@@ -36,6 +47,7 @@ type Cache interface {
 // false if i does not implement Cache.
 func CacheInto(c cache.Cache, i interface{}) (bool, error) {
 	if s, ok := i.(Cache); ok {
+		logWarningMsg()
 		return true, s.InjectCache(c)
 	}
 	return false, nil
@@ -50,6 +62,7 @@ type APIReader interface {
 // Returns false if i does not implement APIReader
 func APIReaderInto(reader client.Reader, i interface{}) (bool, error) {
 	if s, ok := i.(APIReader); ok {
+		logWarningMsg()
 		return true, s.InjectAPIReader(reader)
 	}
 	return false, nil
@@ -65,6 +78,7 @@ type Config interface {
 // false if i does not implement Config.
 func ConfigInto(config *rest.Config, i interface{}) (bool, error) {
 	if s, ok := i.(Config); ok {
+		logWarningMsg()
 		return true, s.InjectConfig(config)
 	}
 	return false, nil
@@ -80,6 +94,7 @@ type Client interface {
 // false if i does not implement Client.
 func ClientInto(client client.Client, i interface{}) (bool, error) {
 	if s, ok := i.(Client); ok {
+		logWarningMsg()
 		return true, s.InjectClient(client)
 	}
 	return false, nil
@@ -95,6 +110,7 @@ type Scheme interface {
 // false if i does not implement Scheme.
 func SchemeInto(scheme *runtime.Scheme, i interface{}) (bool, error) {
 	if is, ok := i.(Scheme); ok {
+		logWarningMsg()
 		return true, is.InjectScheme(scheme)
 	}
 	return false, nil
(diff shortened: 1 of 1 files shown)


coderanger:
Might want to explain a bit about why it is being deprecated?

varshaprasad96:
@vincepri @alvaroaleman : Have only added a comment as a deprecation warning above the interface. Is there logging required anywhere else?

vincepri:
@coderanger The code injection code is going to be deprecated and removed in a future code, it's hard to maintain this codebase and it's highly error prone for users. In addition, when making breaking changes users might not see any compile time errors and some code might start to fail silently, which isn't great.

@varshaprasad96 In terms of deprecation, we'd want to deprecate all of the injectors, not just the client one. We should also add a log line when any of the injection paths is used to warn users that this functionality is going to be removed in `v0.10.x`

pkg/runtime/inject/inject.go
```diff
@@ -72,6 +72,8 @@ func ConfigInto(config *rest.Config, i interface{}) (bool, error) {
 
 // Client is used by the ControllerManager to inject client into Sources, EventHandlers, Predicates, and
 // Reconciles
+// Deprecated: pass a client.Client to the manager directly.
```
vincepri, line 75:
```suggestion
// Deprecated: Injectors are deprecated, and will be removed in v0.10.
```

coderanger:
I would strongly recommend pulling the log into a helper function and making an env var that can be set on CI to turn these into fatal errors. Normal go test runners don't show output unless the test fails.

estroz:
@coderanger defining an env var in the manner you suggest seems like a rather heavy-handed approach to a simple deprecation, which I believe has been done in the past just fine.

coderanger:
@estroz Why not make it better though? Makes it much easier to check if you are clear on deprecation warnings if you can `DEPRECATIONS_AS_ERRORS=1 make test` and call it done :)

estroz:
@coderanger that's what linters are for, which will fail if deprecated functions/packages are used.

coderanger:
@estroz That works too, but is a lot more work to write a custom kubebuilder linter. Compared to make a `logDeprecation()` helper function somewhere :)

estroz:
`golangci-lint` will do this for you with [`staticcheck`](https://staticcheck.io/docs/checks#SA1019). Regardless the feature you suggest should be done in a separate PR. This PR should only contain log messages and a deprecation comment IMO.

coderanger:
I'm not sure that would work since the calls for these functions are not in user code, but I'll try and make a repro and see.

pkg/runtime/inject/inject.go
```diff
@@ -35,6 +40,7 @@ type Cache interface {
 // CacheInto will set informers on i and return the result if it implements Cache.  Returns
 // false if i does not implement Cache.
 func CacheInto(c cache.Cache, i interface{}) (bool, error) {
+	log.Info("Injectors are deprecated, and will be removed in v0.10.x")
```
coderanger, line 43: Shouldn't all of these only be inside the `if`, otherwise they will always log?

alvaroaleman:
@varshaprasad96 can you rename the pull to clarify that this is about deprecating all the Inject functionality?

pkg/runtime/inject/inject.go
```diff
@@ -24,8 +25,15 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
 
+// logWarningMsg logs a warning message if inject is used
+func logWarningMsg() {
+	log := logf.RuntimeLog.WithName("injectors-warning")
```
estroz, line 33:
```suggestion
var log = logf.RuntimeLog.WithName("injectors-warning")

// logWarningMsg logs a warning message if inject is used
func logWarningMsg() {
```

pkg/runtime/inject/inject.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package inject has been deprecated, and will be removed in v0.10.
```
estroz, line 17: Use the golang `// Deprecated:` [convention](https://github.com/golang/go/wiki/Deprecated) directly
```suggestion
// Deprecated: Use manager.Options fields directly. This package will be removed in v0.10.
```

estroz (approved):
/lgtm

vincepri (approved):
/approve

charith-elastic:
Is there an issue for the deprecation effort? The inject methods are still used internally by the controller-runtime (e.g. when [adding a watch](https://github.com/kubernetes-sigs/controller-runtime/blob/52d5ec72934e99f06dfba9e8092e6f72ca3c2446/pkg/internal/controller/controller.go#L113)). So even if calling code gets rid of any inject calls, the log messages won't go away.

hiddeco:
> So even if calling code gets rid of any inject calls, the log messages won't go away.

We are noticing this too for our set of controllers, and given they heavily depend on watches, we are serving [our users](https://github.com/fluxcd) deprecation notices we can not do anything about 😕

estroz:
@charith-elastic @hiddeco reverting part of this in #1382

DirectXMan12:
For next time we do this: there should be a corresponding issue or *something*, and as coderanger noted, there needs to be a more in-depth explanation in the PR & commit messages -- "deprecate XYZ" is insufficient.  People need to be able to look back on PRs and issues to figure out what is going on.

hiddeco:
> For next time we do this: there should be a corresponding issue or something, and as coderanger noted, there needs to be a more in-depth explanation in the PR & commit messages -- "deprecate XYZ" is insufficient. 

Unsolicited consumer advice / perspective:

Tracing this back was actually pretty easy for us, but we do keep a close eye on reading the changelog to understand what's changed. The fact that we were unable to do anything about it, other than commenting here that we ran into deprecation notices we could not do anything about was however not pleasant to our end-users.

So IMHO the issue is/was not with (the ability of) looking back, but looking _forward_ to the impact of changes, so that changes are made in stages (or all at once) but without it having an impact on consumers.

Lastly, thanks all involved for rolling out the patch release, we have updated most of our controllers today. :sunflower:

DirectXMan12:
@hiddeco yeah, totally agree on the impact of this change (log messages passed on to end users w/o any fix).  This was not our best moment, unfortunately, but we'll certainly try to do better in the future.

My comment re: explanation is mainly for folks looking back in the future as to *why* something was changed.  We certainly also need forward-looking impact too though :-).

varshaprasad96:
Sorry for the inconvenience caused here. I agree, this PR was premature, and doesn’t explain the reason for deprecating the functionality properly either. I am starting a discussion on this in the [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/1433), so that we can plan the next steps on removing the inject interface usage internally first and then deprecate it later for the users.

PR #1324, opened 2021-01-08
✨ initialize reconciler metrics when controller is started

joelanford:
This PR initializes the known reconciler metrics for a controller when the controller starts up (rather than waiting for a reconciliation to happen).

While there are ways with promql to work around missing metrics like this, it is a prometheus best practice to initialize all known metrics in advance. See: https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

Issue first reported here: https://github.com/operator-framework/operator-sdk/issues/4372


alvaroaleman (approved):
/lgtm

Issue #1048, opened 2020-07-15
Proposal: a way to inject errors into fake Client for testing 


jpatel-pivotal:
It would be extremely helpful if the fake Client in controller-runtime had a way to inject errors that get returned to help with testing a controller's reconcile logic. 

Consider the following psuedo-code below for example:

```
myObjectBeingReconciled, err := r.Get(ctx, req.NamespacedName)

	if instanceNotFound(err) {
		return noRequeue()
	} else if err != nil {
		return requeue(err)
	}

```

Currently there is no easy way to unit test the handled vs. unhandled error case shown above with the fake Client. I propose that the fake Client could have a field like `errorsToReturn map[errorKey]error` that could be used to hold errors to return for testing as shown below:

```
type errorKey struct {
	action         string
	resourceSchema schema.GroupVersionKind
	resourceKey    client.ObjectKey
}
type fakeClient struct {
	tracker     versionedTracker
	scheme      *runtime.Scheme
	errorsToReturn map[errorKey]error
}
```

Error injection could occur when the fakeClient is initialized with a constructor like `func NewFakeClientWithInjectedErrors(clientScheme *runtime.Scheme, errToReturn map[errorKey]error, initObjs ...runtime.Object) client.Client` 

OR

via an exported method like `func (c *fakeClient) InjectActionWithError(errorKey errorKey, injectedError error)`.

Using the `InjectActionWithError()` we could now inject a `instanceNotFound` error as well as a generic test error like `errors.New("unhandled error case")`to be returned for `myObjectBeingReconciled` so that we can get to 100% test coverage on the psuedo-code shown above.  

```
//InjectActionWithError injects errorKey and error into errorsToReturn map for testing error cases.
func (c *fakeClient) InjectActionWithError(errorKey errorKey, injectedError error) {
	c.errToReturn[errorKey] = injectedError
}
```

The fake Client has to check the `errorsToReturn` map and see if a particular action is meant to return an error. If so, the injected error can be returned. If no injected errors were meant to be returned for this particular action, then it performs the intended action as before. An example `get` implementation would look like:

```
func (c *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
	err = c.getInjectedError(errorKey)
	if err != nil {
                // Injected error matched so we return it
		return err
	}
        // This particular get action was not meant to return an error
	// Do the actual get as before
}

[...]

PR #1325, opened 2021-01-09
:seedling: Add testingclient package with clients for testing

dsharp-pivotal:
This PR is a proposal to add a set of testing clients to controller-runtime. This code is in a fairly early state, and I'm requesting feedback on the direction and acceptability. In particular, I know it needs tests backfilled, and documentation added. Also not sure about any file copyright headers that may need to be added (couldn't find guidelines about that).

This is somewhat an alternative proposal to, but mostly an expansion of, #1048.

- testingclient.Spy: Records calls to the API before delegating to the underlying client. It writes the records to a channel. This means that if a synchronous channel is used, it could also be used for synchronizing timing between test code and the code under test.
- testingclient.ErrorInjector: Intercepts calls to the delegate client to force an error to be returned. Errors can be injected for any combination of action ("get", "create", "update", etc), kind, and object key, with wildcards available.
- testingclient.Reactive: Brings the functionality of client-go's Reactors to the controller-runtime Client API.

I understand that `testingclient.Reactive` might not be desired for inclusion because it's high flexibility can turn it into a footgun. However, it is in active use by a few teams at VMware, with multiple copies existing. Most of its uses could be replaced with the other testing clients, which represent the two most common use cases.

One class of usage of `testingclient.Reactive` in our code that cannot currently covered by `testingclient.ErrorInjector` is filtering the contents of a "patch" to decide if an error should be returned. I can work on adding that, but it's not immediately clear to me what the API for it should be. (Perhaps a separate method, `InjectErrorForPatchContaining(kind client.Object, objectKey client.ObjectKey, patchContains string, injectedError error)`, or `InjectErrorForPatchMatching()` where patchContains could be a `regexp.RE`.)

For the sake of numbers, here's a categorization of all uses of PrependReactor in one operator's unit tests and how they could be replaced with these clients:

```
  28 //Reactive -> Error
  12 //Reactive -> Spy
  11 //Reactive -> Error, but with filtering the patch content
```

CC: @jpatel-pivotal @abg @vincepri

### DIFF
--- a/pkg/client/testingclient/error.go
+++ b/pkg/client/testingclient/error.go
@@ -11,14 +11,24 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type Verb string
+
+const (
+	GetVerb    Verb = "get"
+	CreateVerb Verb = "create"
+	DeleteVerb Verb = "delete"
+	UpdateVerb Verb = "update"
+	PatchVerb  Verb = "patch"
+	AnyVerb    Verb = "*"
+)
+
 type resourceActionKey struct {
-	action    string
+	verb      Verb
 	kind      schema.GroupVersionKind
 	objectKey client.ObjectKey
 }
 
 var (
-	AnyAction  = "*"
 	AnyKind    = &unstructured.Unstructured{}
 	AnyObject  = client.ObjectKey{}
 	anyKindGVK = schema.GroupVersionKind{Kind: "*"}
@@ -49,7 +59,7 @@ func (c ErrorInjector) RESTMapper() meta.RESTMapper {
 }
 
 func (c ErrorInjector) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
-	err := c.getStubbedError("get", obj, key)
+	err := c.getStubbedError(GetVerb, obj, key)
 	if err != nil {
 		return err
 	}
@@ -62,7 +72,7 @@ func (c ErrorInjector) List(ctx context.Context, list client.ObjectList, opts ..
 }
 
 func (c ErrorInjector) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
-	err := c.getStubbedError("create", obj, client.ObjectKeyFromObject(obj))
+	err := c.getStubbedError(CreateVerb, obj, client.ObjectKeyFromObject(obj))
 	if err != nil {
 		return err
 	}
@@ -71,7 +81,7 @@ func (c ErrorInjector) Create(ctx context.Context, obj client.Object, opts ...cl
 }
 
 func (c ErrorInjector) Delete(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error {
-	err := c.getStubbedError("delete", obj, client.ObjectKeyFromObject(obj))
+	err := c.getStubbedError(DeleteVerb, obj, client.ObjectKeyFromObject(obj))
 	if err != nil {
 		return err
 	}
@@ -80,7 +90,7 @@ func (c ErrorInjector) Delete(ctx context.Context, obj client.Object, opts ...cl
 }
 
 func (c ErrorInjector) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
-	err := c.getStubbedError("update", obj, client.ObjectKeyFromObject(obj))
+	err := c.getStubbedError(UpdateVerb, obj, client.ObjectKeyFromObject(obj))
 	if err != nil {
 		return err
 	}
@@ -89,7 +99,7 @@ func (c ErrorInjector) Update(ctx context.Context, obj client.Object, opts ...cl
 }
 
 func (c ErrorInjector) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
-	err := c.getStubbedError("patch", obj, client.ObjectKeyFromObject(obj))
+	err := c.getStubbedError(PatchVerb, obj, client.ObjectKeyFromObject(obj))
 	if err != nil {
 		return err
 	}
@@ -105,18 +115,18 @@ func (c ErrorInjector) Status() client.StatusWriter {
 	return c
 }
 
-func (c ErrorInjector) getStubbedError(action string, kind client.Object, objectKey client.ObjectKey) error {
+func (c ErrorInjector) getStubbedError(action Verb, kind client.Object, objectKey client.ObjectKey) error {
 	gvk := mustGVKForObject(kind, c.Scheme())
 
 	for _, k := range []resourceActionKey{
-		{action, gvk, objectKey},           // (1) 0 wildcards
-		{action, gvk, AnyObject},           // (2) 1 wildcard
-		{AnyAction, gvk, objectKey},        // (3) 1 wildcard
-		{action, anyKindGVK, objectKey},    // (4) 1 wildcard
-		{AnyAction, gvk, AnyObject},        // (5) 2 wildcards
-		{action, anyKindGVK, AnyObject},    // (6) 2 wildcards
-		{AnyAction, anyKindGVK, objectKey}, // (7) 2 wildcards
-		{AnyAction, anyKindGVK, AnyObject}, // (8) 3 wildcards
+		{action, gvk, objectKey},         // (1) 0 wildcards
+		{action, gvk, AnyObject},         // (2) 1 wildcard
+		{AnyVerb, gvk, objectKey},        // (3) 1 wildcard
+		{action, anyKindGVK, objectKey},  // (4) 1 wildcard
+		{AnyVerb, gvk, AnyObject},        // (5) 2 wildcards
+		{action, anyKindGVK, AnyObject},  // (6) 2 wildcards
+		{AnyVerb, anyKindGVK, objectKey}, // (7) 2 wildcards
+		{AnyVerb, anyKindGVK, AnyObject}, // (8) 3 wildcards
 	} {
 		if err, ok := c.errorsToReturn[k]; ok {
 			return err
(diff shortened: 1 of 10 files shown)


pkg/client/testingclient/reactive.go
```diff
@@ -0,0 +1,243 @@
+package testingclient
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
alvaroaleman, line 9: We most certainly do not want to add ginkgo or gomega in any non-test files, downstream projects should be able to decide what testing framework they do or do not want to use

pkg/client/testingclient/error.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	return nil
+}
+
+// InjectError will cause ErrorInjector to return an error for the given (action, kind, objectKey) tuple.
+// To inject an error for all calls with `(action,kind)`, pass objectKey = AnyObject.
+// To inject an error for all calls with `action`, pass kind = "*" and objectKey = AnyObject.
+// To inject an error for any `action`, pass action = "*",  kind = "*" and objectKey = AnyObject.
+func (c *ErrorInjector) InjectError(action string, kind client.Object, objectKey client.ObjectKey, injectedError error) {
```
alvaroaleman, line 129: IMHO this client is an overly specific thing that should instead just be implemented by downstream projects via the reactive client

pkg/client/testingclient/spy.go
```diff
@@ -0,0 +1,143 @@
...
+	gvk, err := listGVK(list, s.Scheme())
+	if err != nil {
+		return err
+	}
+
+	listErr := s.Delegate.List(ctx, list, opts...)
+	s.Calls <- SpyCall{
+		GVK:  gvk,
+		Verb: "list",
+		List: list,
```
alvaroaleman, line 63: You need to DeepCopy everything whenever you store the call, otherwise the object in the saved call might get manipulated later on.

IMHO the functionality here should from an external API POV be part of the reactive client or not exist (internally, it can be a distinct and unexported struct)

alvaroaleman (commented):
IMHO:
* We should either agree on merging something with the flexibility of the reactive client or nothing at all, anything in between will just create questions and feature requests as it can cover some, but not all use cases. The current fake client is at least consistent in that it doesn't do anything other than what a real client would do
* In the end not have more than one client, as more clients will just cause confusion

vincepri:
+1 on the effort, thank you for submitting this!

I'd also add that we need 1-2 folks at minimum that can be maintainers for this new client — We can use nested OWNERS files to have reviewers + approvers groups specific to the client test package. These folks would be in charge of maintaining the client and supporting it going forward. 

The reason I'm bringing this up is that not long ago the fake client almost got deprecated because it's not really kept up to date with a real api-server behavior, and there is a lot of limitations (even for features in controller-runtime itself).

pkg/client/testingclient/reactive.go
dsharp-pivotal: Good point. The uses here are to make some precondition assertions. Do you think it's reasonable to raise a `panic()`, or is it better to bubble up an `error`? My hesitation for returning an error is that it could be hard to diagnose the issue if the error doesn't make it back to the test. My hesitation with panicking is that it will stop all further tests, and the output isn't very friendly. Nevertheless, I'm leaning towards panicking.

pkg/client/testingclient/spy.go
dsharp-pivotal: Thanks, added the `DeepCopyObject()`s. Also had to type-assert the copies back to `client.Object`/`client.ObjectList`. I think that should be safe.

dsharp-pivotal:
@alvaroaleman 

> IMHO:
> 
> * We should either agree on merging something with the flexibility of the reactive client or nothing at all, anything in between will just create questions and feature requests as it can cover some, but not all use cases. The current fake client is at least consistent in that it doesn't do anything other than what a real client would do

TBH, I would be perfectly happy with only Reactive client (in addition to Fake client). I created the others in response to @DirectXMan12 opinion that the reactors of the client-go fake client are a confusing footgun.  IIRC, Solly stated that they would prefer to see more specific, single-purpose clients. I guess we need to get on the same page as a community...

An "all of the above" approach might be appropriate: ErrorInjector and Spy for simple, common uses and a gentler onramp, and Reactive for when they don't cut it.

> * In the end not have more than one client, as more clients will just cause confusion

My hope is that a well-designed toolkit of testing clients that each do one thing well would be less confusing than a single swiss army knife that you have to learn to use correctly in many different ways. So, while I agree that the Reactive client is sufficient for all purposes, I don't agree that having a single tool is what is necessary to eliminate confusion.

alvaroaleman:
What about an interface where we allow to override individual methods for the reactive client, something along the lines of:
```
type ReactiveClient struct {
  Upstream client.Client
  CreateFunc func(ctx.Context, obj client.Obj, opts []client.CreateOption, upstream client.Client) error
  ...
}

func (r *reactiveClient) Create(ctx context.Context, obj client.Obj, opts ...client.CreateOption){
   if r.CreateFunc != nil {
      return r.CreateFunc(ctx, obj, opts, r.Upstream)
  }
  return r.Upstream(ctx, obj, opts...)
} 
```

Then for example injecting an error would be something like:
```
client: := &ReactiveClient{
  Upstream: fakeclient.NewFakeClient(),
  CreateFunc: func(ctx.Context, obj client.Obj, opts []client.CreateOption, upstream client.Client) error{
    if shouldError(obj){
       return errors.New("")
    }
    return upstream.Create(ctx, obj, opts...
  }
```

dsharp-pivotal:
That's a drastic departure from Reactors. I would no longer call that a Reactive client. I would call that something like "Hookable". In particular, it's missing built in support for filtering on object type and identity.

alvaroaleman:
> That's a drastic departure from Reactors. I would no longer call that a Reactive client. I would call that something like "Hookable". In particular, it's missing built in support for filtering on object type and identity.

No, you get the object, you can do any kind of checks you want to do on it and based on that decide what your client is supposed to do, including nothing at all. The advantage is that we don't need a gazillion config option for each reaction and will still be able to cover anything. My example above has this btw, see the `if shouldError(obj)` passage.

dsharp-pivotal:
Yes, I see that. If you want that client, that's fine, but it's not a "Reactive" client. "Reactive" client means it uses "Reactors" like in the client-go fake client. I can add "Hookable" separately if you think it would be valuable. I'd like to keep that separate from Reactive.

alvaroaleman:
> Yes, I see that. If you want that client, that's fine, but it's not a "Reactive" client. "Reactive" client means it uses "Reactors" like in the client-go fake client

My understanding from our last meeting was that that's exactly what we want to avoid, due to the complexity, underdocumentation and footgun potential of client-gos reactor

dsharp-pivotal:
> > Yes, I see that. If you want that client, that's fine, but it's not a "Reactive" client. "Reactive" client means it uses "Reactors" like in the client-go fake client
> 
> My understanding from our last meeting was that that's exactly what we want to avoid, due to the complexity, underdocumentation and footgun potential of client-gos reactors

My understanding was that we agreed that Reactors are useful, but we should take some steps to avoid some of the footguns.
- @DirectXMan12  said we could "make the API more like InjectError". I took this to mean replace the "resource" string with an Object reference, for example.
- Also mentioned was "what happens when you fall off the reaction chain". I think this just needs documentation. Since NewReactive() always adds a reactor that calls the delegate client, and "handles" the request, the only way to fall off the reaction chain is to manually empty `Fake.ReactionChain`. `AddReactor` isn't useful, since the default reactor will always handle the action before the appended reactor.

So, here's a plan:
- Instead of embedding testing.Fake, put it behind a (public?) field. I want to keep it public so it can be used as an escape hatch. We can put a note to the effect that it shouldn't be used unless necessary. You could probably talk me in to making it private.
- Provide a method `func (r *Reactive) PrependReactor(verb string, kind client.Object, reaction ReactionFunc)` that will translate into a call to `r.Fake.PrependReactor(verb, resourceForKind, reaction`. Do not provide `AddReactor`.
- Provide `Actions()` and `ClearActions()` to expose the same from `testing.Fake`. (I haven't used these personally, but seems like they could be useful, and harmless.)

DirectXMan12:
Sorry for being a bit slow to respond to this.  I think in addition to the above, at the very least we'd need clear documentation on expected behavior for the reaction chain, or maybe a concrete return type, as well as clear documentation for any implicit behavior (label selectors will get automatically applied to returned results, for instance).

I'd also like to see the proposed signature of `Reaction` and `PrependReactor` -- try to make the types a bit less unwieldy to use.  If we can wrap the underlying reactors in a safe interface, and we're confident that the interface is actually safe/usable, I think I'd be okay with that as long as we have a set of owners (e.g. @dsharp-pivotal).

DirectXMan12:
in particular, I'd like to avoid any cases of "this is an arbitrary string, who knows what values it's supposed to take".  InjectError's style is a decent start, but still has "action" as a string.  Can we shift that around?  Maybe

```go
type Action string
const (
  GetAction Action = "get"
  ...
)
type Reaction interface {
  Actions() []Action
  Handle(ctx context.Context, object client.Object) (stop bool, err error)
  HandleGet(ctx context.Context, key client.ObjectKey, object client.Object) (stop bool, err error)
}

// helpers for comment types
type Get func(ctx context.Context, key client.ObjectKey, object client.Object, client.GetOptions)
func (f Get) Actions() []Action { retcontinueurn []Action{GetAction} }
func (f Get) Handle(...) (bool, error) { return false, nil }
func (f Get) HandleGet(ctx context.Context, key client.ObjectKey, obj client.Object) (bool, error) {
  opts := OptionsFrom(ctx)
  return f(ctx, key, obj, opts)
}

// comment case is then
cl := fake.Client(...)
cl.On(fake.Get(func(...) {
   // do some stuff on get
})

// to handle multiple things, either wrap the same function twice, or implement the interface on a struct
```

(note: this is just off the top of my head -- feel free to propose something else entirely or change it up as necessary)

jpatel-pivotal:
Definitely +1 on the effort and thank you @dsharp-pivotal for bringing it to this point. I intend to help with the plan so let's co-ordinate.

dsharp-pivotal:
Hi, I think this is ready for another round of review. Sorry it's been so long. I only have time to work on this every couple weeks. I still have a couple of TODOs in the code, but I think we might be able to live without them.

@DirectXMan12 I think I don't understand your proposed interface. It looks like something fancy, but maybe it's too fancy? Or maybe it just needs more fleshing out. All I've done is replace the verb strings with an enum type, and replaced the resource name strings with pointers to empty objects (i.e. `&corev1.Pod{}`).

jpatel-pivotal:
@alvaroaleman I wanted to follow up and see if there was any additional feedback. We would like to make progress.

Thanks and let us know.

fabriziopandini:
ErrorInjector and Spy client wrapper could be really useful, few comments from and user PoV:
- it will be simpler if they are part of `sigs.k8s.io/controller-runtime/pkg/client/fake` instead of a different package
- it will be great if we can use those wrappers directly from the ClientBuilder, something like `fake.NewClientBuilder().WithScheme(scheme.Scheme).WithInjectedError(...).Build()`; NOTE: 
   - this should not prevent us to expose those wrappers as a standalone object, so they can be eventually used with a thestenv client as well
- I'm personally not really concerned about "magic strings"/"magic values" to be passed to those helpers to express verb or any object, however I assume that using options we can get something more expressive.

I have less strong opinions about reactor wrapper, because I can see possible use cases for it, but at the same time the client-go reactor interface seems to me not properly in line the controller runtime clients. I need to think a little bit more about this

PR #1326, opened 2021-01-10
🐛 ignore namespace key when getting cluster-scoped resources

### LINKED ISSUE #1323: Cached client doesn't support Get for cluster scoped resources if namespace is given
cached client `Get` returns not found when looking for cluster-scoped object and passing in a namespace.
client.Get should ignore the namespace for cluster scoped objects and return the object. Logging a
warning is probably a good idea that we are ignoring the namespace.

The APIReader ignores the namespace for cluster-scoped gets. The cache should behave similarly.

joelanford:
This PR makes the controller-runtime cache consistent with the behavior of the client when working with cluster-scoped resources. 

Closes #1323 

Current behavior:
When fetching a cluster-scoped resource using a direct client (e.g. `mgr.GetAPIReader`), the namespace in the object key is ignored. When fetching a cluster-scoped resource via the cache, the namespace key MUST be set to an empty string. Otherwise the cache will return a not found error.

Expected behavior:
When fetching a cluster-scoped resource via the cache, the namespace key is ignored, just like with a direct client. The APIReader and Cache behavior should be consistent.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -35,6 +35,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const testNodeOne = "test-node-1"
 const testNamespaceOne = "test-namespace-1"
 const testNamespaceTwo = "test-namespace-2"
 const testNamespaceThree = "test-namespace-3"
@@ -97,6 +98,8 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 
 			By("creating three pods")
 			cl, err := client.New(cfg, client.Options{})
+			Expect(err).NotTo(HaveOccurred())
+			err = ensureNode(testNodeOne, cl)
 			Expect(err).NotTo(HaveOccurred())
 			err = ensureNamespace(testNamespaceOne, cl)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


pkg/cache/cache_test.go
```diff
@@ -413,17 +416,33 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					Expect(out.Items).Should(HaveLen(1))
 					Expect(out.Items[0].GetNamespace()).To(Equal(testNamespaceOne))
 
-					By("listing all namespaces - should still be able to get a cluster-scoped resource")
-					namespaceList := &unstructured.UnstructuredList{}
-					namespaceList.SetGroupVersionKind(schema.GroupVersionKind{
+					By("listing all nodes - should still be able to list a cluster-scoped resource")
```
alvaroaleman, line 419: Why did you change this from Namespace to Node?

pkg/cache/cache_test.go
```diff
@@ -413,17 +416,33 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+					nodeList := &kmetav1.PartialObjectMetadataList{}
```
alvaroaleman, line 420: This needs to remain an unstructured.UnstructuredList, otherwise the two testcases are identical

pkg/cache/cache_test.go
joelanford: Mainly to avoid confusion of trying to explain "setting a namespace key on a namespace"

pkg/cache/cache_test.go
joelanford: Good catch! That was an accidental copy/paste error.

alvaroaleman (approved):
/lgtm

PR #1331, opened 2021-01-13
🐛: internal: fix a possible temporary directory leak

zhijianli88:
a temporary directory leak was found when running the tests

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/pkg/internal/testing/integration/internal/process.go
+++ b/pkg/internal/testing/integration/internal/process.go
@@ -87,6 +87,13 @@ func DoDefaulting(
 		defaults.URL = *listenURL
 	}
 
+	if path == "" {
+		if name == "" {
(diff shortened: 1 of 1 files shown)


pkg/internal/testing/integration/internal/process.go
```diff
@@ -98,6 +98,9 @@ func DoDefaulting(
 
 	if path == "" {
 		if name == "" {
+			if defaults.DirNeedsCleaning {
```
alvaroaleman, line 101: Can we just move the `TempdDir` ceation past the `if path == ""` condition to not create a temp dir before knowing our inputs are corret?

zhijianli88: Fixed

alvaroaleman (approved):
/lgtm
/approve 

Thanks!

Issue #615, opened 2019-09-30
DelegatingReader cant be used to read unstructured.Unstructured from cache


alvaroaleman:
The delegatingReader always redirects read requests for `unstructured.Unstructured` to the API reader: https://github.com/kubernetes-sigs/controller-runtime/blob/801e12a501606245062a6ff77ac91eef8936d9fb/pkg/client/split.go#L47

However it seems the `InformersMap` is perfectly well capable of establishing watches for `unstructured.Unstructured`: https://github.com/kubernetes-sigs/controller-runtime/blob/801e12a501606245062a6ff77ac91eef8936d9fb/pkg/cache/internal/deleg_map.go#L77

When doing a quick test by creating a client that just uses a `Cache` as `Reader` and requesting `unstructured.Unstructured` through it, everything seemed to work fine and I was able to get objects this way:

```
I0930 12:05:21.031942    2216 reflector.go:131] Starting reflector *unstructured.Unstructured (10h0m0s) from sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196
I0930 12:05:21.031972    2216 reflector.go:169] Listing and watching *unstructured.Unstructured from sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196
```

/kind bug
CC @thetechnick

shawn-hurley:
You have to create the delegating reader to use this behavior yourself on startup. Here is an example of how to do it:

```
	mgr, err := manager.New(cfg, manager.Options{
		NewClient: func(cache cache.Cache, config *rest.Config, options client.Options) (client.Client, error) {
			c, err := client.New(config, options)
			if err != nil {
				return nil, err
			}
			return &client.DelegatingClient{
				Reader:       cache,
				Writer:       c,
				StatusClient: c,
			}, nil
		},
	})
```


We chose to make this the default behavior to make sure that folks were not caching the entire cluster on accident with unstructured.

alvaroaleman:
I can work around the problem, but this issue is about avoiding the workaround :)

> We chose to make this the default behavior to make sure that folks were not caching the entire cluster on accident with unstructured.

Generally I find this  a very valid concern, but why do you think this is more likely and/or more problematic when ppl use `unstructured.Unstructured`?

I would have argued that the types that create a particular high amount of traffic like Pods or Endpoints are mostly in the corev1 api and ppl tend to not have very high number of CRs.

alvaroaleman:
[...]

PR #1332, opened 2021-01-13
🐛 pkg/client: optionally allow caching of unstructured objects

joelanford:
Based on https://github.com/kubernetes-sigs/controller-runtime/issues/615, it seems like there is intent to allow the DelegatingClient to cache unstructured objects. This is behavior that was possible to use before v0.7.0 by using the `NewClientFunc` manager option and creating a custom DelegatingClient.

IMO, this could be considered a regression, but since a large chunk of this code was restructured in v0.7.0, I'm not sure if dropping support for unstructured caching was intentional.

This PR adds a new boolean field to the NewDelegatingClientInput struct that allows users to opt-in to caching unstructured objects.

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -22,8 +22,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"k8s.io/apimachinery/pkg/types"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


joelanford:
/hold

For discussion.

joelanford:
Looks like https://github.com/kubernetes-sigs/controller-runtime/pull/1058 removed the exported `DelegatingClient` struct, which I _think_ is the root cause of the regression, unless I'm missing something about how we can support caching unstructured objects without forcing users to reimplement their own delegating client.

estroz (approved):
Agreed that this is a regression.

/lgtm

vincepri:
I'm +1 having the option to cache unstructured objects as well. 

IIRC, the earlier version of the DelegatingClient was meant to be used to always go to a live client for unstructured objects https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.5/pkg/client/split.go#L45-L61, we kept that behavior. 

That said, as you mentioned the `NewClientFunc` option, which was replaced with `NewClientBuider` and an interface where folks can fill in with their own struct https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.7/pkg/manager/client_builder.go#L25-L33, so I'm +1 supporting this as an option

/approve

coderanger:
Might also be good to set this up as a manager option too?

joelanford:
/hold cancel

I'll add a manager option in a follow-up.

PR #1334, opened 2021-01-14
✨ Adding WithContextFunc for webhook server to pass info to the handler

### LINKED ISSUE #1333: :sparkles: Ability to modify Admissions Request context
I would like to be able to get access to all or some of the http.Request.Context from within an Admission webhook handler. This could be for getting the path, parameters, adding tracing information, etc. This should be something you can opt into since it could bloat the context if we just added the entire request object.

### Proposal

Modify admission.Webhook struct to include optional `WithContextFunc` which allows the controller builder to implement their own context mutation:

```
type Webhook struct {
	// Handler actually processes an admission request returning whether it was allowed or denied,
	// and potentially patches to apply to the handler.
	Handler Handler

	// WithContextFunc will allow you to take the http.Request.Context() and
	// add any additional information such as passing the request path or
	// headers thus allowing you to read them from within the handler
	WithContextFunc func(context.Context, *http.Request) context.Context

	// decoder is constructed on receiving a scheme and passed down to then handler
	decoder *Decoder

	log logr.Logger
}
```

With that defined we can then change the `ServeHTTP` func to call this func before passing the context to the `Handler`.

```
func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()
	if wh.WithContextFunc != nil {
		ctx = wh.WithContextFunc(ctx, r)
	}
        // ...
        reviewResponse = wh.Handle(ctx, req)
}
```

### Usage

Typically in `main.go` this would setup the webhook server and add the path to the context, we can then return this value and take any action we need.

```
	mgr.GetWebhookServer().Register("/webhook", &webhook.Admission{
		Handler: &wh.Service{
			Client: mgr.GetClient(),
		},
		WithContextFunc: ctx context.Context, r *http.Request) context.Context {
	        	return context.WithValue(ctx,"path", r.URL.Path)
		},
	})
```

Webhook implementation:
```
func (a *Service) Handle(ctx context.Context, req admission.Request) admission.Response {
	path := ctx.Value("path").(string)
	fmt.Printf("path: %s", path)
	// ...
}
```

christopherhein:
This closes #1333 and extends #549.

This non-breaking change introduces a new field on the webhook server registration, this allows you to add additional information to the request context from the `ServeHTTP` func. This is useful if you need to get information from the http.Request like parsing the request path for variables.

### Controller Builder Implementation:

This example will write the request path back to 

```
type pathKeyType int
const pathKey pathKeyType = iota
func setPath(ctx context.Context, r *http.Request) context.Context {
	return context.WithValue(ctx, pathKey, r.URL.Path)
}

// Get will return the new request details from the context
func getPath(ctx context.Context) string {
	return ctx.Value(pathKey).(string)
}

mgr.GetWebhookServer().Register("/webhook", &webhook.Admission{
	Handler: &fakeHandler{
		fn: func(ctx context.Context, req Request) Response {
			return Allowed(getPath(ctx))
		},
	},
	WithContextFunc: setPath,
})
```

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -45,6 +45,10 @@ var _ http.Handler = &Webhook{}
 func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var body []byte
 	var err error
+	ctx := r.Context()
+	if wh.WithContextFunc != nil {
+		ctx = wh.WithContextFunc(ctx, r)
+	}
 
 	var reviewResponse Response
 	if r.Body != nil {
@@ -93,7 +97,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	wh.log.V(1).Info("received request", "UID", req.UID, "kind", req.Kind, "resource", req.Resource)
 
 	// TODO: add panic-recovery for Handle
-	reviewResponse = wh.Handle(r.Context(), req)
+	reviewResponse = wh.Handle(ctx, req)
 	wh.writeResponseTyped(w, reviewResponse, actualAdmRevGVK)
 }
 
--- a/pkg/webhook/admission/http_test.go
+++ b/pkg/webhook/admission/http_test.go
@@ -162,6 +162,34 @@ var _ = Describe("Admission Webhooks", func() {
 			webhook.ServeHTTP(respRecorder, req.WithContext(ctx))
 			Expect(respRecorder.Body.String()).To(Equal(expected))
 		})
+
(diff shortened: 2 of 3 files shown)


christopherhein:
@vincepri re:apidiff can we ignore this? Seems like it's expected:

```
[1;33minvoking: 'hack/tools/bin/go-apidiff 17b28b4c90d8736fa8ada8942235635191f85021 --print-compatible'[0m

sigs.k8s.io/controller-runtime/pkg/webhook/admission
  Incompatible changes:
  - Webhook: old is comparable, new is not
  Compatible changes:
  - Webhook.WithContextFunc: added
```

PR #1338, opened 2021-01-14
⚠ Add manager option to enable caching of unstructured objects

joelanford:
This PR adds a new manager/cluster option, called `ClientCacheUnstructured` that configures the ClientBuilder to build a client that caches unstructured objects.

It is a breaking change because it requires a new method on the `cluster.ClientBuilder` interface.

This is a follow-up based on https://github.com/kubernetes-sigs/controller-runtime/pull/1332#issuecomment-760313271

### DIFF
--- a/pkg/cluster/client_builder.go
+++ b/pkg/cluster/client_builder.go
@@ -29,6 +29,10 @@ type ClientBuilder interface {
 	// for this client. This function can be called multiple times, it should append to an internal slice.
 	WithUncached(objs ...client.Object) ClientBuilder
 
+	// CacheUnstructured tells the client whether or not to cache unstructured objects and lists. By default,
+	// unstructured objects and list are not cached.
+	CacheUnstructured(v bool) ClientBuilder
+
 	// Build returns a new client.
 	Build(cache cache.Cache, config *rest.Config, options client.Options) (client.Client, error)
 }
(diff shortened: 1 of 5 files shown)


joelanford:
/hold

For more discussion since this is a breaking change.

shawn-hurley:
Could another approach be to make a new client.Option for this and that would not be a breaking change I think?

I am not saying we should just wondering if we looked at that approach.

joelanford:
@shawn-hurley I'm not following your suggestion. I was also trying to think of a non-breaking change approach, but nothing is coming to me that would make this option available to arbitrary `ClientBuilder`s.

IMO, it's a little bit odd that we would have cache-related methods on the `ClientBuilder` interface because it implies that all clients need to handle caching. We _do_ pass a cache as a `Build()` parameter, but that could theoretically be ignored. Cache-specific methods on the interface for non-caching implementations would need to be no-ops.

I suppose we could make a new interface `CachingClientBuilder` like below and then do type assertions to see if we can call some extra methods:
```go
type CachingClientBuilder interface {
	ClientBuilder
	WithUncached(objs ...client.Object) ClientBuilder
	CacheUnstructured(v bool) ClientBuilder
```

But even with that approach, @estroz made the astute observation that further configurations that affect the `delegatingClient` would require further breaking changes on this interface if we wanted to expose that configuration to users.

shawn-hurley:
Admittedly, have not been involved enough over the last year to know if this makes sense so if it doesn't I am sorry :).

I was wondering if it could be a option here:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L37

such that at `ClientBuilder.Build` it is passed in as an option for the client. 


Maybe this doesn't make sense because the `ClientBuilder` already handles caching concerns and the `Options` don't so maybe this is a bad idea :)

karstland:
I'm quite confused about  manager option(filed ClientBuilder -> filed NewClient) in release v0.8.3 and v0.9.0.alpha1. Why the6. field fallback to release v0.6.5. Could someone explain it?  Which version should I use when I write my own controller...

joelanford:
@hzbcloud #1332 highlights the regression caused by the change to ClientBuilder and #1409 makes the change back to NewClientFunc. These PRs should explain why we needed to revert the ClientBuilder option.

Closing this since the problem this PR solves is no longer a problem with NewClientFunc.

Issue #1337, opened 2021-01-14
0.8.0 has a data race


cflewis:
0.8.0 has introduced a data race:

`go test -v -race ./...`

```
==================
WARNING: DATA RACE
Write at 0x00c00043ec60 by goroutine 8:
  sigs.k8s.io/controller-runtime/pkg/log.(*loggerPromise).Fulfill()
      /app/vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:75 +0xef
  sigs.k8s.io/controller-runtime/pkg/log.(*DelegatingLogger).Fulfill()
      /app/vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:124 +0x12e
  sigs.k8s.io/controller-runtime/pkg/log.init.0.func1()
      /app/vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:67 +0xbb

Previous read at 0x00c00043ec60 by main goroutine:
  [failed to restore the stack]

Goroutine 8 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/log.init.0()
      /app/vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:62 +0x44
==================
```

This error is failing my CI, so I can't upgrade to 0.8 at this time.

cflewis:
Thanks for the great work! :rocket:

PR #1339, opened 2021-01-15
:bug: Fix a race in the delegating logger

alvaroaleman:
For as long as the delegating loggers Fullfill wasn't called, it holds a
loggerPromise and that loggerPromise holds the same delegating logger.

Calls to With{Name,Field} and FullFill of such a loggerPromise and
delegatingLogger pair have to be serialized, because the former accesses
the actual logger and promise, while the latter sets and unsets them.

Furthermore, a call to With{Name,Field} of such a pair returns a new
pair that is attached to the parent pair by making the original promise
keep a reference to the new promise. Because of that, the aforementioned
serialization has to happen recursively for each pair.

https://github.com/kubernetes-sigs/controller-runtime/pull/1309 made this race more visible, because the defaulting of the logger controller-runtime does in the background might race with users code getting a logger (Potentially indirectly by constructing a controller).

Surprisingly, I only managed to reproduce this on tests that import controller-runtime when they take 30 seconds or more. I didn't expect the race detector to take `Sleep` into account, but apparently it does.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1337

/assign @vincepri @joelanford

### DIFF
--- a/pkg/log/deleg.go
+++ b/pkg/log/deleg.go
@@ -72,8 +72,10 @@ func (p *loggerPromise) Fulfill(parentLogger logr.Logger) {
 		logger = logger.WithValues(p.tags...)
 	}
 
+	p.logger.lock.Lock()
 	p.logger.Logger = logger
 	p.logger.promise = nil
+	p.logger.lock.Unlock()
 
 	for _, childPromise := range p.childPromises {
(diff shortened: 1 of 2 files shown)


joelanford:
Wow great find!

/lgtm

Issue #1294, opened 2020-12-10
misspelling 'DestWritter'


tatsuhiro-t:
There is a misspelling 'DestWritter'.  It should be 'DestWriter'.

https://github.com/kubernetes-sigs/controller-runtime/blob/3410e75b087895ff6cd79b6c28dd31c2145c813d/pkg/log/zap/zap.go#L148

PR #1340, opened 2021-01-15
🌱 fix DestWritter var name typo

nader-ziada:
- change name of var from `DestWritter` to `DestWriter`

Fixes: #1294

### DIFF
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -56,10 +56,10 @@ func UseDevMode(enabled bool) Opts {
 }
 
 // WriteTo configures the logger to write to the given io.Writer, instead of standard error.
-// See Options.DestWritter
+// See Options.DestWriter
 func WriteTo(out io.Writer) Opts {
 	return func(o *Options) {
-		o.DestWritter = out
+		o.DestWriter = out
 	}
 }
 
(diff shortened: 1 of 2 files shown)


coderanger:
This lost another `t`, is that intentional?

nader-ziada:
> This lost another `t`, is that intentional?

This was the issue, to replace `Writter` with `Writer` to correct the spelling of it

coderanger:
Yes but you changed `Dest` to `Des` too.

nader-ziada:
> Yes but you changed `Dest` to `Des` too.

Oh sorry. That's by mistake, will fix it.

coderanger:
I defer to others if this is worth fixing with a compat break. The field probably isn't used a ton or we would have noticed this sooner? If we want compat, a more complex fix will be needed.

/lgtm

nader-ziada:
> I defer to others if this is worth fixing with a compat break. The field probably isn't used a ton or we would have noticed this sooner? If we want compat, a more complex fix will be needed.
> 
> /lgtm

its fine if we don't need it, I can close the PR, was just trying to find an easy first PR to do to get familiar with the repo

joelanford (commented):
How hard would it be to add a new DestWriter field, deprecate the misspelled one, and do whatever is necessary to ensure they’re interchangeable (preferring DestWriter if they’re both set), and then remove DestWritter in a later release?

nader-ziada:
> How hard would it be to add a new DestWriter field, deprecate the misspelled one, and do whatever is necessary to ensure they’re interchangeable (preferring DestWriter if they’re both set), and then remove DestWritter in a later release?

should not be too hard, will work on that and update PR, thanks

vincepri (approved):
/lgtm
/assign @alvaroaleman

PR #1345, opened 2021-01-18
:bug: Prevent source.Channel from shutting down immediately

### LINKED ISSUE #1343: source.Channel gets shutdown almost immediately and never processes any events
After upgrading to controller-runtime 0.8.0, I discovered that our external event source implemented using `source.Channel` is not generating any events. Looking at the code, one odd thing that stood out to me was the useless lock here: https://github.com/kubernetes-sigs/controller-runtime/blob/818a2ce212912c980b53b9714b8ebcbf5c36a54f/pkg/source/source.go#L241-L244

It used to protect the [append to the `dest`  array](https://github.com/kubernetes-sigs/controller-runtime/blob/7fc9110156ff699124c108ffa48a84142f677135/pkg/source/source.go#L238-L241), but now the append has moved further up as a fix to #942. After fixing the lock in the code, the event source started working again so I think there must be a race somewhere. The odd thing is that this unprotected append was present in 0.7.0 as well and that didn't affect us. So I am not sure whether my hypothesis is entirely correct.

Either way, I'll be happy to submit a patch to fix the lock unless there are any objections.

charith-elastic:
Because of an incorrectly used context, `source.Channel` gets shutdown almost immediately before it gets a chance to process any events. More details can be found in #1343.

Also fixes a minor bug in `source.Channel` where the destination append operation wasn't protected by a lock as it used to.

Fixes #1343

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -163,9 +163,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		for _, watch := range c.startWatches {
 			c.Log.Info("Starting EventSource", "source", watch.src)
 
-			watchStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
-			defer cancel()
-			if err := watch.src.Start(watchStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
+			if err := watch.src.Start(ctx, watch.handler, c.Queue, watch.predicates...); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/internal/controller/controller.go
```diff
@@ -163,9 +163,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		for _, watch := range c.startWatches {
 			c.Log.Info("Starting EventSource", "source", watch.src)
 
-			watchStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
-			defer cancel()
-			if err := watch.src.Start(watchStartCtx, watch.handler, c.Queue, watch.predicates...); err != nil {
+			if err := watch.src.Start(ctx, watch.handler, c.Queue, watch.predicates...); err != nil {
```
alvaroaleman, line 166: Unfortunately, what we do today is the correct behaviour for `source.Kind`, it has a different interpretation of the purpose of its `ctx` argument. Let's merge this as `source.Channel` doesn't work at all currently, I will open a n issue so we can clarify our interface definitions and update one or the other accordingly

alvaroaleman: (This change will break the waitForSync functionality I believe)

alvaroaleman (approved):
/lgtm
/approve

Issue #1035, opened 2020-07-08
Add option to disable stack traces for zap logs `--zap-stacktrace-level`


bharathi-tenneti:
Description:
Allow users to disable stacktraces , using `--zap-stacktrace-level`, when running the operator.

vincepri:
You should be able to register the flag in your controller, like we do for `klog`. I'm not sure if there is anything for controller-runtime to do in this case.

camilamacedo86:
@vincepri it is related to : https://github.com/kubernetes-sigs/controller-runtime/pull/915  after merge this one we would be nice be allowed to disable/enable it.

camilamacedo86:
Hi @vincepri, 

For we  `You should be able to register the flag in your controller, like we do for klog. I'm not sure if there is anything for controller-runtime to do in this case.` the user requires to do the change in the code, Am I right? If yes, I understand that the flag allows disable/enable the logs on the cluster where the project is running without code changes. 

if I am wrong here, I think we can just add an example here how to do your suggestion and then close this one.

croomes:
I had the same issue but I couldn't find a good way to disable stack traces completely.  I checked klog but couldn't find anything relevant.  Allowing the user to set `panic` as the `--zap-stacktrace-level` seems like a simple solution.

PR #1348, opened 2021-01-20
:sparkles: Allow panic as an option to -zap-stacktrace-level"

croomes:
This allows stack trace logging to be disabled for all but panic log entries.

The Zap `DisableStackTrace` config parameter (https://github.com/uber-go/zap/blob/master/config.go#L72) isn't available via the `Option` interface, so it's not possible (or at least I couldn't find a way) to disable it completely via a user-added flag.

Closes https://github.com/kubernetes-sigs/controller-runtime/issues/1035

### DIFF
--- a/pkg/log/zap/zap_test.go
+++ b/pkg/log/zap/zap_test.go
@@ -423,6 +423,19 @@ var _ = Describe("Zap log level flag options setup", func() {
 			Expect(out.StacktraceLevel.Enabled(zapcore.ErrorLevel)).To(BeTrue())
 		})
 
+		It("Should output stacktrace at panic level, with development mode set to true.", func() {
+			args := []string{"--zap-stacktrace-level=panic", "--zap-devel=true"}
+			fromFlags.BindFlags(&fs)
+			err := fs.Parse(args)
+			Expect(err).ToNot(HaveOccurred())
+			out := Options{}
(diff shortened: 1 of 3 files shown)


croomes:

Looks like un unrelated flake: "race detected during execution of test" in:
sigs.k8s.io/controller-runtime/pkg/manager

pkg/log/zap/zap_test.go
```diff
@@ -423,6 +423,17 @@ var _ = Describe("Zap log level flag options setup", func() {
...
 
+		It("Should output stacktrace at panic level, with development mode set to true.", func() {
+			args := []string{"--zap-stacktrace-level=panic", "--zap-devel=true"}
+			fromFlags.BindFlags(&fs)
+			err := fs.Parse(args)
+			Expect(err).ToNot(HaveOccurred())
+			out := Options{}
+			UseFlagOptions(&fromFlags)(&out)
+
+			Expect(out.StacktraceLevel.Enabled(zapcore.PanicLevel)).To(BeTrue())
```
alvaroaleman, line 434: Isn't the more important test that its disabled on the next level, i.E. `error`?

croomes: Added disable check for `info` and `error`.

alvaroaleman (approved):
/lgtm
/approve

thanks!

PR #1351, opened 2021-01-20
🐛 pkg/client: optionally allow caching of unstructured objects

estroz:
Cherry-pick of #1332 to release-0.7.

>Based on #615, it seems like there is intent to allow the DelegatingClient to cache unstructured objects. This is behavior that was possible to use before v0.7.0 by using the NewClientFunc manager option and creating a custom DelegatingClient.
>
>IMO, this could be considered a regression, but since a large chunk of this code was restructured in v0.7.0, I'm not sure if dropping support for unstructured caching was intentional.
>
>This PR adds a new boolean field to the NewDelegatingClientInput struct that allows users to opt-in to caching unstructured objects.

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -22,8 +22,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"k8s.io/apimachinery/pkg/types"
-
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


vincepri (approved):
/approve
/lgtm

Issue #1350, opened 2021-01-20
Allow creating a controller that doesn't watch anything at the init time.


lzang:
We have a use case that we need to generate a controller that doesn't watch anything at the beginning. 
The source of watch is added after the controller creation as it is from a remote cluster.  

This functionality can be easily achieved if we add one more function to the controller-runtime builder function which is almost the same with Build() but no watch is called.  Will submit the patch soon. Plealse review and let me know if it can be checked in. Thanks!

lzang:
Will do the alternative as suggested in the CL.

PR #1352, opened 2021-01-20
:sparkles: Allow building a controller that doesn't watch anything.

lzang:
In our use case, we need a controller that watches nothing, and the
watch source will be added later from remote clusters.

Issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1350

Signed-off-by: Zang Li <zangli@google.com>

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -204,6 +204,27 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
 	return blder.ctrl, nil
 }
 
+// BuildInterface builds a controller interface that can watch a source later.
+// It is copied from the Build function but doesn't watch anything now.
+func (blder *Builder) BuildInterface(r reconcile.Reconciler) (controller.Controller, error) {
+	if r == nil {
+		return nil, fmt.Errorf("must provide a non-nil Reconciler")
+	}
+	if blder.mgr == nil {
+		return nil, fmt.Errorf("must provide a non-nil Manager")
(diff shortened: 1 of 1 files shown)


coderanger:
-1 from me. I don't think this is common enough that it needs to be exposed in the builder API.

lzang:
> -1 from me. I don't think this is common enough that it needs to be exposed in the builder API.

Hi Noah,  let me describe our use case with more detail. We use cluster-api package to bring multiple child clusters and we use the remote package in cluster-api to watch resources in the child clusters without creating managers for each child cluster.  The remote package will send the watched results to the controller user provides.  So we want to have a controller running in the parent cluster, but doesn't watch resources in the parent cluster.  Instead it will watch the source the remote package provides.

This use case could be quite common for multi-cluster scenarios. To support the case, we could either use the simple and harmless patch in this CL, or we copy all the functions in the controller-runtime to implement a controller with work queue, retry etc. Comparing to that duplicate works, why not allow this little patch to solve the issue?  If you have other simple alternative solution, please propose. Thanks!

coderanger:
Multi-cluster operators are getting official support via the new Cluster package. You would still only use one manager.

All that said, even if you did, for whatever reason, have a need for a reconciler loop with no watches, you would drop down a level and create the controller object yourself rather than using the Builder syntax. The Builder syntax is a shortcut to simplify common use cases, but this is not a common use case.

lzang:
Thanks Noh for your information and explanation. Close the request.

PR #1354, opened 2021-01-21
:bug: Manager.Elected() should beclosed after runnables are started

vincepri:
During debugging a weird issue with some tests failing if they were
running too fast. In other words, calling Start() and Close() on
a manager too fast throws an error where the informers haven't been able
to sync, which then makes Start() fail with an error.

In an effort to improve this behavior, tried to use Elected() to wait
for leader election to start, which also waits for the cache.

During some debugging, this issue happened again and upon digging a bit
more it seems that the channel was closed before starting the runnables
in some cases.

This change reorders the close on cm.elected, which should fix the above
issue.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -473,8 +473,8 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 			}
 		} else {
 			// Treat not having leader election enabled the same as being elected.
+			cm.startLeaderElectionRunnables()
 			close(cm.elected)
-			go cm.startLeaderElectionRunnables()
 		}
 	}()
 
@@ -632,8 +632,8 @@ func (cm *controllerManager) startLeaderElection() (err error) {
 		RetryPeriod:   cm.retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
-				close(cm.elected)
 				cm.startLeaderElectionRunnables()
+				close(cm.elected)
 			},
 			OnStoppedLeading: cm.onStoppedLeading,
 		},


alvaroaleman:
Isn't the purpose of this channel to tell consumers when we are elected? Why do we need to start Runnables first? The starting is async anyways, so I don't think this will do much in terms of consistent behavior

vincepri:
The behavior is inconsistent in the two places, in one place we wait for the cache to sync and then close the channel, while in the other we were firing off a goroutine and then closing the channel.

The reordering of operation is to make sure we're consistent

DirectXMan12:
/lgtm

but I'd like to get one more set of eyes on it to approve, just in case

pkg/manager/internal.go
```diff
@@ -473,8 +473,8 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 			}
 		} else {
 			// Treat not having leader election enabled the same as being elected.
+			cm.startLeaderElectionRunnables()
 			close(cm.elected)
```
alvaroaleman, line 477: @vincepri this is consistent today. We first close the channel, the we start the Runnables.

Also, is it intended to make the `cm.StartLeaderElectionRunnables` synchronous?

vincepri: Yes, like it's here

vincepri:
/hold

for more 👀

pkg/manager/internal.go
alvaroaleman: ah now I get it, this is abusing the fact that `startLeaderElectionRunnables` synchronously waits for the cache to be synced. A tad implicit and has a good chance to be accidentally broken in the future again, but seems reasonable

estroz: Now leader and leaderless starting of pre-election runnables is the same, that makes sense to me.

PR #1355, opened 2021-01-21
✨Fix context.TODO() in the example

### LINKED ISSUE #1353: context.TODO() in the example
https://github.com/kubernetes-sigs/controller-runtime/blob/5c2b42d0dfe264fe1a187dcb11f384c0d193c042/examples/builtins/controller.go#L45

Hi, why do we use `context.TODO()` instead of using the function's `ctx`?
Looks like it's a leftover from before the `Reconcile` function received a context.

rashmigottipati:
Fixes #1353 

Signed-off-by: rashmigottipati <chowdary.grashmi@gmail.com>


vincepri (approved):
/approve
/lgtm
/ok-to-test

PR #1358, opened 2021-01-24
[WIP] :sparkles: Add support for implicit paging in un/structured clients

### LINKED ISSUE #532: Add pagination support to clients
Once #341 merges, it would be nice to add implicit support for list pagination to both `typedClient` and `unstructuredClient`.

I'd be happy to take a stab at this.

nimrodshn:
## What this does?
Following #341  this patch adds implicit paging to the un/structured clients. Fixes #532.

## Thoughts about the API:
This patch opts for adding a new Method called `ListPages`: This method lists large data in "chunks" using the `Continue` curser token with a callback to be used by the user.

## Open Questions:

- Should `ListPages` be part of the `Reader` [interface](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L47-L58)?

@ncdc @DirectXMan12 @vincepri @cben PTAL.

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -23,6 +23,12 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// DefaultPageLimit represents the default limit used for ListPaging when no "Limit"
+	// is specified as ListOption.
+	DefaultPageLimit = 100
+)
+
 // {{{ "Functional" Option Interfaces
 
 // CreateOption is some configuration that modifies options for a create request.
--- a/pkg/client/typed_client.go
+++ b/pkg/client/typed_client.go
@@ -19,6 +19,7 @@ package client
 import (
 	"context"
 
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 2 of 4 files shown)


coderanger:
This feels weird to me. What is the use case for this? Generally when a pagination cursor exists, not using it will lead to lots of weird race condition bugs when the object set changes between requests.

nimrodshn:
@coderanger I tend to agree. I guess the idea is to hide the use of `Continue` from the user. Albeit the user can simply use `Continue` and view pages of data as it is now. Please refer to #532 for more information.

coderanger:
I don't think this is what #532 was referring to. In what situation would you ever want to ask for "page 4 of the results" when you have no idea what that will even contain? The only time I could see someone doing that is in error when they want to get all items via pagination, but using this code they would get all those bugs I mentioned, which is why a pagination cursor is returned.

nimrodshn:
@coderanger I had implemented it this way b/c I thought that is what reported in #532 - how do you view it?

coderanger:
That List() would automatically check if a continue token was returned, and (give or take an option flag) automatically fetch the next page and combine it with the the results you already have, and repeat until no more data is returned. When accessing a huge list of objects that change frequently (e.g. all Pods) this can spread out the load on the API server, trading increased effective latency to remove the massive IO spike that could disrupt API server stability.

coderanger:
Or there's the option @vincepri mentioned of doing a similar looped get with an iterator function callback so you never have to hold the full list of objects in memory at once. Both could be useful in different situations.

nimrodshn:
@coderanger Interesting.. I will try and tackle those ideas. Indeed this implementation if fairly Naive.

nimrodshn:
@coderanger  Please re-tal; specifically at the implementation at the `typed_client.go` - is this the direction you had in mind? what about the `callback` do you think it is necessary? Should we add the `ListPages` method to the `Reader` interface?

pkg/client/options.go
```diff
@@ -23,6 +23,11 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// DefaultPageLimit represents the default limit used when specifyinh the 'Page' ListOption.
+	DefaultPageLimit = 100
```
coderanger, line 28: This probably needs a better name since it's not the default for List() too.

pkg/client/options.go
```diff
@@ -23,6 +23,11 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// DefaultPageLimit represents the default limit used when specifyinh the 'Page' ListOption.
```
coderanger, line 27:
```suggestion
	// DefaultPageLimit represents the default limit used when specifying the 'Page' ListOption.
```

pkg/client/typed_client.go
```diff
@@ -143,6 +144,68 @@ func (c *typedClient) Get(ctx context.Context, key ObjectKey, obj Object) error
...
+	items, err := apimeta.ExtractList(interimResult)
+	if err != nil {
+		return err
+	}
+	allItems = append(allItems, items...)
+
+	// Continue while there are more chunks.
+	for interimResult.GetContinue() != "" {
+		Continue(interimResult.GetContinue()).ApplyToList(&listOpts)
+		err = r.Get().
```
coderanger, line 186: I would re-arrange this code into a do/while structure so the actual API call doesn't have to be duplicated. Also this can use `c.List()` I think? That way we don't have that duplication as well.

```go
for {
  err := c.List(...)
  if err ...
  etc etc

  if result.GetContinue() == "" {
    break
  }
}
```

pkg/client/typed_client.go
```diff
@@ -143,6 +144,68 @@ func (c *typedClient) Get(ctx context.Context, key ObjectKey, obj Object) error
...
+
+		if err := callback(interimResult); err != nil {
+			return err
+		}
+
+		items, err = apimeta.ExtractList(interimResult)
+		if err != nil {
+			return err
+		}
+		allItems = append(allItems, items...)
```
coderanger, line 203: We don't want both a callback function and a huge list return, we should pick one pattern or the other (or move them to separate functions/options). The reason to use a visitor callback function is to avoid keeping the whole list in memory. I think it would be more Go-like to just do the callback style probably, `ExtractList()` requires a lot of runtime `reflect` magic since there is no unified API for accessing list items. But having both available could be convenient.

pkg/client/typed_client.go
nimrodshn: Oh very neat!

pkg/client/typed_client.go
coderanger: @vincepri Do you have opinions on which API to support (or both)?

pkg/client/options.go
nimrodshn: @vincepri @cben WDYT?

pkg/client/typed_client.go
nimrodshn: Gotcha. :+1:

nimrodshn: That is she can pass something like this:
```
out := &corev1.Pods{}
allItems := []runtime.Object
cl.ListPages(ctx, out,  func(obj client.ObjectList) error {
    items, err := apimeta.ExtractList(interimResult)
    if err != nil {
        return err
    }
    allItems = append(allItems, items...)
    return nil
})
```

coderanger: Indeed, or more likely cast it to the correct type and skip needing ExtractList :)

Issue #1359, opened 2021-01-25
Data race in the delegating logger


charith-elastic:
In `DelegatingLogger`, I think the `Fulfil` method needs to be protected by a lock as well. After upgrading to 0.8.1, our test suite reliably fails when run with the Go race detector. After I added locking code to the `Fulfil`  method, the race detector became happy again.

I can raise a PR for the change but I am struggling to produce a simple test case that reliably triggers the issue.

alvaroaleman:
Do you have a sample stacktrace for this race?

charith-elastic:
```
WARNING: DATA RACE
Read at 0x00c0001820c8 by goroutine 9:
  sigs.k8s.io/controller-runtime/pkg/log.(*DelegatingLogger).V()
      <autogenerated>:1 +0x54
  .../pkg/controller/common/watches.(*DynamicEnqueueRequest).AddHandler()
      .../pkg/controller/common/watches/handler.go:72 +0x33d

Previous write at 0x00c0001820c8 by goroutine 8:
  sigs.k8s.io/controller-runtime/pkg/log.(*loggerPromise).Fulfill()
      /home/cell/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/log/deleg.go:76 +0x11a
  sigs.k8s.io/controller-runtime/pkg/log.(*loggerPromise).Fulfill()
      /home/cell/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/log/deleg.go:81 +0x22f
  sigs.k8s.io/controller-runtime/pkg/log.(*DelegatingLogger).Fulfill()
      /home/cell/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/log/deleg.go:133 +0x12e
  sigs.k8s.io/controller-runtime/pkg/log.init.0.func1()
      /home/cell/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.1/pkg/log/log.go:67 +0xbb

Goroutine 9 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1437 +0x612
[...]

PR #1361, opened 2021-01-26
:warning: Fix some more races in the delegating logger

alvaroaleman:
The delegatingLogger had a logger implementation embedded, calling any
method of that would race with FullFill(), which replaces the embedded
logger.

/cc @charith-elastic 
Can you verify this fixes all remaining races?

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1359

### DIFF
--- a/pkg/log/deleg.go
+++ b/pkg/log/deleg.go
@@ -29,11 +29,11 @@ type loggerPromise struct {
 	childPromises []*loggerPromise
 	promisesLock  sync.Mutex
 
-	name *string
-	tags []interface{}
+	name  *string
+	tags  []interface{}
+	level int
 }
 
-// WithName provides a new Logger with the name appended
 func (p *loggerPromise) WithName(l *DelegatingLogger, name string) *loggerPromise {
 	res := &loggerPromise{
 		logger:       l,
@@ -61,6 +61,19 @@ func (p *loggerPromise) WithValues(l *DelegatingLogger, tags ...interface{}) *lo
 	return res
 }
 
+func (p *loggerPromise) V(l *DelegatingLogger, level int) *loggerPromise {
+	res := &loggerPromise{
+		logger:       l,
+		level:        level,
+		promisesLock: sync.Mutex{},
+	}
+
+	p.promisesLock.Lock()
+	defer p.promisesLock.Unlock()
+	p.childPromises = append(p.childPromises, res)
+	return res
+}
+
 // Fulfill instantiates the Logger with the provided logger
 func (p *loggerPromise) Fulfill(parentLogger logr.Logger) {
 	var logger = parentLogger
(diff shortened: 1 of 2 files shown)


charith-elastic:
Seems to work. Thanks @alvaroaleman 

If I understand your change correctly, wouldn't it require a lock to be obtained on every logging call even after the promise is fulfilled? Wouldn't that add some overhead?

alvaroaleman:
> If I understand your change correctly, wouldn't it require a lock to be obtained on every logging call even after the promise is fulfilled? Wouldn't that add some overhead?

Good point. I've replaced the Mutex with a RWMutex, that should minimize the overhead. We will always need something to check if the promise was fulfilled, unfortunately.

alvaroaleman:
/assign @vincepri @estroz 

WDYT about including this in 0.8? This change is incompatible in that if anyone ever did a `delgatingLogger.Logger.Something`, that wouldn't work anymore, but I very much hope no one does that, as it would defeat the purpose of the whole thing. OTOH it is fixing a data race, so that would be good to have in the release IMHO

vincepri:
Given that this is a bug that could cause a race condition (and corruption), I'm +1 merging as ⚠️ and open backport PRs as well

alvaroaleman:
Updated the title to be `:warning:` (what are our title rules for something that is both breaking and a bugfix?)

pkg/log/deleg.go
```diff
@@ -29,11 +29,11 @@ type loggerPromise struct {
...
 	promisesLock  sync.Mutex
 
-	name *string
-	tags []interface{}
+	name  *string
+	tags  []interface{}
+	level int
 }
 
-// WithName provides a new Logger with the name appended
```
estroz, line 36: Removing godocs?

estroz: I guess it doesn't matter because its an interface method

alvaroaleman: Err yeah sorry, I abused this to check if the linter correctly doesn't care about godocs of exported methods on unexported types (which it does). Since it is only visible from within the package it doesn't matter much IMO

pkg/log/deleg.go
```diff
@@ -88,21 +104,75 @@ func (p *loggerPromise) Fulfill(parentLogger logr.Logger) {
...
+
+// Info logs a non-error message with the given key/value pairs as context.
+//
+// The msg argument should be used to add some constant description to
+// the log line.  The key/value pairs can then be used to add additional
+// variable information.  The key/value pairs should alternate string
+// keys and arbitrary values.
+func (l *DelegatingLogger) Info(msg string, keysAndValues ...interface{}) {
+	l.lock.RLock()
+	defer l.lock.RUnlock()
```
estroz, line 129: Should these be read locks? Isn't `Info` writing to the underlying logger here?

alvaroaleman: The Lock protects the reference to the logger, not the method call. The actual logger implementation must also be threadsafe, but that needs to be done in that implementation, not here. Everything except `Fullfill` only reads the reference (as proven by the included tests not failing with a data race)

alvaroaleman: If we care about that, I can update the PR later to only use the lock to get the logger reference and not around the whole method call, which might improve performance slightly

estroz: Up to you.

alvaroaleman: Then I'll keep it as-is, only protecting the reference would require to allocate a pointer so it might actually be worse performance-wise and is more complex

estroz (approved):
/lgtm

PR #1362, opened 2021-01-27
:bug: Don't hotloop on channel source closure

DirectXMan12:
Previously, when we received from the channel in a channel source, we
didn't check if it was still open.  This lead to hotlooping and spamming
the source's handler/queue when the input channel is closed.

Now, we stop the source when the channel is closed, similarly to what
happens if the context is closed.


alvaroaleman (approved):
/lgtm

PR #1365, opened 2021-02-01
✨ Add example for using apply client

Jefftree:
Example for using the server side apply client with CRDs

Took the suggestion of using the Patch to embed the applyconfiguration.

```
cl.Patch(context.TODO(), fooObj, client.ApplyFrom(applyConfig), owner);
```

where `fooObj` is the underlying type (to contain the returned patch object) and `applyConfig` is the applyconfiguration.

I'm not exactly sure how useful the applyconfig example is, let me know if you have any suggestions on how it can be improved. Thanks!


DirectXMan12:
/hold

(work in progress)

pkg/client/typed_client.go
```diff
@@ -108,24 +111,45 @@ func (c *typedClient) DeleteAllOf(ctx context.Context, obj Object, opts ...Delet
 		Error()
 }
 
+// Sets the name and namespace for the returned object if an ApplyConfiguration was supplied
+func setNameNamespace(obj Object, ac interface{}) {
```
Jefftree, line 115: I can add some additional checks for invalid structs, but I think this is the best way to copy over the fields without explicitly modifying k/k in 1.22 to declare some interface.

zoetrope:
> I'm not exactly sure how useful the applyconfig example is, let me know if you have any suggestions on how it can be improved. Thanks!

I believe the applyconfig is very useful for calculating diff when CRD has fields like PodTemplate and ServiceTemplate.
My samples can be found below.
https://github.com/zoetrope/reconcile-tips/blob/typed-ssa/controllers/myapp_controller.go

However, controller-gen doesn't seem to be able to generate deepcopy of CRs including applyconfigs.
https://github.com/kubernetes-sigs/controller-tools/issues/570

Do you have a good solution?

pkg/client/patch.go
```diff
@@ -211,3 +212,34 @@ func (p applyPatch) Data(obj Object) ([]byte, error) {
...
+
+// Type implements Patch.
+func (p applyFromPatch) Type() types.PatchType {
+	return types.ApplyPatchType
+}
+
+// Sets the name and namespace for the returned object if an ApplyConfiguration was supplied
+func setNameNamespace(obj Object, ac interface{}) {
+	obj.SetName(reflect.ValueOf(ac).Elem().FieldByName("Name").Elem().String())
+	obj.SetNamespace(reflect.ValueOf(ac).Elem().FieldByName("Namespace").Elem().String())
```
cbandy, line 229: 🤔 I'm surprised that `ObjectMetaApplyConfiguration` doesn't implement `metav1.Object` with getters and setters.

Jefftree: Implementing the full `metav1.Object` doesn't make sense because other than Name and Namespace, most of the other fields are optional in an ApplyConfiguration while mandatory in an actual metav1.Object. Since the fields are mandatory in the metav1.Object, the interface prevents us from being able to return nil for fields. Eg: CreationTimestamp, Labels, etc.

We could technically implement a partial metav1.Object interface with just the getters and setters for Name and Namespace, but that'll require a change in k/k and push this code to be compatible with 1.22+ instead of 1.21+. 

cc @jpbetz

cbandy: > the interface prevents us from being able to return nil for fields. Eg: CreationTimestamp, Labels, etc.

I haven't thought this through, but I would expect a zero value from these when they are unset. The behavior should be like `metav1.ObjectMeta`, yes?

Looking closer, I see AC that have `*ObjectMetaApplyConfiguration` while their analogs have `metav1.ObjectMeta // +optional`... It's possible to call methods on a zero value of the latter, but calling methods on a `nil` pointer is not going to work out. Forget I asked about `metav1.Object`. 🙇🏻 

I believe this code needs to guard against nil ObjectMetaApplyConfiguration and nil Name and nil Namespace. I'm certain that `reflect.ValueOf(nilPointer).Elem().IsValid() == false`.

🤔 Do we need this at all? None of the other Patch implementations do this.

jpbetz: So it turns out that `metav1.Object` is almost works out really well in apply configurations because all the fields are marked as 'omitempty' already. Honestly, the biggest problem with the `metav1.Object` is that it includes `ManagedFields` which is not allowed in apply configurations (you might notice if you look closely that `ObjectMetaApplyConfiguration` does not have `ManagedFields`).

Note that because we didn't implement `metav1.Object` on the kubernetes native types we've really can't require it in the Kubebuilder APIs (since being able to write controllers using Kubebuilder that applies the native types seems like a pretty reasonable requirement?). I'll admit this is a bit unfortunate.

pkg/client/patch.go
```diff
@@ -211,3 +212,34 @@ func (p applyPatch) Data(obj Object) ([]byte, error) {
...
+type applyFromPatch struct {
+	patch interface{}
+}
+
+// Type implements Patch.
+func (p applyFromPatch) Type() types.PatchType {
+	return types.ApplyPatchType
+}
+
+// Sets the name and namespace for the returned object if an ApplyConfiguration was supplied
```
jpbetz, line 226: very minor nit, still follow godoc conventions here? E.g. "setNameNamespace sets the ..."

pkg/client/patch.go
```diff
@@ -211,3 +212,34 @@ func (p applyPatch) Data(obj Object) ([]byte, error) {
...
+func setNameNamespace(obj Object, ac interface{}) {
+	obj.SetName(reflect.ValueOf(ac).Elem().FieldByName("Name").Elem().String())
```
jpbetz, line 228: These reflections calls seem insufficiently guarded. `ApplyFrom` takes an `interface{}` and we don't check anywhere if the fields we expect to access here are actually present. Maybe perform the assertions in `ApplyFrom`?

pkg/client/patch.go
```diff
@@ -211,3 +212,34 @@ func (p applyPatch) Data(obj Object) ([]byte, error) {
...
+	patch interface{}
+}
+
+// Type implements Patch.
+func (p applyFromPatch) Type() types.PatchType {
+	return types.ApplyPatchType
+}
+
+// Sets the name and namespace for the returned object if an ApplyConfiguration was supplied
+func setNameNamespace(obj Object, ac interface{}) {
```
jpbetz, line 227: Does this work right with cluster scoped objects? I suspect it doesn't?

pkg/client/patch.go
Jefftree: Yeah I agree. I wasn't sure if this was the correct approach so most of the error checking wasn't included. Will improve the handling for errors and edge cases.

pkg/client/patch.go
jpbetz: Overall, I don't think we should implement `metav1.Object` right now because (1) we don't implement it on the apply configurations for the kubernetes native types and (2) we exclude the `ManagedFields` field.

Jefftree:
@vincepri @joelanford Is this something you can take a look at? Thanks!

pkg/client/patch.go
```diff
@@ -211,3 +212,55 @@ func (p applyPatch) Data(obj Object) ([]byte, error) {
...
+}
+
+// setNameNamespace sets name and namespace for the returned object if an ApplyConfiguration was supplied.
+func setNameNamespace(obj Object, ac interface{}) error {
+	if !reflect.ValueOf(ac).Elem().IsValid() || reflect.ValueOf(ac).Elem().Kind() != reflect.Struct {
+		return fmt.Errorf("ApplyConfiguration is not a valid struct")
+	}
+	// Since Name and Namespace are embedded objects, nil checks must be
+	// performed against the embedded ObjectMeta object first.
+	if reflect.ValueOf(ac).Elem().FieldByName("ObjectMetaApplyConfiguration").IsNil() {
```
alvaroaleman, line 233: Instead of using an empty interface and reflect, can we define an interface with getters/setters fro the properties we need here (if it doesn't exist already)?

That would also make the whole thing a lot more intuitive to use and give at least some compile-time verification

examples/applyconfig/main.go
```diff
@@ -0,0 +1,141 @@
...
+	// To use the SSA typed client, use the ApplyConfiguration generated types instead of the types defined in v1
+	applyConfig1 := (&ac.FooApplyConfiguration{
+		TypeMetaApplyConfiguration:   *metav1ac.TypeMeta().WithKind("Foo").WithAPIVersion("applytest.kubebuilder.io/v1"),
+		ObjectMetaApplyConfiguration: metav1ac.ObjectMeta().WithName("acdefault").WithNamespace("default"),
+	}).WithNonNullableField("value1")
+
+	// Without ApplyConfiguration, NonNullableField is not an optional field so it must be specified if Foo{} was used instead of the ApplyConfiguration
+	// The ApplyConfiguration uses pointers for all objects, so omitting a field is equivalent to saying that it is not of interest in this Apply
+	applyConfig2 := (&ac.FooApplyConfiguration{
+		TypeMetaApplyConfiguration: *metav1ac.TypeMeta().WithKind("Foo").WithAPIVersion("applytest.kubebuilder.io/v1"),
```
alvaroaleman, line 119: Do we need to set type meta? We can derive this from the `obj` in patch or not?

examples/applyconfig/api/v1/ac/zz_generated.applyconfigurations.go
```diff
@@ -0,0 +1,49 @@
...
+	b.NullableField = &value
+	return b
+}
+
+// FooApplyConfiguration represents a declarative configuration of the Foo type for use
+// with apply.
+func Foo() *FooApplyConfiguration {
+	return &FooApplyConfiguration{}
+}
+func (ac *FooApplyConfiguration) GetReferenceType() runtime.Object {
```
alvaroaleman, line 39: Can we maybe include this method when dealing with appyconfigs in an interface, so that we end up with an interface that is not fulfilled by `metav1.Object`?

pkg/client/patch.go
Jefftree: I thought about doing this and it's easy to accomplish with just the CRD types. However, the apply configurations for native k/k resources don't implement the interface at the moment. Since it takes some time for k/k changes to propagate, would it be possible to keep this as is and introduce the getters/setters in a future version?

examples/applyconfig/main.go
Jefftree: can you point me in the right direction for finding this? I wasn't find an easy way to derive this since the applyconfig objects are constructed rather than obtained from the apiserver.

imheresamir:
Eagerly awaiting this PR to land!

pkg/client/patch.go
seh: As far as I can tell, the existing `Patch` methods derive the namespace from the supplied `Object`, not from the patch content—at least for ["typed" objects](https://sourcegraph.com/github.com/kubernetes-sigs/controller-runtime@v0.11.1/-/blob/pkg/client/typed_client.go?L125) and [unstructured objects](https://sourcegraph.com/github.com/kubernetes-sigs/controller-runtime@v0.11.1/-/blob/pkg/client/unstructured_client.go?L161), anyway.

seh: As far as I can tell, the existing `Patch` methods derive the namespace from the supplied `Object`, not from the patch content—at least for ["typed" objects](https://sourcegraph.com/github.com/kubernetes-sigs/controller-runtime@v0.11.1/-/blob/pkg/client/typed_client.go?L125) and [unstructured objects](https://sourcegraph.com/github.com/kubernetes-sigs/controller-runtime@v0.11.1/-/blob/pkg/client/unstructured_client.go?L161), anyway.

cannonpalms:
It'd be lovely if this were revived and merged! The SSA story in controller-runtime is woefully underdocumented.

PR #1371, opened 2021-02-05
:warning: Support global controller options in component config

vincepri:
This change adds support for our v1alpha1 ComponentConfig types to
expose configuration options for controllers. The only two current
exposed options are concurrency (done via a GroupKind map) and the cache
sync timeout option.

This is a breaking change, given that we're adding a new required method
to the Manager's interface.


vincepri:
Need some feedback on this one :)

pkg/config/v1alpha1/types.go
```diff
@@ -50,9 +52,14 @@ type ControllerManagerConfigurationSpec struct {
...
 	// To disable graceful shutdown, set to time.Duration(0)
 	// To use graceful shutdown without timeout, set to a negative duration, e.G. time.Duration(-1)
-	// The graceful shutdown is skipped for safety reasons in case the leadere election lease is lost.
+	// The graceful shutdown is skipped for safety reasons in case the leader election lease is lost.
 	GracefulShutdownTimeout *metav1.Duration `json:"gracefulShutDown,omitempty"`
 
+	// Controller contains global configuration options for controllers
+	// registered within this manager.
+	// +optional
+	Controller *ControllerConfigurationSpec `json:"controller,omitempty"`
```
christopherhein, line 61: In the description, you called this out as a breaking change but all the fields have optional? Am I missing something?

pkg/manager/manager.go
```diff
@@ -407,6 +416,14 @@ func (o Options) AndFrom(loader config.ControllerManagerConfiguration) (Options,
 		o.CertDir = newObj.Webhook.CertDir
 	}
 
+	if o.Controller.CacheSyncTimeout == nil && newObj.Controller.CacheSyncTimeout != nil {
```
christopherhein, line 419: should we check if `o.Controller` is `nil` first?

pkg/config/v1alpha1/types.go
```diff
@@ -66,6 +73,29 @@ type ControllerManagerConfigurationSpec struct {
 	Webhook ControllerWebhook `json:"webhook,omitempty"`
 }
 
+// ControllerConfigurationSpec defines the global configuration for
+// controllers registered with the manager.
+type ControllerConfigurationSpec struct {
+	// GroupKindConcurrency is a map from a Kind to the number of concurrent reconciliation
```
christopherhein, line 79: Love this idea, this will make it much easier to make per type configs.

christopherhein (commented):
This looks really good, couple of questions in the review but otherwise 👍

pkg/manager/manager.go
vincepri: o.Controller is a non-pointer, to avoid checking it here

pkg/config/v1alpha1/types.go
vincepri: The manager's interface new method addition is a breaking change

vincepri: Specifically, https://github.com/kubernetes-sigs/controller-runtime/pull/1371/files#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR93-R95

christopherhein: Oh, I see. For the most part the docs (at least the ones in Kubebuilder on `master`) point to embedding the core type so hopefully, anyone that has gone and implemented this would be able to receive the update to resolve this but for folks that aren't that makes sense. luckily we're still alpha 🎉

pkg/manager/manager.go
christopherhein: oh gotcha, confused with the exported controllerconfiguration. makes sense.

vincepri:
/assign @alvaroaleman 

for another set of 👀

christopherhein (approved):
/lgtm

:+1: to this, nice work!

Issue #1346, opened 2021-01-19
source.Kind and source.Channel interpret the ctx arg to their Start method differently


alvaroaleman:
`source.Channel` will shutdown when the `context.Context` passed to its `Start` gets cancelled, whereas `source.Kind` uses that arg go get an Informer and then ignores it.

Because of that, it is currently not possible to implement a proper cacheSyncTimeout without breaking `source.Channel`.

We should:
* Clarify on the interface definition what the meaning of the `context.Context` arg to `Start` is
* Update either `source.Kind` or `source.Channel` accordingly

History:
* Issue to add WaitForCacheSync: https://github.com/kubernetes-sigs/controller-runtime/issues/1219
* WaitForCacheSync implementation: https://github.com/kubernetes-sigs/controller-runtime/pull/1247
* Bugreport that this breaks `source.Channel`: https://github.com/kubernetes-sigs/controller-runtime/issues/1343
* Fix to not break `source.Channel` anymore but break the `waitForCacheSync` functionality: https://github.com/kubernetes-sigs/controller-runtime/pull/1345

charith-elastic:
Hi @alvaroaleman I am most likely missing something; I don't see why #1345 breaks `waitForCacheSync`. Isn't that case being handled here: https://github.com/kubernetes-sigs/controller-runtime/blob/73c52e88006855d20f7bdba45e56c7c1a78c0e52/pkg/internal/controller/controller.go#L174-L196

alvaroaleman:
@charith-elastic because if the cache is already started, which is usually the case as the manager does that, we will:
* Get an informer: https://github.com/kubernetes-sigs/controller-runtime/blob/73c52e88006855d20f7bdba45e56c7c1a78c0e52/pkg/source/source.go#L114
* This then by default calls this implementation: https://github.com/kubernetes-sigs/controller-runtime/blob/73c52e88006855d20f7bdba45e56c7c1a78c0e52/pkg/cache/informer_cache.go#L149
* This will wait for cachesync here and for example in case of rbac issues just hang: https://github.com/kubernetes-sigs/controller-runtime/blob/73c52e88006855d20f7bdba45e56c7c1a78c0e52/pkg/cache/internal/informers_map.go#L187

charith-elastic:
I see. Thanks for the explanation.

charith-elastic:
I have a couple of ideas about supporting cache sync timeout:
[...]

PR #1373, opened 2021-02-06
:sparkles: Add tests in controller that use source.Channel

### LINKED ISSUE #1368: Add integration test that uses source.Channel
We broke source.Channel entirely and didn't notice that until after a release. We should add an integration test that verifies that events written to a source.Channel trigger a reconciliation.

varshaprasad96:
The additional tests verify controller implementation when `source.Channel` is passed.

Fixes #1368
Will also be helpful for discussing #1346

### DIFF
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -28,16 +28,19 @@ import (
 	dto "github.com/prometheus/client_model/go"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/cache/informertest"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllertest"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
...
+			p := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "bar"},
+			}
+			evt := event.GenericEvent{
+				Object: p,
+			}
+
+			ins := &source.Channel{Source: ch}
+			ins.DestBufferSize = 1
+			Expect(inject.StopChannelInto(ctx.Done(), ins)).To(BeTrue())
```
alvaroaleman, line 195: We just deprecated the injection, lets not add more code that uses it

pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
...
+			ins := &source.Channel{}
+			Expect(inject.StopChannelInto(make(<-chan struct{}), ins)).To(BeTrue())
+
+			ctrl.startWatches = []watchDescription{{
+				src: &source.Channel{},
+			}}
+
+			errChan := make(chan error)
+
+			go func() {
```
alvaroaleman, line 264: Same comment re extra routine

pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
...
+			defer cancel()
+
+			ins := &source.Channel{Source: ch}
+			ctrl.startWatches = []watchDescription{{
+				src: ins,
+			}}
+
+			errChan := make(chan error)
+
+			go func() {
```
alvaroaleman, line 239: We can simplify this by not starting in a goroutine that writes into a channel and then waiting for the channel and instead just start in the main goroutine or not?

pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
...
+				},
+			}}
+
+			go func() {
+				defer GinkgoRecover()
+				Expect(ctrl.Start(ctx)).To(Succeed())
+			}()
+
+			select {
+			case <-processed:
```
alvaroaleman, line 220: Nit: You can replace the single case select statement with a simple
```
<- processed
```

pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
 			close(done)
 		}, 10.0)
 
+		It("should not error when source.channel is passed", func(done Done) {
```
alvaroaleman, line 175: I would probably rename this to `should process events from source.Channel`. While it is possible that `source.Channel` could error and that would fail the test, the main failure we are testing for here will manifest as the test timing out and if someone doesn't know the test they'd be like "Well it doesn't error, why is it failing anyways?"

pkg/internal/controller/controller_test.go
```diff
@@ -169,6 +172,108 @@ var _ = Describe("controller", func() {
...
+			ch <- evt
+
+			ctrl.startWatches = []watchDescription{{
+				src: ins,
+				handler: handler.Funcs{
+					GenericFunc: func(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+						defer GinkgoRecover()
+						eventCount++
+
+						if eventCount > 0 {
```
alvaroaleman, line 207: I don't think we need the `eventCount` variable in any way, success is indicated by channel closing which is currently just hidden behind a complicate way of expressing `if true`

alvaroaleman (commented):
Thanks for taking the time for this!

pkg/internal/controller/controller_test.go
varshaprasad96: While starting the source (with channel passed) we check if the stop channel has been injected and return an error if not present ([here](https://github.com/kubernetes-sigs/controller-runtime/blob/a8c19c49e49cfba2aa486ff322cbe5222d6da533/pkg/source/source.go#L208)). I think we may have to modify the implementation while [`starting`](https://github.com/kubernetes-sigs/controller-runtime/blob/a8c19c49e49cfba2aa486ff322cbe5222d6da533/pkg/source/source.go#L197) source.channel to remove this or is there any other way to go about this?

pkg/internal/controller/controller_test.go
varshaprasad96: This is blocking. Starting in main routine is causing the test to time out.

alvaroaleman: I know its blocking, my point was that its needlessly complicated to start it in a distinct goroutine.

```
errChan := make(chan error)
go func(){
  errChan <- someFunc()
}()
e := <-ErrChan
```

Is definitely semantically equivalent to
```
a := someFunc()
```

The test here works fine for me when changing it like described above and does not time out.

alvaroaleman (approved):
/lgtm
/approve

PR #1374, opened 2021-02-07
✨ thread context through to some library functions

eaceaser:
Cleans up some context.TODOs() by plumbing contexts through. This is a breaking change to the envtest setup/teardown API, by threading a context parameter through `Stop`, `Start`, `InstallCRDs`, `UninstallCRDs` etc. I think it's reasonable to expect a context in the `envtest` API,  considering that a user is likely instantiating a context already for interacting with the control plane via a Client.

As for why we need it -- this is an API cleanup that improves the overall code quality by removing `context.TODO`s. I noticed it while familiarizing myself with the codebase and thought it'd be a good way to get used to making changes to the project.

### DIFF
--- a/pkg/builder/builder_suite_test.go
+++ b/pkg/builder/builder_suite_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package builder
 
 import (
+	"context"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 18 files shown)


pkg/builder/builder_suite_test.go
```diff
@@ -44,8 +45,10 @@ func TestBuilder(t *testing.T) {
 
 var testenv *envtest.Environment
 var cfg *rest.Config
+var ctx context.Context
```
vincepri, line 48: This should probably use `SetupSignalHandler()`

PR #1380, opened 2021-02-10
✨  deduplicate repeated requests in a single update event

mars1024:
Now the `Update` handler of `EnqueueRequestForObject` will handle the old and new objects respectively, which will enqueue two requests, even the old and new objects have same name and namespace, just like [https://github.com/kubernetes-sigs/controller-runtime/blob/0247574810d7c7a7dfc234f8064e85a50cc814f2/pkg/handler/enqueue.go#L49-L67](https://github.com/kubernetes-sigs/controller-runtime/blob/0247574810d7c7a7dfc234f8064e85a50cc814f2/pkg/handler/enqueue.go#L49-L67)
IMO, the update event is a level trigger in kubernetes, so the same requests will lead to duplicated reconciliations, which are useless.
So I introduce a `RequestSet` to help collect and deduplicate requests in a single event to solve the issue above.

### DIFF
--- a/pkg/handler/enqueue.go
+++ b/pkg/handler/enqueue.go
@@ -48,30 +48,18 @@ func (e *EnqueueRequestForObject) Create(evt event.CreateEvent, q workqueue.Rate
 
 // Update implements EventHandler
 func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
-	// requestSet will help to deduplicate requests in a single event
-	requestSet := reconcile.NewRequestSet()
-
-	if evt.ObjectOld != nil {
-		requestSet.Insert(reconcile.Request{NamespacedName: types.NamespacedName{
+	if evt.ObjectNew != nil {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+			Name:      evt.ObjectNew.GetName(),
+			Namespace: evt.ObjectNew.GetNamespace(),
+		}})
+	} else if evt.ObjectOld != nil {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
 			Name:      evt.ObjectOld.GetName(),
 			Namespace: evt.ObjectOld.GetNamespace(),
 		}})
 	} else {
-		enqueueLog.Error(nil, "UpdateEvent received with no old metadata", "event", evt)
-	}
-
-	if evt.ObjectNew != nil {
-		requestSet.Insert(reconcile.Request{NamespacedName: types.NamespacedName{
-			Name:      evt.ObjectNew.GetName(),
-			Namespace: evt.ObjectNew.GetNamespace(),
-		}})
-	} else {
-		enqueueLog.Error(nil, "UpdateEvent received with no new metadata", "event", evt)
-	}
-
-	requests := requestSet.List()
-	for i := range requests {
-		q.Add(requests[i])
+		enqueueLog.Error(nil, "UpdateEvent received with no metadata", "event", evt)
 	}
 }
 
--- a/pkg/handler/enqueue.go
+++ b/pkg/handler/enqueue.go
@@ -19,6 +19,7 @@ package handler
 import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
+
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -47,8 +48,11 @@ func (e *EnqueueRequestForObject) Create(evt event.CreateEvent, q workqueue.Rate
 
 // Update implements EventHandler
 func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	// requestSet will help to deduplicate requests in a single event
+	requestSet := reconcile.NewRequestSet()
+
 	if evt.ObjectOld != nil {
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		requestSet.Insert(reconcile.Request{NamespacedName: types.NamespacedName{
 			Name:      evt.ObjectOld.GetName(),
 			Namespace: evt.ObjectOld.GetNamespace(),
 		}})
@@ -57,12 +61,17 @@ func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.Rate
 	}
 
 	if evt.ObjectNew != nil {
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		requestSet.Insert(reconcile.Request{NamespacedName: types.NamespacedName{
 			Name:      evt.ObjectNew.GetName(),
 			Namespace: evt.ObjectNew.GetNamespace(),
 		}})
 	} else {
 		enqueueLog.Error(nil, "UpdateEvent received with no new metadata", "event", evt)
+	}
+
+	requests := requestSet.List()
+	for i := range requests {
+		q.Add(requests[i])
 	}
 }
 
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -58,8 +58,21 @@ func (e *enqueueRequestsFromMapFunc) Create(evt event.CreateEvent, q workqueue.R
 
 // Update implements EventHandler
 func (e *enqueueRequestsFromMapFunc) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
-	e.mapAndEnqueue(q, evt.ObjectOld)
-	e.mapAndEnqueue(q, evt.ObjectNew)
+	// requestSet will help to deduplicate requests in a single event
+	requestSet := reconcile.NewRequestSet()
+
+	// squash requests from both old and new objects with ordered set
+	for _, req := range e.toRequests(evt.ObjectOld) {
+		requestSet.Insert(req)
+	}
+	for _, req := range e.toRequests(evt.ObjectNew) {
+		requestSet.Insert(req)
+	}
+
+	// enqueue all deduplicated requests
+	for _, req := range requestSet.List() {
+		q.Add(req)
+	}
 }
 
 // Delete implements EventHandler
--- a/pkg/handler/enqueue_owner.go
+++ b/pkg/handler/enqueue_owner.go
@@ -66,10 +66,19 @@ func (e *EnqueueRequestForOwner) Create(evt event.CreateEvent, q workqueue.RateL
 
 // Update implements EventHandler
 func (e *EnqueueRequestForOwner) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	// requestSet will help to deduplicate requests in a single event
+	requestSet := reconcile.NewRequestSet()
+
+	// squash requests from both old and new objects with ordered set
 	for _, req := range e.getOwnerReconcileRequest(evt.ObjectOld) {
-		q.Add(req)
+		requestSet.Insert(req)
 	}
 	for _, req := range e.getOwnerReconcileRequest(evt.ObjectNew) {
+		requestSet.Insert(req)
+	}
+
+	// enqueue all deduplicated requests
+	for _, req := range requestSet.List() {
 		q.Add(req)
 	}
 }
--- a/pkg/handler/eventhandler_test.go
+++ b/pkg/handler/eventhandler_test.go
@@ -114,6 +114,26 @@ var _ = Describe("Eventhandler", func() {
 				close(done)
 			})
 
+		It("should enqueue a single Request if both objects have same Name and Namespace in the UpdateEvent.",
+			func(done Done) {
+				newPod := pod.DeepCopy()
+
+				evt := event.UpdateEvent{
+					ObjectOld: pod,
+					ObjectNew: newPod,
+				}
+				instance.Update(evt, q)
+				Expect(q.Len()).To(Equal(1))
+
+				i, _ := q.Get()
+				Expect(i).NotTo(BeNil())
+				req, ok := i.(reconcile.Request)
+				Expect(ok).To(BeTrue())
+				Expect(req.NamespacedName).To(Equal(types.NamespacedName{Namespace: "biz", Name: "baz"}))
+
+				close(done)
+			})
+
 		It("should enqueue a Request with the Name / Namespace of the object in the GenericEvent.", func(done Done) {
 			evt := event.GenericEvent{
 				Object: pod,
--- a/pkg/reconcile/reconcile.go
+++ b/pkg/reconcile/reconcile.go
@@ -18,6 +18,7 @@ package reconcile
 
 import (
 	"context"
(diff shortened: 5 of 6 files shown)


pkg/handler/enqueue.go
```diff
@@ -47,8 +48,11 @@ func (e *EnqueueRequestForObject) Create(evt event.CreateEvent, q workqueue.Rate
 
 // Update implements EventHandler
 func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	// requestSet will help to deduplicate requests in a single event
+	requestSet := reconcile.NewRequestSet()
+
 	if evt.ObjectOld != nil {
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		requestSet.Insert(reconcile.Request{NamespacedName: types.NamespacedName{
```
alvaroaleman, line 55: @mars1024 thanks for your effort! However, the workqueue already deduplicates so this isn't needed: https://github.com/kubernetes/kubernetes/blob/3433f099fb0b5c51d3b549372364f818242f7312/staging/src/k8s.io/client-go/util/workqueue/queue.go#L120-L126

/hold

mars1024: @alvaroaleman Thanks for you reply, but I don't think workqueue could squash all repeated requests into one, as we know, workqueue is not only called by event handler, but also called by controller at the same time.

If workqueue is called as the following steps,
- event handler called `workqueue.Add` with request A, `dirty` set and `queue` slice will have A
- controller called `workqueue.Get` to get A for reconciliation, then `dirty` set and `queue` slice will lose A and `processing` set will have A
- event handler called `workqueue.Add` with request A again in the same update event, `dirty` set will have A again
- controller called `workqueue.Done` after finishing reconciliation A, `processing` set will lose A and `queue` slice will have A because `dirty` set have
- controller called `workqueue.Get` in another loop, then get A for reconciliation again because `queue` slice have
- ...

Just as above, a single update event will trigger two reconciliations for A, IMO, this is unnecessary.

mars1024: friendly ping @alvaroaleman , could you help to take a look at the comments above?

alvaroaleman: I see, yeah it can race with the deqeueing.

pkg/handler/enqueue.go
```diff
@@ -47,8 +48,11 @@ func (e *EnqueueRequestForObject) Create(evt event.CreateEvent, q workqueue.Rate
 
 // Update implements EventHandler
 func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	// requestSet will help to deduplicate requests in a single event
+	requestSet := reconcile.NewRequestSet()
```
alvaroaleman, line 52:
```suggestion
	var namespace, name string
```

Lets simplify this by just storing these two here and adding a check inside the `if evt.ObjectNew != nil` to avoid enqueueing when namespace/name are the same (which I believe will always be the case, I don't think its possible to rename objects).

There is no need for a new type, threadsafety or extending our public api with aspects of such a tiny implementation detail here.

pkg/handler/enqueue.go
coderanger: Can this not be fixed in workqueue?

alvaroaleman: I think the only way to fix that in the workqueue would be to add the ability to take multiple items at once

mars1024: Hi @alvaroaleman , so your final suggestion is fixing this issue in `workqueue` package ? Maybe I can try, but I don't see the difference of implementing deduplication in `workqueue` or `handler`, could you explain this for me?

alvaroaleman: That was a comment for @coderanger and I was referring to the upstream `workqueue` package. I would expect that to cause some debate upstream, though. I don't see how we could get the workqueue itself to deduplicate in such a case safely without atomically taking more than one item.

Downstream, this needs to be in the handler.

coderanger: Understandable that doing it here is probably quicker but this feels a bit off to me, redundant layers that are very similar :) I guess I still don't see the core issue. If the second `Add` happens after the first `Get`, then reconciling twice is correct. We don't know where in the process things would be. In some cases it might be redundant but in the general case?

alvaroaleman: The reconciliation will only look at the state of the object at the time of reconciliation. If we enqueue twice for one event, one of those will end up being a no-op. I agree its a micro optimization. I personally don't have strong feelings one way or another.

mars1024: Personally I prefer to do code changes on `handler`, just because in the old days of `client-go`, `sample-controller` was enqueuing object only once for update event, like [https://github.com/kubernetes/sample-controller/blob/80a6a3c0f86bb65beec9ab6a44fe7ea967ae3ede/controller.go#L131-L140](https://github.com/kubernetes/sample-controller/blob/80a6a3c0f86bb65beec9ab6a44fe7ea967ae3ede/controller.go#L131-L140)
Deduplication seems more reasonable to be done by `handler`.

pkg/handler/enqueue.go
mars1024: @alvaroaleman  Thanks for your comment, but your simple way seems only work for `EnqueueRequestForObject` handler, in some other handlers, like `EnqueueRequestsForMapFunc` [https://github.com/kubernetes-sigs/controller-runtime/blob/e1a725df2743147795e5dfc8275365f7ada24805/pkg/handler/enqueue_mapped.go#L60-L63](https://github.com/kubernetes-sigs/controller-runtime/blob/e1a725df2743147795e5dfc8275365f7ada24805/pkg/handler/enqueue_mapped.go#L60-L63), it's hard to replace `RequeueSet` simply by two variables namespace & name because there probably will be several requests.
I provide a more general way to deduplicate requests in a single event, though it looks a little bit complicated.
But I have no objections on your suggestion, just we need to find a better way to fix this issue for **all** handlers.

alvaroaleman: You can simply split `mapAndEnqueue` into one func that creates a slice of `types.NamespacedName` and one func that iterates and enqueues. Then just use the former in `Update` and deduplicate before enqueuing. Also again, this is an internal detail that does not need extending of our public api.

pkg/handler/enqueue.go
coderanger: So the goal is just to deal with the short time window between `Get` and the reconcile actually starting?

alvaroaleman: Yeah, if we do add, then reconcile gets it, then add, it gets reconciled twice. If we do both adds before the reconcile gets it, this won't make a difference, as the workqueue will de-duplicate.

coderanger: I mean that overall seems correct. If another event comes in during the middle of a reconcile, we need to reconcile again since that change might not have been fully seen by the running one, depending on the code and how far along it was.

alvaroaleman: @coderanger that is true but the issue here is that one update even might trigger two reconciles if the update event handler enqueues oldObj-> reconciler pops -> update event handler enqueues newObj

A different event must always trigger another reconciliation, that is correct.

coderanger: I've never really understood why we enqueue ObjOld since by definition it shouldn't be relevant. The only case where it would matter is detecting object renames which Kubernetes doesn't support anyway. Maybe some kind of edge case with forced overwrites and finalizers?

coderanger: But in a more concrete answer, an `AddMany` API in workqueue seems like a reasonable thing to add :) Would be the same as Add but a `...` args list (or even just make Add take ...) and make it atomic-y? Might be literally the same code as this but inside Add.

alvaroaleman: > I've never really understood why we enqueue ObjOld since by definition it shouldn't be relevant. The only case where it would matter is detecting object renames which Kubernetes doesn't support anyway. Maybe some kind of edge case with forced overwrites and finalizers?

Yeah I wasn't entirely sure about that, does the kube api guarantee that updates have an old object and a new object?

> But in a more concrete answer, an AddMany API in workqueue seems like a reasonable thing to add :) Would be the same as Add but a ... args list (or even just make Add take ...) and make it atomic-y? Might be literally the same code as this but inside Add.

Give it a stab upstream? :)

coderanger: > Yeah I wasn't entirely sure about that, does the kube api guarantee that updates have an old object and a new object?

Even if they didn't, simpler code would be more like "if new enqueue new else if old enqueue old else wat". That would be a lot less tracking stuff in here, and removes allocating a set on each event.

coderanger: Yeah the more I look at this code, there is no reason why any event should ever result in more than one reconcile request. The existing code is just wrong, or is trying to plan for impossible events (like object renames or something).

coderanger: Also both the Owner and Mapped handlers do similar thing, though there it's more defensible since you can change an owner in an update and reconciling both seems reasonable. But that would need yet more complexity to do the fast check >_<

mars1024: Emmm... So a clear and precise solution hasn't came up, `if (new) {enqueue new} else if (old) {enqueue old} else {wat}` seems a good way for Object handler, but for Owner and Mapped handlers, both old and new objects should be handled, ref to [pod update event handler in replica set controller](https://github.com/kubernetes/kubernetes/blob/c417626c4912161eb9aecafe53e8b081b7105e7b/pkg/controller/replicaset/replica_set.go#L399-L469). We have to admit that a single update event can trigger one or several requests.

pkg/handler/enqueue.go
mars1024: Just a little afraid of doing some similar inline codes... :)

pkg/handler/enqueue.go
alvaroaleman: This PR doesn't change owner and mapping handler, so why are we debating them? I like the idea of @coderanger to just check upstream if they are ok changing the workqueue to allow adding multiple items at once, that is by far the most elegant solution of all

coderanger: > This PR doesn't change owner and mapping handler, so why are we debating them?

Because those have exactly the same issue and if we fix it for one, we should probably fix it for all of them? If we were going to move forward with this solution, it would need to be applied there too :)

mars1024: > > This PR doesn't change owner and mapping handler, so why are we debating them?
> 
> Because those have exactly the same issue and if we fix it for one, we should probably fix it for all of them? If we were going to move forward with this solution, it would need to be applied there too :)

Yes, I didn't do some code changes on other handlers is not because they don't have similar issues, is just because i'm not sure that this is really an issue (of course, we believe that it is now) ...

@alvaroaleman @coderanger Could you help to invoke more members for a discussion or a decision on changing`workqueue` package?

coderanger: I'm working on a PR for client-go now. Or rather working out how to test it since that's the hard part :)

coderanger: https://github.com/kubernetes/kubernetes/pull/99185 filed.

sttts:
```
// Update implements EventHandler
func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
	var oldReq reconcile.Request
	if evt.MetaOld != nil {
		oldReq = reconcile.Request{NamespacedName: types.NamespacedName{
			Name:      evt.MetaOld.GetName(),
			Namespace: evt.MetaOld.GetNamespace(),
		}}
		q.Add(oldReq)
	} else {
		enqueueLog.Error(nil, "UpdateEvent received with no old metadata", "event", evt)
	}

	if evt.MetaNew != nil {
		newReq := reconcile.Request{NamespacedName: types.NamespacedName{
			Name:      evt.MetaNew.GetName(),
			Namespace: evt.MetaNew.GetNamespace(),
		}}
		if oldReq != newReq {
			q.Add(newReq)
		}
	} else {
		enqueueLog.Error(nil, "UpdateEvent received with no new metadata", "event", evt)
	}
}
```

alvaroaleman: @sttts the upstream change isn't really wanted because of this handler, as you noted, fixing it here is trivial and won't change anything about the number of allocations we do.

It is about something like this https://github.com/kubernetes-sigs/controller-runtime/blob/2faebdb710c1626efaa1e4df7cfb05c32ce1bdf5/pkg/handler/enqueue_mapped.go#L61-L62

There, we now suddenly need to allocate a new set for every update event to build deduplication in front of the already existing set(s) that is the workqueue. Having to build an ephemeral second set because the first set doesn't allow to atomically insert multiple entries seems needlessly inefficient.

pkg/handler/enqueue.go
vincepri: Should we implement something for https://github.com/kubernetes-sigs/controller-runtime/issues/857 that provides a useful feature for folks and we can put behind a flag for the time being, instead of changing critical code paths?

alvaroaleman: #857 is orthogonal I think, this here is about deduplicating events from a single event, #857 is about ratelimiting reconciliations, if I understand it correctly

vincepri: If I understood 857 correctly though, it's about ratelimiting events for the same resource in a set time window. It seems here we're kind of doing the same, where the time window is small?

alvaroaleman: The difference is that 857 is across multiple events (which is something we can absolutely not do by default) and this is only deduplicating requests from the same event, which is something we can do by default.

mars1024:
Hi @alvaroaleman , so we abandon this and switch to #1390 just because of introducing a new type?

mars1024:
How about using `namespace+name` in Object handler and using `request set` in Map and Owner handlers?

coderanger:
@mars1024 I don't think the issue is the new type per se, it's just both overkill for the place its used here and isn't set up for the other handlers where more complex logic is needed.

mars1024:
> @mars1024 I don't think the issue is the new type per se, it's just both overkill for the place its used here and isn't set up for the other handlers where more complex logic is needed.

So maybe I should use the new type in Map and Owner handlers to see whether it is overkill? Till now, I still think the inline logic implemented in #1390  is similar to an ordered set of requests.

mars1024:
> > @mars1024 I don't think the issue is the new type per se, it's just both overkill for the place its used here and isn't set up for the other handlers where more complex logic is needed.
> 
> So maybe I should use the new type in Map and Owner handlers to see whether it is overkill? Till now, I still think the inline logic implemented in #1390 is similar to an ordered set of requests.

Friendly ping @alvaroaleman @coderanger , I've updated this PR in the way I mentioned above, and do you still think `request set` is overkill? IMO, it seems more like a generic utility package, which will make codes brief and clean.

coderanger:
Yes, it is still overkill. There is no need for a mutex since this object isn't shared across goroutines.

coderanger:
I'm really not trying to invalidate your contributions here, you found a good thing to fix and I thank you for it, but we don't need something this complicated and I think the simpler version will be fine, I just need to finish a few final tweaks once $DAYJOB crunch time is finished.

PR #1381, opened 2021-02-10
🐛 [relase-0.8] Fix a race condition between leader election and recorder

vincepri:
This change introduces better syncronization between the leader election
code and the event recorder. Running tests with -race flag, we often saw
a panic on a closed channel, the channel was the one that the event
recorder was using internally.

After digging more through the code, it seems that we weren't properly
waiting for leader election code to stop completely, but instead we were
only calling the cancel() function asking the leader election to stop.

With this change, during a shutdown, we now wait for leader election to
finish up any internal task before we return and close an internal
channel. Only after leader election signals that the channel has been
closed, and  Run(...) has properly returned, we return execution to the
stop procedure, where the event recorder is then stopped.

Backport for #1379 for the release-0.8 branch

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -117,6 +117,10 @@ type controllerManager struct {
 	// it must be deferred until after gracefulShutdown is done.
 	leaderElectionCancel context.CancelFunc
 
+	// leaderElectionStopped is an internal channel used to signal the stopping procedure that the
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
/lgtm

alvaroaleman:
We have no tests on this branch :/

vincepri:
🤔 Seems we need to fix that in test infra

PR #1390, opened 2021-02-21
✨ Deduplicate events before sending them into the workqueue.

coderanger:
This avoids race conditions where extra reconciles can happen rarely.

I still think https://github.com/kubernetes/kubernetes/pull/99185 is a better solution but failing that, here is a "userspace" solution.

### DIFF
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -73,7 +73,11 @@ func (e *enqueueRequestsFromMapFunc) Generic(evt event.GenericEvent, q workqueue
 }
 
 func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
+	reqs := map[reconcile.Request]bool{}
 	for _, req := range e.toRequests(object) {
+		reqs[req] = true
+	}
+	for req := range reqs {
 		q.Add(req)
 	}
 }
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -73,12 +73,13 @@ func (e *enqueueRequestsFromMapFunc) Generic(evt event.GenericEvent, q workqueue
 }
 
 func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
-	reqs := map[reconcile.Request]bool{}
+	reqs := map[reconcile.Request]empty{}
 	for _, req := range e.toRequests(object) {
-		reqs[req] = true
-	}
-	for req := range reqs {
-		q.Add(req)
+		_, ok := reqs[req]
+		if !ok {
+			q.Add(req)
+			reqs[req] = empty{}
+		}
 	}
 }
 
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -53,27 +53,30 @@ type enqueueRequestsFromMapFunc struct {
 
 // Create implements EventHandler
 func (e *enqueueRequestsFromMapFunc) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
-	e.mapAndEnqueue(q, evt.Object)
+	reqs := map[reconcile.Request]empty{}
+	e.mapAndEnqueue(q, evt.Object, reqs)
 }
 
 // Update implements EventHandler
 func (e *enqueueRequestsFromMapFunc) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
-	e.mapAndEnqueue(q, evt.ObjectOld)
-	e.mapAndEnqueue(q, evt.ObjectNew)
+	reqs := map[reconcile.Request]empty{}
+	e.mapAndEnqueue(q, evt.ObjectOld, reqs)
+	e.mapAndEnqueue(q, evt.ObjectNew, reqs)
 }
 
 // Delete implements EventHandler
 func (e *enqueueRequestsFromMapFunc) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
-	e.mapAndEnqueue(q, evt.Object)
+	reqs := map[reconcile.Request]empty{}
+	e.mapAndEnqueue(q, evt.Object, reqs)
 }
 
 // Generic implements EventHandler
 func (e *enqueueRequestsFromMapFunc) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
-	e.mapAndEnqueue(q, evt.Object)
+	reqs := map[reconcile.Request]empty{}
+	e.mapAndEnqueue(q, evt.Object, reqs)
 }
 
-func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
-	reqs := map[reconcile.Request]empty{}
+func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object, reqs map[reconcile.Request]empty) {
 	for _, req := range e.toRequests(object) {
 		_, ok := reqs[req]
 		if !ok {
--- a/pkg/handler/enqueue_owner.go
+++ b/pkg/handler/enqueue_owner.go
@@ -59,31 +59,37 @@ type EnqueueRequestForOwner struct {
 
 // Create implements EventHandler
 func (e *EnqueueRequestForOwner) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
-	for _, req := range e.getOwnerReconcileRequest(evt.Object) {
+	reqs := map[reconcile.Request]bool{}
+	e.getOwnerReconcileRequest(evt.Object, reqs)
+	for req := range reqs {
 		q.Add(req)
 	}
 }
 
 // Update implements EventHandler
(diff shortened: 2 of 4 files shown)


coderanger:
Replaces #1380 but covers all three handler types instead.

coderanger:
If we got with this approach, should probably rewrite this to use k8s.io/apimachinery/pkg/util/sets.

pkg/handler/enqueue_mapped.go
```diff
@@ -73,7 +73,11 @@ func (e *enqueueRequestsFromMapFunc) Generic(evt event.GenericEvent, q workqueue
 }
 
 func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
+	reqs := map[reconcile.Request]bool{}
```
mars1024, line 76: I don't think a simple map is enough here, what we need is **an ordered set**, just like the `RequestSet` implementation in #1380 .

coderanger: I think that's a question of API promises. The return value from the map function is an array but do we consider that a promise that things will end up in the workqueue in that order? It's non-deterministic how things will actually run across multiple workers, though overall ordering would be respected. I wouldn't think that reconciles within the same controller (i.e. of the same root object type) would be order sensitive in general though.

mars1024: Got your point, but if the user-defined map function will generate an order-sensitive array, should we shuffle the order with a map even for the deduplication?

coderanger: I defer to the wisdom of the group :) I wouldn't consider order to be preserved there since it's not really in line with the convergence-based aesthetic of controllers. But we can totally do it if we want to spend the CPU cycles on slightly more tracking :D

coderanger:
```
Will keep running them until they fail.
This was attempt #107
No, seriously... you can probably stop now.
```

Should be good on ordering. Still leaving owners as unordered since the underlying order is meaningless and can be shuffled.

pkg/handler/enqueue_owner.go
```diff
@@ -59,31 +59,37 @@ type EnqueueRequestForOwner struct {
...
+	e.getOwnerReconcileRequest(evt.ObjectNew, reqs)
+	for req := range reqs {
 		q.Add(req)
 	}
 }
 
 // Delete implements EventHandler
 func (e *EnqueueRequestForOwner) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
-	for _, req := range e.getOwnerReconcileRequest(evt.Object) {
+	reqs := map[reconcile.Request]empty{}
```
alvaroaleman, line 81: Is this needed here and in Generic? I would have assumed that kube doesn't allow multiple identical ownerRefs

pkg/handler/eventhandler_test.go
```diff
@@ -281,20 +277,16 @@ var _ = Describe("Eventhandler", func() {
 				Expect(q.Len()).To(Equal(4))
 
 				i, _ := q.Get()
-				Expect(i).To(Equal(reconcile.Request{
-					NamespacedName: types.NamespacedName{Namespace: "foo", Name: "baz-bar"}}))
+				Expect(i).To(Equal(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: "foo", Name: "baz-bar"}}))
```
alvaroaleman, line 280: What about changing this tests to not give the new pod a new name (which can not happen anyways with a real apiserver) so it tests the deduplication?

pkg/handler/enqueue_owner.go
coderanger: Hmm, I don't actually know what happens if you try to give the same non-controller owner twice.

coderanger: I bet you could have duplicates, `+patchMergeKey=uid` so it will theoretically dedup on UID but you could have invalid owner refs with the wrong UID and duplicate names? I would also be 100% fine with not handling that cleanly and risking spurious reconciles since that should be very hard to make happen.

alvaroaleman: Yeah I would prefer to omit that here and in Generic

pkg/handler/eventhandler_test.go
alvaroaleman: Yeah, lets do that

pkg/handler/enqueue_mapped.go
alvaroaleman: So first off, there is no obvious rule if if we were to sort this the result of oldObj or newObj comes first. Second, deduplicating is already an microoptimization that should be irrelevant for most controllers (if it isn't, you should probably reconsider the architecture of your controller, you shouldn't be blocking for long-running things in `Reconcile`). So I am opposed to make ordering-when-deduplicating part of the api and adding additional code for that.

coderanger: I did actually change it to preserve the ordering on mapfuncs because it was easy :)

pkg/handler/enqueue_mapped.go
```diff
@@ -73,8 +73,13 @@ func (e *enqueueRequestsFromMapFunc) Generic(evt event.GenericEvent, q workqueue
 }
 
 func (e *enqueueRequestsFromMapFunc) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
+	reqs := map[reconcile.Request]empty{}
```
mars1024, line 76: I'm afraid this deduplication is not global enough when `mapAndEnqueue` is called several time in a single update, such as `enqueueRequestsFromMapFunc.Update`.

pkg/handler/enqueue_owner.go
```diff
@@ -111,17 +117,16 @@ func (e *EnqueueRequestForOwner) parseOwnerTypeGroupKind(scheme *runtime.Scheme)
 
 // getOwnerReconcileRequest looks at object and returns a slice of reconcile.Request to reconcile
 // owners of object that match e.OwnerType.
-func (e *EnqueueRequestForOwner) getOwnerReconcileRequest(object metav1.Object) []reconcile.Request {
```
mars1024, line 114: Since the function is changed wholly, should we add some comments for the new parameters?

pkg/handler/enqueue_mapped.go
coderanger: Fair, I should make it work more like _owner.

cbandy: If `EnqueueRequestForOwner` had a constructor, it could share by embedding `enqueueRequestsFromMapFunc` and set `toRequests = getOwnerReconcileRequest`. 😇

pkg/handler/enqueue_owner.go
cbandy: If not in Delete nor Generic, why have it in Create?

cbandy:
🦆 I tried to wrap my head around if/why we should dedupe in the single-object events (Create, Delete, Generic). It seems to me that if we do it in the Update event, we should do it in all events, right?

🦆 Should a user expect `EnqueueRequestsFromMapFunc` to dedupe the one slice their `MapFunc` returns? If we do it in Update, would it be unexpected for the other events to not?

🦆 Perhaps `MapFunc` is what it is and there should be a new EventHandler implementation that deals in unordered sets of `Request`. We could pass in a mutable set and they could call `set.Insert` themselves. That might avoid one slice and an iteration. (e.g. the Owner implementation here.)

🦆 If `workqueue` starts accepting slices, we would want to stick with slices... 😖 In that case, the two EventHandler implementations could (would?) have the same semantics.

❓ If `workqueue` starts accepting slices, would we definitely append the two `MapFunc` slices in Update and pass them to the queue at once?

---

🤔 after all that...

I think we should dedupe as little or as much as we like. Any amount is an improvement and approaches the ideal, where it happens directly in `workqueue`.

Perhaps we should change the documentation on any/all of these Handlers now to allow deduplication to change over time.

alvaroaleman:
> I tried to wrap my head around if/why we should dedupe in the single-object events (Create, Delete, Generic). It seems to me that if we do it in the Update event, we should do it in all events, right?

Update events contain two objects, old obj and new obj which is why there is a good chance this might happen. Adding de-duplication for anything else is essentially assuming that users provide buggy mappers. I am apposed to that.

> Should a user expect EnqueueRequestsFromMapFunc to dedupe the one slice their MapFunc returns? If we do it in Update, would it be unexpected for the other events to not?

No, see above.

> Perhaps MapFunc is what it is and there should be a new EventHandler implementation that deals in unordered sets of Request. We could pass in a mutable set and they could call set.Insert themselves. That might avoid one slice and an iteration. (e.g. the Owner implementation here.)

I am opposed to that. What this change is doing in the first place is a performance optimization that for most controllers shouldn't matter. I am strictly opposed to complicating our API and dev user experience even the slightest for this. Adding a redundant way to do the same thing would fall into the "complicating API" category. 

>  If workqueue starts accepting slices, we would want to stick with slices... ￼ In that case, the two EventHandler implementations could (would?) have the same semantics.

It won't, they rejected the upstream PR.

pkg/handler/enqueue_owner.go
coderanger: Looking at the code again, I think we should probably just leave it. It would complicate the flow to only do it sometimes, would have to switch back to returning a slice and then process it again, the extra memory allocations and code complexity seems not worth it.

coderanger:
@alvaroaleman I think this should be all set now?

alvaroaleman (approved):
/lgtm
/approve

Sorry, I didn't have time initially and then it fell off my radar :upside_down_face:

Issue #1191, opened 2020-09-30
envtest: no matches for kind "Infrastructure" in version "config.openshift.io/v1"


gee4vee:
Dependency versions:
`sigs.k8s.io/controller-runtime v0.6.3`
`github.com/openshift/api v0.0.0-20200326152221-912866ddb162`

I'm getting the error above in my test even though I've added the relevant schema like so:

```
...
	configv1 "github.com/openshift/api/config/v1"
...
var _ = BeforeSuite(func(done Done) {
...
	err = configv1.AddToScheme(scheme.Scheme)
	Expect(err).NotTo(HaveOccurred())
...
```

I have some other schemas I've added including the one for our CRD, and those don't give me problems. The controller does start and begins reconciling our CRD. I also checked the version of the OpenShift API that I'm using to make sure it has `Infrastructure` and it does: https://pkg.go.dev/github.com/openshift/api@v0.0.0-20200623075207-eb651a5bb0ad/config/v1

Am I missing something else? Full code is here: https://github.com/IBM/starter-kit-operator/blob/operator-sdk-upgrade/controllers/suite_test.go

ktravis:
I was also facing this issue without much indication what the problem was - I think I've just realized the problem, maybe someone can correct me if I have it wrong:

You are registering the object types with your `scheme.Scheme`, but there is nothing telling the (test) apiserver about the CRD itself. kubebuilder bootstraps a snippet that loads the CRDs you are defining and testing directly:

https://github.com/IBM/starter-kit-operator/blob/adffbc16c42d9d899432c67cd9c1e63a27729104/controllers/suite_test.go#L92-L94 

But there is nothing to define the rest of those types with the apiserver (registering Go types with your scheme does not install the associated CRD, it's only telling your local client how to handle that type of object).

For this to work the way you are expecting I think you would need to include the CRDs and add them to that list - otherwise the test apiserver will always return an error when you send a request for a type it hasn't heard of.

Hope that helps.

hfgbarrigas:
Any updates on this? Getting the same error. Using go mod vendor won't help because non go files are pruned.

Apart from crd paths, the environment has an attibute CRDs that looks promising but having issues using it since I can't find any documentation around it . Maybe @camilamacedo86 can shed some light?

hfgbarrigas:
For those running into this, here's how to properly add the external crd manifests to envtest:

```golang
CRDDirectoryPaths: []string{
[...]

PR #1393, opened 2021-02-23
📖 : add faq section on 3rd party APIs in envtest

varshaprasad96:
This commit adds a section on how to use third party APIs
in envtest.

Signed-off-by: varshaprasad96 <varshaprasad96@gmail.com>

Addresses - #1191

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -79,3 +79,36 @@ from the API groups that it needs (be they Kubernetes types or your own).
 See the [scheme builder
 docs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/scheme) for
 more information.
+
+### Q: When using envtest with a third party API, why do I see the error: `no matches for kind "Kind" in version "<group>/<version>"`?
+
+**A**: While testing a third party API with `envtest`, in addition to registering the API type with the test client, you would also
+have to register it with the test server. To do so, download the associated CustomResourceDefinition's manifest locally
+and add its path to [`Environment.CRDDirectoryPaths`](https://pkg.go.dev/github.com/kubernetes-sigs/controller-runtime/pkg/envtest#Environment).
+
+For example, to use `customapiv1` in `envtest`:
+- Register the API type with test client by adding it to the `scheme` used in the tests:
+
+```go
+...
+var _ = BeforeSuite(func (done Done)) {
+  ...
+  testScheme := scheme.Scheme
+  err := customapiv1.AddToScheme(testScheme)
+  Expect(err).NotTo(HaveOccurred())
+}
+```
+
+- Specify the CRD path to install it in the test environment:
+
+```go
+...
+var _ = BeforeSuite(func(done Done) {
+  ...
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{<location_of_the_CRD_for_`customapi`>},
+	}
+...
+```
+


coderanger:
Looks good, if we want to go even further maybe an example?

FAQ.md
```diff
@@ -79,3 +79,11 @@ from the API groups that it needs (be they Kubernetes types or your own).
 See the [scheme builder
 docs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/scheme) for
 more information.
+
+### Q: How to use third party APIs in `envtest`?
```
estroz, line 83: This could match the actual error better
```suggestion
### Q: When using envtest, why do I see the error: `no matches for kind "Kind" in version "<group>/<version>"`?
```

coderanger: That's the error for the CRD not being found, there's a different error for the scheme not being registered I think?

varshaprasad96: If its not added to scheme, it would be `no "<Kind>" is registered in scheme`, I think. That's explained in the previous question

FAQ.md
```diff
@@ -79,3 +79,11 @@ from the API groups that it needs (be they Kubernetes types or your own).
...
+
+**A**:  In order to use third party APIs in `envtest`, you would have to
+register the relevant custom object types with the `Scheme` being used in the tests. In addition to this,
+the CRD associated with the object also needs to be installed, so that the test API server
+knows how to handle the custom objects. To do so, you would have to specify the location of
+CRDs while [setting up the environment](https://pkg.go.dev/github.com/kubernetes-sigs/controller-runtime/pkg/envtest#Environment) by adding it to `CRDDirectoryPaths`. This will create the test environment with the specified CRDs installed. If this is not done, you may get `no matches for Kind <object_kind>` error from the API server.
```
estroz, line 89:
```suggestion
**A**:  testing a third party API with `envtest` requires you to do two things:
1. Register the API type with the test client by adding the relevant [custom `Scheme`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.2/pkg/envtest/envtest_test.go#L47-L55) to the client's scheme.
1. Register the API type with the server by downloading the associated CustomResourceDefinition's manifest locally then adding its path to [`Environment.CRDDirectoryPaths`](https://pkg.go.dev/github.com/kubernetes-sigs/controller-runtime/pkg/envtest#Environment).
```

FAQ.md
estroz: In that case, remove point (1) from my suggestion below.

estroz (approved):
/lgtm

FAQ.md
```diff
@@ -79,3 +79,10 @@ from the API groups that it needs (be they Kubernetes types or your own).
 See the [scheme builder
 docs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/scheme) for
 more information.
+
+### Q: When using envtest with a third party API, why do I see the error: `no matches for kind "Kind" in version "<group>/<version>"`?
+
+**A**: While testing a third party API with `envtest`, in addition to registering the API type with the test client, you would also
+have to register it with the test server. To do so, download the associated CustomResourceDefinition's manifest locally
+and add its path to [`Environment.CRDDirectoryPaths`](https://pkg.go.dev/github.com/kubernetes-sigs/controller-runtime/pkg/envtest#Environment).
```
camilamacedo86, line 87: The solution in the issue is to use the go module : `	filepath.Join(build.Default.GOPATH, "pkg", "mod", "github.com", "owner", "theirs@version", "config", "crd", "bases"),`

I do not think that we should tell fr the users add a CRD manually in the config/crds which is the place where it will be in kubebuilder/sdk because: 
- users will apply that with the make targets when should not
- sdk features to generate a bundle will add the crd as well. so, would it be really desirable?  

Also, could we add here an example of a valid config? wdyt?

estroz: >The solution in the issue is to use the go module

That path is meant to be opaque, i.e. shouldn't be recommended to users. Side note: this is no different than downloading a CRD locally using `curl https://my.domain/crd.yaml`.

FAQ.md
```diff
@@ -79,3 +79,36 @@ from the API groups that it needs (be they Kubernetes types or your own).
...
+
+For example, to use `customapiv1` in `envtest`:
```
varshaprasad96, line 89: Since this involves adding the api to scheme (which is explained in the previous question), and just specifying the path in `CRDDirectoryPaths`, is this example section required ? @coderanger @camilamacedo86

FAQ.md
camilamacedo86: Hi @estroz, 

I do not think we are addressing it properly yet. It is a very common question and I think it can be better covered.  

- IMO: It is missing an explanation where describe that EnvTest requires to have access to the CRD of the external-types schemas added and used. Then, we need to clarify that has a few options to do that. 
- Also, I think it is missing an example by point out from the go mod. 

Then, see that the kubebuilder scaffold is: 

```go
By("bootstrapping test environment")
	CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
	ErrorIfCRDPathMissing: true,
}
```

So, users will read this explanation and add the manifest manually in the config/crd dir as you suggested. So, I have a few questions: Could not it bring side effects for the project/tools helpers? In sdk, will not the `make bundle` use the CRD added there to generate the bundle? In KB/SDK, will not the makefile targets start to install the CRDs. Should we really suggest that? Have we other options? Should we make a small note here and describe that if you are using sdk/kb then, in order to avoid side effects you might prefer to use the go option or create a dir to copy all crds for there and add the external ones as well?

estroz: Sounds like you're asking for the envtest docs to be improved, not the FAQ response.

camilamacedo86 (changes requested):
It is a widespread scenario that brings a lot of confusion, and then, I believe that we need to try to cover it in a better way. 

Could EnvTest check/use the schema(s) on the cluster instead of expecting us to inform the CRDs indeed for the external types? Would not that possible? Should not it indeed that a better solution? If we reach the consensus that yes, then I think we can add only in the FAQ and clarify it is a workaround until the issue #nnnn be addressed. @vincepri @DirectXMan12 wdyt? 

Please, check the **Why** section below. 

**Suggested Solution (only if is mandatory inform the path / or until we have a better one)**

- Tell the user that currently, EnvTest requires to have access to all schemas definitions (CRDs) that will be used by it and then, because of this, the user needs to provide the CRD of the external type.
- Recommend first with the example of the go mod option, which is better than downloading the CRD imo:

```
CRDDirectoryPaths: []string{
			filepath.Join("..", "config", "crd", "bases"),
			filepath.Join(build.Default.GOPATH, "pkg", "mod", "github.com", "owner", "theirs@version", "config", "crd", "bases"),
},
```
- Then, describe that also is possible to add external type CRD schema(s) by adding the manifest on the project
- Recommend to the user **to create a new directory and not suggested it or lead them to believe that they should download it in the `config/crd/bases`.** PS.: IMO the best place in this scenario would be where the suite_test is (e.g controllers/testdata/external)  
- Tell to them add the downloaded CRD in the new directory. See that [CRDDirectoryPaths](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/server.go#L122) is `a list of paths` so, the user can inform the new directory as well:

```go
               CRDDirectoryPaths: []string{
			filepath.Join("..", "config", "crd", "bases"),
                        filepath.Join("testdata", "external"), 
		},
```

**Why**

The config/ has the project config manifests. So, I do not think that we should recommend or provide any info that leads to the idea that people should download and add the external CRDs there used only on tests for envtests. It can cause harmful side effects since the tools (sdk/kb) expected found in the `config/crd/bases` only the CRDs which are defined in the project e.g:

- after adding the CRD the users will run [make install](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L68-L69) and try to apply on cluster a schema that should exist already there 
- SDK will use all CRDs to generate the bundle and will add the external-type to it. @estroz do we want to have the external-type CRD defined in the [csv](https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/bundle/manifests/memcached-operator.clusterserviceversion.yaml#L23) and copied to [bundle/manifests/](https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/bundle/manifests/). Will OLM not try to install them on the cluster? Will that work if we have operator A which is the ower installing the CRD via the bundle, and then, operator B which is the consumer trying to install the same CRD again in the same cluster? I do not think so. 

PS.: @pwittrock's initial idea about allows exploding files from [config-gen](https://github.com/kubernetes-sigs/kubebuilder/pull/1831) would be only for that one which makes sense for users customize, ditto so it would not work with this recommendation and could lead a misunderstands. 

Note that we can provide helpers in SDK or KB that make life easier for the users; however, we cannot suggest a recommendation that hurts the project's config. Otherwise, the helpers will not work well as expected.

FAQ.md
camilamacedo86: I hope that I could clarify all my thoughts in ; https://github.com/kubernetes-sigs/controller-runtime/pull/1393#pullrequestreview-604042047. So, I am closing this one to avoid misunderstands and make it easier for others to follow up on this one.

mpreu:
(I am sorry if this is not the right place for my objection)

In the current discussions about this topic (operator-sdk, kubebuilder and here) there are basically two promoted solutions (at least as a workaround until better support): 

1. Downloading the external CRDs
2. Referring to a go module containing the external CRDs

1.) has potential downsides as already mentioned, while 2.) introduces the problem of having a statically defined path floating around, which needs to be updated manually if dependencies in the project change. 2.) can also lead to unexpected errors if multiple versions are available in the go module cache and someone forgets to update the tests to refer to the new version of external CRDs.

As external CRDs should usually be a dependency of the project anyway (as testdata or within the controller), could it be possible to promote using go modules in the regular way and dynamically add the CRDs to `envtest` using the `go.mod` file?

E.g. adding OCP CRDs programmatically could look (very) roughly like this:

```go
func getOCPCRDs() []string {
	const ocpModule = "github.com/openshift/api"
	modFilePath := filepath.Join("..", "..", "go.mod")

	goModBytes, err := ioutil.ReadFile(modFilePath)

	file, err := modfile.Parse(modFilePath, goModBytes, nil)

	ocpVersion := ""
	for _, r := range file.Require {
		if r.Mod.Path == ocpModule {
			ocpVersion = r.Mod.Version
		}
	}
	Expect(ocpVersion).ToNot(BeEmpty())

	basePath := filepath.Join(
		build.Default.GOPATH,
		"pkg",
		"mod",
		ocpModule+"@"+ocpVersion,
	)

	netNamespaceCRDPath := filepath.Join(
		basePath,
		"network",
		"v1",
		"003-netnamespace-crd.yaml",
	)

	networkConfigCRDPath := filepath.Join(
		basePath,
		"config",
		"v1",
		"0000_10_config-operator_01_network.crd.yaml",
	)

	return []string{
		netNamespaceCRDPath,
		networkConfigCRDPath,
	}
}
...
By("extracting relevant OpenShift API CRDs")
ocpCRDs := getOCPCRDs()

By("bootstrapping test environment")
    ErrorIfCRDPathMissing: true,
        CRDDirectoryPaths: []string{
	   filepath.Join("..", "..", "config", "crd", "bases"),
	},
}
testEnv.CRDDirectoryPaths = append(testEnv.CRDDirectoryPaths, ocpCRDs...)
```

This is obviously not at all perfect, but at least removes the need to update the static go module path in the tests with every little dependency update in the overall project.

PR #1395, opened 2021-02-24
⚠️ Change client.Patch to take client.Object for performance

cbandy:
Most uses of merge-patch will be for built-in types or kubebuilder-generated custom resources which have accessors for ResourceVersion. Using these with DeepCopyObject() is simple and fast.

```
name                            old time/op    new time/op    delta
MergeFrom/NoOptions               91.6µs ± 4%   107.3µs ±26%     ~     (p=0.417 n=7+10)
MergeFrom/NoOptions-2              112µs ±18%      88µs ±11%  -21.15%  (p=0.000 n=10+9)
MergeFrom/WithOptimisticLock       163µs ± 3%     121µs ± 3%  -25.66%  (p=0.000 n=10+8)
MergeFrom/WithOptimisticLock-2     137µs ± 4%     101µs ± 4%  -26.28%  (p=0.000 n=10+10)

name                            old alloc/op   new alloc/op   delta
MergeFrom/NoOptions               20.3kB ± 0%    20.3kB ± 0%     ~     (all equal)
MergeFrom/NoOptions-2             20.3kB ± 0%    20.3kB ± 0%     ~     (all equal)
MergeFrom/WithOptimisticLock      34.1kB ± 0%    26.7kB ± 0%  -21.89%  (p=0.000 n=10+10)
MergeFrom/WithOptimisticLock-2    34.2kB ± 0%    26.7kB ± 0%  -21.89%  (p=0.000 n=10+8)

name                            old allocs/op  new allocs/op  delta
MergeFrom/NoOptions                  359 ± 0%       359 ± 0%     ~     (all equal)
MergeFrom/NoOptions-2                359 ± 0%       359 ± 0%     ~     (all equal)
MergeFrom/WithOptimisticLock         579 ± 0%       390 ± 0%  -32.64%  (p=0.000 n=10+10)
MergeFrom/WithOptimisticLock-2       579 ± 0%       390 ± 0%  -32.64%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/client/patch.go
+++ b/pkg/client/patch.go
@@ -96,8 +96,42 @@ func (s *mergeFromPatch) Type() types.PatchType {
 	return types.MergePatchType
 }
 
+func (*mergeFromPatch) data(original, modified Object, opts MergeFromOptions) ([]byte, error) {
+	if opts.OptimisticLock {
+		version := original.GetResourceVersion()
+		if len(version) == 0 {
+			return nil, fmt.Errorf("cannot use OptimisticLock, object %q does not have any resource version we can use", original)
+		}
+
+		original = original.DeepCopyObject().(Object)
+		original.SetResourceVersion("")
+
+		modified = modified.DeepCopyObject().(Object)
+		modified.SetResourceVersion(version)
+	}
+
+	originalJSON, err := json.Marshal(original)
+	if err != nil {
+		return nil, err
+	}
+
+	modifiedJSON, err := json.Marshal(modified)
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonpatch.CreateMergePatch(originalJSON, modifiedJSON)
+}
+
 // Data implements Patch.
 func (s *mergeFromPatch) Data(obj runtime.Object) ([]byte, error) {
+	fromObject, fromOk := s.from.(Object)
+	objObject, objOk := obj.(Object)
+
+	if fromOk && objOk {
+		return s.data(fromObject, objObject, s.opts)
+	}
+
 	originalJSON, err := json.Marshal(s.from)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


cbandy:
I signed it

pkg/client/patch.go
```diff
@@ -96,8 +96,42 @@ func (s *mergeFromPatch) Type() types.PatchType {
...
+	modifiedJSON, err := json.Marshal(modified)
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonpatch.CreateMergePatch(originalJSON, modifiedJSON)
+}
+
 // Data implements Patch.
 func (s *mergeFromPatch) Data(obj runtime.Object) ([]byte, error) {
```
vincepri, line 127: Should we just move this to use `client.Object` instead? I don't think we necessarily want to support all runtime objects, because we still require ObjectMeta.

vincepri: cc @alvaroaleman

vincepri:
/milestone v0.9.x

If we switch the interface accepted in the method, we need to mark this as ⚠️

pkg/client/patch.go
alvaroaleman: Yeah, sounds like a good idea

cbandy:
If we're changing the signatures, perhaps we should add options vararg to MergeFrom like other patch constructors.

vincepri:
@cbandy Could we discuss that change in a separate issue?

cbandy:
> @cbandy Could we discuss that change in a separate issue?

Certainly. Is the signature (with runtime.Object) acceptable as-is in this PR?

vincepri:
We should move to `client.Object` and always use the fast path, there is no need to use the old runtime.Object

pkg/client/patch.go
cbandy: Done in aef159a.

vincepri:
LGTM please squash commits

alvaroaleman (approved):
/lgtm

PR #1399, opened 2021-02-26
⚠ Handle finalizers in the fake client

tjamet:
The fake client differs significantly from the real implementation.
With the real client implementation, upon the deletion of a resource,
when finalizers are set the DeletionTimestamp is filled with a non-null
value.
This triggers a number of reconciling loops for all controllers that then
can take the opportunity to inspect the resource being deleted, perform
the required cleaning actions and then update the resource by removing
their finalizers as the cleaning job have been done.

In the current implementation, the Delete function implements a straight
delete and hence triggers a false positive or negative test when trying
to run tests at the client level, without interacting with the objects.

For example:

```
client.CreateObject()
controller.Reconcile()
Expect(Something).To(BeDone())
client.DeleteObject()
controller.Reconcile()
Expect(SomethingElse).To(BeDone())
```

In the real life, SomethingElse would actually be done as the controller
would still have access to the resource. In the current fake implementation,
this is not the case as the controller does not have access to the
resource any longer.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -392,20 +392,7 @@ func (c *fakeClient) Delete(ctx context.Context, obj client.Object, opts ...clie
 	delOptions := client.DeleteOptions{}
 	delOptions.ApplyOptions(opts)
 
-	old, err := c.tracker.Get(gvr, accessor.GetNamespace(), accessor.GetName())
-	if err == nil {
-		oldAccessor, err := meta.Accessor(old)
-		if err == nil {
-			if len(oldAccessor.GetFinalizers()) > 0 {
-				now := metav1.Now()
-				oldAccessor.SetDeletionTimestamp(&now)
-				return c.tracker.Update(gvr, old, accessor.GetNamespace())
-			}
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


tjamet:
I think the PR is ready on its own. I didn't implement yet the feature in `DeleteAllOf` which would probably also be required

alvaroaleman (commented):
/ok-to-test

This is a breaking change and needs to be marked as such

tjamet:
> This is a breaking change and needs to be marked as such

Sure! shall I use ⚠ then?

alvaroaleman:
> Sure! shall I use ￼ then?

Yes, please.

> I think the PR is ready on its own. I didn't implement yet the feature in DeleteAllOf which would probably also be required

Can you implement that as well, please? Not having it there will surprise users.

tjamet:
> Can you implement that as well, please? Not having it there will surprise users

Sure! I will push it soon

alvaroaleman (approved):
/lgtm
/approve

thank you!

alvaroaleman:
/hold cancel
Labeler gets fixed in https://github.com/kubernetes/test-infra/pull/21119

PR #1400, opened 2021-02-27
:running: Builder: Remove unused restconfig attribute

alvaroaleman:
This probably was used back in the days when the config was a property
of the controller, but is definitely unused today. The builder can just
directly call `mgr.GetConfig`, should it need one.


vincepri (approved):
/approve
/lgtm
/milestone v0.9.x

PR #1401, opened 2021-02-27
:bug: Manager leader election: Don't reset restcfg UserAgent

alvaroaleman:
In pkg.LeaderElection.NewResourceLock we call rest.AddUserAgent
which resets the restcfgs useragent and sets it to the default one plus
a suffix. This resets whatever was originally set as UserAgent and since
we do not copy our restcfg before passing it in there, this leads to the
UserAgent being set to the leader-election one globally.


vincepri (approved):
/lgtm
/milestone v0.9.x

PR #1403, opened 2021-03-01
🐛 controllerutil.CreateOrPatch doesn't update not empty Status fields if Spec fields are specified

### LINKED ISSUE #1392: controllerutil.CreateOrPatch doesn't update not empty Status fields if Spec fields are specified
If we would like to patch custom resource with some `Spec` fields and `Status` which are not empty, the controllerutil.CreateOrPatch patches `Spec` only fields, but any `Status` keep unchanged,
`result` returns `OperationResultUpdatedStatus` which means that both `Spec` and `Status` where updated
The reason why this happens is that the original `Status` of the object is returned when the" Patch "procedure is exited and, as a result, no` Status` fields are updated

dmvolod:
If we would like to patch custom resource with some `Spec` fields and `Status` which are not empty, the controllerutil.CreateOrPatch patches `Spec` only fields, but any `Status` keep unchanged,
`result` returns `OperationResultUpdatedStatus` which means that both `Spec` and `Status` where updated
The reason why this happens is that the original `Status` of the object returned when the `Patch` procedure is exited and, as a result, no` Status` fields are updated

Fixes #1392

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -309,6 +309,20 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
 		// Only issue a Status Patch if the resource has a status and the beforeStatus
 		// and afterStatus copies differ
+		if result == OperationResultUpdated {
+			// If Status was replaced by Patch before, set it to afterStatus
+			objectAfterPatch, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+			if err != nil {
+				return result, err
+			}
+			if err = unstructured.SetNestedField(objectAfterPatch, afterStatus, "status"); err != nil {
+				return result, err
+			}
+			// If Status was replaced by Patch before, restore patched structure to the obj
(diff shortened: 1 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -309,6 +309,20 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
 		// Only issue a Status Patch if the resource has a status and the beforeStatus
 		// and afterStatus copies differ
+		if result == OperationResultUpdated {
+			// If Status was replaced by Patch before, set it to afterStatus
+			objectAfterPatch, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&obj)
```
cbandy, line 314: Should this be `ToUnstructured(obj)` (no address operator) like L282?

I see that L257 makes a deep copy while L282 does not. Is it possible that `Patch()` on L303 has manipulated `afterStatus` as well? I think a test that calls `CreateOrPatch()` with an `*unstructured.Unstructured` will confirm.

dmvolod: Probably should use `obj`, not its address, thanks @cbandy, fixed. But tests both variants, it seems to be working fine.

dmvolod:
@vincepri is there any chance to merge this PR?
Thanks in advance.

alvaroaleman (approved):
/approve

PR #1404, opened 2021-03-01
:sparkles: Add cache option to filter by field

qinqon:
All instance for a same resources are being cached by
controller-runtime, for some use cases this consumes a lot of memory and
CPU. This change add a option to the cache so resources can be filtered
by field.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -103,6 +104,13 @@ type Options struct {
 	// Namespace restricts the cache's ListWatch to the desired namespace
 	// Default watches all namespaces
 	Namespace string
+
+	// FieldSelectorByResource restricts the cache's ListWatch to the desired
+	// fields per resource, the map's value must implement Selector [1] using
+	// for example a Set [2]
+	// [1] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selector
+	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
+	FieldSelectorByResource map[schema.GroupResource]fields.Selector
 }
 
 var defaultResyncTime = 10 * time.Hour
@@ -113,8 +121,18 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace)
+	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace, opts.FieldSelectorByResource)
 	return &informerCache{InformersMap: im}, nil
+}
+
+// BuilderWithFieldSelectors returns a Cache constructor that will filter
+// contents using fieldSelectorByResource
+// WARNING: filtered out resources are not returned.
+func BuilderWithFieldSelectors(fieldSelectorByResource map[schema.GroupResource]fields.Selector) NewCacheFunc {
+	return func(config *rest.Config, opts Options) (Cache, error) {
+		opts.FieldSelectorByResource = fieldSelectorByResource
+		return New(config, opts)
+	}
 }
 
 func defaultOpts(config *rest.Config, opts Options) (Options, error) {
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -26,6 +26,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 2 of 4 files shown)


pkg/cache/internal/informers_map.go
```diff
@@ -76,15 +70,10 @@ type MapEntry struct {
 // specificInformersMap create and caches Informers for (runtime.Object, schema.GroupVersionKind) pairs.
 // It uses a standard parameter codec constructed based on the given generated Scheme.
 type specificInformersMap struct {
-	// Scheme maps runtime.Objects to GroupVersionKinds
-	Scheme *runtime.Scheme
```
vincepri, line 80: This change seems outside the scope of this PR?

pkg/cache/internal/informers_map.go
```diff
@@ -76,15 +70,10 @@ type MapEntry struct {
...
 
 	// config is used to talk to the apiserver
 	config *rest.Config
 
-	// mapper maps GroupVersionKinds to Resources
-	mapper meta.RESTMapper
-
```
vincepri, line 87: Same as above, could we keep these?

pkg/cache/internal/informers_map.go
```diff
@@ -97,11 +86,6 @@ type specificInformersMap struct {
 	// stop is the stop channel to stop informers
 	stop <-chan struct{}
 
-	// resync is the base frequency the informers are resynced
-	// a 10 percent jitter will be added to the resync period between informers
-	// so that all informers will not send list requests simultaneously.
-	resync time.Duration
-
```
vincepri, line 104: Same as above, could we keep these?

pkg/cache/internal/informers_map.go
```diff
@@ -117,9 +101,7 @@ type specificInformersMap struct {
 	// unstructured objects.
 	createListWatcher createListWatcherFunc
 
-	// namespace is the namespace that all ListWatches are restricted to
-	// default or empty string means all namespaces
-	namespace string
```
vincepri, line 122: Same as above, could we keep these?

pkg/cache/cache.go
```diff
@@ -103,6 +103,10 @@ type Options struct {
 	// Namespace restricts the cache's ListWatch to the desired namespace
 	// Default watches all namespaces
 	Namespace string
+
+	// FieldSelectorByResource restricts the cache's ListWatch to the desired
+	// fields per resource
+	FieldSelectorByResource map[string]string
```
vincepri, line 109: What's the key and value expected to look like here? Would it better to use a GroupKind as key instead?

vincepri: Or GroupResource?

pkg/cache/internal/options.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// Options are the optional arguments for creating a new InformersMap object
+type Options struct {
+	Scheme                  *runtime.Scheme
+	Mapper                  meta.RESTMapper
+	Resync                  *time.Duration
+	Namespace               string
+	FieldSelectorByResource map[string]string
+}
```
vincepri, line 17: Given that this is all internal, how are users supposed to use this new feature?

vincepri (commented):
Hi @qinqon, thanks for the contribution. Left a few comments, let's please reduce the PR to the absolute minimum set of changes needed for the requested feature addition. Also this doesn't seem to be a 🐛, but rather a new feature ✨

qinqon:
Remove the commit that encapsulate some internal attributes to an internal Options struct and create a small helper to find the fieldselector on the map using the resource from mapper, @vincepri you can take a look now

pkg/cache/internal/informers_map.go
qinqon: Restored

pkg/cache/internal/informers_map.go
qinqon: Restored

pkg/cache/internal/informers_map.go
qinqon: Done

pkg/cache/internal/informers_map.go
qinqon: Done

pkg/cache/cache.go
qinqon: Used GroupResource so we can match all the versions.

pkg/cache/internal/options.go
qinqon: Removed the internal Options

pkg/cache/cache.go
```diff
@@ -103,6 +103,10 @@ type Options struct {
 	// Namespace restricts the cache's ListWatch to the desired namespace
 	// Default watches all namespaces
 	Namespace string
+
+	// FieldSelectorByResource restricts the cache's ListWatch to the desired
+	// fields per resource
+	FieldSelectorByResource map[schema.GroupResource]string
```
alvaroaleman, line 109: Please add a sample field selector here and link to the kube docs about field selectors, many ppl are not aware that they only work on a very limited set of fields and not at all for CRDs

alvaroaleman:
@qinqon this doesn't close #244 as it doesn't allow filtering by labels (which I would assume to be more common), so please remove that from the PR body

pkg/cache/internal/informers_map.go
```diff
@@ -231,6 +235,15 @@ func (ip *specificInformersMap) addInformerToMap(gvk schema.GroupVersionKind, ob
 	return i, ip.started, nil
 }
 
+func (ip *specificInformersMap) findFieldSelectorByGVR(gvr schema.GroupVersionResource) string {
+	gr := schema.GroupResource{Group: gvr.Group, Resource: gvr.Resource}
+	fieldSelector, ok := ip.fieldSelectorByResource[gr]
```
alvaroaleman, line 240: Nit: This is an empty string if there isn't a match in the map, so you do not need the `ok` check

qinqon: Done

qinqon:
> @qinqon this doesn't close #244 as it doesn't allow filtering by labels (which I would assume to be more common), so please remove that from the PR body

I can try to add labels filtering at a follow up PR to keep this small.

pkg/cache/cache.go
qinqon: Done

alvaroaleman: its `medadata.name`, not `meta.name` or not?

alvaroaleman (commented):
Looks good except for the doc sample.

> I can try to add labels filtering at a follow up PR to keep this small.

If  you have the time for that, that would be great but I didn't mean to say that you are in any way obligated to that, just that it would be premature to close the issue

pkg/cache/cache.go
qinqon: Done, correct value is `metadata.name`.

qinqon:
@vincepri do you know if Is the issue at pull-controller-runtime-apidiff-master  important ?

qinqon:
[force-push](https://github.com/kubernetes-sigs/controller-runtime/compare/03473b4df63daf398deebb562c0cf2e156461c86..a4e89aedfd10f66d915f4f99d23acf4d0be1aa87): Rebase on master

alvaroaleman:
@qinqon The one thing this change ignores entirely is what impact this has for the cache-backed client. Could you write a small design doc in a different PR that:
* Outlines the motivation for this and possible alternatives
* What this means for the cache-backed client
* Also: Are there ways to defined a fieldSelector other than via a plain string?

Doesn't have to be anything big, but we asked for that multiple times in https://github.com/kubernetes-sigs/controller-runtime/issues/244#issuecomment-451317507 and declined other PRs because of that. I personally like the simplicity of just saying "We allow this on the cache and if you use a custom cache on your manager, you imply that you know better" _but_ its a bigger design question and I want to be sure there is agreement about that.

qinqon:
> @qinqon The one thing this change ignores entirely is what impact this has for the cache-backed client. Could you write a small design doc in a different PR

@alvaroaleman done a PR with the proposal https://github.com/kubernetes-sigs/controller-runtime/pull/1419

qinqon:
[force-push](https://github.com/kubernetes-sigs/controller-runtime/compare/a4e89aedfd10f66d915f4f99d23acf4d0be1aa87..b6217c63f49236118d0d10c4282da2f6f7260aed): Rebased and converting the map value to `fields.Selector`

Issue #1375, opened 2021-02-08
Support creating strategic merge patch


timebertt:
Currently, strategic merge patches can be sent using `client.RawPatch` like this:
```go
var deploy *appsv1.Deployment
deployCopy := deploy.DeepCopy()

// do changes
i, container := findContainerByName(deployCopy.Spec.Template.Spec.Containers, "nginx")
container.Image = "nginx:updated"
deployCopy.Spec.Template.Spec.Containers[i] = container

// calculate patch
oldDeployJSON, _ := json.Marshal(deployCopy)
deployJSON, _ := json.Marshal(deploy)

patchBytes, _ := strategicpatch.CreateTwoWayMergePatch(oldDeployJSON, deployJSON, &appsv1.Deployment{})

// send patch
return c.Patch(ctx, deploy, client.RawPatch(types.StrategicMergePatchType, patchBytes))
```

This will sent a patch looking like this, which will leverage the API's `mergeKey`:
```
{"spec":{"template":{"spec":{"$setElementOrder/containers":[{"name":"nginx"},{"name":"sidecar"}],"containers":[{"image":"nginx:some-version","name":"nginx"}]}}}}
```

It would be nice to support creating such strategic merge patch similar to `client.MergeFrom` (which will create a json merge patch), like this:
```go
// do changes
i, container := findContainerByName(deployCopy.Spec.Template.Spec.Containers, "nginx")
container.Image = "nginx:updated"
deployCopy.Spec.Template.Spec.Containers[i] = container

// send patch
return c.Patch(ctx, deploy, client.StrategicMergeFrom(deployCopy))
```

PR #1406, opened 2021-03-01
✨ Add client.StrategicMergeFrom

timebertt:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1375

This PR adds `client.StrategicMergeFrom`.
The function is similar to `client.{MergeFrom,MergeFromWithOptions}`, but it creates a `strategic-merge-patch` instead of a `merge-patch`. See doc strings for more details about the differences of the two patch types and to figure out when to use which one.

Note, that strategic merge patches are [not supported by CRDs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility).

### DIFF
--- a/pkg/client/patch.go
+++ b/pkg/client/patch.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
 var (
@@ -85,13 +86,15 @@ type MergeFromOptions struct {
 }
 
 type mergeFromPatch struct {
-	from Object
-	opts MergeFromOptions
+	patchType   types.PatchType
+	createPatch func(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
+	from        Object
+	opts        MergeFromOptions
 }
 
 // Type implements patch.
 func (s *mergeFromPatch) Type() types.PatchType {
-	return types.MergePatchType
+	return s.patchType
 }
 
 // Data implements Patch.
@@ -122,7 +125,7 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
 		return nil, err
 	}
 
-	data, err := jsonpatch.CreateMergePatch(originalJSON, modifiedJSON)
+	data, err := s.createPatch(originalJSON, modifiedJSON, obj)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


coderanger:
Is there any way to refactor this so it's not quite as much copy-paste from MergeFrom? It feels like a lot of duplication.

timebertt:
@coderanger Yes, you're right. It turned out to be very much the same like `MergeFrom`.
Deduplicated it now. PTAL.

pkg/client/patch.go
```diff
@@ -148,7 +164,21 @@ func MergeFromWithOptions(obj runtime.Object, opts ...MergeFromOption) Patch {
...
+	return &mergeFromPatch{patchType: types.MergePatchType, createPatch: createMergePatch, from: obj, opts: *options}
+}
+
+// StrategicMergeFrom creates a Patch that patches using the strategic-merge-patch strategy with the given object as base.
+func StrategicMergeFrom(obj runtime.Object) Patch {
+	return &mergeFromPatch{patchType: types.StrategicMergePatchType, createPatch: createStrategicMergePatch, from: obj}
+}
+
+// StrategicMergeFromWithOptions creates a Patch that patches using the strategic-merge-patch strategy with the given object as base.
+func StrategicMergeFromWithOptions(obj runtime.Object, opts ...MergeFromOption) Patch {
```
cbandy, line 176: In my opinion, there should be just one constructor `StrategicMergeFrom` and it should have the variadic `opts`.

cbandy, line 176: Is it appropriate to use `MergeFromOption` here? My concern is that `merge-patch` and `strategic-merge-patch` could diverge in the options they take in the future.

timebertt: I was contemplating about the same, glad you bring it up here.

The reason I chose this way, is that if options specific to either patch type will be added in the future, we could simply add another interface `StrategicMergeFromOption`, implement both interfaces in the options that we already have and just switch the parameter type in `StrategicMergeFrom`. This way, folks using the options wouldn't get any compile tim error when upgrading.
And as it's not required to differentiate right now, I chose to reuse `MergeFromOption` for simplicity's sake.

Let me know, if you have any concerns with this approach.

timebertt: Fair enough, done.
PTAL :)

pkg/client/patch.go
```diff
@@ -148,7 +164,16 @@ func MergeFromWithOptions(obj runtime.Object, opts ...MergeFromOption) Patch {
 	for _, opt := range opts {
 		opt.ApplyToMergeFrom(options)
 	}
-	return &mergeFromPatch{from: obj, opts: *options}
+	return &mergeFromPatch{patchType: types.MergePatchType, createPatch: createMergePatch, from: obj, opts: *options}
+}
+
+// StrategicMergeFrom creates a Patch that patches using the strategic-merge-patch strategy with the given object as base.
+func StrategicMergeFrom(obj runtime.Object, opts ...MergeFromOption) Patch {
```
alvaroaleman, line 171: Can you please add godocs here and to `MergeFrom` that explain how StrategicMergePatch and MergePatch differ and when to use which?

timebertt: Sure, good idea. Done.
WDYT now?

timebertt:
> @timebertt: PR needs rebase.

Done.

cbandy:
I see a failure now:

```
Patch
controller-runtime/pkg/client/client_test.go:3242
  StrategicMergeFrom
  controller-runtime/pkg/client/client_test.go:3307
    creates a strategic merge patch with the modifications applied during the mutation, using optimistic locking [It]
    controller-runtime/pkg/client/client_test.go:3357

    Unexpected error:
        <mergepatch.ErrPreconditionFailed>: {
            message: "precondition failed for: map[metadata:map[resourceVersion:10] spec:map[template:map[spec:map[$setElementOrder/containers:[map[name:main] map[name:sidecar]] containers:[map[image:foo:v2 name:main]]]]]]",
        }
        precondition failed for: map[metadata:map[resourceVersion:10] spec:map[template:map[spec:map[$setElementOrder/containers:[map[name:main] map[name:sidecar]] containers:[map[image:foo:v2 name:main]]]]]]
    occurred

    controller-runtime/pkg/client/client_test.go:3376
```

timebertt:
@cbandy yes, sorry about that.
The `preconditions` and the respective test case are not needed anymore since your changes in #1395.
I fixed it accordingly. PTAL.

alvaroaleman:
/lgtm
/approve

thank you!

PR #1409, opened 2021-03-03
:warning: Allow setting NewClientFunc w/o implementing an interface

alvaroaleman:
When we introduced the ClientBuilder we lost the ability to easily
overwrite the NewClient func, as interfaces can not be implemented
inline. This change reverts us back to using a NewClientFunc but with a
slightly changed signature so objects that should not be cached can
still be configured. In order to not provide two ways to do the same
thing, the ClientBuilder is removed


pkg/cluster/client_builder.go
```diff
@@ -39,16 +42,34 @@ func NewClientBuilder() ClientBuilder {
...
-	// Create the Client for Write operations.
+	if n.newClientFunc == nil {
+		n.newClientFunc = DefaultNewClientFunc
+	}
+
+	return n.newClientFunc(cache, config, options, n.uncached...)
+}
+
+// NewClientFunc is the signature of the func that is called to construct a new client.
+type NewClientFunc func(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error)
```
alvaroaleman, line 68: @vincepri on second thought at this point where we have to export NewClienFunc again (with a slightly different signature than what it had before) it might be easier to simply remove the ClientBuilder and go back to using a NewClientFunc directly. WDYT?

vincepri: Could we instead create a builder from a function?

alvaroaleman: > Could we instead create a builder from a function?

Can you elaborate, I don't think I understand what you mean by that?

vincepri (approved):
/approve
/lgtm

pkg/cluster/cluster.go
```diff
@@ -253,3 +251,20 @@ func setOptionsDefaults(options Options) Options {
 
 	return options
 }
+
+// NewClientFunc allows a user to define how to create a client
+type NewClientFunc func(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error)
+
+// DefaultNewClient creates the default caching client
+func DefaultNewClient(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
```
DirectXMan12, line 259: we might want to consider making this options or something in the future -- this is getting a bit unwieldy

PR #1410, opened 2021-03-04
🐛 Fix client.Apply and client.Merge to satisfy Patch

cbandy:
I missed this in #1395 and so did the tests. 😬 This changes the variable declarations of `client.Apply` and `client.Merge` to ensure they satisfy `client.Patch`. The compiler caught the other types because their constructors return `Patch`.


alvaroaleman (approved):
/lgtm
/approve

PR #1419, opened 2021-03-09
:seedling: Add a design document to filter cache by selectors

qinqon:
The controller-runtime is caching all the resource instance even if not
all of them end up being reconciled, this design document describe a
proposal to that problem by use selectors at the ListWatch informers
mechanism.

This is related to https://github.com/kubernetes-sigs/controller-runtime/pull/1404

### DIFF
--- a/designs/use-selectors-at-cache.md
+++ b/designs/use-selectors-at-cache.md
@@ -1,0 +1,125 @@
+Filter cache ListWatch using selectors
+===================
+## Motivation
+
+Controller-Runtime controllers use a cache to subscribe to events from
+Kubernetes objects and to read those objects more efficiently by avoiding
+to call out to the API. This cache is backed by Kubernetes informers.
+
+The only way to filter this cache is by namespace and resource type.
+In cases where a controller is only interested in a small subset of objects
+(for example all pods on a node), this might end up not being efficient enough.
+
+Requests to a client backed by a filtered cache for objects that do not match
+the filter will never return anything, so we need to make sure that we properly
+warn users to only use this when they are sure they know what they are doing.
+
+This proposal sidesteps the issue of "How to we plug this into the cache-backed
+client so that users get feedback when they request something that is
+not matching the caches filter" by only implementing the filter logic in the
+cache package. This allows advanced users to combine a filtered cache with the
+already existing `NewCacheFunc` option in the manager and cluster package,
+while simultaneously hiding it from newer users that might not be aware of the
+implications and the associated foot-shoot potential.
+
+The only alternative today to get a filtered cache with controller-runtime is
+to build it out-of tree. Because such a cache would mostly copy the existing
+cache and add just some options, this is not great for consumers.
+
+This proposal is related to the following issue [2]
+
+## Proposal
+
+Add a new selector code at `pkg/cache/internal/selector.go` with common structs
+and helpers
+```golang
+package internal
+
+...
+
+// SelectorsByObject associate a runtime.Object to a field/label selector
+type SelectorsByObject map[client.Object]Selector
+
+// SelectorsByGVK associate a GroupVersionResource to a field/label selector
+type SelectorsByGVK map[schema.GroupVersionKind]Selector
+
+// Selector specify the label/field selector to fill in ListOptions
+type Selector struct {
+	Label labels.Selector
+	Field fields.Selector
+}
+
+// ApplyToList fill in ListOptions LabelSelector and FieldSelector if needed
+func (s Selector) ApplyToList(listOpts *metav1.ListOptions) {
+...
+}
+
+
+Add a type alias to `pkg/cache/cache.go` to internal
+
+```golang
+
+type SelectorsByObject internal.SelectorsByObject
+
+```
+
+Extend `cache.Options` as follows:
+
+```golang
+type Options struct {
+	Scheme            *runtime.Scheme
+	Mapper            meta.RESTMapper
+	Resync            *time.Duration
+	Namespace         string
+	SelectorsByObject SelectorsByObject
+}
+```
+
+Add new builder function that will return a cache constructor using the passed
+cache.Options, users can set SelectorsByObject there to filter out cache, it
+will convert SelectorByObject to SelectorsByGVK
+
+
+```golang
+func BuilderWithOptions(options cache.Options) NewCacheFunc {
+...
+}
+```
+
+is passed to informer's ListWatch and add the filtering option:
+
+```golang
+
+# At pkg/cache/internal/informers_map.go
+
+ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
+            ip.selectors[gvk].ApplyToList(&opts)
+...
+```
+
+Here is a PR with the implementatin at the `pkg/cache` part [3]
+
+## Example
+
+User will override `NewCache` function to make clear that they know exactly the
+implications of using a different cache than the default one
+
+```golang
+ ctrl.Options.NewCache = cache.BuilderWithOptions(cache.Options{
+                            SelectorsByObject: cache.SelectorsByObject{
+                                    &corev1.Node{}: {
+                                        Field: fields.SelectorFromSet(fields.Set{"metadata.name": "node01"}),
+                                    }
+                                    &v1beta1.NodeNetworkState{}: {
+                                        Field: fields.SelectorFromSet(fields.Set{"metadata.name": "node01"}),
+                                        Label: labels.SelectorFromSet(labels.Set{"app": "kubernetes-nmstate})",
+                                    }
+                                }
+                            }
+                        )
+```
+
+
+[1] https://github.com/nmstate/kubernetes-nmstate/pull/687
+[2] https://github.com/kubernetes-sigs/controller-runtime/issues/244
+[3] https://github.com/kubernetes-sigs/controller-runtime/pull/1404


alvaroaleman:
@qinqon thanks for putting this up! Could you join the kubebuilder and controller-runtime biweekly meeting this Thursday 18 CET and show it briefly?

qinqon:
@alvaroaleman sure, I will be there.

qinqon:
@alvaroaleman do I have to add a bullet to the agenda just rise a hand on the meeting ?

alvaroaleman:
@qinqon a bullet would be great

qinqon:
/cc @alvaroaleman @vincepri 

Let me know if I have to pin someone else here, so we can discuss it.

designs/filter-cache-by-field.md
```diff
@@ -0,0 +1,89 @@
...
+deployed as a daemonset instead of a deployment since all the nodes with
+daemonset's pods will expect to receive notification of all the instances and
+that can end with scalation problems depending on the type or resource it's
+watching.
+
+The alternative that can be done with the current controller-runtme is
+to implement a custom cache that do the filtering since cache is pluggable
+into the controller-runtime [1]
+
+This proposal is related to the following issue [2]
```
alvaroaleman, line 22: What we need to mention is that with this proposal, we defer/ignore the problem of "The default cache client won't be able to give you any feedback if you request something through it that doesn't match the caches filter" by saying that if ppl replace the default new cache on the manager/cluster, they are essentially waving the warnings and saying they know better than the default.

designs/filter-cache-by-field.md
```diff
@@ -0,0 +1,89 @@
...
+
+Incrase `cache.Options` as follow:
+
+```golang
+type Options struct {
+	Scheme                  *runtime.Scheme
+	Mapper                  meta.RESTMapper
+	Resync                  *time.Duration
+	Namespace               string
+	FieldSelectorByResource map[schema.GroupResource]string
```
alvaroaleman, line 34: What about using https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selector instead of a string?

qinqon: Totally and add some example using https://github.com/kubernetes/apimachinery/blob/v0.20.4/pkg/fields/fields.go#L34 as an example.

designs/filter-cache-by-field.md
qinqon: Done

qinqon:
[force-push](https://github.com/kubernetes-sigs/controller-runtime/compare/69c3cbf680f56293ea032a9512066cc8c92fc3bd..e2bc32262621d2703b45c7c0c6ea834ef4e50be6): Removed pkg/cache changes from this PR, converting map value to `fields.Selector` and added warning to proposal.

qinqon:
[force-push](https://github.com/kubernetes-sigs/controller-runtime/compare/e2bc32262621d2703b45c7c0c6ea834ef4e50be6..95dbc434d1e239e486dc95d8ff79d9d834f7ec35): Added an additional proposal to pass the selector using the builder instead of manager.Options as implemented in https://github.com/kubernetes-sigs/controller-runtime/pull/1425

designs/filter-cache-by-field.md
alvaroaleman: Where?

alvaroaleman: We should also explicitly mention that the implication of that is that we are going to make this a bit hidden by forcing ppl to define a NewCache func to make use of this

qinqon: But the idae os the proposal is to use a new manager.Options field not overwritting the NewCache func, I added an example of overwritting it as an illustration of a workaround with the current controller-runtime option.

Or do you mean that we should just add the field at `cache.Options` level and force people to overwrite the NewCache func ?

alvaroaleman: > Or do you mean that we should just add the field at cache.Options level and force people to overwrite the NewCache func ?

Yeah, because otherwise ppl will use it and be surprised that the client which is also cache-backed gives them neither an error nor a result. If someone overwrites the NewCacheFunc they very explicitly say they know what they are doing. The discussion around this is the reason nothing ever happened here. We will also discuss this in the next meeting and hear other ppls opinions.

qinqon: ok, I will rewrite this with just one proposal of cache.Options + NewCacheFunc.

qinqon: btw are you sure the https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer return no result if you ask for stuff not retrieved at ListWatch ? I think it just go to apiserver, I am pretty possitive the PoC I did at kubernetes-nmstate was not failing in that case.

alvaroaleman: If the indexer doesn't have data because you filtered it out in your list opts, it will just give you a 404, yes.

qinqon:
[force-push](https://github.com/kubernetes-sigs/controller-runtime/compare/95dbc434d1e239e486dc95d8ff79d9d834f7ec35..abadf6544bbea98ea5107b1a7dc3357fd6f5a8e0): Removed proposal 2 and adapterd former proposal to use `NewCache` using a builder function that will return a constructor.

qinqon:
Does it make sense to include labels in this proposal, the interfaces to use them are very similar and we will be able to close the issue.

qinqon:
@alvaroaleman are you ok about adding also labels in this design document following the PoC from last comment ?

alvaroaleman:
> @alvaroaleman are you ok about adding also labels in this design document following the PoC from last comment ?

Sure

DirectXMan12:
Recording what we talked about in the meeting:

generally seems fine, esp since it sidesteps the longstanding issues we've had with figuring out interactions between controllers, autocaching, etc.

Please add a brief section about how this sidesteps the past issues (for posterity's sake)

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Example
+
+User will override `NewCache` function to make clear that they know exactly the
+implications of using a different cache than the default one
+
+```golang
+ ctrl.Options.NewCache = cache.BuilderWithSelectors(
+                            selector.SelectorsByGroupResource{
+							    {Group: "", Resource: "nodes"}: {
```
timebertt, line 103: I would probably go for `client.Object` instead of `schema.GroupResource` for configuring different caches (similar to how it is done in [`NewDelegatingClient`](https://github.com/kubernetes-sigs/controller-runtime/blob/eb0f1d4aa81794a38206965be25d118af5e18afc/pkg/client/split.go#L44).
It's more safe and consistent.
```suggestion
							    &corev1.Node: {
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,117 @@
...
+
+Here is a PR with the implementatin at the `pkg/cache` part [3]
+
+## Example
+
+User will override `NewCache` function to make clear that they know exactly the
+implications of using a different cache than the default one
+
+```golang
+ ctrl.Options.NewCache = cache.BuilderWithSelectors(
```
timebertt, line 101: General remark/ask: Would this proposal support setting up different caches for the same resource but for different controllers?
We have use cases for leveraging different optimized watches (e.g. with mutually exclusive label selectors / different namespaces) in different controllers. Would be nice, if we could build this right into this optimized cache.

timebertt (commented):
Like the proposal in general, one minor suggestion and one general question inline :)

designs/use-selectors-at-cache.md
qinqon: We have one cache per Manager, to do so you would need different manager (maybe running at different containers) or running multiple managers in the same container (I suppose this can be problematic).

Making a cache per controller is out of the scope of this design document.

designs/use-selectors-at-cache.md
qinqon: Make sense, I there is a `apiutil.GVKForObject(obj, in.Client.Scheme())` to compose the GVK from a client.Object.

timebertt: The map for lookup can be built up once before hand, just like in the `delegatingReader`:
https://github.com/kubernetes-sigs/controller-runtime/blob/eb0f1d4aa81794a38206965be25d118af5e18afc/pkg/client/split.go#L46-L52

Anyways, this can be discussed in detail during the implementation review.

qinqon:
Addressed @DirectXMan12 comments

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+The alternative that can be done with the current controller-runtme is
+to implement a custom cache that do the filtering since cache is pluggable
+into the controller-runtime [1]
+
+This proposal is related to the following issue [2]
+
+## Proposal
+
+Add a new selector code at `pkg/cache/selector/selector.go` with common structs
```
alvaroaleman, line 37: Why a new `selector` package rather than putting this into the existing `cache` package?

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// FindByGR use the GVR group and resource to find the field/label selector
+func (s SelectorsByGroupResource) FindByGR(gvr schema.GroupVersionResource) Selector {
+...
+}
+
+```
+
+Incrase `cache.Options` as follow:
```
alvaroaleman, line 65: Nit:

```suggestion
Extend `cache.Options` as follows:
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+Add new builder function that will return a cache constructor that will return
+a cache filtered by selectors
+
+```golang
+// BuilderWithSelectors returns a Cache constructor that will filter
+// contents using selectors
+// WARNING: filtered out resources are not returned.
+func BuilderWithSelectors(selectors selector.SelectorsByGroupResource) NewCacheFunc {
```
alvaroaleman, line 84: Why just selectors and not `Options`?

designs/use-selectors-at-cache.md
alvaroaleman: @timebertt you can already do "one cache per controller" today by using `source.KindWithCache`. It is not particularly ergonomic, but it works fine

designs/use-selectors-at-cache.md
alvaroaleman: I agree on the `client.Object` remark

designs/use-selectors-at-cache.md
qinqon: `pkg/cache/internal` is importing it, at golang you cannot import "parent" packages.

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
timebertt: Ah, right. Missed that option. 
Thanks for pointing it out 👍

designs/use-selectors-at-cache.md
alvaroaleman: So define it in pkg/cache/internal and alias it from pkg/cache?

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
...
+As side effect of this proposal the request that does not match the filtered
+will not be returned, so users has to be warned that if this filtering
+is used they know it and they are careful with the request they are doing.
+
+As a side note this proposal sidesteps some of the long standing issues with
+other [solutions](https://github.com/kubernetes-sigs/controller-runtime/issues/244#issuecomment-451336202)
+like problematic interactions between controllers or the need for a complex
+refactorings by adding the possibility for users to define the ListWatch
+selectors but with an API that will be used by users that know what are the
+implications.
```
alvaroaleman, line 27:
```suggestion
This proposal sidesteps the issue of "How to we plug this into the cache-backed client so that users get feedback when they request something that is not matching the caches filter" by only implementing the filter logic in the cache
package. This allows advanced users to combine a filtered cache with the already existing `NewCacheFunc` option in the manager and cluster package, while simultaneously hiding it from newer users that might not be aware of the implications and the associated foot-shoot potential.
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
...
+specific fields it will still cache all the instances on that type increasing
+the memory and CPU load. This also get worse in case controller-runtime is
+deployed as a daemonset instead of a deployment since all the nodes with
+daemonset's pods will expect to receive notification of all the instances and
+that can end with scalation problems depending on the type or resource it's
+watching.
+
+As side effect of this proposal the request that does not match the filtered
+will not be returned, so users has to be warned that if this filtering
+is used they know it and they are careful with the request they are doing.
```
alvaroaleman, line 20:
```suggestion
Requests to a client backed by a filtered cache for objects that do not match the filter will never return anything, so
we need to make sure that we properly warn users to only use this when they are sure they know what they are doing.
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
...
+===================
+## Motivation
+
+The controller-runtime subscribe to changes using a cache mechanism that get
+updated using kubernetes informers, those informers implement a `ListWatch`
+that specify the namespace and type specified at controller builder time.
+
+So the only possible filtering of that cache is namespace and resource type,
+this means that if some user is interested only on Reconcile some resource with
+specific fields it will still cache all the instances on that type increasing
```
alvaroaleman, line 11:
```suggestion
The only way to filter this cache is by namespace and resource type. In cases where a controller is only interested in a small subset of objects (for example all pods on a node), this might end up not being efficient enough.
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
...
+the memory and CPU load. This also get worse in case controller-runtime is
+deployed as a daemonset instead of a deployment since all the nodes with
```
alvaroaleman, line 13: I don't think how you deploy your controller is particularly relevant, the relevant aspect from your example is that your controller instance is only interested in a small subset of all objects that can not be filtered via namespace or type, right?

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
+Filter cache ListWatch using selectors
+===================
+## Motivation
+
+The controller-runtime subscribe to changes using a cache mechanism that get
+updated using kubernetes informers, those informers implement a `ListWatch`
+that specify the namespace and type specified at controller builder time.
```
alvaroaleman, line 7:
```suggestion
Controller-Runtime controllers use a cache to subscribe to events from Kubernetes objects and to read those objects more efficiently by avoiding to call out to the API. This cache is backed by Kubernetes informers.
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,123 @@
...
+As a side note this proposal sidesteps some of the long standing issues with
+other [solutions](https://github.com/kubernetes-sigs/controller-runtime/issues/244#issuecomment-451336202)
+like problematic interactions between controllers or the need for a complex
+refactorings by adding the possibility for users to define the ListWatch
+selectors but with an API that will be used by users that know what are the
+implications.
+
+The alternative that can be done with the current controller-runtme is
+to implement a custom cache that do the filtering since cache is pluggable
+into the controller-runtime [1]
```
alvaroaleman, line 31:
```suggestion
The only alternative today to get a filtered cache with controller-runtime is to build it out-of tree. Because such a cache would mostly copy the existing cache and add just some options, this is not great for consumers.
```

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Sure, removing it, your previous correction is good enough.

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Make sense, done.

alvaroaleman (approved):
/ok-to-test
/lgtm
/approve
/assign @DirectXMan12 @vincepri @estroz  

Anything to add from your side?

alvaroaleman:
/hold
for others

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,125 @@
...
+```golang
+package internal
+
+...
+
+// SelectorsByObject associate a client.Object to a field/label selector
+type SelectorsByObject map[client.Object]Selector
+
+// SelectorsByObject associate a GroupResource to a field/label selector
+type SelectorsByGVK map[schema.GVK]Selector
```
estroz, line 44: Nit
```suggestion
// SelectorsByObject associate a GroupVersionKind to a field/label selector
type SelectorsByGVK map[schema.GroupVersionKind]Selector
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,125 @@
...
+
+Add a new selector code at `pkg/cache/internal/selector.go` with common structs
+and helpers
+```golang
+package internal
+
+...
+
+// SelectorsByObject associate a client.Object to a field/label selector
+type SelectorsByObject map[client.Object]Selector
```
estroz, line 41: This should be restricted to `runtime.Object` so users don't get the idea that you can filter by object name/namespace
```suggestion
type SelectorsByObject map[runtime.Object]Selector
```

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,125 @@
...
+type SelectorsByGVK map[schema.GVK]Selector
+
+// Selector specify the label/field selector to fill in ListOptions
+type Selector struct {
+	Label labels.Selector
+	Field fields.Selector
+}
+
+// FillInListOpts fill in ListOptions LabelSelector and FieldSelector if needed
+func (s Selector) FillInListOpts(listOpts *metav1.ListOptions) {
```
estroz, line 53: How about
```suggestion
func (s Selector) ApplyToList(listOpts *client.ListOptions) {
```
so it implements `client.ListOption`

designs/use-selectors-at-cache.md
```diff
@@ -0,0 +1,125 @@
...
+
+Extend `cache.Options` as follows:
+
+```golang
+type Options struct {
+	Scheme            *runtime.Scheme
+	Mapper            meta.RESTMapper
+	Resync            *time.Duration
+	Namespace         string
+	SelectorsByObject SelectorsByObject
```
estroz, line 74: It would be nice to generalize this to something like 
```suggestion
	Selectors SelectorSet
```
Where `SelectorSet` is any set of selectors that implements
```go
type SelectorSet interface {
	client.ListOption
	Select()
}
```
where `Select()` is just an identifier method so users don't pass in other `client.ListOption`s.

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

designs/use-selectors-at-cache.md
qinqon: Done

qinqon: Do you mean ?

```golang
type SelectorSet interface {
	Select(runtime.Object) client.ListOption
}
```

qinqon: We have apply options per GVK so it doe not make sense to implement ListOption at the map level.

designs/use-selectors-at-cache.md
alvaroaleman: @estroz why? It is a map, the values can not access the keys. Making this `runtime.Object` means we have to deal with stuff like ppl putting List types in there

designs/use-selectors-at-cache.md
alvaroaleman: @estroz why? It is a map, the values can not access the keys. Making this `runtime.Object` means we have to deal with stuff like ppl putting List types in there

qinqon: @alvaroaleman don't we have the same problem with client.Object since it's embedding runtime.Object ?

estroz: @alvaroaleman it may be confusing to users to have all these object fields exposed by the `metav1.Object`, since they may assume the key can actually apply its own selector values, ex. by `SetName()`. Instead it would be better to constrain the contract to just `GroupVersionKind()`.

There can be some internal validation step to make sure the object is not a list type.

designs/use-selectors-at-cache.md
estroz: Sure, get rid of `client.ListOption` from that interface. Regardless, I may want to construct a GVK myself instead of using a typed object or `Unstructured`, so why not also expose `SelectorsByGVK`?

designs/use-selectors-at-cache.md
alvaroaleman: Using GVK is inconvenient. Using the smaller interface (runtime.Object) doesn't prevent ppl from putting in objects that fulfill a lager interface (client.Object). It does however keep them from putting something like List types in there and makes the compiler verify that rather than getting a runtime error. Also, I think the idea that ppl might think the map key has  any further relevance for filtering when the value is a `Selector` struct seems a bit far fetched to me.

We also have plenty of prior art where we use client.Object to only derive the type, e.G. in source.Kind or the clients delete method.

estroz: You could similarly argue that someone would not try to map a selector to a list type because that's not the type they add to `For()`, `Owns()`, or `Watches()`.

However since there is prior art I am fine with `client.Object`. There can always be another selector map type added later for `runtime.Object` without breakage if we see bugs filed because of confusion. It's not worth worrying about too much now.

qinqon: Going back to `client.Object` as proposed originally.

designs/use-selectors-at-cache.md
qinqon: As we have previously stated we are going to expose the map key as `client.Object`.

qinqon:
Before we merge this, I want to be sure that we don't want to have something like `client.ListOptions` as the value to open the door to filter the cache with all the options from there.

alvaroaleman:
> Before we merge this, can you confirm that we don't want to have something like client.ListOptions as the map value to allow users to filter with all the possible options not only selectors.

What other options exist that we could filter by? Only the namespace or not?

qinqon:
> > Before we merge this, can you confirm that we don't want to have something like client.ListOptions as the map value to allow users to filter with all the possible options not only selectors.
> 
> What other options exist that we could filter by? Only the namespace or not?

The other options are more dangerous than useful, so yes.

alvaroaleman:
> The other options are more dangerous than useful, so yes.

Yeah, the namespace is a bit redundant as its also a global option. Generally, we can easily extend the Selector in the future in a backwards-compatible way if we see a reason, so I am not too concerned about missing something there.

qinqon:
@estroz are you ok with the changes ?

estroz (approved):
/lgtm

alvaroaleman:
/hold cancel
Lets get this done!

qinqon:
/retest

I think we can override the job clearly is not related

PR #1425, opened 2021-03-12
:sparkles: Add Selector option to Builder

qinqon:
As a follow up of [1] this change instruct the ListWatcher from the
informers to select by label/field but instead of using a manager/cache
Options it pass the select using the builder.

[1] #1404


alvaroaleman (commented):
/hold

The implication of "We do ignore the effects of this change to the client' is IMHO that we should not add any option other than overwriting the NewCacheFunc to get access to this functionality. Otherwise ppl will use this, not get their objects that do not match the selector via the client anymore and create support issues.

Issue #1424, opened 2021-03-12
Graceful shutdown doesn't wait for reconcile workers to complete


RobEarl:
Following the release of https://github.com/kubernetes-sigs/controller-runtime/pull/967, I was hoping it would allow me to set a timeout big enough for any ongoing reconcile to complete. On testing, however, I discovered the reconcile still exits immediately.

I have tried injecting the stop channel into my reconciler then using a `sync.WaitGroup` to block until any ongoing reconcile completes:

```
var waitGroup sync.WaitGroup

func (r *MyReconciler) InjectStopChannel(stopChannel <-chan struct{}) error {
	logger := r.log.WithValues("Test", "Request.Name")

	go func() {
		<-stopChannel
		logger.Info("Shutdown signaled. Block until reconcile completes")
		// Stop signalled. Stall until any ongoing reconcile ends.
		waitGroup.Wait()
	}()

	return nil
}

func (r *MyReconciler) Reconcile(context context.Context, request reconcile.Request) (reconcile.Result, error) {
	waitGroup.Add(1)
	defer waitGroup.Done()

	reqLogger := r.log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
	reqLogger.Info("Reconciling", "type", r.instanceFactory.Name())
    
	for i := 1; i <= 30; i++ {
		reqLogger.Info("Waiting")
		time.Sleep(time.Second)
	}

	return reconcile.Result{}, nil
}
```
If I run this and hit CTRL+C in the middle of a reconcile, I do see the `waitGroup.Wait()` function log output but the process still exits immediately:

```
2021-03-12T10:39:57.628Z	INFO	controller-runtime.manager.controller.my	Starting Controller	{"reconciler group": "mygroup", "reconciler kind": "MyKind"}
2021-03-12T10:39:57.628Z	INFO	controller-runtime.manager.controller.my	Starting workers	{"reconciler group": "mygroup", "reconciler kind": "MyKind", "worker count": 1}
2021-03-12T10:39:57.628Z	INFO	controller_my	Reconciling	{"Request.Namespace": "test", "Request.Name": "test", "type": "MyKind"}
2021-03-12T10:39:57.628Z	INFO	controller_my	Waiting	{"Request.Namespace": "test", "Request.Name": "test"}
2021-03-12T10:39:58.629Z	INFO	controller_my	Waiting	{"Request.Namespace": "test", "Request.Name": "test"}
[...]

PR #1427, opened 2021-03-14
:bug: Controller: Wait for all reconciliations before shutting down

alvaroaleman:
Currently, the controller will instantly shutdown and return when its
context gets cancelled, leaving active reconciliations be. This change
makes it wait for those before finishing shutdown.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1424

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/go-logr/logr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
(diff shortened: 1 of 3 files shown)


pkg/internal/controller/controller.go
```diff
@@ -203,19 +203,17 @@ func (c *Controller) Start(ctx context.Context) error {
...
 
-		if c.JitterPeriod == 0 {
-			c.JitterPeriod = 1 * time.Second
-		}
-
 		// Launch workers to process resources
 		c.Log.Info("Starting workers", "worker count", c.MaxConcurrentReconciles)
+		wg.Add(c.MaxConcurrentReconciles)
 		for i := 0; i < c.MaxConcurrentReconciles; i++ {
-			go wait.UntilWithContext(ctx, func(ctx context.Context) {
```
alvaroaleman, line 213: From what I can see, the `wait.UntilWithContext` was harmless, but not useful because it would start the closure every c.JitterPeriod for as long as the ctx is not cancelled. The closure will however stay in the `for c.processNextWorkItem(ctx)` loop until the workqueue is shutdown and the workqueue is only shutdown when the `ctx` is cancelled.

I think we still have this because before b4a0212360dfc048233a01ee193de7b1d4dd1ab4 it was possible for the closure to end.

With the changes in this PR, using it may lead to a deadlock:
* We add the workers to the `WaitGroup`
* Context gets cancelled before the closure is started
* `wait.UntilWithContext` will not start the closure so nothing will ever call `wg.Done()`
* The `wg.Wait` is deadlocked

vincepri: This makes lots of sense, and it'd actually explain why I was still noticing some errors during tests that wouldn't properly clean up or reconcile everything before shutting stuff down (including the event recorder, which panics).

vincepri (approved):
/lgtm

PR #1428, opened 2021-03-14
:bug: Fix cache sync timeout functionality

alvaroaleman:
So far, timing out the cache sync in most realistic scenarios didn't work,
bceause source.Kind gets an already started cache from the Manager. A
cache that is already started will block forever on GetInformer which we
called in source.Kinds start and not its WaitForSync. The context passed
to Start however defines the entire lifecycle of the Source, not the
Start timeout.

This change makes source.Kind call GetInformer in Start but in a new
go routine and WaitForSync just wait for that to finish or for its
context to be cancelled. This preserves the existing semantic of
starting in Start but waint for ready in WaitForSync.

/cc @varshaprasad96 @estroz @vincepri 

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1346

### DIFF
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/cache/informertest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllertest"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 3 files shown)


pkg/internal/controller/controller_test.go
```diff
@@ -104,40 +106,21 @@ var _ = Describe("controller", func() {
 			close(done)
 		})
 
-		It("should wait for each informer to sync", func(done Done) {
```
alvaroaleman, line 107: The only thing this test ever tested was that `controller.Start` with a cancelled context doesn't error. With the changes in this PR however it will always error because the context passed from the Controller to WaitForSync is cancelled

pkg/source/source_test.go
```diff
@@ -90,6 +90,7 @@ var _ = Describe("Source", func() {
 					},
 				}, q)
 				Expect(err).NotTo(HaveOccurred())
+				Expect(instance.WaitForSync(context.Background())).NotTo(HaveOccurred())
```
alvaroaleman, line 93: We need to call WaitForSync here because before that finished, the Informer won't have the EventHandler attached. Not waiting makes the test racy.

alvaroaleman:
apidiff fails because:
```
  Incompatible changes:
  - Kind: old is comparable, new is not
```

I don't think Kind being comparable is something we consider part of the API, WDYT?

pkg/source/source.go
```diff
@@ -110,16 +115,29 @@ func (ks *Kind) Start(ctx context.Context, handler handler.EventHandler, queue w
...
-				"kind", kindMatchErr.GroupKind)
+	// cache.GetInformer will block until its context is cancelled if the cache was already started and it can not
+	// sync that informer (most commonly due to RBAC issues).
+	ctx, ks.startCancel = context.WithCancel(ctx)
+	ks.started = make(chan error)
+	go func() {
+		// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
+		i, err := ks.cache.GetInformer(ctx, ks.Type)
+		if err != nil {
+			if kindMatchErr, ok := err.(*meta.NoKindMatchError); ok {
```
estroz, line 126:
```suggestion
			kindMatchErr := &meta.NoKindMatchError{}
			if errors.As(err, &kindMatchErr) {
```

pkg/internal/controller/controller_test.go
estroz: In that case do you think it's worth having a test case that checks that an error is always returned if the context is cancelled?

alvaroaleman: IMHO the only thing someone who starts a Controller with a cancelled context can realistically expect is that it ends quickly, if it errors or not along the way ends up being an implementation detail. It is not going to result in anything useful.

alvaroaleman:
@estroz addressed your comment, ptal. I've also renamed the pull as the issue wasn't with the cache sync per se but with the cache sync timeout.

Just to mention that, another possible approach to this would have been to make the `GetInformer` call not wait for the given informer to be synced after the cache is started and only do that in Get/List, that would allow folks like ourselves to have one non-blocking and one explicitly blocking call. The reason I didn't do that is that it might break others that rely on GetInformer blocking. WDYT?

vincepri (approved):
/approve
/lgtm

vincepri:
/hold 

for @estroz

estroz:
>Just to mention that, another possible approach to this would have been to make the GetInformer call not wait for the given informer to be synced after the cache is started and only do that in Get/List, that would allow folks like ourselves to have one non-blocking and one explicitly blocking call.

@alvaroaleman that depends, can you think of any reason `GetInformer()` returning an unsync'd cache would be beneficial?

The current approach is fine with me.

/lgtm

PR #1429, opened 2021-03-16
✨ Simple helper for unmanaged webhook server

### LINKED ISSUE #1255: It should be a bit easier to setup webhook handlers without a manger
It's currently intended to be *possible* to set up webhook handlers w/o a manager (e.g. in your own HTTP mux, in a separate binary that doesn't need to have a REST config, etc), but it's not *easy* due to the dependency injection setup from the manager.

We should have a helper to make the dependencies more explicit, and to make this a bit easier.

/kind feature

kevindelgado:
This adds helper functions to setup webhooks server without a manager. It adds 
1. `webhook.Unmanaged` that creates a `webhook.Server` without a manager / cluster connection information
2. `admission.StandalonWebhook` that creates an `http.Handler` that can be run on any arbitrary Go HTTP server (`http.SeverMux`) rather than a `webhook.Server`)

Fixes #1255

### DIFF
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -113,6 +113,9 @@ func (f HandlerFunc) Handle(ctx context.Context, req Request) Response {
 }
 
 // Webhook represents each individual webhook.
+//
+// It must be registered with a webhook.Server or
+// populated by StandaloneWebhook to be ran on an arbitrary HTTP server.
 type Webhook struct {
 	// Handler actually processes an admission request returning whether it was allowed or denied,
 	// and potentially patches to apply to the handler.
@@ -221,8 +224,11 @@ type StandaloneOptions struct {
 	Path string
 }
 
-// StandaloneWebhook transforms a Webhook that needs to be registered
-// on a webhook.Server into one that can be ran on any arbitrary mux.
+// StandaloneWebhook prepares a webhook for use without a webhook.Server,
+// passing in the information normally populated by webhook.Server
+// and instrumenting the webhook with metrics.
+//
+// Use this to attach your webhook to an arbitrary HTTP server or mux.
 func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, error) {
 	if opts.Scheme == nil {
 		opts.Scheme = scheme.Scheme
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -22,13 +22,18 @@ import (
 	"net/http"
 
 	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	jsonpatch "gomodules.xyz/jsonpatch/v2"
 	admissionv1 "k8s.io/api/admission/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/client-go/kubernetes/scheme"
 
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics"
 )
 
 var (
@@ -202,4 +207,63 @@ func (w *Webhook) InjectFunc(f inject.Func) error {
 	}
 
 	return setFields(w.Handler)
+}
+
+// InstrumentedHook adds some instrumentation on top of the given webhook.
+func InstrumentedHook(path string, hookRaw http.Handler) http.Handler {
+	lbl := prometheus.Labels{"webhook": path}
+
+	lat := metrics.RequestLatency.MustCurryWith(lbl)
+	cnt := metrics.RequestTotal.MustCurryWith(lbl)
+	gge := metrics.RequestInFlight.With(lbl)
+
+	// Initialize the most likely HTTP status codes.
+	cnt.WithLabelValues("200")
+	cnt.WithLabelValues("500")
+
+	return promhttp.InstrumentHandlerDuration(
+		lat,
+		promhttp.InstrumentHandlerCounter(
+			cnt,
+			promhttp.InstrumentHandlerInFlight(gge, hookRaw),
+		),
+	)
+}
+
+// StandaloneOptions let you configure a StandaloneWebhook.
+type StandaloneOptions struct {
+	// Scheme is the scheme used to resolve runtime.Objects to GroupVersionKinds / Resources
+	// Defaults to the kubernetes/client-go scheme.Scheme, but it's almost always better
+	// idea to pass your own scheme in.  See the documentation in pkg/scheme for more information.
+	Scheme *runtime.Scheme
+	// Logger to be used by the webhook.
+	// If none is set, it defaults to log.Log global logger.
+	Logger logr.Logger
+	// Path the webhook will be served at.
+	// Used for labelling prometheus metrics.
+	Path string
+}
+
+// StandaloneWebhook transforms a Webhook that needs to be registered
+// on a webhook.Server into one that can be ran on any arbitrary mux.
+func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, error) {
+	if opts.Scheme == nil {
+		opts.Scheme = scheme.Scheme
+	}
+
+	var err error
+	hook.decoder, err = NewDecoder(opts.Scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.Logger == nil {
+		opts.Logger = logf.RuntimeLog.WithName("webhook")
+	}
+	hook.log = opts.Logger
+
+	if opts.Path == "" {
+		return hook, nil
+	}
+	return InstrumentedHook(opts.Path, hook), nil
 }
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -219,9 +219,10 @@ type StandaloneOptions struct {
 	// Logger to be used by the webhook.
 	// If none is set, it defaults to log.Log global logger.
 	Logger logr.Logger
-	// Path the webhook will be served at.
-	// Used for labelling prometheus metrics.
-	Path string
+	// MetricsPath is used for labelling prometheus metrics
+	// by the path is served on.
+	// If none is set, prometheus metrics will not be generated.
+	MetricsPath string
 }
 
 // StandaloneWebhook prepares a webhook for use without a webhook.Server,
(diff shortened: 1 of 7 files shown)


kevindelgado:
It’s worth noting that dependency injection is in the process of [being deprecated](https://github.com/kubernetes-sigs/controller-runtime/pull/1322). 

I plan to follow this up with an issue/attempt at doing this without dependency injection. This is more involved though (and probably touches a lot of how the webhook code is organized) and it doesn’t seem like we should prevent publishing a simple helper for unmanaged webhook servers until dependency injection is sorted out.

kevindelgado (commented):
Left some questions inline PTAL @DirectXMan12

pkg/webhook/admission/webhook.go
```diff
@@ -203,3 +208,62 @@ func (w *Webhook) InjectFunc(f inject.Func) error {
 
 	return setFields(w.Handler)
 }
+
+// InstrumentedHook adds some instrumentation on top of the given webhook.
+func InstrumentedHook(path string, hookRaw http.Handler) http.Handler {
```
kevindelgado, line 213: I moved `InstrumentedHook` from server.go here into `admission/webhook.go` because I'm using it in `StandaloneWebhook` but also because it seemed more appropriate because its wrapping instrumentation around the handler rather than the server stuff. Let me know if you think somewhere else is more appropriate.

pkg/webhook/webhook_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+				err = c.Create(context.TODO(), obj)
+				return errors.ReasonForError(err) == metav1.StatusReason("Always denied")
+			}, 1*time.Second).Should(BeTrue())
+
+			cancel()
+			close(done)
+		})
+	})
+	Context("when running a standalone webhook", func() {
+		It("should reject create request for webhook that rejects all requests", func(done Done) {
```
kevindelgado, line 113: This is the only test I have for `StandaloneWebhook`. I tried to pull the minimal setup I could out of `server.go#Start` to mimic what an arbitrary Go HTTP server could look like. Let me know if you had something else in mind for testing it.

kevindelgado (commented):
Added `StandaloneWebhook` helper for attach c-r webhooks to an existing arbitrary http.ServeMux (part two of https://github.com/kubernetes-sigs/controller-runtime/issues/1255#issuecomment-800618840). PTAL @DirectXMan12

rohitagarwal003:
Can you also add some docs on who to use this? It's hard to discover these without these being in the docs.

pkg/webhook/admission/webhook.go
DirectXMan12: this applies to all webhooks, not just admission ones, so it belongs in `pkg/webhook`.

DirectXMan12: or you could put it in a subpackage to avoid import loops, if necessary, like `pkg/webhook/internal/metrics` or something

pkg/webhook/admission/webhook.go
```diff
@@ -203,3 +208,62 @@ func (w *Webhook) InjectFunc(f inject.Func) error {
...
+	// Logger to be used by the webhook.
+	// If none is set, it defaults to log.Log global logger.
+	Logger logr.Logger
+	// Path the webhook will be served at.
+	// Used for labelling prometheus metrics.
+	Path string
+}
+
+// StandaloneWebhook transforms a Webhook that needs to be registered
+// on a webhook.Server into one that can be ran on any arbitrary mux.
```
DirectXMan12, line 248: This documentation is technically correct but slightly unclear.  I'd

1. Update the documentation on `Webhook` to mention that it needs to be registered with a `webhook.Server` or populated by `StandaloneWebhook`
2. maybe reword to something like
   
    ```go
// StandaloneWebhook prepares a webhook for use without a webhook.Server, passing in the information normally
// populated by webhook.Server and instrumenting the webhook with metrics.
// 
// Use this to attach your webhook to an arbitrary HTTP server or mux.
```

File: pkg/webhook/admission/webhook.go [hunk 4]
```diff
@@ -203,3 +208,62 @@ func (w *Webhook) InjectFunc(f inject.Func) error {
...
+
+	if opts.Logger == nil {
+		opts.Logger = logf.RuntimeLog.WithName("webhook")
+	}
+	hook.log = opts.Logger
+
+	if opts.Path == "" {
+		return hook, nil
+	}
+	return InstrumentedHook(opts.Path, hook), nil
```
DirectXMan12, line 268: Should this instrument since we don't actually know the path?  I almost feel like it should just be up to the caller to do this, or the caller's webhook server, but I'm open to opinions.

pkg/webhook/server.go
```diff
@@ -105,6 +105,67 @@ func (s *Server) setDefaults() {
...
+
+	// Scheme is the scheme used to resolve runtime.Objects to GroupVersionKinds / Resources
+	// Defaults to the kubernetes/client-go scheme.Scheme, but it's almost always better
+	// idea to pass your own scheme in.  See the documentation in pkg/scheme for more information.
+	Scheme *runtime.Scheme
+}
+
+// NewUnmanaged provides a webhook server that can be ran without
+// a controller manager.
+func NewUnmanaged(options Options) (*Server, error) {
```
DirectXMan12, line 144: nit: rename to `NewStandaloneServer` or something?

DirectXMan12: actually, since server is designed to be initialized from a struct, I think we could even switch this around and have a method called just `Standalone` that took an existing server instance.  That way we don't need an extra options struct, etc.

So something like

```go
func Standalone(server *Server) error {
  // ...
}

File: pkg/webhook/server.go [hunk 6]
```diff
@@ -105,6 +105,67 @@ func (s *Server) setDefaults() {
...
+	server := &Server{
+		Host:       options.Host,
+		Port:       options.Port,
+		CertDir:    options.CertDir,
+		CertName:   options.CertName,
+		KeyName:    options.KeyName,
+		WebhookMux: options.WebhookMux,
+	}
+	server.setDefaults()
```
DirectXMan12, line 153: this *must* be called under the `defaultingOnce`.  That aside, why are you calling this here?

pkg/webhook/server.go
DirectXMan12: or even

```go
func (s *Server) Standalone() error { /* ... */ }
```

pkg/webhook/server_test.go
```diff
@@ -174,6 +174,34 @@ var _ = Describe("Webhook Server", func() {
 			Expect(handler.injectedField).To(BeTrue())
 		})
 	})
+
+	Context("when using an unmanaged webhook server", func() {
```
DirectXMan12, line 178: nit: I'd just write this as `it should be able to serve in unmanaged mode` (no need for the context, IMO)

pkg/webhook/webhook_integration_test.go
```diff
@@ -0,0 +1,117 @@
...
+				err = c.Create(context.TODO(), obj)
+				return errors.ReasonForError(err) == metav1.StatusReason("Always denied")
+			}, 1*time.Second).Should(BeTrue())
+
+			cancel()
+			close(done)
+		})
+	})
+	Context("when running a webhook server without a manager ", func() {
+		It("should reject create request for webhook that rejects all requests", func(done Done) {
```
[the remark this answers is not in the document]
DirectXMan12: either `JustBeforeEach` and `BeforeEach`, or use a helper method.

DirectXMan12 (changes requested):
we should also probably have a couple godoc examples (see `example_test.go` files) and a bit more verbose docs on when you might want to use these.

kevindelgado (commented):
Addressed all feedback and added examples to `example_test.go`. Let me know if you think there's anywhere else I should document how/why to use this.

PTAL @DirectXMan12

stevekuznetsov:
@kevindelgado :wave:

pkg/webhook/webhook_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+				net.JoinHostPort(testenv.WebhookInstallOptions.LocalServingHost,
+					strconv.Itoa(int(testenv.WebhookInstallOptions.LocalServingPort))), cfg)
+			Expect(err).NotTo(HaveOccurred())
+
+			// create and register the standalone webhook
+			hook, err := admission.StandaloneWebhook(&webhook.Admission{Handler: &rejectingValidator{}}, admission.StandaloneOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			http.Handle("/failing", hook)
+
+			// run the http server
```
DirectXMan12, line 140: in the future, use `By("running the http server")` instead of just comments like this.  No need to fix this now though

DirectXMan12 (approved):
/lgtm
/approve

pkg/webhook/admission/webhook.go
```diff
@@ -203,3 +209,45 @@ func (w *Webhook) InjectFunc(f inject.Func) error {
...
+// and instrumenting the webhook with metrics.
+//
+// Use this to attach your webhook to an arbitrary HTTP server or mux.
+func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, error) {
+	if opts.Scheme == nil {
+		opts.Scheme = scheme.Scheme
+	}
+
+	var err error
+	hook.decoder, err = NewDecoder(opts.Scheme)
```
stevekuznetsov, line 239: @kevindelgado what actually injects this into the `hook.Handler`?

kevindelgado: 🤦 It doesn't... because I'm dumb

See https://github.com/kubernetes-sigs/controller-runtime/pull/1490

kevindelgado: fyi, fixed now @stevekuznetsov

pkg/webhook/example_test.go
```diff
@@ -70,3 +80,72 @@ func Example() {
...
+	hookServer := &Server{
+		Port: 8443,
+	}
+
+	// Register the webhooks in the server.
+	hookServer.Register("/mutating", mutatingHook)
+	hookServer.Register("/validating", validatingHook)
+
+	// Start the server without a manger
+	err := hookServer.StartStandalone(signals.SetupSignalHandler(), scheme.Scheme)
```
stevekuznetsov, line 97: @kevindelgado one more thing - maybe we could write an e2e for examples to help catch these, but it looks like the cert and key are not optional on this server? When using code similar to this, I get:

```
time="2021-04-23T20:34:54Z" level=error msg="Failed to serve admission webhooks." error="open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"
```

DirectXMan12: that's intentional -- same with the main webhook server.  More broadly, the main point of the webhook server is to help with managing TLS setup properly.  If you don't want TLS, use StandaloneWebhook.

We could certainly document that better, though.

kevindelgado: https://github.com/kubernetes-sigs/controller-runtime/pull/1501

PR #1430, opened 2021-03-16
:bug: Deflake should update CRDs if already present in the cluster

alvaroaleman:
Reproduces easily under load, e.G. by running `for s in $(seq 1 10); do go test ./pkg/envtest & done`. Doesn't reproduce for me with this patch.

/assign @vincepri


vincepri (approved):
/approve
/lgtm
/milestone v0.9.x

PR #1437, opened 2021-03-17
:bug: Pass webhook decoder through sub-dependencies when scheme is injected after injector

impl:
It is currently impossible to chain validating/defaulting webhooks using the multi-handlers because the decoder never gets properly passed down through the dependency chain if the scheme happens to be injected after the injection func. This change fixes that logic and adds helper methods to create standalone Handlers for Defaulters and Validators.

Without this change, we get a panic in the webhook when it receives a request. Here's an example of the kind of chaining we're trying to do: https://github.com/puppetlabs/pvpool/blob/12321f9da263ade52595493e1af60abb61c86050/pkg/webhook/checkout.go#L153-L163

I also have a cherry-pick for release-0.8 ready to go: https://github.com/kubernetes-sigs/controller-runtime/compare/release-0.8...puppetlabs:multi-webhook-sub-dependencies-release-0.8

### DIFF
--- a/pkg/webhook/admission/defaulter.go
+++ b/pkg/webhook/admission/defaulter.go
@@ -33,8 +33,13 @@ type Defaulter interface {
 // DefaultingWebhookFor creates a new Webhook for Defaulting the provided type.
 func DefaultingWebhookFor(defaulter Defaulter) *Webhook {
 	return &Webhook{
-		Handler: &mutatingHandler{defaulter: defaulter},
+		Handler: DefaultingHandlerFor(defaulter),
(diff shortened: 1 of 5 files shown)


impl:
Hi @DirectXMan12 and @pwittrock, I'd love to get your feedback on this PR if you're able to take a look! Happy to make any adjustments needed. Thanks so much!

alvaroaleman:
Fixes in https://github.com/kubernetes-sigs/controller-runtime/pull/1502
Thanks for your work!

PR #1439, opened 2021-03-18
✨ Make client.Patch implementations gomock-friendly

### LINKED ISSUE #1438: client.Patch implementations are no longer gomock-friendly
After kubernetes-sigs/controller-runtime#1406, the client.Patch implementations are no longer gomock-friendly.

So we have unit tests that use gomock. Let's have the following example to get a better understanding.

- `pkg/utils.go`

```go
package utils

import (
	"context"

)

func RemoveAnnotation(ctx context.Context, c client.Client, obj client.Object, annotation string) error {
	withAnnotation := obj.DeepCopyObject()

	annotations := obj.GetAnnotations()
	delete(annotations, annotation)
	obj.SetAnnotations(annotations)

	return c.Patch(ctx, obj, client.MergeFrom(withAnnotation))
}
```

- `pkg/utils_test.go`

```go
package utils_test

import (
	"context"
	"testing"

	"github.com/ialidzhikov/example/pkg/utils"
	
	"github.com/golang/mock/gomock"
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	corev1 "k8s.io/api/core/v1"
)

func TestUtils(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Utils Suite")
}

var _ = Describe("Utils", func() {

	var (
		ctrl *gomock.Controller
		c    *utils.MockClient
	)

	BeforeEach(func() {
		ctrl = gomock.NewController(GinkgoT())
		c = utils.NewMockClient(ctrl)
	})

	AfterEach(func() {
		ctrl.Finish()
	})

	Describe("#RemoveAnnotation", func() {
		It("should delete specific annotation", func() {
			obj := &corev1.Secret{
					Name:        "test-name",
					Namespace:   "test-namespace",
					Annotations: map[string]string{
						"foo": "bar",
						"bar": "baz",
					},
				},
			}
			objWithAnnotation := obj.DeepCopy()
			objWithoutAnnotation := &corev1.Secret{
					Name:        "test-name",
					Namespace:   "test-namespace",
					Annotations: map[string]string{
						"bar": "baz",
					},
				},
			}

			ctx := context.TODO()
			c.EXPECT().Patch(ctx, objWithoutAnnotation, client.MergeFrom(objWithAnnotation))

			err := utils.RemoveAnnotation(ctx, c, obj, "foo")
		})
	})
})
```

Note: `MockClient` is a mock generated with gomock

With sigs.k8s.io/controller-runtime < v0.8.3 this test is working perfectly fine.

```
$ go test -mod=vendor -timeout 30s -run ^TestUtils$ github.com/ialidzhikov/example/pkg/utils
ok  	github.com/ialidzhikov/example/pkg/utils	0.787s
```

kubernetes-sigs/controller-runtime#1406 (also cherry-picked into release-0.8 branch) introduces a field of type func (`createPatch `) in the `mergeFromPatch` struct. This breaks the naive gomock expectations. The reason is that the
`mergeFromPatch` struct has now a function attribute (`createPatch`), and as gomock is using `reflect.DeepEqual` under the hood, the comparison now fails:

> Func values are deeply equal if both are nil; otherwise they are not deeply equal.
> (from https://golang.org/pkg/reflect/#DeepEqual)

Example of such failure with ` sigs.k8s.io/controller-runtime@v0.8.3`:

```
$ go test -mod=vendor -timeout 30s -run ^TestUtils$ github.com/ialidzhikov/example/pkg/utils
Running Suite: Utils Suite
==========================
Random Seed: 1616082953
Will run 1 of 1 specs

• Failure [0.001 seconds]
Utils
/git/example/pkg/utils/utils_test.go:23
  #RemoveAnnotation
  /git/example/pkg/utils/utils_test.go:39
    should delete specific annotation [It]
    /git/example/pkg/utils/utils_test.go:40

    Unexpected call to *utils.MockClient.Patch([context.TODO &Secret{ObjectMeta:{test-name  test-namespace    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[bar:baz] [] []  []},Data:map[string][]byte{},Type:,StringData:map[string]string{},Immutable:nil,} 0xc000407a00]) at /git/example/pkg/utils/mocks.go:138 because:
    expected call at /git/example/pkg/utils/utils_test.go:63 doesn't match the argument at index 2.
    Got: &{application/merge-patch+json 0x1dbea60 0xc00042f680 {false}}
    Want: is equal to &{application/merge-patch+json 0x1dbea60 0xc00042f400 {false}}

    /git/example/pkg/utils/mocks.go:138
------------------------------


Summarizing 1 Failure:

[Fail] Utils #RemoveAnnotation [It] should delete specific annotation
/git/example/pkg/utils/mocks.go:138

Ran 1 of 1 Specs in 0.001 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestUtils (0.00s)
FAIL
FAIL	github.com/ialidzhikov/example/pkg/utils	0.671s
FAIL
```

ialidzhikov:
Fixes #1438

### DIFF
--- a/pkg/client/patch.go
+++ b/pkg/client/patch.go
@@ -84,11 +84,15 @@ type MergeFromOptions struct {
 	OptimisticLock bool
 }
 
+type patchHandler interface {
+	createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
+}
+
 type mergeFromPatch struct {
-	patchType   types.PatchType
-	createPatch func(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
-	from        Object
-	opts        MergeFromOptions
+	patchType    types.PatchType
+	patchHandler patchHandler
+	from         Object
+	opts         MergeFromOptions
 }
 
 // Type implements Patch.
@@ -124,7 +128,7 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
 		return nil, err
 	}
 
-	data, err := s.createPatch(originalJSON, modifiedJSON, obj)
+	data, err := s.patchHandler.createPatch(originalJSON, modifiedJSON, obj)
 	if err != nil {
 		return nil, err
 	}
@@ -132,11 +136,15 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
 	return data, nil
 }
 
-func createMergePatch(originalJSON, modifiedJSON []byte, _ interface{}) ([]byte, error) {
+type mergePatchHandler struct{}
+
+func (mergePatchHandler) createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
 	return jsonpatch.CreateMergePatch(originalJSON, modifiedJSON)
 }
 
-func createStrategicMergePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
+type strategicMergePatchHandler struct{}
+
+func (strategicMergePatchHandler) createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
 	return strategicpatch.CreateTwoWayMergePatch(originalJSON, modifiedJSON, dataStruct)
 }
 
@@ -148,7 +156,7 @@ func createStrategicMergePatch(originalJSON, modifiedJSON []byte, dataStruct int
 // See https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/ for more details on
 // the difference between merge-patch and strategic-merge-patch.
 func MergeFrom(obj Object) Patch {
-	return &mergeFromPatch{patchType: types.MergePatchType, createPatch: createMergePatch, from: obj}
+	return &mergeFromPatch{patchType: types.MergePatchType, patchHandler: mergePatchHandler{}, from: obj}
 }
 
 // MergeFromWithOptions creates a Patch that patches using the merge-patch strategy with the given object as base.
@@ -158,7 +166,7 @@ func MergeFromWithOptions(obj Object, opts ...MergeFromOption) Patch {
 	for _, opt := range opts {
 		opt.ApplyToMergeFrom(options)
 	}
-	return &mergeFromPatch{patchType: types.MergePatchType, createPatch: createMergePatch, from: obj, opts: *options}
+	return &mergeFromPatch{patchType: types.MergePatchType, patchHandler: mergePatchHandler{}, from: obj, opts: *options}
 }
 
 // StrategicMergeFrom creates a Patch that patches using the strategic-merge-patch strategy with the given object as base.
@@ -175,7 +183,7 @@ func StrategicMergeFrom(obj Object, opts ...MergeFromOption) Patch {
 	for _, opt := range opts {
 		opt.ApplyToMergeFrom(options)
 	}
-	return &mergeFromPatch{patchType: types.StrategicMergePatchType, createPatch: createStrategicMergePatch, from: obj, opts: *options}
+	return &mergeFromPatch{patchType: types.StrategicMergePatchType, patchHandler: strategicMergePatchHandler{}, from: obj, opts: *options}
 }
 
 // mergePatch uses a raw merge strategy to patch the object.
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -19,6 +19,7 @@ package client_test
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync/atomic"
 	"time"
 
@@ -3302,6 +3303,12 @@ var _ = Describe("Patch", func() {
 			By("returning a patch with data containing the annotation change and the resourceVersion change")
 			Expect(data).To(Equal([]byte(fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"},"resourceVersion":"%s"}}`, annotationKey, annotationValue, cm.ResourceVersion))))
 		})
+
+		It("deeply equals to the same patch", func() {
+			patch1 := client.MergeFrom(cm)
+			patch2 := client.MergeFrom(cm)
+			Expect(reflect.DeepEqual(patch1, patch2)).To(BeTrue())
+		})
 	})
 
 	Describe("StrategicMergeFrom", func() {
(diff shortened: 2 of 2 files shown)


pkg/client/patch.go
```diff
@@ -124,19 +128,23 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
...
 		return nil, err
 	}
 
 	return data, nil
 }
 
-func createMergePatch(originalJSON, modifiedJSON []byte, _ interface{}) ([]byte, error) {
+type mergePatchHandler struct{}
+
+func (h mergePatchHandler) CreatePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
```
rfranzke, line 141:
```suggestion
func (mergePatchHandler) CreatePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
```

pkg/client/patch.go
```diff
@@ -124,19 +128,23 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
...
 	return jsonpatch.CreateMergePatch(originalJSON, modifiedJSON)
 }
 
-func createStrategicMergePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
+type strategicMergePatchHandler struct{}
+
+func (h strategicMergePatchHandler) CreatePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
```
rfranzke, line 147:
```suggestion
func (strategicMergePatchHandler) CreatePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error) {
```

pkg/client/patch.go
```diff
@@ -84,11 +84,15 @@ type MergeFromOptions struct {
 	OptimisticLock bool
 }
 
+type patchHandler interface {
+	CreatePatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
```
rfranzke, line 88:
```suggestion
	createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
```

pkg/client/patch.go
```diff
@@ -84,11 +84,15 @@ type MergeFromOptions struct {
...
+	createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
+}
+
 type mergeFromPatch struct {
-	patchType   types.PatchType
-	createPatch func(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
-	from        Object
-	opts        MergeFromOptions
+	patchType    types.PatchType
+	patchHandler patchHandler
```
alvaroaleman, line 93: So the definition of "gomock friendly" is "can not use function types"?

alvaroaleman: Generally speaking, everything that is unexported is considered to be an implementation detail consumers can not rely on. This is for example why cmp will refuse to compare private types, see https://github.com/google/go-cmp third bullet point

pkg/client/patch.go
```diff
@@ -84,11 +84,15 @@ type MergeFromOptions struct {
 	OptimisticLock bool
 }
 
+type patchHandler interface {
+	createPatch(originalJSON, modifiedJSON []byte, dataStruct interface{}) ([]byte, error)
+}
+
 type mergeFromPatch struct {
```
timebertt, line 91: If you want to ensure, that this doesn't break in any upcoming release, can you add a test case like the one in https://github.com/kubernetes-sigs/controller-runtime/issues/1438?

pkg/client/patch.go
ialidzhikov: Sure, I agree that it is an implementation details.
On the other side this PR tries to improve the user experience for users that consume the controller-runtime client pkg - in more details it improves the readability of unit tests with controller-runtime client that use gomock (the most popular mocking library in the golang ecosystem).

This is how an expectation for aPatch call with gomock looks like with the current master

```go
	expectedPatch := client.MergeFrom(mergeFrom)
	expectedData, expectedErr := expectedPatch.Data(obj)
	Expect(expectedErr).To(BeNil())

	c.EXPECT().Patch(ctx, obj, gomock.Any()).DoAndReturn(func(_ context.Context, _ client.Object, patch client.Patch, _ ...client.PatchOption) error {
		data, err := patch.Data(obj)
		Expect(err).To(BeNil())
		Expect(patch.Type()).To(Equal(expectedPatch.Type()))
		Expect(data).To(Equal(expectedData))
		return nil
	})
```

This is how it would look like with this PR:

```go
	 c.EXPECT().Patch(ctx, obj, client.MergeFrom(mergeFrom))
```

alvaroaleman: How exactly do you assess in the second example that  the patch type is correct, that the data is correct and that it won't error if patch.Data is called?

ialidzhikov: Why I have to asses this when the given client.Patch arg equals to the expected one? In my unit test I would be only interested if the client.Client is called with the expected args (expected obj, expected patch, etc). The fact that Patch func of client.Client internally is using the Data func of the passed patch arg is not something that I should verify in my unit test.

alvaroaleman: Because you are looking at how the Patch looks, not at how it behaves even though you only care about the latter (and we export the latter and not the former). If we were to introduce a bug that made a patch always return no data, your tests would happily pass but your product would be broken.

ialidzhikov: Yeah, but I think this is the fundamental principle of the unit testing. When you write a unit test, you are only interested in the functional correctness of your unit (your type). The functional correctness of the external types you use should be ensured with (unit) tests in the corresponding library/pkg.
That's why we have mocks and spies that help us to isolate the unit under test (our type) from all dependant types (external ones). When you start testing also the dependent types, this is no longer a unit test.

alvaroaleman: Unit tests just means that you are testing one isolated unit (here: That Patch() is called with the correct value). You are testing for "correct value" by comparing two unexported struct instances. This is just invalid and has nothing to do with unit testing or not. The things you as a user of controller-runtime are allowed to make any assumptions about are exported.
Please feel free to bring this up in the next kubebuilder and controller-runtime meeting if you disagree.
/hold

PR #1441, opened 2021-03-22
🌱 Adding certwatcher as an external package

### LINKED ISSUE #1292: Export the certwatcher package
Consumers might want to react to externally-managed + rotated certificates, and the certwatcher package handles this, but it is an internal package. Could we consider exporting it?

christopherhein:
Per #1292 this exports the `certwatcher` package for use with other packages, for example custom `Runnable`'s passed into a manager.

Closes #1292

### DIFF
--- a/pkg/certwatcher/certwatcher_test.go
+++ b/pkg/certwatcher/certwatcher_test.go
@@ -1,0 +1,187 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package certwatcher_test
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net"
+	"os"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
+)
+
+var _ = Describe("CertWatcher", func() {
+	var _ = Describe("certwatcher New", func() {
+		It("should errors without cert/key", func() {
+			_, err := certwatcher.New("", "")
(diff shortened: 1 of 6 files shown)


pkg/certwatcher/certwatcher_test.go
```diff
@@ -0,0 +1,193 @@
...
+					if err != nil {
+						return err
+					}
+					continue
+				}
+
+				return nil
+			}).Should(Succeed())
+
+			watcher, _ = certwatcher.New(certPath, keyPath)
```
alvaroaleman, line 69: Why the unhandled error?

pkg/certwatcher/certwatcher_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+			ctxCancel()
+			Eventually(doneCh, "4s").Should(BeClosed())
+		})
+	})
+})
+
+func publicKey(priv interface{}) interface{} {
+	switch k := priv.(type) {
+	case *rsa.PrivateKey:
```
alvaroaleman, line 116: can we properly type the function signature if this effectively has to be a `*rsa.PrivateKey`?

pkg/certwatcher/certwatcher_test.go
christopherhein: no particular reason, I can add and `Expect` here as well.

pkg/certwatcher/certwatcher_test.go
christopherhein: Yeah, I'll be honest cert gen was a bit of copy/pasta where they had the ability to create multiple algo based certs ecdsa/rsa etc and I just trimmed this down. Given this is only one algo I can just hard code this type on line https://github.com/kubernetes-sigs/controller-runtime/pull/1441/files/efd29024698fb095a8c004dea2641925655fb4ad#diff-cc4e056d68e41389001be7de46dfac27f508da1039369db677584b5721321ee4R162

```
derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, publicKey(priv), priv)
```

christopherhein: Updated.

pkg/certwatcher/certwatcher_test.go
christopherhein: Updated.

vincepri (commented):
Changes LGTM, @christopherhein could we add a `doc.go` in this package to explain what this should/can be used for?

christopherhein:
> Changes LGTM, @christopherhein could we add a `doc.go` in this package to explain what `certwatcher` should/can be used for?

Definitely! :+1:

christopherhein:
@vincepri I did ya one a little better, added a `doc.go` and an `example_test.go` to make these docs even more thorough.

vincepri (approved):
/approve
/assign @alvaroaleman

alvaroaleman (approved):
/lgtm

PR #1446, opened 2021-03-23
Warn if logger is not explicitly set

justinsb:
Though we have a 30 second fallback, there is no indication that this
is happening or hint to users that they should set a logger.

This path being hit is not ideal: we may still buffer a substantial
amount of log messages on larger clusters, and we aren't getting any
log messages.  We should log something; even though the best we can do
is to just log an unformatted message to stderr.


coderanger:
I'm 99% sure the default skeleton sets a logger but should make sure this doesn't incorrectly activate during tests.

alvaroaleman:
> This path being hit is not ideal: we may still buffer a substantial
amount of log messages on larger clusters, and we aren't getting any
log messages. 

@justinsb You mean in the first 30 seconds?

justinsb:
The default generated skeleton does indeed set a logger, but user reactions to zap output are ... mixed :-)  The failure case I've observed is that users have simply commented out the zap logger, not realizing the implications.

The fix to stop buffering log messages in ram after 30 seconds is a good one, but even with this users may wonder why their RAM usage is higher in the first 30 seconds (and we would expect that for a typical controller a lot of work happens on startup as it syncs and re-reconciles everything).

I'm proposing here that we warn users that they should explicitly call SetLogger (even if they're going to pass a null logger, that's still better than just commenting it out).

pkg/log/log.go
```diff
@@ -64,6 +66,9 @@ func init() {
 		loggerWasSetLock.Lock()
 		defer loggerWasSetLock.Unlock()
 		if !loggerWasSet {
+			// Alert the user that this is not a good thing to rely on
+			fmt.Fprintln(os.Stderr, "warning: log.SetLogger was not called; this is not recommended and may be unsupported in a future version of controller-runtime")
```
alvaroaleman, line 70: Can you mention here that they should explicitly call `SetLogger(NullLogger{})`? I would assume ppl who don't want the c-r logging will ignore that message otherwise

coderanger: +1 for actionable output.

justinsb: Good idea - the more actionable the better!  Done!

pkg/log/log.go
```diff
@@ -64,6 +66,9 @@ func init() {
 		loggerWasSetLock.Lock()
 		defer loggerWasSetLock.Unlock()
 		if !loggerWasSet {
+			// Alert the user that this is not a good thing to rely on
+			fmt.Fprintln(os.Stderr, "warning: log.SetLogger was not called; this is not recommended and may be unsupported in a future version of controller-runtime.  Call log.SetLogger(log.NullLogger{}) if you want no log output.")
```
alvaroaleman, line 70: Ci failure is genuine, maybe add a lint exception?
```
pkg/log/log.go:70: line is 223 characters (lll)
```

Ref https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1446/pull-controller-runtime-test-master/1375495983600766976#1:build-log.txt%3A214

weilaaa: Hi, buddies. It seems something like log buffer is need, cause If the program is directly exiting within 30 seconds, we neither got stderr as line 70 showed nor other logs anyhow.

@alvaroaleman Hi, buddy, can I have a try for this?

weilaaa:
I submit a pr for this, I hope it's helpful.
https://github.com/kubernetes-sigs/controller-runtime/pull/1493

Warn if logger is not explicitly set in two situations:
logger was called before 30s without setLogger
setLogger was not called exceeds 30s

cndoit18:
Fixes #1475

PR #1451, opened 2021-03-24
:bug: Event recorder should not send events when broadcaster is stopped

vincepri:
This change solves a race condition when the manager is shutting down
and reconcilers are still in the process of sending events through the
recorder. If the shutdown timeout is hit too early, the event recorder
panics because we're still trying to send events through the recorder.

To avoid this issue, we introduce a RWMutex between the provider and
event recorder. When the broadcaster is shutdown, events are now
dropped.

/assign @alvaroaleman


alvaroaleman (approved):
/lgtm

PR #1460, opened 2021-03-28
:sparkles: Add ClientWithWatch for use in CLIs

alvaroaleman:
This change adds a new WithWatch interface and implementation to the
Client package. It is a superset of the existing Client interface that
additionally contains a Watch method. It is intended to be used by CLI
apps that need to wait for a condition to arise.

Because this is not intended to be used with controllers and might cause
confusion there, the existing interface and all references to it are
kept as-is.

The basic idea of adding this was discussed and agreed-on in the March 25th meeting.


vincepri (approved):
/lgtm

alvaroaleman:
Timeout creating an envtest env
/retest

Issue #702, opened 2019-11-29
Fakeclient requires types read through unstructured to be registered in the scheme


alvaroaleman:
Writing a test that uses the fakeClient for code that uses `*unstructured.UnstructuredList` to list something, the test fails with a

```
failed to list ImageRegistryConfigs: non-list type *unstructured.UnstructuredList (kind "my-api/v1, Kind=Config") passed as output
```

The same code works fine when using a non-fake client.
/kind bug

alvaroaleman:
Okay so actually it seems I should have just read the error message, because I didn't pass in the ListType which happens to work with the non-fake client.

However after fixing that, the next error is this:

```
deletion_test.go:155: Deletion failed: failed to cleanup TYPE: failed to list TYPE: no kind "TYPEList" is registered for version "APIVERSION" in scheme "k8s.io/client-go/kubernetes/scheme/register.go:65"
```

This issue seems to be rooted in the underlying object tracker used by the fake client. Not sure if and how to fix it. Maybe register `*unstructured.Unstructured` for the given GVK?

Not being able to use `*unstructured.Unstructered` with an unregistered GVK defeats the reason of using it in the first place.

DirectXMan12:
I think the objectracker needs to use the passed-in scheme, not the kubernetes scheme -- use `NewFakeClientWithScheme` instead of `NewFakeClient` (which totally violates our controller-runtime API conventions, but that's a whole different story...)

alvaroaleman:
> I think the objectracker needs to use the passed-in scheme, not the kubernetes scheme -- use NewFakeClientWithScheme instead of NewFakeClient (which totally violates our controller-runtime API conventions, but that's a whole different story...)

The problem is that it requires the type I am trying to get via an `unstructured.Unstructured.` to be registered. But the reason I am using `unstructured.Unstructured` in the first place is that I do not have the go types.

DirectXMan12:
oh, yeah, sorry, triaging too fast :-P

DirectXMan12:
need to dig into objecttracker here probably

/help

mittachaitu:
Is there any specific reason to add [this](https://github.com/kubernetes-sigs/controller-runtime/blob/f52b6180d515bca5f8faa551c3784fee5ce89a83/pkg/client/fake/client.go#L199) check?

alvaroaleman:
> Is there any specific reason to add this check?

[...]

PR #1467, opened 2021-04-03
:bug: Fakeclient: Do not require ListKind

alvaroaleman:
The fake client will currently error out of given a Kind in an
UnstructuredList that doesn't end with list. This is different from what
we do in the actual client, where we tolerate that:
https://github.com/kubernetes-sigs/controller-runtime/blob/10ae090c1d3ac0c560dfa1a29b2517eb8d74442b/pkg/client/unstructured_client.go#L204

This change brings the fake client in line with the real client.

Related: The comments on https://github.com/kubernetes-sigs/controller-runtime/issues/702 although the Bug itself is about something else.

/assign @vincepri

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -290,13 +290,11 @@ func (c *fakeClient) List(ctx context.Context, obj client.ObjectList, opts ...cl
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


vincepri (approved):
/lgtm

PR #1468, opened 2021-04-04
✨ Surface API Server warnings in client

### LINKED ISSUE #1150: Surface warnings sent by apiserver via client-go
We should look into surfacing warnings sent in responses by 1.19+ API servers.

Examples of these warnings include:
* Use of deprecated core APIs
* Use of deprecated CRD-defined APIs
* Warnings returned by admission webhooks

See https://kubernetes.io/blog/2020/09/03/warnings/ for details.

MadhavJivrajani:
- Add a `WarningHandler` that writes to `stderr`
    - Deduplicates warnings
    - Highlight warning output with color when the outut supports it

Fix #1150 

/hold
/cc joelanford
- Currently, by default warnings are surfaced, should an option be provided to turn off the warnings if needed as illustrated [here](https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling)?

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -31,7 +32,22 @@ import (
 	"k8s.io/client-go/rest"
 
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
+
+// WarningHandlerOptions are options for configuring a
+// warning handler for the client which is responsible
+// for surfacing API Server warnings.
+type WarningHandlerOptions struct {
+	// SuppressWarnings decides if the warnings from the
+	// API server are suppressed or surfaced in the client.
+	SuppressWarnings bool
+	// AllowDuplicateLogs does not deduplicate the to-be
+	// logged surfaced warnings messages. See
+	// log.WarningHandlerOptions for considerations
+	// regarding deuplication
+	AllowDuplicateLogs bool
+}
 
 // Options are creation options for a Client
 type Options struct {
@@ -40,6 +56,10 @@ type Options struct {
 
 	// Mapper, if provided, will be used to map GroupVersionKinds to Resources
 	Mapper meta.RESTMapper
+
+	// Opts is used to configure the warning handler responsible for
+	// surfacing and handling warnings messages sent by the API server.
+	Opts WarningHandlerOptions
 }
 
 // New returns a new Client using the provided config and Options.
@@ -56,6 +76,27 @@ func New(config *rest.Config, options Options) (Client, error) {
 	if config == nil {
 		return nil, fmt.Errorf("must provide non-nil rest.Config to client.New")
 	}
+	var logger logr.Logger
+
+	if !options.Opts.SuppressWarnings {
+		// surface warnings
+		logger = log.Log.WithName("KubeAPIWarningLogger")
+	} else {
+		// suppress warnings
+		logger = log.NullLogger{}
+	}
+
+	// Set a WarningHandler, the default WarningHandler is log.WarningLogger with
+	// deduplication enabled. See log.WarningLoggerOptions for considerations regarding
+	// deduplication.
+	rest.SetDefaultWarningHandler(
+		log.NewWarningLogger(
+			logger,
+			log.WarningLoggerOptions{
+				Deduplicate: !options.Opts.AllowDuplicateLogs,
+			},
+		),
+	)
 
 	// Init a scheme if none provided
 	if options.Scheme == nil {
@@ -109,6 +150,7 @@ func New(config *rest.Config, options Options) (Client, error) {
 var _ Client = &client{}
 
 // client is a client.Client that reads and writes directly from/to an API server.  It lazily initializes
+// new clients at the time they are used, and caches the client.
 // new clients at the time they are used, and caches the client.
 type client struct {
 	typedClient        typedClient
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -57,6 +59,15 @@ func New(config *rest.Config, options Options) (Client, error) {
 		return nil, fmt.Errorf("must provide non-nil rest.Config to client.New")
 	}
 
+	// Copy to avoid mutating the passed in config
+	config = rest.CopyConfig(config)
+
+	// Set a WarningHandler
+	config.WarningHandler = rest.NewWarningWriter(os.Stderr, rest.WarningWriterOptions{
```
alvaroaleman, line 66: Can we somehow use a logger for this?

alvaroaleman (commented):
/ok-to-test

pkg/client/client.go
MadhavJivrajani: Hi @alvaroaleman! Thanks for the feedback :smile:

This could work. This implements the `HandleWarningHeader` function as well but [uses](https://github.com/kubernetes/kubernetes/blob/f137c4777095b3972e2dd71a01365d47be459389/staging/src/k8s.io/client-go/rest/warnings.go#L66) `klog` for logging.
```go
config.WarningHandler = rest.WarningLogger{}
```

alvaroaleman: I think I was more thinking about something that would use the logger from `pkg/log`. This might imply a small wrapper?
It might also be a good idea to somehow get the deduplication logic, otherwise we might end up spamming the log

MadhavJivrajani: Hey @alvaroaleman, I tried thinking about the above suggestion, I'm a little unsure about the following which is why I was hoping to maybe get some feedback on the same:

- For writing a wrapper, did you have something such as the following in mind:
    ```go
    func (l DelegatingLogger) HandleWarningHeader(code int, agent string, text string) {
        ...
    }
    ```
    - And then a `DelegatingLogger` can be assigned as the `config.WarningHandler`
- For deduplication, a new attribute could be added to the `DelegatingLogger` called `written` which can be used to check for duplicate messages similar to as done [here](https://github.com/kubernetes/kubernetes/blob/b0abe89ae259d5e891887414cb0e5f81c969c697/staging/src/k8s.io/client-go/rest/warnings.go#L81).

alvaroaleman: The idea was more to build a wrapper that takes a `logr.Logger` rather than making it part of the `DelegatingLogger` itself in order to keep this composable

MadhavJivrajani: Oh I see, that makes sense!

To make sure I understand correctly, what you had in mind was a wrapper that is similar to the latest commit, but changed to something that would be used as follows:

```go
rest.SetDefaultWarningHandler(log.NewWarningLogger(
        log.Log.WithName("api-server-warning"),
        log.WarningLoggerOptions{
	        Deduplicate: true,
        }
    )
)
```
Is that what you had in mind?

alvaroaleman: Yes, that looks good. I would probably call it `KubeAPIWarningLogger`, but that is just a detail

MadhavJivrajani: Awesome! I'll make the changes 😄 

One last question: would it be a good idea to provide a field in [`Options`](https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/client/client.go#L37) to enable/disable surfacing of warnings should there be a case where said warnings aren't desirable?

alvaroaleman: I am agnostic there, I can not think of a reason why someone would not want deduplication but that doesn't mean its impossible.

MadhavJivrajani: Oh, I meant surfacing of warnings itself and not specifically de-duplication

pkg/client/client.go
```diff
@@ -56,6 +76,27 @@ func New(config *rest.Config, options Options) (Client, error) {
...
+	// deduplication enabled. See log.WarningLoggerOptions for considerations regarding
+	// deduplication.
+	rest.SetDefaultWarningHandler(
+		log.NewWarningLogger(
+			logger,
+			log.WarningLoggerOptions{
+				Deduplicate: !options.Opts.AllowDuplicateLogs,
+			},
+		),
+	)
```
alvaroaleman, line 99: I am wondering a bit if the client is the best place to plug this in. Maybe pkg/cluster would be better? Or both?
Many ppl will use the cache-backed client and all its read operations will not have this enabled since they go through the cache.
cc @vincepri

pkg/log/warning_handler.go
```diff
@@ -0,0 +1,74 @@
...
+			return
+		}
+		l.written[message] = struct{}{}
+	}
+	l.logger.Info(message)
+}
+
+// NewWarningLogger returns an implementation of rest.WarningHandler that logs warnings
+// with code = 299 using a DelegatingLogger.
+func NewWarningLogger(l logr.Logger, opts WarningLoggerOptions) *WarningLogger {
```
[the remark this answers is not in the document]
MadhavJivrajani: Yes, that seems more apt, I will make the changes 😄

pkg/log/warning_handler.go
```diff
@@ -0,0 +1,76 @@
...
+	if l.opts.Deduplicate {
+		l.writtenLock.Lock()
+		defer l.writtenLock.Unlock()
+
+		if _, alreadyLogged := l.written[message]; alreadyLogged {
+			return
+		}
+		l.written[message] = struct{}{}
+	}
+	l.logger.Info(message)
```
alvaroaleman, line 65:
```suggestion
	l.logger.Warn(message)
```

MadhavJivrajani: One clarification: if I understand correctly, the [`logr.Logger`](https://github.com/go-logr/logr/blob/e347122aefba5ae0d0d4bfdc14732ce482b3647a/logr.go#L148) interface doesn't implement a `Warn` level and instead recommends specifying verbosity levels for `Info` logs itself as explained [here](https://github.com/go-logr/logr#differences-from-daves-ideas).

alvaroaleman: I see, lets keep it at Info then

alvaroaleman (approved):
/lgtm
/approve

Thanks for your work!

MadhavJivrajani:
Thanks so much for the constant feedback @alvaroaleman! I had a great time working on this 😄

Issue #1357, opened 2021-01-22
envtest with Kubernetes 1.20


cbandy:
There is a kubebuilder-tools-1.20.2 since kubernetes-sigs/kubebuilder@3147a65d1a90f10c2d9c48c7e8497be6e6576f7b, but as pointed out in kubernetes-sigs/kubebuilder#1902 this version has stronger opinions about `--insecure` flags. See kubernetes/kubernetes@cfc2b330a7027e109ac26fc6a86b68898883099e.

When I try to use controller-runtime@v0.8.1 with kubebuilder-tools-1.20.2, the control plane does not start:

```
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 56497: only zero is allowed
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 56497: only zero is allowed
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 56497: only zero is allowed
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 56497: only zero is allowed
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 56497: only zero is allowed

    test.go:37: failed to start the controlplane. retried 5 times: timeout waiting for process kube-apiserver to start
```

alvaroaleman:
Yeah, getting this fixed and not using `insecure-port` in envtest would be great. In addition to working with newer kube versions, it would also allow us to write tests that require rbac and not be implemented via the `allow-all` authorized that is used for the insecure port, xref https://github.com/kubernetes-sigs/controller-runtime/issues/1346#issuecomment-779569663

/help

DirectXMan12:
(v old at this point though :-) )

DirectXMan12:
I'm poking at this now

[...]

PR #1477, opened 2021-04-14
🌱 Envtest exports secure config, helpers for simplify users adding

### LINKED ISSUE #983: 'envtest' should support test scenarios with Authn/Authz configurations (should expose secure endpoint info of kube-apiserver)

lobziik:
fixes #983

Bit extended #984, added `token-auth-file` generation with "kubeadmin" user, added couple helpers to envstest for users creation.

This PR added:
    `internal.integration.APIServer`: expose `TLSClientConfig` to connect its secure endpoint (@everpeace)
    `envtest.Environment`: introduce SecureConfig that just contains kube-apiserver's secure endpoint and its TLSClientConfig. (@everpeace)
     `envtest.AddAPIServerUser` helper for adding additional users
     `envtest.GetConfigForUser` helper for getting rest.Config with respective credentials
     token-auth-file creates automatically during APIServer startup
     fixed issue with missing certificates for APIServer after its restart, certs regenerates after controlplane start/stop

### DIFF
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -23,11 +23,16 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	authv1 "k8s.io/api/authentication/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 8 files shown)


lobziik:
cc @alvaroaleman, please take a look

pkg/envtest/envtest_test.go
```diff
@@ -845,4 +850,106 @@ var _ = Describe("Test", func() {
...
+
+		It("should throw error on attempt get config on stopped ControlPlane", func(done Done) {
+			env := &Environment{}
+			_, err := env.GetConfigForUser("test")
+			Expect(err).Should(HaveOccurred())
+			Expect(err.Error()).Should(Equal("control plane not started"))
+			close(done)
+		})
+
+		It("should successfully add user and able to get valid config if ControlPlane not yet running", func(done Done) {
```
elmiko, line 919: minor nit
```suggestion
		It("should successfully add user and be able to get valid config if ControlPlane not yet running", func(done Done) {
```

pkg/envtest/server.go
```diff
@@ -279,6 +302,39 @@ func (te *Environment) Start() (*rest.Config, error) {
 	return te.Config, err
 }
 
+//AddAPIServerUser helper function for adding APIServer user. Should be using while ControlPlane NOT running
```
elmiko, line 305: minor nit
```suggestion
//AddAPIServerUser helper function for adding APIServer user. Should be used while ControlPlane NOT running
```

pkg/envtest/server.go
```diff
@@ -279,6 +302,39 @@ func (te *Environment) Start() (*rest.Config, error) {
...
+		return nil, fmt.Errorf("users cannot be added while ControlPlane running")
+	}
+
+	if te.KubeAPIServerUsers == nil {
+		te.KubeAPIServerUsers = make(map[string]*integration.User)
+	}
+	newUser := integration.User{
+		Token:  uuid.New().String(),
+		Name:   username,
+		UID:    username,
```
elmiko, line 317: i'm guessing this does not need to be a numeric value?

(i see it's a string in the struct, just curious)

pkg/envtest/server.go
```diff
@@ -279,6 +302,39 @@ func (te *Environment) Start() (*rest.Config, error) {
...
+		Groups: groups,
+	}
+	te.KubeAPIServerUsers[username] = &newUser
+	return &newUser, nil
+}
+
+// GetConfigForUser returns *rest.Config for given username if user exists. Should be using while ControlPlane IS running
```
elmiko, line 324: minor nit
```suggestion
// GetConfigForUser returns *rest.Config for given username if user exists. Should be used while ControlPlane IS running
```

pkg/internal/testing/integration/apiserver.go
```diff
@@ -50,6 +67,16 @@ type APIServer struct {
 	// directory, and the Stop() method will clean it up.
 	CertDir string
 
+	// AuthFiletDir is a path to a directory containing token-auth-file which needed for setup
```
elmiko, line 70: minor nit
```suggestion
	// AuthFiletDir is a path to a directory containing token-auth-file which is needed to setup
```

elmiko (commented):
i am not overly familiar with this code, but the pr generally makes sense to me. i have a few minor suggestions and a question.

pkg/envtest/server.go
lobziik: Should be a string, as I understand from doc:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authentication-strategies

>UID: a string which identifies the end user and attempts to be more consistent and unique than username.

alvaroaleman:
The underlying issue this fixes is assigned to @DirectXMan12 (ref https://github.com/kubernetes-sigs/controller-runtime/issues/1357#issuecomment-799757100) and they mentioned over Slack that something to review will be ready in the next couple of days. I'll refrain from reviewing this one for now.

lobziik:
/close
because of #1486 . These is cooler.

Issue #1476, opened 2021-04-14
CreateOrUpdate will panic when MutateFunc is nil


joshrosso:
[CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/controller/controllerutil/controllerutil.go#L197) will panic when the `MutateFn` argument is nil. This is because [mutate](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/controller/controllerutil/controllerutil.go#L325-L334) and on nil would panic, as demonstrated in this [simple go playground example](https://play.golang.org/p/_U-E08geRMY).

PR #1478, opened 2021-04-14
:bug: ensure mutate does not call func when nil

joshrosso:
This commit ensures the mutate call will not call the function passed in
(via argument) when the argument received is nil. This prevents a nil
pointer dereference.

Note this protects mutate from within, which IMO is a better model because functions don't need to know how to safely call it.

resolves: https://github.com/kubernetes-sigs/controller-runtime/issues/1476

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -237,10 +237,8 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 		if !errors.IsNotFound(err) {
 			return OperationResultNone, err
 		}
-		if f != nil {
-			if err := mutate(f, key, obj); err != nil {
-				return OperationResultNone, err
-			}
+		if err := mutate(f, key, obj); err != nil {
+			return OperationResultNone, err
 		}
 		if err := c.Create(ctx, obj); err != nil {
 			return OperationResultNone, err
@@ -324,12 +322,17 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 
 // mutate wraps a MutateFn and applies validation to its result
 func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
-	if err := f(); err != nil {
-		return err
+	// if f is nil, there is no mutate operation to be done, in which case just return
+	if f != nil {
+
+		if err := f(); err != nil {
+			return err
+		}
+		if newKey := client.ObjectKeyFromObject(obj); key != newKey {
+			return fmt.Errorf("MutateFn cannot mutate object name and/or object namespace")
+		}
 	}
-	if newKey := client.ObjectKeyFromObject(obj); key != newKey {
-		return fmt.Errorf("MutateFn cannot mutate object name and/or object namespace")
-	}
+
 	return nil
 }
 


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -237,10 +237,8 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 		if !errors.IsNotFound(err) {
 			return OperationResultNone, err
 		}
-		if f != nil {
-			if err := mutate(f, key, obj); err != nil {
-				return OperationResultNone, err
-			}
```
vincepri, line 243: Wasn't this covered by this if branch already? The mutate function wasn't being called if `f` is nil

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -324,12 +322,17 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
...
 // mutate wraps a MutateFn and applies validation to its result
 func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
-	if err := f(); err != nil {
-		return err
-	}
-	if newKey := client.ObjectKeyFromObject(obj); key != newKey {
-		return fmt.Errorf("MutateFn cannot mutate object name and/or object namespace")
+	// if f is nil, there is no mutate operation to be done, in which case just return
+	if f != nil {
+
```
vincepri, line 327:
```suggestion
	// if f is nil, there is no mutate operation to be done, in which case just return
	if f == nil {
	  return nil
	}
```

Prefer to return right away rather than having everything in an if block

pkg/controller/controllerutil/controllerutil.go
joshrosso: @vincepri in CreateOrPatch, yes, but it was not being checked in [CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/18cb30ed465c607607dbd53dbda44ff99d265ed9/pkg/controller/controllerutil/controllerutil.go#L213), thus triggering a nil pointer dereference when mutate was nil.

Rather than adding another check in `CreateOrUpdate`, I figured it was smarter to make `mutate` handle this check as IMO callers of mutate should not have to be concerned with how to call it safely.

pkg/controller/controllerutil/controllerutil.go
joshrosso: @vincepri happy to do this. The trade-off here is now you have 2 locations (top and bottom) of func where a successful return can occur.

Just want to check you prefer that relative to the `if` as it stands.

Let me know.

alvaroaleman:
@vincepri @joshrosso why are we adding nilchecks for things? This is very surprising to me, in all the go code bases I've seen so far it is not common to have functions nilcheck their parameters unless they explicitly declare a nil param valid. Calling `CreateOrPatch` with a `nil` arg is a programming bug. I would really prefer not to scatter nil checks all over our codebase. If anything, we could make the `Controller` recover panics in the reconciler.

alvaroaleman:
We already did this before but I have to say that I find code that silently does nothing when used wrongly extremely questionable.

vincepri:
Given that this was the behavior before, how would you suggest to change the function signature or the contract?

alvaroaleman:
> Given that this was the behavior before, how would you suggest to change the function signature or the contract?

Make a breaking change and remove the handling of nil args.

pkg/controller/controllerutil/controllerutil.go
aayushrangwala: +1 on the @vincepri suggestion. having multiple if-else creates branching. Should just check the terminating condition first

joshrosso:
> @vincepri @joshrosso why are we adding nilchecks for things? This is very surprising to me, in all the go code bases I've seen so far it is not common to have functions nilcheck their parameters unless they explicitly declare a nil param valid. Calling CreateOrPatch with a nil arg is a programming bug. I would really prefer not to scatter nil checks all over our codebase. If anything, we could make the Controller recover panics in the reconciler.

Fair. I guess i'm more interested in the pragmatic solution relative to the most idiomatic/philosophically sound.

At the end of the day `mutate` is a private function that gets used in 2 contexts. I wouldn't lose sleep over this claiming that a nil func is a noop. But anyway, this isn't my codebase to maintain over time.

How about I setup `mutate` to return an error if the func is nil? Then, at least it can be better handled relative to the current panic that happens.

If not, no worries. I'm using a different lib now so this isn't a pressing issue for me, happy to let someone do the more "proper" change and just close this PR :).

PR #1479, opened 2021-04-16
✨ Cache-Backed Client: Support `listOpts.Limit`

### LINKED ISSUE #1422: `delegatingClient` with `client.informerCache` does not respect `client.Limit` option
When listing resources, I found that `client.delegatingClient` with `client.informerCache` does not respect `client.Limit` option. Scanning the source, I could not find any provisions to apply the limit when reading from the cache. This should be trivial to implement, no? 

Also, I believe that `kubernetesfakes` suffers from the same issue (but that is less concerning to me).

MadhavJivrajani:
Enabled list options in `List` implementation of `Reader` by the infromer cache in order to respect `client.Limit` if set. 

Fix #1422

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -100,6 +100,8 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			knownPod2           client.Object
 			knownPod3           client.Object
 			knownPod4           client.Object
+			knownPod5           client.Object
+			knownPod6           client.Object
 		)
 
 		BeforeEach(func() {
@@ -122,14 +124,20 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			knownPod2 = createPod("test-pod-2", testNamespaceTwo, kcorev1.RestartPolicyAlways)
 			knownPod3 = createPodWithLabels("test-pod-3", testNamespaceTwo, kcorev1.RestartPolicyOnFailure, map[string]string{"common-label": "common"})
 			knownPod4 = createPodWithLabels("test-pod-4", testNamespaceThree, kcorev1.RestartPolicyNever, map[string]string{"common-label": "common"})
+			knownPod5 = createPod("test-pod-5", testNamespaceOne, kcorev1.RestartPolicyNever)
+			knownPod6 = createPod("test-pod-6", testNamespaceTwo, kcorev1.RestartPolicyAlways)
+
 			podGVK := schema.GroupVersionKind{
 				Kind:    "Pod",
 				Version: "v1",
 			}
+
 			knownPod1.GetObjectKind().SetGroupVersionKind(podGVK)
 			knownPod2.GetObjectKind().SetGroupVersionKind(podGVK)
 			knownPod3.GetObjectKind().SetGroupVersionKind(podGVK)
 			knownPod4.GetObjectKind().SetGroupVersionKind(podGVK)
+			knownPod5.GetObjectKind().SetGroupVersionKind(podGVK)
+			knownPod6.GetObjectKind().SetGroupVersionKind(podGVK)
 
 			By("creating the informer cache")
 			informerCache, err = createCacheFunc(cfg, cache.Options{})
@@ -149,6 +157,8 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 			deletePod(knownPod2)
 			deletePod(knownPod3)
 			deletePod(knownPod4)
+			deletePod(knownPod5)
+			deletePod(knownPod6)
 
 			informerCacheCancel()
 		})
@@ -226,7 +236,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 
 					By("verifying that the returned pods have GVK populated")
 					Expect(out.Items).NotTo(BeEmpty())
-					Expect(out.Items).Should(SatisfyAny(HaveLen(3), HaveLen(4)))
+					Expect(out.Items).Should(SatisfyAny(HaveLen(5), HaveLen(6)))
 					for _, p := range out.Items {
 						Expect(p.GroupVersionKind()).To(Equal(kcorev1.SchemeGroupVersion.WithKind("Pod")))
 					}
@@ -240,9 +250,10 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 
 					By("verifying that the returned pods are in test-namespace-1")
 					Expect(listObj.Items).NotTo(BeEmpty())
-					Expect(listObj.Items).Should(HaveLen(1))
-					actual := listObj.Items[0]
-					Expect(actual.Namespace).To(Equal(testNamespaceOne))
+					Expect(listObj.Items).Should(HaveLen(2))
+					for _, item := range listObj.Items {
+						Expect(item.Namespace).To(Equal(testNamespaceOne))
+					}
 				})
 
 				It("should deep copy the object unless told otherwise", func() {
@@ -295,20 +306,12 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					Expect(errors.IsTimeout(err)).To(BeTrue())
 				})
 
-				It("should limit number of objects listed to 1", func() {
-					By("setting the Limit option to 1")
-					opts := &client.ListOptions{Limit: int64(1)}
-					limitListOpts := &client.ListOptions{}
-					opts.ApplyToList(limitListOpts)
-					Expect(limitListOpts).To(Equal(opts))
-
-					By("verifying that only Limit (1) number of objects are retrieved from the cache")
+				It("should set the Limit option and limit number of objects to Limit when List is called", func() {
+					opts := &client.ListOptions{Limit: int64(3)}
+					By("verifying that only Limit (3) number of objects are retrieved from the cache")
 					listObj := &kcorev1.PodList{}
-					Expect(informerCache.List(context.Background(), listObj, limitListOpts)).To(Succeed())
-					Expect(listObj.Items).Should(HaveLen(1))
-
-					By("verifying the Continue field is set")
-					Expect(listObj.ListMeta.Continue).ShouldNot(HaveLen(0))
+					Expect(informerCache.List(context.Background(), listObj, opts)).To(Succeed())
+					Expect(listObj.Items).Should(HaveLen(3))
 				})
 			})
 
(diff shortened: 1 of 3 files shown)


alvaroaleman:
@MadhavJivrajani can you please add a test for the `Limit` option?

MadhavJivrajani:
Oh woops, sorry about that. I'll add it :)

Thanks!

MadhavJivrajani:
@alvaroaleman I've added the test, could you PTAL?

pkg/cache/informer_cache_unit_test.go
```diff
@@ -75,4 +76,11 @@ var _ = Describe("ip.objectTypeForListObject", func() {
 			Expect(obj).To(Equal(referenceObject))
 		})
 	})
+
+	It("should set Limit field", func() {
```
alvaroaleman, line 80: We need to test behavior, not implementation details, i.E. here that it returns at most `Limit` objects even if the cache has more

MadhavJivrajani: Oh I see. Sorry about that. To be sure, such a test would reside in `cache_test.go`?

alvaroaleman: In pkg/cache, the file name is not particularly relevant

MadhavJivrajani:
@alvaroaleman could you PTAL?

pkg/cache/informer_cache.go
```diff
@@ -71,6 +71,9 @@ func (ip *informerCache) Get(ctx context.Context, key client.ObjectKey, out clie
 // List implements Reader
 func (ip *informerCache) List(ctx context.Context, out client.ObjectList, opts ...client.ListOption) error {
 
+	listOpts := client.ListOptions{}
+	listOpts.ApplyOptions(opts)
```
alvaroaleman, line 75: I don't understand how this changes anything, this variable is never accessed or am I completely missing something?

MadhavJivrajani: No, you're right, this was a very silly oversight on my part, I don't remember why I didn't implement the remainder of it, I apologize. I will fix it.

MadhavJivrajani:
Hey @alvaroaleman, the `Limit` option doesn't seem to be enabled for multi-namespace cached either, is that required? And if so is it okay if I try and implement it as part of this PR or should I open another issue for the same?

alvaroaleman:
> Hey @alvaroaleman, the Limit option doesn't seem to be implemented for multi-namespace cached either, would it be okay to have a Limit option there as well? And if so is it okay if I try and implement it as part of this PR or should I open another issue for the same?

It would be good to support it everywhere at once, if that isn't too much work.

MadhavJivrajani:
Sure, no worries. 😄

MadhavJivrajani:
Hey @alvaroaleman, could you PTAL?

alvaroaleman (commented):
@MadhavJivrajani thanks, sorry, this fell of my radar. The MultiNamespaceCache implementation is incorrect and just passes the tests by coincidence, because we only have one pod per namespace in the test scenario.

If you do something it instantly starts failing:

```diff
diff --git a/pkg/cache/cache_test.go b/pkg/cache/cache_test.go
index 9a41662a..6f0cdd39 100644
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -100,6 +100,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
                        knownPod2           client.Object
                        knownPod3           client.Object
                        knownPod4           client.Object
+                       knownPod99          client.Object
                )
 
                BeforeEach(func() {
@@ -119,6 +120,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
                        Expect(err).NotTo(HaveOccurred())
                        // Includes restart policy since these objects are indexed on this field.
                        knownPod1 = createPod("test-pod-1", testNamespaceOne, kcorev1.RestartPolicyNever)
+                       knownPod99 = createPod("test-pod-99", testNamespaceOne, kcorev1.RestartPolicyNever)
                        knownPod2 = createPod("test-pod-2", testNamespaceTwo, kcorev1.RestartPolicyAlways)
                        knownPod3 = createPodWithLabels("test-pod-3", testNamespaceTwo, kcorev1.RestartPolicyOnFailure, map[string]string{"common-label": "common"})
                        knownPod4 = createPodWithLabels("test-pod-4", testNamespaceThree, kcorev1.RestartPolicyNever, map[string]string{"common-label": "common"})
@@ -149,6 +151,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
                        deletePod(knownPod2)
                        deletePod(knownPod3)
                        deletePod(knownPod4)
+                       deletePod(knownPod99)
 
                        informerCacheCancel()
                })
@@ -295,7 +298,7 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
                                        Expect(errors.IsTimeout(err)).To(BeTrue())
                                })
 
-                               It("should set the Limit option and limit number of objects to Limit when List is called", func() {
+                               FIt("should set the Limit option and limit number of objects to Limit when List is called", func() {
                                        By("setting the Limit option to 1")
                                        opts := &client.ListOptions{Limit: int64(1)}
                                        limitListOpts := &client.ListOptions{}
```

I am fine with both leaving it for a follow-up (but then, please disable the test for the multi-namespace cache) or fixing it here (I think it actually shouldn't be too hard).

If you fix it, please also change the tests to actually have a namespace with more than one object :)

pkg/cache/multi_namespace_cache.go
```diff
@@ -173,6 +182,7 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 		allItems = append(allItems, items...)
 		// The last list call should have the most correct resource version.
 		resourceVersion = accessor.GetResourceVersion()
+		alreadyFetched++
```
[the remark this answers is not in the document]
MadhavJivrajani: Ohh you're right, sorry about that. I will fix that.

pkg/cache/multi_namespace_cache.go
```diff
@@ -155,8 +155,17 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
...
 	}
+
+	alreadyFetched := int64(len(allItems))
+	limitSet := listOpts.Limit > 0
+
 	var resourceVersion string
 	for _, cache := range c.namespaceToCache {
+		// if the Limit option is set and the number of items
+		// listed exceeds this limit, then stop reading.
+		if limitSet && alreadyFetched >= listOpts.Limit {
```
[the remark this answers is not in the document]
MadhavJivrajani: That sounds great. I will make the change.

MadhavJivrajani:
> If you fix it, please also change the tests to actually have a namespace with more than one object :)

Will do 👍🏻

pkg/cache/multi_namespace_cache.go
```diff
@@ -155,8 +155,17 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 	if err != nil {
 		return err
 	}
+
+	alreadyFetched := int64(len(allItems))
```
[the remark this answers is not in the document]
MadhavJivrajani: Yeah you're right. I think I got a little confused, and for some reason thought that the client might pass a pre-existing `listObj` to the `List` (potentially from a previously `Limit`ed read) which now that I think more about it, doesn't make sense to do at all.

Will change. 👍🏻

MadhavJivrajani:
@alvaroaleman any idea as to why it says:
>  testing: warning: no tests to run 

For some reason, both `${MOD_OPT}` and `${WHAT}` in `hack/test-all.sh` seem to be empty?

alvaroaleman:
> For some reason, both ${MOD_OPT} and ${WHAT} in hack/test-all.sh seem to be empty?

Lol, thanks for noticing. It seems we don't run any unittests in CI currently. I've opened https://github.com/kubernetes-sigs/controller-runtime/pull/1503 for that (And we will not merge anything until that one is in).

pkg/cache/multi_namespace_cache.go
```diff
@@ -166,6 +174,18 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
...
 		}
+		if limitSet {
+			// decrement Limit by the number of items
+			// fetched from the current namespace.
+			listOpts.Limit -= int64(len(items))
+			// if the Limit is exceeded while
+			// reading from the current namespace,
+			// then keep only the number of items
+			// needed to not overshoot the Limit.
+			if listOpts.Limit < 0 {
```
alvaroaleman, line 185: Since each individual cache will respect the limit, this shouldn't be possible?

pkg/cache/multi_namespace_cache.go
```diff
@@ -166,6 +174,18 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 		if err != nil {
 			return err
 		}
+		if limitSet {
+			// decrement Limit by the number of items
+			// fetched from the current namespace.
+			listOpts.Limit -= int64(len(items))
```
alvaroaleman, line 180: I think it would be more obvious if we just returned here when `Limit == 0`

pkg/cache/cache_test.go
```diff
@@ -295,7 +305,20 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					Expect(errors.IsTimeout(err)).To(BeTrue())
 				})
 
+				It("should set the Limit option and limit number of objects to Limit when List is called", func() {
+					By("setting the Limit option to 3")
+					opts := &client.ListOptions{Limit: int64(3)}
+					limitListOpts := &client.ListOptions{}
+					opts.ApplyToList(limitListOpts)
+					Expect(limitListOpts).To(Equal(opts))
```
alvaroaleman, line 313:
```suggestion
```

I would expect this to be tested in the client already

pkg/cache/multi_namespace_cache.go
MadhavJivrajani: My understanding was that, since `opts` is being passed into the `List`, if the original `Limit` was `3`, and if we go with a similar configuration as in the test, from the first namespace it would read out `2` items (`limitOpts.Limit` now being `1`), in the `List` of the next namepsace, the passed in `ListOptions` still would have a `Limit` of `3`, it would read out `2` items from the second namespace and the `limitOpts.Limit` would now be `-1` and the total `List`ed items would be `4`, which overshoots the original `Limit`

wdyt?

alvaroaleman: You are forgetting that we adjust `listOpts.Limit` after every time we read something: `listOpts.Limit -= int64(len(items))`

pkg/cache/multi_namespace_cache.go
alvaroaleman: nvm, we can not return as we need to go through the rest of the function

pkg/cache/multi_namespace_cache.go
MadhavJivrajani: If I understand correctly, the decremented `listOpts.Limit` won't really be reflected in `opts`?
https://github.com/kubernetes-sigs/controller-runtime/blob/3b25aa6a620cd80323abc0f31613bb1ab2014d08/pkg/cache/multi_namespace_cache.go#L192

alvaroaleman: Well, replace `opts...` with `listOpts`

alvaroaleman (approved):
/lgtm
/approve
Thanks!

Issue #1453, opened 2021-03-25
[Feature Request] Finalizer helper library


rashmigottipati:
In Kubernetes garbage collection model, OwnerReferences are allowed within a namespace wherein the "dependent" object has anOwnerReference field referencing the "parent" object. If the parent object is being deleted, Kubernetes garbage collector will delete all of its dependent objects. As cross-namespace owner references are not allowed by design; a Finalizer should be used to perform cleanup.

Finalizers are more powerful and broadly applicable than owner references because they permit customizable deletion steps for a controller to define how an object is cleaned up, ex. updating a CR's status during deletion with a particular message. This gives Operator authors control over CR deletion proceeds. 

Operator projects would benefit from a finalizer helper library in controller-runtime which will make it easy for users to define their own finalizer struct implementations and use the abstractions we include to register the Finalizers and hand the Finalizers to their Reconciler and call Finalize function.

Please see @joelanford's high level thought on the implementation details: https://github.com/kubernetes-sigs/kubebuilder/issues/2010#issuecomment-785558621.

/kind feature

Adirio:
Pinging @joelanford as he suggested this approach.

coderanger:
https://github.com/coderanger/controller-utils/blob/main/core/reconciler.go#L277-L290 is my implementation of the general helper pattern for this. It checks on initialization (up above) for if the component implements the finalizer interface and if so, it will call Finalizer() on that component when needed.

joelanford:
My high level thought was to have something like this:
```go
type Registerer interface {
	Register(key string, f Finalizer) error
}

type Finalizer interface {
	Finalize(client.Object) (requeue bool, err error)
}

type Finalizers interface {
	// implements Registerer and Finalizer to finalize
	// all registered finalizers if the provided object
	// has a deletion timestamp or set all registered
	// finalizers if it doesn't
	Registerer
	Finalizer
}

func NewFinalizers() Finalizers {
	return finalizers(map[string]Finalizer{})
}

type finalizers map[string]Finalizer

func (f finalizers) Register(key string, finalizer Finalizer) error {
	if _, ok := f[key]; ok {
		return fmt.Errorf("finalizer for key %q already registered")
	}
	f[key] = finalizer
	return nil
}

func (f finalizers) Finalize(obj client.Object) (bool, error) {
	requeue := false
	for key, finalizer := range f {
[...]

PR #1481, opened 2021-04-19
✨[WIP] Finalizer helper library

rashmigottipati:
PR for adding a finalizer helper library to controller-runtime 
Issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1453

Signed-off-by: rashmigottipati <chowdary.grashmi@gmail.com>


pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,55 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func NewFinalizers() Finalizers {
+	return finalizers(map[string]Finalizer{})
+}
+
+func (f finalizers) Register(key string, finalizer Finalizer) error {
+	if _, ok := f[key]; ok {
+		return fmt.Errorf("finalizer for key %q already registered")
```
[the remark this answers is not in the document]
joelanford: I'd suggest keeping `%q`, which is a quoted string. That way if the key is an empty string, the message does not end up being: 

```
finalizer for key  already registered
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (f finalizers) Finalize(ctx context.Context, obj client.Object) (bool, error) {
+	needsUpdate := false
+	for key, finalizer := range f {
+		if obj.GetDeletionTimestamp().IsZero() && !controllerutil.ContainsFinalizer(obj, key) {
+			controllerutil.AddFinalizer(obj, key)
+			needsUpdate = true
+		} else if obj.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(obj, key) {
+			needsUpdate, err := finalizer.Finalize(ctx, obj)
```
coderanger, line 44: It should still try to run other finalizers that might be on the object, since they are all independent and might be async. Also this assumes that finalizers always complete which may not be the case if the cleanup takes time.

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("finalizer for key %q already registered", key)
+	}
+	f[key] = finalizer
+	return nil
+}
+
+func (f finalizers) Finalize(ctx context.Context, obj client.Object) (bool, error) {
+	needsUpdate := false
+	for key, finalizer := range f {
+		if obj.GetDeletionTimestamp().IsZero() && !controllerutil.ContainsFinalizer(obj, key) {
```
estroz, line 40:
```suggestion
		if dt := obj.GetDeletionTimestamp(); (dt == nil || dt.IsZero()) && !controllerutil.ContainsFinalizer(obj, key) {
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+			controllerutil.AddFinalizer(obj, key)
+			needsUpdate = true
+		} else if obj.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(obj, key) {
```
estroz, line 43:
```suggestion
		} else if dt != nil && controllerutil.ContainsFinalizer(obj, key) {
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+			ret, err := finalizer.Finalize(ctx, obj)
+			if err != nil {
+				return ret, fmt.Errorf("finalize failed for key %q: %v", key, err)
```
estroz, line 46: Keys are finalizer names here
```suggestion
				return ret, fmt.Errorf("finalizer %q failed: %v", key, err)
```

pkg/finalizer/types.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Finalizer holds Finalize that will add/remove a finalizer based on the
+// deletion timestamp being set and return an indication of whether the
+// obj needs an update or not
+type Finalizer interface {
+	Finalize(context.Context, client.Object) (needsUpdate bool, err error)
+}
+
+type finalizers map[string]Finalizer
```
estroz, line 36: Nit: move this to finalizer.go since that's where the concrete implementation lives.

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// NewFinalizers returns the Finalizers interface
+func NewFinalizers() Finalizers {
+	return finalizers(map[string]Finalizer{})
```
estroz, line 26:
```suggestion
	return finalizers{}
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (f finalizers) Register(key string, finalizer Finalizer) error {
+	if _, ok := f[key]; ok {
```
estroz, line 30: I'm kinda thinking this should be removed, unless we want to prevent re-registering a finalizer in all cases.

pkg/finalizer/finalizer.go
coderanger: I'm pretty sure IsZero has an `== nil` check in it already.

estroz: If the `*time.Time` returned by `GetDeletionTimestamp()` is nil, then calling a method on it will panic.

pkg/finalizer/finalizer.go
estroz: The only case I could see re-registering to be ok is when the finalizer logic implemented in a `Finalizer` has been updated based on some runtime information. But I guess there are workarounds to that case that don't involve removing this re-registration restriction.

pkg/finalizer/finalizer.go
coderanger: I don't think that's correct, https://play.golang.org/p/JtZblcMxSd9 seems to work for me. Pretty sure Go allows defining methods on null pointers like any other typed pointer, the value passed in will just be nil.

estroz: If a method has to dereference a pointer receiver to run and the receiver is nil, then the method will panic. See https://play.golang.org/p/1VZxyB9uq-c. Since `IsZero()` is not implemented on a pointer, it will panic.

coderanger: GetDeletionTimestamp() returns a *metav1.Time. Our version of IsZero is defined with a pointer receiver https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L59

estroz: Ah silly me, read it as std time. Resolving this comment.

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (f finalizers) Finalize(ctx context.Context, obj client.Object) (bool, error) {
+	var errList []error
+	needsUpdate := false
+	for key, finalizer := range f {
+		if dt := obj.GetDeletionTimestamp(); dt.IsZero() && !controllerutil.ContainsFinalizer(obj, key) {
+			controllerutil.AddFinalizer(obj, key)
+			needsUpdate = true
+		} else if !dt.IsZero() && controllerutil.ContainsFinalizer(obj, key) {
+			ret, err := finalizer.Finalize(ctx, obj)
```
estroz, line 48: Nit: more obvious name
```suggestion
			finalizerNeedsUpdate, err := finalizer.Finalize(ctx, obj)
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,59 @@
...
+			if err != nil {
+				errList = append(errList, fmt.Errorf("finalizer %q failed: %v", key, err))
+			}
+			if ret == true {
+				needsUpdate = true
+			}
```
estroz, line 54:
```suggestion
			needsUpdate = needsUpdate || finalizerNeedsUpdate
```

pkg/finalizer/finalizer.go
estroz: Yeah that's what I'm thinking too.

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,59 @@
...
+			controllerutil.RemoveFinalizer(obj, key)
```
joelanford, line 55: We should only remove the finalizer if it succeeded, and in that case, we should always set `needsUpdate` to `true`.

```suggestion
			finalizerNeedsUpdate, err := finalizer.Finalize(ctx, obj)
			if err != nil {
				// Even when the finalizer fails, it may need to signal to update the primary
				// object (e.g. it may set a condition and need a status update).
				needsUpdate = needsUpdate || finalizerNeedsUpdate
				errList = append(errList, fmt.Errorf("finalizer %q failed: %v", key, err))
			} else {
				// If the finalizer succeeds, we remove the finalizer from the primary 
				// object's metadata, so we know it will need an update.
				needsUpdate = true
				controllerutil.RemoveFinalizer(obj, key)
			}
```

> it may set a condition and need a status update

This makes me wonder if we need separate `needsUpdate` and `needsStatusUpdate` return values?

pkg/finalizer/finalizer_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Expect(pod).NotTo(BeNil())
+
+		finalizers = NewFinalizers()
+		Expect(finalizers).NotTo(BeNil())
+
+		f := mockFinalizer{}
+		Expect(f).NotTo(BeNil())
+
+	})
+	Describe("Finalizer helper library", func() {
```
joelanford, line 46: Nit:
```suggestion
	Describe("Register", func() {
```

And then put just the Register tests in this block. Ditto for `Finalize`

And then the `It` tests can be `It("successfully registers a finalizer")` to make it read more natually.

pkg/finalizer/finalizer_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			pod.DeletionTimestamp = nil
+			pod.Finalizers = []string{}
+
+			needsUpdate, err := finalizers.Finalize(context.TODO(), pod)
+			Expect(err).To(BeNil())
+			Expect(needsUpdate).To(BeTrue())
+		})
+	})
+})
```
joelanford, line 92: Can you add some other tests to cover:
1. The combinations of the pod having deletion timestamp (or not) and the finalizers (or not)
2. The combinations of the finalizer returning needsUpdates true/false and err nil/non-nil.

Based on the change to collect errors from finalizers instead of returning on the first error, we probably also need some cases to cover adding and removing multiple finalizers.

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	f[key] = finalizer
+	return nil
+}
+
+func (f finalizers) Finalize(ctx context.Context, obj client.Object) (bool, error) {
+	var errList []error
+	needsUpdate := false
+	for key, finalizer := range f {
+		if dt := obj.GetDeletionTimestamp(); dt.IsZero() && !controllerutil.ContainsFinalizer(obj, key) {
```
aayushrangwala, line 44: This could be a switch case to make It more clean

pkg/finalizer/types.go
```diff
@@ -0,0 +1,42 @@
...
+// deletion timestamp being set and return an indication of whether the
+// obj needs an update or not
+type Finalizer interface {
+	Finalize(context.Context, client.Object) (needsUpdate bool, err error)
+}
+
+// Finalizers implements Registerer and Finalizer to finalize all registered
+// finalizers if the provided object has a deletion timestamp or set all
+// registered finalizers if it does not
+type Finalizers interface {
```
aayushrangwala, line 39: Should we add a sample example of a finaliser to be implemented by the consumers

estroz:
I don't think the failing tests are related to changes in this PR.

pkg/finalizer/types.go
rashmigottipati: Good idea Aayush. I will do that in a follow-up PR.

joelanford (commented):
A quick pass, but looking good!

pkg/finalizer/finalizer_test.go
```diff
@@ -0,0 +1,197 @@
...
+		})
+
+		It("should return no error and return false for needsUpdate when deletion timestamp is nil and finalizer doesn't exist", func() {
+			pod.DeletionTimestamp = nil
+			pod.Finalizers = []string{}
+
+			needsUpdate, err := finalizers.Finalize(context.TODO(), pod)
+			Expect(err).To(BeNil())
+			Expect(needsUpdate).To(BeFalse())
+		})
```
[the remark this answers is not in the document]
rashmigottipati: Yes, they're effectively the same. Removed the initial one and left the other one intact.

pkg/finalizer/finalizer_test.go
```diff
@@ -0,0 +1,197 @@
...
+		It("successfully finalizes and returns true for needsUpdate when deletion timestamp is nil and finalizer does not exist", func() {
+			err = finalizers.Register("finalizers.sigs.k8s.io/testfinalizer", f)
+			Expect(err).To(BeNil())
+
+			pod.DeletionTimestamp = nil
+			pod.Finalizers = []string{}
+
+			needsUpdate, err := finalizers.Finalize(context.TODO(), pod)
+			Expect(err).To(BeNil())
+			Expect(needsUpdate).To(BeTrue())
```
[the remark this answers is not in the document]
rashmigottipati: Done!

rashmigottipati:
@joelanford thanks for the review! Addressed the latest set of comments. Could you please take another look?

joelanford:
I think the last question I have is whether we want to split the `needsUpdate` value in two: `needsUpdate` and `needsStatusUpdate`.

Without that, callers using CRDs with the status subresource will likely need to call both `client.Update()` and `client.Status().Update()`

On the otherhand, an API with three return values starts to get into "is this the right API?" territory.

estroz:
@joelanford I agree; something like [OperationResult](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil?utm_source=godoc#OperationResult) would be useful here.

joelanford:
+1 @estroz. Are you thinking of using that directly or adding a separate type in the finalizer package that conveys just what a finalizer would need? I vote the latter.

IF you agree, I think there's two options for how to do this:
```go
type Result struct {
   Updated bool
   StatusUpdated bool
}
```
or

```go
type Result string

const (
	ResultUnchanged     Result = "unchanged"
	ResultUpdated       Result = "updated"
	ResultUpdatedStatus Result = "updatedStatus"
    ResultUpdatedAll    Result = "updatedAll"
)
```

I think I prefer the result struct rather than the result string, since it seems easier to use in conditional expressions that drive the client update calls.

estroz:
A new type for sure, and I'm fine with the struct approach (pretty similar to `Option<T>` return types in functional languages that get pattern-matched).

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,73 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type finalizers map[string]Finalizer
+
+// Result struct holds Updated and StatusUpdated fields
+type Result struct {
+	Updated       bool
+	StatusUpdated bool
```
estroz, line 30:
```suggestion
// Result struct holds information about what parts of an object were updated by finalizer(s).
type Result struct {
	// Updated will be true if at least one of the object's non-status field
	// was updated by some registered finalizer.
	Updated bool
	// StatusUpdated will be true if at least one of the object's status' fields
	// was updated by some registered finalizer.
	StatusUpdated bool
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,73 @@
...
+				// Even when the finalizer fails, it may need to signal to update the primary
+				// object (e.g. it may set a condition and need a status update).
+				res.Updated = res.Updated || finalizerRes.Updated
+				res.StatusUpdated = res.StatusUpdated || finalizerRes.StatusUpdated
+				errList = append(errList, fmt.Errorf("finalizer %q failed: %v", key, err))
+			} else {
+				// If the finalizer succeeds, we remove the finalizer from the primary
+				// object's metadata, so we know it will need an update.
+				res.Updated = true
+				controllerutil.RemoveFinalizer(obj, key)
```
estroz, line 68:
```suggestion
				controllerutil.RemoveFinalizer(obj, key)
				// The finalizer may have updated the status too.
				res.StatusUpdated = res.StatusUpdated || finalizerRes.StatusUpdated
```

estroz (approved):
/lgtm

alvaroaleman:
@estroz @joelanford Can we in the future please either tell our contributors to squash their commits or use the label to have them squashed on merge? Something like this clutters the git history bigtimes for no good reason.

estroz:
@alvaroaleman yes, a worthwhile callout.

rashmigottipati:
@alvaroaleman, my bad I didn't realize that the PR would get automatically merged after it's approved. Will be mindful of that going forward. Thanks for mentioning it.

joelanford:
My bad! Perhaps we should configure the merge bot to [squash by default](https://github.com/kubernetes/test-infra/blob/cf6821b75f0d2ccd077066ff54b31972be9533d6/config/prow/config.yaml#L584)? I feel like I'm guilty of approving several PRs without using the label or asking for a squash because I'm used to other projects where that happens automatically.

PR #1486, opened 2021-04-20
:warning: Envtest refactor, support for 1.20+ API servers (secure serving)

### LINKED ISSUE #1357: envtest with Kubernetes 1.20

DirectXMan12:
This introduces (pluggable) authentication support to envtest, which is required for 1.20+ API servers. The default authentication provider is cert-based, since we already have TinyCA, and cert-based auth can provision users both before & after starting the API server.  Theoretically, other options should be available as well.  Correspondingly, RBAC is now turned on for authorization, but this should not impact existing users (see below).

We continue to provide support for the legacy helpers on controlplane to fetch a rest config & kubectl, but deprecate them in favor of methods to provision a new user.  The `RESTConfig` field on envtest, however, is not deprecated.
   
We also deprecate insecure serving, since it's going away anyway, and add some new fields to support this.  Theoretically, this should be largely transparent to end users, since I tried to backfill all legacy functionality with secure equivalents automatically (e.g. all "give me connection details" should now give an admin user, which should be transparent in terms of permissions).  ***However*** if any users are manually constructing a RESTConfig without using the envtest fields or helper methods (i.e. just grabbing the API server URL from envtest), this will break them, so it's technically a breaking change.

Lmk if you spot anything that looks more seriously breaking, but AFAICT by-and-large this should just be a transparent switch with a nice new benefit of "also you can now test controller RBAC".

Closes #1357 

Depends on #1541 
Depends on #1540


pkg/internal/testing/controlplane/plane.go
```diff
@@ -0,0 +1,241 @@
...
+func (u *AuthenticatedUser) Kubectl() (*KubeCtl, error) {
+	if u.kubectl != nil {
+		return u.kubectl, nil
+	}
+	if len(u.plane.APIServer.CertDir) == 0 {
+		panic("the API server has not yet been started, please do that before accessing connection details")
+	}
+
+	// cleaning this up is handled when our tmpDir is deleted
+	out, err := os.CreateTemp(u.plane.APIServer.CertDir, "*.kubecfg")
```
lobziik, line 195: Don't know if it's matter, but it strictly requires go 1.16. Some people might still use 1.15, i think

lobziik:
One thing i've noticed - control plane wont start again after stop because temp directory with certificates was removed within stop procedure and didn't create again. Throws panic like this:
```
panic: unable to start control plane itself: failed to start the controlplane. retried 5 times: unable to save the client certificate CA to /tmp/k8s_test_framework_269505290/client-cert-auth-ca.crt: open /tmp/k8s_test_framework_269505290/client-cert-auth-ca.crt: no such file or directory
```

I was solve it with crutch like this: https://github.com/kubernetes-sigs/controller-runtime/pull/1477/files#diff-c6321f65a2defa4f1d2a0810e6262d27a4a7e55c75db83330bf396a3148bc72aR109. Probably, it might be worked around more elegant way.

Also, suggest to get back such test, which seemd to be removed here: https://github.com/kubernetes-sigs/controller-runtime/pull/1486/files#diff-c4ae583bca86061bdfa2f60d76bdf50eba0f945fd1f3b56bea8edcb4e268ee0cL147 ( pkg/internal/testing/integration/internal/integration_tests/integration_test.go Line 147) :)

pkg/internal/testing/controlplane/plane.go
DirectXMan12: We're already on go 1.16 as off the bump to 1.21: ec2aeaf35e8d45c62e631f99ae0d30cfe509ba7b

DirectXMan12:
Good catch on the restart issue.  I'll take a look.  There's some weird interplay between "I want a manually specified certdir" & "I want to be able to restart and get a new temp dir".

DirectXMan12:
Actually, I think the extent of this should probably be discussed.  It's not clear to me that this is a use case we can safely support while keeping our existing semantics -- if we clean up the work dir, we invalidate all existing users that have been provisioned under most auth strategies. 

We can just recreate the work dir, but it's not clear to me that that's not gonna be surprising.  I'll put the functionality to do that back in for now, but I think we probably want to deprecate that bit of functionality and just say "if you want a new control plane instance, ask for one".

I wonder what the use case here is?

DirectXMan12:
Ok, I've fixed the "can't re-start a ControlPlane" issue and added a basic test.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -101,9 +147,51 @@ func (s *APIServer) Start() error {
...
+
+	// Insecure: URL, InsecureServing
+	if s.URL != nil {
+		s.InsecureServing = &process.ListenAddr{
+			Address: s.URL.Hostname(),
+			Port:    s.URL.Port(),
+		}
+	} else if insec := s.InsecureServing; insec != nil {
+		if insec.Port == "" || insec.Address == "" {
+			port, host, err := addr.Suggest("")
```
timebertt, line 193: Should we pass `InsecureServing.Address` down to `addr.Suggest`?
```suggestion
		if insec.Port == "" {
			port, host, err := addr.Suggest(insec.Address)
```

timebertt (commented):
Nice work, looking forward to using it :)
Left you a few comments inline, will give this a test spin later this week with the envtest suites in our project.

pkg/internal/testing/process/process.go
```diff
@@ -10,172 +11,188 @@ import (
...
 
 	select {
 	case <-ready:
 		ps.ready = true
 		return nil
+	case <-ps.waitDone:
+		if pollerStopCh != nil {
+			close(pollerStopCh)
+		}
+		return fmt.Errorf("timeout waiting for process %s to start", path.Base(ps.Path))
```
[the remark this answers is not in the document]
DirectXMan12: `timeout waiting for process %s to start succesfully (it may have failed to star
t, or stopped unexpectedly before becoming ready)` sound better?

pkg/internal/testing/process/arguments.go
```diff
@@ -27,3 +29,192 @@ func RenderTemplates(argTemplates []string, data interface{}) (args []string, er
...
+
+// Disable prevents this flag from be passed.
+func (a *Arguments) Disable(key string) *Arguments {
+	a.values[key] = DontPass
+	return a
+}
+
+// Append adds additional values to this flag.  If this flag has
+// yet to be set, initial values will include defaults.  If you want
+// to intentionally ignore defaults/start from scratch, call AppendOverridingDefaults.
```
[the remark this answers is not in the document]
DirectXMan12: good catch, changed the name of this a couple times during refactors :-)

pkg/internal/testing/process/arguments.go
```diff
@@ -3,11 +3,13 @@ package process
...
 	"bytes"
 	"html/template"
+	"sort"
+	"strings"
 )
 
 // RenderTemplates returns an []string to render the templates
 //
-// Deprecated: this should be removed when we remove APIServer.Args.
+// Deprecated: will be removed in favor of Arguments
```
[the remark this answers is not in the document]
DirectXMan12: Ah!  It's a bit less straightforward (wasn't actually certain how commonly people used the templates beyond the default args -- haven't seen a lot of it), and maybe I could make it more straightforward, but it was intended that you could do something like:

```go
// in your code
type CertDirArg struct {
  server *APIServer
}
func (a CertDirArg) Append(...string) Arg { return a } // ignore appends

func (a CertDirArg) Get(defaults []string) []string { 
  return []string{filepath.Join(a.server.CertDir, "blah")}
}

// then later

server.Configure().SetRaw("--proxy-client-key-file", CertDirArg{server})
```

Then, you can do ``

On option to make this easier would be to add a helper to CR:

```go
// in CR:
type FuncArg func(defaults []string) []string
func (a FuncArg) Append(...string) Arg { return a } // ignore appends
func (a FuncArg) Get(defaults []string) []string {
  return a(defaults)
}

// in your code

server.Configure().SetRaw("--proxy-client-key-file", envtest.FuncArg(func([]string) {
  return []string{filepath.Join(server.CertDir, "blah")} // theoretically works on windows, too!
})
```

Finally, if you're just after the cert dir, and are using the pre-existing client CA certs as a nice generated set of certs, you could wrap the auth to have your own certs instead:

```go
type WithProxyCA struct {
  // delegate most stuff to the other authn
  envtest.Authn

  CA envtest.TinyCA
  CertDir string
}

func (a *WithProxyCA) Configure(workDir string, args, *process.Arguments) error {
  a.CertDir = workDir
  args.Set("--proxy-client-cert-file", filepath.Join(a.CertDir, "proxy-client.crt"))
  // etc
  return nil
}
func (a *WithProxyCA) Start() error {
  // write stuff out to disk
  a.CA.DoTheStuffYouWant()
  
  // delegated to the other authn stuff
  return a.Other.Start()
}
```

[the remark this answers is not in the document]
DirectXMan12: Do any of these seem reasonable?

pkg/internal/testing/controlplane/apiserver.go
DirectXMan12: I was mostly intending on porting the suggestion logic as-is from what it is currently to avoid breaking things -- the old code never passed anything to `addr.Suggest` AFAICT

pkg/internal/testing/process/process.go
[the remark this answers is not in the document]
timebertt: Sounds good 👍

pkg/internal/testing/process/arguments.go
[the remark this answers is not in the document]
timebertt: Yes, good point. `SetRaw` would still support such use cases.

You're right, that's probably not straightforward / obvious how to do it (at least I didn't recognize it directly). I like the `FuncArg` idea. I think, if you add this (with a nice example like the above), it would already make things a bit more obvious. 👍 

> Finally, if you're just after the cert dir, and are using the pre-existing client CA certs as a nice generated set of certs, you could wrap the auth to have your own certs instead:

Yeah, that's actually what we're doing there. However, my comment was indeed rather about the general case and not tied to the special one :)

pkg/internal/testing/controlplane/apiserver.go
timebertt: Yep, also checked how it was done before. This was just a suggestion on top, nothing critical.

examples/scratch-env/main.go
```diff
@@ -68,39 +61,47 @@ func runMain() int {
...
-	kubeconfigFile, err := ioutil.TempFile("", "scratch-env-kubeconfig-")
+	// NB(directxman12): this group is unfortunately named, but various
+	// kubernetes versions require us to use it to get "admin" access.
+	user, err := env.ControlPlane.AddUser(envtest.User{
+		Name:   "envtest-admin",
+		Groups: []string{"system:masters"},
+	}, nil)
 	if err != nil {
-		log.Error(err, "unable to create kubeconfig file, continuing on without it")
+		log.Error(err, "unable to provision admin user, continuing on without it")
```
alvaroaleman, line 81: This effectively makes it unusable, shouldn't we return an error (perhaps a well known one in case ppl want to check for it?)

pkg/manager/manager_suite_test.go
```diff
@@ -53,8 +53,14 @@ var _ = BeforeSuite(func(done Done) {
...
 	Expect(err).NotTo(HaveOccurred())
 
-	clientTransport = &http.Transport{}
-	cfg.Transport = clientTransport
+	cfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
+		// NB(directxman12): we can't set Transport *and* use TLS options,
+		// so we grab the transport right after it gets created so that we can
+		// type-assert on it (hopefully)?
+		// hopefully this doesn't break 🤞
+		clientTransport = rt.(*http.Transport)
```
alvaroaleman, line 61: What about intead of asserting here just changing `var clientTransport *http.Transport` to `http.RoundTripper` and doing the type assertion when we close the keep-alives? That would make this easier to debug if something fails (right now we'd just get a NPD, right?)

pkg/envtest/server.go
```diff
@@ -46,54 +46,66 @@ It's possible to override some defaults, by setting the following environment va
 
 */
 const (
-	envUseExistingCluster  = "USE_EXISTING_CLUSTER"
-	envKubeAPIServerBin    = "TEST_ASSET_KUBE_APISERVER"
```
alvaroaleman, line 50: Did you mean to entirely remove these? That makes this PR a breaking change, in a subtile way. They are also still documented in `pkg/envtest/server.go`

pkg/internal/testing/certs/tinyca_test.go
```diff
@@ -0,0 +1,254 @@
...
+			Expect(ca.CA.Cert).NotTo(BeNil(), "should have a cert")
+		})
+
+		It("should be usable for signing & verifying", func() {
+			Expect(ca.CA.Cert.KeyUsage&x509.KeyUsageCertSign).NotTo(Equal(0), "should be usable for cert signing")
+			Expect(ca.CA.Cert.KeyUsage&x509.KeyUsageDigitalSignature).NotTo(Equal(0), "should be usable for signature verifying")
+		})
+	})
+
+	It("should produce unique serials among all generated certificates of all types", func() {
```
alvaroaleman, line 55: Out of curiosity, does this have any practical relevance? I know the RFC requires it, but is it used for anything other than CRLs?

pkg/internal/testing/certs/tinyca_test.go
```diff
@@ -0,0 +1,254 @@
...
+			})
+
+			It("should add the corresponding IP address(es) (as IP SANs) for DNS names", func() {
+				// NB(directxman12): we currently fail if the lookup fails.
+				// I'm not certain this is the best idea (both the bailing on
+				// error and the actual idea), so if this causes issues, you
+				// might want to reconsider.
+
+				localhostAddrs, err := net.LookupHost("localhost")
+				Expect(err).NotTo(HaveOccurred(), "should be able to find IPs for localhost")
```
alvaroaleman, line 121: Are you sure you want to do this lookup? I have seen it returning very surprising results at times :P

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+		"secure-port":               []string{s.SecureServing.Port},
+		// NB(directxman12): previously we didn't set the bind address for the secure
+		// port.  It *shouldn't* make a difference unless people are doing something really
+		// funky, but if you start to get bug reports look here ;-)
+		"bind-address": []string{s.SecureServing.Address},
+
+		// required on 1.20+, fine to leave on for <1.20
+		"service-account-issuer":           []string{s.SecureServing.URL("https", "/").String()},
+		"service-account-key-file":         []string{filepath.Join(s.CertDir, saCertFile)},
+		"service-account-signing-key-file": []string{filepath.Join(s.CertDir, saKeyFile)},
```
alvaroaleman, line 331: Why does the Apiserver needs this if the CM creates the SAs?

pkg/internal/testing/controlplane/kubectl.go
```diff
@@ -0,0 +1,108 @@
...
+	kubeConfig := kcapi.NewConfig()
+	protocol := "https"
+	if !rest.IsConfigTransportTLS(*cfg) {
+		protocol = "http"
+	}
+	kubeConfig.Clusters["envtest"] = &kcapi.Cluster{
+		// TODO(directxman12): if client-go ever decides to expose defaultServerUrlFor(config),
+		// we can just use that.  Note that this is not the same as the public DefaultServerURL,
+		// which requires us to pass a bunch of stuff in manually.
+		Server:                   (&url.URL{Scheme: protocol, Host: cfg.Host, Path: cfg.APIPath}).String(),
```
timebertt, line 48: `cfg.Host` is set to something like `https://127.0.0.1:52197/`, so this will actually result in a kubeconfig containing something like this:
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t...
    server: https://https:%2F%2F127.0.0.1:52197%2F
  name: envtest
```

pkg/internal/testing/process/arguments.go
```diff
@@ -0,0 +1,313 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package process
+
+import (
+	"bytes"
+	"html/template"
```
timebertt, line 21: I assume, this is should actually use `text/template`, but uses `html/template` by accident (was already the case before this PR).
This might result in some undesired encoding of special characters command line arguments. I was bitten by this one time, but I actually don't remember what character was encoded.
So just leaving this side-note here, in case you wanna change it :)

pkg/envtest/server.go
```diff
@@ -46,54 +46,66 @@ It's possible to override some defaults, by setting the following environment va
...
+	// EmptyArguments constructs a new set of flags with nothing set.
+	//
+	// This is mostly useful for testing helper methods -- you'll want to call
+	// Configure on the APIServer (or etcd) to configure their arguments.
+	EmptyArguments = process.EmptyArguments()
+)
 
 // Environment creates a Kubernetes test environment that will start / stop the Kubernetes control plane and
 // install extension APIs
 type Environment struct {
```
alvaroaleman, line 106: How about adding a method `GetConfigFor(username string, groups ...string)(*rest.Config, error`)? Right now this is somewhat hidden in `Environent.Controlplane.AddUser`

timebertt (commented):
I gave this a test spin now in our envtest suites and found a few issues, you might wanna address (mainly to make it more backwards-compatible and easier to upgrade to for users).
Also noticed some nits along the way.

Apart from the mentioned issues, everything worked pretty well! :)

alvaroaleman (commented):
Initial, somewhat cursory pass. This change is simply too big for me to be able to properly review it :upside_down_face:

DirectXMan12:
> This change is simply too big for me to be able to properly review it

@alvaroaleman I think what I'm gonna do here is finish addressing the comments from @timebertt and you, and then break this up into 3-4 smallers PRs, which should be more digestable.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}
+	return s.Authn.Stop()
+}
+
+// APIServerDefaultArgs exposes the default args for the APIServer so that you
+// can use those to append your own additional arguments.
+//
+// Deprecated: use APIServer.Configure()
+var APIServerDefaultArgs = []string{
```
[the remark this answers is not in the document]
DirectXMan12: hmm, the intention was that it should auto-adjust to the old flag passing mechanism -- there's a special adapter that's called internally.  Maybe something broke there?  Weird, let me double-check.

examples/scratch-env/main.go
DirectXMan12: ah, yeah, I think that was an artifact of the way that control flow worked before.  Good point :+1:

pkg/envtest/server.go
DirectXMan12: they're still used, just lower-down in the code.  These constants are private, so it's not breaking to remove them here -- they're actually just duplicates from `testing/internal` from when we couldn't tweak testing_frameworks directly

pkg/internal/testing/controlplane/kubectl.go
estroz: Is that always true?

pkg/envtest/server.go
```diff
@@ -168,27 +174,6 @@ func (te *Environment) Stop() error {
 	return te.ControlPlane.Stop()
 }
 
-// getAPIServerFlags returns flags to be used with the Kubernetes API server.
-// it returns empty slice for api server defined defaults to be applied if no args specified
-func (te Environment) getAPIServerFlags() []string {
-	// Set default API server flags if not set.
-	if len(te.KubeAPIServerFlags) == 0 {
```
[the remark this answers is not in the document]
DirectXMan12: ah whoops, it was not intended to remove these, just deprecate them -- there's fallback logic that's supposed to switch to these if they're overridden.  Good catch.

pkg/envtest/server.go
DirectXMan12: (FWIW, they're now used natively in the binpath finder, which we were just short-circuiting before because we couldn't make it do what we wanted it to)

pkg/envtest/server.go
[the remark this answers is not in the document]
DirectXMan12: should be fixed as soon as I push a new batch of code up

pkg/internal/testing/certs/tinyca_test.go
DirectXMan12: I don't actually remember.  I remember thinking when I wrote TinyCA, "hmm, I should do this semi-correctly and generate unique serials at least", but I don't remember exactly why I thought that

pkg/internal/testing/certs/tinyca_test.go
DirectXMan12: we actually do this currently within the code (perhaps unfortunately, depending on how you're looking at it), so it's there for compat reasons.  This just adds a test for that behavior.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+	//
+	// If this is not specified, we default to a random free port on localhost.
+	//
+	// Deprecated: use InsecureServing (for the insecure URL) or SecureServing, ideally.
+	URL *url.URL
+
+	// SecurePort is the additional secure port that the APIServer should listen on.
+	//
+	// Deprecated: use SecureServing.
+	SecurePort int
```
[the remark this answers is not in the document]
DirectXMan12: ah, whoops, again, another thing that must've gotten dropped during a refactor :-(.  This was not supposed to be ignored, especially not silently.

pkg/internal/testing/controlplane/apiserver.go
DirectXMan12: You know, I'm not sure, but it does :woman_shrugging: I'll look it up.

pkg/envtest/server.go
DirectXMan12: :+1:

DirectXMan12:
I've split the refactor of testing and the introduction of structured arguments into two separate PRs.

DirectXMan12:
Gotta rebase juggle now.

DirectXMan12:
Ok, I split the license change out into a separate PR to hopefully reduce the noise on this one as well a bit.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+// APIServerDefaultArgs exposes the default args for the APIServer so that you
+// can use those to append your own additional arguments.
+//
+// Deprecated: use APIServer.Configure()
+var APIServerDefaultArgs = []string{
+	"--advertise-address=127.0.0.1",
+	"--etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}",
+	"--cert-dir={{ .CertDir }}",
+	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
+	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
```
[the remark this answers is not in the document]
DirectXMan12: I'll add a note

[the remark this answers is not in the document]
DirectXMan12: I went back and forth on this since these are deprecated anyway, but this minimal fix is easy enough.  This won't work with the newest API servers though.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
```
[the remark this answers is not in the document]
DirectXMan12: I was just keeping these the same as before -- we default SecurePort for compatibility, and these'll go away at the same time as `.SecurePort`, so it should be fine.

pkg/internal/testing/process/arguments.go
DirectXMan12: keeping it the same for now, since we'll be deprecating this part eventually anway.

pkg/internal/testing/controlplane/kubectl.go
DirectXMan12: host should never have a scheme.  Let me double check -- I didn't observe that behavior in my testing.

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	return s.Authn.Stop()
+}
+
+// APIServerDefaultArgs exposes the default args for the APIServer so that you
+// can use those to append your own additional arguments.
+//
+// Deprecated: use APIServer.Configure()
+var APIServerDefaultArgs = []string{
+	"--advertise-address=127.0.0.1",
```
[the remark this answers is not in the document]
DirectXMan12: hmm... yeah.  I was trying not to mess up people that were intentionally using the insecure port with legacy args though by enabling RBAC.  Let me think about this one.

pkg/internal/testing/controlplane/apiserver.go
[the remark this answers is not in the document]
DirectXMan12: ah, fixed it -- we missing passing in args at the top level

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -0,0 +1,452 @@
...
+	// on the flags, and hope that we've "parsed" them properly.
+	//
+	// TODO(directxman12): once we support 1.20 as the min version (might be when 1.24 comes out,
+	// might be around 1.25 or 1.26), remove this logic and the corresponding line in API server's
+	// default args.
+	if err := s.discoverFlags(); err != nil {
+		return err
+	}
+
+	s.processState.Args, err = process.TemplateAndArguments(s.Args, s.Configure(), process.TemplateDefaults{
```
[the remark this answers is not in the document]
DirectXMan12: Think I know why this is, will double check to confirm (legacy args get converted to non-legacy args during processing, but the legacy args stay in place)

pkg/internal/testing/controlplane/apiserver.go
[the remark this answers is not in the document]
DirectXMan12: made it so that we enable RBAC if the user doesn't enable something else.

pkg/internal/testing/controlplane/kubectl.go
DirectXMan12: ah, hmm, I see it now.  Let me fix that :-)

examples/scratch-env/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	github.com/spf13/pflag v1.0.5
-	k8s.io/client-go v0.19.2
+	k8s.io/client-go v0.21.0-beta.1
```
vincepri, line 7: v0.21.1?

pkg/manager/manager_suite_test.go
DirectXMan12: meh, then we have more type assertions, though :-/

I can def put a better panic message in though, which should help there.

vincepri (approved):
/lgtm

Issue #1234, opened 2020-10-31
Install and looking for envtest bins in the $GOBIN/ by default


camilamacedo86:
**Description**

Currently, the default path where controller-runtime is looking for the binaries which are required for the env tests is `"/usr/local/kubebuilder/bin"`. See: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/server.go#L57

This issue is to purpose the idea of:
- Change the default path for $GOBIN/ (e.g $GOBIN/kubectl-1.15.6) 
- Change the scripts to looking for and use the $GOBIN instead of this path. See: https://github.com/kubernetes-sigs/controller-runtime/blob/master/hack/setup-envtest.sh

**Motivation**

- Do not require sudo permissions to install the required binaries 
- Simplifies its installation and usage 
- Issue raised in Kubebuilder: https://github.com/kubernetes-sigs/kubebuilder/issues/1732

estroz:
Expanding on this: it would be nice to have
- a set of simple loading rules (like with kubeconfigs)
- a helper to install envtest binaries to one of those locations
- a helper to compare binary versions

camilamacedo86:
We might consider addressing these requirements toghether with https://github.com/kubernetes-sigs/controller-runtime/issues/1300.

DirectXMan12:
I think this is likely to be subsumed by #1488, closing for now

PR #1488, opened 2021-04-21
:sparkles: EnvTest Binaries Setup Tool

### LINKED ISSUE #1300: Replace the shell script to install the env test binaries by a go model in the tool
#### Describe the problem you need a feature to resolve.

Currently, in Kubebuilder where the binaries are used we have : 

```shell
# Run tests
ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
test: generate fmt vet manifests
	mkdir -p ${ENVTEST_ASSETS_DIR}
	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/{{ .ControllerRuntimeVersion }}/hack/setup-envtest.sh
	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out

```

Note that it is NOT intuitive for the users. As I user of Kubebuilder, I'd expected that the tool provides a much more readable, intuitive and understandable code to reduce the learning curve and keep its maintainability such as: 

```shell
# Run tests
test: generate fmt vet manifests setup 
	go test ./... -coverprofile cover.out

# Setup the envtest setup local envinroment 
setup:
	sigs.k8s.io/controller-tools@versio; go run envtest_setup.go 
```

However, since was decide that the tool/module should be in the controller-runtime and not Kubebuilder then, it means that the script `/hack/setup-envtest.sh` would need to be replaced by a go module maybe from its `/hack/tools`.
   
#### Describe the solution you'd like.

Note that here we also might nat to consider address together the requirements described in https://github.com/kubernetes-sigs/controller-runtime/issues/1234. However, in order to help who will work on this task, following the module that was done in the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1711.  It is very closer to the code that should be done here. 

```go
package main

import (
	"fmt"
	"io"
	"net/http"
	"os"
	"os/exec"
	"path/filepath"
	"strings"

	"k8s.io/klog"
)

const (
	// defaultKubebuilderPath is the fixed value defined in the controller-runtime to looking for the
	// binaries.
	defaultKubebuilderPath = "/usr/local/kubebuilder/bin" // see that needs to be changed https://github.com/kubernetes-sigs/controller-runtime/issues/1234

	// kubernetesVersion defines the version used to download the binaries
	kubernetesVersion = "1.16.4"
)

func main() {
	klog.Info("Checking EnvTest binaries")
	assetsEnv := os.Getenv("KUBEBUILDER_ASSETS")
	if strings.TrimSpace(assetsEnv) != "" {
		klog.Infof("EnvTest binaries configured via KUBEBUILDER_ASSETS: %s", assetsEnv)
		os.Exit(0)
	}

	binPath := filepath.Join("..", "bin")
	if hasBinaries(binPath) {
		klog.Info("EnvTest binaries found in bin/")
		os.Exit(0)
	}

	klog.Infof("Downloading EnvTest tools")
	envtest_tools_archive_name := getEnvToolsArchiveName()
	envtest_tools_download_url := "https://storage.googleapis.com/kubebuilder-tools/" + envtest_tools_archive_name

	err := downloadFile(envtest_tools_archive_name, envtest_tools_download_url)
	if err != nil {
		klog.Fatalf("unable to download the file (%v) from (%v): %v", envtest_tools_archive_name, envtest_tools_download_url, err)

	}

	if _, err := os.Stat(binPath); os.IsNotExist(err) {
		klog.Infof("Creating bin/ directory")
		err = os.Mkdir(binPath, 0755)
		if err != nil {
			klog.Fatalf("error to create the bin/ directory : %s", err)
		}
	}

	cmd := exec.Command("tar", "-C", binPath, "--strip-components=2", "-zvxf", envtest_tools_archive_name)
	if err := cmd.Run(); err != nil {
		klog.Fatalf("error to untar bin %v: %v", filepath.Join(binPath, envtest_tools_archive_name), err)
	}

	cmd = exec.Command("rm", "-rf", envtest_tools_archive_name)
	if err := cmd.Run(); err != nil {
		klog.Fatal(err)
	}
	klog.Infof("EnvTest binaries are in the directory %v", binPath)
}

// getEnvToolsArchiveName will return the name of the env tools archive based or the arch and so
func getEnvToolsArchiveName() string {
	cmd := exec.Command("go", "env", "GOARCH")
	goarch, err := cmd.CombinedOutput()
	if err != nil {
		klog.Fatal(err)
	}

	cmd = exec.Command("go", "env", "GOOS")
	goos, err := cmd.CombinedOutput()
	if err != nil {
		klog.Fatal(err)
	}
	return fmt.Sprintf("kubebuilder-tools-%s-%s-%s.tar.gz", kubernetesVersion, strings.TrimSpace(string(goos)), strings.TrimSpace(string(goarch)))
}

// hasBinaries will return true when the envtest binaries are found
func hasBinaries(path string) bool {
	if _, err := os.Stat(path); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "etcd")); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "kube-apiserver")); os.IsNotExist(err) {
		return false
	}
	if _, err := os.Stat(filepath.Join(path, "kubectl")); os.IsNotExist(err) {
		return false
	}
	return true
}

// downloadFile will download a url to a local file.
func downloadFile(filepath string, url string) error {
	// Get the data
	resp, err := http.Get(url)
	if err != nil {
		return err
	}
	defer resp.Body.Close()

	// Create the file
	out, err := os.Create(filepath)
	if err != nil {
		return err
	}
	defer out.Close()

	// Write the body to file
	_, err = io.Copy(out, resp.Body)
	return err
}
```

DirectXMan12:
This introduces a tool (in Go) to manage installed versions of the envtest binaries.
It can download new ones (both always and if necessary), enumerate installed and available versions, and remove versions.

By default, it installs to an OS-specific cache dir (as per OS conventions), but that can be customized if need-be.

The tool itself is in a separate module, so it can just be `go installed`-ed on 1.16 w/o pulling in extra deps.

The README in the subdirectory and the command's `--help` flag have more details, but the TL;DR of usage is:

```shell
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
source <(setup-envtest fetch -k '1.19.*' -p env)
# or, for the latest
source <(setup-envtest fetch -p env)
```

Fixes #1300


DirectXMan12:
Quick FAQ on downloading and "latest":

- `fetch` is lazy when given a non-latest version (e.g. `-k '1.19.*' or `-k 1.19.2`) -- if you already have something on disk that matches, it won't make any network connections
- `fetch` can be told to fetch the latest in general (`-k latest`), or of a particular subversion (`-k 1.20.+`), in which case it'll query GCS
- to find available versions, use the `list` command, to find installed ones use either `list` or `list-installed`

DirectXMan12:
Other notes: this currently writes the archive to a temp file then reads that back, and clean it up at the end.. Not strictly necessary -- we could stream directly to the gzip -> tar readers, but it theoretically could make debugging easier. :woman_shrugging:

tools/setup-envtest/versions/types.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+}
+
+// Version is an X.Y.Z version.
+type Version struct {
+	Major, Minor, Patch PointVersion
+}
+
+// IsWildcard checks if any part of this version is a wildcard.
+func (v Version) IsWildcard() bool {
```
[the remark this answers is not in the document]
DirectXMan12: I wanted a quick way to distinguish "get me the latest" from "get me anything from disk if I have it, otherwise download", but you may be right that it's easier.  I'll play around a bit and see if I can find something that works nicely

DirectXMan12:
refactoring a bit too enable no-internet support, and to combine commands, etc

tools/setup-envtest/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+For example, on Linux this is `$XDG_CACHE_DIR`, or just `~/.config` if
+that's unset.
+
+There's an overall folder that holds all files, and inside that is
+a folder for each version/platform pair.  The exact directory structure is
+not guarnateed, except that the leaf directory will contain the names
+expected by envtest.  You should always use `setup-envtest fetch` or
+`setup-envtest switch` (generally with the `-p path` or `-p env` flags) to
+get the directory that you should use.
```
camilamacedo86, line 47: Would not first be required to move forward with https://github.com/kubernetes-sigs/controller-runtime/issues/1234?

Is this one working if: 
- We remove the bins from: /usr/local/kubebuilder/bin
- Then, use it to install the binaries in the bin/install as we do in KB. 

Note that currently in KB side we can just run the tests with make test and not directly because of it. (https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L56)

tools/setup-envtest/remote/client.go
```diff
@@ -0,0 +1,217 @@
...
+	// Log allows us to log.
+	Log logr.Logger
+}
+
+// ListVersions lists all available tools versions in the given bucket, along
+// with supported os/arch combos and the corresponding hash.
+//
+// The results are sorted with newer versions first.
+func (c *Client) ListVersions(ctx context.Context) ([]versions.Set, error) {
+	loc := &url.URL{
```
estroz, line 59: Can you not use the [google storage API](https://pkg.go.dev/cloud.google.com/go/storage) here?

DirectXMan12: didn't want to pull in a whole bunch of deps for basically one API call, plus testing is a bit easier.

DirectXMan12: I can switch it out if you want, but it's not really much code.

tools/setup-envtest/README.md
DirectXMan12: This is an alternate approach to installing in GOBIN.  We could, for instance, follow up with something along those lines where we have a well-known symlink in the store directory, and have `use` or `switch` change that instead of having you fiddle with the environment variable, but I figured this was a decent first step that's backwards compatible (all CR versions support using `$KUBEBUILDER_ASSETS`)

DirectXMan12:
Added some tests, made the version selector more semver like:

* `1.6.3`: exact
* `1.6.x` or `1.6.*` or just `1.6`: any 1.6 (copied from a couple of the existing semver libraries)
* `~1.6.3`: usual meaning (pin on 1.6, 1.6.3 and up)
* `<1.17`: older than 1.17 (useful for cleanup, mainly)
* `1.6.x!`: no really, get the latest remote version
* `1.22.0?`: include prerelease versions when searching

We don't use an existing semver library because one of them (blang) is missing support for '~' (maybe not strictly needed?) and the other (Masterminds) has no easy way to construct a version from parts except by round tripping it through a string.  We need to keep our own version type because blang's version isnt hashable, and mastermind's is technically hashable but is opaque so it's not guaranteed to stay that way.

tools/setup-envtest/versions/types.go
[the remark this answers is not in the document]
DirectXMan12: also, FWIW: all semver selectors are bespoke as far as I can tell -- there doesn't appear to be a spec for the selectors, just the version itself.

DirectXMan12:
I might simplify the selector logic down a bit closer to what you suggested above tomorrow -- right now we handle a lot of cases that don't currently apply (they could in the future, which is why I added them, but :woman_shrugging: )

tools/setup-envtest/remote/client.go
estroz: Nah that's a fair point.

DirectXMan12:
Simplified the version code a bit to remove a bunch of cases that we don't need atm, switched the version to be specified with a positional argument instead of a flag (e.g. `use 1.17` instead of `use -k 1.17`)

hack/check-everything.sh
```diff
@@ -20,16 +20,27 @@ set -o pipefail
...
+(
+    # don't presume to install for the user
+    cd ${hack_dir}/../tools/setup-envtest
+    GOBIN=${tmp_bin} go install .
+)
+source <(${tmp_bin}/setup-envtest use --use-env=always -p env ${ENVTEST_K8S_VERSION})
+
+# link the assets into integration
+for tool in kube-apiserver etcd kubectl; do
+    ln -f -s "${KUBEBUILDER_ASSETS:?unable find envtest assets}/${tool}" "${hack_dir}/../pkg/internal/testing/integration/assets/bin/${tool}"
```
estroz, line 42: Shouldn't this respect the default `/usr/local/kubebuilder/bin`?

tools/setup-envtest/main.go
```diff
@@ -0,0 +1,292 @@
...
+	"sigs.k8s.io/controller-runtime/tools/setup-envtest/remote"
+	"sigs.k8s.io/controller-runtime/tools/setup-envtest/versions"
+	"sigs.k8s.io/controller-runtime/tools/setup-envtest/workflows"
+)
+
+
+const (
+	// envNoDownload is an env variable that can be set to always force
+	// the --installed-only, -i flag to be set.
+	envNoDownload = "ENVTEST_INSTALLED_ONLY"
```
estroz, line 29: Thoughts on `CONTROLLER_RUNTIME_ENVTEST_` as a prefix instead of `ENVTEST_`? Eventually it would be nice to add similarly prefixed variables to the set for specifying envtest config options, so it would also be nice to have all envtest-related config variables aligned in name. I guess the tool's config != envtest config, so maybe `CONTROLLER_RUNTIME_ENVTEST_TOOL_`.

Corrollary: remove variables prefixed with `KUBEBUILDER_` prior to 1.0 since envtest is more general than kubebuilder.

tools/setup-envtest/main.go
```diff
@@ -0,0 +1,292 @@
...
+	// that it's the last one executed.
+	defer envp.HandleExitWithCode()
+
+	// set up flags
+	flag.Usage = func() {
+		name := os.Args[0]
+		fmt.Fprintf(os.Stderr, "Usage: %s [FLAGS] use|list|cleanup|sideload [VERSION]\n", name)
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr,
+			`
```
estroz, line 146: WDYT about making this tool `alpha` for now so we can get feedback on usage before defining a stable CLI?
```suggestion
			`
**Note:** this command is currently alpha, and it's CLI may change.

```

estroz:
I was kinda hoping to keep the tool pretty simple off the bat, and add features as needed/requested. Having simple use/uninstall commands that optionally respect `KUBEBUILDER_ASSETS`, and follow semver with some pattern matching is an MVP that covers the cases I'd expect most users to follow:
1. Use, and optionally install, some version pattern globally or from a local directory.
2. Uninstall some version pattern globally or from a local directory.

I can see the above modes of `setup-envtest` being used in the 2 environment types below (assuming `setup-envtest` was installed by `go get sigs.k8s.io/controller-runtime/tools/setup-envtest`). The current implementation does these things, but what I describe below is simpler IMO.

### Development locally

In this mode, both global and local modes are relevant, and being able to switch versions is important.

#### Use/install

A user could run the following to install and use bins at the default `usr/local/kubebuilder/bin`:

```sh
setup-envtest use <1.17
go test ./...
```

To set a custom directory:

```sh
export KUBEBUILDER_ASSETS=/foo/bin
setup-envtest use <1.17
go test ./...
```

Maybe add a simple `--export` flag, and `--assets-dir` to override `KUBEBUILDER_ASSETS`:

```sh
source <(setup-envtest use <1.17 --assets-dir=./testbin --export)
go test ./...
```

I'd also expose `--no-install` to force `setup-envtest` to fail if no bins are already installed. In this case, running `setup-envtest use --no-install <version>` without `--export` would be a no-op and should result in an error. 

#### Switching

Switching should be as simple as

```sh
source <(setup-envtest use 1.18.3 --assets-dir=./otherbin --export)
go test ./...
```

#### Uninstall

Globally:

```sh
setup-envtest uninstall 1.17.*
```

Locally:

```sh
setup-envtest uninstall --assets-dir=./testbin 1.17.*
```

### In CI/CD

Installation is really the only thing that matters, and should be dead simple.

#### Use/install

Install one version for a particular CI environment/run:

```sh
export KUBEBUILDER_ASSETS=/tmp/testbin # Probably set by CI config, or not at all and use default.
setup-envtest use 1.17.3
go test ./...
```

---

>EDIT: I realize I mis-read what you were suggesting and I guess technically made this more complicated, but at least it's a bit more familiar? And the < selector is nice for cleanup.

Specifying version specifiers (the current implementation looks close to those used by [python](https://www.python.org/dev/peps/pep-0440/#version-specifiers)) is fine with me.

DirectXMan12:
## Inline responses

> A user could run the following to install and use bins at the default usr/local/kubebuilder/bin

`/usr/local` is frequently not writable by normal users.  This causes issues on initial installation.  I had intended to do a follow-up PR with a suggested local symlink instead and tweaking CR to read that, esp since folks have mentioned the /usr/local location as an issue before.

> To set a custom directory:

Are you suggesting we clobber whatever's stored at `KUBEBUILDER_ASSETS` here?  That seems like a pretty significant behavioral difference from what the current scripts so (which is "use KUBEBUILDER_ASSETS if it's set, otherwise use a new thing"), and I kinda wanted to make switching over pretty straightforward

> Switching should be as simple as

Just to be clear: the current prototype has this pretty much as `source <(setup-envtest use 1.16 -p env)`, is there something you didn't like about that, was it not obvious, or were you just stating the full vision you had here (just want to make sure I'm not missing something).

## Overall

So, I'm definitely not opposed to having a more global-centric workflow with this, which is what it looks like you're target at.  I had intended to follow up with that, with the first PR being minimally destructive to the user's existing FS & setup.

Basically, I had three constraints in mind:

1. /usr/local is not writable generally
2. If someone is just setting `KUBEBUILDER_ASSETS` manually, I don't want to disrupt them suddenly
3. I want to be able to bounce between multiple versions quickly, running them all at once (idk how common this is, but I frequently find myself needing to run experiments where I test behavior against 2 or more API server versions.  Technically they don't have to be in parallel, but it's nice just having two different shells open with different versions)

For "global" swapping, I had envisioned adding an search path option (after KUBEBUILDER_ASSETS) that was "$STORE_DIR/current`.  Then, `setup-envtest` could just make that a symlink and swap it around with something like `setup-envtest use` or `setup-envtest switch`.

If we had that, making `use` swap that by default instead of just printing out env/path/info, and putting the current behavior behind a flag or different command would be fine by me.  Does that match up with what you were thinking?

DirectXMan12:
Walked through this a bit with @estroz the other day, and I think it's probably worth it for me to write down some of the design considerations that I made.

## Design Inspiration & Background

The design here is loosely inspired by tools like GVM, RVM, and rustup.

The basic premise behind each of these tools is that you have one or more projects installed that need to use one or more versions of X, and you don't want to have to manually manage the different versions, or have 17 copies of X laying around your various project folders.

Instead, you have a global store of different versions of X, and then you {set the PATH/point env vars in the right place/update a symlink/have a wrapper script}.

This gives you:

-  a single place to store all of your different versions (so you don't have to redownload)
- a way to use multiple versions across different projects
- a way to swap quickly between versions

## What, specifically, we do in setup-envtest

setup-envtest follows the "set environment variables" pattern mentioned above.  We store all of our downloaded versions in a data dir, and the return a path to the version you ask for that you can set KUBEBUILDER_ASSETS to (or with `-p env` we just return the full shell line that you can `source`).

## Why don't we...

- **have a global version too?**
  
  It's not really super useful for our cases.  Generally, you don't want to stop on existing uses (e.g. you might be running a project against two different versions to figure out the differences that are causing a bug). 

  This is particularly true from shell scripts & make files, which *must* use project-specific settings for this reason.  As a normal user (if you're just playing around), you probably still want this if you're working with a couple of projects, and if not, it's not too hard to just say `souce <(setup-envtest -p env XYZ)` instead of just `setup-envtest XYZ` (but there's an easy shell wrapper that removes that part too -- see the README)

- **install to KUBEBUILDER_ASSETS?**

  1. We don't install binaries there because we don't want many copies of the same binaries floating around our system, and we want to make it quick to swap back and forth
  2. We don't set up symlinks there because we want to preserve the existing semantics where KUBEBUILDER_ASSETS acts as an *override* to change what shell scripts are doing.  If my makefile normally uses `setup-envtest --use-env 1.16` and I want to test with 1.19, it should be possible to just do `source <(setup-envtest -p env 1.19)` and have the override kick in without needing to modify my Makefile or setup script.

- **use KUBEBUILDER_ASSETS by default instead of requiring `--use-env`**
 
  I see this as a tradeoff between "interactive UX" and "script UX" -- that is: if I'm a user writing `setup-envtest 1.19` on the shell, my intent is most likely "I want to use 1.19".  However, if I'm in a script, I most always want "maybe give me 1.19, unless overrides are set".  Since scripts are typed once-ish, but shell commands are typed frequently, it seems more reasonable to me to optimize around the interactive user's convenience and desired behavior, and make the shell scripter type a few more characters.

hack/check-everything.sh
estroz: Resolving this, as `/usr/local` is often not writeable by non-superusers.

tools/setup-envtest/env/env.go
```diff
@@ -0,0 +1,473 @@
...
+
+	// don't bother checking the version, the user explicitly asked us to use this
+	// we don't know the version, so set it to wildcard
+	e.Version = versions.AnyVersion
+	e.Platform.OS = "*"
+	e.Platform.Arch = "*"
+	e.manualPath = value
+	return true
+
+	return false
```
estroz, line 452:
```suggestion
```

tools/setup-envtest/main.go
```diff
@@ -0,0 +1,273 @@
...
+		Store: store.NewAt(*binDir),
+		Out: os.Stdout,
+	}
+
+	switch version {
+	case "", "latest":
+		env.Version = versions.LatestVersion
+	case "latest-on-disk":
+		env.Version = versions.AnyVersion
+		env.NoDownload = true
```
estroz, line 106: Will this combination of settings match the first version found on disk, or the latest version on disk?

tools/setup-envtest/main.go
DirectXMan12: I'm kinda :woman_shrugging: on variable names, except that I don't like holding down the shift key much :-P.

Seriously though, maybe just ENVTEST_TOOL_XYZ?

tools/setup-envtest/main.go
DirectXMan12: latest -- we always sort by latest so that we give you the latest on disk & get deterministic behavior.

DirectXMan12: added a comment to this effect

estroz (approved):
/lgtm

Issue #1463, opened 2021-03-31
options.AndFrom overriding previously set value


criscola:
Hello everyone, I think I've found a small bug. Here's my code:

```
var err error
ctrlConfig := operatorconfigv1.OperatorConfig{}
options := ctrl.Options{
	Scheme:                 scheme,
	MetricsBindAddress:     viper.GetString(MetricsBindAddressKey),
	Port:                   viper.GetInt(WebhookPortKey),
	HealthProbeBindAddress: viper.GetString(HealthProbeBindAddressKey),
	LeaderElection:        	viper.GetBool(LeaderElectEnableKey),
}

// Build and pass the configuration file to the controller.
if cfgFile := viper.GetString(LoadConfigKey); cfgFile != "" {
	options, err = options.AndFrom(ctrl.ConfigFile().AtPath(cfgFile).OfKind(&ctrlConfig))
	if err != nil {
		setupLog.Error(err, "unable to load the config file into the controller")
		os.Exit(1)
	}
}

// Initialize manager
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), options)
if err != nil {
	setupLog.Error(err, "unable to initialize manager")
	os.Exit(1)
}
```

What I'm expecting is that `options.AndFrom` doesn't override _any_ previously set value. The problem is that the config file specified through `AtPath(cfgFile)` is overriding the field `LeaderElection` even though it is was set previously through a CLI flag, i.e. I have a config file like that:

```
[..]
leaderElection:
  leaderElect: true
[..]
```
so now if I set leaderElect to false using a CLI flag, it is still going to be set to true unlike the other fields.

The culprit seems to be located [here](https://github.com/kubernetes-sigs/controller-runtime/blob/0d84ce43073cefb813563dd7586c943fa2c97783/pkg/manager/manager.go#L423). 

```
if o.LeaderElection == false && obj.LeaderElection.LeaderElect != nil {
	o.LeaderElection = *obj.LeaderElection.LeaderElect
}
```
if the field is set to false and the value is specified in the config file, `LeaderElection` will be set to the value of the config file even though it was meant to be overridden by a flag.
A solution could be using a `*bool` type. What do you think?

aayushrangwala:
@criscola I think the expectation of CLI flags to be given preference over the config file is valid. Considering the `AndFrom` comment
```
// AndFrom will use a supplied type and convert to Options
// any options already set on Options will be ignored, this is used to allow
[...]

PR #1492, opened 2021-04-25
 ⚠️ fix the leader election precedence over config file

aayushrangwala:
Fixes: #1463 

According to the documentation, flags to the cli should be considered over the properties in config files. But in the case of `--enable-leader-election` flag, there was a corner case which was not compliant to the statement.

case: If the user explicitly gives `false` to the leader election flag, but config file has true, the manager option is actually having the config value. Because of the fact that since it is a bool variable, its difficult to differentiate if the flag is given with value false or the zero value of the variable is being considered

Update LeaderElection Manager Option type to *bool

- Changed the datatype of leader election for manager option from bool to *bool

- Changed the flag parsing logic in Webhook main according to the new datatype
If the flag is not passed, it will be nil, otherwise false or true

- Added test to check the condition if true in config file and false in flag

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -21,6 +21,7 @@ package apiutil
 
 import (
 	"fmt"
+	"strconv"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -168,4 +169,26 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
 	}
 
 	return cfg
+}
+
+// BoolPointerFlagFunc returns a function to be associated with the custom bool flag func
+//
+// Example:
+//
+// var fooPtr *bool
+// flag.Func("foo", "foo usage", apiutil.BoolPointerFlagFunc(fooPtr))
+func BoolPointerFlagFunc(ptr *bool) func(flagVal string) error {
+	return func(flagVal string) error {
+		// flagVal is empty, consider flag as not passed
+		if flagVal == "" {
+			return nil
+		}
+
+		boolVal, err := strconv.ParseBool(flagVal)
+		if err != nil {
+			return err
+		}
+		ptr = &boolVal
+		return nil
+	}
 }
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -370,9 +370,8 @@ var _ = Describe("manger.Manager", func() {
 
 		Context("with leader election enabled", func() {
 			It("should only cancel the leader election after all runnables are done", func() {
-				leaderElectOpt := true
 				m, err := New(cfg, Options{
-					LeaderElection:          &leaderElectOpt,
+					LeaderElection:          pointer.BoolPtr(true),
 					LeaderElectionNamespace: "default",
 					LeaderElectionID:        "test-leader-election-id-2",
 					HealthProbeBindAddress:  "0",
@@ -416,9 +415,8 @@ var _ = Describe("manger.Manager", func() {
 
 			})
 			It("should disable gracefulShutdown when stopping to lead", func() {
-				leaderElectOpt := true
 				m, err := New(cfg, Options{
-					LeaderElection:          &leaderElectOpt,
+					LeaderElection:          pointer.BoolPtr(true),
 					LeaderElectionNamespace: "default",
 					LeaderElectionID:        "test-leader-election-id-3",
 					HealthProbeBindAddress:  "0",
@@ -446,9 +444,8 @@ var _ = Describe("manger.Manager", func() {
 			})
 			It("should default ID to controller-runtime if ID is not set", func() {
 				var rl resourcelock.Interface
-				leaderElectOpt := true
 				m1, err := New(cfg, Options{
-					LeaderElection:          &leaderElectOpt,
+					LeaderElection:          pointer.BoolPtr(true),
 					LeaderElectionNamespace: "default",
 					LeaderElectionID:        "test-leader-election-id",
 					newResourceLock: func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error) {
@@ -466,10 +463,9 @@ var _ = Describe("manger.Manager", func() {
 				m1cm, ok := m1.(*controllerManager)
 				Expect(ok).To(BeTrue())
 				m1cm.onStoppedLeading = func() {}
-				leaderElectOpt = true
 
 				m2, err := New(cfg, Options{
-					LeaderElection:          &leaderElectOpt,
+					LeaderElection:          pointer.BoolPtr(true),
 					LeaderElectionNamespace: "default",
 					LeaderElectionID:        "test-leader-election-id",
 					newResourceLock: func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error) {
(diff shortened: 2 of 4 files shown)


alvaroaleman:
Looks good, but is a breaking change.
/retitle :bug: fix the leader election precedence over config file
/ok-to-test

@vincepri WDYT, should we merge this or wait for a 0.9 release?

alvaroaleman:
wow, fail.
/retitle :warning: fix the leader election precedence over config file

vincepri:
Seems the code is now panicking.

christopherhein:
These changes look good to me, thanks for catching that!

aayushrangwala:
> Seems the tests are failing because of a panic, probably due to the pointer change.

@vincepri Yes, I handled the invalid pointer error. Tests seems to be passing now

pkg/webhook/conversion/testdata/main.go
```diff
@@ -46,10 +47,19 @@ func init() {
...
+	flag.Func("enable-leader-election",
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.",
+		func(flagVal string) error {
+			boolVal, err := strconv.ParseBool(flagVal)
+			if err != nil {
+				return err
+			}
+			enableLeaderElection = &boolVal
+			return nil
+		},
```
christopherhein, line 61: I wonder if we could provide a helper for this since this is a big chunk people will have to repeat.

aayushrangwala: I think the main code lines the users will repeat is to convert normal bool var to pointer which can be done by k8s util `pointer.BoolPtr()`. If we think about it in a generic way this function is taking the string and assigning/returning the bool ptr and error `IF` its a valid string. If user wants that they can directly `strong.ParseBool()`

Or I am I missing some point?

vincepri: Why do we need to change the flag?

vincepri: We can probably just use the plain bool, and dereference it, or use pointer.BoolPtr(enableLeaderElection) instead.

aayushrangwala: That was the whole point for the change actually. If we keep the flag as plain bool, we won't be able to differentiate if the user explicitly gave it false or it was the zero value of that variable.

If user gave it false then we have to pass it down to the `Options` so that the config file option is not considered. If user did not give anything for this flag, then we will consider the config file option

christopherhein: I was more thinking of a helper function so every controller that currently uses:

```golang
flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
```

is now going to have to repeat your flap parsing func, while not hard it's breaking and something folks will have to do when upgrading, thus it could be nice to have a helper func people could define here instead. Something like:

```golang
var enableLeaderElection *bool
flag.Func("enable-leader-election",
		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.",
		ctrl.BoolPointerFlag(enableLeaderElection),
)
```

then in `pkg/manager` we could have `BoolPointerFlag()` as a helper.

```golang
func BoolPointerFlag(value *bool) func(flagVal string) error {
	return func(flagVal string) error {
		boolVal, err := strconv.ParseBool(flagVal)
		if err != nil {
			return err
		}
		value = &boolVal
		return nil
	}
}
```

Might be unnecessary 🤷‍♂️

vincepri: Does returning an error from BoolPointerFlag actually do what you want this to do? See https://play.golang.org/p/-VBOv0ZkZ1k, if the string is empty, the bool conversion fails

aayushrangwala: @vincepri Yeah I understood your point. So the current behaviour would be following:
- If `--enable-leader-election` is not passed -> the function block to pointer conversion won't run
- If `--enable-leader-election` is passed but with either empty string `""` or with some invalid string -> It will fail with showing the usage message
- If `--enable-leader-election` is passed with either `true/false` -> it will accept and move ahead

aayushrangwala: @christopherhein I got your point. I will add that helper

pkg/client/apiutil/apimachinery.go
```diff
@@ -169,3 +170,20 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
...
 }
+
+// BoolPointerFlagFunc returns a function to be associated with the custom bool flag func
+//
+// Example:
+//
+// var fooPtr *bool
+// flag.Func("foo", "foo usage", apiutil.BoolPointerFlagFunc(fooPtr))
+func BoolPointerFlagFunc(ptr *bool) func(flagVal string) error {
+	return func(flagVal string) error {
```
vincepri, line 181: Shouldn't we say `if flagVal == "" { return nil }` ?

aayushrangwala: I think instead of returning nil, we should assign nil to the bool pointer. Which would imply that
- If the user gave the flag but with empty value, it will be considered as flag not passed
- If the user gave the flag but with some invalid value, it will be error
-  If the user gave the flag but with some valid value, it will be handled properly

aayushrangwala:
@vincepri @christopherhein  Please approve if LG

christopherhein:
This LGTM.

aayushrangwala:
@vincepri Can you please take a final look and add approved/lgtm label. Since this is api breaking change we can have it in the next patch release

alvaroaleman (approved):
/hold

pkg/client/apiutil/apimachinery.go
```diff
@@ -169,3 +170,25 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
...
+		// flagVal is empty, consider flag as not passed
+		if flagVal == "" {
+			return nil
+		}
+
+		boolVal, err := strconv.ParseBool(flagVal)
+		if err != nil {
+			return err
+		}
+		ptr = &boolVal
```
alvaroaleman, line 191: Actually, this doesn't do what you think it does. The `ptr` variable holds a pointer and is scoped to this function. You will need to:
* Set its value (which is problematic here I think, because then it will always be non-nil?)
* Use a double pointer `**bool`
/lgtm cancel

pkg/manager/manager_test.go
```diff
@@ -452,7 +536,8 @@ var _ = Describe("manger.Manager", func() {
 				Expect(err).To(MatchError(ContainSubstring("expected error")))
 			})
 			It("should return an error if namespace not set and not running in cluster", func() {
-				m, err := New(cfg, Options{LeaderElection: true, LeaderElectionID: "controller-runtime"})
+				leaderElectOpt := true
+				m, err := New(cfg, Options{LeaderElection: &leaderElectOpt, LeaderElectionID: "controller-runtime"})
```
alvaroaleman, line 540: Somewhat nitpicky, but if we import the `utils/pointer` image anyways, why not use `pointer.Boolptr` everywhere? It makes the code IMHO easier to read as you don't have to go back to check what the value of `leaderElectOpt` is

aayushrangwala: Agree. Changed

pkg/client/apiutil/apimachinery.go
aayushrangwala: Okay, got it. Let me think and work on it. Thanks for pointing that out

alvaroaleman:
@aayushrangwala will you be able to rebase and finish this?

Issue #1475, opened 2021-04-14
GetConfigOrDie log nothing


weilaaa:
Hi there.

When we try to get *rest.Config if there is not kubeconfig exactly, it's supposed to log an error "no configuration has been provided, try setting KUBERNETES_MASTER environment variable" like code below:
```golang
func GetConfigOrDie() *rest.Config {
	config, err := GetConfig()
	if err != nil {
		log.Error(err, "unable to get kubeconfig")
		os.Exit(1)
	}
	return config
}
```
but I got "exit 1" directly without error log print:
**got**
```
➜  helloworld go run .   
exit status 1
```
**want**
```
2021/04/14 21:02:37 error no configuration has been provided, try setting KUBERNETES_MASTER environment variable
exit status 1
```

With permission，I'm willing to fix it.

```
Go deps:
k8s.io/api v0.20.5
k8s.io/apimachinery v0.21.0
k8s.io/client-go v0.20.5
sigs.k8s.io/controller-runtime v0.7.0
```

/kind bug

joelanford:
@weilaaa You are likely not seeing logs because you have not set the logger before making this call. For an example, here is how the default kubebuilder scaffolding sets the logger: https://github.com/kubernetes-sigs/kubebuilder/blob/e6f3a70339b03f0509f19cd02cfa41c307d9a96c/testdata/project-v3/main.go#L66

weilaaa:
@joelanford ctrl.SetLogger() works, but if should we initialize a default logger to avoid situation that logger not set or user forget it instead of logging nothing?

joelanford:
Looks like this PR may be what you're after: https://github.com/kubernetes-sigs/controller-runtime/pull/1446

weilaaa:
Thanks, that's what I want, count me in.

shenlanse:
I think if  fail to `GetConfig`, should return error, instead of os.exit.  It requires log implementation `log.Error` method to flush in-memory log to disk or else the log would be lost. 
@joelanford @weilaaa

PR #1493, opened 2021-04-26
🐛Warn if logger is not explicitly set in two situations

weilaaa:
What type of PR is this?
/kind bugfix

What this PR does / why we need it:
print error reason when call `config.GetConfigOrDie()` failed instead of log nothing

Which issue(s) this PR fixes:
Part of #1475

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -150,7 +150,7 @@ func loadConfigWithContext(apiServerURL string, loader clientcmd.ClientConfigLoa
 func GetConfigOrDie() *rest.Config {
 	config, err := GetConfig()
 	if err != nil {
-		log.Error(err, "unable to get kubeconfig")
+		fmt.Fprintln(os.Stderr, "error: unable to get kubeconfig")
 		os.Exit(1)
 	}
 	return config


cndoit18 (approved):
/LGTM

pkg/log/null.go
```diff
@@ -25,23 +29,46 @@ import (
 // all binaries.
 
 // NullLogger is a logr.Logger that does nothing.
-type NullLogger struct{}
+type NullLogger struct {
```
alvaroaleman, line 32: I don't think this makes sense. The null logger shouldn't consider itself to be problematic. I am fine with the warning when it is being set in the delgating logger, but not here

pkg/log/log_test.go
```diff
@@ -128,7 +128,7 @@ var _ = Describe("logging", func() {
 		BeforeEach(func() {
 			root = &fakeLoggerRoot{}
 			baseLog = &fakeLogger{root: root}
-			delegLog = NewDelegatingLogger(NullLogger{})
+			delegLog = NewDelegatingLogger(DefaultNullLogger())
```
vincepri, line 131: This seems kind of a breaking change if we now require to use `DefaultNullLogger` instead of `NullLogger{}`. 

Could we achieve something similar without breaking users by leveraging the `NullLogger{}` struct instead?

Also, given that `NullLogger{}` is not a pointer, doesn't sync.Once get copied?

pkg/log/null.go
```diff
@@ -25,23 +29,46 @@ import (
...
+func (w warnOnce) do() {
+	w.once.Do(func() {
+		fmt.Fprintln(os.Stdin, w.warnPrint)
+	})
+}
 
 var _ logr.Logger = NullLogger{}
 
+// DefaultNullLogger a default NullLogger with specified warn print
+func DefaultNullLogger() NullLogger {
```
alvaroaleman, line 51: Why export this?

weilaaa: Sorry for disconnection for a long while! 
I abandoned this intrusive way, used a simple way to print error log.

PR #1494, opened 2021-04-26
[WIP]🌱 Add OWNERS for pkg/client/fake to support maintaining the fake client

detiber:
Following up on the issue related to the deprecation/undeprecation of the fake client: https://github.com/kubernetes-sigs/controller-runtime/issues/768

Adds an OWNERS file for pkg/client/fake and related aliases in OWNERS_ALIASES

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -34,6 +34,15 @@ aliases:
   - apelisse
   - hoegaarden
 
+  # folks who can approve things related to the fake client
+  fake-client-approvers:
+  - detiber
+
+  # folks wo can review and LGTM any PRs related to the fake client
+  fake-client-reviewers:
+  - detiber
+  - sbueringer
+
   # folks who may have context on ancient history,
   # but are no longer directly involved
   # controller-runtime-emeritus-maintainers:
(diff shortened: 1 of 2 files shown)


sbueringer:
@detiber I would help, but I don't have the review experience here, so not sure if this would be appropriate.

alvaroaleman:
> @detiber I would help, but I don't have the review experience here, so not sure if this would be appropriate.

Having more reviewers even if they do not have approvers right would already be very helpful and getting experience is basically the reason why the possibility to first become a reviewer without approval rights exists :)

detiber:
I agree with @alvaroaleman, went ahead and added you as a reviewer @sbueringer. Thank you for volunteering.

phil9909:
I would be willing to help as well. I'm currently not a member of the org, but I guess that would be a prerequisite, right?
Anything else I can do to support you?

detiber:
> I would be willing to help as well. I'm currently not a member of the org, but I guess that would be a prerequisite, right?
> Anything else I can do to support you?

Membership is indeed a requirement, here are the guidelines for org membership: https://github.com/kubernetes/community/blob/master/community-membership.md

PR #1499, opened 2021-04-29
:sparkles: Expose function to find controller namespace in controllerutil

randomvariable:
A number of downstream CRs have been re-implementing the ability to
find the namespace the controller is running under, so we make this
a public function, and ensure memoization of the resolution.

Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

Follow up from https://github.com/kubernetes-sigs/cluster-api/pull/4514

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -18,11 +18,14 @@ package controllerutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"reflect"
+	"sync"
 
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -31,6 +34,91 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
+
+var (
+	// Will store memoized controller namespace
+	inClusterNamespace      string
+	inClusterNamespacePath  = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+	inClusterNamespaceMutex sync.Mutex
+	// ErrNotRunningInCluster is the error for when the controller is not running in cluster
+	ErrNotRunningInCluster = &NotRunningInClusterError{}
+	// ErrNamespaceReadError is the error when the controller cannot read its namespace
+	ErrNamespaceReadError = &NamespaceReadError{}
+)
+
+// ReadInClusterNamespace returns the namespace by which the controller is running under.
+// read from "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+func ReadInClusterNamespace() (string, error) {
+	data, err := os.ReadFile(inClusterNamespacePath)
+	if errors.Is(err, os.ErrNotExist) {
+		return "", &NotRunningInClusterError{cause: err}
+	} else if err != nil {
+		return "", &NamespaceReadError{cause: err}
+	}
+
+	return string(data), nil
+}
+
+// ReadInClusterNamespaceCached returns the namespace by which the controller is running under.
+// read from "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+// Note that the resolved namespace is cached after it is successfully read from disk.
+func ReadInClusterNamespaceCached() (string, error) {
+	inClusterNamespaceMutex.Lock()
+	ns := inClusterNamespace
+	if ns != "" {
+		inClusterNamespaceMutex.Unlock()
+		return ns, nil
+	}
+	ns, err := ReadInClusterNamespace()
+	if err != nil {
+		inClusterNamespaceMutex.Unlock()
+		return "", err
+	}
+	// memoize result for next call
+	inClusterNamespace = ns
+	inClusterNamespaceMutex.Unlock()
+	return ns, nil
+}
+
+// NotRunningInClusterError is an error returned if the controller is not running in the cluster
+// determined by "/var/run/secrets/kubernetes.io/serviceaccount/namespace" not existing.
+type NotRunningInClusterError struct {
+	cause error
+}
+
+func (e *NotRunningInClusterError) Error() string {
+	return "not running in-cluster"
+}
+
+// Unwrap returns the error cause
+func (e *NotRunningInClusterError) Unwrap() error {
+	return e.cause
+}
+
+// Is determines if the target error is a NotRunningInClusterError
+func (e *NotRunningInClusterError) Is(target error) bool {
+	return target.Error() == e.Error()
+}
+
+// NamespaceReadError is an error returned if the controller cannot read
+// from "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+type NamespaceReadError struct {
+	cause error
+}
+
+func (e *NamespaceReadError) Error() string {
+	return fmt.Sprintf("error reading namespace file at %q", inClusterNamespacePath)
+}
+
+// Unwrap returns the error cause
+func (e *NamespaceReadError) Unwrap() error {
+	return e.cause
+}
+
+// Is determines if the target error is a NamespaceReadError
+func (e *NamespaceReadError) Is(target error) bool {
+	return target.Error() == e.Error()
+}
 
 // AlreadyOwnedError is an error returned if the object you are trying to assign
 // a controller reference is already owned by another controller Object is the
@@ -197,7 +285,7 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
 	key := client.ObjectKeyFromObject(obj)
 	if err := c.Get(ctx, key, obj); err != nil {
-		if !errors.IsNotFound(err) {
+		if !apierrors.IsNotFound(err) {
 			return OperationResultNone, err
 		}
 		if err := mutate(f, key, obj); err != nil {
@@ -234,7 +322,7 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f M
 func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
 	key := client.ObjectKeyFromObject(obj)
 	if err := c.Get(ctx, key, obj); err != nil {
-		if !errors.IsNotFound(err) {
+		if !apierrors.IsNotFound(err) {
 			return OperationResultNone, err
 		}
 		if f != nil {
--- a/pkg/leaderelection/leader_election.go
+++ b/pkg/leaderelection/leader_election.go
@@ -19,17 +19,15 @@ package leaderelection
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/recorder"
 )
-
-const inClusterNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
 // Options provides the required configuration to create a new resource lock
 type Options struct {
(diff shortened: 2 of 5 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,33 @@ import (
...
+		return "", fmt.Errorf("error checking namespace file: %w", err)
+	}
+
+	// Load the namespace file and return its content
+	namespace, err := ioutil.ReadFile(inClusterNamespacePath)
+	if err != nil {
+		return "", fmt.Errorf("error reading namespace file: %w", err)
+	}
+	inClusterNamespace = string(namespace)
+	return inClusterNamespace, nil
```
vincepri, line 61:
```suggestion
	return string(namespace), nil
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,33 @@ import (
...
+	if os.IsNotExist(err) {
+		return "", fmt.Errorf("not running in-cluster")
+	} else if err != nil {
+		return "", fmt.Errorf("error checking namespace file: %w", err)
+	}
+
+	// Load the namespace file and return its content
+	namespace, err := ioutil.ReadFile(inClusterNamespacePath)
+	if err != nil {
+		return "", fmt.Errorf("error reading namespace file: %w", err)
```
vincepri, line 58:
```suggestion
		return "", fmt.Errorf("error reading namespace file at %q: %v", inClusterNamespacePath, err)
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,33 @@ import (
...
+	if inClusterNamespace != "" {
+		return inClusterNamespace, nil
+	}
+	// Check whether the namespace file exists.
+	// If not, we are not running in cluster so can't guess the namespace.
+	_, err := os.Stat(inClusterNamespacePath)
+	if os.IsNotExist(err) {
+		return "", fmt.Errorf("not running in-cluster")
+	} else if err != nil {
+		return "", fmt.Errorf("error checking namespace file: %w", err)
```
vincepri, line 52:
```suggestion
		return "", fmt.Errorf("error checking namespace file at %q: %v", inClusterNamespacePath, err)
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,33 @@ import (
...
+	}
+
+	// Load the namespace file and return its content
+	namespace, err := ioutil.ReadFile(inClusterNamespacePath)
```
vincepri, line 56:
```suggestion
	namespace, err := os.ReadFile(inClusterNamespacePath)
```

pkg/controller/controllerutil/controllerutil.go
vincepri, line 52: Do we actually need this part? Couldn't wew check the result of `ReadFile` instead?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,33 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+const inClusterNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+var inClusterNamespace string
+
+// GetInClusterNamespace returns the namespace by which the controller is running under.
+func GetInClusterNamespace() (string, error) {
```
vincepri, line 42: Can we get some tests in place?

pkg/controller/controllerutil/controllerutil.go
randomvariable: That was a deliberate memoization (and pretty much the one thing I changed to the existing func), so we don't have to read the file on every call.

vincepri: Ah ok

pkg/controller/controllerutil/controllerutil.go
randomvariable: Going to check. I do remember doing this before somewhere else for some obscure POSIX error reason.

randomvariable: Wasn't necessary to do Stat separately, have cleaned up.

pkg/controller/controllerutil/controllerutil.go
randomvariable: Done, though it's a bit clunky because of having of envtest not running the controller in a real cluster.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,35 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+const inClusterNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+var (
+	// Will store memoized controller namespace
+	inClusterNamespace string
```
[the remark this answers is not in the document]
randomvariable: No, not really, don't need to optimise that far. Having a few extra file reads are fine if that's what ends up happening.

That said, have changed this to a map over the passed in filename, though no one should do that in real usage.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,37 @@ import (
...
+	// Will store memoized controller namespace
+	inClusterNamespace = make(map[string]string)
+)
+
+// GetInClusterNamespace returns the namespace by which the controller is running under.
+// Use file to pass in a different file. If "", will use the default location,
+// "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+// Note that the resolved namespace is persisted after it is read from disk.
+func GetInClusterNamespace(file string) (string, error) {
+	if inClusterNamespace, ok := inClusterNamespace[file]; ok {
```
vincepri, line 49: Could we start with no arguments? Let's see if folks need this to be configurable, passing an empty string seems a bit odd

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,37 @@ import (
...
+var (
+	// Will store memoized controller namespace
+	inClusterNamespace = make(map[string]string)
+)
+
+// GetInClusterNamespace returns the namespace by which the controller is running under.
+// Use file to pass in a different file. If "", will use the default location,
+// "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+// Note that the resolved namespace is persisted after it is read from disk.
+func GetInClusterNamespace(file string) (string, error) {
```
estroz, line 48: The name should be explicit about what it's doing
```suggestion
func ReadCachedInClusterNamespace(file string) (string, error) {
```
`ReadInClusterNamespaceCached` works too

pkg/controller/controllerutil/controllerutil.go
randomvariable: what's the best way to handle the test of the function?

vincepri: We could make `inClusterNamespacePath` a variable and change it for testing? Or can we write at that path during unit tests?

randomvariable:
Alrighty, I've restored the original implementation I started with (no arguments required), but created a new test file in the controllerutil package so it can override the file path for the purposes of testing.

randomvariable:
etcd pod not starting can't be me, is it?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,86 @@ import (
...
+		return inClusterNamespace, nil
+	}
+	ns, err := ReadInClusterNamespace()
+	if err != nil {
+		return "", err
+	}
+
+	// memoize result for next call
+	inClusterNamespacePath := ns
+	return inClusterNamespacePath, nil
```
estroz, line 74:
```suggestion
	inClusterNamespace = ns
	return inClusterNamespace, nil
```

vincepri:
@randomvariable Test is a bit flaky

pkg/controller/controllerutil/controllerutil.go
randomvariable: Argh.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,86 @@ import (
...
+	if inClusterNamespace != "" {
+		return inClusterNamespace, nil
+	}
+	ns, err := ReadInClusterNamespace()
+	if err != nil {
+		return "", err
+	}
+
+	// memoize result for next call
+	inClusterNamespace := ns
```
estroz, line 73: This still isn't storing `ns` globally, it's redefining `inClusterNamespace` in the local scope
```suggestion
	inClusterNamespace = ns
```

randomvariable: I should probably retire at this stage :)

pkg/leaderelection/leader_election.go
```diff
@@ -45,6 +43,10 @@ type Options struct {
 	// election resource will be created.
 	LeaderElectionNamespace string
 
+	// LeaderElectionUncachedNamespaceLookup if set to true will do an uncached read of the
+	// in-cluster namespace
+	LeaderElectionUncachedNamespaceLookup bool
```
estroz, line 48: Hmm I'm not sure if this is worth surfacing as an option. It is kinda confusing as well (why cache this namespace in the first place? why do I care?). Is it really that hard to call `options.LeaderElectionNamespace = ReadInClusterNamespaceCached()` or `options.LeaderElectionNamespace = ReadInClusterNamespace()` depending on what mode you want when setting up `Options`, with a default to `ReadInClusterNamespace()`?

randomvariable: Yeah, I agree. First of all was trying to fix the test, so added the uncached, but then after doing that, there's no reason to do the cached lookup for leader election where it's not going to be called very often unlike consumers within reconciliation loops. Deleting.

estroz (approved):
/lgtm

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -18,11 +18,13 @@ package controllerutil
...
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
```
alvaroaleman, line 27: Didn't know mixed-case import aliasing is allowed

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,86 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+var (
+	// Will store memoized controller namespace
+	inClusterNamespace     string
+	inClusterNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```
alvaroaleman, line 40: Should this be a const?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +34,86 @@ import (
...
+	}
+
+	return string(data), nil
+}
+
+// ReadInClusterNamespaceCached returns the namespace by which the controller is running under.
+// read from "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+// Note that the resolved namespace is cached after it is successfully read from disk.
+func ReadInClusterNamespaceCached() (string, error) {
+	if inClusterNamespace != "" {
```
alvaroaleman, line 64: I think we should make this threadsafe

randomvariable: Seemed a bit unnecessary given the very low impact, but have done so.

pkg/controller/controllerutil/controllerutil.go
randomvariable: changed to lowercase anyway

pkg/controller/controllerutil/controllerutil.go
randomvariable: Normally yes, but am overriding the value for testing the function, and the alternatives weren't much better.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +35,91 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+var (
+	// Will store memoized controller namespace
+	inClusterNamespace      string
+	inClusterNamespacePath  = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```
varshaprasad96, line 41: could this be an alternative instead of declaring the path as a variable?
```go
var readServiceAccountFile = func() ([]byte, error) {
	return ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
}



func ReadInClusterNamespace() (string, error) {
      nsBytes, err :=  readServiceAccountFile()
      ......
}
```

this would make testing easier, in the sense:
```go
It("should return the right namespace", func() {
     // overwriting the variable
     readServiceAccountFile = func() ([]byte, error) {
          return []byte("testnamespace"), nil
     }

     ns, err := ReadInClusterNamespace()
     Expect(err).Should(BeNil())
     Expect(ns).To(Equal("testnamespace"))
}) 
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -32,6 +35,91 @@ import (
...
+		return "", &NamespaceReadError{cause: err}
+	}
+
+	return string(data), nil
+}
+
+// ReadInClusterNamespaceCached returns the namespace by which the controller is running under.
+// read from "/var/run/secrets/kubernetes.io/serviceaccount/namespace".
+// Note that the resolved namespace is cached after it is successfully read from disk.
+func ReadInClusterNamespaceCached() (string, error) {
```
joelanford, line 65: What's the use case behind the cached version of this function? I feel like we should be telling operator authors to use the non-cached function once during operator startup and keep track of the namespace as necessary themselves.

PR #1500, opened 2021-04-29
:bug: add manager.Options.WebhookServer so webhook server can be set

estroz:
This PR adds `pkg/manager.Options.WebhookServer` to allow external configuration of a webhook server. Currently there is no way to add a custom webhook server, since `Manager.Add()` will not set `controllerManager.webhookServer`. I am treating this as a bug since the manager's internals appear to allow external webhook configuration, but no code API was ever exposed to allow this.

/kind bug

Signed-off-by: Eric Stroczynski <ericstroczynski@gmail.com>

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -365,7 +365,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		return nil, err
 	}
 
-	cm := &controllerManager{
+	return &controllerManager{
 		cluster:                 cluster,
 		recorderProvider:        recorderProvider,
 		resourceLock:            resourceLock,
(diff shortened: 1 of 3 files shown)


estroz:
There are probably a few ways to add a webhook server. This method preserves most existing logic without requiring injection, so I figured it is the easiest/forward-compatible way.

pkg/manager/manager.go
```diff
@@ -379,7 +387,17 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		gracefulShutdownTimeout: *options.GracefulShutdownTimeout,
 		internalProceduresStop:  make(chan struct{}),
 		leaderElectionStopped:   make(chan struct{}),
-	}, nil
+	}
+
+	// A webhook server set by New's caller should be added now
+	// so GetWebhookServer can construct a new one if unset and add it only once.
+	if cm.webhookServer != nil {
+		if err := cm.Add(cm.webhookServer); err != nil {
```
coderanger, line 395: I'm not sure New() is the right place to be adding runnables? That seems like it will make testing complicated.

estroz: I'm trying to think of a way to call `Add()` after `New()` without changing much. Perhaps some boolean flag to check/set in `GetWebhookServer()`.

coderanger: Yeah, I think some rearranging of the code in there would blend this more naturally with the existing flow.

estroz: Decided to use `sync.Once` instead.

cndoit18 (approved):
/LGTM

cndoit18 (approved):

vincepri (approved):
/approve

PR #1501, opened 2021-04-29
📖 Better document TLS requirements for webhook servers

kevindelgado:
This better documents the fact that `webhook.Server` requires valid TLS configurations and that `admission.StandaloneWebhook` requires the user to configure TLS on their own server.

Based on feedback from https://github.com/kubernetes-sigs/controller-runtime/pull/1429#discussion_r619719298


DirectXMan12 (approved):
/lgtm
/approve

Issue #1487, opened 2021-04-20
v0.8.3: admission.Handler is being injected with a nil admission.Decoder


nesv:
Hi there!

Recently, I have been upgrading a code base from `sigs.k8s.io/controller-runtime` from v0.6.0 &rarr; v0.8.3. In v0.6.0, everything was working fine, but after updating to v0.8.3, our `admission.Handler`s are being injected with nil `*admission.Decoder`s. 

Nothing has changed with how we are registering any CRDs.

I am still doing some digging through the changesets between 0.6.0 and 0.8.3 to see if anything changed with regards to decoder injection, but I wanted to file this issue just to get the ball rolling. I will post a comment for anything I find along the way.

### Dependencies

As a side note about dependencies, any dependencies in the `k8s.io/` tree had already been updated to a v.20.x tag. The `k8s.io/` packages are updated separately, and when updating to controller-runtime v0.8.3, no `k8s.io/` packages were updated as a side-effect, since the `k8s.io/` packages were already at compatible version.

I felt it necessary to mention this aspect, to preempt any discussion about `k8s.io/` package versions, since the only change in dependencies that occurred before this issue started, was bumping `sigs.k8s.io/controller-runtime` from 0.6.0 &rarr; 0.8.3.

nesv:
I'm officially stuck on where this issue could be happening.

With regards to the `runtime.Scheme` we pass when instantiating a `manager.Manager`, we are:

- Registering our CRDs with the [scheme.Builder](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/scheme#Builder) following the examples in the `scheme` package's godoc;
- Creating a [runtime.Scheme](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Scheme), passing it to `"k8s.io/client-go/kubernetes/scheme".AddToScheme`, as well as the `AddToScheme` functions in the packages for our CRDs;
- Creating a new [manager.Manager](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/manager#Manager) (via [manager.New](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/manager#New)), and passing the `runtime.Scheme` in via the [manager.Options](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/manager#Options).Scheme field.

[...]

PR #1307, opened 2020-12-23
✨ Move cluster-specifics from Manager into new pkg/cluster

alvaroaleman:
This change is the first step towards implementing the corresponding
[proposal][1]. It essentially consists of moving a bunch if code from
pkg/manager into pkg/cluster, slightly adjusting some tests in
pkg/manager and copying applicable tests from there into pkg/cluster.

It is not yet the full implementation, because the `Manager` will only
correctly start it's own cache and not other caches before starting the
remaining `Runnables`. This behavior matches current behavior when
building something that uses multiple caches and will be fixed in a
follow-up.

[1]: https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/move-cluster-specific-code-out-of-manager.md

PR #1502, opened 2021-05-01
🐛 fix nil Decoder in multiValidating and multiMutating handlers by implementing DecoderInjector

d-honeybadger:
**TL;DR**
Webhooks with `multiValidating` and `multiMutating` handlers currently fail to inject decoders into their sub-handlers, resulting in nil pointer panics. This PR is fixing it.

**Full description**
Before the change in https://github.com/kubernetes-sigs/controller-runtime/pull/1307, webhook method `webhook.InjectScheme` was called before `webhook.InjectFunc`, meaning that decoder intstantiation was happening before injecting it  by `webhook.InjectFunc`.
After that PR (specifically, the changes to `controllerManager.SetFields` in `pkg/manager/internal.go`) , the order has changed, and now `webhook.InjectFunc` gets called before `webhook.InjectScheme`, meaning that we  `webhook.InjectFunc` injects a `nil` decoder.
It appears that such re-ordering was anticipated, because in the `webhook.InjectScheme` it [tries to inject decoder again, just in case](https://github.com/kubernetes-sigs/controller-runtime/blob/2c238def1a48d4ff480f4745be2b258958acade0/pkg/webhook/admission/webhook.go#L165), with this comment:
```
// inject the decoder here too, just in case the order of calling this is not
// scheme first, then inject func
if w.Handler != nil {
	if _, err := InjectDecoderInto(w.GetDecoder(), w.Handler); err != nil {
		return err
	}
}
```
This second attempt at decoder injection works just fine for regular webhooks, but fails if the weebhook's handler is a `multiValidating` or `multiMutating` handler, because those handlers do not implement `DecoderInjector` interface.
Note that for regular handlers, it is the responsibility of the user to implement `DecoderInjector` interface, but because both `multiValidating` and `multiMutating` are unexported types, a user cannot do so for them.

The fix proposed in this PR is not the only possible solution. Another one could be to switch back the order of `inject.InjectorInto` and `cm.cluster.SetFields` in https://github.com/kubernetes-sigs/controller-runtime/blob/2c238def1a48d4ff480f4745be2b258958acade0/pkg/manager/internal.go#L227. And yet another one is proposed in https://github.com/kubernetes-sigs/controller-runtime/pull/1437

Link to the issue that was filed for this bug: https://github.com/kubernetes-sigs/controller-runtime/issues/1487

Tagging @porridge since you were looking at that issue and @alvaroaleman as the author of the PR that re-ordered `webhook.InjectScheme` and `webhook.InjectFunc`.

Happy to make any changes or implement a different fix for this bug.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -225,6 +225,9 @@ func (cm *controllerManager) Add(r Runnable) error {
 
 // Deprecated: use the equivalent Options field to set a field. This method will be removed in v0.10.
 func (cm *controllerManager) SetFields(i interface{}) error {
+	if err := cm.cluster.SetFields(i); err != nil {
+		return err
+	}
 	if _, err := inject.InjectorInto(cm.SetFields, i); err != nil {
 		return err
 	}
@@ -232,9 +235,6 @@ func (cm *controllerManager) SetFields(i interface{}) error {
 		return err
 	}
 	if _, err := inject.LoggerInto(cm.logger, i); err != nil {
-		return err
-	}
-	if err := cm.cluster.SetFields(i); err != nil {
 		return err
 	}
 
--- a/pkg/webhook/admission/multi.go
+++ b/pkg/webhook/admission/multi.go
@@ -77,6 +77,16 @@ func (hs multiMutating) InjectFunc(f inject.Func) error {
 	return nil
 }
 
+// InjectDecoder injects the decoder into the handlers.
+func (hs multiMutating) InjectDecoder(d *Decoder) error {
+	for _, handler := range hs {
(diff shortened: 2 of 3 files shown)


alvaroaleman (commented):
> The fix proposed in this PR is not the only possible solution. Another one could be to switch back the order of inject.InjectorInto and cm.cluster.SetFields in

@d-honeybadger would you mind to also do that, just on the off chance that something else also relied on the order there?

/ok-to-test

d-honeybadger:
> @d-honeybadger would you mind to also do that, just on the off chance that something else also relied on the order there?

:+1:  done!
And fixed a linting problem.

alvaroaleman (approved):
/lgtm
/approve

thanks!

PR #1503, opened 2021-05-01
:bug: Run unittests in CI

alvaroaleman:
This accidentally got disabled in
cd065bf2f63c3057db307990c21316483fa60ce8 under the assumption that the
CI entrypoint is the Makefile in an attempt to make this configurable.
Since the script is called `test_all.sh`, it is reasonable to assume it
always test verything. If ppl want something else, they can just add
another script/maketarget.

/assign @estroz @vincepri

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,6 @@
 SHELL:=/usr/bin/env bash
 .DEFAULT_GOAL:=help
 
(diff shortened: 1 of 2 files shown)


vincepri (approved):
/approve
/lgtm

PR #1504, opened 2021-05-02
✨ Cleanup Webhook server setup

### LINKED ISSUE #1496: envtest.Environment.Stop() leaves orphan /tmp/envtest-serving-certs-* directories
Executing the following snippet in tests:
```go
    testEnv := &envtest.Environment{}

    cfg, err := testEnv.Start()
    if err != nil {
      t.Fatalf("starting test environment: %v", err)
    }

    t.Cleanup(func() {
      t.Logf("stopping test environment")

      if err := testEnv.Stop(); err != nil {
        t.Logf("stopping test environment: %v", err)
      }
    })
```

Leaves orphan files in `/tmp`:
```console
$ ls /tmp/envtest-serving-certs-*
/tmp/envtest-serving-certs-210428206:
tls.crt  tls.key

/tmp/envtest-serving-certs-672117224:
tls.crt  tls.key

/tmp/envtest-serving-certs-778787569:
tls.crt  tls.key

/tmp/envtest-serving-certs-912074827:
tls.crt  tls.key
```

They should be cleaned up.

aayushrangwala:
- Called the webhook server cleanup function

- Only ignore controlplane cleanup when using existing cluster

- Added test to check the cleanup for Webhook after `Stop()`

Fixes #1496

### DIFF
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -23,11 +23,13 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -65,8 +67,8 @@ var _ = Describe("Test", func() {
 				Name: crd.GetName(),
 			}
 			var placeholder v1beta1.CustomResourceDefinition
-			err := c.Get(context.TODO(), crdObjectKey, &placeholder)
-			if err != nil && apierrors.IsNotFound(err) {
+			if err = c.Get(context.TODO(), crdObjectKey, &placeholder); err != nil &&
+				apierrors.IsNotFound(err) {
 				// CRD doesn't need to be deleted.
 				continue
 			}
@@ -842,6 +844,19 @@ var _ = Describe("Test", func() {
 			_, err := env.Start()
 			Expect(err).NotTo(HaveOccurred())
 			Expect(env.Stop()).To(Succeed())
+			close(done)
+		}, 30)
+	})
+
+	Describe("Stop", func() {
+		It("should cleanup webhook /tmp folder with no error when using existing cluster", func(done Done) {
+			env := &Environment{}
+			_, err := env.Start()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(env.Stop()).To(Succeed())
+
+			// check if the /tmp/envtest-serving-certs-* dir doesnt exists any more
+			Expect(env.WebhookInstallOptions.LocalServingCertDir).ShouldNot(BeADirectory())
 			close(done)
 		}, 30)
 	})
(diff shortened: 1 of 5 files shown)


invidian (commented):
Should there be some test that the directory has been properly cleaned up?

aayushrangwala:
> Should there be some test that the directory has been properly cleaned up?

Yes, definitely this needs to be part of unit tests and I was actually about to add too. But then I found there needs to be a comprehensive set of tests for checking the complete `Stop()` functionality of `evntest` as well as independent component such as `controlplan.Stop()` and `webhook.Stop()`. 

This seems to be the part of another PR to improve tests

invidian:
@aayushrangwala what about something like the following:
```go
	Describe("Stop", func() {
                Context("With existing cluster", func() {
	  	        It("should remove local serving certificate directory", func(done Done) {
			        env = &Environment{UseExistingCluster: pointer.Bool(true)}
			        _, err = env.Start()
			        Expect(env.Stop()).To(Succeed())
                                Expect(env.WebhookInstallOptions.LocalServingCertDir).ShouldNot(BeADirectory())
			        close(done)
		        }, 30)
                })
	})
```

I'm not sure if this is going to work, as it seems `UseExistingCluster` has no tests at all...

pkg/envtest/envtest_test.go
```diff
@@ -65,15 +65,15 @@ var _ = Describe("Test", func() {
 				Name: crd.GetName(),
 			}
 			var placeholder v1beta1.CustomResourceDefinition
-			err := c.Get(context.TODO(), crdObjectKey, &placeholder)
+			err = c.Get(context.TODO(), crdObjectKey, &placeholder)
```
alvaroaleman, line 68:
```suggestion
			err := c.Get(context.TODO(), crdObjectKey, &placeholder)
```

Please do not do scope changes for variables like this (This comment doesn't only apply to this line but to all places where you did that).

aayushrangwala: Got it. Will change it back. Did that as the part of go-lint with `no new variable`

alvaroaleman: not sure what `no new variable` is but IMHO it is best to keep the scope for variables as small as possible and this kind of change is the very opposite to that. Reduces the odds of accidentally propagating values to the wrong place and makes it easier for the garbage collector

aayushrangwala:
> @aayushrangwala what about something like the following:
> 
> ```go
> 	Describe("Stop", func() {
>                 Context("With existing cluster", func() {
> 	  	        It("should remove local serving certificate directory", func(done Done) {
> 			        env = &Environment{UseExistingCluster: pointer.Bool(true)}
> 			        _, err = env.Start()
> 			        Expect(err).NotTo(HaveOccurred())
> 			        Expect(env.Stop()).To(Succeed())
>                                 Expect(env.WebhookInstallOptions.LocalServingCertDir).ShouldNot(BeADirectory())
> 			        close(done)
> 		        }, 30)
>                 })
> 	})
> ```
> 
> I'm not sure if this is going to work, as it seems `UseExistingCluster` has no tests at all... Maybe it can be added with something like:
> 
> ```go
> if useExisting := os.Getenv("USE_EXISTING_CLUSTER"); useExisting == "" {
>         Skip("USE_EXISTING_CLUSTER environment variable required")
> }
> ```
> 
> So at least one can run it locally until there is a CI facility to execute this test automatically.

Yeah, will add a case to check the cleanup but I think we dont need to worry about the existing cluster scenario, because as per the code even for the new cluster the same setup happens for Webhook install. So its better to keep it generic

aayushrangwala:
@alvaroaleman  Please re-review

cndoit18 (approved):
LGTM

alvaroaleman (approved):
thanks!

Issue #1507, opened 2021-05-03
metadata client does not apply patch options


furkhat:
The `opts` argument is ignored. I think `.ApplyOptions(opts)` should be called here https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/metadata_client.go#L107

PR #1508, opened 2021-05-03
🐛 metadata client applies patch options

furkhat:
Signed-off-by: Furkhat Kasymovgeniiuulu <vailodf@gmail.com>

Patch options provided by caller are not ignored. #1507


alvaroaleman:
@furkhat thanks! Could you please add a test for this?

furkhat:
Sure, I will add test for this.

furkhat:
There seems to be no tests for `Patch`. Is this "technical dept" or there were other reasons to not add tests?

It also puzzles me that metadata client has `PatchStatus` method. How that would work? I would guess that this was added just to satisfy interface, otherwise I don't know.

alvaroaleman:
> There seems to be no tests for Patch. Is this "technical dept" or there were other reasons to not add tests?

Seems to be an oversight.

> It also puzzles me that metadata client has PatchStatus method. How that would work? I would guess that this was added just to satisfy interface, otherwise I don't know.

Why would it not work? Not being able to retrieve the status doesn't mean you can't set it.

furkhat:
> Not being able to retrieve the status doesn't mean you can't set it.

I am sorry I could not understand, what are you referring here?

Let me elaborate on what puzzles me. I think `PatchStatus` of metadata client would not work because the object passed to `PatchStatus` is always `metav1.PartialObjectMetadata` which AFAIK doesn't have any fields associated with "status".

alvaroaleman:
>  I think PatchStatus of metadata client would not work because the object passed to PatchStatus is always metav1.PartialObjectMetadata which AFAIK doesn't have any fields associated with "status"

No, it takes a client.Object. I agree that in all practical terms this is of limited use, as someone would need to construct an object with a status by themselves, they can not get it through the client.

furkhat:
> I agree that in all practical terms this is of limited use, as someone would need to construct an object with a status by themselves, they can not get it through the client.

No I don't mean that. I mean that according to code to trigger metadata client's PatchStatus the client.Object must be of `*metav1.PartialObjectMetadata` and I don't understand the the use..

Here
* https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/client/metadata_client.go#{L169:L172} - exits with error if type isn't `*metav1.PartialObjectMetadata`
* https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/client/client.go#L259-L260 - calls metadata_client.PatchStatus only if type is `*metav1.PartialObjectMetadata`

alvaroaleman:
@furkhat ok sure, feel free to open an issue for that. It is unrelated to this PR though, isn't it?

furkhat:
> It is unrelated to this PR though, isn't it?

`PatchStatus` doesn't apply patch options it receives in the same way so I wanted to fix it too.. I opened the issue #1514

alvaroaleman (approved):
thanks!

PR #1510, opened 2021-05-04
:bug: Fakeclient: Support WithWatch

alvaroaleman:
When adding client.WithWatch, we forgot to extend the fakeclient to also
support that, this change adds that support.

/assign @vincepri 

WDYT about changing the existing constructors to return a `client.WithWatch` vs adding a new set of constructors? I opted for extending them, because its easier to use, as `client.WithWatch` is a superset of `client.Client`, so will work as a replacement for both.

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -33,6 +34,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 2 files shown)


vincepri:
Would it makes sense to just have Watch within the Client interface?

alvaroaleman:
> Would it makes sense to just have Watch within the Client interface?

I think that would confuse a lot of ppl and make them try to use that for getting cached data. In the context of a cache-backed client it would also result in establishing a second watch. The usecase for this are mainly CLIs that wait for an event to happen.

vincepri:
Got it, that makes sense. Having the constructors to return a complete interface makes sense to me, although I wonder in that case if can also cause confusion.

alvaroaleman:
> Got it, that makes sense. Having the constructors to return a complete interface makes sense to me, although I wonder in that case if can also cause confusion.

Yeah, that is what I was unsure about as well. Explicitly splitting here means three new constructors though.

pkg/client/fake/client_test.go
```diff
@@ -541,6 +543,32 @@ var _ = Describe("Fake client", func() {
...
+			objWatch, err := cl.Watch(context.Background(), &corev1.ServiceList{})
+			Expect(err).NotTo(HaveOccurred())
+
+			defer objWatch.Stop()
+
+			go func() {
+				defer GinkgoRecover()
+				// It is likely starting a new goroutine is slower than progressing
+				// in the outer routine, sleep to make sure this is always true
+				time.Sleep(100 * time.Millisecond)
```
stevekuznetsov, line 557: Why not a channel or semaphore to enforce it?

alvaroaleman: What we are waiting for is that the main routine starts to read, how would you sync that with a channel or semaphore?

vincepri (approved):

PR #1512, opened 2021-05-05
:bug: Webhook servers should always start before cache syncs

vincepri:

Fixes a race condition between conversion webhooks to start and the cache being populated.


alvaroaleman (approved):

alvaroaleman (commented):
/lgtm

PR #1513, opened 2021-05-05
:bug: Quote path with -p env in envtest-setup

DirectXMan12:
The path could have spaces, single quotes, etc in in, so this quotes the
path and escapes single quotes so we can export properly.

This is very relevant on darwin, where our default store path contains spaces (it includes "Application Support")

### DIFF
--- a/tools/setup-envtest/env/env.go
+++ b/tools/setup-envtest/env/env.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"path/filepath"
 	"sort"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


DirectXMan12:
FYI @vincepri

vincepri (approved):

DirectXMan12:
hmm... looks like a flaky timing test?  Occaisional timeouts waiting for etcd of kube-apiserver to start, esp on the "it should be fast to set up and tear down" tests.

/retest

tools/setup-envtest/env/env.go
```diff
@@ -372,7 +373,14 @@ func (e *Env) PrintInfo(printFmt PrintFormat) {
...
 		fmt.Fprint(e.Out, path) // NB(directxman12): no newline -- want the bare path here
 	case PrintEnv:
-		fmt.Fprintf(e.Out, "export KUBEBUILDER_ASSETS=%s\n", path)
+		// quote in case there are spaces, etc in the path
+		// the weird string below works like this:
+		// - you can't escape quotes in shell
+		// - shell strings that are next to each other are concatenated (so "a""b""c" == "abc")
+		// - you can intermix quote styles using the above
+		// - so `'"'"'` --> CLOSE_QUOTE + "'" + OPEN_QUOTE
+		shellQuoted := strings.ReplaceAll(path, "'", `'"'"'`)
```
estroz, line 382: So the result is supposed to look something like the following?
```console
$ export FOO='/foo/'"'"'Application Support'"'"''
$ echo $FOO
/foo/'Application Support'
```

PR #1516, opened 2021-05-07
🌱 add a returning error for the Default method of the Defaulter interface

cndoit18:
The error triggered in mutation will be discarded, even when validation is being checked, the error details are lost.
Return the error directly when mutation, this will not lose the error details.

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -98,6 +98,14 @@ WAIT_LOOP:
 	Fail(fmt.Sprintf("timed out waiting for namespace %q to be deleted", ns.Name))
 }
 
+type mockPatchOption struct {
+	applied bool
+}
+
+func (o *mockPatchOption) ApplyToPatch(_ *client.PatchOptions) {
+	o.applied = true
+}
+
 // metaOnlyFromObj returns PartialObjectMetadata from a concrete Go struct that
 // returns a concrete *metav1.ObjectMeta from GetObjectMeta (yes, that plays a
 // bit fast and loose, but the only other options are serializing and then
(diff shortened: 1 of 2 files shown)


cndoit18:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.**
> 
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.
> 
> * If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address.  Check your existing CLA data and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * If you signed the CLA as a corporation, please sign in with your organization's credentials at https://identity.linuxfoundation.org/projects/cncf to be authorized.
> * If you have done the above and are still having issues with the CLA being reported as unsigned, please log a ticket with the Linux Foundation Helpdesk: https://support.linuxfoundation.org/
> * Should you encounter any issues with the Linux Foundation Helpdesk, send a message to the backup e-mail support address at: [login-issues@jira.linuxfoundation.org](mailto:login-issues@jira.linuxfoundation.org)

ok

cndoit18 (commented):
/check-cla

coderanger:
The general use case of this simplified layer doesn't really expect anything to be capable of failing. If you need a more complex webhook, you would drop down to the lower-level API shown in https://book.kubebuilder.io/reference/webhook-for-core-types.html. That allows for direct control over the hook response.

Also this would be an interface break which would require all users to update, and we try to not do those super often. If we were going to revise these interfaces, it should probably be in a more overall way so we do the API changes all together.

coderanger:
(notably you're sending back 400 Bad Request on an error but a failure in the Defaulter code would probably better be considered a 500 in most cases if I had to pick one)

cndoit18:
> The general use case of this simplified layer doesn't really expect anything to be capable of failing. If you need a more complex webhook, you would drop down to the lower-level API shown in https://book.kubebuilder.io/reference/webhook-for-core-types.html. That allows for direct control over the hook response.
> 
> Also this would be an interface break which would require all users to update, and we try to not do those super often. If we were going to revise these interfaces, it should probably be in a more overall way so we do the API changes all together.

I understand.
my mutation is not complicated, so I hope that the error will be returned in Defaulter, not to write one by Handler. 
so defaulter returns an error, I think it is an effect of expectations.

cndoit18:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.**
> 
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.
> 
> * If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address.  Check your existing CLA data and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * If you signed the CLA as a corporation, please sign in with your organization's credentials at https://identity.linuxfoundation.org/projects/cncf to be authorized.
> * If you have done the above and are still having issues with the CLA being reported as unsigned, please log a ticket with the Linux Foundation Helpdesk: https://support.linuxfoundation.org/
> * Should you encounter any issues with the Linux Foundation Helpdesk, send a message to the backup e-mail support address at: [login-issues@jira.linuxfoundation.org](mailto:login-issues@jira.linuxfoundation.org)

/check-cla

pkg/webhook/admission/defaulter.go
```diff
@@ -27,7 +27,7 @@ import (
 // Defaulter defines functions for setting defaults on resources
 type Defaulter interface {
 	runtime.Object
-	Default()
+	Default() error
```
alvaroaleman, line 30: /hold

Defaulting shouldn't error, use `admission.Validator` for that.

cndoit18: my commits are overwritten by the pullbot.
what I want is similar to the Validator interface, and multiplex the same path.
can I add this?
```
type Mutator interface {
	runtime.Object
	MutatorCreate() error
	MutatorUpdate(old runtime.Object) error
}
```

PR #1520, opened 2021-05-08
:bug: filter global namespace while looking for cluster scoped resources

varshaprasad96:
In the current implementation of multinamespaced cache, we look at
global namespace while listing objects from all namespaced. Global namespace
should not be looked at while fetching namespaced resources.

### DIFF
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -50,9 +50,18 @@ func MultiNamespacedCacheBuilder(namespaces []string) NewCacheFunc {
 		if err != nil {
 			return nil, err
 		}
-		// create a cache for cluster scoped resources
-		namespaces = append(namespaces, globalCache)
+
 		caches := map[string]Cache{}
+
+		// create a cache for cluster scoped resources
+		gCache, err := New(config, opts)
+		if err != nil {
+			return nil, fmt.Errorf("error creating global cache %v", err)
+		}
+
+		// add global cache to the cacheMap
+		caches[globalCache] = gCache
+
 		for _, ns := range namespaces {
 			opts.Namespace = ns
 			c, err := New(config, opts)
@@ -190,7 +199,10 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList, 
 	limitSet := listOpts.Limit > 0
 
 	var resourceVersion string
-	for _, cache := range c.namespaceToCache {
+	for ns, cache := range c.namespaceToCache {
+		if ns == globalCache {
+			continue
+		}
 		listObj := list.DeepCopyObject().(client.ObjectList)
 		err = cache.List(ctx, listObj, &listOpts)
 		if err != nil {
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -41,7 +41,7 @@ const globalCache = "_cluster-scope"
 // MultiNamespacedCacheBuilder - Builder function to create a new multi-namespaced cache.
 // This will scope the cache to a list of namespaces. Listing for all namespaces
 // will list for all the namespaces that this knows about. By default this will create
-// a global cache for cluster scoped resource (having empty namespace). Note that this is not intended
+// a global cache for cluster scoped resource. Note that this is not intended
 // to be used for excluding namespaces, this is better done via a Predicate. Also note that
 // you may face performance issues when using this with a high number of namespaces.
 func MultiNamespacedCacheBuilder(namespaces []string) NewCacheFunc {
@@ -59,9 +59,6 @@ func MultiNamespacedCacheBuilder(namespaces []string) NewCacheFunc {
 			return nil, fmt.Errorf("error creating global cache %v", err)
 		}
 
-		// add global cache to the cacheMap
-		caches[globalCache] = gCache
-
 		for _, ns := range namespaces {
 			opts.Namespace = ns
 			c, err := New(config, opts)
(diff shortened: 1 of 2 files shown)


varshaprasad96:
In the `GET` calls, the namespace of the object should be populated right ([here](https://github.com/kubernetes-sigs/controller-runtime/blob/745c7c98a7223069fcaab6b2e1b373f40d82e15a/pkg/cache/multi_namespace_cache.go#L149)) since its taken from the `ObjectKey`,  so I believe calling of global cache wouldn't occur. 
I think this should solve the bug caused from PR #1418

pkg/cache/multi_namespace_cache.go
```diff
@@ -190,7 +190,10 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 	limitSet := listOpts.Limit > 0
 
 	var resourceVersion string
-	for _, cache := range c.namespaceToCache {
+	for ns, cache := range c.namespaceToCache {
```
alvaroaleman, line 193: Can we somehow test that after we do a List for a namespaced object (say pods), only the namespaced caches have an informer for that object?

varshaprasad96: actually, nope. The global cache also has an informer for the namespaced object. That seems to be wrong.

varshaprasad96:
Have moved the global cache creation above the loop, so that the `_cluster-scope` namespace is not passed to `opts.Namespace`. In a follow up PR, will refactor this to not add global cache to the map itself, since as Eric mentioned, the namespace to cache map assumes that the namespaces are actually valid.

pkg/cache/multi_namespace_cache.go
alvaroaleman: How about that test? We could maybe also just make the global cache a new attribute in the `multiNamespaceCache` to avoid this exact issue of looping and accidentally including it

varshaprasad96: have modified this. Created a separate clusterCache in `multinamespaceCache`

pkg/cache/multi_namespace_cache.go
```diff
@@ -99,10 +112,24 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
 		}
 		informers[ns] = informer
 	}
+	clusterCacheInf, err := c.clusterCache.GetInformerForKind(ctx, gvk)
+	if err != nil {
+		return nil, err
+	}
+	informers[globalCache] = clusterCacheInf
```
varshaprasad96, line 119: we could ideally create a separate informer in [`multiNamespaceInformer`](https://github.com/kubernetes-sigs/controller-runtime/blob/0d84ce43073cefb813563dd7586c943fa2c97783/pkg/cache/multi_namespace_cache.go#L184) for cluster scoped cache. But that's not needed for now, since the namespace is not specifically used in any of the methods.

pkg/cache/multi_namespace_cache.go
```diff
@@ -87,6 +94,12 @@ func (c *multiNamespaceCache) GetInformer(ctx context.Context, obj client.Object
 		}
 		informers[ns] = informer
 	}
+	clusterCacheInf, err := c.clusterCache.GetInformer(ctx, obj)
```
alvaroaleman, line 97: I don't think this is correct, this should only happen if the obj is not namespaced

pkg/cache/multi_namespace_cache.go
```diff
@@ -99,10 +112,24 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
 		}
 		informers[ns] = informer
 	}
+	clusterCacheInf, err := c.clusterCache.GetInformerForKind(ctx, gvk)
```
alvaroaleman, line 115: Same here, we shouldn't call this unless the object is cluster-scopped (and then we should _only_ call this)

pkg/cache/multi_namespace_cache.go
varshaprasad96: Done, modified this to have `clusterCacheInformer` only when object is cluster scoped.

pkg/cache/multi_namespace_cache.go
```diff
@@ -87,6 +94,21 @@ func (c *multiNamespaceCache) GetInformer(ctx context.Context, obj client.Object
 		}
 		informers[ns] = informer
 	}
+
+	isNamespaced, err := objectutil.IsAPINamespaced(obj, c.Scheme, c.RESTMapper)
+	if err != nil {
+		return nil, err
+	}
+
+	if !isNamespaced {
```
alvaroaleman, line 103: This is mutually exclusive with getting the namespaced informers, we should either get the namespaced ones or the global ones, not both like we do now

varshaprasad96: ah, yes. right. The obj would either be namespaced or cluster scoped. will make the change.

pkg/cache/multi_namespace_cache.go
```diff
@@ -131,6 +189,10 @@ func (c *multiNamespaceCache) IndexField(ctx context.Context, obj client.Object,
 			return err
 		}
 	}
+
+	if err := c.clusterCache.IndexField(ctx, obj, field, extractValue); err != nil {
```
estroz, line 193: I think this also needs to be called only if obj is namespaced, otherwise iterate through `c.namespaceToCache`.

estroz (approved):
/lgtm

alvaroaleman (approved):

PR #1521, opened 2021-05-10
:bug: Fakeclient: Fix List for unregistered type through unstructured

alvaroaleman:
The objecttracker requires a List type to be registered for List to
work, hence we have to register UnstructuredList for the given list GVK
if an unstructuredList is targeted and nothing is registered yet.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/702

/assign @vincepri @estroz


vincepri (approved):
/lgtm

Issue #1522, opened 2021-05-12
LeaderElectionReleaseOnCancel is Not Honored


laverite:
Manager option `LeaderElectionReleaseOnCancel` (added in https://github.com/kubernetes-sigs/controller-runtime/commit/93e6db8d5f04222e42accaa91af027e8bc4ec52b) does not take effect when set.  It appears this option is simply not plumbed through all the way to the internal `controllerManager`.

PR #1523, opened 2021-05-12
:bug: honor LeaderElectionReleaseOnCancel

laverite:
Manager option `LeaderElectionReleaseOnCancel` (added in https://github.com/kubernetes-sigs/controller-runtime/commit/93e6db8d5f04222e42accaa91af027e8bc4ec52b) does not take effect when set. It appears this option is simply not plumbed through all the way to the internal controllerManager.  This PR plumbs the `LeaderElectionReleaseOnCancel` option through so that it is honored.

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/1522


laverite:
CLA signed

alvaroaleman (approved):
thanks!

PR #1524, opened 2021-05-12
🐛 typeMeta: typeMeta is empty when using client.Get()

### LINKED ISSUE #1517: TypeMeta is empty when using client.Get()
I use `client.New()` to build a `Client` and call its `Get()` method to get some object. But the returned object has an empty TypeMeta.

```go
package main

import (
	"context"
	"fmt"

	corev1 "k8s.io/api/core/v1"
	"k8s.io/client-go/kubernetes/scheme"
	_ "k8s.io/client-go/plugin/pkg/client/auth"
)

func main() {
	cli, err := client.New(config.GetConfigOrDie(), client.Options{Scheme: scheme.Scheme})
	if err != nil {
		panic(err)
	}
	ns := &corev1.Namespace{}
	if err := cli.Get(context.Background(), client.ObjectKey{Name: "default"}, ns); err != nil {
		panic(err)
	}
	fmt.Printf("TypeMeta: %+v\n", ns.TypeMeta)
}
```

Expected output:

```
TypeMeta: {Kind: Namespace APIVersion: v1}
```

Actual output:

```
TypeMeta: {Kind: APIVersion:}
```

cndoit18:
Signed-off-by: cndoit18 [cndoit18@outlook.com](mailto:cndoit18@outlook.com)
 
**What type of PR is this?**
 
**What this PR does / why we need it**:

This is a temporary repair that solves the problem of typeMate is empty.

**Which issue(s) this PR fixes**:

Fixes #1517
 
**Special notes for your reviewer**:
 
**Does this PR introduce a user-facing change?**:
 
```release-note
none
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -250,6 +250,17 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object) error {
 		defer c.resetGroupVersionKind(obj, obj.GetObjectKind().GroupVersionKind())
 		return c.metadataClient.Get(ctx, key, obj)
 	default:
+		// Metadata only object should always preserve the GVK coming in from the caller.
+		if obj != nil {
+			gvks, _, err := c.scheme.ObjectKinds(obj)
+			if err != nil {
+				return err
+			}
+			// The GVK obtained by reflect.Type, there is only one
+			if len(gvks) == 1 {
+				defer c.resetGroupVersionKind(obj, gvks[0])
+			}
+		}
 		return c.typedClient.Get(ctx, key, obj)
 	}
 }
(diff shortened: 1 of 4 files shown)


coderanger:
Echo'ing comments from Slack and #1517, I'm not sure we should fix this. An upstream thing would be better. And on top of that, this is a really niche issue and I would worry that even a small performance hit on such a common operation could be problematic when it has no upside for almost anyone.

cndoit18:
> Echo'ing comments from Slack and #1517, I'm not sure we should fix this. An upstream thing would be better. And on top of that, this is a really niche issue and I would worry that even a small performance hit on such a common operation could be problematic when it has no upside for almost anyone.



> Echo'ing comments from Slack and #1517, I'm not sure we should fix this. An upstream thing would be better. And on top of that, this is a really niche issue and I would worry that even a small performance hit on such a common operation could be problematic when it has no upside for almost anyone.

yes, I will close this PR and provide general solutions to the Issue.

PR #1525, opened 2021-05-14
⚠️ Envtest should modify CRDs appropriately when using webhooks

vincepri:
Before this change, the webhook code was patching mutation or admission
webhooks to point to the temporary URL and CA bundle generated when the
webhook server would start.

Conversion webhooks are handled at the CRD level in Kubernetes, and we
need to make sure to patch those client configurations as well.

This patch also unexports a lot of methods that should have been private
from the very beginning, so it's marked as a breaking change.

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -20,12 +20,15 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -35,6 +38,7 @@ import (
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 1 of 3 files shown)


pkg/envtest/crd.go
```diff
@@ -340,6 +348,33 @@ func renderCRDs(options *CRDInstallOptions) ([]client.Object, error) {
 	return res, nil
 }
 
+func modifyConversionWebhooks(crds []client.Object, webhookOptions WebhookInstallOptions) error {
+	// generate host port.
+	hostPort, err := webhookOptions.generateHostPort()
+	if err != nil {
+		return err
+	}
+	url := pointer.StringPtr(fmt.Sprintf("https://%s/convert", hostPort))
```
DirectXMan12, line 357: is conversion always on `/convert`?  Should we be keeping the user's configured path, and just messing with the host-port part?

DirectXMan12 (changes requested):
lets split the "un-exporting" part out into a separate PR for discussion.  otherwise, 1 comment inline.

pkg/envtest/crd.go
vincepri: Yeah by convention it's always `/convert`, see  https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/builder/webhook.go#L139-L141

DirectXMan12: theoretically it could be not convert if you don't go through the builder, but I'm not sure how much we need to worry about that ATM.

pkg/envtest/webhook.go
```diff
@@ -179,10 +180,6 @@ func (o *WebhookInstallOptions) PrepWithoutInstalling() error {
 
 // Install installs specified webhooks to the API server
 func (o *WebhookInstallOptions) Install(config *rest.Config) error {
-	if err := o.PrepWithoutInstalling(); err != nil {
```
DirectXMan12, line 182: maybe make this call conditional so we don't break people if anyone's using it manually

DirectXMan12: (i.e. don't re-call if we've already called it, but if someone's using it standalone still call)

DirectXMan12: if it's not too complicated

DirectXMan12 (approved):
/lgtm
/approve

DirectXMan12 (commented):
/lgtm

PR #1526, opened 2021-05-14
:seedling: Handle unstructured CRD objects in envtest webhook mods

vincepri:

Follow-up to #1525


christopherhein:
Nice 👍 

/lgtm

PR #1527, opened 2021-05-15
🌱 [WIP] Conditional Controllers

kevindelgado:
This is an updated implementation of the design to add conditional controller support from https://github.com/kubernetes-sigs/controller-runtime/pull/1192 

It offers a way to configure controllers to seamlessly start/stop/restart themselves based on whether the CRD it watches exists on the cluster or not. It does so by:

1. Implementing a stop channel on the informers map `MapEntry` that can be
   retrieved from the cache's `Informers` interface that fires when the informer has
   stopped.
2. A new Source is created called `ConditionalSource` that is implemented by a
   `ConditionalKind` that has a `Ready()` method that blocks until the kind's
   type is installed on the cluster and ready to be controlled, and a
   `StartNotifyDone()` that starts the underlying source with a channel that
   fires when the source (and its underlying informer have stopped).
3. Controllers maintain a list of `ConditonalSource`s known as
   `conditionalWatches` that can be added like regular watches via the
   controllers `Watch` method. For any `ConditionalSource` that `Watch()` is
   called with, it uses `Ready()` to determine when to `StartNotifyDone()` it
   and uses the done channel returned by `StartNotifyDone()` to determine when
   the watch has started and that it should wait on `Ready()` again.
4. The manager recognize a special type or `Runnable` known as a
   `ConditionalRunnable` that has a `Ready()` method to indicate when the
   underlying `Runnable` can be ran.
5. The controller builder enables users supply a boolean `conditional` option to
   the builder's `For`, `Owns`, and `Watches` inputs that creates
   `ConditionalRunnables` to be ran by the manager.

Currently it has been manually tested and there is a single integration test that checks the whole CRD install/uninstall/reinstall flow. I intend to add more unit tests for each individual component that has changed, but wanted to get feedback on the overall design first.


coderanger:
I think I'm missing some background here, but what use case is this for? This seems like it would only come up in operators which can do _stuff_ to any specified object type based on runtime inputs? That seems legit but want to clarify since we've gotten questions about this kind of feature before when "two normal controllers that both have an early return check" was the far simpler solution :)

kevindelgado:
> I think I'm missing some background here, but what use case is this for? This seems like it would only come up in operators which can do _stuff_ to any specified object type based on runtime inputs? That seems legit but want to clarify since we've gotten questions about this kind of feature before when "two normal controllers that both have an early return check" was the far simpler solution :)

The use case we are targeting is when the owner of the operator is not an admin of the cluster(s) it is operating on and thus it doesn’t have any control over when CRDs are installed and uninstalled. 

As the owner of the controller/operator, my operator should still run successfully on such a cluster and gracefully transition from a state of reconciling when the CRD I’m operating on is present to waiting for that CRD to be reinstalled when it no longer exists on the cluster (and vice versa).

Running two separate controllers that intentionally stop themselves is not always viable when you’re running multiple operators in a single binary (e.g. multi-cluster, or just sharing a cache for perf reasons like kubernetes’ controller-manager).

tossmilestone:
@kevindelgado The feature is an awsome feature for us, we implemented a same one but with a bad method due to the sealed stop channel of informer.  Hope the PR to be merged soon.

tossmilestone:
@kevindelgado Is the feature will be abandoned? If you are busy, I can help to work on it and get it done.

Issue #1436, opened 2021-03-17
TokenReview.authentication.k8s.io resource helpers for Wehbook Server


christopherhein:
Would Controller Runtime be interested in being able to support non-AdmissionReview based HTTP handlers in the `Webhook` server package. I'm thinking of resources like `authentication.k8s.io.TokenReview` `imagepolicy.k8s.io.ImageReview`. 

EG if I could use:

```go
mgr.GetWebhookServer().Register("/tokenreview", &authentication.Webhook{Handler: &wh.TokenReview{}})
```

And it would then have something like:

```go
import "k8s.io/api/authentication/v1"
// ...
type Request struct {
  v1.TokenReview
}
```

and my `Handler` func could be:

```
func (a *TokenReview) Handle(ctx context.Context, req authentication.Request) authentication.Response {
  // ... implement authentication request
  return authentication.Response{}
}
```

This would help to standardize the building of `ImageReviews` and `TokenReviews` for code bases like https://sigs.k8s.io/aws-iam-authenticator where it's currently just a normal go HTTP server implementation and where you need to add controllers (when we did) we just used `client-go` directly and we could have used CR easier.

coderanger:
We already also support conversion hooks, so this seems fine. I think we should probably keep it scoped to K8s-specific hooks just to avoid getting into generic web framework territory but these should work similarly to the existing hooks.

christopherhein:
Yea, was thinking that as well, as long as it's exposed by core k8s APIs we _could_ support it.

…and agree, avoiding the web framework side since you can always use a new `Runnable`.

alvaroaleman:
> We already also support conversion hooks, so this seems fine. I think we should probably keep it scoped to K8s-specific hooks just to avoid getting into generic web framework territory but these should work similarly to the existing hooks.

So right now, the signature allows you to register anything: `Register(path string, hook http.Handler)`. IMHO that is fine?

Adding helpers to make writing other Kubernetes hook handlers easier sounds like a good idea to me :+1:

christopherhein:
> So right now, the signature allows you to register anything: Register(path string, hook http.Handler). IMHO that is fine?

Yea I ended up implementing that for what I was doing but I'm basically maintaining a near replica of the admissions `ServeHTTP` func and felt like it could be upstreamed. Cool sounds like this might be of use.

coderanger:
[...]

PR #1535, opened 2021-05-20
🌱 adding SubjectAccessReview.authorization.k8s.io/v1 webhook support

rfranzke:
Similar to #1440, this PR adds a new `pkg/webhook/authorization/` package for setting up the `ServeHTTP` func for the webook server to be able to handle `SubjectAccessReview.authorization.k8s.io/(v1,v1beta1)` requests.

It allows third-party programs to focus on their business logic while controller-runtime providers the necessary boilerplate for handling such HTTP requests.

For more background, see also https://github.com/kubernetes-sigs/controller-runtime/issues/1436#issuecomment-801709381.

### DIFF
--- a/pkg/webhook/authorization/http.go
+++ b/pkg/webhook/authorization/http.go
@@ -68,8 +68,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// verify the content type is accurate
-	contentType := r.Header.Get("Content-Type")
-	if contentType != "application/json" {
+	if contentType := r.Header.Get("Content-Type"); contentType != "application/json" {
 		err = fmt.Errorf("contentType=%s, expected application/json", contentType)
 		wh.log.Error(err, "unable to process a request with an unknown content type", "content type", contentType)
 		reviewResponse = Errored(err)
@@ -117,11 +116,7 @@ func (wh *Webhook) writeSubjectAccessReviewResponse(w io.Writer, ar authorizatio
 		wh.log.Error(err, "unable to encode the response")
 		wh.writeResponse(w, Errored(err))
 	}
-	res := ar
-	if log := wh.log; log.V(1).Enabled() {
-		log.V(1).Info("wrote response", "UID", res.UID, "authorized", res.Status.Allowed)
-	}
-	return
+	log.V(1).Info("wrote response", "UID", ar.UID, "authorized", ar.Status.Allowed)
 }
 
 func (wh *Webhook) decodeRequestBody(body []byte) (unversionedSubjectAccessReview, *schema.GroupVersionKind, error) {
--- a/pkg/webhook/authorization/http.go
+++ b/pkg/webhook/authorization/http.go
@@ -55,7 +55,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		err = errors.New("request body is empty")
 		wh.log.Error(err, "bad request")
 		reviewResponse = Errored(err)
-		wh.writeResponse(w, reviewResponse)
+		wh.writeResponse(w, nil, reviewResponse)
 		return
 	}
 
@@ -63,7 +63,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if body, err = ioutil.ReadAll(r.Body); err != nil {
 		wh.log.Error(err, "unable to read the body from the incoming request")
 		reviewResponse = Errored(err)
-		wh.writeResponse(w, reviewResponse)
+		wh.writeResponse(w, nil, reviewResponse)
 		return
 	}
 
@@ -72,7 +72,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		err = fmt.Errorf("contentType=%s, expected application/json", contentType)
 		wh.log.Error(err, "unable to process a request with an unknown content type", "content type", contentType)
 		reviewResponse = Errored(err)
-		wh.writeResponse(w, reviewResponse)
+		wh.writeResponse(w, nil, reviewResponse)
 		return
 	}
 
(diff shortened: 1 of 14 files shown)


pkg/webhook/authorization/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Handler can handle an SubjectAccessReview.
+type Handler interface {
+	// Handle yields a response to an SubjectAccessReview.
+	//
+	// The supplied context is extracted from the received http.Request, allowing wrapping
+	// http.Handlers to inject values into and control cancelation of downstream request processing.
+	Handle(context.Context, Request) Response
+}
```
rfranzke, line 65: In Gardener ([ref](https://github.com/gardener/gardener/blob/7c4388ee8bb0eaecfbae451772dc2577df45cc3f/pkg/admissioncontroller/webhooks/auth/seed/handler.go#L63)), the interface for the handler is [`k8s.io/apiserver/pkg/authorization/authorizer.Authorizer`](https://github.com/kubernetes/apiserver/blob/c13376c1a6cd7f67c1b72f1f96256164d6b592f2/pkg/authorization/authorizer/interfaces.go#L67-L72), i.e.

```golang
type Authorizer interface {
	Authorize(ctx context.Context, a Attributes) (authorized Decision, reason string, err error)
}
```

The reason was that this is also the interface for the in-tree authorizers in k/k.

Do you think it makes sense to also change the `Request` type in this PR to inline `auth.Attributes` instead of the plain `SubjectAccessReview` type?

christopherhein: I personally think it's nice to align them with the rest of controller-runtime webhooks like you've done but @DirectXMan12 might have thoughts here.

christopherhein (approved):
/lgtm

pkg/webhook/authorization/http.go
```diff
@@ -0,0 +1,165 @@
...
+// writeSubjectAccessReviewResponse writes ar to w.
+func (wh *Webhook) writeSubjectAccessReviewResponse(w io.Writer, ar authorizationv1.SubjectAccessReview) {
+	if err := json.NewEncoder(w).Encode(ar); err != nil {
+		wh.log.Error(err, "unable to encode the response")
+		wh.writeResponse(w, Errored(err))
+	}
+	res := ar
+	if log := wh.log; log.V(1).Enabled() {
+		log.V(1).Info("wrote response", "UID", res.UID, "authorized", res.Status.Allowed)
+	}
```
estroz, line 123: Any reason for all this?
```suggestion
	wh.log.V(1).Info("wrote response", "UID", ar.UID, "authorized", ar.Status.Allowed)
```

pkg/webhook/authorization/http.go
```diff
@@ -0,0 +1,165 @@
...
+		ar.SetGroupVersionKind(*subjRevGVK)
+	}
+	wh.writeSubjectAccessReviewResponse(w, ar)
+}
+
+// writeSubjectAccessReviewResponse writes ar to w.
+func (wh *Webhook) writeSubjectAccessReviewResponse(w io.Writer, ar authorizationv1.SubjectAccessReview) {
+	if err := json.NewEncoder(w).Encode(ar); err != nil {
+		wh.log.Error(err, "unable to encode the response")
+		wh.writeResponse(w, Errored(err))
```
estroz, line 118: Unless I'm mistaken this is a recursive call.

christopherhein: Interesting catch, you might be right but also reviewing other webhooks we do the same things in those… https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/authentication/http.go#L135-L138 and https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/http.go#L129-L131 I wonder if we've seen errors from these too and never root caused it 🤔

pkg/webhook/authorization/http.go
christopherhein: For cross-referencing is this what we do in other webhook HTTP handlers, I believe it's cause technically this should work without a logger.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/http.go#L134-L139
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/authentication/http.go#L140-L142

At least that was my thinking when I kept it in the `authentication` webhook.

estroz: I'm pretty sure `Enabled()` exists so you can write another action to take if the logger at some level is not turned on. If you do `wh.log.V(notEnabledInt).Info(...)` the log message won't be written.

pkg/webhook/authorization/http.go
estroz: Unless `Errored(someError)` is guaranteed to encode without error, this could cause issues. If the other webhook impl's are doing the same thing then maybe it is?

pkg/webhook/authorization/http.go
christopherhein: TIL! Thanks, that's super good to know. there seems to be a couple of these places we should/can clean up throughout the entire `pkg/webhook/...` packages then. For the others should we file some issues to make sure they all get cleaned up?

pkg/webhook/authorization/http.go
christopherhein: 🥴 Yeah… they definitely aren't guaranteed since they are all pretty much the same response type so if it was an error with handling of the underlying struct you're right this will definitely loop.

pkg/webhook/authorization/http.go
rfranzke: Do you want me to change this for `authorization` in this PR? Or should we proceed as is for now so that all three occasions (admission, authentication, authorization) can be cleaned up in one go?

pkg/webhook/authorization/http.go
rfranzke: Do you want me to change this for `authorization` in this PR? Or should we proceed as is for now so that all three occasions (admission, authentication, authorization) can be cleaned up in one go?

pkg/webhook/authorization/http.go
estroz: IMO the less changes later, the better (do it here).

pkg/webhook/authorization/http.go
estroz: Refactoring `writeResponse` into `write` or something like that, that both `writeResponse` and `writeSubjectAccessReviewResponse` can call sans recursion, would be nice here.

pkg/webhook/authorization/http.go
rfranzke: done with 2767220 (for all webhook handlers)

rfranzke:
@estroz Could you have another look, please?

rfranzke:
I've rebased and adapted some lines based on the newly introduced linting suggestions. I'd be grateful if somebody could take a look :)

christopherhein:
/remove-lifecycle stale

@vincepri are you able to take a look at this?

christopherhein:
@rfranzke @alvaroaleman what should we do with this?

rfranzke:
Well, I don't know why this PR was not yet approved, especially in the light that it is very similar to #1440.
Can I do something on my end? I would still be happy to get this merged.

/remove-lifecycle rotten

everpeace:
@rfranzke @alvaroaleman @christopherhein 

Although it is closed by stale long time ago, I'm now very interested in this PR. I would like to contribute to getting this PR merged. What can I do??

/reopen

PR #1537, opened 2021-05-21
🌱 Optimize client cache

ash2k:
Currently the client might be created more than once. There is no need to do the work twice, instead just check if the work has been done already after acquiring the write lock.

### DIFF
--- a/pkg/client/client_cache.go
+++ b/pkg/client/client_cache.go
@@ -82,13 +82,13 @@ func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
 	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
 	isUnstructured = isUnstructured || isUnstructuredList
 
-	// It's better to do creation work twice than to not let multiple
-	// people make requests at once
-	c.mu.RLock()
-	resourceByType := c.structuredResourceByType
+	var resourceByType map[schema.GroupVersionKind]*resourceMeta
 	if isUnstructured {
 		resourceByType = c.unstructuredResourceByType
+	} else {
+		resourceByType = c.structuredResourceByType
 	}
+	c.mu.RLock()
 	r, known := resourceByType[gvk]
 	c.mu.RUnlock()
 
@@ -99,6 +99,10 @@ func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
 	// Initialize a new Client
 	c.mu.Lock()
 	defer c.mu.Unlock()
+	r, known = resourceByType[gvk] // check again to handle potential race
+	if known {
+		return r, nil
+	}
 	r, err = c.newResource(gvk, meta.IsListType(obj), isUnstructured)
 	if err != nil {
 		return nil, err


pkg/client/client_cache.go
```diff
@@ -82,13 +82,13 @@ func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
...
-	// It's better to do creation work twice than to not let multiple
-	// people make requests at once
-	c.mu.RLock()
-	resourceByType := c.structuredResourceByType
+	var resourceByType map[schema.GroupVersionKind]*resourceMeta
 	if isUnstructured {
 		resourceByType = c.unstructuredResourceByType
+	} else {
+		resourceByType = c.structuredResourceByType
 	}
```
ash2k, line 90: Field reads don't need to be under the lock as the fields themselves are not mutated, the contents of the maps are mutated. I.e. it's safe to read the map pointer into a variable.

vincepri:
Can we add some benchmarks and tests?

ash2k:
I do not have time for that, sorry.

ash2k:
ok, alright. I'm not sure what should be benchmarked - it's not a performance-sensitive piece of code, is it? If the next thing that's going to happen is a network call then it's not that important if there is an additional `if` inside of the lock. Also, what testing is needed? That the client is not created twice? It's hard to test because it's about concurrency. How to test that?

vincepri:
Hi @ash2k,

Thanks for replying back, I originally asked for a benchmark in response to the PR's description of "There is no need to do the work twice". My ask for a test or benchmark is to make sure we're effectively fixing an issue, I'm not super comfortable moving locks around for code that was written more than 3 years ago (I'm also not super familiar with this part of the codebase).

cc @DirectXMan12

ash2k:
Actually, I've looked at the code once again and I think it's not doing what the old comment is saying - there is no way for "creation work" to happen twice because it's under the write lock (well, yes, without the `if` I'm adding it can do it twice). I think the code is weird and should be simplified even further. What's not clear is what is the desired behavior:

- Comment implies it's better to call `c.newResource` twice than to hold the exclusive lock for too long.
- But it does not work this way, since the lock is held around `c.newResource`.

So, what is the desired behavior? Depending on it the code should be changed.

If we want to do the work twice (i.e. without the lock), it should be:
```go
func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
	gvk, err := apiutil.GVKForObject(obj, c.scheme)
	if err != nil {
		return nil, err
	}

	_, isUnstructured := obj.(*unstructured.Unstructured)
	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
	isUnstructured = isUnstructured || isUnstructuredList

	var resourceByType map[schema.GroupVersionKind]*resourceMeta
	if isUnstructured {
		resourceByType = c.unstructuredResourceByType
	} else {
		resourceByType = c.structuredResourceByType
	}
	c.mu.RLock()
	r, known := resourceByType[gvk]
	c.mu.RUnlock()

	if known {
		return r, nil
	}

	// It's better to do creation work twice than to not let multiple
	// people make requests at once

	// Initialize a new Client
	r, err = c.newResource(gvk, meta.IsListType(obj), isUnstructured)
	if err != nil {
		return nil, err
	}
	c.mu.Lock()
	defer c.mu.Unlock()
	concurrentR, known := resourceByType[gvk]
	if known {
		return concurrentR, nil
	}
	resourceByType[gvk] = r
	return r, nil
}
```

If we want to keep the current behavior of holding the lock and avoiding the double work, then (note that we no longer need `RWLock` in this case):
```go
func (c *clientCache) getResource(obj runtime.Object) (*resourceMeta, error) {
	gvk, err := apiutil.GVKForObject(obj, c.scheme)
	if err != nil {
		return nil, err
	}

	_, isUnstructured := obj.(*unstructured.Unstructured)
	_, isUnstructuredList := obj.(*unstructured.UnstructuredList)
	isUnstructured = isUnstructured || isUnstructuredList

	var resourceByType map[schema.GroupVersionKind]*resourceMeta
	if isUnstructured {
		resourceByType = c.unstructuredResourceByType
	} else {
		resourceByType = c.structuredResourceByType
	}

	c.mu.Lock()
	defer c.mu.Unlock()
	r, known := resourceByType[gvk]
	if known {
		return r, nil
	}
	// Initialize a new Client
	r, err = c.newResource(gvk, meta.IsListType(obj), isUnstructured)
	if err != nil {
		return nil, err
	}
	resourceByType[gvk] = r
	return r, nil
}
```

PR #1539, opened 2021-05-23
⚠ pkg/webhook/admission: use Result.Message instead of Result.Reason

bhcleek:
Use Result.Message instead of Result.Reason for the user supplied
portion. Reason is intended to be machine readable while Message is
intended to be human readable. While each is documented as being
informational only, the Is* family of functions in
k8s.io/apimachinery/pkg/api/errors rely on the StatusReason.

This change allows controllers to rely on
"k8s.io/apimachinery/pkg/api/errors".IsForbidden to deal with errors
consistently regardless of whether the operation failed due to an
admission webhook implemented with controller-runtime, a standard
kubernetes failure (e.g related to RBAC), or another controller not
implemented with controller-runtime.

See kubernetes/kubernetes#101926 for more information

### DIFF
--- a/pkg/webhook/admission/response.go
+++ b/pkg/webhook/admission/response.go
@@ -26,21 +26,21 @@ import (
 
 // Allowed constructs a response indicating that the given operation
 // is allowed (without any patches).
-func Allowed(reason string) Response {
-	return ValidationResponse(true, reason)
+func Allowed(message string) Response {
+	return ValidationResponse(true, message)
 }
 
 // Denied constructs a response indicating that the given operation
 // is not allowed.
-func Denied(reason string) Response {
-	return ValidationResponse(false, reason)
+func Denied(message string) Response {
+	return ValidationResponse(false, message)
 }
 
 // Patched constructs a response indicating that the given operation is
 // allowed, and that the target object should be modified by the given
 // JSONPatch operations.
-func Patched(reason string, patches ...jsonpatch.JsonPatchOperation) Response {
-	resp := Allowed(reason)
+func Patched(message string, patches ...jsonpatch.JsonPatchOperation) Response {
+	resp := Allowed(message)
 	resp.Patches = patches
 
 	return resp
@@ -60,21 +60,24 @@ func Errored(code int32, err error) Response {
 }
 
 // ValidationResponse returns a response for admitting a request.
-func ValidationResponse(allowed bool, reason string) Response {
+func ValidationResponse(allowed bool, message string) Response {
 	code := http.StatusForbidden
+	reason := metav1.StatusReasonForbidden
 	if allowed {
 		code = http.StatusOK
+		reason = ""
 	}
 	resp := Response{
 		AdmissionResponse: admissionv1.AdmissionResponse{
 			Allowed: allowed,
 			Result: &metav1.Status{
-				Code: int32(code),
+				Code:   int32(code),
+				Reason: reason,
 			},
 		},
 	}
-	if len(reason) > 0 {
-		resp.Result.Reason = metav1.StatusReason(reason)
+	if len(message) > 0 {
+		resp.Result.Message = message
 	}
 	return resp
 }
--- a/pkg/envtest/webhook_test.go
+++ b/pkg/envtest/webhook_test.go
@@ -19,6 +19,7 @@ package envtest
 import (
 	"context"
 	"path/filepath"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -83,7 +84,7 @@ var _ = Describe("Test", func() {
 
 			Eventually(func() bool {
 				err = c.Create(context.TODO(), obj)
-				return apierrors.ReasonForError(err) == metav1.StatusReason("Always denied")
+				return err != nil && strings.HasSuffix(err.Error(), "Always denied") && apierrors.ReasonForError(err) == metav1.StatusReasonForbidden
 			}, 1*time.Second).Should(BeTrue())
 
 			cancel()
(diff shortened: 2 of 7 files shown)


coderanger:
Trying to think of ways this compat break could bite someone. We should be very certain we can't think of any.

bhcleek:
> Trying to think of ways this compat break could bite someone. We should be very certain we can't think of any.

I was torn on whether to mark this with ⚠️  or not, but decided to err on the conservative side.

I don't think this will break any controller-runtime clients, but it will cause `"k8s.io/apimachinery/pkg/errors".IsForbidden` calls to match `Deny` responses where previously it would not. This is actually a good thing, because controller code often checks `IsForbidden` (and other functions in the `Is*` family) to decide whether to retry an operation or not.

bhcleek:
Is there anything we can do to move this forward?

pkg/webhook/admission/response.go
```diff
@@ -60,21 +60,24 @@ func Errored(code int32, err error) Response {
...
 			Result: &metav1.Status{
-				Code: int32(code),
+				Code:   int32(code),
+				Reason: reason,
 			},
 		},
 	}
-	if len(reason) > 0 {
-		resp.Result.Reason = metav1.StatusReason(reason)
+	if len(message) > 0 {
```
joelanford, line 79: Nit: Do we need to check the length of message or can we just always set `resp.Result.Message`? `metav1.Status.Message` is a string type, so leaving it "unset" is the same as setting it to an empty string.

joelanford (commented):
I think the piece of this that gives me some pause is that this change is API compatible, but it breaks semantics. So calling code will compile and run just fine, but what's happening may not be what a caller expects (e.g. perhaps a caller _really does_ want to set a custom reason)

I _do_ agree that it makes more sense to expose the message and auto-populate the reason.

I would almost prefer that we find a way to break the Go API so that users are confronted with this breakage.

pkg/webhook/admission/response.go
bhcleek: You're right, we no longer need to check the length here. Is it worth changing or should we keep it as-is if there aren't other changes to be made?

bhcleek (commented):
> I think the piece of this that gives me some pause is that this change is API compatible, but it breaks semantics. So calling code will compile and run just fine, but what's happening may not be what a caller expects (e.g. perhaps a caller really does want to set a custom reason)

> I do agree that it makes more sense to expose the message and auto-populate the reason.

> I would almost prefer that we find a way to break the Go API so that users are confronted with this breakage.

Given that controller-runtime is still pre-1.0, this kind of breaking change in semantics seems appropriate to me, but I'll happily defer to you if you feel strongly. I see a couple of possibilities for addressing your concerns about the changing semantics:
1. break the API by removing the second parameter of `ValidationResponse` entirely so that existing code will fail on compile and introduce a new `ValidationResponseWithReasonAndMessage` to allow both the reason and message to be provided.
2. Keep API compatibility, but introduce `func ValidationResponseWithReason(allowed bool, reason, message string) Response` to allow users to migrate if they really do want to provide a custom reason.

What do you think?

alvaroaleman:
@bhcleek is this still needed now that https://github.com/kubernetes/kubernetes/pull/101926 merged? Like Joe I'd prefer not to introduce breaking changes the compiler can't point out.

bhcleek:
While this isn't strictly needed with the next version of Kubernetes or later, it's still useful for people on the current version of Kubernetes or earlier.

I'd also add, that IMHO this is closer to the expectations that people may have given how this message can be surfaced to users.

bhcleek:
Where do we stand with this? Should we close it or is there something we can do to push forward on the alignment with kubernetes error expectations?

alvaroaleman (approved):
/hold
@joelanford are you good with this?

sbueringer:
I think it would be good if this change is surfaced via compile errors to consumers.

I would suggest:
* Allowed() Response
  * Doesn't need any reason or message. If folks want to set them anyway for an allowed request they can use ValidationResponse
* Denied() Response
  * Essentially a shortcut to return deny with `metav1.StatusReasonForbidden` set
* Patched(patches ...jsonpatch.JsonPatchOperation) Response
  * Usually doesn't need reason/message as it's also an allowed request.
* ValidationResponse(allowed bool, reason, message string) Response
  * Just one simple func where everything can be provided if reason is "" and allowed is false I would set `metav1.StatusReasonForbidden`

I think this would signal clearly to conusmers that the API changed and the new set of funcs should cover relevant use cases.

bhcleek:
What value should be used for the `reason` when `allowed` is set?

bhcleek:
I can make that change. The reason, though, should probably be a `metav1.StatusReason` instead of a `string` so that users cannot easily swap the reason and the message arguments unintentionally. But what value should be used for the `reason` when `allowed` is set?

bhcleek:
Yeah, the few places I need it, `metav1.StatusReasonUnknown` can be used for success (e.g. `Allowed()`)

sbueringer:
> Yeah, the few places I need it, `metav1.StatusReasonUnknown` can be used for success (e.g. `Allowed()`)

Why is a reason needed for Allowed?

bhcleek:
If `ValidationResponse` is to have three parameters as you suggested with `ValidationResponse(allowed bool, reason, message string)`, then an argument for the reason has to be provided.

bhcleek:
After making the changes locally that you suggested, @sbueringer, I'm less convinced that's the direction this should go. Being able to provide the human readable message to `Allowed` and `Denied` is useful and requiring users provide the `reason` to `ValidationResponse` puts more burden on users to understand all the possible `StatusReason` values.

I know you're trying to avoid unexpected breakages, but given that controller-runtime regularly breaks across releases and effectively abuses semantic versioning anyway in order to be able to break backward compatibility without violating semantic versioning coupled with the greater burden that those suggestions would put on users, I'm not sure we should make those changes. 🤔 

Having expressed my skepticism, I want to be clear that I will submit the changes if there's consensus.

sbueringer:
Okay no worries :). Just my opinion, I'm fine either way.

bhcleek:
/remove-lifecycle rotten

@alvaroaleman with only 21 open PRs, I'm wondering if there's anything we can do to move this forward.

PR #1540, opened 2021-05-25
:sparkles: Clean up pkg/internal/testing

DirectXMan12:
This cleans up `pkg/internal/testing`, hosting a series of nested packages into a flatter hierarchy that makes more sense for its current location, and prunes some old, unused functionality.

The old state matches, more-or-less directly, the structure of `sigs.k8s.io/testing_frameworks`.  When we took ownership of that code, we pretty much dropped it in `internal/testing`.  This made it easier to import, but leaves us with a complicated mess that's more difficult to refactor and understand.

We also ended up with some unused functionality -- some of this was due to mismatches between how CR needed to use `testing_frameworks` vs how it was written, other stuff was due to older versions of software that we no longer support (e.g. all the etcd versions we support have HTTP health check end points).

This should make it easier to build on this code to add new features (e.g. support for secure serving).

Part of #1486.


pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -123,9 +133,12 @@ func (s *APIServer) setProcessState() error {
 		return err
 	}
 
-	s.processState.Args, err = internal.RenderTemplates(
-		internal.DoAPIServerArgDefaulting(s.Args), s,
-	)
+	args := s.Args
+	if len(args) == 0 {
+		args = APIServerDefaultArgs
+	}
```
vincepri, line 139: Should we opt to merge the Args given from users with the default one (and potentially overwriting some)?

pkg/internal/testing/controlplane/etcd.go
```diff
@@ -55,48 +57,61 @@ type Etcd struct {
...
 
 // Start starts the etcd, waits for it to come up, and returns an error, if one
 // occoured.
 func (e *Etcd) Start() error {
 	if e.processState == nil {
-		if err := e.setProcessState(); err != nil {
+		if err := e.setState(); err != nil {
 			return err
 		}
 	}
```
vincepri, line 71: Is there any reason why we need to protect this?

Could we put everything inside a sync.Once?

pkg/internal/testing/controlplane/apiserver.go
DirectXMan12: this PR is supposed to generally not change behavior from what happened previously, and this non-merging behavior is what the existing code does.

pkg/internal/testing/controlplane/etcd.go
DirectXMan12: ditto here -- focusing on not changing functionality in this PR

DirectXMan12:
this PR is supposed to generally not change behavior from what happened previously -- just refactor or prune a couple of blatantly unused things.

vincepri (approved):
/approve

PR #1541, opened 2021-05-25
✨ Structured args in Testing

DirectXMan12:
This adds structured arguments to the testing package, deprecating the old style where we do templating in favor of a fluent API with defaults that can be overridden.

This should make it easier for consumers to override individual arguments while not breaking the necessary flags that are required for the API server to function, and should make it easier for us to deal with upcoming changes in the k8s API server that requiring setting flags in some conditions, but not in others that are unrelated to the value of the flag (e.g. certain flags *must* be passed for older versions, but must *NOT* be passed for newer versions).

The new API looks roughly like:

```go
apiServer.Configure().
  Disable("insecure-port").
  Append("vmodule", "httplog.go=10").
  Set("v", "5")
```

There's also a helper to merge the new arguments with the older style ones.  To preserve existing behavior, if the end user explicitly passes a template, we'll consider those to be the "defaults", otherwise we'll use our normal new default logic.

Depends on #1540
Part of #1486

### DIFF
--- a/pkg/internal/testing/controlplane/apiserver.go
+++ b/pkg/internal/testing/controlplane/apiserver.go
@@ -44,6 +44,11 @@ type APIServer struct {
 	//
 	// If not specified, the minimal set of arguments to run the APIServer will
 	// be used.
+	//
+	// They will be loaded into the same argument set as Configure.  Each flag
+	// will be Append-ed to the configured arguments just before launch.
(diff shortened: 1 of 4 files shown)


vincepri (approved):
/lgtm
/retitle ✨ Structured args in Testing

PR #1542, opened 2021-05-26
:seedling: Add missing license headers to files

DirectXMan12:
A few existing files, plus some new files from #1486, were missing license headers, so this fixes that.

Depends on #1486


vincepri:
Need a rebase?

PR #1548, opened 2021-06-07
✨ Allow TLS minimum version to be configured

### LINKED ISSUE #1431: Default TLS implementations leave consuming systems open to multiple vulnerabilities.
The default implementation for TLS endpoints expose the long deprecated TLS 1.0 and TLS 1.1 versions along with dangerous ciphers that expose systems to version downgrade attacks, SWEET32 (CVE-2016-2183, CVE-2016-6329), BEAST (CVE-2011-3389), LUCKY13 (CVE-2013-0169).  There should be a way for consumers to control which versions and ciphers are exposed.

This was confirmed with a default endpoint and tested with version 3.0.4 of testssl.sh.

willthames:
Some environments have automated security scans that trigger
on TLS versions or insecure cipher suites. Setting TLS to 1.3
would solve both problems (setting to 1.2 only solves the former
as the default 1.2 cipher suites are insecure).

Default TLS minimum version of 1.0 remains.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -71,7 +71,7 @@ type Server struct {
 	ClientCAName string
 
 	// TLSVersion is the minimum version of TLS supported. Accepts
-	// "1.1", "1.2" and "1.3" - anything else will result in "1.0"
+	// "", "1.0", "1.1", "1.2" and "1.3" only ("" is equivalent to "1.0" for backwards compatibility)
 	TLSMinVersion string
 
 	// WebhookMux is the multiplexer that handles different webhooks.
@@ -80,10 +80,6 @@ type Server struct {
 	// webhooks keep track of all registered webhooks for dependency injection,
 	// and to provide better panic messages on duplicate webhook registration.
 	webhooks map[string]http.Handler
-
-	// tlsMinVersion is the result of the conversion from human-readable TLS version (for example "1.1")
-	// to the values accepted by tls.Config (for example 0x301)
-	tlsMinVersion uint16
 
 	// setFields allows injecting dependencies from an external source
 	setFields inject.Func
@@ -116,17 +112,6 @@ func (s *Server) setDefaults() {
 
 	if len(s.KeyName) == 0 {
 		s.KeyName = "tls.key"
-	}
-
-	switch s.TLSMinVersion {
-	case "1.1":
-		s.tlsMinVersion = tls.VersionTLS11
-	case "1.2":
-		s.tlsMinVersion = tls.VersionTLS12
-	case "1.3":
-		s.tlsMinVersion = tls.VersionTLS13
-	default:
-		s.tlsMinVersion = tls.VersionTLS10
 	}
 }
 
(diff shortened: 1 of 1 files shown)


willthames:
One could argue that only 1.3 makes any real sense but I've gone with the same TLS version options as go's crypto package

coderanger:
/ok-to-test

@willthames You'll need to rebase/amend to remove the `Fixes` from the commit message. Because of how Kubernetes vendoring works, it can get noisy to leave those in the commits themselves.

As for default version, I would be fine setting it to 1.3 if all recent kube-apiservers support that. That's the only thing which actually matters as a client.

willthames:
See #1431 for related issue

willthames:
@coderanger TLS 1.3 support was enabled by default with go 1.13 (https://golang.org/doc/go1.13#tls_1_3) but was available in go 1.12 through a GODEBUG setting. 

Looks like kubernetes builds started to use go 1.13 from kubernetes 1.17 (from https://github.com/kubernetes/kubernetes/commit/e3ff39ffa068bbd1d9808c54923e527248e4fa87)

AWS still supports 1.16 but kubernetes itself doesn't.

If you're happy for me to make 1.3 the default I can make that change (it'll make this PR a lot simpler!)

coderanger:
I would want to make sure that's okay with OperatorSDK crew because they have to support older stuff than the rest of us (usually) but I think 1.2 as the default would definitely make sense no matter what, and maybe 1.3 if we can swing it. I think the overall structure of your change is good though, just playing with which is the default value.

pkg/webhook/server.go
```diff
@@ -109,6 +117,17 @@ func (s *Server) setDefaults() {
...
 	}
+
+	switch s.TLSMinVersion {
+	case "1.1":
+		s.tlsMinVersion = tls.VersionTLS11
+	case "1.2":
+		s.tlsMinVersion = tls.VersionTLS12
+	case "1.3":
+		s.tlsMinVersion = tls.VersionTLS13
+	default:
```
alvaroaleman, line 128: IMHO we should error if `TLSMinVersion` is set but not in the list, silently picking an old version is not a good idea IMHO

alvaroaleman:
Also nit, but I don't think this qualifies as a bugfix, a bug is IMHO something that was already built and supposed to work but didn't, since we never head support for configuring the TLS version this is IMHO a feature.

willthames:
@alvaroaleman the description for bug also says patch (although I guess any change is a patch), and there is an issue raised for this - that was my reasoning but I'm not against changing it to feature

pkg/webhook/server.go
willthames: Do you mind suggesting a mechanism for this (setDefaults doesn't currently have any error handling so I'm not sure what pattern to follow)

willthames:
@alvaroaleman changed title to use ✨

pkg/webhook/server.go
alvaroaleman: I would just move this into `Start`. The only thing that really qualifies as defaulting is to set it to `tls.VersionTLS10` if s.TLSMinVersion is an empty string

willthames: thanks for that suggestion - hopefully this iteration does what you're after.

alvaroaleman (approved):

alvaroaleman:
thanks!

coderanger:
@alvaroaleman Is there a reason we didn't make 1.2 the default?

alvaroaleman:
> @alvaroaleman Is there a reason we didn't make 1.2 the default?

Yeah, to not change what is currently happening, golang defaults to 1.0:

```
 // If zero, TLS 1.0 is currently taken as the minimum.
```

coderanger:
Sure but that seems like a bad default because kube-apiserver should support 1.2 in all cases? 1.3 was questionable but I don't think there's a reason to not pin up to 1.2 to block some funky MITM attacks?

alvaroaleman:
> Sure but that seems like a bad default because kube-apiserver should support 1.2 in all cases? 1.3 was questionable but I don't think there's a reason to not pin up to 1.2 to block some funky MITM attacks?

So I am not opposed to changing the default, golang defaulting to 1.0 is quite terrible actually. It just happens to be a change that _might_ cause issues (I don't know if it does) and this change right here seemed to me just about adding a knob.

coderanger:
Legit, I'll open a new PR to debate changing the default :)

PR #1551, opened 2021-06-07
:book: Remove reference to ClientBuilder in Cluster docs

DirectXMan12:
The Cluster interface docs still had a reference to client, despite
being removed in aa1bfee605cea988a4a79e9021ad1d4334b09eb8.

This restores those docs to talk about NewClient instead.

Follow up from #1409


vincepri (approved):
/approve
/lgtm

PR #1552, opened 2021-06-07
:book: Mark all templated flag pieces as deprecated

DirectXMan12:
The underlying templated flag machinery was marked as deprecated as part
of the envtest refactor (see ddfdfdff7), but a couple fields in the
higher-level bits of envtest were not marked as deprecated.  This fixes
that.

See also #1541


vincepri (approved):
/approve
/lgtm

PR #1562, opened 2021-06-22
:bug: Remove advertise-address from testenv api-server

vincepri:
Kubernetes 1.21 doesn't let us use a loopback address anymore, we can
let the api-server to figure out its own advertise address by removing
the argument.


coderanger:
/lgtm

Mildly annoying for security but that's a battle for upstream.

PR #1563, opened 2021-06-22
🌱  addr.Suggest should lock a file instead of memory

vincepri:
Envtest is often running in parallel when using go test, which spins up
multiple indipendent go test processes that cannot talk to each other.
The address suggestion code, mostly used to find an open port, can
cause port collisions and a race condition between different envtests
running at the same time.

This change switches the internal memory to use a file based system that
creates a file.

### DIFF
--- a/pkg/internal/testing/addr/manager.go
+++ b/pkg/internal/testing/addr/manager.go
@@ -18,78 +18,109 @@ package addr
 
 import (
 	"fmt"
+	"io/fs"
 	"net"
-	"sync"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/internal/flock"
 )
 
 // TODO(directxman12): interface / release functionality for external port managers
 
 const (
-	portReserveTime   = 1 * time.Minute
+	portReserveTime   = 10 * time.Minute
 	portConflictRetry = 100
+	portFilePrefix    = "port-"
 )
 
-type portCache struct {
-	lock  sync.Mutex
-	ports map[int]time.Time
+var (
+	cacheDir string
+)
+
+func init() {
+	baseDir, err := os.UserCacheDir()
+	if err != nil {
+		baseDir = os.TempDir()
+	}
+	cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
+	if err := os.MkdirAll(cacheDir, 0750); err != nil {
+		panic(err)
+	}
 }
 
-func (c *portCache) add(port int) bool {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-	// remove outdated port
-	for p, t := range c.ports {
-		if time.Since(t) > portReserveTime {
-			delete(c.ports, p)
+type portCache struct{}
+
+func (c *portCache) add(port int) (bool, error) {
+	// Remove outdated ports.
+	if err := fs.WalkDir(os.DirFS(cacheDir), ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
 		}
+		if d.IsDir() || !d.Type().IsRegular() || !strings.HasPrefix(path, portFilePrefix) {
+			return nil
+		}
+		info, err := d.Info()
+		if err != nil {
+			return err
+		}
+		if time.Since(info.ModTime()) > portReserveTime {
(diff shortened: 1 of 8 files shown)


coderanger:
This would help specifically for parallel testing of one package but not multiple packages, right? Since those each get their own entrypoint binary so would be independent tempdirs?

vincepri:
Are you referring to darwin?

alvaroaleman:
> Envtest is often running in parallel when using go test, which spins up
multiple indipendent go test processes

TILd

pkg/internal/testing/addr/manager.go
```diff
@@ -28,33 +32,68 @@ import (
...
+		if d.IsDir() || !d.Type().IsRegular() || !strings.HasPrefix(path, portFilePrefix) {
+			return nil
+		}
+		info, err := d.Info()
+		if err != nil {
+			return err
+		}
+		if time.Since(info.ModTime()) > portReserveTime {
+			if err := os.Remove(filepath.Join(cacheDir, path)); err != nil {
+				panic(err)
```
alvaroaleman, line 72: the func returns an error, why not just return the err?

pkg/internal/testing/addr/manager.go
```diff
@@ -28,33 +32,68 @@ import (
...
+	cacheDir string
+)
+
+func init() {
+	baseDir, err := os.UserCacheDir()
+	if err != nil {
+		baseDir = os.TempDir()
+	}
+	cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
+	_ = os.MkdirAll(cacheDir, 0750)
```
alvaroaleman, line 48: maybe `panic`?

pkg/internal/testing/addr/manager.go
```diff
@@ -28,33 +32,68 @@ import (
...
 type portCache struct {
-	lock  sync.Mutex
-	ports map[int]time.Time
+	lock sync.Mutex
 }
 
-func (c *portCache) add(port int) bool {
+func (c *portCache) add(port int) (bool, error) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
```
alvaroaleman, line 57: Do we still need the lock if we lock through the fs?

pkg/internal/testing/addr/manager.go
vincepri: Oops, forgot during testing

pkg/internal/testing/addr/manager.go
vincepri: I think so, mostly because of the WalkDir and the fact that envtest could be spun up multiple times within a single package

coderanger:
> Are you referring to darwin?

Looks like you took care of it with UserCacheDir, nice :)

christopherhein:
Does this stop one test process from stopping envtest on test completion causing flakes across multiple packages?

pkg/internal/testing/addr/manager.go
alvaroaleman: But the whole point of this change is to make the locking work cross-binary so this must be threadsafe without a lock anyways or not?

vincepri:
> Does this stop one test process from stopping envtest on test completion causing flakes across multiple packages?

This stops a race condition that could happen in the event that the same port is allocated and closed quickly between two processes, where one end up not starting because either etcd or api-server is reusing a port.

pkg/internal/testing/addr/manager.go
vincepri: PTAL

alvaroaleman (approved):

Porges:
@vincepri We just updated to `controller-runtime@v0.9.2` and it looks like this code is causing problems for us. We run envtest in parallel and are getting test timeouts with goroutines sitting in `unix.Flock`; I am wondering if the acquire call should be made with `LOCK_NB|LOCK_EX` to ensure it doesn’t block?

vincepri:
@Porges Thanks for the report, running some tests locally first before opening a PR

Porges:
@vincepri at the moment we don’t limit it (and use `t.Parallel()`), so potentially up to 19 instances. An example [failure is here](https://github.com/Azure/azure-service-operator/pull/1627/checks?check_run_id=3119089026) and downgrading to v0.9.1 fixed it.

PR #1567, opened 2021-06-22
🌱 Limit parallel go test executions on MacOS

vincepri:


christopherhein:
/lgtm

Do we have another issue to track the root cause so we can leverage parallelism again at some point? If not can we get one filed.

PR #1570, opened 2021-06-27
🌱  Pass Object's GVK version when reconcile.

haosdent:
This PR tries to pass the Object's GVK info when reconciling resources in the controller. This would be helpful when the controller is watching multiple resources.

e.g. Suppose we have a CRD `PodHealthySet`, which status contains `ReadyPods` as a counter to record how many `Pod`s are healthy under a specific selector.

In our `PodHealthySet Controller`, we watch `PodHealthySet` and `Pod` resources together.
When Pod healthy status changed, we would like to update `PodHealthySet.Status.ReadyPods`.

If we use `EnqueueRequestsFromMapFunc`, to count `PodHealthySet.Status.ReadyPods`, request chain is

1. Find all matched `PodHealthySet` for a specific `Pod` in `EnqueueRequestsFromMapFunc`, then we enqueue `PodHealthySet` reconcile request.
2. Go through all Pods in `PodHealthySet.Reconcile`, and calculate `ReadyPods`. => This would become slower as more and more `Pod` and `PodHealthySet` objects.

If we support pass object's GVK when reconcile, we could identify reconcile from which kind of object, and then don't need the slow Step 2 above.

### DIFF
--- a/pkg/handler/enqueue.go
+++ b/pkg/handler/enqueue.go
@@ -41,25 +41,34 @@ func (e *EnqueueRequestForObject) Create(evt event.CreateEvent, q workqueue.Rate
 		enqueueLog.Error(nil, "CreateEvent received with no metadata", "event", evt)
 		return
 	}
-	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
-		Name:      evt.Object.GetName(),
-		Namespace: evt.Object.GetNamespace(),
-	}})
+	q.Add(reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      evt.Object.GetName(),
+			Namespace: evt.Object.GetNamespace(),
+		},
+		GroupVersionKind: evt.Object.GetObjectKind().GroupVersionKind(),
+	})
 }
 
 // Update implements EventHandler.
 func (e *EnqueueRequestForObject) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
 	switch {
 	case evt.ObjectNew != nil:
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
-			Name:      evt.ObjectNew.GetName(),
-			Namespace: evt.ObjectNew.GetNamespace(),
-		}})
+		q.Add(reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      evt.ObjectNew.GetName(),
+				Namespace: evt.ObjectNew.GetNamespace(),
+			},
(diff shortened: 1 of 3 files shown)


haosdent:
@mengqiy @pwittrock @gerred Could you help to take a look if this PR makes sense? I would fix fail test cases if you are OK with this appoarch.

coderanger:
This feature is very specifically eschewed from controller-runtime as controllers watching multiple root objects is both an anti-pattern and about as impossible as we can make it (ex. calling `For()` twice is disallowed). `EnqueueRequestsFromMapFunc` has to map things back to the root type of the controller in all cases.

haosdent:
Hi, @coderanger thanks a lot for your comment. We understand it is an anti-pattern, but we still would like to find an approach to address performance issue.

coderanger:
Why do you think there is a performance issue?

haosdent:
Hi, @coderanger the performance issue is because 
> Go through all Pods in PodHealthySet.Reconcile, and calculate ReadyPods. => This would become slower as more and more Pod and PodHealthySet objects.

coderanger:
@haosdent While technically yes, that is O(N) operations, they are extremely fast because it's working with a local cache of the objects. Do you have any evidence of actual performance issues causing problems in real use? I don't think we're interested in extremely footgun-y options to deal with hypothetical cases :) You can also cache one direction of the association in ownership references or annotations pretty easily, or just keep a local in-memory cache of it. I think all of those would be a better option than breaking our "stateless reconcile" model.

FillZpp:
> If we support pass object's GVK when reconcile, we could identify reconcile from which kind of object, and then don't need the slow Step 2 above.

@haosdent The question is, even if you can pass `Pod` into queue, what do you plan to do in reconcile? You may get all related `PodHealthySet` for the `Pod` got from queue, then get all matched `Pods` for each of them and calculate their `ReadyPods`?

But there will be a race condition that a same `PodHealthySet` may be handled in multiple workers at once, if the `MaxConcurrentReconciles` is more than `1`. It goes against the design of k8s controller model.

haosdent:
Hi, @FillZpp thx a lot for your comment. 
For the example I mentioned above, when I am able to get the object's GKV, I could distinguish the reconcile is from PodHealthySet or from Pod.
If the object is PodHealthySet, I could `PodHealthySet.Status.ReadyPods++` or `PodHealthySet.Status.ReadyPods--` according to the Pod status. Would not have race condition at here.
The current situation is I able to get the object name, while I could not get the object type. This looks not make sense.

haosdent:
@coderanger Sry for missing your comment. Although we have the local cache, when there are 200k pods, it would become very slow.

PR #1579, opened 2021-07-01
🐛manager recorderProvider only used by resource lock use spec makeBroadcaster to overload cluster's makeBroadcaster

niuyueyang1996:
manager recorderProvider only used by resource lock use spec makeBroadcaster to overload cluster's makeBroadcaster
use spec managerBroadcaster to replace cluster's makeBroadcaster


niuyueyang1996:
I signed it

PR #1580, opened 2021-07-01
🐛manager recorderProvider only used by resource lock

### LINKED ISSUE #1576: when i use mgr.GetEventRecorderFor will send event multiple times

niuyueyang1996:
use spec makeBroadcaster to overload cluster's makeBroadcaster

if i use EventBroadcaster options to set event qps/BurstSize like
```
		Scheme:                 scheme,
		MetricsBindAddress:     o.MetricsAddr,
		Port:                   9443,
		HealthProbeBindAddress: o.ProbeAddr,
		LeaderElection:         o.EnableLeaderElection,
		LeaderElectionID:       o.LeaderElectionID,
		DryRunClient:           o.DryRun,
		SyncPeriod:             &o.SyncPeriod,
		EventBroadcaster: record.NewBroadcasterWithCorrelatorOptions(record.CorrelatorOptions{
			BurstSize: 50,
			QPS:       15,
		}),
	})
```
then makeBroadcaster options will return EventBroadcaster,will not build a NewBroadcaster
```
if options.EventBroadcaster == nil {
		// defer initialization to avoid leaking by default
		options.makeBroadcaster = func() (record.EventBroadcaster, bool) {
			return record.NewBroadcaster(), true
		}
	} else {
		options.makeBroadcaster = func() (record.EventBroadcaster, bool) {
			return options.EventBroadcaster, false
		}
	}
```
cluster.go/manager.go all use the same EventBroadcaster to init newRecorderProvider
will two times exec StartEventWatcher 
then will send event multiple times

but manager recorderProvider only used by resource lock
use makeBroadcaster to overload cluster's makeBroadcaster

fixes #1576


niuyueyang1996:
@droot 
can you review this pr?

BenjaminHuang:
@varshaprasad96 could we  remove tide/merge-method-squash label for now?


@droot @shawn-hurley does it require further review?

BenjaminHuang:
@alvaroaleman could we  remove tide/merge-method-squash label for now?

@droot @shawn-hurley does it require further review?

BenjaminHuang:
@droot @shawn-hurley does it require further review?

/reopen

vincepri:
Closing due to inactivity

/close

vincepri:
@niuyueyang1996 Feel free to reopen, I'm not sure what the purpose of the PR is and what's trying to fix looking at the code changes, please update the description once you're ready

niuyueyang1996:
/reopen
I added a more detailed description with thie pr
@vincepri

pkg/manager/manager.go
```diff
@@ -580,6 +580,14 @@ func defaultBaseContext() context.Context {
 	return context.Background()
 }
 
+// NewProvider use spec managerBroadcaster to replace cluster's makeBroadcaster.
+func NewProvider(config *rest.Config, httpClient *http.Client, scheme *runtime.Scheme, logger logr.Logger, makeBroadcaster intrec.EventBroadcasterProducer) (*intrec.Provider, error) {
+	managerBroadcaster := func() (record.EventBroadcaster, bool) {
+		return record.NewBroadcaster(), true
+	}
+	return intrec.NewProvider(config, httpClient, scheme, logger, managerBroadcaster)
+}
```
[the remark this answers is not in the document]
niuyueyang1996:
```
pkg/manager/manager.go
setOptionsDefaults()
if options.EventBroadcaster == nil {
		// defer initialization to avoid leaking by default
		options.makeBroadcaster = func() (record.EventBroadcaster, bool) {
			return record.NewBroadcaster(), true
		}
	} else {
		options.makeBroadcaster = func() (record.EventBroadcaster, bool) {
			return options.EventBroadcaster, false
		}
	}
```
i see options can use EventBroadcaster.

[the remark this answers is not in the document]
vincepri: Could we actually add a test first that shows the issue you're referring to? I'm still not sure how does the change above actually solve the issue

pkg/internal/controller/controller.go
```diff
@@ -246,19 +246,17 @@ func (c *Controller) Start(ctx context.Context) error {
...
 	}
-
 	// We call Done here so the workqueue knows we have finished
 	// processing this item. We also must remember to call Forget if we
 	// do not want this work item being re-queued. For example, we do
 	// not call Forget if a transient error occurs, instead the item is
 	// put back on the workqueue and attempted again after a back-off
 	// period.
-	defer c.Queue.Done(obj)
-
```
vincepri, line 261: Revert this change please

pkg/internal/controller/controller.go
```diff
@@ -246,19 +246,17 @@ func (c *Controller) Start(ctx context.Context) error {
 // attempt to process it, by calling the reconcileHandler.
 func (c *Controller) processNextWorkItem(ctx context.Context) bool {
 	obj, shutdown := c.Queue.Get()
+	c.GetLogger().V(7).Info("queue get obj", "obj", obj, "shutdown", shutdown)
 	if shutdown {
 		// Stop working
 		return false
 	}
-
```
vincepri, line 253: Revert

example_test.go
```diff
@@ -62,6 +64,54 @@ func Example() {
...
+			}
+		}
+	}()
+	if err := manager.Start(ctrl.SetupSignalHandler()); err != nil {
+		log.Error(err, "could not start manager")
+		os.Exit(1)
+	}
+	log.Info("success")
+}
+
```
niuyueyang1996, line 114: if not fix and use the example.
will see this log.
one times 'eventf end' log and two times 'EventBroadcaster test' log.
```
I0221 11:40:46.893743   93446 leaderelection.go:258] successfully acquired lease test1/123456qbcdqe
2023-02-21T11:40:46+08:00	DEBUG	events	recorder/recorder.go:105	niuyueyangdeMacBook-Pro.local_28a6f8b8-3c7a-478c-afde-7d46495999fc became leader	{"type": "Normal", "object": {"kind":"Lease","namespace":"test1","name":"123456qbcdqe","uid":"5f7083cc-315a-4888-bacd-66e9b0e6fe78","apiVersion":"coordination.k8s.io/v1","resourceVersion":"14055689"}, "reason": "LeaderElection"}



2023-02-21T11:40:56+08:00	DEBUG	event-examples	main/main.go:58	eventf end
2023-02-21T11:40:56+08:00	DEBUG	events	recorder/recorder.go:105	EventBroadcaster test	{"type": "Normal", "object": {"kind":"Pod","namespace":"default","name":"example","apiVersion":"v1"}, "reason": "event-test"}
2023-02-21T11:40:56+08:00	DEBUG	events	recorder/recorder.go:105	EventBroadcaster test	{"type": "Normal", "object": {"kind":"Pod","namespace":"default","name":"example","apiVersion":"v1"}, "reason": "event-test"}
```

vincepri: Thanks for the example, I'm wondering though if we can add a failing test to the manager that makes sure events don't get duplicated

pkg/manager/manager_test.go
```diff
@@ -1739,6 +1780,27 @@ func (c *startClusterAfterManager) GetCache() cache.Cache {
 	return c.informer
 }
 
+type FakeEventBroadcaster struct {
+	once sync.Once
+	record.EventBroadcaster
+}
+
```
niuyueyang1996, line 1787: panic test.
if call StartRecordingToSink func many times,will create/patch event duplicate.
if duplicate,will panic.

pkg/manager/manager.go
```diff
@@ -632,6 +632,15 @@ func defaultBaseContext() context.Context {
 	return context.Background()
 }
 
+// NewProvider replace makeBroadcaster to create NewBroadcaster avoid conflict with cluster's makeBroadcaster
+// because manager's provider only used by leader election.
+func NewProvider(config *rest.Config, httpClient *http.Client, scheme *runtime.Scheme, logger logr.Logger, makeBroadcaster intrec.EventBroadcasterProducer) (*intrec.Provider, error) {
```
vincepri, line 637: Can this be a private function?

niuyueyang1996: yes,already replace

pkg/manager/manager.go
```diff
@@ -641,7 +650,7 @@ func setOptionsDefaults(options Options) Options {
 
 	// Allow newRecorderProvider to be mocked
 	if options.newRecorderProvider == nil {
-		options.newRecorderProvider = intrec.NewProvider
+		options.newRecorderProvider = newProvider
```
sbueringer, line 653: To be honest. Not sure if this is the right way to solve this problem.

With this PR, none of the implementations of the `newRecorderProvider` will use the handed over `makeBroadcaster` parameter.
This means as far as I can tell that options.makeBroadcaster is never used now (for the manager)

If that is in fact what we want to do, why not drop the makeBroadcaster parameter from newRecorderProvider and the makeBroadcaster field from the options and do this

```go
	// Allow newRecorderProvider to be mocked
	if options.newRecorderProvider == nil {
		options.newRecorderProvider = func(config *rest.Config, httpClient *http.Client, scheme *runtime.Scheme, logger logr.Logger) (*intrec.Provider, error) {
			return intrec.NewProvider(config, httpClient, scheme, logger, func() (record.EventBroadcaster, bool) {
				return record.NewBroadcaster(), true
			})
		}
	}
```

Is that actually the new behavior we want? My implementation is equivalent to the current implementation just simpler.

It seems the effect is basically that the manager is not using the passed in EventBroadcaster anymore?

example_test.go
```diff
@@ -62,6 +64,54 @@ func Example() {
...
+				return
+			case <-time.After(time.Second * 10):
+				er := manager.GetEventRecorderFor("event-test")
+				// Like this,will push the same event two times.
+				er.Eventf(examplePod, corev1.EventTypeNormal, "event-test", "EventBroadcaster test")
+				log.V(1).Info("eventf end")
+			}
+		}
+	}()
+	if err := manager.Start(ctrl.SetupSignalHandler()); err != nil {
```
sbueringer, line 108: This example seems to panic because ctrl.SetupSignalHandler() is called twice?

pkg/manager/manager.go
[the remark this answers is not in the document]
sbueringer: @niuyueyang1996 with you're code we still use the EventBroadCaster to build this function and set it in makeBroadcaster, but makeBroadcaster is not used anymore

pkg/manager/manager.go
sbueringer: Do we know why the current implementation produces duplicate events?

sbueringer: Trying to summarize (https://github.com/kubernetes-sigs/controller-runtime/issues/1576#issuecomment-871987952)

In the case that an EventBroadcaster is passed in the following happens:
* we create 1 recorderProvider with this EventBroadCaster in manager
* we create 1 recorderProvider with this EventBroadCaster in cluster
* The result is that on the passed in EventBroadCaster we end up with duplicate watches
* Because of this every event will be sent twice

The current solution in this PR is to just use a separate broadcaster (record.NewBroadcaster()) in manager instead of the passed in broadcaster.

I have some doubts if just ignoring the passed in EventBroadcaster for the manager is the right way to solve this.

Wouldn't it be better to ensure we just have one recorderProvider that we use in both the manager and the cluster?



Side note: let's imagine we implement some sort of multi cluster support and then we have more than one cluster. I think we probably would get the same issue again.

niuyueyang1996: > Trying to summarize ([#1576 (comment)](https://github.com/kubernetes-sigs/controller-runtime/issues/1576#issuecomment-871987952))
> 
> In the case that an EventBroadcaster is passed in the following happens:
> 
> * we create 1 recorderProvider with this EventBroadcaster in manager
> * we create 1 recorderProvider with this EventBroadcaster in cluster
> * The result is that on the passed in EventBroadcaster we end up with duplicate watches
> * Because of this every event will be sent twice
> 
> The current solution in this PR is to just use a separate broadcaster (record.NewBroadcaster()) in manager instead of the passed in broadcaster.
> 
> I have some doubts if just ignoring the passed in EventBroadcaster for the manager is the right way to solve this.
> 
> Wouldn't it be better to ensure we just have one recorderProvider that we use in both the manager and the cluster?
> 
> Side note: let's imagine we implement some sort of multi cluster support and then we have more than one cluster. I think we probably would get the same issue again.

i want to set event qps/burst,so need EventBroadcaster options.

example_test.go
niuyueyang1996: > This example seems to panic because ctrl.SetupSignalHandler() is called twice?

This example to test event broadcaster to set qps and burst size.
panic test is in pkg/manager/manager_test.go.

inteon: @niuyueyang1996 you are calling SetupSignalHandler again on line 98

sbueringer:
/close

Given no recent activity. Feel free to reopen if you're still interested in continuing this work.

PR #1582, opened 2021-07-07
:warning: Fakeclient: Reject Delete with mismatched ResourceVersion

2uasimojo:
`Delete` takes optional varargs comprising `DeleteOption`s. Previously fakeclient was validating that these could be run through `ApplyOptions`, but otherwise ignoring them.

With this commit, fakeclient will observe a `DeleteOption` specifying a `Precondition.ResourceVersion`, rejecting the deletion if the value does not match what's on the "server", bringing fakeclient's behavior more in line with a real client.

Note that, while technically a breaking change, any test that was relying on the previous behavior was already broken.

xref #832 / 25ef372 which did something similar for `Create` and `Update`. Major difference being that those methods do `ResourceVersion` matching by default, whereas `Delete` only does it if opted in via the `Precondition`.

/assign @vincepri 
/assign @mengqiy 
/cc @alvaroaleman
/cc @djzager

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -568,9 +568,50 @@ var _ = Describe("Fake client", func() {
 			Expect(obj.ObjectMeta.ResourceVersion).To(Equal(trackerAddResourceVersion))
 		})
 
-		It("should be able to Delete", func() {
+		It("should reject Delete with a mismatched ResourceVersion", func() {
+			bogusRV := "bogus"
+			By("Deleting with a mismatched ResourceVersion Precondition")
+			err := cl.Delete(context.Background(), dep, client.Preconditions{ResourceVersion: &bogusRV})
+			Expect(apierrors.IsConflict(err)).To(BeTrue())
+
+			list := &appsv1.DeploymentList{}
+			err = cl.List(context.Background(), list, client.InNamespace("ns1"))
+			Expect(err).To(BeNil())
+			Expect(list.Items).To(HaveLen(2))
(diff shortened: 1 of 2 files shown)


alvaroaleman (commented):
/ok-to-test

2uasimojo:
> 2uasimojo force-pushed the 2uasimojo:fake-delete-ResourceVersion branch from 49338fb to e8c6d31 now

Updated the message to match what happens IRL. (This probably doesn't matter, test-wise, but I thought it would be nice to have parity with the real behavior as much as possible.)

pkg/client/fake/client_test.go
```diff
@@ -568,7 +568,33 @@ var _ = Describe("Fake client", func() {
...
+			list := &appsv1.DeploymentList{}
+			err = cl.List(context.Background(), list, client.InNamespace("ns1"))
+			Expect(err).To(BeNil())
+			Expect(list.Items).To(HaveLen(2))
+			Expect(list.Items).To(ConsistOf(*dep, *dep2))
+		})
+
+		It("should successfully Delete with a matching ResourceVersion", func() {
+			goodRV := trackerAddResourceVersion
+			By("Deleting with a mismatched ResourceVersion Precondition")
```
alvaroaleman, line 586:
```suggestion
			By("Deleting with a matching ResourceVersion Precondition")
```

2uasimojo: whoops, copypasta, good catch.

alvaroaleman (approved):
thanks!

PR #1586, opened 2021-07-09
:warning: Drop rest client request latecy metric

### LINKED ISSUE #1423: Disable rest client latency metric
We vendor in `controller-runtime` and use it to expose its controller specific metrics in addition to our own.
Recently we observed that [rest_client_request_latency_seconds](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/client_go_adapter.go#L55-L60) was eating up a significant portion of Prometheus storage, as can be seen from a snapshot of the highest cardinality metrics in our codebase:
```
42765 kube_secret_type
42799 kube_secret_metadata_resource_version
461312 rest_client_request_latency_seconds_count
461312 rest_client_request_latency_seconds_sum
5074432 rest_client_request_latency_seconds_bucket
```

I understand this metric was deprecated in kube 1.14 and turned off in [1.17](https://github.com/kubernetes/kubernetes/pull/83836).
Is it possible to either disable this metric or to allow us to unregister it?

2uasimojo:
Remove the `rest_client_request_latency_seconds` metric.

This metric has been deprecated since kube 1.14 and [turned off in kube 1.17](https://github.com/kubernetes/kubernetes/pull/83836).

fixes #1423

Co-Authored-By: @suhanime

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,8 +18,6 @@ package metrics
 
 import (
 	"context"
-	"net/url"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	reflectormetrics "k8s.io/client-go/tools/cache"
@@ -52,13 +50,6 @@ const (
 
 var (
 	// client metrics.
-	requestLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
-		Subsystem: RestClientSubsystem,
-		Name:      LatencyKey,
-		Help:      "Request latency in seconds. Broken down by verb and URL.",
-		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 10),
-	}, []string{"verb", "url"})
-
 	requestResult = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Subsystem: RestClientSubsystem,
 		Name:      ResultKey,
@@ -127,13 +118,11 @@ func init() {
 // registerClientMetrics sets up the client latency metrics from client-go.
 func registerClientMetrics() {
 	// register the metrics with our registry
-	Registry.MustRegister(requestLatency)
 	Registry.MustRegister(requestResult)
 
 	// register the metrics with client-go
 	clientmetrics.Register(clientmetrics.RegisterOpts{
-		RequestLatency: &latencyAdapter{metric: requestLatency},
-		RequestResult:  &resultAdapter{metric: requestResult},
+		RequestResult: &resultAdapter{metric: requestResult},
 	})
 }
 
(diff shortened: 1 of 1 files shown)


logicalhan (commented):
This is going to break a bunch of people who alert off of this metric, is that going to be okay?

2uasimojo:
> This is going to break a bunch of people who alert off of this metric, is that going to be okay?

This is where it would help me to understand what the actual effect of https://github.com/kubernetes/kubernetes/pull/83836 was. Wouldn't *that* have already broken such consumers?

alvaroaleman (commented):
/ok-to-test

logicalhan:
> This is where it would help me to understand what the actual effect of [kubernetes/kubernetes#83836](https://github.com/kubernetes/kubernetes/pull/83836) was. Wouldn't _that_ have already broken such consumers?

We actually have a deprecation policy, we announce a metric deprecation, we use our prometheus wrapper to annotate our metrics, our framework, on a version bump will auto-hide the metric (which yes breaks consumers). However, they can re-enable it for a release in order to properly migrate off of it. After that, the metric is deleted. 

We offer a staged way to softly push people from using a metric that we intend to delete, in this PR, it'll just abruptly break people...

alvaroaleman:
> This is going to break a bunch of people who alert off of this metric, is that going to be okay?

@logicalhan is that an assumption or do you personally know of a case where someone uses this? The metric includes the whole url as a label, which is obviously a cardinality problem

logicalhan:
> @logicalhan is that an assumption or do you personally know of a case where someone uses this? The metric includes the whole url as a label, which is obviously a cardinality problem

I'm assuming people use recording rules to aggregate the metric into something more useable.

logicalhan:
FYI (We also have a way to enforce cardinality bounds): https://github.com/kubernetes/enhancements/pull/1692

logicalhan:
We try really, really hard not to break people upstream...

alvaroaleman:
> We try really, really hard not to break people upstream...

Sure but deactivating a metric and forcing ppl to do sth to get it back is always going to break them. I am also fine with adding an explicit opt-in for this and removing it in a future release, but the only value add I see in that is for cases where ppl use this metric and don't know how to properly the code that is currently being removed.

2uasimojo:
Okay; if either

> recording rules

or

> a way to enforce cardinality bounds

allows us to effectively nix this metric from taking up space, we're happy to do that. What does the former look like? /me starts reading the latter...

2uasimojo:
> adding an explicit opt-in for this

like this? https://github.com/kubernetes-sigs/controller-runtime/pull/1587

logicalhan:
> Sure but deactivating a metric and forcing ppl to do sth to get it back is always going to break them. 

Yes but in order to unbreak in this case is to roll-back a version. In our case, you just need to enable a flag. This means you can bump the version and use it in an unbroken way, but you must now explicitly acknowledge that this metric is not one you can depend on in a future release. It is a forcing function which is softer than, you can't use this release if you can't tolerate not having this metric.

logicalhan:
> > adding an explicit opt-in for this
> 
> like this? #1587

Yeah I probably prefer the optional approach, it's softer and friendly to users.

2uasimojo:
Okay, let's pursue this via #1587 (or something like it). Thanks!

Issue #1396, opened 2021-02-24
Import of controller-runtime calls flag for "kubeconfig"


chrischdi:
Hi there :-),

we are consuming `cluster-api-provider-openstack` which imports `kubernetes-sigs/controller-runtime`. 

Because [alias.go](https://github.com/kubernetes-sigs/controller-runtime/blob/master/alias.go#L23) is importing `sigs.k8s.io/controller-runtime/pkg/client/config` it always defines the ["kubeconfig" flag](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/config/config.go#L39) which then conflicts with other places where the same flag is getting defined. (e.g. inside kubernetes' e2e test framework https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/test_context.go#L321)

Of course we are able to workaround it by not directly using the functions via `alias.go` but I think this is against the reasons why it does exist?

One alternative would be to define a `RegisterFlag` function instead of defining the flag via `func init() {...}`, however this leads to a breaking change.

<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH, 
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Imprint</a>
</sup>

chrischdi:
Addition: the same thing happens when importing `sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha4`.

johannesfrey:
FYI: I created following WIP PR: https://github.com/kubernetes-sigs/controller-runtime/pull/1999

johannesfrey:
Should we close this issue now that #1999 has been merged? Or should we keep this issue open until the init function is removed completely (if this is desired at all)?

sbueringer:
> Should we close this issue now that #1999 has been merged? Or should we keep this issue open until the init function is removed completely (if this is desired at all)?

Sorry for the late reply. I'm not sure if we want to drop the init function. 

Pro:
* it would be better if no flags are just implicitly registered because someone imports the config package
Con:
* nobody seems to complain about it so it seems to work for most people

On the other hand, I think almost all controller are just run inside of a Pod and barely anyone is using a --kubeconfig command line flag.

IIRC component-base/logs also dropped the implicit registering of flags. So I'm slightly in favor of getting rid of the implicit flag registration (after whatever deprecation period / release notes is appropriate)

@alvaroaleman @vincepri WDYT?

PR #1587, opened 2021-07-09
:warning: Disable the `rest_client_request_latency_seconds` metric by default

### LINKED ISSUE #1423: Disable rest client latency metric

2uasimojo:
This metric has been deprecated since kube 1.14 and [turned off in kube 1.17](https://github.com/kubernetes/kubernetes/pull/83836). Stop registering it by default, but export the appropriate variables so
consumers can register it explicitly if they wish.

fixes #1423

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -52,7 +52,24 @@ const (
 
 var (
 	// client metrics.
-	requestLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+
+	// RequestLatency reports the request latency in seconds per verb/URL.
+	// Deprecated: This metric is deprecated for removal in a future release: using the URL as a
+	// dimension results in cardinality explosion for some consumers. It was deprecated upstream
+	// in k8s v1.14 and hidden in v1.17 via https://github.com/kubernetes/kubernetes/pull/83836.
+	// It is not registered by default. To register:
+	//	import (
+	//		clientmetrics "k8s.io/client-go/tools/metrics"
+	//		clmetrics "sigs.k8s.io/controller-runtime/metrics"
+	//	)
+	//
+	//	func init() {
+	//		clmetrics.Registry.MustRegister(clmetrics.RequestLatency)
+	//		clientmetrics.Register(clientmetrics.RegisterOpts{
+	//			RequestLatency: clmetrics.LatencyAdapter
+	//		})
+	//	}
+	RequestLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Subsystem: RestClientSubsystem,
 		Name:      LatencyKey,
 		Help:      "Request latency in seconds. Broken down by verb and URL.",
@@ -127,13 +144,11 @@ func init() {
 // registerClientMetrics sets up the client latency metrics from client-go.
 func registerClientMetrics() {
 	// register the metrics with our registry
-	Registry.MustRegister(requestLatency)
 	Registry.MustRegister(requestResult)
 
 	// register the metrics with client-go
 	clientmetrics.Register(clientmetrics.RegisterOpts{
-		RequestLatency: &latencyAdapter{metric: requestLatency},
-		RequestResult:  &resultAdapter{metric: requestResult},
+		RequestResult: &resultAdapter{metric: requestResult},
 	})
 }
 
@@ -159,11 +174,13 @@ func registerReflectorMetrics() {
 // copied (more-or-less directly) from k8s.io/kubernetes setup code
 // (which isn't anywhere in an easily-importable place).
 
-type latencyAdapter struct {
+// LatencyAdapter implements LatencyMetric.
+type LatencyAdapter struct {
 	metric *prometheus.HistogramVec
 }
 
-func (l *latencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
+// Observe increments the request latency metric for the given verb/URL.
+func (l *LatencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
 	l.metric.WithLabelValues(verb, u.String()).Observe(latency.Seconds())
 }
 


pkg/metrics/client_go_adapter.go
```diff
@@ -52,7 +52,21 @@ const (
 
 var (
 	// client metrics.
-	requestLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+
+	// RequestLatency reports the request latency in seconds per verb/URL.
```
alvaroaleman, line 56: Please document that this is deprecated, going to get removed and what the issue with it is.

logicalhan (commented):
what is this? this isn't really making it optional.. you should have a command line flag which allows you to enable this metric and plumb the value of that flag to the metric registration function

alvaroaleman:
> what is this? this isn't really making it optional.. you should have a command line flag which allows you to enable this metric and plumb the value of that flag to the metric registration function

No. We are a library. We are not going to register a flag. How and if controller authors expose options in their controllers to end-users is up to them.

logicalhan:
component-base is also a library, it includes flags which can be used by consuming binaries. these include the ones for metrics.

2uasimojo:
...but I'm with you on the plumbing part. If there was a way we could set a package variable such that it's present by the time the import `init()`s, we could switch it on/off from client code. Is that even possible in go?

logicalhan:
> ...but I'm with you on the plumbing part. If there was a way we could set a package variable such that it's present by the time the import `init()`s, we could switch it on/off from client code. Is that even possible in go?

yes it is possible, we literally do this in upstream kubernetes.

logicalhan:
obviously this means you can't use the init for these though.

2uasimojo:
Okay; I was looking for a way to enable opt-out so the change could be non-breaking. But if

> you can't use the init for these

then I'm not sure how that would work.

logicalhan:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/metrics/metrics.go#L506-L546

...is one way.

alvaroaleman:
Kubelet is a binary, we are a library. I am going to veto any change that registers a flag. The one case where we do this creates enough trouble already: https://github.com/kubernetes-sigs/controller-runtime/issues/1396

logicalhan:
> Okay; I was looking for a way to enable opt-out so the change could be non-breaking. 

I think it's okay to make it opt-in, it's more friendly than just deleting it. And the approach of manually invoking Register is now an option.

logicalhan:
> Kubelet is a binary, we are a library. I am going to veto any change that registers a flag. The one case where we do this creates enough trouble already: #1396

That's fine, I don't have any skin in the game here, I am just saying that it's not friendly to outright break people. What you guys want to do with your library and what your priorities are 🤷 to me.. I'm stating my opinion about deleting metrics and breaking people.. I've had it happen to me and it sucks.

logicalhan:
Actually, as a library, it would make sense to make all your metrics opt-in...

2uasimojo:
So if we don't

> registers a flag

and we want to

> make it opt-in

then is the approach in this PR acceptable? Or would it be better to wrap the registration of this metric in its own `Register` func?

Alternatively, I would be happy to propose

> make all your metrics opt-in

logicalhan:
Yeah I would probably make all the metrics public and auto-register them for a release and have a release note saying that in the next release, these will all be opt in. That way, people have some sort of warning, rather than just randomly have everything break. You have several options (as I see it):

- expose all metrics individually (this is kinda a pain because now you have to know all the ones you want to register)
- expose groups of metrics (this is slightly easier for consumers)
- expose a single Register which registers all metrics (possibly taking a var args of metrics not to enable).

2uasimojo:
I'm happy to do anything in this range, from [outright removal](https://github.com/kubernetes-sigs/controller-runtime/pull/1586) to immediate opt-in (this PR) to a multi-release no-change-deprecation => opt-in dance.

@alvaroaleman what is your wish?

alvaroaleman:
This change modulo my comment https://github.com/kubernetes-sigs/controller-runtime/pull/1587/files#r667037223 is IMHO fine. I don't think we should require opt-in for every single metric but I am happy to discuss if, which and how we register metrics by default in a new issue.

pkg/metrics/client_go_adapter.go
2uasimojo: Done.

2uasimojo:
> This change modulo my comment

✓

> I don't think we should require opt-in for every single metric but I am happy to discuss

See https://github.com/kubernetes-sigs/controller-runtime/pull/1589 for one approach: opt in gets you all by default, but you have the ability to opt *out* of individual metrics.

alvaroaleman:
I am going to retitle this to make sure that its properly understood when it ends up in a release note:
/retitle :warning: Disable the `rest_client_request_latency_seconds` metric by default

alvaroaleman (approved):
thanks!

2uasimojo:
Okay, so what's the process for getting this into a tagged release?

alvaroaleman:
@2uasimojo we deliver breaking changes with a new minor release and do a new minor release whenever there is a new kube minor release. So this will take some time.

PR #1588, opened 2021-07-09
✨  add health check for webhook server

### LINKED ISSUE #723: Add default readiness check for webhook server?

sbueringer:

Fixes #723 

Supersedes #1124

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	kscheme "k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics"
 )
@@ -86,6 +87,10 @@ type Server struct {
 
 	// defaultingOnce ensures that the default fields are only ever set once.
 	defaultingOnce sync.Once
+
+	// started is set to true immediately before the server is started
+	// and thus can be used to check if the server has been started
+	started bool
 
 	// mu protects access to the webhook map & setFields for Start, Register, etc
 	mu sync.Mutex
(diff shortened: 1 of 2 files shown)


sbueringer:
I implemented it this way as it's the easiest implementation and there didn't seem to be major concerns in #1124 (which has a similar implementation)

My biggest concern is that the `started` field is not 100% correct as the `Serve` func is executed after setting `started` to true. An alternative would be to check if the server has been started by trying a tcp tial on the webhook port (similar to https://github.com/kubernetes-sigs/cluster-api/blob/master/internal/envtest/environment.go#L225-L241).

pkg/webhook/server.go
```diff
@@ -280,6 +288,19 @@ func (s *Server) Start(ctx context.Context) error {
 	return nil
 }
 
+// StartedChecker returns an healthz.Checker which is healthy after the
+// server has been started.
+func (s *Server) StartedChecker() healthz.Checker {
```
sbueringer, line 293: I can also move this func into the healthz package. The consequence would be that either the healthz package has a dependency onto the webhook package or I would have to add an interface with a `Started() bool` func.

vincepri: Could we maybe also tcp ping the server?

sbueringer: Yup definitely. Should we then still keep the started flag? 

Or is the idea to combine that, so we only start pinging after started is true?

I'm not sure if it's worth it keeping the started field around.

vincepri: It can make sense to only start tcp-pinging the port only after the started field is set to true

sbueringer: Yup, will do.

sbueringer:
I'll squash before merge.

pkg/webhook/server.go
```diff
@@ -280,6 +289,27 @@ func (s *Server) Start(ctx context.Context) error {
...
+	return func(req *http.Request) error {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+
+		if !s.started {
+			return fmt.Errorf("webhook server has not been started yet")
+		}
+
+		conn, err := net.DialTimeout("tcp", net.JoinHostPort(s.Host, strconv.Itoa(s.Port)), 10*time.Second)
```
sbueringer, line 303: 10 seconds is just a wild guess. Not sure after which threshold we want to declare the webhook server not ready.

cndoit18 (approved):

vincepri (approved):
/approve
/lgtm

PR #1589, opened 2021-07-09
:warning: Make all metrics opt-in

2uasimojo:
This commit stops registering metrics by default. Consumers may now register them explicitly via two newly-exported methods:

```go
import (
	crmetrics "sigs.k8s.io/controller-runtime/metrics"
)

func init() {
	crmetrics.RegisterClientMetrics()
	crmetrics.RegisterReflectorMetrics()
}
```

Each of these methods accepts varargs to opt *out* of any of the metrics it is responsible for. For example:

```go
crmetrics.RegisterClientMetrics(crmetrics.LatencyKey)
```

This will still register `rest_client_requests_total`, but will avoid
registering `rest_client_request_latency_seconds`.

fixes: #1423

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -119,34 +119,53 @@ var (
 	}, []string{"name"})
 )
 
-func init() {
-	registerClientMetrics()
-	registerReflectorMetrics()
-}
+// RegisterClientMetrics sets up the client latency metrics from client-go.
+// To opt *out* of specific metrics, specify their *Key consts to optOut.
+func RegisterClientMetrics(optOut ...string) {
+	toRegister := map[string]bool{
+		LatencyKey: true,
+		ResultKey:  true,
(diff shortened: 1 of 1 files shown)


2uasimojo:
/close

We're going to do #1587 instead, but I wanted to get this out of my brain.

pkg/metrics/client_go_adapter.go
```diff
@@ -119,34 +119,53 @@ var (
 	}, []string{"name"})
 )
 
-func init() {
-	registerClientMetrics()
-	registerReflectorMetrics()
-}
+// RegisterClientMetrics sets up the client latency metrics from client-go.
+// To opt *out* of specific metrics, specify their *Key consts to optOut.
```
2uasimojo, line 123: Realistically this is lousy UX without a better indication that the varargs are opt *out* rather than *in*. It would probably be better to flip it.

PR #1591, opened 2021-07-10
[WIP] :sparkles: Add GVKCaches and DynamicMultiNamespaceCache

### LINKED ISSUE #1590: Introduce a dynamic multi-namespace cache
The DynamicMultinamespacedCache will dynamically create a cache for a given namespace on demand. Like the MultinamespacedCache, it supports cluster and namespaced resources. It also able to handle permissions granted after starting the cache, being able to create more caches on demand hence the dynamic nature. The DynamicMultinamespacedCache uses the MultinamespacedCache underneath.

Update:
We have operators that are moving from namespace scoped to cluster scoped. An operator uses caches for dependent resources, these caches are usually spun up with list/watchers for those resources using controller-runtime pattern at startup. 

If the operator is started with access only to the owned CRDs to start and not given all of the required permissions at startup. The caches will not have the dependent resources and when the admin grants permissions later the caches will not be able to react to the change.

The operator would need access to a cache that can handle dynamically changing permissions.

jmrodri:
Fixes #1590 

We have operators that are moving from namespace scoped to cluster scoped. An operator uses caches for dependent resources, these caches are usually spun up with list/watchers for those resources using controller-runtime pattern at startup.

If the operator is started with access only to the owned CRDs to start and not given all of the required permissions at startup. The caches will not have the dependent resources and when the admin grants permissions later the caches will not be able to react to the change.

The operator would need access to a cache that can handle dynamically changing permissions.

### DIFF
--- a/pkg/cache/gvk_caches.go
+++ b/pkg/cache/gvk_caches.go
@@ -1,0 +1,173 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+type GVKCachesOptions struct {
+	ObjectCacheFuncs map[client.Object]NewCacheFunc
+	DefaultCacheFunc NewCacheFunc
+}
+
+// GVKCachesBuilder - Builder function to create a new composite cache which contains
+// one or more caches based on a mapping of GVKs. Each GVK can be configured to
+// use a separate cache.
+func GVKCachesBuilder(gvkCachesOpts GVKCachesOptions) NewCacheFunc {
+	return func(config *rest.Config, opts Options) (Cache, error) {
+		opts, err := defaultOpts(config, opts)
+		if err != nil {
+			return nil, err
+		}
+
+		if gvkCachesOpts.DefaultCacheFunc == nil {
+			gvkCachesOpts.DefaultCacheFunc = New
+		}
+		defaultCache, err := gvkCachesOpts.DefaultCacheFunc(config, opts)
+		if err != nil {
+			return nil, err
+		}
+
+		gvkToCache := map[schema.GroupVersionKind]Cache{}
+		for obj, newCacheFunc := range gvkCachesOpts.ObjectCacheFuncs {
+			gvk, err := apiutil.GVKForObject(obj, opts.Scheme)
+			if err != nil {
+				return nil, err
+			}
+			gvkToCache[gvk], err = newCacheFunc(config, opts)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		return &gvkCaches{
+			gvkToCache:   gvkToCache,
+			defaultCache: defaultCache,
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@jmrodri could you please extend your PR body to explain what this does and what use case that fulfills?

pkg/cache/gvk_caches.go
```diff
@@ -0,0 +1,173 @@
...
+	if cache, ok := c.gvkToCache[gvk]; ok {
+		return cache.GetInformer(ctx, obj)
+	}
+
+	return c.defaultCache.GetInformer(ctx, obj)
+}
+
+func (c *gvkCaches) GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind) (Informer, error) {
+	if cache, ok := c.gvkToCache[gvk]; ok {
+		return cache.GetInformerForKind(ctx, gvk)
```
varshaprasad96, line 127: what in general would be the use case of having caches based on gvk? In the sense, what would be the scenario where it is more optimized than having caches for each namespace?

pkg/cache/gvk_caches.go
```diff
@@ -0,0 +1,173 @@
...
+	gvk, err := apiutil.GVKForObject(obj, c.scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	if cache, ok := c.gvkToCache[gvk]; ok {
+		return cache.GetInformer(ctx, obj)
+	}
+
+	return c.defaultCache.GetInformer(ctx, obj)
```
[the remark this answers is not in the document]
joelanford: This is _probably_ fine, but it's worth pointing out that the underlying cache's `GetInformerForKind` implementation might not actually be equivalent to looking up the GVK and calling `GetInformer`.

With the existing implementation, we're always calling the same underlying cache function that was originally called.

pkg/cache/gvk_caches.go
```diff
@@ -0,0 +1,173 @@
...
+			gvkToCache:   gvkToCache,
+			defaultCache: defaultCache,
+			scheme:       opts.Scheme,
+		}, nil
+	}
+}
+
+type gvkCaches struct {
+	gvkToCache   map[schema.GroupVersionKind]Cache
+	defaultCache Cache
```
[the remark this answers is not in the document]
joelanford: It would contain any objects whose GVK is not explicitly in the GVK cache mapping. So the idea is that if you have a use case where a specific GVK needs a custom cache, then you can use this implementation to specify that cache for the GVK you care about and another (default) cache for all other GVKs your clients use.

pkg/cache/gvk_caches.go
joelanford: The primary use case I can think of is this:
1. Primary object - global cache
2. Secondary resources - global cache or possibly dynamic multinamespace cache, depending on expected number of namespaces that would contain primary objects and RBAC permissions.
3. Controller configuration objects (possibly configmaps, secrets, or CRD config objects) - controller namespace

This solves two problems:
1. Memory savings: 
    - if using the dynamic multinamespace cache, an operator would only cache secondary objects in the namespaces where a CR is or has been present since the controller started
    - if a controller expects some if its objects to exist _only_ in the namespace it is running, it can watch that GVK in its namespace only, again saving cache space by not caching unrelated objects of the same GVK in other namespaces.
    - if a controller can trigger a reconciliation for its primary object based solely on the metadata of another object, it can setup a metadata-only cache, which avoids storing unnecessary spec/status contents.
2. Controller RBAC: In basically all cases, a controller should be given global get/list/watch its primary objects and update on its primary objects status, so that it is always able to inform users of the state of custom objects across the cluster. However, in many cases, a cluster admin may want to limit permissions of a controller to get/list/watch/create/update/delete secondary objects in specific namespaces. This setup fails in a single global cache scenario because the controller would not have global get/list/watch/create/update/delete permission on secondary resources.

alvaroaleman: Could we maybe clarify the use case in an issue before we dump code? The referenced issue only describes how this works, not what problem it solves. That makes it impossible to provide any useful design feedback.
/hold

jmrodri: @alvaroaleman I will update the issue.

jmrodri: @alvaroaleman I've updated the issue. 
> We have operators that are moving from namespace scoped to cluster scoped. An operator uses caches for dependent resources, these caches are usually spun up with list/watchers for those resources using controller-runtime pattern at startup.
> 
> If the operator is started with access only to the owned CRDs to start and not given all of the required permissions at startup. The caches will not have the dependent resources and when the admin grants permissions later the caches will not be able to react to the change.
> 
> The operator would need access to a cache that can handle dynamically changing permissions.

PR #1593, opened 2021-07-15
⚠️  Bump github.com/go-logr/zapr & github.com/go-logr/logr to v1.1.0

morlay:
Bump github.com/go-logr/zapr & github.com/go-logr/logr to v1.0.0
with logr apis refactor


vincepri:
/retitle ⚠️  Bump github.com/go-logr/zapr & github.com/go-logr/logr to v1.0.0

joelanford:
There _might_ be some impact on this change with respect to the [zap-log-level and zap-stacktrace-level flags ](https://github.com/kubernetes-sigs/controller-runtime/blob/40cffdcd5fe1b842fdfb3dd3514763f8a471e383/pkg/log/zap/zap.go#L244-L246) because [negative V values are now treated the same as `0`](https://github.com/go-logr/logr/blob/622c97b4eb90a99cc628f4773662eb0015ce0f2a/logr.go#L226-L231).

The zap `error` and `panic` levels correspond to negative logr levels, so we need to either verify that those flags continue working as they always have, or if they don't, we need to tweak `pkg/log/zap` to compensate for this change.

For example, this test may fail with the change to logr/zapr v1.0.0: https://github.com/kubernetes-sigs/controller-runtime/blob/40cffdcd5fe1b842fdfb3dd3514763f8a471e383/pkg/log/zap/zap_test.go#L538-L539

morlay:
Block by `k8s.io/klog/v2`, waiting for a new release 
branch main already upgraded https://github.com/kubernetes/klog

joelanford:
Moving this to a future milestone. The [kubernetes/kubernetes project is still on logr 0.4.0](https://github.com/kubernetes/kubernetes/blob/v1.22.1/go.mod#L257) for the 1.22 release, so controller-runtime will stay on 0.4.0 for v0.10.x to stay aligned with upstream.

/milestone Next

morlay:
klog released v2.20.0 
https://github.com/kubernetes/klog/releases/tag/v2.20.0

Could continue do the logr upgrade.


it is safe to using with `k8s.io/client-go@v0.22.x`
However the panic log with break because of `Negative V-levels are treated as 0` .
https://github.com/go-logr/logr/blob/master/logr.go#L269-L279

randomvariable:
We should probably merge this in, but is there going to be some downstream impact for consumers, in which case this should go to v0.11?

varshaprasad96:
If we are going to cut a release with this breaking change + k8s 1.22 deps, can we merge this before #1709 ?

joelanford (commented):
In the GoDoc for `logr.Logger`, I see:
> Implementations of LogSink should provide their own constructors that return Logger, not LogSink.

It seems like we're violating this guidance with the `DelegatingLogSink` constructor.

I'm guessing the reason is that we somehow need to call `Fulfill()` eventually, but we can't if our concrete `DelegatingLogger` is lost behind the `logr.Logger` abstraction.

Is there a way to overcome this? Perhaps we can fudge the guidance and have a `NewDelegatingLogger` that returns a `DelegatingLogger` rather than a `logr.Logger`? And then callers could call `Fulfill()` but then also unwrap it to get at an underlying `logr.Logger`?

Maybe related, do we need `DelegatingLogSink` to be exported? I wonder if we can restore `DelegatingLogger` and unexport `DelegatingLogSink`.

I'm totally spitballing here. Not sure what cascading effects this would cause.

morlay:
@joelanford 

Have to return  `logr.Logger`, it is an struct. if we return custom struct, will be not compatible with other pkgs.

So I have upgrade codes, and make `DelegatingLogSink` as  private.

and added new `NewDelegatingLogger(logSink logr.LogSink) logr.Logger`

For `FullFill` could be use 
```golang
Log.GetSink().(CanFulfill).Fulfill(l.GetSink())
```
or added a new function to safe fill
```golang
func Fulfill(l * logr.Logger, logSink logr.LogSink)
```

morlay:
Sames previus version already meged. Close this

PR #1595, opened 2021-07-16
✨ Adds DeleteIfExist Variants for ControllerUtil

sujil02:
Adds DeleteIfExist Variants for ControllerUtil

Signed-off-by: sujil02 <sujil.shah@gmail.com>

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -179,6 +179,8 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 	OperationResultCreated OperationResult = "created"
 	// OperationResultUpdated means that an existing resource is updated.
 	OperationResultUpdated OperationResult = "updated"
+	// OperationResultDeleted means that an existing resource is deleted.
+	OperationResultDeleted OperationResult = "deleted"
 	// OperationResultUpdatedStatus means that an existing resource and its status is updated.
 	OperationResultUpdatedStatus OperationResult = "updatedStatus"
 	// OperationResultUpdatedStatusOnly means that only an existing status is updated.
@@ -332,6 +334,20 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 	}
 
 	return result, nil
+}
+
+// DeleteIfExist deletes the given object in the Kubernetes
+// cluster.
+//
+// It returns the executed operation and an error.
+func DeleteIfExist(ctx context.Context, c client.Client, obj client.Object) (OperationResult, error) {
+	if err := c.Delete(ctx, obj); err != nil {
+		if apierrors.IsNotFound(err) {
+			return OperationResultNone, nil
+		}
+		return OperationResultNone, err
+	}
+	return OperationResultDeleted, nil
 }
 
 // mutate wraps a MutateFn and applies validation to its result.
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -407,6 +407,89 @@ var _ = Describe("Controllerutil", func() {
 		})
 	})
 
+	Describe("DeleteIfExist", func() {
+		var deploy *appsv1.Deployment
+		var deplSpec appsv1.DeploymentSpec
+		var deplKey types.NamespacedName
+		var specr controllerutil.MutateFn
+
+		BeforeEach(func() {
+			deploy = &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("deploy-%d", rand.Int31()), //nolint:gosec
+					Namespace: "default",
+				},
+			}
+
+			deplSpec = appsv1.DeploymentSpec{
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"foo": "bar"},
+				},
+				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"foo": "bar",
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  "busybox",
+								Image: "busybox",
+							},
+						},
+					},
+				},
+			}
+
+			deplKey = types.NamespacedName{
+				Name:      deploy.Name,
+				Namespace: deploy.Namespace,
+			}
+
+			specr = deploymentSpecr(deploy, deplSpec)
+		})
+
+		It("creates a new object and delete it", func() {
+			op, err := controllerutil.CreateOrUpdate(context.TODO(), c, deploy, specr)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning OperationResultCreated")
+			Expect(op).To(BeEquivalentTo(controllerutil.OperationResultCreated))
+
+			By("actually having the deployment created")
+			fetched := &appsv1.Deployment{}
+			Expect(c.Get(context.TODO(), deplKey, fetched)).To(Succeed())
+
+			By("being mutated by MutateFn")
+			Expect(fetched.Spec.Template.Spec.Containers).To(HaveLen(1))
+			Expect(fetched.Spec.Template.Spec.Containers[0].Name).To(Equal(deplSpec.Template.Spec.Containers[0].Name))
+			Expect(fetched.Spec.Template.Spec.Containers[0].Image).To(Equal(deplSpec.Template.Spec.Containers[0].Image))
+
+			ope, err := controllerutil.DeleteIfExist(context.TODO(), c, deploy)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning OperationResultDeleted")
+			Expect(ope).To(BeEquivalentTo(controllerutil.OperationResultDeleted))
+
+		})
+
+		It("Delete a nonexisting object should return status unchanged", func() {
+
+			ope, err := controllerutil.DeleteIfExist(context.TODO(), c, deploy)
+
+			By("returning no error")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("returning OperationResultNone")
+			Expect(ope).To(BeEquivalentTo(controllerutil.OperationResultNone))
+
+		})
+	})
 	Describe("CreateOrPatch", func() {
 		var deploy *appsv1.Deployment
 		var deplSpec appsv1.DeploymentSpec


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -334,6 +336,47 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
...
+//
+// It returns the executed operation and an error.
+func DeleteIfExist(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
+	key := client.ObjectKeyFromObject(obj)
+	if err := c.Get(ctx, key, obj); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return OperationResultNone, err
+		}
+		if err := c.Delete(ctx, obj); err != nil {
+			return OperationResultNone, err
```
coderanger, line 372: Generally in code like this you want to allow a 404 response to still be considered a success, helps with concurrency conditions.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -334,6 +336,47 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 	return result, nil
 }
 
+// CreateIfNotExist creates the given object in the Kubernetes
+// cluster. The object's desired state must be reconciled with the existing
+// state inside the passed in callback MutateFn.
+//
+// It returns the executed operation and an error.
+func CreateIfNotExist(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
```
coderanger, line 344: Not sure this is common enough to warrant a helper. Controllers really shouldn't be doing this except in very rare cases as it's a sign of non-convergent behavior.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -334,6 +336,47 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
...
+		return OperationResultCreated, nil
+	}
+	return OperationResultNone, nil
+}
+
+// DeleteIfExist deletes the given object in the Kubernetes
+// cluster.
+//
+// It returns the executed operation and an error.
+func DeleteIfExist(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
```
coderanger, line 365: MutateFn isn't used and isn't needed.

sujil02: Fixed.

pkg/controller/controllerutil/controllerutil.go
sujil02: @coderanger So if I got it right, irrespective of the error has occurred or not the response should be OperationResultDeleted? right?

pkg/controller/controllerutil/controllerutil.go
jmrodri: @coderanger we have a teams that use a variety of *IfNotExist methods. They created them because controller-runtime didn't have them. We distilled the methods do this specific `CreateIfNotExist` variant. We were hoping to let the teams know that controller-runtime now supports their use case and stop maintaining their variants. That's what brought this PR on.

coderanger: Could you explain the use cases that lead to needing them? I don't doubt that you have code that does it, what I doubt is if that code is a good idea and should be encouraged :)

pkg/controller/controllerutil/controllerutil.go
coderanger: I would probably return `OperationResultNone, nil` if the delete returns a 404? This code didn't do a delete but someone else did.

sujil02: Well if the object did not exist(404) it will never enter the if statement and default returns `OperationResultNone, nil` but if it does exist and delete could not process the action hence returned the error.

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -407,6 +407,92 @@ var _ = Describe("Controllerutil", func() {
...
+
+			deplKey = types.NamespacedName{
+				Name:      deploy.Name,
+				Namespace: deploy.Namespace,
+			}
+
+			specr = deploymentSpecr(deploy, deplSpec)
+		})
+
+		It("creates a new object if one doesn't exists", func() {
```
varshaprasad96, line 453: could be helpful to have the case tested where no error is returned when object exists.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -195,16 +197,7 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
...
-			return OperationResultNone, err
-		}
-		if err := mutate(f, key, obj); err != nil {
-			return OperationResultNone, err
-		}
-		if err := c.Create(ctx, obj); err != nil {
-			return OperationResultNone, err
-		}
-		return OperationResultCreated, nil
+		return CreateIfNotExist(ctx, c, obj, f)
```
varshaprasad96, line 200: `CreateIfNotExist` and `CreateOrUpdate` are similar. I am wondering if we need a separate helper for `CreateIfNotExist`. One can use both the helpers to do the same, right?

pkg/controller/controllerutil/controllerutil.go
coderanger: That would not have returned a 404 though, a 404 specifically means "object not found".

pkg/controller/controllerutil/controllerutil.go
joelanford: I agree. Is there a use case for `CreateIfNotExist` that is not already covered by `CreateOrUpdate`?

sujil02:
Reviewing the discussion removed the createsIfExist helper.

pkg/controller/controllerutil/controllerutil.go
joelanford: It's possible for `c.Get` to hand you an object and `c.Delete` to return a 404. The reason is that `c.Get` returns an object from the cache (which might be outdated), and `c.Delete` issues the request directly to the apiserver. And even if `c.Get` read from the apiserver directly, it would be possible for the apiserver to receive and handle a deletion request from another client between the `c.Get` and `c.Delete` calls.

I think you can simplify to this instead:
```go
if err := c.Delete(ctx, obj); client.IgnoreNotFound(err) != nil {
}
return OperationResultDeleted, nil
```

But given that calling code would take up just as many lines as this wrapper, I wonder if there's any actual value to the `DeleteIfExists` wrapper function?

coderanger: Given the eventual destiny of CreateOrUpdate is be removed in favor of Server Side Apply (some day, don't anyone go panic this will be soon) I could see an argument we shouldn't really add more of these helpers in general. But maybe having a "common patterns" doc to help folks coming up to speed with using the API?

coderanger: (OTOH treating this file as "executable documentation" wouldn't be the worst)

jmrodri: @joelanford I tend to agree. But there are few libraries that keep adding these types of methods.

coderanger: @jmrodri A lot of folks struggle to understand writing convergent code and try to shoehorn procedural patterns into operators where they don't belong. Very understandable, but that doesn't mean we should encourage those patterns. We should instead improve our resources to teach people convergent patterns :)

sujil02: If we update the deleteIfExits to
```
if err := c.Delete(ctx, obj); client.IgnoreNotFound(err) != nil {
}
return OperationResultDeleted, nil
```
It will return `OperationResultDeleted` even if the object doesn't exist. If that's okay? I will go ahead and update the helper.

joelanford: I'm trying to think of a scenario where I would call this and need to handle `OperationResultDeleted, nil` differently from `OperationResultNone, nil`.

In either case, the thing I want deleted is not present after this returns a non-nil error.

coderanger: @joelanford The usual case is emitting Events for debugging or analysis purposes.

joelanford: @coderanger Good point. Slight tweak then:

```go
if err := c.Delete(ctx, obj); err != nil {
	if apierrors.IsNotFound(err) {
		return OperationResultNone, nil
	}
}
return OperationResultDeleted, nil
```

sujil02: @joelanford @coderanger thanks for the feedback. Updated the helper.

varshaprasad96: As @joelanford previously mentioned, I'm thinking of the benefit this would add to the generic `client.Delete`. We are additionally wrapping `IsNotFound` error and returning `OperationResultNone`. Regarding handling of `OperationResultDeleted` and `OperationResultNone`, if we want to find if the object exists in the cluster or not, shouldn't the right method be to send the request again and use `IsNotFound` error directly. It seems counterintuitive that if someone wants to delete an object either way, they would be handing the case where object wasn't present in the cluster differently. Or is there any scenrio I am missing?

sujil02: @varshaprasad96 @coderanger @varshaprasad96 @jmrodri 
Any further discussion on this do we need any amends?

varshaprasad96: @joelanford @coderanger @jmrodri - would like to know your opinion. I am still a bit doubtful on whether having this helper would be useful. If not, we could have this documented in FAQ instead of creating a separate wrapper.

pkg/controller/controllerutil/controllerutil.go
jmrodri: @joelanford the one comment I got from the developers that wrote the Create*IfNotExist utilties was the CreateOrUpdate was hard to use with the mutations. So maybe there is a place to better document that call with better examples.

pkg/controller/controllerutil/controllerutil.go
jmrodri: So what I'm reading is that call to `Delete` would look like this:
```go
if err := c.Delete(ctx, obj); err != nil {
	if apierrors.IsNotFound(err) {
		// handle not found
	}
	// handle error
}
// otherwise deleted
```

and a call `DeleteIfExists` would look like:

```go
if _, err := controllerutil.DeleteIfExist(ctx, c, obj); err != nil {
    // handle error
}
// otherwise deleted
```

I still see value in not having to handle the `IsNotFound` case each time. But I can see that it's not that much more code. I still see the value of `DeleteIfExists`, that's just my opinion.

coderanger:
I think this is mergable?

/lgtm

varshaprasad96 (approved):
/lgtm

PR #1608, opened 2021-07-29
🐛feat(testing): add env ETCD_UNSUPPORTED_ARCH for etcd in ARM

### LINKED ISSUE #1609: Etcd needs extra env in ARM
ETCD_UNSUPPORTED_ARCH=arm is necessary for etcd in ARM

wu-cl:
fixes #1609

ETCD_UNSUPPORTED_ARCH=arm is necessary for etcd in ARM

This PR will check the current GOARCH, if it is arm64, ETCD_UNSUPPORTED_ARCH=arm will  be set for launching etcd.

### DIFF
--- a/pkg/internal/testing/controlplane/etcd.go
+++ b/pkg/internal/testing/controlplane/etcd.go
@@ -20,6 +20,8 @@ import (
 	"io"
 	"net"
 	"net/url"
+	"os"
+	"runtime"
 	"strconv"
 	"time"
 
@@ -91,6 +93,13 @@ type Etcd struct {
 func (e *Etcd) Start() error {
 	if err := e.setProcessState(); err != nil {
 		return err
+	}
+
+	//Set extra env for etcd in arm
+	if runtime.GOARCH == "arm64" {
+		if err := os.Setenv("ETCD_UNSUPPORTED_ARCH", "arm64"); err != nil {
+			return err
+		}
 	}
 	return e.processState.Start(e.Out, e.Err)
 }


alvaroaleman:
/hold
I presume there is a reason etcd folks require this env var and we shouldn't just unconditionally set it like this. Ppl can simply set and export this variable prior to running their tests and things will work.

vincepri:
+1 to what @alvaroaleman said, https://github.com/etcd-io/etcd/blob/9a53601a1867c2ec2da244b0f9a5f14449f5e3f6/Documentation/op-guide/supported-platform.md are you running on a 32bit system?

wu-cl:
It seems that etcd v3.5.0(latest) is not supported in Ubuntu20.04 ARM64.
```
root@arm:~/etcd-v3.5.0-linux-arm64 # cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.2 LTS"
root@arm:~/etcd-v3.5.0-linux-arm64 # arch
aarch64
root@arm:~/etcd-v3.5.0-linux-arm64 # etcd
etcd on unsupported platform without ETCD_UNSUPPORTED_ARCH=arm64 set
root@arm:~/etcd-v3.5.0-linux-arm64 #
```

coderanger:
@wu-cl We understand that it's not supported on all platforms, the issue is we don't want to "fix" that in a way that's not a real fix. If you are personally willing to accept the caveats of an unsupported platform you can `export ETCD_UNSUPPORTED_ARCH=arm` before running your tests (or in the Makefile, envrc, whatever).

vincepri:
@wu-cl Should we file a bug in etcd directly to ask for further clarification?

From their code, arm64 should be supported https://github.com/etcd-io/etcd/blob/v3.5.0/server/etcdmain/etcd.go#L467-L472

Although in their testing they do test the flag https://github.com/etcd-io/etcd/blob/932d42b0279c21b6e73134729ee7f2831106bc24/tests/e2e/main_test.go#L37 in a forced way and they do the same in the Dockerfile for arm64 https://github.com/etcd-io/etcd/blob/932d42b0279c21b6e73134729ee7f2831106bc24/Dockerfile-release.arm64#L8

wu-cl:
Yes, I agree. It should be blame on etcd.

PR #1612, opened 2021-07-29
:bug: Envtest should try to allocate a port for etcd listenPeerURL

vincepri:

### DIFF
--- a/pkg/internal/testing/controlplane/etcd.go
+++ b/pkg/internal/testing/controlplane/etcd.go
@@ -84,6 +84,10 @@ type Etcd struct {
 	// args contains the structured arguments to use for running etcd.
 	// Lazily initialized by .Configure(), Defaulted eventually with .defaultArgs()
 	args *process.Arguments
+
+	// listenPeerURL is the address the Etcd should listen on for peer connections.
+	// It's automatically generated and a random port is picked during execution.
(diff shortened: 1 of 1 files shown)


pkg/internal/testing/controlplane/etcd.go
```diff
@@ -150,7 +167,7 @@ func (e *Etcd) Stop() error {
 
 func (e *Etcd) defaultArgs() map[string][]string {
 	args := map[string][]string{
-		"listen-peer-urls": {"http://localhost:0"},
+		"listen-peer-urls": {e.listenPeerURL.String()},
```
alvaroaleman, line 170: What difference does this make? Shouldn't etcd get a free port from the kernel (and then our allocation not consider that one)?

vincepri: If there are multiple parallel packages running at the same time, process A could internally allocate a port (on file) which is then used for either etcd listen or api-server, while process B (etcd) is already listening on it. 

There is always a race condition between the code that allocates the port and the actual program startup, because the port gets closed right after we have a file lock, another program not knowing about the file allocation can snatch it.

alvaroaleman: Makes sense, thanks for the explanation.

/lgtm
/approve

PR #1598, opened 2021-07-20
✨   make it possible to monitor if the webhook server has been started

sbueringer:
Signed-off-by: Stefan Büringer buringerst@vmware.com

Cherry-pick of #1588

PR #1616, opened 2021-07-30
 🐛   0.9: fix webhook health check tls handshake timeout errors

sbueringer:

My previous PR #1598 introduced a health check for the webhooks. The health check works as intended (it reports healthy when the webhook server was up), but it also produces tls handshake timeout errors and thus a lot of log messages like:
```bash
TLS handshake error from 127.0.0.1:42498: EOF
```

xref: https://kubernetes.slack.com/archives/CD6U2V71N/p1627643594012600


sbueringer:
/retest
probably flaky

vincepri (approved):
/approve
/lgtm

PR #1618, opened 2021-07-30
:bug: Fix a bug with `Limit` list opt returning less than expected number of objects in the result.

ankel:
Previously, `Limit(n)` only looks at the first `n` objects. This could cause `List` to return less than `n` objects if other opts apply further filtering, even though there are more than `n` total number of objects matching the list options.

With the fix, `List` will loop until it collects `n` objects before returning.


alvaroaleman (approved):
Thanks!

jpfourny:
Can this fix be backported to 0.9? Some folks are stuck at K8s 1.21 deps (due to compatibility with K8s api-server).

PR #1152, opened 2020-09-07
:running: Allow specifying custom TLS Config for webhooks

varshaprasad96:
Allow specifying custom TLS Config for webhooks

* This implements a Manager option WebhookTLSConfig which allows for
  specifying a custom TLS configuration when running webhooks using
  controller-runtime
* Added CustomTLSConfig attribute to WebhookInstallOptions
* Added test to show this config overrides the config generated by
  CertDir
* If cert-dir and tls-config are provided, custom tls-config is used

Authored by: Matt Smith (@alastor-erinyes)
Closes: #1135 #852 
Taken from: [PR 853](https://github.com/kubernetes-sigs/controller-runtime/pull/853#) 

c/c @alastor-erinyes

Issue #1135, opened 2020-08-24
Ability to specify the name of server key and certificate while setting up Webhooks


varshaprasad96:
**Context:**
Currently, when we create a manager, we have the ability to specify the [`CertDir`](https://github.com/kubernetes-sigs/controller-runtime/blob/011cd8ae62dba632cf63f0637971cbca813cbf29/pkg/manager/manager.go#L209) which contains the server key and certificate. However, this comes with a clause that the server key and certificate files should be named as tls.key and tls.crt, respectively.

Eg:
```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
		Scheme:             scheme,
		MetricsBindAddress: metricsAddr,
		Port:               9443,
		LeaderElection:     enableLeaderElection,
		LeaderElectionID:   "86f835c3.example.com",
		CertDir:            "/apiserver.local.config/certificates",
	})
```

In this example, `/apiserver.local.config/certificates` directory is expected to have `tls.key` and `tls.cert`

**Problem:**

Operator SDK intends to support OLM for creating and managing of Webhooks. Currently, OLM  does not use cert manager, and creates self signed certificates for all the deployments mentioned in the CRD. The OLM uses the [API Service logic](https://github.com/operator-framework/olm-docs/blob/master/content/en/docs/advanced-tasks/adding-an-admission-webhook.md#certificate-authority-constraints), and hence the server key and certificates are named as `apiserver.key` and `apiserver.cert`.

**Issue**:

When we use OLM to manage operators built using SDK, we need not create a webhook server manually. However, while using controller-runtime's `SetupWebhook`, the manager looks for the certificates mentioned above during the initial setup. Though there is an option provided for specifying the directory path of the certificates, there is no flexibility in mentioning the name of the files. 

Having the flexibility of specifying the name of the server key and certificate would be helpful.

PR #1620, opened 2021-08-02
✨ Allow TLSMinVersion to be set in Manager configuration

willthames:
Routines like GetWebhookServer need to have configuration exposed
to be able to set TLS minimum version

### DIFF
--- a/pkg/config/v1alpha1/types.go
+++ b/pkg/config/v1alpha1/types.go
@@ -139,6 +139,10 @@ type ControllerWebhook struct {
 	// must be named tls.key and tls.crt, respectively.
 	// +optional
 	CertDir string `json:"certDir,omitempty"`
+
+	// TLSMinVersion is the minimum version of TLS for the webhook server
+	// Defaults to TLS 1.0 if not set.
+	TLSMinVersion string `json:"tlsMinVersion,omitempty"`
 }
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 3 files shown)


willthames:
Follow up to #1548 which allowed TLS minimum version to be set when creating a webhook.Server, but didn't expose that to e.g. GetWebhookServer

joelanford:
/hold

I'm curious if we could solve the general case of managing TLS configuration by simply allowing a custom `tls.Config` to be used. It looks like several PRs have already attempted to do this (#853 and  #1135, both closed by the stale bot), and I didn't see any major reason for those _not_ to be merged.

I worry that adding `TLSMinVersion` will open the floodgates for feature requests to add other aspects of the TLS configuration, so IMO we ought to just expose the whole thing now.

willthames:
@joelanford I'd be happy with that approach, if it can be done without going stale again (and I realise that's not always in the control of the project)

Would you like me to take any concrete next steps to help with resurrecting #853?

joelanford:
@willthames If possible, I'd suggest checking out that PR, re-basing onto the current master, and then making further commits to address the comments in #853 and #1152.

I don't see any major hurdles for getting it in; it just needs some attention (which you have this time, at least from me).

willthames:
@joelanford no problem, I'll see what I can do

willthames:
@joelanford #1623 added to update #853

PR #1621, opened 2021-08-02
🐛 fix List method for metadataClient

Iceber:
Signed-off-by: Iceber Gu <wei.cai-nat@daocloud.io>

I think the `List` method should not modify the `GVK` of the `obj`

### DIFF
--- a/pkg/client/metadata_client.go
+++ b/pkg/client/metadata_client.go
@@ -145,7 +145,8 @@ func (mc *metadataClient) List(ctx context.Context, obj ObjectList, opts ...List
 		return fmt.Errorf("metadata client did not understand object: %T", obj)
 	}
 
-	gvk := metadata.GroupVersionKind()
+	listGVK := metadata.GroupVersionKind()
+	gvk := listGVK
 	if strings.HasSuffix(gvk.Kind, "List") {
 		gvk.Kind = gvk.Kind[:len(gvk.Kind)-4]
 	}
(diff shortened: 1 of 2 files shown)


alvaroaleman (commented):
@Iceber Could you please add a testcase for this?

/ok-to-test

vincepri:
+1 for a test case

Iceber:
@alvaroaleman @vincepri I observed that it is common in `controller-runtime` to write BDD tests for exportable methods or structures, is it necessary to add a separate test for `metadataClient` as a non-exportable structure?

The bug in the List method of `metadataClient` is circumvented by reset in `(*client) List`.
https://github.com/kubernetes-sigs/controller-runtime/blob/9fe547c87bf12c6c0177eb830fcc85cf083231de/pkg/client/client.go#L258-L270
Here is the fix for the obvious problem with `(*metadataClient).List`.

vincepri:
@Iceber You could probably add a new test file in the same package to test the internals only

Iceber:
> @Iceber You could probably add a new test file in the same package to test the internals only

Okay, I'll do it. 😀

vincepri:
@Iceber Any updates for a simple test case?

vincepri:
@Iceber do you still have time to tackle this?

Iceber:
@vincepri Sorry for the long delay in adding the test.

PR #1623, opened 2021-08-04
✨ Allow specifying custom TLS Config for webhooks

willthames:
* This implements a Manager option WebhookTLSConfig which allows for
  specifying a custom TLS configuration when running webhooks using
  controller-runtime
* Added CustomTLSConfig attribute to WebhookInstallOptions
* Added test to show this config overrides the config generated by
  CertDir

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -299,7 +299,12 @@ func (te *Environment) Start() (*rest.Config, error) {
 	te.CRDs = crds
 
 	log.V(1).Info("installing webhooks")
-	if err := te.WebhookInstallOptions.Install(te.Config); err != nil {
+	if te.WebhookInstallOptions.CustomTLSConfig != nil {
+		err = te.WebhookInstallOptions.InstallWithCustomTLS(te.Config)
+	} else {
+		err = te.WebhookInstallOptions.Install(te.Config)
+	}
+	if err != nil {
 		return nil, fmt.Errorf("unable to install webhooks onto control plane: %w", err)
 	}
 	return te.Config, nil
--- a/pkg/envtest/webhook.go
+++ b/pkg/envtest/webhook.go
@@ -15,6 +15,7 @@ package envtest
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
@@ -68,6 +69,11 @@ type WebhookInstallOptions struct {
 
 	// LocalServingHostExternalName is the hostname to use to reach the webhook server.
 	LocalServingHostExternalName string
+
+	// CustomTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
+	CustomTLSConfig *tls.Config
 
 	// MaxTime is the max time to wait
 	MaxTime time.Duration
@@ -180,6 +186,25 @@ func (o *WebhookInstallOptions) Install(config *rest.Config) error {
 		if err := o.PrepWithoutInstalling(); err != nil {
 			return err
 		}
+	}
+
+	if err := createWebhooks(config, o.MutatingWebhooks, o.ValidatingWebhooks); err != nil {
+		return err
+	}
+
+	return WaitForWebhooks(config, o.MutatingWebhooks, o.ValidatingWebhooks, *o)
+}
+
+// InstallWithCustomTLS installs specified webhooks to the API server but
+// doesn't setup a CA/cert via tinyca.
+func (o *WebhookInstallOptions) InstallWithCustomTLS(config *rest.Config) error {
+	if err := parseWebhook(o); err != nil {
+		return err
+	}
+
+	err := o.ModifyWebhookDefinitions()
+	if err != nil {
+		return err
 	}
 
 	if err := createWebhooks(config, o.MutatingWebhooks, o.ValidatingWebhooks); err != nil {
--- a/pkg/envtest/webhook_test.go
+++ b/pkg/envtest/webhook_test.go
@@ -18,6 +18,7 @@ package envtest
 
 import (
 	"context"
+	"crypto/tls"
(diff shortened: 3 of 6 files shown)


willthames:
Rebase of #853 

I've preserved the tlsMinVersion change even though the tlsConfig change renders it unnecessary.

As the tlsMinVersion change didn't make it to any release notes and is missing the changes in #1620 that would make it easier to consume, we could just strip out the tlsMinVersion code entirely.

pkg/envtest/server.go
```diff
@@ -299,7 +299,12 @@ func (te *Environment) Start() (*rest.Config, error) {
 	te.CRDs = crds
 
 	log.V(1).Info("installing webhooks")
-	if err := te.WebhookInstallOptions.Install(te.Config); err != nil {
+	if te.WebhookInstallOptions.CustomTLSConfig != nil {
+		err = te.WebhookInstallOptions.InstallWithCustomTLS(te.Config)
+	} else {
+		err = te.WebhookInstallOptions.Install(te.Config)
+	}
```
joelanford, line 306: Do we need a separate `InstallWithCustomTLS` method? Or could we just make `Install` know how to handle the `CustomTLSConfig` field?

pkg/envtest/webhook.go
```diff
@@ -69,6 +70,11 @@ type WebhookInstallOptions struct {
 	// LocalServingHostExternalName is the hostname to use to reach the webhook server.
 	LocalServingHostExternalName string
 
+	// CustomTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
+	CustomTLSConfig *tls.Config
```
joelanford, line 76: What do you think about just calling this `TLSConfig`? IMO the "Custom" prefix is unnecessary.

pkg/envtest/webhook.go
```diff
@@ -69,6 +70,11 @@ type WebhookInstallOptions struct {
 	// LocalServingHostExternalName is the hostname to use to reach the webhook server.
 	LocalServingHostExternalName string
 
+	// CustomTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
```
joelanford, line 75: I think there was a suggestion in the other PR to make setting both CertDir and TLSConfig an error condition, to help users avoid troubleshooting what's happening when both are set.

pkg/envtest/webhook_test.go
```diff
@@ -109,6 +110,76 @@ var _ = Describe("Test", func() {
...
+			dir := env.WebhookInstallOptions.LocalServingCertDir
+			certFile := filepath.Join(dir, "tls.crt")
+			keyFile := filepath.Join(dir, "tls.key")
+			Expect(err).NotTo(HaveOccurred())
+			sCert, err := tls.LoadX509KeyPair(certFile, keyFile)
+			Expect(err).NotTo(HaveOccurred())
+			m, err := manager.New(env.Config, manager.Options{
+				Port: env.WebhookInstallOptions.LocalServingPort,
+				Host: env.WebhookInstallOptions.LocalServingHost,
+				// CertDir: env.WebhookInstallOptions.LocalServingCertDir,
```
joelanford, line 126: Nit: remove this commented line.

pkg/envtest/webhook_test.go
```diff
@@ -109,6 +110,76 @@ var _ = Describe("Test", func() {
...
+							},
+						},
+					},
+				},
+			}
+
+			Eventually(func() bool {
+				err = c.Create(context.TODO(), obj)
+				// Bad Certificate shows up as InternalError
+				return apierrors.ReasonForError(err) == metav1.StatusReason("InternalError")
```
joelanford, line 177: nit:
```suggestion
				return apierrors.IsInternalError(err)
```

pkg/manager/manager.go
```diff
@@ -229,6 +230,11 @@ type Options struct {
 
 	// Functions to all for a user to customize the values that will be injected.
 
+	// WebhookTLSConfig is a *tls.Config object that will be used in place of
+	// the one normally created by reading certs in the CertDir.  If both
+	// CertDir and this are specified CertDir will be ignored
```
joelanford, line 235: Same comment as above about making it an error to specify both CertDir and WebhookTLSConfig.

Perhaps we should also:
1. Introduce an exported function (maybe in the `certwatcher` package?) that can build the same `tls.Config` that is built when `CertDir` and any other user-accessible TLS config settings are used.
2. Deprecate `CertDir` and any related fields, pointing folks to to the function exported from (1)
3. In a later release, remove `CertDir` and related fields.

joelanford (commented):
Thanks for doing this @willthames! Just a few more comments and follow-ups.

pkg/manager/manager.go
joelanford: Hmm. Thinking about this more, it might be too much to ask users to setup the cert watcher and configure it as a runnable in the manager to get equivalent behavior to what they have now.

Thoughts @vincepri @alvaroaleman @varshaprasad96?

alvaroaleman: > Hmm. Thinking about this more, it might be too much to ask users to setup the cert watcher and configure it as a runnable in the manager to get equivalent behavior to what they have now.

Yeah. How about instead using functional like this `opts ...func(*tls.Config)` and if we ultimatively have file(s) in there, start the certwatcher?

willthames: > How about instead using functional opts like this `opts ...func(*tls.Config)`

Sorry, I'm not sure I fully understand that pattern - can you point me at an example/documentation?

joelanford: It's this pattern: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

But I'm not sure exactly what @alvaroaleman has in mind for where to put this.

PR #1627, opened 2021-08-09
✨Add an option to recover panic for controller reconcile

### LINKED ISSUE #797: Reconciler panics should not crash the manager

FillZpp:

Add an option to recover panic for controller reconcile. It ensures that any panics raised by controller reconciler are converted to normal errors.

fixes #797

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -85,6 +85,9 @@ type Controller struct {
 
 	// Log is used to log messages to users during reconciliation, or for example when a watch is started.
 	Log logr.Logger
+
+	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
+	RecoverPanic bool
 }
 
 // watchDescription contains all the information necessary to start a watch.
@@ -95,7 +98,17 @@ type watchDescription struct {
 }
 
 // Reconcile implements reconcile.Reconciler.
-func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
+	if c.RecoverPanic {
+		defer func() {
+			if r := recover(); r != nil {
+				for _, fn := range utilruntime.PanicHandlers {
+					fn(r)
+				}
+				err = fmt.Errorf("panic: %v [recovered]", r)
+			}
+		}()
+	}
 	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
 	ctx = logf.IntoContext(ctx, log)
 	return c.Do.Reconcile(ctx, req)
@@ -295,7 +308,7 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
 
 	// RunInformersAndControllers the syncHandler, passing it the Namespace/Name string of the
 	// resource to be synced.
-	result, err := c.Do.Reconcile(ctx, req)
+	result, err := c.Reconcile(ctx, req)
 	switch {
 	case err != nil:
 		c.Queue.AddRateLimited(req)
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -88,6 +88,39 @@ var _ = Describe("controller", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(result).To(Equal(reconcile.Result{Requeue: true}))
 		})
+
+		It("should not recover panic if RecoverPanic is false by default", func() {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			defer func() {
+				Expect(recover()).ShouldNot(BeNil())
+			}()
+			ctrl.Do = reconcile.Func(func(context.Context, reconcile.Request) (reconcile.Result, error) {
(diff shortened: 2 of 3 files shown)


pkg/internal/controller/controller.go
```diff
@@ -321,6 +322,15 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
 	}
 }
 
+func (c *Controller) doReconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
+	defer func() {
```
alvaroaleman, line 326: This should be optional and opt-in, similiar to what upstream does: https://github.com/kubernetes/apimachinery/blob/f916759cb6b8547418dc7708876ecab5c1961448/pkg/util/runtime/runtime.go#L29

Also, please add tests for this.

FillZpp: Okay. I have added tests and changed all `recover()` to `utilruntime.HandleCrash()`, so that developers can set it via `utilruntime. ReallyCrash`.

pkg/internal/controller/controller.go
```diff
@@ -95,7 +95,10 @@ type watchDescription struct {
 }
 
 // Reconcile implements reconcile.Reconciler.
-func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
```
alvaroaleman, line 98: Sorry, what I meant by the other comment was:
1. Please add an option to `Controller` to make this configurable (i.E. `RecoverPanic bool`)
2. Make the handling here conditional (i.E. `if c.RecoverPanic { defer recover() }`)
3. Expose the option in `pkg/controller.Options`


The reason for favoring that over the current approach is that with the current approach we require ppl to read internal packages to understand how they can get panics recovered, which is not good UX.

Sorry for being unclear about this before.

FillZpp: Understand, I will fix it. So what do you think of webhook handler? Should its panic default to be recovered or another option field to configure it?

alvaroaleman: It should also have an option to set this up and by default panic

FillZpp: Updated. It is hard to add the option for webhook. Maybe we can firstly add an option for controller reconcile.

pkg/internal/controller/controller.go
```diff
@@ -95,7 +98,17 @@ type watchDescription struct {
...
+	if c.RecoverPanic {
+		defer func() {
+			if r := recover(); r != nil {
+				for _, fn := range utilruntime.PanicHandlers {
+					fn(r)
+				}
+				err = fmt.Errorf("panic: %v [recovered]", r)
```
alvaroaleman, line 108: Could we add the stack here, otherwise this is almost impossible to debug? From `runtime/debug.Stack()`

pkg/internal/controller/controller_test.go
```diff
@@ -88,6 +88,41 @@ var _ = Describe("controller", func() {
...
+			defer func() {
+				Expect(recover()).ShouldNot(BeNil())
+			}()
+			ctrl.Do = reconcile.Func(func(context.Context, reconcile.Request) (reconcile.Result, error) {
+				var res *reconcile.Result
+				return *res, nil
+			})
+			_, _ = ctrl.Reconcile(ctx,
+				reconcile.Request{NamespacedName: types.NamespacedName{Namespace: "foo", Name: "bar"}})
+			By("this should not be called")
```
alvaroaleman, line 105: Nit: Is this needed? The `defer` should make the test fail if it doesn't work or not?

pkg/internal/controller/controller.go
FillZpp: Maybe the `utilruntime.PanicHandlers` above can print the stack of panic more properly 
 https://github.com/kubernetes/apimachinery/blob/f916759cb6b8547418dc7708876ecab5c1961448/pkg/util/runtime/runtime.go#L61

pkg/internal/controller/controller_test.go
FillZpp: Have removed this.

alvaroaleman (approved):
Thank you!

PR #1636, opened 2021-08-19
✨return a bool from AddFinalizer and RemoveFinalizer

hatfieldbrian:
When a finalizer is added or removed to an object, typically a client requests the object be updated on the server.  Instead of making an update request unconditionally, it may be desirable to do so only when the object is modified, as an optimization.  Since the add and remove finalizer routines do not return an indication whether a finalizer was add or removed respectively, a caller needs to determine this information some other way.

One way is to call `ContainsFinalizer` before calling `AddFinalizer` or `RemoveFinalizer`.  But the time complexity of each of those routines is linear, resulting in two linear time functions when a finalizer is added or removed.

Another solution is to save the length of a the finalizer slice before calling add or remove finalizer and checking if the length is different upon return.  This improves over the previous solution presuming `len()` of a slice is a constant time complexity operation.

This solution provided in this pull request simplifies further by not requiring the add or remove finalizer caller save slice length prior to call.  Instead, the caller can simply call add or remove finalizer and if it returns true, then make the object update request.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -349,26 +349,31 @@ func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
 type MutateFn func() error
 
 // AddFinalizer accepts an Object and adds the provided finalizer if not present.
-func AddFinalizer(o client.Object, finalizer string) {
+// It returns an indication of whether it updated the object's list of finalizers.
+func AddFinalizer(o client.Object, finalizer string) (finalizersUpdated bool) {
 	f := o.GetFinalizers()
 	for _, e := range f {
 		if e == finalizer {
-			return
+			return false
 		}
 	}
 	o.SetFinalizers(append(f, finalizer))
+	return true
 }
 
 // RemoveFinalizer accepts an Object and removes the provided finalizer if present.
-func RemoveFinalizer(o client.Object, finalizer string) {
+// It returns an indication of whether it updated the object's list of finalizers.
+func RemoveFinalizer(o client.Object, finalizer string) (finalizersUpdated bool) {
 	f := o.GetFinalizers()
 	for i := 0; i < len(f); i++ {
 		if f[i] == finalizer {
 			f = append(f[:i], f[i+1:]...)
 			i--
+			finalizersUpdated = true
 		}
 	}
 	o.SetFinalizers(f)
+	return
 }
 
 // ContainsFinalizer checks an Object that the provided finalizer is present.
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -700,6 +700,62 @@ var _ = Describe("Controllerutil", func() {
 			})
 		})
 
+		Describe("AddFinalizer, which returns an indication of whether it modified the object's list of finalizers,", func() {
+			deploy = &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Finalizers: []string{},
+				},
+			}
+
+			When("the object's list of finalizers has no instances of the input finalizer", func() {
+				It("should return true", func() {
+					Expect(controllerutil.AddFinalizer(deploy, testFinalizer)).To(BeTrue())
+				})
+				It("should add the input finalizer to the object's list of finalizers", func() {
+					Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{testFinalizer}))
+				})
+			})
+
+			When("the object's list of finalizers has an instance of the input finalizer", func() {
+				It("should return false", func() {
+					Expect(controllerutil.AddFinalizer(deploy, testFinalizer)).To(BeFalse())
+				})
+				It("should not modify the object's list of finalizers", func() {
+					Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{testFinalizer}))
+				})
+			})
+		})
+
+		Describe("RemoveFinalizer, which returns an indication of whether it modified the object's list of finalizers,", func() {
+			When("the object's list of finalizers has no instances of the input finalizer", func() {
+				It("should return false", func() {
+					Expect(controllerutil.RemoveFinalizer(deploy, testFinalizer1)).To(BeFalse())
+				})
+				It("should not modify the object's list of finalizers", func() {
+					Expect(deploy.ObjectMeta.GetFinalizers()).To(Equal([]string{testFinalizer}))
+				})
+			})
+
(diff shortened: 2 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -348,27 +348,31 @@ func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
...
 
-// AddFinalizer accepts an Object and adds the provided finalizer if not present.
-func AddFinalizer(o client.Object, finalizer string) {
+// AddFinalizer accepts an Object and adds the provided finalizer and returns true if not present.
+func AddFinalizer(o client.Object, finalizer string) bool {
 	f := o.GetFinalizers()
 	for _, e := range f {
 		if e == finalizer {
-			return
+			return false
```
varshaprasad96, line 356: This addition does seem useful. But the other way which I could think of it is this:
1. Case 1: Finalizer is present and user wants it to be or finalizer is not present and user wants it removed - In this case, we can just use `ConatinsFinalizer()`, which is a linear operation of complexity O(n).
2. Case 2: When the finalizer is present and user wants it removed or vice versa: Here there needs to be an update to the object which needs to happen. As you have explained, we would have to use `ContainsFinalizer` + `Add/RemoveFinalizer` where there would be linear iteration twice. The complexity still remains linear. I am not sure if there would be much performance improvement by saving on one iteration, since updating the object is anyway a necessity. 

We could still return a signal in Add or Remove Finalizer saying it has been updated or not, but I think boolean may not be apt. Returning `false` in `AddFinalizer` can also be interpreted by users that it has not been added to the object at all (though with proper docs we could explain that its just adding element to a list). We can instead maybe return an indication that is already present. 

```go
type FinalizerStatus string

const (
	FinalizerUpdated    FinalizerStatus = "updated"
	FinalizerNotUpdated FinalizerStatus = "notupdated"
)

// AddFinalizer accepts an Object and adds the provided finalizer if not present.
func AddFinalizer(o client.Object, finalizer string) FinalizerStatus {
	f := o.GetFinalizers()
	for _, e := range f {
		if e == finalizer {
			return FinalizerUpdated
		}
	}
	o.SetFinalizers(append(f, finalizer))
	return FinalizerNotUpdated
}
```

However, this would be a breaking change. Also, I think the [finalizer](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/finalizer/finalizer.go) library already implements this in a better way.

hatfieldbrian: Thanks for the fast response!

The finalizer library contains one of the issues I intend to solve, namely two linear iterations, e.g. for Add:
https://github.com/kubernetes-sigs/controller-runtime/blob/4e7f0c968b90a306fb9ef22f8e92fa36b98b4ef1/pkg/finalizer/finalizer.go#L57
https://github.com/kubernetes-sigs/controller-runtime/blob/4e7f0c968b90a306fb9ef22f8e92fa36b98b4ef1/pkg/finalizer/finalizer.go#L58
It does provide a bool indicating whether the object was updated: https://github.com/kubernetes-sigs/controller-runtime/blob/4e7f0c968b90a306fb9ef22f8e92fa36b98b4ef1/pkg/finalizer/finalizer.go#L59

The functions are called `AddFinalizer` and `RemoveFinalizer` so I think it is intuitive and logical to interpret a boolean return code as an indication of whether they did what their names indicate they should.

I would be ok with wrapping the boolean in a struct just to give it a name that the programmer must type, like the library does, but its struct has more than one field.

The string solution takes more work to match.

I propose naming the bool return value `finalizersUpdated` in the function signature to avoid misinterpretation.  I updated the pull request with that solution.

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -677,27 +677,33 @@ var _ = Describe("Controllerutil", func() {
 			}
 
 			It("should add the finalizer when not present", func() {
-				controllerutil.AddFinalizer(deploy, testFinalizer)
+				Expect(controllerutil.AddFinalizer(deploy, testFinalizer)).To(BeTrue())
```
dmvolod, line 680: It would be nice to keep old test "as is" without return validation to check for backward compatibility and create an additional tests for the new functionality validation

hatfieldbrian: A change was force-pushed that leaves the old tests unchanged and adds the new ones

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -700,6 +700,42 @@ var _ = Describe("Controllerutil", func() {
 			})
 		})
 
+		Describe("AddFinalizer", func() {
```
alvaroaleman, line 703: I think you need to change the name, otherwise Ginko de-duplicates, the junit shows two instances of `AddFinalizer` and not four: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1636/pull-controller-runtime-test-master/1512120552670629888

hatfieldbrian: The new tests' descriptions have been changed to be different than the originals

hatfieldbrian:
The `apidiff` job failed:
```sh
verifying api diff
invoking: 'hack/tools/bin/go-apidiff 954449d15797527d32eb9071c4b95b3e44c[26](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1636/pull-controller-runtime-apidiff-master/1514882772970246144#1:build-log.txt%3A26)115 --print-compatible'
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil
  Incompatible changes:
  - AddFinalizer: changed from func(sigs.k8s.io/controller-runtime/pkg/client.Object, string) to func(sigs.k8s.io/controller-runtime/pkg/client.Object, string) bool
  - RemoveFinalizer: changed from func(sigs.k8s.io/controller-runtime/pkg/client.Object, string) to func(sigs.k8s.io/controller-runtime/pkg/client.Object, string) bool
```
It seems an API function may not be changed to return something, so new versions of the routines were defined named `AddFinalizerV2` and `RemoveFinalizerV2`.

alvaroaleman:
> The apidiff job failed:

Yes, but it is not mandatory, it is only intending to inform us about compatibility breaks.

I don't think an added non-error return to a func that didn't return anything before can actually break anyone, because nothing can not be assigned to a variable nor `return`ed and linters won't care either.

alvaroaleman:
I meant to say: Please revert that and keep only one version of the funcs with the old name.

alvaroaleman (approved):
thanks!

PR #1637, opened 2021-08-20
Construct the typed clients we need

justinsb:
:sparkles: Construct typed clients directly

The pattern of constructing all the clients to get a
kubernetes.Interface, only to extract one or two typed clients has the
side-effect of importing a lot of packages that are otherwise
unneeded.

Avoid doing so by constructing the clients we need.

This isn't sufficient to remove all these extra dependencies (because
of the use of scheme.Scheme as default), but it is a step closer.

### DIFF
--- a/pkg/internal/recorder/recorder.go
+++ b/pkg/internal/recorder/recorder.go
@@ -24,8 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes"
-	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 )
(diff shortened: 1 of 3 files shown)


pkg/internal/recorder/recorder.go
```diff
@@ -24,8 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes"
-	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
```
cndoit18, line 27: Maybe we can keep the name: `typedcorev1`

alvaroaleman (approved):

ilimtvtr-alt:
@k8s-ci-robot

ilimtvtr-alt:
@k8s-ci-robot

ilimtvtr-alt (commented):
@k8s-ci-robot 
@alvaroaleman @ilimtvtr-alt @mhrivnak

sbueringer:
@ilimtvtr-alt please stop bumping old PRs like this

PR #1638, opened 2021-08-20
Use proto only if we have proto types

justinsb:
:sparkles: WIP: Use proto only if we have proto definitions

Even if the apiserver supports proto for a type, if we don't know the
proto schema definition we can't use it.

Therefore, detect if the type implements proto.Message when deciding
whether to use proto encoding.

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/discovery"
-	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 )
@@ -40,17 +39,8 @@ var (
 	protobufSchemeLock sync.RWMutex
 )
 
-func init() {
-	// Currently only enabled for built-in resources which are guaranteed to implement Protocol Buffers.
-	// For custom resources, CRDs can not support Protocol Buffers but Aggregated API can.
-	// See doc: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility
-	if err := clientgoscheme.AddToScheme(protobufScheme); err != nil {
-		panic(err)
-	}
-}
-
 // AddToProtobufScheme add the given SchemeBuilder into protobufScheme, which should
-// be additional types that do support protobuf.
+// be additional types where we do want to support protobuf.
 func AddToProtobufScheme(addToScheme func(*runtime.Scheme) error) error {
 	protobufSchemeLock.Lock()
 	defer protobufSchemeLock.Unlock()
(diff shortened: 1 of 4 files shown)


pkg/client/apiutil/apimachinery.go
```diff
@@ -169,3 +144,23 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
...
+func haveProtoDefinition(scheme *runtime.Scheme, gvk schema.GroupVersionKind) bool {
+	gvkType, found := scheme.AllKnownTypes()[gvk]
+	if !found {
+		return false
+	}
+	// We normally get the raw struct e.g. v1.Pod, not &v1.Pod
+	if gvkType.Kind() == reflect.Struct {
+		gvkType = reflect.PtrTo(gvkType)
+	}
+	return gvkType.Implements(protoMessageType)
```
coderanger, line 165: Is this slow enough that we should cache it? I know reflect operations can get pretty cranky sometimes.

FillZpp, line 165: I'm afraid the gvk type has proto definition in scheme doesn't mean the API Server could accept its requests via Protobuf. Actually for a gvk type defined by CRD, API Server can only accept JSON requests even if we implement its proto types in project.

justinsb: Do you know if there's a way to know if the apiserver supports proto (for a given kind)?  I think we need "destination-type-is-proto and apiserver-supports-proto", because if someone queries a PodList into an UnstructuredList we still want to not use proto (?).

BTW the reason that I'm looking at this code is that I'd like to eliminate most of the imports of the kubernetes API types, for controllers that generally just use unstructured.Unstructured.

FillZpp: You can see this PR [Use application/vnd.kubernetes.protobuf as content-type if possible](https://github.com/kubernetes-sigs/controller-runtime/pull/1149) I merged before, which brings Protobuf support into controller-runtime.

```
// Currently only enabled for built-in resources which are guaranteed to implement Protocol Buffers.
// For custom resources, CRDs can not support Protocol Buffers but Aggregated API can.
// See doc: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility
```

So if you want to extend the no-built-in types to use Protobuf, you have to find out whether they are defined by CRD or Aggregated API, which is hard to check. Since all types created via kubebuilder/operator-sdk are in CRD definition, I think the built-in types are enough for most ppl using controller-runtime.

WDYT @alvaroaleman

alvaroaleman:
@justinsb what problem does this PR solve? I agree it is currently a bit hard to discover how to register your api for proto usage, however the only case where this could be wanted is an extension api server that supports proto. To my knowledge, there aren't a lot of them, so this is a bit of a niche problem.

pkg/client/apiutil/apimachinery.go
justinsb: It did involve a linear scan, so I added memoization - thanks!

justinsb: Good point, I guess it is possible (and something I would love to try out!) to use proto to define CRDs locally.

I tweaked the logic therefore, there's a list of built-in types, and we have a unit test to make sure that the core kubernetes types are recognized appropriately.  This should detect if there's a new built-in library added upstream, and I added a comment so that it's obvious how to add the newly-added built-in group.

This also means that we can continue to support AddToProtobufScheme.

justinsb:
Thanks for the reviews and comment @alvaroaleman and @coderanger .  The intention here is actually the other way round; I want to eliminate all dependencies on kubernetes/scheme.  Merely importing that package pulls in all the k8s API types, and that adds up to a lot more memory usage than is really needed (>500 entries in the scheme, instead of around a dozen that are really needed).  It also causes

This is by no means the last place we need this, but ... it's a journey!

Issue #1639, opened 2021-08-21
client.Get() does not zero out slice fields on CRD objects


nathanperkins:
I've encountered some unusual behavior when using Get() on CRD objects with the controller-runtime client. When the client.Object you pass into Get() has a slice field populated with values and the object on the API server has an empty or nil slice for that field, the client does not properly zero/empty those fields. Other field types like int and string update fine.

This is in contrast with the behavior for core objects, which does zero/empty the fields as expected.

The example test below demonstrates the inconsistency. I tested this on v0.7.0 and v0.9.6.

```
var _ = Describe("client.Get", func() {
	ctx := context.Background()

	var (
		crdObject  *v1alpha1.CRDType
		coreObject *corev1.ConfigMap
	)

	BeforeEach(func() {
		crdObject = &v1alpha1.Type{
			ObjectMeta: metav1.ObjectMeta{
				Name:      "crd-object",
				Namespace: "default",
				// Finalizers empty.
			},
		}
		coreObject = &corev1.ConfigMap{
			ObjectMeta: metav1.ObjectMeta{
				Name:      "core-object",
				Namespace: "default",
				// Finalizers empty.
			},
		}
	})

	JustBeforeEach(func() {
		Expect(k8sClient.Create(ctx, crdObject)).Should(Succeed())
		Expect(k8sClient.Create(ctx, coreObject)).Should(Succeed())
	})

	AfterEach(func() {
		Expect(k8sClient.Delete(ctx, crdObject)).Should(Succeed())
		Expect(k8sClient.Delete(ctx, coreObject)).Should(Succeed())
	})

	Context("CRD object", func() {
		It("Should overwrite slices that should be empty", func() {
			crdObject.Finalizers = []string{"1.1.1.1"}

			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(crdObject), crdObject)).
				Should(Succeed())
      			
			// Fails with crdObject.Finalizers = []string{"1.1.1.1"}
			Expect(crdObject.Finalizers).To(BeEmpty())
		})
	})

	Context("Core object", func() {
		It("Should overwrite slices that should be empty", func() {
			coreObject.Finalizers = []string{"1.1.1.1"}

			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(coreObject), coreObject)).
				Should(Succeed())
      
			// Passes.
			Expect(coreObject.Finalizers).To(BeEmpty())
		})
	})
})
```

[...]

PR #1640, opened 2021-08-22
:bug: Client: Ensure no stale data remains in target object

alvaroaleman:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1639

The json deserializer of the stdlib and the one from Kube which aims to
be compatible won't zero out all field types in the object it
deserializes into, for example it lets slices be if the json does not
contain that field. This means that if a non-empty variable is used for
any api call with the client, the resulting content might be a mixture
of previous content and what is on the server.

This PR adds a wrapper around the deserializer that will first zero the
target object.

/assign @joelanford @vincepri

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -21,6 +21,7 @@ package apiutil
 
 import (
 	"fmt"
+	"reflect"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


vincepri:
Shouldn't clients make sure that the object passed in is clean?

alvaroaleman:
> Shouldn't clients make sure that the object passed in is clean?

Ideally yes, realistically there are a bunch of situation where that might not happen, for example in "Wait for change" loops. An extreme example of this is `Update`, if for example a mutatingwebhook would patch a slice out, the returned object would still have it (as we re-use the passed-in object to decode the result into).

coderanger:
Is the overhead small enough to handle this unconditionally? We could probably get away with a heuristic like skipping it if .Name == "" but I do like how clean the code is so if we think it's fast enough, maybe just leave it as is.

alvaroaleman:
> Is the overhead small enough to handle this unconditionally? We could probably get away with a heuristic like skipping it if .Name == "" but I do like how clean the code is so if we think it's fast enough, maybe just leave it as is.

I haven't done any benchmarking so I don't know how much performance overhead exactly this entails. I did check the generated clients though, which always and uncoditionally allocate a new object to serialize into.

It should also be noted that this only affects the api client. I would think that its generally unlikely someone makes so many mutating request that the performance overhead of this becomes noticeable and for reading requests, it would make sense to use the cache-backed client if one cares about performance. That one doesn't have this issue, as it doesn't deserialize json into the passed object.

vincepri:
Not too worried about performance impacts of this change, but more of masking user errors. This change is definitely a behavioral change in the client code (should be marked as a breaking change), although I'm not sure if we should zero out objects altogether for users.

Shouldn't our users be responsible to either clean out the object or just use a new one?

alvaroaleman:
> Not too worried about performance impacts of this change, but more of masking user errors. This change is definitely a behavioral change in the client code (should be marked as a breaking change), although I'm not sure if we should zero out objects altogether for users.



It is a change in behavior, but I can not think of an example where someone would want existing data plus the response to be mixed in an object which led me to the conclusion that this is not going to break anyones code, hence it is marked as bugfix.

> Shouldn't our users be responsible to either clean out the object or just use a new one?

I don't think we have anything to that effect documented. It also isn't really possible to test this for an existing codebase. I can say that I have written code that re-uses a variable to fetch data to avoid unnecessary allocations (And assuming wrongly that the deserialization would make there the object gets cleared)

joelanford:
/lgtm

/hold
For @vincepri in case he's got a legit example of this being a breaking change.

alvaroaleman:
Just testing something...
/cherrypick release-0.9

alvaroaleman:
Lets give this another shot...
/cherrypick release-0.9

alvaroaleman:
Lets try again:
/cherrypick release-0.9

spiffxp:
/cherrypick release-0.9
testing again to see if the new bot is used

spiffxp:
/cherrypick release-0.9
now that https://github.com/kubernetes/test-infra/pull/23556 landed

PR #1651, opened 2021-09-03
🐛 Fix object overwrite on patch for fake client

FillZpp:

Previously the fake client tried to decode the new object after patching to the object outside, which will cause some fields that should be removed in this patch remain in the object.

It should overwrite the object using reflect set, just like [the way in cache get](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L99).

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -284,6 +285,7 @@ func (c *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj client.O
 		return err
 	}
 	decoder := scheme.Codecs.UniversalDecoder()
+	zero(obj)
 	_, _, err = decoder.Decode(j, nil, obj)
 	return err
 }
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -544,13 +545,13 @@ func (c *fakeClient) Patch(ctx context.Context, obj client.Object, patch client.
 	ta.SetKind(gvk.Kind)
 	ta.SetAPIVersion(gvk.GroupVersion().String())
 
-	j, err := json.Marshal(o)
-	if err != nil {
-		return err
+	objVal := reflect.ValueOf(obj)
```
alvaroaleman, line 548: If we use reflect to assign the value, we need to deepcopy first

alvaroaleman (commented):
This is essentially the same issue as https://github.com/kubernetes-sigs/controller-runtime/issues/1639 which was fixed by zeroing the target object in the decoder (but your approach is likely more efficient here).

Could you also do this for `Get` and `List`? They  have the same issue

/ok-to-test

FillZpp:
> This is essentially the same issue as #1639 which was fixed by zeroing the target object in the decoder (but your approach is likely more efficient here).
> 
> Could you also do this for `Get` and `List`? They have the same issue

Sure. But I find that we can not simply use `reflect.Set` to overwrite the object, for it fails when use an unstructured object to get a typed object in client, which causes `AssignableTo` check error like this:

```
client had type *v1.DeploymentList, but *unstructured.UnstructuredList was asked for
```

So I have to *zero* each object before decode into it, just like the thing you do in `targetZeroingDecoder`.

alvaroaleman (approved):

Issue #1511, opened 2021-05-04
fakeClient.List tries to populate UnstructuredList with objects of underlying type


wbrefvem:
If I have an UnstructuredList with a known GVK, passing it to `fakeClient.List` returns an error:

```
item[0]: can't assign or convert unstructured.Unstructured into <myGV>.<myKind>
```

Running in the debugger revealed that `fakeClient.List` passes the underlying type to `client-go`, which returns a list of the underlying type. So it makes sense that fake client would error when trying to populate the `UnstructuredList` with objects it gets from `client-go`.

There should be some way for `fakeClient.List` to handle `Unstructured*` objects. IIUC the real client delegates to an unstructured client. That seems heavy-handed here, but it may be the only way to avoid a bunch of spaghetti in `fakeClient.List` just to handle `UnstructuredList` as a special case.

This is related to #702 but is not the same issue.

alvaroaleman:
@wbrefvem I am a bit surprised this fails, we have testcases for this: https://github.com/kubernetes-sigs/controller-runtime/blob/745c7c98a7223069fcaab6b2e1b373f40d82e15a/pkg/client/fake/client_test.go#L124-L142

> Running in the debugger revealed that fakeClient.List passes the underlying type to client-go, which returns a list of the underlying type. So it makes sense that fake client would error when trying to populate the UnstructuredList with objects it gets from client-go.

We don't do that, we serialize and deserialize: https://github.com/kubernetes-sigs/controller-runtime/blob/745c7c98a7223069fcaab6b2e1b373f40d82e15a/pkg/client/fake/client.go#L333-L338

* Is your type registered in the Scheme you are using?
* Would it be possible for you to provide a small repro?

wbrefvem:
> Is your type registered in the Scheme you are using?

Yes.

> Would it be possible for you to provide a small repro?

controller-runtime version: v0.9.0-beta.0

I'll try to be as brief as possible:

```
func (v *ValidateHandler) Handle(ctx context.Context, request admission.Request) admission.Response {
        // logic that retrieves the GVK for this type omitted
        // "service" and "serviceGVK" refer to the unkown CR type, which encapsulate configs for microservices
        // the value of labelSelector is unique to the type, e.g. for type Foo it's something like `myService==foo`
        // the key point is that the labelSelectors for the existing CR and the incoming request are identical, because the types are the same.

	unstructuredServiceList := &unstructured.UnstructuredList{}
[...]

PR #1662, opened 2021-09-11
:bug: Fakeclient: Allow manipulating registered types through unstructured

alvaroaleman:
Currently, manipulating an object that is registered through
unstructured will break all subsequent List requests that include said
object. This happens because the tracker unconditionally saves objects but
it's List() assumes that what it finds can be assigned to the .Items slice
of an object produced through scheme.New(ListGVK).

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1511

/assign @vincepri @joelanford

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -169,6 +169,11 @@ func (t versionedTracker) Add(obj runtime.Object) error {
 			// be recognized
 			accessor.SetResourceVersion(trackerAddResourceVersion)
 		}
+
+		obj, err = convertFromUnstructuredIfNecessary(t.scheme, obj)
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@joelanford was it intended to not lgtm the change and only approve?

vincepri (approved):
/lgtm

nathaniel-lehrer-ib:
Hello, thanks for the fix :) I need to upgrade my packages to get this fix -- which version/package is this fixed in? I'm currently using:
```
	k8s.io/api v0.21.4
	k8s.io/apimachinery v0.21.4
	k8s.io/client-go v0.21.4
```

Issue #1654, opened 2021-09-03
rest.Config.NegotiatedSerializer causes watches to be created with the wrong scheme


howardjohn:
We are seeing errors like `watch of *v1alpha2.Gateway ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"Gateway\" is registered for version \"gateway.networking.k8s.io/v1alpha2\" in scheme \"pkg/runtime/scheme.go:100\"") has prevented the request from succeeding?`

But...
* It works with a standard client-go informer
* I turned on debug logs, got the exact watch request to the api-server, and run it with kubectl get --raw and it works
* The controller actually works - its reading the resource its failing to watch. It almost seems like it can list but not watch?

After a lot of debugging, the root cause was setting `rest.Config.NegotiatedSerializer` in the rest config passed to the manager creation.

I am not sure if this is a bug or user error. If its a bug, the root cause is here: https://github.com/kubernetes-sigs/controller-runtime/blob/8f9d0dfc84464e8a8fd5f5def0837d748be3bb8f/pkg/client/apiutil/apimachinery.go#L161

The reason we have it set is we copy+pasted from https://github.com/kubernetes/kubectl/blob/652881798563c00c1895ded6ced819030bfaa4d7/pkg/cmd/util/kubectl_match_version.go#L115. I am not sure if there is a valid reason beyond that.

PR #1664, opened 2021-09-13
:bug: Unconditionally set NegotiatedSerializer

alvaroaleman:
We currently only set the NegotiatedSerializer when there is none set
yet. Unfortunately, various things will break if someone sets their own
in the rest config, hence always set it.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1654

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -158,14 +158,12 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
 		protobufSchemeLock.RUnlock()
 	}
 
-	if cfg.NegotiatedSerializer == nil {
-		if isUnstructured {
-			// If the object is unstructured, we need to preserve the GVK information.
-			// Use our own custom serializer.
-			cfg.NegotiatedSerializer = serializerWithDecodedGVK{serializer.WithoutConversionCodecFactory{CodecFactory: codecs}}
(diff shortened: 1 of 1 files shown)


pkg/client/apiutil/apimachinery.go
```diff
@@ -158,14 +158,12 @@ func createRestConfig(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
...
 
-	if cfg.NegotiatedSerializer == nil {
-		if isUnstructured {
-			// If the object is unstructured, we need to preserve the GVK information.
-			// Use our own custom serializer.
-			cfg.NegotiatedSerializer = serializerWithDecodedGVK{serializer.WithoutConversionCodecFactory{CodecFactory: codecs}}
-		} else {
-			cfg.NegotiatedSerializer = serializerWithTargetZeroingDecode{NegotiatedSerializer: serializer.WithoutConversionCodecFactory{CodecFactory: codecs}}
-		}
+	if isUnstructured {
```
alvaroaleman, line 161: Not entirely sure if we should mark this at breaking. It is bu tut then again things are unlikely to have worked if anyone set this before.

alvaroaleman: Note that we copy the config already so we won't manipulate the original one

vincepri (approved):
/approve
/lgtm

PR #1665, opened 2021-09-16
🐛  envtest: get gvk from hook struct instead of forcing set TypeMeta

sbueringer:

Without this fix it's necessary to set TypeMeta in ValidatingWebhookConfiguration and MutatingWebhookConfiguration, e.g. like this:
```golang
				gvk, err := apiutil.GVKForObject(webhook, scheme.Scheme)
				if err != nil {
					klog.Fatalf("failed to get gvk for ValidatingWebhookConfiguration %s", o.GetName())
				}
				webhook.SetGroupVersionKind(gvk)
```

With this fix we will just get the type from the "struct".

xref: https://github.com/kubernetes-sigs/cluster-api/pull/5249/files#r709959229


vincepri (approved):
/approve
/lgtm

PR #1666, opened 2021-09-16
🐛  envtest: fix CRD installation

sbueringer:

Without this PR the CRD installation did not work:
* it panic'ed when `.Conversion` is nil
* it didn't add the Conversion config (because of copy in the array)
* we also have to add `ConversionReviewVersions`

Notes:
* in our case the base CRD files don't have conversion as they are added via patches: https://github.com/kubernetes-sigs/cluster-api/blob/master/config/crd/bases/cluster.x-k8s.io_clusters.yaml
* this didn't occur with v0.9 because there the conversion webhook was modified via unstructured. (this changed through this PR: https://github.com/kubernetes-sigs/controller-runtime/commit/b5eeb7106d2360ba7e282c66cd7c65d6c57c6be4#diff-b6b1b02e41870fbffb46bc5880965e53a92a1c6807bb3fae0e8473154fb416a4)

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -359,21 +359,26 @@ func modifyConversionWebhooks(crds []apiextensionsv1.CustomResourceDefinition, s
 	}
 	url := pointer.StringPtr(fmt.Sprintf("https://%s/convert", hostPort))
 
-	for _, c := range crds {
+	for i := range crds {
 		// Continue if we're preserving unknown fields.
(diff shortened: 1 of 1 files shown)


vincepri (approved):
/approve
/lgtm

sbueringer:
> /hold there is at least one other issue, I'll investigate

/hold cancel 
already fixed it and updated the PR description accordingly

PR #1676, opened 2021-09-27
:sparkles: Allow webhooks to register custom validators/defaulter types

vincepri:
This changeset allows our webhook builder to take in a handler any other
struct other than a runtime.Object.

Today having an object as the primary source of truth for both
Defaulting and Validators makes API types carry a lot of information and
business logic alongside their definitions.

Moreover, lots of folks in the past have asked for ways to have an
external type to handle these operations and use a controller runtime
client for validations.

This change brings a new way to register webhooks, which admission.For
handler any type (struct) can be a defaulting or validating handler for
a runtime Object.

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -17,6 +17,7 @@ limitations under the License.
 package builder
 
 import (
+	"errors"
 	"net/http"
 	"net/url"
 	"strings"
@@ -32,10 +33,12 @@ import (
 
 // WebhookBuilder builds a Webhook.
 type WebhookBuilder struct {
-	apiType runtime.Object
-	gvk     schema.GroupVersionKind
-	mgr     manager.Manager
-	config  *rest.Config
+	apiType       runtime.Object
+	withDefaulter admission.CustomDefaulter
+	withValidator admission.CustomValidator
+	gvk           schema.GroupVersionKind
+	mgr           manager.Manager
+	config        *rest.Config
 }
 
 // WebhookManagedBy allows inform its manager.Manager.
@@ -50,6 +53,18 @@ func WebhookManagedBy(m manager.Manager) *WebhookBuilder {
 // If the given object implements the admission.Validator interface, a ValidatingWebhook will be wired for this type.
 func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 	blder.apiType = apiType
+	return blder
+}
+
+// WithDefaulter takes a admission.WithDefaulter interface, a MutatingWebhook will be wired for this type.
+func (blder *WebhookBuilder) WithDefaulter(defaulter admission.CustomDefaulter) *WebhookBuilder {
+	blder.withDefaulter = defaulter
+	return blder
+}
+
+// WithValidator takes a admission.WithValidator interface, a ValidatingWebhook will be wired for this type.
+func (blder *WebhookBuilder) WithValidator(validator admission.CustomValidator) *WebhookBuilder {
+	blder.withValidator = validator
 	return blder
 }
 
@@ -69,9 +84,13 @@ func (blder *WebhookBuilder) loadRestConfig() {
 }
 
 func (blder *WebhookBuilder) registerWebhooks() error {
+	typ, err := blder.getType()
+	if err != nil {
+		return err
+	}
+
 	// Create webhook(s) for each type
-	var err error
-	blder.gvk, err = apiutil.GVKForObject(blder.apiType, blder.mgr.GetScheme())
+	blder.gvk, err = apiutil.GVKForObject(typ, blder.mgr.GetScheme())
 	if err != nil {
 		return err
 	}
@@ -88,12 +107,7 @@ func (blder *WebhookBuilder) registerWebhooks() error {
 
 // registerDefaultingWebhook registers a defaulting webhook if th.
 func (blder *WebhookBuilder) registerDefaultingWebhook() {
-	defaulter, isDefaulter := blder.apiType.(admission.Defaulter)
-	if !isDefaulter {
-		log.Info("skip registering a mutating webhook, admission.Defaulter interface is not implemented", "GVK", blder.gvk)
-		return
-	}
-	mwh := admission.DefaultingWebhookFor(defaulter)
+	mwh := blder.getDefaultingWebhook()
 	if mwh != nil {
 		path := generateMutatePath(blder.gvk)
 
@@ -108,13 +122,21 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
 	}
 }
 
-func (blder *WebhookBuilder) registerValidatingWebhook() {
-	validator, isValidator := blder.apiType.(admission.Validator)
-	if !isValidator {
-		log.Info("skip registering a validating webhook, admission.Validator interface is not implemented", "GVK", blder.gvk)
-		return
+func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
+	if defaulter := blder.withDefaulter; defaulter != nil {
+		return admission.WithCustomDefaulter(blder.apiType, defaulter)
 	}
-	vwh := admission.ValidatingWebhookFor(validator)
+	if defaulter, ok := blder.apiType.(admission.Defaulter); ok {
+		return admission.DefaultingWebhookFor(defaulter)
+	}
+	log.Info(
+		"skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called",
+		"GVK", blder.gvk)
+	return nil
+}
+
+func (blder *WebhookBuilder) registerValidatingWebhook() {
+	vwh := blder.getValidatingWebhook()
 	if vwh != nil {
 		path := generateValidatePath(blder.gvk)
 
@@ -127,6 +149,19 @@ func (blder *WebhookBuilder) registerValidatingWebhook() {
 			blder.mgr.GetWebhookServer().Register(path, vwh)
 		}
 	}
+}
+
+func (blder *WebhookBuilder) getValidatingWebhook() *admission.Webhook {
+	if validator := blder.withValidator; validator != nil {
+		return admission.WithCustomValidator(blder.apiType, validator)
+	}
+	if validator, ok := blder.apiType.(admission.Validator); ok {
+		return admission.ValidatingWebhookFor(validator)
+	}
+	log.Info(
+		"skip registering a validating webhook, object does not implement admission.Validator or WithValidator wasn't called",
+		"GVK", blder.gvk)
+	return nil
 }
 
 func (blder *WebhookBuilder) registerConversionWebhook() error {
@@ -143,6 +178,13 @@ func (blder *WebhookBuilder) registerConversionWebhook() error {
 	}
 
 	return nil
+}
+
+func (blder *WebhookBuilder) getType() (runtime.Object, error) {
+	if blder.apiType != nil {
+		return blder.apiType, nil
+	}
+	return nil, errors.New("For() must be called with a valid object")
 }
 
 func (blder *WebhookBuilder) isAlreadyHandled(path string) bool {
--- a/pkg/webhook/admission/defaulter_custom.go
+++ b/pkg/webhook/admission/defaulter_custom.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package admission
+
+import (
+	"context"
+	"encoding/json"
+
+	"errors"
+	"net/http"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// CustomDefaulter defines functions for setting defaults on resources.
+type CustomDefaulter interface {
+	Default(ctx context.Context, obj runtime.Object) error
+}
+
+// WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
+func WithCustomDefaulter(obj runtime.Object, defaulter CustomDefaulter) *Webhook {
+	return &Webhook{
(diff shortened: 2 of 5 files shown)


pkg/webhook/admission/defaulter_handler.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"net/http"
+
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// DefaulterFor defines functions for setting defaults on resources.
+type DefaulterFor interface {
+	For
+	Default(ctx context.Context, obj runtime.Object)
```
fabriziopandini, line 30:
```suggestion
	Default(ctx context.Context, obj runtime.Object) error
```

fabriziopandini (commented):
I see two aspect for this PR
- Allowing registering webhook implemented outside of the API types
- Improving validator and default interface with context and always returning error.

For the first part, I have the impression that what we need is just somethings that implements the two methods defined in runtime.Object, but I still need to investigate this a little bit. 
For the second part, might be worth also considering a breaking change, given that those seems best practices to me

pkg/webhook/admission/defaulter_handler.go
vincepri: Why?

fabriziopandini: This could be handy if for any reason the default logic could fail (e.g we make an API call inside the defaulter)

vincepri: Fair enough :), added

vincepri:
> For the first part, I have the impression that what we need is just somethings that implements the two methods defined in runtime.Object, but I still need to investigate this a little bit.

Would that be a preferred solution to this problem? It seems like that overwriting the deep copy methods is a hacky way to force an external type into the current webhooks builder logic.

vincepri:
> For the second part, might be worth also considering a breaking change, given that those seems best practices to me
This is something we can consider breaking in a future release, once we can streamline the interface as well

fabriziopandini:
>> For the first part, I have the impression that what we need is just somethings that implements the two methods defined in runtime.Object, but I still need to investigate this a little bit.
>
> Would that be a preferred solution to this problem? It seems like that overwriting the deep copy methods is a hacky way to force an external type into the current webhooks builder logic.

That's right, I'm trying to understand if we can be explicit using HandlerFor in the builder but avoiding to duplicate the Handle code in validation/defaulter handler

pkg/builder/webhook.go
```diff
@@ -53,6 +55,15 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 	return blder
 }
 
+// HandlerFor takes a admission.admissionFor interface.
+//
+// If the given object implements the admission.DefaulterFor interface, a MutatingWebhook will be wired for this type.
+// If the given object implements the admission.ValidatorFor interface, a ValidatingWebhook will be wired for this type.
+func (blder *WebhookBuilder) HandlerFor(forType admission.For) *WebhookBuilder {
```
coderanger, line 62: Would something like `WebhookFor` be better? I worry that `Handler` is such a generic term.

vincepri: Not sure if that's clear either, wouldn't the `WebhookBuilder` be in charge of returning the webhook?

sbueringer: I tried to do that, but I couldn't get it to work with the DeepCopy and subsequent Decode. (because the webhook struct is not registered as type for the decoding). Do you have a link to an existing implementation?

pkg/builder/webhook.go
```diff
@@ -108,10 +118,22 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
 	}
 }
 
+func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
+	if defaulter, ok := blder.apiType.(admission.Defaulter); ok {
+		return admission.DefaultingWebhookFor(defaulter)
+	}
+	if defaulter, ok := blder.forType.(admission.DefaulterFor); ok {
```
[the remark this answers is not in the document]
sbueringer: I think I like how it would be less magical that way. Could/Should we use the type set via `WebhookBuilder.For` in that case?

pkg/builder/webhook.go
```diff
@@ -108,10 +118,22 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
...
+		return admission.DefaulterForToWebhook(defaulter)
+	}
+	log.Info(
+		"skip registering a mutating webhook, admission.Defaulter or admission.DefaulterFor interface is not implemented",
+		"GVK", blder.gvk)
+	return nil
+}
+
 func (blder *WebhookBuilder) registerValidatingWebhook() {
 	validator, isValidator := blder.apiType.(admission.Validator)
```
sbueringer, line 135: Must also be refactored like `registerDefaultingWebhook`

pkg/builder/webhook.go
alvaroaleman: ah ok, I didn't test it I just assumed that it should work.

pkg/builder/webhook.go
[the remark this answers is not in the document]
alvaroaleman: Yeah, just continuing to use the `For` would be the idea

pkg/builder/webhook.go
sbueringer: Okay, I just thought maybe there is a way and I just couldn't find it :)

pkg/builder/webhook.go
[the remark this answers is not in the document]
vincepri: PTAL, I've adjusted to remove the magical aspect of it

pkg/webhook/admission/defaulter_handler.go
```diff
@@ -0,0 +1,81 @@
...
+
+// WithDefaulter defines functions for setting defaults on resources.
+type WithDefaulter interface {
+	Default(ctx context.Context, obj runtime.Object) error
+}
+
+// WithCustomDefaulter creates a new Webhook for a WithDefaulter interface.
+func WithCustomDefaulter(obj runtime.Object, defaulter WithDefaulter) *Webhook {
+	return &Webhook{
+		Handler: &defaulterForType{handler: defaulter},
```
sbueringer, line 37: `obj` must also be set

pkg/webhook/admission/defaulter_handler.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+type defaulterForType struct {
+	handler WithDefaulter
```
sbueringer, line 42:
```suggestion
	defaulter WithDefaulter
```
nit, Not sure if it's worth it but we could also call this defaulter to keep it more similar to the other defaulter implementation (but it's probably also the better name)

pkg/webhook/admission/validator_handler.go
```diff
@@ -0,0 +1,122 @@
...
+
+	v1 "k8s.io/api/admission/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// WithValidator defines functions for validating an operation.
+type WithValidator interface {
+	ValidateCreate(ctx context.Context, obj runtime.Object) error
+	ValidateUpdate(ctx context.Context, old runtime.Object, new runtime.Object) error
```
sbueringer, line 32:
```suggestion
	ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error
```
nit

pkg/webhook/admission/validator_handler.go
```diff
@@ -0,0 +1,122 @@
...
+
+// WithCustomValidator creates a new Webhook for validating the provided type.
+func WithCustomValidator(obj runtime.Object, validator WithValidator) *Webhook {
+	return &Webhook{
+		Handler: &validatorForType{obj: obj, handler: validator},
+	}
+}
+
+type validatorForType struct {
+	handler WithValidator
```
sbueringer, line 44:
```suggestion
	validator WithValidator
```
nit (same reasons as for defaulter)

pkg/webhook/admission/validator_handler.go
```diff
@@ -0,0 +1,125 @@
...
+	"net/http"
+
+	v1 "k8s.io/api/admission/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// WithValidator defines functions for validating an operation.
+type WithValidator interface {
+	ValidateCreate(ctx context.Context, obj runtime.Object) error
```
alvaroaleman, line 31: Can we use `client.Object` for these?

pkg/webhook/admission/defaulter_handler.go
```diff
@@ -0,0 +1,81 @@
...
+	"encoding/json"
+	goerrors "errors"
+	"net/http"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// WithDefaulter defines functions for setting defaults on resources.
+type WithDefaulter interface {
```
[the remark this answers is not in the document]
alvaroaleman: Yeah, I am not super happy about it either. From the suggestions above I like `ObjectDefaulter` the most

pkg/webhook/admission/defaulter_handler.go
```diff
@@ -0,0 +1,84 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package admission
+
+import (
+	"context"
+	"encoding/json"
+	goerrors "errors"
```
alvaroaleman, line 22: Nit: Can we call this just `errors` without the alias like everywhere else?

pkg/webhook/admission/validator_handler.go
```diff
@@ -0,0 +1,125 @@
...
+	if h.validator == nil {
+		panic("validator should never be nil")
+	}
+	if h.object == nil {
+		panic("object should never be nil")
+	}
+
+	// Get the object in the request
+	obj := h.object.DeepCopyObject()
+	if req.Operation == v1.Create {
```
alvaroaleman, line 68: Nit: We could switch on the operation to be clear that these are mutually excluse (and maybe have a erorring `default` clause). Also, the error final error handling can be moved out of the conditions and deduplicated

pkg/webhook/admission/validator_handler.go
vincepri: Not really, unless we make it a breaking change? I was trying to change any interface until 0.11, so we can backport this change

pkg/webhook/admission/defaulter_handler.go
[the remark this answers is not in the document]
vincepri: Went with `CustomDefaulter`, `ObjectDefaulter` sounds like what the current interface should be called 😄

pkg/webhook/admission/validator_handler.go
vincepri: Done! ptal

pkg/builder/webhook.go
```diff
@@ -32,10 +33,12 @@ import (
 
 // WebhookBuilder builds a Webhook.
 type WebhookBuilder struct {
-	apiType runtime.Object
-	gvk     schema.GroupVersionKind
-	mgr     manager.Manager
-	config  *rest.Config
+	apiType       runtime.Object
+	withDefaulter admission.CustomDefaulter
+	withValidator admission.CustomValidator
```
sbueringer, line 38:
```suggestion
	customDefaulter admission.CustomDefaulter
	customValidator admission.CustomValidator
```
nit (also fine for me to keep it as is, but compared to the other fields it seems consistent to me to drop the with prefix)

pkg/webhook/admission/validator_custom.go
```diff
@@ -0,0 +1,108 @@
...
+		err = h.validator.ValidateUpdate(ctx, oldObj, obj)
+	case v1.Delete:
+		// In reference to PR: https://github.com/kubernetes/kubernetes/pull/76346
+		// OldObject contains the object being deleted
+		if err := h.decoder.DecodeRaw(req.OldObject, obj); err != nil {
+			return Errored(http.StatusBadRequest, err)
+		}
+
+		err = h.validator.ValidateDelete(ctx, obj)
+	}
```
sbueringer, line 95: Do we want to return an error in the default case?

pkg/webhook/alias.go
```diff
@@ -29,6 +29,12 @@ type Defaulter = admission.Defaulter
 // Validator defines functions for validating an operation.
 type Validator = admission.Validator
 
+// WithDefaulter defines functions for setting defaults on resources.
+type WithDefaulter = admission.CustomDefaulter
+
+// WithValidator defines functions for validating an operation.
+type WithValidator = admission.CustomValidator
```
sbueringer, line 36:
```suggestion
type CustomDefaulter = admission.CustomDefaulter

// WithValidator defines functions for validating an operation.
type CustomValidator = admission.CustomValidator
```

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -0,0 +1,85 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// CustomDefaulter defines functions for setting defaults on resources.
+type CustomDefaulter interface {
+	Default(ctx context.Context, obj runtime.Object) error
+}
+
+// WithCustomDefaulter creates a new Webhook for a WithDefaulter interface.
```
sbueringer, line 35:
```suggestion
// WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
```
nit (or something generic like the comment of `WithCustomValidator`)

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -0,0 +1,85 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
sbueringer, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```
nit (same in validator_custom.go)

sbueringer:
No more nits from my side :)

Thank you very much for implementing this!!

sbueringer:
/hold 
@vincepri  Not sure if you want to wait for feedback from others.

pkg/webhook/admission/validator_handler.go
alvaroaleman: These are new interfaces, how would using `client.Object` here be breaking?

I was thinking that this would make usage slightly better, i.E. you wouldn't be able to pass a List object in here

vincepri: While the interfaces are new, the object passed in comes from the webhook builder, we'd have to cast it before passing creating the webhook. Moreover, when we use DeepCopyObject() that returns a `runtime.Object`, which we'd have to cast again if we want to use client.Object everywhere

alvaroaleman:
/hold cancel
Thanks for your work!

PR #1681, opened 2021-10-04
🐛  Start the Cache if the Manager has already started

### LINKED ISSUE #1673: Add cluster after Manager has already started
I am working on a multi-cluster operator project. We following the changes described in https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/move-cluster-specific-code-out-of-manager.md initializing a `Cluster` for each remote k8s cluster. It works great, so thanks for that :)

I would like to add (and remove) `Clusters` at runtime after the `Manager` has already started. It doesn't look like that is possible today. I have been trying to get more familiar with the `Manager` code to get a better sense of what I need to do to manage the `Cluster` on my own. It definitely looks non-trivial. Maybe I am wrong but it looks like I would have to reimplement a fair bit of logic from `controllerManager`.

This brings me to my question. Are there any plans to support adding a `Cluster` to the `Manager` after it has been started? What are some of the caveats with doing this?

jsanda:
Fixes #1673. 

This is needed to better support multi-cluster controllers where you want to dynamically add Clusters after the Manager has already started and have their caches started and synced.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -211,6 +211,12 @@ func (cm *controllerManager) Add(r Runnable) error {
 		cm.nonLeaderElectionRunnables = append(cm.nonLeaderElectionRunnables, r)
 	} else if hasCache, ok := r.(hasCache); ok {
 		cm.caches = append(cm.caches, hasCache)
+		if cm.started {
+			cm.startRunnable(hasCache)
+			if !hasCache.GetCache().WaitForCacheSync(cm.internalCtx) {
+				return fmt.Errorf("could not sync cache")
+			}
+		}
 	} else {
 		shouldStart = cm.startedLeader
 		cm.leaderElectionRunnables = append(cm.leaderElectionRunnables, r)


jsanda:
@alvaroaleman if my changes look reasonable I can add a new test in manager_test.go.

pkg/manager/internal.go
```diff
@@ -211,6 +211,12 @@ func (cm *controllerManager) Add(r Runnable) error {
 		cm.nonLeaderElectionRunnables = append(cm.nonLeaderElectionRunnables, r)
 	} else if hasCache, ok := r.(hasCache); ok {
 		cm.caches = append(cm.caches, hasCache)
+		if cm.started {
+			cm.startRunnable(hasCache)
+			if !hasCache.GetCache().WaitForCacheSync(cm.internalCtx) {
```
alvaroaleman, line 216: This will never timeout, right? Does that match the standard path, where we start the caches before the `cm` is started?

/ok-to-test

jsanda: It could definitely time out. The remote cluster might not be accessible. Should the cache be started and synced with a different context? If the sync times out, I would then cancel the context. I suppose it also makes more sense to append the cache after the sync completes successfully.

I believe this matches the standard path.  I can't simply call `waitForCache` because it returns immediately if `cm.started` is true. It first iterates through the caches and calls `cm.startRunnable` for each one. Then it iterates through the caches again and calls `cache.GetCache().WaitForCacheSync(ctx)` where `ctx` is `cm.internalCtx`.

jsanda: After looking through the code some more,  and the `WaitForCacheSync` function in `shared_informer.go` in client-go in particular, I am less certain that it will time out.

jsanda:
@alvaroaleman I added a test. I added a fake cluster impl as a hook to verify the cluster is started. It feels a bit clunky. Open to suggestions if you have a better/simpler approach.

alvaroaleman:
thanks!

Issue #1685, opened 2021-10-07
Webhooks are not guaranteed to start before cache sync is started


randomvariable:
In the process of debugging https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2834 we spot the following behaviour:

If resources are present in API Server/etcd at an older storage version AND  `--leader-election` is set to false when starting controllers with a newer CRD version, then there is a high probability that conversion webhooks will not start before cache syncs start, which due to conversion webhooks not being present and a mutex being held, indefinitely delay the start of the webhook server. On some infrastructures, this is closer to 90% (e.g. Amazon EC2 t3.large (2 CPU, 8GB RAM)).

When leader election is enabled, then the probability drops to 20%, and when using healthchecks for the webhooks, sufficient restarts by kubelet lead to the conversion webhooks running sufficiently early to allow cache syncs to occur.

https://github.com/kubernetes-sigs/controller-runtime/commit/4e548fa8011587ccf1e6f6b98f9b3747da98d285 was intended to ensure that webhooks start first, but this doesn't seem to have had the desired effect.

We believe that the occurrence is lower when leader election happens since the leader election process delays the call of `cm.startLeaderElectionRunnables()` until [after election](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L481) is completed. 

Example logs are here: [capi-logs.txt]

[capi-logs.txt]: https://github.com/kubernetes-sigs/controller-runtime/files/7303307/capi-logs.txt

Controller-runtime-version: v0.10.2
Affected consumers: Cluster API v1.0.0

cc @fabriziopandini @sbueringer

PR #1690, opened 2021-10-11
🐛 Start web hooks first

fabriziopandini:
This PR test a simpler and possible cleaner solution to #1685.
We should test how this impact start time

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -476,6 +476,11 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 		go cm.serveHealthProbes()
 	}
 
+	// Webhooks MUST start before any cache is populated, otherwise there is a race condition
+	// between conversion webhooks and the cache sync (usually initial list) which causes the webhooks
+	// to never start because no cache can be populated.
+	cm.startWebhookRunnables()
+
 	go cm.startNonLeaderElectionRunnables()
 
 	go func() {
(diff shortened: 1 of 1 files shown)


fabriziopandini:
I will take a look about tests tomorrow morning, at least this is a chance to learn this part as well

alvaroaleman:
> @alvaroaleman When I first moved the code to run webhooks first there wasn't a nice way to do it from the outside.

Maybe just set up a manager with a webhook and a runnable that does a http request to the webhook in its `Start()`? It is a bit annoying because if we race like before it will flake and not reliably fail, but I guess that is unavoidable when it comes to race conditions that are not data races?

fabriziopandini:
@vincepri rebased on to of the release 0.10 branch

sbueringer:
@fabriziopandini Can you please include the change in: https://github.com/fabriziopandini/controller-runtime/pull/1 ?

The problem is that otherwise we'll have the same deadlock just with the healthprobes.

I would then take this version and rebase my 2 test PRs iN CAPI to run all our tests with and without leader election

fabriziopandini:
> @fabriziopandini Can you please include the change in: fabriziopandini#1 ?

Done!

sbueringer:
Thx! Tests are running here:
* https://github.com/kubernetes-sigs/cluster-api/pull/5421
* https://github.com/kubernetes-sigs/cluster-api/pull/5407

P.S. I might need a few tries, just refactored those branches back to use CR release-0.10 instead of main.

sbueringer:
@fabriziopandini WDYT, ready for review/squash/un-wip? :)

vincepri (approved):
/lgtm
/milestone v0.10

Issue #1708, opened 2021-10-29
Cache selectors should allow setting a default


alvaroaleman:
It is currently possible to set a selector in the cache per GVK, however it is not possible to set a default selector which might be very useful for scenarios like "I want all caches in this namespace except the cache for footype, which should be in a different namespace".

joelanford:
I definitely agree that a default selector would be useful in many cases (e.g. label `my.domain/managedBy == "my-operator"`), but I'm not sure a namespace selector is appropriate at this layer.

If I'm understanding this correctly, it would result in the informers using a cluster-wide list/watch and the apiserver receiving _all_ objects from etcd before filtering down to the desired namespace.

Clients using this setup would require cluster-wide RBAC list/watch for all of the watched kinds and (barring any gaps in my understanding of apiserver <-> etcd communication) would cause unnecessary resource usage from objects not in the selected namespaces being unnecessarily sent from etcd to the apiserver.

It seems like the `GVKCaches` approach in #1591 would be better from an RBAC and resource utilization perspective.

alvaroaleman:
> If I'm understanding this correctly, it would result in the informers using a cluster-wide list/watch and the apiserver receiving all objects from etcd before filtering down to the desired namespace.

No, the selector can be used to not even receive the objects and make the apiserver filter for us. Using a default namespace was just an example btw, could also be a default selector. What exactly makes sense depends on what you want to do :)

joelanford:
> No, the selector can be used to not even receive the objects and make the apiserver filter for us.

Understood, but if `cache.Options.Namespace` is unset, the underlying apiserver call is going to be cluster-scoped regardless of the selectors. So it would generally be a bad idea to use a cluster-scoped informer with a namespace selector for the reasons I stated.

Nonetheless, I agree a default selector is valuable.

alvaroaleman:
No, we had a change a while back that made us do a namespace-scoped call and only require rbac for that namespace if a fieldselector for a namespace is set: https://github.com/kubernetes-sigs/controller-runtime/pull/1602

joelanford:
Ah ha! That was the piece I was missing. Very cool!

PR #1710, opened 2021-11-01
:sparkles: Allow configuring a default cache selector

alvaroaleman:
It is already possible to configure cache selectors per gvk, but it is
not possible to default this selector for all types. This change adds
that.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1708

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -90,6 +90,8 @@ type Informer interface {
 type ObjectSelector internal.Selector
 
 // SelectorsByObject associate a client.Object's GVK to a field/label selector.
+// There is also `DefaultSelector` to set a global default (which will be overridden by
+// a more specific setting here, if any).
 type SelectorsByObject map[client.Object]ObjectSelector
 
 // Options are the optional arguments for creating a new InformersMap object.
@@ -117,6 +119,10 @@ type Options struct {
 	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
 	SelectorsByObject SelectorsByObject
 
+	// DefaultSelector will be used as selectors for all object types
+	// that do not have a selector in SelectorsByObject defined.
+	DefaultSelector ObjectSelector
+
 	// UnsafeDisableDeepCopyByObject indicates not to deep copy objects during get or
 	// list objects per GVK at the specified object.
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
@@ -132,7 +138,7 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	selectorsByGVK, err := convertToSelectorsByGVK(opts.SelectorsByObject, opts.Scheme)
+	selectorsByGVK, err := convertToSelectorsByGVK(opts.SelectorsByObject, opts.DefaultSelector, opts.Scheme)
 	if err != nil {
 		return nil, err
 	}
@@ -194,7 +200,7 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 	return opts, nil
 }
 
-func convertToSelectorsByGVK(selectorsByObject SelectorsByObject, scheme *runtime.Scheme) (internal.SelectorsByGVK, error) {
+func convertToSelectorsByGVK(selectorsByObject SelectorsByObject, defaultSelector ObjectSelector, scheme *runtime.Scheme) (internal.SelectorsByGVK, error) {
 	selectorsByGVK := internal.SelectorsByGVK{}
 	for object, selector := range selectorsByObject {
 		gvk, err := apiutil.GVKForObject(object, scheme)
@@ -203,6 +209,7 @@ func convertToSelectorsByGVK(selectorsByObject SelectorsByObject, scheme *runtim
 		}
 		selectorsByGVK[gvk] = internal.Selector(selector)
 	}
+	selectorsByGVK[schema.GroupVersionKind{}] = internal.Selector(defaultSelector)
 	return selectorsByGVK, nil
 }
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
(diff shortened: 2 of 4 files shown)


pkg/cache/cache.go
```diff
@@ -90,8 +91,22 @@ type Informer interface {
...
+// selector for all kinds that do not have a specific selector set up.
+type SelectorsByObjectDefaultKey struct{ metav1.Object }
+
+// GetObjectKind implements runtime.Object.
+func (s *SelectorsByObjectDefaultKey) GetObjectKind() schema.ObjectKind {
+	return schema.EmptyObjectKind
+}
+
+// DeepCopyObject implements runtime.Object.
+func (s *SelectorsByObjectDefaultKey) DeepCopyObject() runtime.Object { return s }
```
joelanford, line 108: Rather than encoding the default selector into the `SelectorsByObject` option which requires this new exported `SelectorsByObjectDefaultKey` type, is it possible to add a `DefaultSelector` field to the options struct and then handle the setup internally?

alvaroaleman: That's a great idea, updated the PR, PTAL

FillZpp (commented):
In fact, now we already have [ObjectAll](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/cache.go#L215) and [GroupVersionKindAll](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/disabledeepcopy.go#L22).

For `type DisableDeepCopyByObject map[client.Object]bool`, If developers use `ObjectAll` as its key, it will be the default value. So how about reuse `ObjectAll` in `SelectorsByObject` to avoid defining a new option?

joelanford:
@FillZpp Brings up a good point that there's already a mechanism for this with `ObjectAll`.

I'm torn though. On one hand, I think a separate `Default*` field is much better from a user discoverability standpoint. On the other, there is already precedent for storing the default value in the map with a special key.

Interested on folks thoughts about introducing `Default*` fields for all of these `*ByObject` maps?

alvaroaleman:
> In type DisableDeepCopyByObject map[client.Object]bool, If developers use ObjectAll as its key, it will be the default value. So how about reuse ObjectAll in SelectorsByObject to avoid defining a new option?

Right, that is where I got the idea from, I thought I remembered we did that but then couldn't find it :)

So tbh I am thorn as well, the UX of having a top-level field is much better than a magic key but does that warrant making things differently? Also I feel like the selector option warrants it because it is likely going to be used more than the somewhat exotic `DisableDeepCopyByObject` but I might be biased in that opinion as I need one but not the other.

To not get stuck in this question I would suggest to only add a magic key and not a top-level option for now. If we later find that ppl want to use this but have trouble finding the magic key we can still add a top-level option without breaking compatibility (by copying its value into the magic key entry). WDYT @joelanford ?

alvaroaleman:
@joelanford could you have another look, please?

pkg/cache/cache.go
```diff
@@ -117,6 +119,10 @@ type Options struct {
 	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
 	SelectorsByObject SelectorsByObject
 
+	// DefaultSeletor will be used as selectors for all object types
+	// that do not have a selector in SelectorsByObject defined.
+	DefaultSeletor ObjectSelector
```
vincepri, line 124:
```suggestion
	DefaultSelector ObjectSelector
```

pkg/cache/cache.go
```diff
@@ -90,6 +90,8 @@ type Informer interface {
 type ObjectSelector internal.Selector
 
 // SelectorsByObject associate a client.Object's GVK to a field/label selector.
+// Use SelectorsByObjectDefaultKey as key to specify a default selector that
+// will be used for all types that do not have a selector configured.
```
vincepri, line 94: Is this comment still valid?

pkg/cache/cache.go
```diff
@@ -203,6 +209,7 @@ func convertToSelectorsByGVK(selectorsByObject SelectorsByObject, scheme *runtim
 		}
 		selectorsByGVK[gvk] = internal.Selector(selector)
 	}
+	selectorsByGVK[schema.GroupVersionKind{}] = internal.Selector(defaultSelector)
```
fabriziopandini, line 212: nit not blocking 
const defalutSelectorKey = schema.GroupVersionKind{}

pkg/cache/cache.go
alvaroaleman: err after the back-and-forth regarding magic key vs config option I decided to use the magic key, as it is easy to add a config option later but not easy to remove it again: https://github.com/kubernetes-sigs/controller-runtime/pull/1710#issuecomment-962218495

Apparently I forgot to update the code though :grimacing: did that now.

I ended up not re-using the `ObjectAll` because it is a default that doesn't apply to everything, just to everything that doesn't have a setting so the name might cause confusion.

pkg/cache/cache.go
alvaroaleman: Needs to be a var but good idea, updated the code.

alvaroaleman:
/retest
The failure seems unrelated

pkg/cache/cache.go
FillZpp: > I ended up not re-using the ObjectAll because it is a default that doesn't apply to everything, just to everything that doesn't have a setting so the name might cause confusion.

Emm.. I'm confused that the **default** and **all** here seem to be the same semantics? They all mean the default configuration for all kinds of object, except the objects that have their specific values.

pkg/cache/cache_test.go
```diff
@@ -93,6 +121,76 @@ var _ = Describe("Multi-Namespace Informer Cache", func() {
...
+		for idx, namespace := range []string{testNamespaceOne, testNamespaceTwo} {
+			_ = createSA("test-sa-"+strconv.Itoa(idx), namespace, cl)
+			_ = createSvc("test-svc-"+strconv.Itoa(idx), namespace, cl)
+		}
+
+		opts := cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				&corev1.ServiceAccount{}:             {Field: fields.OneTermEqualSelector("metadata.namespace", testNamespaceOne)},
+				&cache.SelectorsByObjectDefaultKey{}: cache.ObjectSelector{Field: fields.OneTermEqualSelector("metadata.namespace", testNamespaceTwo)},
+			},
```
vincepri, line 150: Looking at how this is being used, if I understood what we're trying to do here correctly, we can summarize it as:
- For objects of type ServiceAccount, always cache only the objects in namespace 1
- For any other object, cache objects in namespace 2

If that's the case, I'm not super sure if I like this APIs, it's confusing that the fallback is also a "selector by object", which seems like a wildcard, although it could be confusing to users.

Have we thought instead of adding another option, like `DefaultSelector`, or similar?

alvaroaleman: @vincepri yes, see the discussion here: https://github.com/kubernetes-sigs/controller-runtime/pull/1710#issuecomment-961179320

vincepri: I would suggest to not add and deprecate any magic key in favor of a different field, it makes a nicer UX and easier understanding of the codebase. Without this test/example, it wouldn't have been clear what the PR was trying to achieve and how.

alvaroaleman: Okay, reverted it back to add a top-level `DefaultSelector` field, ptal.

vincepri (approved):

PR #1716, opened 2021-11-13
:sparkles: Fix SetupSignalHandler godoc

sgreene570:
This PR is a small follow up to https://github.com/kubernetes-sigs/controller-runtime/pull/1205, which updated `SetupSignalHandler` in `pkg/manager/signals/signal.go` to use a context instead of a stop channel without updating the godoc for that function.


alvaroaleman (approved):
thanks!

PR #1718, opened 2021-11-15
🐛 sample validatingwebhook annotation key

varas:
This webhook is a validating, not a mutating one, key value seemed misleading.


fabriziopandini:
+1 from my side, changes seems ok

vincepri (approved):
/lgtm
/ok-to-test

PR #1739, opened 2021-12-07
✨ Pass caller context instead creating new in ListWatcher

ardaguclu:
This PR passes context to ListWatcher instead creating a new one,
basically fixes the mentioned as TODO.

### DIFF
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -102,9 +102,6 @@ type specificInformersMap struct {
 	// paramCodec is used by list and watch
 	paramCodec runtime.ParameterCodec
 
-	// stop is the stop channel to stop informers
-	stop <-chan struct{}
-
 	// resync is the base frequency the informers are resynced
 	// a 10 percent jitter will be added to the resync period between informers
 	// so that all informers will not send list requests simultaneously.
@@ -135,6 +132,9 @@ type specificInformersMap struct {
 
 	// disableDeepCopy indicates not to deep copy objects during get or list objects.
 	disableDeepCopy DisableDeepCopyByGVK
+
+	// ctx is the main context to be used managing watchers.
+	ctx context.Context
 }
 
 // Start calls Run on each of the informers and sets started to true.  Blocks on the context.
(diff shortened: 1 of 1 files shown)


pkg/cache/internal/informers_map.go
```diff
@@ -192,7 +192,7 @@ func (ip *specificInformersMap) Get(ctx context.Context, gvk schema.GroupVersion
 
 	if !ok {
 		var err error
-		if i, started, err = ip.addInformerToMap(gvk, obj); err != nil {
+		if i, started, err = ip.addInformerToMap(ctx, gvk, obj); err != nil {
```
alvaroaleman, line 195: I think the context we need to use here is the one we get from `Start`, `Get` is called for example in the implementation of `client.Reader` and the `ctx` we pass there is assumed to be the lifetime of the request, not the lifetime of the underlying informer: https://github.com/kubernetes-sigs/controller-runtime/blob/558ef2e0b7c5735aff64aa5a1f72a298744f456f/pkg/cache/informer_cache.go#L54

/hold

ardaguclu: Thanks for review @alvaroaleman.

I deliberately did this. Because just focusing on Get requests in watchers(https://github.com/kubernetes-sigs/controller-runtime/blob/f236f0345ad2933912ebf34bfcf0f93620769654/pkg/cache/internal/informers_map.go#L284) made me think; their lifetimes can be connected to parent's `GET` requests.

But that is a mistake and you are right. Although it is not a suggested thing to use context as field in struct, it seems the only way to pass `Start`'s context to watchers. 

If you don't like the idea, we can leave code as is and close the PR. But maybe it is a good time to resolve at least one of TODO's :), what do you think?

ardaguclu: After looking at deeper, this change makes context life cycle management much more complex. If parent context is cancelled, watchers need to handle this and create new one, etc.  

It is safer to leave that as is, since we are not getting much benefit comparing to the complexity comes along.

ardaguclu:
I'm closing this PR due to the reasons stated in here https://github.com/kubernetes-sigs/controller-runtime/pull/1739#discussion_r765510778

/close

pkg/cache/internal/informers_map.go
alvaroaleman: Yeah, we would have to store the `context` we get in `Start`, we do the same in the `Manager` to be able to manage `Runnables` that are added after `Start`. I don't really see a way around that.

Issue #1660, opened 2021-09-10
It is too easy to create duplicate caches in different projections


maelvls:
Since v0.7.0, it has become possible to use a "metadata-only" cache (https://github.com/kubernetes-sigs/controller-runtime/pull/1174). One caveat that is (I think) not highlighted enough in the Go comments is that you may end up writing racy code when mixing multiple types of clients.

For example, the following snippet watches the same type (v1.Pod) using two watchers:

- a metadata-only client used by the watcher,
- a concrete client used in the reconciliation function.

A simple use-case could be the one of reconciling ReplicaSets whenever a Pod changes.

```go
ctrl.NewControllerManagedBy(mgr).
	For(&appsv1.ReplicaSet{}).
	Watches(
		&source.Kind{Type: &corev1.Pod{}},
		&handler.EnqueueRequestForOwner{OwnerType: &corev1.ReplicaSet{}, IsController: true},
		builder.OnlyMetadata, // 🔥
	).
	Complete(reconcile.Func(func(_ context.Context, r reconcile.Request) (reconcile.Result, error) {
		// This is the ReplicaSet reconciliation function. Let's imagine that a pod
		// has to be fetched in this reconciliation.

        nsAndName := findPodAssociatedWithRS(r.NamespaceAndName)
		pod := &corev1.Pod{}
		_ = mgr.GetClient().Get(context.Background(), nsAndName, pod) // 🔥

        // ...
	}))
```

When a Pod changes, the apiserver sends a `MODIFIED` event twice since two watchers are running (one is watching `metav1.PartialMetadataObject`, the other `v1.Pod`). This means that the metadata cache may get updated first, triggering the reconciliation, while the second has not been propagated.

This race may lead to a "not found" error in the `Get` call above. The issue https://github.com/kubernetes-sigs/controller-runtime/issues/1454 also talks about this unexpected behavior.

**Update:** I had no motivation or purpose in mixing the metadata and concrete projections. I did not know that I was mixing things. I just wanted to use the medatada projection.

Does it make sense to document this caveat?

alvaroaleman:
> When a Pod changes, the apiserver sends a MODIFIED event twice since two watchers are running (one is watching metav1.PartialMetadataObject, the other v1.Pod)

Sorry, I don't follow. The code you are showing results in two watches, one for ReplicaSets and one for Pods in the metadata projection, but not in two watches for pods in different projections.

[...]

PR #1747, opened 2021-12-14
📖 Explain what may happen when using OnlyMetadata and Get on the concrete type instead of PartialMetadataObject

maelvls:
In #1660, we found that it is too easy for users to duplicate cache projections, for example by creating a watch with `OnlyMetadata`, which creates a metadata cache, and then using the concrete type to get the watched object, which automatically creates a cache of the concrete projection.

Duplicating the cached projections is probably never what the user wanted to do, and leads to:
- Increased memory usage,
- Race conditions since caches aren't synchronized.

At first, I wanted to follow the suggestion in https://github.com/kubernetes-sigs/controller-runtime/issues/1660#issuecomment-918242097:

> We could add a check for this into pkg/cache and error out if someone requests something that we already have in a different projection

I then realized that I don't really know how to make this change.

Can I suggest starting with a quick warning that people would see when looking at  `OnlyMetadata`? In this PR, I added the warning.

To sum up my goal in this PR, I wish to document the following:

1. The three following Get calls will create three different caches:

	```go
	// Concrete client:
	pod := &v1.Pod{}
	mgr.GetClient().Get(ctx, nsAndName, pod)
	
	// Unstructured client:
	pod := &unstructured.Unstructured{}
	mgr.GetClient().Get(ctx, nsAndName, pod)

	// Metadata-only client:
	pod := &metav1.PartialObjectMetadata{}
	mgr.GetClient().Get(ctx, nsAndName, pod)
	```

2. Caches are not synchronized, meaning that you may end up with data races when using two different cache projections.
3. Using OnlyMetadata option should not be used with the concrete nor the unstructured clients.

### DIFF
--- a/pkg/builder/options.go
+++ b/pkg/builder/options.go
@@ -107,6 +107,20 @@ var (
 	// metav1.PartialObjectMetadata to the client when fetching objects in your
 	// reconciler, otherwise you'll end up with a duplicate structured or
 	// unstructured cache.
+	//
+	// When watching a resource with OnlyMetadata, for example the v1.Pod, you
+	// should not Get and List using the v1.Pod type. Instead, you should use
+	// the special metav1.PartialObjectMetadata type:
+	//
+	//   pod := &v1.Pod{}                             // ❌
+	//   mgr.GetClient().Get(ctx, nsAndName, pod)     // ❌
+	//
+	//   pod := &metav1.PartialObjectMetadata{}       // ✅
+	//   mgr.GetClient().Get(ctx, nsAndName, pod)     // ✅
+	//
+	// In the first case, controller-runtime will create another cache for the
+	// concrete type on top of the metadata cache; this increases memory
+	// consumption and leads to race conditions as caches are not in sync.
 	OnlyMetadata = projectAs(projectAsMetadata)
 
 	_ ForOption     = OnlyMetadata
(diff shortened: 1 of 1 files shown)


joelanford:
@maelvls is this still a work in progress? As far as I can tell, this looks like a great clarification and something we could go ahead and merge.

maelvls:
Thanks, I think it is ready to go!

pkg/builder/options.go
```diff
@@ -107,6 +107,20 @@ var (
...
 	// unstructured cache.
+	//
+	// When watching a resource with OnlyMetadata, for example the v1.Pod, you
+	// should not Get and List using the v1.Pod type. Instead, you should use
+	// the special metav1.PartialObjectMetadata type:
+	//
+	//   pod := &v1.Pod{}                             // ❌
+	//   mgr.GetClient().Get(ctx, nsAndName, pod)     // ❌
+	//
+	//   pod := &metav1.PartialObjectMetadata{}       // ✅
```
alvaroaleman, line 118: Doesn't this need typemeta set

maelvls: That's a very good point, controller-runtime would have a hard time figuring out which type we are referring to with just `&metav1.PartialObjectMetadata{}` 😅

I'll fix this.

alvaroaleman (approved):
thank you!

Issue #1760, opened 2021-12-30
controller will hang when set  renewDeadline  greater than leaseDuration


yuzp1996:
Hi  I  encounter a problem when I set renewDeadline greater than leaseDuration  and the controller will hang and no error will be report

for example 

```
       lease := time.Second * 5
	renew := time.Second * 6
	options := ctrl.Options{LeaderElection: true, LeaderElectionID: "test", LeaseDuration: &lease, RenewDeadline: &renew}
	manager, err := ctrl.NewManager(config, options)
	if err != nil {
		log.Fatalln(err)
	}
	err = manager.Start(ctx)
	if err != nil {
		log.Fatalln(err)
	}
	return
```

PR #1761, opened 2021-12-30
🐛  fix bug when set renewDeadline greater than leaseDuration the control…

yuzp1996:
fix bug when set renewDeadline greater than leaseDuration the controller will hang and no error will be return

Issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1760

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -478,11 +478,9 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 
 	// Start the leader election and all required runnables.
 	{
-		ctx, cancel := context.WithCancel(context.Background())
-		cm.leaderElectionCancel = cancel
 		go func() {
 			if cm.resourceLock != nil {
-				if err := cm.startLeaderElection(ctx); err != nil {
+				if err := cm.startLeaderElection(); err != nil {
 					cm.errChan <- err
 				}
 			} else {
@@ -552,12 +550,14 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 	defer cm.recorderProvider.Stop(cm.shutdownCtx)
 	defer func() {
 		// Cancel leader election only after we waited. It will os.Exit() the app for safety.
-		if cm.resourceLock != nil {
-			// After asking the context to be cancelled, make sure
-			// we wait for the leader stopped channel to be closed, otherwise
-			// we might encounter race conditions between this code
-			// and the event recorder, which is used within leader election code.
+		// After asking the context to be cancelled, make sure
+		// we wait for the leader stopped channel to be closed, otherwise
+		// we might encounter race conditions between this code
+		// and the event recorder, which is used within leader election code.
+		if cm.leaderElectionCancel != nil {
 			cm.leaderElectionCancel()
+		}
+		if cm.leaderElectionStopped != nil {
 			<-cm.leaderElectionStopped
 		}
 	}()
@@ -604,7 +604,7 @@ func (cm *controllerManager) startLeaderElectionRunnables() error {
 	return cm.runnables.LeaderElection.Start(cm.internalCtx)
 }
 
-func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error) {
+func (cm *controllerManager) startLeaderElection() (err error) {
 	l, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
 		Lock:          cm.resourceLock,
 		LeaseDuration: cm.leaseDuration,
@@ -636,6 +636,10 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 	if err != nil {
 		return err
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cm.leaderElectionCancel = cancel
+	cm.leaderElectionStopped = make(chan struct{})
 
 	// Start the leader elector process
 	go func() {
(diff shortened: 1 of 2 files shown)


pkg/manager/internal.go
```diff
@@ -634,6 +634,7 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 		ReleaseOnCancel: cm.leaderElectionReleaseOnCancel,
 	})
 	if err != nil {
```
yuzp1996, line 636: if err happend the leaderElectionStopped will not close and engageStopProcedure will hang in this code
```
     defer func() {
        // Cancel leader election only after we waited. It will os.Exit() the app for safety.
        if cm.resourceLock != nil {
	        // After asking the context to be cancelled, make sure
	        // we wait for the leader stopped channel to be closed, otherwise
	        // we might encounter race conditions between this code
	        // and the event recorder, which is used within leader election code.
	        cm.leaderElectionCancel()
	        <-cm.leaderElectionStopped
        }
  }()
```

so I think when error happened we should close this channel

FillZpp:
@yuzp1996 Yeah, it is a bug here, but not only for the NewLeaderElector error.

The cause of the manager hang during stopping, is the `cm.leaderElectionStopped` might not be closed. Sometimes the `cm.leaderElectionCancel` might be `nil` so it could also panic here. For example it fails to `cm.add(cm.cluster)` or `cm.runnables.Webhooks.Start(cm.internalCtx)` in the `Start()`.

```
        // Cancel leader election only after we waited. It will os.Exit() the app for safety.
        if cm.resourceLock != nil {
	        // After asking the context to be cancelled, make sure
	        // we wait for the leader stopped channel to be closed, otherwise
	        // we might encounter race conditions between this code
	        // and the event recorder, which is used within leader election code.
	        cm.leaderElectionCancel()
	        <-cm.leaderElectionStopped
        }
```

So it might need to change like this, IMHO

1. Init `leaderElectionStopped` only when `NewLeaderElector` succeeds.
2. Should check if `cm.leaderElectionCancel` is nil before `cm.leaderElectionCancel()` and check if `cm.leaderElectionStopped` is nil before `<-cm.leaderElectionStopped`. If so, maybe there is no need to check `if cm.resourceLock != nil` here any more.

pkg/manager/internal.go
```diff
@@ -637,6 +638,10 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 		return err
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
```
yuzp1996, line 641: Following @FillZpp's suggestion, I changed the location of initialization `cm.leaderElectionCancel` and `cm.leaderElectionStopped`.  So these values ​​will not be nil if `NewLeaderElector` succeeds.

 When `engageStopProcedure` needs to check `cm.leaderElectionCancel` and `cm.leaderElectionStopped`, it will first check whether they are nil.

pkg/manager/internal.go
```diff
@@ -552,13 +550,16 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
...
-		if cm.resourceLock != nil {
-			// After asking the context to be cancelled, make sure
-			// we wait for the leader stopped channel to be closed, otherwise
-			// we might encounter race conditions between this code
-			// and the event recorder, which is used within leader election code.
+		// After asking the context to be cancelled, make sure
+		// we wait for the leader stopped channel to be closed, otherwise
+		// we might encounter race conditions between this code
+		// and the event recorder, which is used within leader election code.
+		if cm.leaderElectionCancel != nil {
```
FillZpp, line 557: It seems you only have to check one of `cm.leaderElectionCancel` and `cm.leaderElectionStopped` for now they are assigned at the same place.

FillZpp (commented):
Mostly LG. Some nits.
/ok-to-test

FillZpp:
@yuzp1996 And you have to fix the failed job, which you can test locally with `make lint`.

pkg/manager/internal.go
yuzp1996: Yes but I think maybe checking them individually is safer 😄

yuzp1996:
> @yuzp1996 And you have to fix the failed job, which you can test locally with `make lint`.

FillZpp:
Thanks for fixing this.
/lgtm
/cc @vincepri @alvaroaleman  PTAL

pkg/manager/internal.go
vincepri: I'm not sure if this is correct, if we set the context and cancel here it'll be set in a goroutine which can cause a race condition if the controller closes too quickly.

vincepri (commented):
Why do you want renewDeadline greater than lease duration? The lease should always be greater than its renewal, otherwise you might end up losing leader election which is probably what's happening here.

We could validate that the rule above applies and enforce it when configuring the manager, instead of changing the code around to avoid leader election.

/hold

yuzp1996:
> 

Yes, I think we have had the validation but when someone set renewDeadline greater than lease duration the whole process will block and the error will not throw. I don't think this is the right way to do that so I make some changes to make sure errors can be throw

FillZpp:
@vincepri As I said above, it is not only caused by renewDeadline greater than lease duration.

Any errors occurred between `defer engageStopProcedure` in `controllerManager.Start()` and `// Start the leader elector process` in `controllerManager.startLeaderElection()`, such as:

1. Error of `cm.add(cm.cluster)`
2. Error of `cm.runnables.Webhooks.Start(cm.internalCtx)`
3. Error of `cm.runnables.Caches.Start(cm.internalCtx)`
4. Error of `cm.runnables.Others.Start(cm.internalCtx)`
5. Error of `leaderelection.NewLeaderElector`

They will all return error, trigger `engageStopProcedure` in defer, and finally hang on `<-cm.leaderElectionStopped`, because the channel will not be closed forever.

FillZpp:
WDYT @vincepri

vincepri:
Can we add a failing test that covers the above bug first which can help explain and prevent the deadlock above? I cannot reproduce locally, so it'd be good to have a test for this

PR #1767, opened 2022-01-06
✨ Add Kubernetes Gomega extension with to make testing controllers easier

schrej:
This is an attempt to revive #1364 by @JoelSpeed. It's essentially the same with a few changes:

- I've added the EqualObject matcher from kubernetes-sigs/cluster-api (written by @sbueringer and @killianmuldoon - I hope you're fine with that!)
- Added package global functions so it can be used by importing `. sigs.k8s.io/controller-runtime/pkg/envtest/komega` and use it similar to Gomega
- renamed `Matcher` to `komega` since it's actually not a Gomega Matcher
- changed the `With...` functions to copy komega

This still needs a lot of tests and documentation.

cc @Danil-Grigorev @joelanford @fabriziopandini  since you were interested in the original PR

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,12 +8,12 @@ require (
 	github.com/go-logr/logr v1.2.0
 	github.com/go-logr/zapr v1.2.0
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.17.0
+	github.com/onsi/gomega v1.18.1
 	github.com/prometheus/client_golang v1.11.1
 	github.com/prometheus/client_model v0.2.0
 	go.uber.org/goleak v1.1.12
 	go.uber.org/zap v1.19.1
-	golang.org/x/sys v0.0.0-20211029165221-6e7872819dc8
+	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
 	gomodules.xyz/jsonpatch/v2 v2.2.0
 	k8s.io/api v0.23.0
--- a/pkg/envtest/komega/OWNERS
+++ b/pkg/envtest/komega/OWNERS
@@ -1,0 +1,14 @@
+approvers:
+  - controller-runtime-admins
+  - controller-runtime-maintainers
+  - controller-runtime-approvers
+  - schrej
+  - JoelSpeed
+  - sbueringer
+reviewers:
+  - controller-runtime-admins
+  - controller-runtime-reviewers
+  - controller-runtime-approvers
+  - schrej
+  - JoelSpeed
+  - sbueringer
--- a/pkg/envtest/komega/default.go
+++ b/pkg/envtest/komega/default.go
@@ -1,0 +1,92 @@
+package komega
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// defaultK is the Komega used by the package global functions.
+var defaultK = &komega{ctx: context.Background()}
+
+// SetClient sets the client used by the package global functions.
+func SetClient(c client.Client) {
+	defaultK.client = c
+}
+
+// SetContext sets the context used by the package global functions.
+func SetContext(c context.Context) {
+	defaultK.ctx = c
+}
+
+func checkDefaultClient() {
+	if defaultK.client == nil {
+		panic("Default Komega's client is not set. Use SetClient to set it.")
+	}
+}
+
+// Get returns a function that fetches a resource and returns the occurring error.
+// It can be used with gomega.Eventually() like this
+//   deployment := appsv1.Deployment{ ... }
+//   gomega.Eventually(komega.Get(&deployment)).To(gomega.Succeed())
+// By calling the returned function directly it can also be used with gomega.Expect(komega.Get(...)()).To(...)
+func Get(obj client.Object) func() error {
+	checkDefaultClient()
+	return defaultK.Get(obj)
+}
+
+// List returns a function that lists resources and returns the occurring error.
+// It can be used with gomega.Eventually() like this
+//   deployments := v1.DeploymentList{ ... }
+//   gomega.Eventually(k.List(&deployments)).To(gomega.Succeed())
+// By calling the returned function directly it can also be used as gomega.Expect(k.List(...)()).To(...)
+func List(list client.ObjectList, opts ...client.ListOption) func() error {
+	checkDefaultClient()
+	return defaultK.List(list, opts...)
+}
+
+// Update returns a function that fetches a resource, applies the provided update function and then updates the resource.
+// It can be used with gomega.Eventually() like this:
+//   deployment := appsv1.Deployment{ ... }
+//   gomega.Eventually(k.Update(&deployment, func (o client.Object) {
+//     deployment.Spec.Replicas = 3
+//     return &deployment
+//   })).To(gomega.Scucceed())
+// By calling the returned function directly it can also be used as gomega.Expect(k.Update(...)()).To(...)
+func Update(obj client.Object, f func(), opts ...client.UpdateOption) func() error {
+	checkDefaultClient()
+	return defaultK.Update(obj, f, opts...)
+}
+
+// UpdateStatus returns a function that fetches a resource, applies the provided update function and then updates the resource's status.
+// It can be used with gomega.Eventually() like this:
+//   deployment := appsv1.Deployment{ ... }
+//   gomega.Eventually(k.Update(&deployment, func (o client.Object) {
+//     deployment.Status.AvailableReplicas = 1
+//     return &deployment
+//   })).To(gomega.Scucceed())
+// By calling the returned function directly it can also be used as gomega.Expect(k.UpdateStatus(...)()).To(...)
+func UpdateStatus(obj client.Object, f func(), opts ...client.UpdateOption) func() error {
+	checkDefaultClient()
+	return defaultK.UpdateStatus(obj, f, opts...)
+}
+
+// Object returns a function that fetches a resource and returns the object.
+// It can be used with gomega.Eventually() like this:
+//   deployment := appsv1.Deployment{ ... }
+//   gomega.Eventually(k.Object(&deployment)).To(HaveField("Spec.Replicas", gomega.Equal(pointer.Int32(3))))
+// By calling the returned function directly it can also be used as gomega.Expect(k.Object(...)()).To(...)
+func Object(obj client.Object) func() (client.Object, error) {
+	checkDefaultClient()
+	return defaultK.Object(obj)
+}
+
+// ObjectList returns a function that fetches a resource and returns the object.
+// It can be used with gomega.Eventually() like this:
+//   deployments := appsv1.DeploymentList{ ... }
+//   gomega.Eventually(k.ObjectList(&deployments)).To(HaveField("Items", HaveLen(1)))
+// By calling the returned function directly it can also be used as gomega.Expect(k.ObjectList(...)()).To(...)
+func ObjectList(list client.ObjectList, opts ...client.ListOption) func() (client.ObjectList, error) {
+	checkDefaultClient()
+	return defaultK.ObjectList(list, opts...)
+}
--- a/pkg/envtest/komega/interfaces.go
+++ b/pkg/envtest/komega/interfaces.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package komega
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Komega is a collection of utilites for writing tests involving a mocked
+// Kubernetes API.
+type Komega interface {
+	// Get returns a function that fetches a resource and returns the occurring error.
+	// It can be used with gomega.Eventually() like this
+	//   deployment := appsv1.Deployment{ ... }
+	//   gomega.Eventually(k.Get(&deployment)).To(gomega.Succeed())
+	// By calling the returned function directly it can also be used with gomega.Expect(k.Get(...)()).To(...)
+	Get(client.Object) func() error
+
+	// List returns a function that lists resources and returns the occurring error.
+	// It can be used with gomega.Eventually() like this
+	//   deployments := v1.DeploymentList{ ... }
+	//   gomega.Eventually(k.List(&deployments)).To(gomega.Succeed())
+	// By calling the returned function directly it can also be used as gomega.Expect(k.List(...)()).To(...)
+	List(client.ObjectList, ...client.ListOption) func() error
+
+	// Update returns a function that fetches a resource, applies the provided update function and then updates the resource.
+	// It can be used with gomega.Eventually() like this:
+	//   deployment := appsv1.Deployment{ ... }
+	//   gomega.Eventually(k.Update(&deployment, func (o client.Object) {
+	//     deployment.Spec.Replicas = 3
+	//     return &deployment
+	//   })).To(gomega.Scucceed())
+	// By calling the returned function directly it can also be used as gomega.Expect(k.Update(...)()).To(...)
+	Update(client.Object, func(), ...client.UpdateOption) func() error
+
+	// UpdateStatus returns a function that fetches a resource, applies the provided update function and then updates the resource's status.
+	// It can be used with gomega.Eventually() like this:
+	//   deployment := appsv1.Deployment{ ... }
+	//   gomega.Eventually(k.Update(&deployment, func (o client.Object) {
+	//     deployment.Status.AvailableReplicas = 1
+	//     return &deployment
+	//   })).To(gomega.Scucceed())
+	// By calling the returned function directly it can also be used as gomega.Expect(k.UpdateStatus(...)()).To(...)
+	UpdateStatus(client.Object, func(), ...client.UpdateOption) func() error
+
+	// Object returns a function that fetches a resource and returns the object.
+	// It can be used with gomega.Eventually() like this:
+	//   deployment := appsv1.Deployment{ ... }
+	//   gomega.Eventually(k.Object(&deployment)).To(HaveField("Spec.Replicas", gomega.Equal(pointer.Int32(3))))
+	// By calling the returned function directly it can also be used as gomega.Expect(k.Object(...)()).To(...)
+	Object(client.Object) func() (client.Object, error)
+
+	// ObjectList returns a function that fetches a resource and returns the object.
+	// It can be used with gomega.Eventually() like this:
+	//   deployments := appsv1.DeploymentList{ ... }
+	//   gomega.Eventually(k.ObjectList(&deployments)).To(HaveField("Items", HaveLen(1)))
+	// By calling the returned function directly it can also be used as gomega.Expect(k.ObjectList(...)()).To(...)
+	ObjectList(client.ObjectList, ...client.ListOption) func() (client.ObjectList, error)
+
+	// WithContext returns a copy that uses the given context.
+	WithContext(context.Context) Komega
+}
--- a/pkg/envtest/komega/komega.go
+++ b/pkg/envtest/komega/komega.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package komega
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// komega is a collection of utilites for writing tests involving a mocked
+// Kubernetes API.
+type komega struct {
+	ctx    context.Context
+	client client.Client
+}
+
+var _ Komega = &komega{}
+
+// New creates a new Komega instance with the given client.
+func New(c client.Client) Komega {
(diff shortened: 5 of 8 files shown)


pkg/envtest/komega/interfaces.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"time"
+
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Komega is the root interface that the Matcher implements.
+type Komega interface {
```
fabriziopandini, line 29: I'm struggling a little bit in fully understanding this interface, might be some example on how it should be used could help.

What confuses me is that the interface is a mix of a Client with CRUD operations, with a bit of client-go (UpdateStatus), something from gomega (Consistently Eventually), and the "With" methods.

Might be it could help to:
- Drop extras (seems out of scope)
- Drop shared context and pass context to each CRUD operation
- Drop shared Timeout/Pollinterval and allow users to pass to CRUD operations as options (otherwise rely on gomega defaults)

But given my poor understanding of the intent, I might be totally out of track...

fabriziopandini (commented):
@schrej thanks for reviving this effort.

While I get EqualObject and HaveField, and I would really like to have them merged, I'm still grokking the Komega client.
If others feel the same, might be breaking this in separated PR could help...

go.mod
```diff
@@ -4,11 +4,13 @@ go 1.17
 
 require (
 	github.com/evanphx/json-patch v4.12.0+incompatible
+	github.com/evanphx/json-patch/v5 v5.6.0
```
sbueringer, line 7: Not sure if we should use two versions of the same library. Maybe we should only use v4.12 for now (like k/k)

go.mod
```diff
@@ -4,11 +4,13 @@ go 1.17
...
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/go-logr/logr v1.2.0
 	github.com/go-logr/zapr v1.2.0
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.17.0
+	github.com/pkg/errors v0.9.1
```
sbueringer, line 13: Not sure how errors are usually wrapped in controller-runtime, but given that pkg/errors is only added with this PR it's probably not the same as in CAPI

Maybe we shouldn't introduce it (but let's see what CR maintainers think)

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,260 @@
...
+	client       client.Client
+	extras       []interface{}
+	timeout      time.Duration
+	pollInterval time.Duration
+}
+
+var _ Komega = &komega{}
+
+// NewKomega creates a new instance with a given client.
+func NewKomega(c client.Client) *komega {
```
sbueringer, line 43: Should this return the exposed interface and not the unexported struct? (same in the funcs below)

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,260 @@
...
+func (k komega) WithTimeout(timeout time.Duration) Komega {
+	k.timeout = timeout
+	return &k
+}
+
+// context returns the matcher context if one has been set.
+// Else it returns the context.TODO().
+func (k *komega) context() context.Context {
+	if k.ctx == nil {
+		return context.TODO()
```
sbueringer, line 94: Should this be `context.Background()`? 

```go
// TODO returns a non-nil, empty Context. Code should use context.TODO when
// it's unclear which Context to use or it is not yet available (because the
// surrounding function has not yet been extended to accept a Context
// parameter).
```

pkg/envtest/komega/interfaces.go
sbueringer: Some unit tests to see how komega is supposed to be used would be really great

JoelSpeed: It's been a while since I threw together the POC, and I'm still trying to understand the changes that have been made on top of it, but perhaps if I were to look at the naming of the functions, it would be better/easier to understand if it instead looks like
```
type Komega interface {
  ExpectCreate()
  ExpectGet()
  ExpectList()
  ExpectUpdate()
  ExpectUpdateStatus()
  ExpectDelete()

  ConsistentlyGet()
  EventuallyGet()

  ...
}
```

So then, for a quick example, if you are creating an object and you expect that the controller should set some status field to some value it has processed the object, you would write

```
myObj := &myapiv1.MyObject{}
Eventually(myObj).WithContext(ctx).WithTimeout(5*time.Second).Should(HaveField(".status.myField", Equal("expectedValue")))
```

The confusing/annoying part of this at the moment is the `WithContext` and `WithTimeout` modifiers that modify the assertion, I couldn't think of a better way to put those in at the time though

I think the original PR had smaller interfaces that better grouped the different functions which made this a little easier to understand, keen to know the reasoning for merging the interfaces into 1 here

JoelSpeed (commented):
In the interest of keeping the review surface small, I'm wondering if we should break this PR up a bit. The addition of the `EqualObject` matcher makes this PR quite a lot bigger and, as it is code that is already reviewed and accepted within CAPI, maybe should be reviewed separately if we intend to transfer it to this repo. That way we can focus the discussion on the design of the gomega extension and other new code within the PR? WDYT?

pkg/envtest/komega/transforms.go
```diff
@@ -25,20 +25,20 @@ import (
 	gtypes "github.com/onsi/gomega/types"
 )
 
-// WithField gets the value of the named field from the object.
+// HaveField gets the value of the named field from the object.
```
JoelSpeed, line 28: This was originally `WithField` because of the naming of the `WithTransform` in gomega, so this then matches that naming scheme. I don't have a strong opinion on which it should be called though, just wanted to open up for discussion in case anyone had some opinions about keeping this in line with gomega

pkg/envtest/komega/default.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Default is the Komega used by the package global functions.
+// NOTE: Replace it with a custom instance or use WithDefaultClient.
+//       Otherwise the package global functions will panic.
+var Default Komega = &komega{}
```
JoelSpeed, line 12: Is the intention here that you can import this and use it in the same way that you would with the default Gomega? Rather than having to have the `Komega` in scope, you just `komega.CreateObject().Should(Succeed())`?

Only nit is that these names don't match the interface at the moment, would it be good to be consistent?

schrej: The intention for the different names was to be a bit more clear that it's about Kubernetes objects when there is no `komega` in front of it. Also there would be collision with Eventually and Consistently.

Naming isn't great yet though.

pkg/envtest/komega/transforms.go
schrej: I changed it because it's a little nicer to read imo.
```
Expect(...).To(HaveField(..., Equal(...))
```
vs
```
Expect(...).To(WithField (..., Equal(...))
```

pkg/envtest/komega/interfaces.go
schrej: I moved the `With..` methods to the main interface so they can return a `Komega` again. And since I didn't really see the reason for separation I just merged all into one.

I'm also not sure if some of these functions/methods even make sense. `Create` is just a wrapper around `client.Create` that strips away a few options for example. Maybe we should try to only provide helper functions that can then be fed into Gomega.

So instead of `Get().Should(Succeed())` you would write `Eventually(Get()).Should(Succeed())`. That way that `Get()` function can take the regular arguments related to fetching objects, while `Eventually()` is just the regular Gomega function that takes it's arguments.
Instead of returning a Gomega call we would just return the function instead, which can then be fed into a Gomega call.

- `Eventually()` should be replaced with `Eventually(Object(...))` and `Eventually(List())` (or maybe `Resource()`?) - where Object and List are part of komega
- Same applies to `Consistently()`
- `Get()`, `List()`, `Update()`, `UpdateStatus()` stay, but are used as `Expect(Get())` and so on
- We can probably ditch `Create()`
- The difference between `Get()` and `Object()` is that the former returns an error and the latter returns an object.

schrej: > The original reason for separation was to make it obvious which of the With... methods applied to which of the assertions. For example the timeout only applies to the async operations, so if we have a separate async interface we can ensure that a user can't program Get().WithTimeout() which would be invalid.

Hmm. But `Get()` would return a `GomegaAssertion` anyway, so that wouldn't have worked. I guess the intention then was to only allow `ak.WithTimeout().Create()`, but not `ak.WithTimeout().Get()`.

> Likewise, with the async assertions, you would be able to set the timeout globally, and if you wanted to, override it for a single assertion

It's possible to set default timeouts on Gomega level, imo that should be sufficient: 
https://pkg.go.dev/github.com/onsi/gomega@v1.17.0#SetDefaultConsistentlyDuration

> So your proposal then is that, for example, `Get` would be implemented as:

Exactly. Sorry for not explaining that properly.

> I guess the only things that would make sense in these cases would be ctx and client, in which case, it's not so helpful maybe?

I think having at least the client, which is often just a single instance within a package, would be helpful, and so would be to default to a background context. If you need to use a specific context or client you could create a new instance from the global one using the `With...` functions, e.g.
```go
# uses default client set with komega.DefaultClient(c), e.g. in suite_test.go, and context.Background()
Expect(Get()).Should(Succeed())

# to use specific client and context
k := komega.WithClient(fakeClient).WithContext(ctx)
Expect(k.Get()).Should(Succeed())
```
I'll change the PR to reflect these changes and I'll also add some tests if I manage to do so :smile:

JoelSpeed: > Hmm. But Get() would return a GomegaAssertion anyway, so that wouldn't have worked. I guess the intention then was to only allow ak.WithTimeout().Create(), but not ak.WithTimeout().Get().

I think you're right, I did this a while back, I'll be honest, I can't remember 😅 

> It's possible to set default timeouts on Gomega level, imo that should be sufficient:
https://pkg.go.dev/github.com/onsi/gomega@v1.17.0#SetDefaultConsistentlyDuration

Ack, agreed!

> I'll change the PR to reflect these changes and I'll also add some tests if I manage to do so 😄

Cool, I'm on board with what's being discussed here, sounds like it will be a lot simpler, will drop a review in once you've pushed the new changes

pkg/envtest/komega/transforms.go
schrej: Turns out this already exists: https://pkg.go.dev/github.com/onsi/gomega@v1.17.0#HaveField
I'll remove it.

JoelSpeed: Oh sweet, they literally just introduced that!

sbueringer: Nice catch!

schrej: Unfortunately that one doesn't support pointers, which are used within the API structs quite a bit. With the new `Object()` function it's then quite important, so I'll leave it in for now. If we decide on using this API we can open an Issue/PR to Gomega and see if we can get pointer support.

> Nice catch!

Heh, I tried to `import . "github.com/onsi/gomega"` and it complained that `HaveField` already exists :smile:

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,133 @@
...
+// consistentlyclient.Object gets an individual object from the API server.
+func (k *komega) Object(obj client.Object) func() client.Object {
+	key := types.NamespacedName{
+		Name:      obj.GetName(),
+		Namespace: obj.GetNamespace(),
+	}
+	return func() client.Object {
+		err := k.client.Get(k.context(), key, obj)
+		if err != nil {
+			panic(err)
```
JoelSpeed, line 118: Panicking here will cause the test to end on any error, which isn't what we want for an async assertion.
What you want to do instead is make this function return two values, the object and the error.

When Gomega sees two values, it will error if the second value is not nil/zero valued, so you can get away with returning errors and, if using this in an async assertion, it will wait until the error goes away. That means you can `Eventually(Object(&Doesn'tExistAtTheStart{})).ShouldNot(BeEmpty())`, or create the object and wait for it to be populated in the client cache if that makes sense.

So I would make this function:

```
func (k *komega) Object(obj client.Object) func() (client.Object, error) {
	key := types.NamespacedName{
		Name:      obj.GetName(),
		Namespace: obj.GetNamespace(),
	}
	return func() (client.Object, error) {
		err := k.client.Get(k.context(), key, obj)
		if err != nil {
			return nil, err
		}
		return obj, nil
	}
}
```
Same applies for the list below as well

pkg/envtest/komega/transforms.go
JoelSpeed: It would be good if we can improve the gomega version to support pointers, otherwise I'd suggest renaming this to something that doesn't conflict with the gomega function

pkg/envtest/komega/default.go
schrej: We can, and that even reduces friction with importing both gomega and komega as `.`.

schrej:
is `pkg/envtest/komega` a good location for that package? Or should we maybe move it to `pkg/client/komega`, since it's heavily relying on that one, or just to `pkg/komega` directly?

JoelSpeed (commented):
Apart from the `HaveField` naming conflict, I think this is looking pretty good

pkg/envtest/komega/transforms.go
JoelSpeed: I created an issue on Gomega to suggest adding pointer support and will work on fixing there if we can get agreement from the maintainer https://github.com/onsi/gomega/issues/494

schrej: I was about to do the same, thanks!

schrej:
> is `pkg/envtest/komega` a good location for that package? Or should we maybe move it to `pkg/client/komega`, since it's heavily relying on that one, or just to `pkg/komega` directly?

@fabriziopandini @vincepri do you have an opinion on that?

JoelSpeed:
/lgtm

I don't have a strong opinion on the naming, but the Komega is looking really clean now and with the gomega HaveField should be really useful for trimming the fat in controller tests

alvaroaleman:
Hey @schrej , thanks for your work here. Would you or anyone of the ppl that commented in here be open to maintain this going forward? If yes, please add an OWNERS file with them and we can merge this.

The reason for the ask is that I personally don't use gomega so can't really review this very well nor do I want to be responsible for something I am definitely not going to use. Simultaneous I don't necessary want that to block things that might be useful for others. Having sub owners might be a nice compromise.

schrej:
I would be open to maintain this. I'd assume @vincepri might also be fine with maintaining this, since he's already controller-runtime maintainer and gomega is used heavily within cluster-api.

@alvaroaleman do you have any opinion on [module location](https://github.com/kubernetes-sigs/controller-runtime/pull/1767#issuecomment-1023070992)?

JoelSpeed:
I'm also happy to help maintain here if required. I use gomega frequently and intend to use this package in a project I'm starting soon, so will have a vested interest in it

sbueringer:
I can also help maintain it as I think it will be heavily used in CAPI.

alvaroaleman:
I don't have a strong opinion on location, maybe something like pkg/testutils/komega might be easier to discover, not sure though,

Can you please add an owners file to the package?

schrej:
I will once I'm back from vacation. But I also thought about waiting for go 1.18 to add generics to the Update function since it's a new API anyway 😄

JoelSpeed:
> But I also thought about waiting for go 1.18 to add generics to the Update function since it's a new API anyway 

Hmm interesting idea, I have no idea what that would look like, have you done a POC on one of the 1.18 betas? What's the advantage of adding generics here?

pkg/envtest/komega/default.go
```diff
@@ -0,0 +1,85 @@
...
+var defaultK = &komega{}
+
+// SetDefaultClient sets the client used by the package global functions.
+func SetDefaultClient(c client.Client) {
+	defaultK = &komega{client: c}
+}
+
+func checkDefaultClient() {
+	if defaultK.client == nil {
+		panic("Default Komega's client is not set. Use WithDefaultClient to set it.")
```
sbueringer, line 17:
```suggestion
		panic("Default Komega's client is not set. Use SetDefaultClient to set it.")
```
nit. Should this be `SetDefaultClient`?

pkg/envtest/komega/default.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// defaultK is the Komega used by the package global functions.
+var defaultK = &komega{}
+
+// SetDefaultClient sets the client used by the package global functions.
+func SetDefaultClient(c client.Client) {
```
sbueringer, line 11: I wonder if this should be just SetClient()

I'm not sure in which way it is a default client.

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// WithContext returns a copy that uses the given context.
+func (k komega) WithContext(ctx context.Context) Komega {
+	k.ctx = ctx
+	return &k
+}
+
+// WithClient returns a copy that uses the given client.
+func (k komega) WithClient(c client.Client) Komega {
```
sbueringer, line 49: I wonder when this function would be used. 
If my understanding is correct the client already has to be passed in with `New`

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,128 @@
...
+	k.client = c
+	return &k
+}
+
+// context returns the matcher context if one has been set or context.Background() otherwise.
+func (k *komega) context() context.Context {
+	if k.ctx == nil {
+		return context.Background()
+	}
+	return k.ctx
```
sbueringer, line 59: WDYT about just setting context.Background() in the New func?

 Then we should get the same behavior when we just use k.ctx in the funcs below.

pkg/envtest/komega/komega.go
```diff
@@ -0,0 +1,128 @@
...
+		Name:      obj.GetName(),
+		Namespace: obj.GetNamespace(),
+	}
+	return func() error {
+		return k.client.Get(k.context(), key, obj)
+	}
+}
+
+// List returns a function that lists resources and returns the occurring error.
+func (k *komega) List(obj client.ObjectList, opts ...client.ListOption) func() error {
```
sbueringer, line 74:
```suggestion
func (k *komega) List(list client.ObjectList, opts ...client.ListOption) func() error {
```
nit. There are probably a few other similar cases.

pkg/envtest/komega/interfaces.go
```diff
@@ -0,0 +1,82 @@
...
+	ObjectList(client.ObjectList, ...client.ListOption) func() (client.ObjectList, error)
+
+	// WithClient returns a copy that uses the given client.
+	WithClient(client.Client) Komega
+	// WithContext returns a copy that uses the given context.
+	WithContext(context.Context) Komega
+}
+
+// UpdateFunc receives an object and expects a modified version of it to be returned.
+type UpdateFunc func(client.Object) client.Object
```
sbueringer, line 82: Afaik when modifying this object the "input parameter" will be modified automatically.

Is the expectation that the input client.Object is deep copied before modification?

Or maybe we don't need a return value?

I think it would be good to be clear about the expectation.

sbueringer (commented):
Looks good to me overall. 

It would be good to know if the PR is essentially on hold until Go 1.18 or not.

My current take is that we should get go 1.18 in a few weeks once Kubernetes picked it up and CR picked up the Kubernetes release.

Downside is that all users of this package then have to use go 1.18 too

go.mod
```diff
@@ -8,12 +8,12 @@ require (
 	github.com/go-logr/logr v1.2.0
 	github.com/go-logr/zapr v1.2.0
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.17.0
+	github.com/onsi/gomega v1.18.1
```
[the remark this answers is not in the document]
schrej: Doesn't contain any significant changes, but is required for the tests implemented in this PR: https://github.com/onsi/gomega/releases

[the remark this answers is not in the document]
sbueringer: Just wanted to ask if it should be separate. I don't know how CR does it, in core CAPI we usually prefer separate PRs to bump dependencies

pkg/envtest/komega/interfaces.go
sbueringer: Okay got it. So it sounds like it might be worth to see an alternative implementation with Go generics before we move forward?

schrej:
After further investigation I don't think it makes sense to add generics at this time. As Stefan pointed out it will obviously require go 1.18, and force all depending projects to use 1.18 as well. Since methods don't support type parameters (only structs), it's also hard to implement it in a beneficial way.
It's only feasible on the [package functions](https://github.com/schrej/controller-runtime/blob/feature/generic-komega/pkg/envtest/komega/default.go) that use the default instance.
But [using it](https://github.com/schrej/controller-runtime/blob/feature/generic-komega/pkg/envtest/komega/default_test.go) is pretty nice.

pkg/envtest/komega/interfaces.go
sbueringer: Okay saw your answer below

pkg/envtest/komega/komega.go
schrej: Leftover from when timeouts were part of that struct, to allow creating a copy with a new client, but preserving those settings.

schrej:
Ok. All comments should be addressed now. No generics.
I've added an OWNERS file containing @sbueringer, @JoelSpeed and myself in addition to the regular reviewers and approvers. Is that ok @alvaroaleman?

schrej:
not related to the changes. /retest

schrej:
Not related to the changes.
/retest

alvaroaleman:
Leaving lgtm for one of the new approvers for this

pkg/envtest/komega/OWNERS
```diff
@@ -0,0 +1,14 @@
...
+  - schrej
+  - JoelSpeed
+  - sbueringer
+reviewers:
+  - controller-runtime-admins
+  - controller-runtime-reviewers
+  - controller-runtime-approvers
+  - schrej
+  - JoelSpeed
+  - sbueringer
```
JoelSpeed, line 14: Nit: Should make sure files have an empty new line at the end

pkg/envtest/komega/default.go
```diff
@@ -0,0 +1,87 @@
+package komega
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// defaultK is the Komega used by the package global functions.
+var defaultK = &komega{ctx: context.Background()}
```
JoelSpeed, line 10: No option to set the context on the default komega?

JoelSpeed (commented):
Couple of nits otherwise LGTM

pkg/envtest/komega/default.go
schrej: Not sure how useful that is, but I've added it for completeness.

JoelSpeed (approved):
/lgtm

Thanks for all your work on this @schrej

Issue #1545, opened 2021-06-01
Ginkgo v2 migration


DirectXMan12:
Ginkgo has started warning about the Ginkgo v2 upgrade, which deprecates the `done` argument to tests (the suggestion is to use `Eventually` and such).  We should look at migrating our tests over.

/help

ystkfujii:
Shoud I change the ginkyo v1.16.0 to v2.0.0-rc2 ?

There are some changes other than AAA, should I modify them?
For example, the argument of the BeforeSuite function.

PR #1792, opened 2022-02-05
⚠️  Update ginkgo@v2.1.1 and gomega@v1.18.0

acumino:
Migrate ginkgo to v2. 

Fixes: #1545

**Important Note:**
-> Per node `timeout` has been removed to prevent the wrongly marking a test as failure which is caused by some other test. Read more about it [here](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing)
-> Custom reporters have been deprecated in ginkgo v2 and will be removed in future releases.


hack/test-all.sh
```diff
@@ -20,7 +20,7 @@ source $(dirname ${BASH_SOURCE})/common.sh
 
 header_text "running go test"
 
-go test -race ${P_FLAG} ${MOD_OPT} ./...
+go test ${P_FLAG} ${MOD_OPT} ./...
```
alvaroaleman, line 23: Why?

pkg/builder/builder_suite_test.go
```diff
@@ -36,10 +37,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+var suiteName = "application Suite"
```
alvaroaleman, line 40:
```suggestion
const suiteName = "application Suite"
```

pkg/builder/builder_suite_test.go
```diff
@@ -63,7 +65,7 @@ var _ = BeforeSuite(func() {
 
 	webhook.DefaultPort, _, err = addr.Suggest("")
 	Expect(err).NotTo(HaveOccurred())
-}, 60)
```
alvaroaleman, line 66: Why is the custom timeout gone?

alvaroaleman (commented):
It seems this doesn't build?

pkg/builder/builder_suite_test.go
acumino: This has been removed in v2.

acumino: I checked, but I guess there is no support. Please check https://onsi.github.io/ginkgo/MIGRATING_TO_V2#major-additions-and-improvements.

hack/test-all.sh
acumino: Since custom reporters have been deprecated, the deprecation code behavior is different. Until we move to a new change we can remove this flag.

alvaroaleman: No, we can not remove the `race` flag, we want the racedector to be enabled during tests. And I don't understand how this relates to reportes?

pkg/builder/builder_suite_test.go
alvaroaleman: Please describe in your PR how the timeout behavior for these kind of tests was before and how it is now. Just removing things without a proper explanation is not the way to go.

hack/test-all.sh
acumino: Can you check the log of `pull-controller-runtime-test-master`? There is data race. This seems to happen because we now use `ReportViaDeprecatedReporter` and its implementation is not the same as `RunSpecsWithDefaultAndCustomReporters`. This is for migration purposes. Once we remove Custom Reporter we can introduce the race flag again.

acumino: @alvaroaleman PTAL.

alvaroaleman: Well, then we need to fix it, not disable the race detector

acumino: The steps will be to remove the custom reporter in future updates and then only we can enable race. For now, till the time we use `ReportViaDeprecatedReporter `, we should disable it. Or should we directly remove the custom reporter? WDYT?
It was mentioned in the gingkov2 migration guide to use `ReportViaDeprecatedReporter`. Please check https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-custom-reporters

alvaroaleman: We can not disable the race detector because we need it to find data races.

I don't have the cycles to look into why ginkgo v2 with the `ReportViaDeprecatedReporter` causes a data race, figuring this out is part the v2 migration work.

If you want to avoid using `ReportViaDeprecatedReporter`, please find out why this reporter exists (I think because the junit reporting doesn't exist the process with a RC != 0 if there was an error) and find an alternate approach to the problem that reporter is solving.

pkg/builder/builder_suite_test.go
acumino: I have updated the PR and resolved your concern, and also opened an issue on gingko [here](https://github.com/onsi/ginkgo/issues/940). PTAL

hack/test-all.sh
acumino: I have updated the PR and resolved your concern, and also opened an issue on gingko [here](https://github.com/onsi/ginkgo/issues/940). PTAL

acumino: friendly ping @alvaroaleman

pkg/builder/builder_suite_test.go
acumino: Previously custom timeouts were there to run tests asynchronously, now ginkgo doesn't support async testing. I have updated it in the PR description.

pkg/client/client_test.go
```diff
@@ -431,7 +429,7 @@ var _ = Describe("Client", func() {
 				Expect(err).To(HaveOccurred())
 				// TODO(seans): Add test to validate the returned error. Problems currently with
 				// different returned error locally versus travis.
-			}, serverSideTimeoutSeconds)
```
camilamacedo86, line 434: That does not seem the right way to move forward with and ensure the same behaviour.
See: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/go-logr/logr v1.2.0
 	github.com/go-logr/zapr v1.2.0
-	github.com/onsi/ginkgo v1.16.5
+	github.com/onsi/ginkgo/v2 v2.1.1
```
camilamacedo86, line 10: Is not it a breaking change?
Will not this change affects who is using C+R and testing their projects with ENV TEST for example?
Will not be required to migrate the code to do the same with v2 as well?

pkg/client/client_test.go
acumino: @camilamacedo86 can you check?

go.mod
schrej: No. The package is only used in tests. All changed files are also `_test.go` files and those can't be imported by other packages/programs. So there is no way this would affect consumers of controller-runtime.

The dependency in `go.mod` also doesn't seem to have an effect, at least cluster-api tests are still executing fine when using a `replace` directive.

camilamacedo86: @schrej,

Controller runtime provides EnvTest, which allows us to write the tests.
It seems that it is a breaking change ( changes the behaviour for who is importing kubebuilder and writing tests with. Am I right? if yes, I think we need to clarify that in the title as ⚠️ at least.  )

schrej: Ah, you are right. There actually is one package that does import ginkgo outside of tests: `pkg/envtest/printer`
But it seems like the `v2` suffix of the new version allows to import both versions at the same time. At least the compiler didn't complain when I used this PR with cluster-api, and that one is still on v1.

camilamacedo86: Note that the ginkgo made a MAJOR bump. That is only required when we have breaking changes ( unless you have a version that is 0.MINOR releases which means initial development )

The MAJOR bump occurred because they removed some deprecated options.

Therefore, consumers of controller runtime might be impacted because they will need to update and change their tests. So, we need to update the title of this PR ( which is used for generating the release notes ) with ⚠️ instead of use ✨. Also, it would be nice to have in the description of the PR the link for the release notes of the dep as the guidance to upgrade from v1 to v2.

schrej (commented):
From what I can see, the removal of those two helpers for gingko v1 in `pkg/envtest/printer` makes this a breaking change.

Yes, Ginkgo does a major bump, but it is possible to have both v1 and v2 as a dependency in the same project. Bumping Ginkgo alone only influences the tests of controller-runtime, but not the tests of dependant projects. Cluster API is still using Ginkgo v1 and works fine when running its test with this PR.
Since Cluster API doesn't use the two removed helpers, it's not affected by this PR at all.

pkg/envtest/printer/ginkgo.go
```diff
@@ -1,53 +0,0 @@
...
-	"github.com/onsi/ginkgo/config"
-	"github.com/onsi/ginkgo/types"
-)
-
-var _ ginkgo.Reporter = NewlineReporter{}
-
-// NewlineReporter is Reporter that Prints a newline after the default Reporter output so that the results
-// are correctly parsed by test automation.
-// See issue https://github.com/jstemmer/go-junit-report/issues/31
-type NewlineReporter struct{}
```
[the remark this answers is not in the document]
camilamacedo86: See that you are changing the test_.go files. Those with controller runtime projects will also need to change their _test.go files as well, right? They might use methods that will be affected or not. For example, the default scaffolds done by Kubebuilder will be affected if we need to bump v2.

So, the correct mojo here would be ⚠️.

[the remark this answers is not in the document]
schrej: They will need to change their tests if they are using either `printer.NewLineReporter()` or `printer.NewProwReporter()`.

Any change that _we_ do in _our_ `_test.go` files does _not_ affect any users of controller-runtime. Code in `_test.go` files can't be imported, so it also can't affect their implementation. That's why I initially thought this is not a breaking change.

I would propose to move the removal of these two reporters to a separate PR, so this one does not have to be breaking. And then maybe postpone the removal of those helpers a bit to allow more time for migration to v2.

pkg/builder/builder_suite_test.go
matteoolivi: @alvaroaleman in ginkgo v1 `BeforeSuite` honored the timeout if and only if the function to `BeforeSuite` accepted a `done` channel - which is not the case with the code before this change. So even before this issue the timeout was ignored, so removing it here is fine.

pkg/client/client_test.go
matteoolivi: @camilamacedo86 ginkgo honours timeouts like the one removed here if and only if the function passed to `It` accepts a `done` channel, otherwise it ignores it. 

Before this change the function passed to the `It` the removed timeout is for didn't have a `done` channel argument - so the timeout was completely ignored.

So removing the timeout here doesn't alter behavior.

matteoolivi:
Why was this closed? Can someone provide an explanation/a status update?

alvaroaleman:
@matteoolivi presumably bceause @acumino didn't have the cycles to finish it

matteoolivi:
@alvaroaleman . Oh. I can definitely relate to that.

Issue #1799, opened 2022-02-10
Tests crash if os.UserCacheDir() returns an unusable directory


skitt:
Since commit 58c17f696263 (“addr.Suggest should lock a file instead of memory”), `pkg/internal/testing/addr/manager.go`’s `init()` function tries to create a directory using either `os.UserCacheDir()` or `os.TempDir()`, whichever succeeds first. In many build environments, `$HOME` is non-empty but points to an unusable directory; in such cases, `os.UserCacheDir()` returns an unusable directory, which causes `init()` to panic.

It could instead try `os.TempDir()`, which is usable in many more scenarios than `os.UserCacheDir()`.

PR #1800, opened 2022-02-10
Try a temporary directory if the user cache fails

skitt:
Since commit 58c17f6 ("addr.Suggest should lock a file instead of
memory"), pkg/internal/testing/addr/manager.go’s init() function tries
to create a directory using either os.UserCacheDir() or os.TempDir(),
whichever succeeds first. In many build environments, $HOME is
non-empty but points to an unusable directory; in such cases,
os.UserCacheDir() returns an unusable directory, which causes init()
to panic.

This changes init() to first try os.UserCacheDir(), including creating
the desired directory; if that fails, the whole operation is tried
again with os.TempDir().

Fixes: #1799
Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/pkg/internal/testing/addr/manager.go
+++ b/pkg/internal/testing/addr/manager.go
@@ -43,11 +43,17 @@ var (
 
 func init() {
 	baseDir, err := os.UserCacheDir()
-	if err != nil {
-		baseDir = os.TempDir()
+	if err == nil {
+		cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
+		err = os.MkdirAll(cacheDir, 0o750)
 	}
-	cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
-	if err := os.MkdirAll(cacheDir, 0750); err != nil {
+	if err != nil {
+		// Either we didn't get a cache directory, or we can't use it
+		baseDir = os.TempDir()
+		cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
+		err = os.MkdirAll(cacheDir, 0o750)
+	}
+	if err != nil {
 		panic(err)
 	}
 }


pkg/internal/testing/addr/manager.go
```diff
@@ -42,16 +42,30 @@ var (
 )
 
 func init() {
-	baseDir, err := os.UserCacheDir()
+	var err error
+	cacheDir, err = tryUserCacheDir()
 	if err != nil {
-		baseDir = os.TempDir()
+		// Either we didn't get a cache directory, or we can't use it
+		cacheDir, err = tryCacheInDir(os.TempDir())
```
alvaroaleman, line 49: I would actually vote for trying os.TempDir first, because a) it is more likely to succeed and more importantly b) it gets cleared upon reboot, which is not the case for UserCacheDir in most envs I think?

skitt: Yup, `UserCacheDir` doesn’t get cleaned up. I wanted to mostly preserve existing behaviour in this PR; were I to change that, I’d actually argue that `UserCacheDir` isn’t really appropriate. It is [defined as](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

> There is a single base directory relative to which user-specific non-essential (cached) data should be written.

(Plus there’s the general idea in XDG that it’s for “real” users in a desktop environment...)

For lock-style files as used here, it seems to me that the appropriate directories are either `/tmp` (as used historically, *e.g.* for X11 locks or SSH agent sockets) or `/run` (but Go doesn’t wrap that).

alvaroaleman: Hm yeah, I am really kinda surprised at seeing this here and wondering why that was done in the first place. Maybe os.Tempdir on MacOS is problematic? I don't have a Mac so I can't test that (quick google didn't show something, but that doesn't mean there is no issue). Do you remember @vincepri ?

Other than that this looks fine, except that the commit and pr title should be prefixed with `:bug:` as documented in the PR template, could you please add that? We also have a check that should verify that, but it appears it is not working.

vincepri: Tempdir on Mac isn't really reliable so we switched to the CacheDir instead

pkg/internal/testing/addr/manager.go
```diff
@@ -42,16 +42,30 @@ var (
...
 	}
-	cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
-	if err := os.MkdirAll(cacheDir, 0750); err != nil {
+	if err != nil {
 		panic(err)
```
vincepri, line 52: Instead of moving everything to two functions, can we just add the fallback here?

skitt: We can, it ends up looking like

```go
func init() {
	baseDir, err := os.UserCacheDir()
	if err == nil {
		cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
		err = os.MkdirAll(cacheDir, 0o750)
	}
	if err != nil {
		// Either we didn't get a cache directory, or we can't use it
		baseDir = os.TempDir()
		cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
		err = os.MkdirAll(cacheDir, 0o750)
	}
	if err != nil {
		panic(err)
	}
}
```

The first `err` check isn’t idiomatic and might trip up people making changes here in future, which is why I split everything up, but if that looks OK to you I’ll update the PR accordingly.

skitt: @vincepri which do you prefer? The version in the PR, or the version in the comment above?

vincepri: The one above works

skitt: Done, thanks!

Issue #1798, opened 2022-02-09
Leader election events are recorded with the wrong config


JustinKuli:
I'm working with a controller that needs to synchronize resources between two different Kubernetes clusters. The controller runs in what we'll call the "spoke" cluster, but we want it to watch resources in another cluster, which we call the "hub." To handle the leader election inside the spoke cluster, we make the manager like this:

```
options := manager.Options{
    ...
    LeaderElectionConfig: spokeCfg,
}

mgr, err := ctrl.NewManager(hubCfg, options)
```

This seems to work, and the configmap/lease that the leader election uses are on the spoke cluster. However, the leader election Events are going to the hub cluster.

We noticed this because we didn't give the kubeconfig on the hub cluster permissions to create events, so we get an error log like this (some details elided) :
```
E0208 21:45:38.523903       1 event.go:264] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ..., Reason:"LeaderElection", Message:"... became leader", ...}': 'events is forbidden: User "..." cannot create resource "events" in API group "" in the namespace "..."' (will not retry!)
```

I think the issue is that the `recorderProvider` used here doesn't use the `LeaderElectionConfig`: https://github.com/kubernetes-sigs/controller-runtime/blob/273e608d8a6cd7f34b6b56a6bd4626878319aed7/pkg/manager/manager.go#L332-L342

PR #1803, opened 2022-02-14
:bug: Use leader config for election events

JustinKuli:
Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/1798

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

This creates a new Recorder for the leader election events (using the leader config, if provided).

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -335,11 +335,21 @@ func New(config *rest.Config, options Options) (Manager, error) {
 	}
 
 	// Create the resource lock to enable leader election)
-	leaderConfig := options.LeaderElectionConfig
-	if leaderConfig == nil {
+	var leaderConfig *rest.Config
+	var leaderRecorderProvider *intrec.Provider
+
+	if options.LeaderElectionConfig == nil {
 		leaderConfig = rest.CopyConfig(config)
+		leaderRecorderProvider = recorderProvider
+	} else {
+		leaderConfig = rest.CopyConfig(options.LeaderElectionConfig)
+		leaderRecorderProvider, err = options.newRecorderProvider(leaderConfig, cluster.GetScheme(), options.Logger.WithName("events"), options.makeBroadcaster)
+		if err != nil {
+			return nil, err
+		}
 	}
-	resourceLock, err := options.newResourceLock(leaderConfig, recorderProvider, leaderelection.Options{
+
+	resourceLock, err := options.newResourceLock(leaderConfig, leaderRecorderProvider, leaderelection.Options{
 		LeaderElection:             options.LeaderElection,
 		LeaderElectionResourceLock: options.LeaderElectionResourceLock,
 		LeaderElectionID:           options.LeaderElectionID,


pkg/manager/manager.go
```diff
@@ -339,7 +339,13 @@ func New(config *rest.Config, options Options) (Manager, error) {
 	if leaderConfig == nil {
 		leaderConfig = rest.CopyConfig(config)
 	}
-	resourceLock, err := options.newResourceLock(leaderConfig, recorderProvider, leaderelection.Options{
+
+	leaderRecorderProvider, err := options.newRecorderProvider(leaderConfig, cluster.GetScheme(), options.Logger.WithName("events"), options.makeBroadcaster)
```
FillZpp, line 343: Can we new this `leaderRecorderProvider`  only if `options.LeaderElectionConfig` is not nil? Otherwise it should be redundant to `recorderProvider`.

pkg/manager/manager.go
```diff
@@ -335,11 +335,21 @@ func New(config *rest.Config, options Options) (Manager, error) {
...
-	leaderConfig := options.LeaderElectionConfig
-	if leaderConfig == nil {
+	var leaderConfig *rest.Config
+	var leaderRecorderProvider *intrec.Provider
+
+	if options.LeaderElectionConfig == nil {
 		leaderConfig = rest.CopyConfig(config)
+		leaderRecorderProvider = recorderProvider
+	} else {
+		leaderConfig = options.LeaderElectionConfig
```
alvaroaleman, line 345: Please also copy the config here, the leader election manipulates it down the line and we don't want that to get propagated anywhere else

JustinKuli: I'm not sure what you mean. If given, that config would only be used for the leader election, so I don't see where manipulating it would be an issue.

I've updated the PR to copy it, since that's also not a problem. I'm just curious if you can provide some more information to help me learn.

pkg/manager/manager.go
```diff
@@ -335,11 +335,21 @@ func New(config *rest.Config, options Options) (Manager, error) {
...
-	leaderConfig := options.LeaderElectionConfig
-	if leaderConfig == nil {
+	var leaderConfig *rest.Config
+	var leaderRecorderProvider *intrec.Provider
+
+	if options.LeaderElectionConfig == nil {
 		leaderConfig = rest.CopyConfig(config)
+		leaderRecorderProvider = recorderProvider
+	} else {
+		leaderConfig = rest.CopyConfig(options.LeaderElectionConfig)
```
alvaroaleman, line 345: The config is a pointer to a *rest.Config. We set some fields down the line for User-Agent and similiar which means we manipulate the value behind the pointer, so it appears in the original kubeconfig, i.E.:

```
leCfg := GetConfig()
mgr.New(cfg, manager.Options{LeaderElectionConfig: leCfg}
// If we don't do the CopyConfig, the User-Agent in the leCfg got changed
```

alvaroaleman (approved):
thanks!

PR #1805, opened 2022-02-15
✨ Add TransformFuncByObject Option for Informer Cache

alexzielenski:
This PR is a complement to a recent addition to client-go https://github.com/kubernetes/kubernetes/pull/107507.

Adds two new `ForOption`s, `OwnsOption`s, and `WatchesOption`s to be used with the existing `OnlyMetadata` called `WithoutAnnotations` and `WithoutManagedFields`.

### Motivation
We have teams which make extensive use of metadata-only watches (tens of thousands of objects), which have seen sizable reductions in RAM usage when these fields are removed. This new options makes it easy and ergonomic to opt into stripping `managedFields` and `annotations` metadata before they are persisted in the cache, to avoid wasting resources in controllers which do not use those fields.

### Implementation 
Behind the scenes uses `SetTransform` on `SharedIndexInformer`, and uses a transformer which respects the builders' elections for each GVK.

### Example Usage

```go
bldr := ControllerManagedBy(mgr).
        For(&appsv1.Deployment{}, OnlyMetadata, WithoutAnnotations, WithoutManagedFields)
```

The above line of code creates a controller whose cache is metadata-only and the annotations/managed fields have been stripped from the metadata.

### Comments
I'm open to other ways to expose this new `SetTransform` API, but would prefer to have it specified near the builder since `OnlyMetadata` is specified there as well. It seems natural to keep the different options for cache projections together.

Additionally included is a go.mod update to a version of client-go which includes `SetTransform`. I am unsure if this PR is the best place to put it, or if another PR should be created updating all components in one shot.

It should be noted that in my implementation `WithoutAnnotations` and `WithoutManagedFields` only apply to the `PartialObjectMeta` cache, and not unstructured/structured caches. This was intentional, since I did not see a reason to use it against other cache types. But this point may be worth discussing.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -128,6 +128,18 @@ type Options struct {
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
+
+	// TransformFuncByObject is a map from GVKs to transformer functions which
+	// get applied when objects of the transformation are about to be committed
+	// to cache.
+	//
+	// This function is called both for new objects to enter the cache,
+	// 	and for updated objects.
+	TransformFuncByObject TransformFuncByObject
+
+	// DefaultTransform is the transform used for all GVKs which do
+	// not have an explicit transform func set in TransformFuncByObject
+	DefaultTransform toolscache.TransformFunc
 }
 
 var defaultResyncTime = 10 * time.Hour
@@ -146,7 +158,12 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace, selectorsByGVK, disableDeepCopyByGVK)
+	transformByGVK, err := convertToTransformByKindAndGVK(opts.TransformFuncByObject, opts.DefaultTransform, opts.Scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	im := internal.NewInformersMap(config, opts.Scheme, opts.Mapper, *opts.Resync, opts.Namespace, selectorsByGVK, disableDeepCopyByGVK, transformByGVK)
 	return &informerCache{InformersMap: im}, nil
 }
 
@@ -240,4 +257,19 @@ func convertToDisableDeepCopyByGVK(disableDeepCopyByObject DisableDeepCopyByObje
 		}
 	}
 	return disableDeepCopyByGVK, nil
+}
+
+// TransformFuncByObject associate a client.Object's GVK to a transformer function
+// to be applied when storing the object into the cache.
+type TransformFuncByObject map[client.Object]toolscache.TransformFunc
+
+func convertToTransformByKindAndGVK(t TransformFuncByObject, defaultTransform toolscache.TransformFunc, scheme *runtime.Scheme) (internal.TransformFuncByObject, error) {
+	result := internal.NewTransformFuncByObject()
+	for obj, transformation := range t {
+		if err := result.Set(obj, scheme, transformation); err != nil {
+			return nil, err
+		}
+	}
+	result.SetDefault(defaultTransform)
+	return result, nil
 }
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -129,16 +129,16 @@ type Options struct {
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
 
-	// TransformFuncByObject is a map from GVKs to transformer functions which
+	// TransformByObject is a map from GVKs to transformer functions which
 	// get applied when objects of the transformation are about to be committed
 	// to cache.
 	//
 	// This function is called both for new objects to enter the cache,
 	// 	and for updated objects.
-	TransformFuncByObject TransformFuncByObject
+	TransformByObject TransformByObject
 
 	// DefaultTransform is the transform used for all GVKs which do
-	// not have an explicit transform func set in TransformFuncByObject
+	// not have an explicit transform func set in TransformByObject
 	DefaultTransform toolscache.TransformFunc
 }
 
@@ -158,7 +158,7 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	transformByGVK, err := convertToTransformByKindAndGVK(opts.TransformFuncByObject, opts.DefaultTransform, opts.Scheme)
+	transformByGVK, err := convertToTransformByKindAndGVK(opts.TransformByObject, opts.DefaultTransform, opts.Scheme)
 	if err != nil {
 		return nil, err
 	}
@@ -259,11 +259,11 @@ func convertToDisableDeepCopyByGVK(disableDeepCopyByObject DisableDeepCopyByObje
 	return disableDeepCopyByGVK, nil
 }
 
-// TransformFuncByObject associate a client.Object's GVK to a transformer function
+// TransformByObject associate a client.Object's GVK to a transformer function
 // to be applied when storing the object into the cache.
-type TransformFuncByObject map[client.Object]toolscache.TransformFunc
+type TransformByObject map[client.Object]toolscache.TransformFunc
 
-func convertToTransformByKindAndGVK(t TransformFuncByObject, defaultTransform toolscache.TransformFunc, scheme *runtime.Scheme) (internal.TransformFuncByObject, error) {
+func convertToTransformByKindAndGVK(t TransformByObject, defaultTransform toolscache.TransformFunc, scheme *runtime.Scheme) (internal.TransformFuncByObject, error) {
 	result := internal.NewTransformFuncByObject()
 	for obj, transformation := range t {
 		if err := result.Set(obj, scheme, transformation); err != nil {
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -186,38 +186,43 @@ var _ = Describe("Cache with transformers", func() {
 		By("creating the informer cache")
 		informerCache, err = cache.New(cfg, cache.Options{
 			DefaultTransform: func(i interface{}) (interface{}, error) {
-				if obj := i.(runtime.Object); obj != nil {
-					accessor, err := meta.Accessor(obj)
-					if err == nil {
-						annotations := accessor.GetAnnotations()
-						if _, exists := annotations["transformed"]; !exists {
-							if annotations == nil {
-								annotations = make(map[string]string)
-							}
-							annotations["transformed"] = "default"
-							accessor.SetAnnotations(annotations)
-						}
-					}
+				obj := i.(runtime.Object)
+				Expect(obj).NotTo(BeNil())
 
+				accessor, err := meta.Accessor(obj)
+				Expect(err).To(BeNil())
+				annotations := accessor.GetAnnotations()
+
+				if _, exists := annotations["transformed"]; exists {
+					// Avoid performing transformation multiple times.
+					return i, nil
 				}
+
+				if annotations == nil {
+					annotations = make(map[string]string)
+				}
+				annotations["transformed"] = "default"
+				accessor.SetAnnotations(annotations)
 				return i, nil
 			},
 			TransformByObject: cache.TransformByObject{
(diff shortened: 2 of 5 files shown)


FillZpp:
@alexzielenski Thanks. This is the first time I have seen https://github.com/kubernetes/kubernetes/pull/107507 , it is awesome that we can reduce the cache size with `TransformFunc`. So I'm thinking why not expose that for controller-runtime cache, so that developers could have their transform functions for different GVKs in cache?

Also this might have to wait for k8s v1.24 release to avoid alpha dependency in go.mod.

alexzielenski:
reviving this PR now that 1.24 is being release

alexzielenski:
@FillZpp What do you think of the new interface example usage below? It is a bit different from what we talked about on slack, because I still want to give control to only apply transformer to partialMetadata vs full objects.

```
nodeMetadata := &metav1.PartialObjectMetadata{}
nodeMetadata.SetGroupVersionKind(schema.GroupVersionKind{
						Group:   "",
						Version: "v1",
						Kind:    "Node",
					})
ctrl.Options.NewCache = cache.BuilderWithOptions(cache.Options{
                            TransformFuncByObject: cache.SelectorsByObject{
                                    &corev1.Node{}: func(val interface{}) interface{} {
                                        // Transformer for structured v1.Node
                                        ....
                                    }
                                    nodeMetadata: func(val interface{}) interface{} {
                                        /// Transformer for metadata v1.Node
                                        ....
                                    }
                                }
                            }
                        )
```

pkg/manager/manager_test.go
```diff
@@ -470,8 +471,10 @@ var _ = Describe("manger.Manager", func() {
 				Expect(ok).To(BeTrue())
 				multilock, isMultiLock := cm.resourceLock.(*resourcelock.MultiLock)
 				Expect(isMultiLock).To(BeTrue())
-				_, primaryIsConfigMapLock := multilock.Primary.(*resourcelock.ConfigMapLock)
```
alexzielenski, line 473: This type is no longer exported in 1.24. So I replaced it with closest equivalent code I could think of

pkg/cache/internal/transformers.go
```diff
@@ -0,0 +1,72 @@
...
+	case *metav1.PartialObjectMetadata:
+		return runtimeObjectKindMetadata
+	case *metav1.PartialObjectMetadataList:
+		return runtimeObjectKindMetadata
+	default:
+		return runtimeObjectKindStructured
+	}
+}
+
+type TransformFuncByObject map[runtimeObjectKind]map[schema.GroupVersionKind]cache.TransformFunc
```
FillZpp, line 36: This type and the enums may be a little confusing. How about this:

```golang
type TransformFuncByObject struct {
    structured TransformFuncByGVK
    unstructured TransformFuncByGVK
    metadata TransformFuncByGVK
}
```

And in the `Set(...)` you can use switch to choose which one should be added in.

pkg/cache/cache_test.go
```diff
@@ -121,6 +122,237 @@ var _ = Describe("Multi-Namespace Informer Cache", func() {
...
+			},
+		}
+
+		err = informerCache.List(informerCacheCtx, podsList)
+		Expect(err).ToNot(HaveOccurred())
+
+		//!TODO: I do not know a way to force the underlying listerwatcher
+		// to update. Please advise. Sleep here temporarily since it is required
+		// to reliably pass.
+		time.Sleep(1 * time.Second)
```
alexzielenski, line 282: @FillZpp or someone with knowledge: please advise on the best way to wait/trigger listerwatcher events for these tests?

alexzielenski, line 282: @FillZpp or someone with knowledge: please advise on the best way to wait/trigger listerwatcher events for these tests?

pkg/cache/cache_test.go
```diff
@@ -121,6 +122,237 @@ var _ = Describe("Multi-Namespace Informer Cache", func() {
...
```
alexzielenski, line 350: Also here

pkg/cache/cache_test.go
alvaroaleman: You can't, at least not within the integration tests, it happens asynchronously. The only way to do this I think would be to use a fake listwatcher, but that would be a largish refactoring: https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/cache/internal/informers_map.go#L223

What we do elsewhere in this situation is using `Eventually`, which polls: https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/cache/cache_test.go#L1142

That is also not ideal either ofc but IMHO better than a sleep, because depending on the performance of the computer that executes these tests, the required duration of the `sleep` varies which effectively means that you have to use the worst case duration which in turn artificially slows down the tests on faster computers.

pkg/cache/cache.go
```diff
@@ -128,6 +128,14 @@ type Options struct {
...
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
+
+	// TransformFuncByObject is a map from GVKs to transformer functions which
+	// get applied when objects of the transformation are about to be committed
+	// to cache.
+	//
+	// This function is called both for new objects to enter the cache,
+	// 	and for updated objects.
+	TransformFuncByObject TransformFuncByObject
```
alvaroaleman, line 138: Would it make sense to also add a `DefaultTransform` that is used if there is not a more specific one?

alexzielenski: That sounds potentially very useful to some cases. I will add that in.

pkg/cache/internal/transformers.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func (t TransformFuncByObject) Set(obj runtime.Object, scheme *runtime.Scheme, transformer cache.TransformFunc) error {
+	gvk, err := apiutil.GVKForObject(obj, scheme)
+	if err != nil {
+		return err
+	}
+
+	switch obj.(type) {
```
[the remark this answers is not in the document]
alvaroaleman: So you are ram pressed, you watch the same things in two different representations and in order to make that more efficient, you strip the data you already have in the metadata representation from the full representation...? Can you explain why you don't use the full representation throughout instead?

fwiw, `TransformFunc` is defined as  `func(interface{}) (interface{}, error)` and doesn't even godoc what you should expect to get except for `cache.DeletedFinalStateUnknown` - I personally would need to read the surrounding code or trial and error to figure out how to use it correctly one way or another.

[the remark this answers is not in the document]
alexzielenski: Apologies, I misrepresented/remembered the use case. You are right that it would be so simple to use the structured object the way I explained it!

In the case of the metadata watches, only the partial metadata is actually watched so in fact your proposal to split by only GVK is not an issue for this case. I was mistaken.

To your other comment: I've taken a note to improve the documentation/type requirements of the TransformFunc upstream in client-go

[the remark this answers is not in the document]
alexzielenski: Given the above, will wait a tiny bit to see if there is any more feedback on this API design before changing it to use a single transform for all representations of a GVK.

FillZpp:
/retitle ✨ Add TransformFuncByObject Option for Informer Cache

pkg/cache/cache.go
FillZpp: > Would it make sense to also add a `DefaultTransform` that is used if there is not a more specific one?

I'm not sure if we really need the `DefaultTransform`. This is different with the object selector IMHO.
There might be the same  `LabelSelector` for watching multiple GVKs, but since different kinds of object have different spec/status structure, they should be unlikely to have the same `Transform` logic.

pkg/cache/internal/transformers.go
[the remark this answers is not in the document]
FillZpp: > If someone writes a transform that expects a specific representation and then creates a cache for the same type in a different representation, that is effectively a bug.

This may happen if someone creates duplicate caches for the same GVK, as https://github.com/kubernetes-sigs/controller-runtime/issues/1660 recorded. But turns out only a comment doc has been added.
Should we have some mechanisms to check and warn users? Otherwise the one `Transform` for one GVK may work unexpectedly and have risk of panic.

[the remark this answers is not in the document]
alexzielenski: I've changed the API to use one transform for each representation. Still need to update tests/polish. Marking this one as resolved.

pkg/cache/cache.go
alexzielenski: I am in agreement with Alvaro. My first instinct was that one might want to use a global transform to strip fields from ALL objects. Additionally, it may be useful to write a transformer that modifies all objects conforming to a given interface. A `DefaultTransformer` would allow that.

pkg/cache/cache_test.go
alexzielenski: I've replaced the sleeps with eventually/consistently usages. Marking as resolved.

pkg/cache/cache.go
alexzielenski: Default transformer & test has been added. Marking as resolved.

pkg/cache/cache.go
```diff
@@ -128,6 +128,18 @@ type Options struct {
...
+
+	// DefaultTransform is the transform used for all GVKs which do
+	// not have an explicit transform func set in TransformFuncByObject
+	DefaultTransform toolscache.TransformFunc
```
FillZpp, line 142: Nit: option fields should be matched. `TransformFuncByObject` with `DefaultTransformFunc`, or `TransformByObject` with `DefaultTransform`. (I prefer the lower ones)

alexzielenski: I've renamed the public-facing API to use the form without `Func`

pkg/cache/internal/informers_map.go
```diff
@@ -227,6 +233,12 @@ func (ip *specificInformersMap) addInformerToMap(gvk schema.GroupVersionKind, ob
 	ni := cache.NewSharedIndexInformer(lw, obj, resyncPeriod(ip.resync)(), cache.Indexers{
 		cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
 	})
+
+	// Check to see if there is a transformer for this gvk
+	if err := ni.SetTransform(ip.transformers.Get(gvk)); err != nil {
```
alvaroaleman, line 238: Is the informer doing the right thing when `SetTransform` is called with a `nil` argument something we can rely on?

pkg/cache/cache_test.go
```diff
@@ -121,6 +123,216 @@ var _ = Describe("Multi-Namespace Informer Cache", func() {
...
+		knownPod2.GetObjectKind().SetGroupVersionKind(podGVK)
+		knownPod3.GetObjectKind().SetGroupVersionKind(podGVK)
+		knownPod4.GetObjectKind().SetGroupVersionKind(podGVK)
+		knownPod5.GetObjectKind().SetGroupVersionKind(podGVK)
+		knownPod6.GetObjectKind().SetGroupVersionKind(podGVK)
+
+		By("creating the informer cache")
+		informerCache, err = cache.New(cfg, cache.Options{
+			DefaultTransform: func(i interface{}) (interface{}, error) {
+				if obj := i.(runtime.Object); obj != nil {
```
alvaroaleman, line 189: I would prefer it if we could [flatten this arrow code](https://blog.codinghorror.com/flattening-arrow-code/) and check errors where we don't expect them to happen rather than just gating on `err == nil`

alvaroaleman (commented):
Two small points, other than that it looks good to me. @FillZpp is there anything left on your end?

pkg/cache/internal/informers_map.go
alexzielenski: If `SetTransform` is called with `nil` then no transform will be applied. That is the intent for this line when no transformer is returned for the gvk by `Get`.

pkg/cache/cache_test.go
alexzielenski: I've updated the code to flatten out the control flow. Marking this comment thread as resolved.

FillZpp:
> @FillZpp is there anything left on your end?

Yeah, I'm not sure if we should define some common transform functions so that most users don't have to search the source code to see how to implement it.
For example:

```
func TransformStripManagedFields(obj interface{}) (interface{}, error) {
    meta, err := apimeta.Accessor(obj)
    if err != nil {
        return nil, err
    }
    meta.SetManagedFields(nil)
    return obj, nil
}
```

It can be used for typed, Unstructured and PartialbjectMetadata objects.
WDYT @alvaroaleman 

Another question, should we deep copy first or just modify the given object in the transform func? @alexzielenski

Issue #1794, opened 2022-02-08
Set controller name as k/v pair in logger


sbueringer:
Hey folks,

I saw that the controller name is add via `WithName` calls on the logger of a controller. This results in a logger name of "controller/\<name\>". I wonder if we should set it as a k/v pair instead.

An example: (I'm using k/component-base/logs for logging)

### Current implementation: 

`Log:                     options.Log.WithName("controller").WithName(name),` 
(https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L138)

logging-format: text
> I0208 10:09:22.914540   30061 controller.go:224] controller/cluster "msg"="Starting workers" "reconciler group"="cluster.x-k8s.io" "reconciler kind"="ClusterClass"

logging-format: json
> {"ts":1644313339262.228,"caller":"controller/controller.go:224","msg":"controller/cluster: Starting workers","v":0,"reconciler group":"cluster.x-k8s.io","reconciler kind":"Cluster"}

### Using k/v: 

`Log:                     options.Log.WithValues("controller", name),`

logging-format: text
> I0208 10:40:01.223292   45529 controller.go:224]  "msg"="Starting workers" "controller"="cluster" "reconciler group"="cluster.x-k8s.io" "reconciler kind"="Cluster"

logging-format: json
> {"ts":1644313298144.636,"caller":"controller/controller.go:224","msg":"Starting workers","v":0,"reconciler group":"cluster.x-k8s.io","reconciler kind":"Cluster","controller":"cluster"}

As you can see the logger name is used as a prefix in the message (that should always be the case with klog, as klog sets the WithName as a prefix, not sure about other log implementations). I think it would be better to have it as a k/v pair instead as that makes it possible to filter logs based on the controller name, without having to parse the log message to do it.

ykakarap:
+1 to the general idea. It would make filtering logs for a specific controller much simpler.

sbueringer:
I'll close this issue for now as the k/v pair has been added. I'll include Fabrizio's requirement in a new issue that I'll open in a bit.

/close

PR #1806, opened 2022-02-15
🌱 Add controller-name as kv pair to logger instead of as name

sbueringer:

Addresses #1794. More context in the initial post of #1794

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"k8s.io/client-go/util/workqueue"
+
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
I like the idea but it is a breaking change as existing log queries will stop working. Are we comfortable with that?
/hold

pkg/controller/controller.go
```diff
@@ -135,7 +135,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		CacheSyncTimeout:        options.CacheSyncTimeout,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
-		Log:                     options.Log.WithName("controller").WithName(name),
+		Log:                     options.Log.WithValues("controller", name),
```
sbueringer, line 138: @alvaroaleman 
Maybe we can set the name and k/v?

I just saw this user story in the kubernetes contextual logging KEP:
> When kube-scheduler starts processing a pod, it creates a new logger with logger.WithValue("pod", klogr.KObj(pod)) and passes that around. While iterating over plugins in certain operations, another logger gets created with logger.WithName(<operation>).WithName(<plugin name>) and then is used when invoking that plugin. This adds a prefix to each log message which represents the call path to the specific log message, like for example NominatedPods/Filter/VolumeBinding.
> https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging#story-1

To me it seemed strange to have a msg prefix with the controller. But looking at that user story it kind of makes sense.

alvaroaleman: Yeah, that makes sense and seems a good idea

alvaroaleman (approved):
thanks!
/hold cancel

PR #1807, opened 2022-02-15
🐛  Check to see if obj is nil so there is no panic trying to get the gvk

emmjohnson:
If object is nil, this func will panic.

Co-authored-by: Marty Spiewak <mspiewak@vmware.com>

### DIFF
--- a/pkg/log/zap/zap_test.go
+++ b/pkg/log/zap/zap_test.go
@@ -257,7 +257,7 @@ var _ = Describe("Zap logger setup", func() {
 			})
 
 			It("should not panic with nil obj", func() {
-				pod := &corev1.Pod{}
+				var pod *corev1.Pod
 				pod = nil
 				logger.Info("here's a kubernetes object", "thing", pod)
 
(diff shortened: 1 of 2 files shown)


pkg/log/zap/zap_test.go
```diff
@@ -255,6 +255,15 @@ var _ = Describe("Zap logger setup", func() {
 					"namespace": name.Namespace,
 				}))
 			})
+
+			It("should not panic with nil obj", func() {
+				pod := &corev1.Pod{}
```
alvaroaleman, line 260: Nit to get a typed nilpointer:
```suggestion
				var pod *corev1.Pod
```

vincepri (approved):
/ok-to-test
/lgtm

nice find!

alvaroaleman:
/test pull-controller-runtime-apidiff-master
Not sure why the job doesn't actually start :thinking:

alvaroaleman:
Actually its optional anyways
/lgtm
/approve

Thanks!

PR #1810, opened 2022-02-16
🌱 log panic in reconcile even if panic should not be recovered

### LINKED ISSUE #1793: Logging a panic in Reconcile even if panic should not be recovered
We have the situation where we don't want to recover panics in Reconcile (RecoverPanic==false), so that panics are still immediately surfaced to users. But it would be great if we could log more information in that case.

I was thinking about a code change ~ like this:
```go
func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
	defer func() {
		if r := recover(); r != nil {
			if c.RecoverPanic {
				for _, fn := range utilruntime.PanicHandlers {
					fn(r)
				}
				err = fmt.Errorf("panic: %v [recovered]", r)
				return
			}

			log.Info(fmt.Sprintf("Observed a panic: %v", r))
			panic(r)
		}
	}()
	ctx = logf.IntoContext(ctx, log)
	return c.Do.Reconcile(ctx, req)
}
```
xref: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L100

The behavior for RecoverPanic=true wouldn't change. But for RecoverPanic=false we would essentially: recover, log, re-panic.

This way:
* The controller/Pod still goes immediately into CrashLoopBackoff 
* We additionally get a log line like this: (before the panic)
> I0208 10:09:26.613129   30061 controller.go:111] controller/cluster "msg"="Observed a panic: \<panic-error-text\>" "name"="my-cluster" "namespace"="default" "reconciler group"="cluster.x-k8s.io" "reconciler kind"="Cluster" 

So it's way easier to figure out which resource was reconciled when the panic occurred.

sbueringer:

Details in #1793

Fixes #1793
Fixes https://github.com/kubernetes-sigs/cluster-api/issues/5709

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -99,17 +99,21 @@ type watchDescription struct {
 
 // Reconcile implements reconcile.Reconciler.
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
-	if c.RecoverPanic {
-		defer func() {
-			if r := recover(); r != nil {
+	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	defer func() {
+		if r := recover(); r != nil {
+			if c.RecoverPanic {
 				for _, fn := range utilruntime.PanicHandlers {
 					fn(r)
 				}
 				err = fmt.Errorf("panic: %v [recovered]", r)
+				return
 			}
-		}()
-	}
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+
+			log.Info(fmt.Sprintf("Observed a panic in reconciler: %v", r))
+			panic(r)
+		}
+	}()
 	ctx = logf.IntoContext(ctx, log)
 	return c.Do.Reconcile(ctx, req)
 }


pkg/internal/controller/controller.go
```diff
@@ -99,17 +100,21 @@ type watchDescription struct {
...
 					fn(r)
 				}
 				err = fmt.Errorf("panic: %v [recovered]", r)
+				return
 			}
-		}()
-	}
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+
+			log.Info(fmt.Sprintf("Observed a panic: %v", r))
```
FillZpp, line 114: This line only prints the panic message like `Invalid memory address or nil pointer dereference`. Should we also use `utilruntime.PanicHandlers` to print the panic stack, so that user can track where it happened?

sbueringer: I'm not sure. The panic in l.115 will print the full panic stack, though.

I don't know enough about the idea behind PanicHandlers if it should be called it that case too.

Previously the "API" was that they are only called if RecoveryPanic was set to true.

FillZpp: > The panic in l.115 will print the full panic stack

But if the panic itself is enough, why should we recover and panic it again? This just add the duplicated message print.

vincepri, line 114:
```suggestion
			log.Info(fmt.Sprintf("Observed a panic in reconciler: %v", r))
```

vincepri (commented):
Small nit

pkg/internal/controller/controller.go
sbueringer: The benefit is that this log line correlates the panic with a concrete reconciled object.

For example we had the situation that a controller panic'ed at some point, but as there were ~50 objects which could have been reconciled at that moment it was impossible to figure out which object triggered the panic without a go debugger. The logger at this time contains the namespace and name of the current object, which makes it trivial to see which object "triggers" the panic.

sbueringer: Thx. fixed

sbueringer:
Findings should be addressed, PTAL.

Issue #1811, opened 2022-02-16
Enable per-reconcile logger customization


sbueringer:
It would be great if it would be possible to modify loggers for individual reconciles. 

Some context:
* When using the standard controller `pkg/builder.Builder` today you can set a logger which is later adjusted:
  * during Builder.Build():
    * "reconciler group" and "reconciler kind" k/v pairs are added
    * "controller" and "\<controller-name\>" are added as logger name
    * "controller": "\<controller-name\>" is added as k/v pair
  * per reconcile in reconcileHandler:
    * "name": "\<object-name\>", "namespace": "\<object-namespace\>" k/v pairs are added

The name and the k/v pair are added on top of either the logger provided in the builder or on top of `mgr.GetLogger()`.

We would like to be able to customize the logger according to our requirements. Concrete this means we would like to add a "traceID" and "\<kind\>": "\<namespace\>/\<name\>" k/v pair to the logger (more background info below). We could do this during the `Reconcile` call, but then the error logged by controller-runtime wouldn't have this context ([this one](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L317)).

As I'm pretty sure there are different opinions about what the name of a logger should look like and which k/v pairs should or shouldn't be added, I would propose to add a "hook" so that folks are able to customize the logger in the way they want.
Also e.g. dropping the current "name"/"namespace" k/v pairs would be a breaking change in controller-runtime.

### Background info about "traceID" and "\<kind\>": "\<namespace\>/\<name\>"

* "traceID": we want to be able to correlate logs of an individual reconcile. Furthermore we want to be able to correlate them with traces and metrics in the future
* "\<kind\>": "\<namespace\>/\<name\>": We are trying to follow the upstream [Kubernetes structured logging KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging#references-to-kubernetes-objects). The KEP standardizes the "value" of the k/v pair to be "'\<namespace\>/\<name\>". It doesn't exactly standardize the key of it but in CAPI we will probably follow the examples there and use the lower case singular kind.

I just wanted to provide this as background information. I think the discussion if this is a good or bad idea is out-of-scope (and still TBD in the CAPI project). But I think it should illustrate that it's worth making the logger customizable, so it's possible to accommodate requirements like this.

I've opened the following PR to illustrate how a solution could look like:
[...]

PR #1812, opened 2022-02-16
[WIP] 🌱 Enable per-reconcile logger customization

sbueringer:

For more details please see #1811

PR is currently [WIP] as it's only meant to illustrate a possible solution for #1811.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-logr/logr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/util/workqueue"
+
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -62,6 +63,9 @@ type Controller struct {
 	// SetFields is used to inject dependencies into other objects such as Sources, EventHandlers and Predicates
 	// Deprecated: the caller should handle injected fields itself.
 	SetFields func(i interface{}) error
+
+	// LoggerCustomizer customizes the logger for individual reconciliations.
+	LoggerCustomizer func(logr.Logger, reconcile.Request) logr.Logger
 
 	// mu is used to synchronize Controller setup
 	mu sync.Mutex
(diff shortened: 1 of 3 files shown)


pkg/internal/controller/controller.go
```diff
@@ -109,8 +113,6 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ re
 			}
 		}()
 	}
```
sbueringer, line 115: This was done twice before. Here and one layer above

pkg/internal/controller/controller.go
```diff
@@ -303,7 +305,12 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
 		return
 	}
 
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	var log logr.Logger
+	if c.LoggerCustomizer != nil {
+		log = c.LoggerCustomizer(c.Log, req)
```
sbueringer, line 310: We could also call c.LoggerCustomizer after l.312 instead of making it an if/else

I think I would prefer the current variant as there are no request specific k/v pairs in c.Log this way

pkg/internal/controller/controller.go
```diff
@@ -303,7 +305,12 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
...
+	} else {
+		log = c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	}
 	ctx = logf.IntoContext(ctx, log)
```
sbueringer, line 314:
```suggestion
	var log logr.Logger
	if c.LoggerCustomizer != nil {
		ctx, log = c.LoggerCustomizer(ctx, c.Log, req)
	} else {
		log = c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
		ctx = logf.IntoContext(ctx, log)
	}
```

I wonder if we should make this a bit more powerful. This way the `LoggerCustomizer` could 
use additional context to customize the ctx and the logger (obviously the name doesn't fit that good anymore).

I'm especially thinking about the tracing use case. By passing the ctx in and out of the ctx it would be possible to:
* use the traceID which is potentially passed in with the context in the future
* create a new trace / span and add it to the context

(P.S. this could almost also be done inside of the Reconcile func, but then the log.Error wouldn't have the traceID as k/v pair)

sbueringer:
/close
in favor of #1827

PR #1816, opened 2022-02-25
Delete OWNERS

muha0039:


palnabarun:
This is not adding any value.

/close

PR #1822, opened 2022-03-02
:bug: Adding "unmanaged" option for the "controller builder"

### LINKED ISSUE #1821: Add support to use controller builder to build unmanaged controller
## Problem
Currently, the [controller builder](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go) doesn't support building an unmanaged controller.

To achieve this, developers are required to manually configure and wire the controller (which is the benefit of the builder).

## Solution
I suggest adding a new method for the builder that specifies if it should use "unmanaged" mode:
```golang
func (blder *Builder) Unmanaged(unmanaged bool) *Builder 
```

The implementation for that is relatively easy and requires adding only a flag to the builder.

AlmogBaku:
Adding "unmanaged" option for the "controller builder".
Fixes #1821 , see this issue for more details.

The implementation is actually quite small (12 LOC + tests) :P

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -36,6 +36,7 @@ import (
 
 // Supporting mocking out functions for testing.
 var newController = controller.New
+var newUnmanagedController = controller.NewUnmanaged
 var getGvk = apiutil.GVKForObject
 
 // project represents other forms that the we can use to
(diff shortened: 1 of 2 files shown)


vincepri (commented):
The builders were meant to be used only for controllers that rely on a manager. If you'd like to use an unmanaged controller, why not call the function directly?

AlmogBaku:
@vincepri
Currently, the [controller builder](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go) doesn't support building an unmanaged builder.

To achieve this, developers are required to manually configure and wire the controller and write a lot of code (which is actually the idea of the builder)

AlmogBaku:
PS. it’s a bit tricky - an “Unmanaged controller” is actually does wired to the manager, but just not adding the Runnable to the manager:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L87

joelanford:
Closing this based on the lengthy Slack discussion on this topic: https://kubernetes.slack.com/archives/C02MRBMN00Z/p1646938170423789

PR #1827, opened 2022-03-04
⚠ logging: align to Kubernetes structured logging, add reconcileID

### LINKED ISSUE #1826: Improve logging
I would like to propose to adjust CR logging.

Today, controller-runtime logs have the following format: (when JSON logging is used)
```json
{
  "ts": 1646398972336.9634,
  "caller": "cluster/cluster_controller.go:105",
  "controller": "cluster",
  "reconciler group": "cluster.x-k8s.io",
  "reconciler kind": "Cluster",
  "msg": "controller/cluster: Waiting for DockerCluster",
  "v": 0,
  "name": "capi-quickstart",
  "namespace": "default"
}
```

I would like to propose the following format instead:

```json
{
  "ts": 1646398972336.9634,
  "caller": "cluster/cluster_controller.go:105",
  "controller": "cluster",
  "controllerGroup": "cluster.x-k8s.io",
  "controllerKind": "Cluster",
  "msg": "Waiting for DockerCluster",
  "v": 0,
  "cluster": {
    "name": "capi-quickstart",
    "namespace": "default"
  },
  "reconcileID": "56e044eb38fb76741ffe41cd83e7bed5"
}
```

The delta is:
```diff
-  "msg": "controller/cluster: Waiting for DockerCluster",
+  "msg": "Waiting for DockerCluster",
-  "name": "capi-quickstart",
-  "namespace": "default",
+  "cluster": {
+     "name": "capi-quickstart",
+     "namespace": "default"
+   },
-  "reconciler group": "cluster.x-k8s.io",
-  "reconciler kind": "Cluster",
+  "controllerGroup": "cluster.x-k8s.io",
+  "controllerKind": "Cluster",
+  "reconcileID": "56e044eb38fb76741ffe41cd83e7bed5",
```

The ideas behind that are:

**Drop the message prefix**
* I think the message prefix in every message is not necessary and just makes the logs unnecessarily verbose.
* With structured logging it's easily possible to just filter on / show the controller k/v pair instead.

**Log `"<kind>": "<namespace>/<name>"` instead of `"namespace"` and `"name"`**

* Let's align to the structured logging KEP. Specifically to: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging#object-reference-format-in-logs
    * Note: the value is formatted differently for JSON/text format.
* This produces logs consistent with the logs in upstream Kubernetes
* It is a lot easier to correlate logs for a specific object across controllers with this format    

**Add a reconcileID**

With a reconcileID it is very easy to identify logs belong to the same reconcile operation. 

**Change reconciler kind/ group to controllerKind and controllerGroup**

Given that Reconciler only seems to be the "inner" reconciler provided by the user, I think we should rather use controller (e.g. we're setting the name and object on the `NewControllerManagedBy` Builder )

Also to align with "casing" best practices outlined here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments

I've opened a PR to show how a potential solution would look like: [⚠ logging: align to Kubernetes structured logging, add reconcileID](https://github.com/kubernetes-sigs/controller-runtime/pull/1827)

P.S.
* This is an alternative to: https://github.com/kubernetes-sigs/controller-runtime/issues/1811 where we considered making logging customizable. 
* I know I made a related change recently to just add the controller k/v pair. We discussed this a bit and now want to propose to align CR logging overall.

sbueringer:

For more details, please see: https://github.com/kubernetes-sigs/controller-runtime/issues/1826

This PR introduces the `LogConstructor` controller option which replaces the previous `Log` option.
`LogConstructor` creates a `Logger` on-demand either within the context of a reconciliation based on a *reconcile.Request or outside of it (then nil is passed in).

Through this PR the logs change as follows:

When the controller is created with the builder:
```diff
-  "msg": "controller/cluster: Waiting for DockerCluster",
+  "msg": "Waiting for DockerCluster",
+  "cluster": {
+     "name": "capi-quickstart",
+     "namespace": "default"
+   },
-  "reconciler group": "cluster.x-k8s.io",
-  "reconciler kind": "Cluster",
+  "controllerGroup": "cluster.x-k8s.io",
+  "controllerKind": "Cluster",
+  "reconcileID": "56e044eb38fb76741ffe41cd83e7bed5",
```

When the controller is created with `controller.New` or `controller.NewUnmanaged`:
```diff
-  "msg": "controller/cluster: Waiting for DockerCluster",
+  "msg": "Waiting for DockerCluster",
+  "object": {
+     "name": "capi-quickstart",
+     "namespace": "default"
+   },
+  "reconcileID": "56e044eb38fb76741ffe41cd83e7bed5",
```

The output of the text format changed similarly. Although the objects are just formatted as `<namespace>/<name>` in this case (when using klog).

Please note by providing a LogConstructor via controller options everything except the `reconcileID` can be overwritten.

Fixes #1826

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-logr/logr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -148,9 +149,9 @@ func (blder *Builder) WithOptions(options controller.Options) *Builder {
 	return blder
 }
 
-// WithLogger overrides the controller options's logger used.
-func (blder *Builder) WithLogger(log logr.Logger) *Builder {
-	blder.ctrlOptions.Log = log
+// WithLogConstructor overrides the controller options's LogConstructor.
+func (blder *Builder) WithLogConstructor(logConstructor func(*reconcile.Request) logr.Logger) *Builder {
+	blder.ctrlOptions.LogConstructor = logConstructor
 	return blder
 }
 
@@ -304,13 +305,31 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 		ctrlOptions.CacheSyncTimeout = *globalOpts.CacheSyncTimeout
 	}
 
+	controllerName := blder.getControllerName(gvk)
+
 	// Setup the logger.
-	if ctrlOptions.Log.GetSink() == nil {
-		ctrlOptions.Log = blder.mgr.GetLogger()
+	if ctrlOptions.LogConstructor == nil {
+		log = blder.mgr.GetLogger().WithValues(
+			"controller", controllerName,
+			"controllerGroup", gvk.Group,
+			"controllerKind", gvk.Kind,
+		)
+
+		lowerCamelCaseKind := strings.ToLower(gvk.Kind[:1]) + gvk.Kind[1:]
+
+		ctrlOptions.LogConstructor = func(req *reconcile.Request) logr.Logger {
+			log := log
+			if req != nil {
+				log = log.WithValues(
+					lowerCamelCaseKind, klog.KRef(req.Namespace, req.Name),
+					"namespace", req.Namespace, "name", req.Name,
+				)
+			}
+			return log
+		}
 	}
-	ctrlOptions.Log = ctrlOptions.Log.WithValues("reconciler group", gvk.Group, "reconciler kind", gvk.Kind)
 
 	// Build the controller and return.
-	blder.ctrl, err = newController(blder.getControllerName(gvk), blder.mgr, ctrlOptions)
+	blder.ctrl, err = newController(controllerName, blder.mgr, ctrlOptions)
 	return err
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
@@ -46,9 +47,9 @@ type Options struct {
 	// The overall is a token bucket and the per-item is exponential.
 	RateLimiter ratelimiter.RateLimiter
 
-	// Log is the logger used for this controller and passed to each reconciliation
-	// request via the context field.
-	Log logr.Logger
+	// LogConstructor is used to construct a logger used for this controller and passed
+	// to each reconciliation via the context field.
+	LogConstructor func(request *reconcile.Request) logr.Logger
 
 	// CacheSyncTimeout refers to the time limit set to wait for syncing caches.
 	// Defaults to 2 minutes if not set.
@@ -105,8 +106,20 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		return nil, fmt.Errorf("must specify Name for Controller")
 	}
 
-	if options.Log.GetSink() == nil {
-		options.Log = mgr.GetLogger()
+	if options.LogConstructor == nil {
+		log := mgr.GetLogger().WithValues(
+			"controller", name,
+		)
+		options.LogConstructor = func(req *reconcile.Request) logr.Logger {
+			log := log
+			if req != nil {
+				log = log.WithValues(
+					"object", klog.KRef(req.Namespace, req.Name),
+					"namespace", req.Namespace, "name", req.Name,
+				)
+			}
+			return log
+		}
 	}
 
 	if options.MaxConcurrentReconciles <= 0 {
@@ -136,7 +149,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		CacheSyncTimeout:        options.CacheSyncTimeout,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
-		Log:                     options.Log.WithName("controller").WithName(name).WithValues("controller", name),
+		LogConstructor:          options.LogConstructor,
 		RecoverPanic:            options.RecoverPanic,
 	}, nil
 }
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-logr/logr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
@@ -83,8 +84,11 @@ type Controller struct {
 	// startWatches maintains a list of sources, handlers, and predicates to start when the controller is started.
 	startWatches []watchDescription
 
-	// Log is used to log messages to users during reconciliation, or for example when a watch is started.
-	Log logr.Logger
+	// LogConstructor is used to construct a logger to then log messages to users during reconciliation,
+	// or for example when a watch is started.
+	// Note: LogConstructor has to be able to handle nil requests as we are also using it
+	// outside the context of a reconciliation.
+	LogConstructor func(request *reconcile.Request) logr.Logger
 
 	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
 	RecoverPanic bool
@@ -99,7 +103,6 @@ type watchDescription struct {
 
 // Reconcile implements reconcile.Reconciler.
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
 	defer func() {
 		if r := recover(); r != nil {
 			if c.RecoverPanic {
@@ -110,11 +113,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ re
 				return
 			}
 
+			log := logf.FromContext(ctx)
 			log.Info(fmt.Sprintf("Observed a panic in reconciler: %v", r))
 			panic(r)
 		}
 	}()
-	ctx = logf.IntoContext(ctx, log)
 	return c.Do.Reconcile(ctx, req)
 }
 
@@ -144,7 +147,7 @@ func (c *Controller) Watch(src source.Source, evthdler handler.EventHandler, prc
 		return nil
 	}
 
-	c.Log.Info("Starting EventSource", "source", src)
+	c.LogConstructor(nil).Info("Starting EventSource", "source", src)
 	return src.Start(c.ctx, evthdler, c.Queue, prct...)
 }
 
@@ -179,7 +182,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		// caches to sync so that they have a chance to register their intendeded
 		// caches.
 		for _, watch := range c.startWatches {
-			c.Log.Info("Starting EventSource", "source", fmt.Sprintf("%s", watch.src))
+			c.LogConstructor(nil).Info("Starting EventSource", "source", fmt.Sprintf("%s", watch.src))
 
 			if err := watch.src.Start(ctx, watch.handler, c.Queue, watch.predicates...); err != nil {
 				return err
@@ -187,7 +190,7 @@ func (c *Controller) Start(ctx context.Context) error {
 		}
 
 		// Start the SharedIndexInformer factories to begin populating the SharedIndexInformer caches
-		c.Log.Info("Starting Controller")
+		c.LogConstructor(nil).Info("Starting Controller")
 
 		for _, watch := range c.startWatches {
 			syncingSource, ok := watch.src.(source.SyncingSource)
@@ -204,7 +207,7 @@ func (c *Controller) Start(ctx context.Context) error {
 				// is an error or a timeout
 				if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
 					err := fmt.Errorf("failed to wait for %s caches to sync: %w", c.Name, err)
-					c.Log.Error(err, "Could not wait for Cache to sync")
+					c.LogConstructor(nil).Error(err, "Could not wait for Cache to sync")
 					return err
 				}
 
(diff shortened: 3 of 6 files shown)


pkg/internal/controller/controller.go
```diff
@@ -353,3 +381,46 @@ func (c *Controller) InjectFunc(f inject.Func) error {
 func (c *Controller) updateMetrics(reconcileTime time.Duration) {
 	ctrlmetrics.ReconcileTime.WithLabelValues(c.Name).Observe(reconcileTime.Seconds())
 }
+
+// KRef returns ObjectRef from name and namespace
+// Note: This is a copy of the func from klog. It has been copied to avoid
+// introducing a dependency to klog, while still implement logging according
+// to the Kubernetes structured logging KEP.
```
sbueringer, line 388: This is based on the assumption that we don't want to introduce a dependency to klog.
I don't know if that's true.

pkg/controller/controller.go
```diff
@@ -84,8 +88,8 @@ type Controller interface {
 
 // New returns a new Controller registered with the Manager.  The Manager will ensure that shared Caches have
 // been synced before the Controller is Started.
-func New(name string, mgr manager.Manager, options Options) (Controller, error) {
-	c, err := NewUnmanaged(name, mgr, options)
+func New(name string, gvk *schema.GroupVersionKind, mgr manager.Manager, options Options) (Controller, error) {
```
sbueringer, line 91: I considered adding gvk to Options instead. But to me it doesn't look like it would fit there.

sbueringer:
Not sure if it's related, but let's see.

pkg/controller/controller.go
vincepri: Why not?

pkg/controller/controller.go
```diff
@@ -126,6 +130,19 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
...
+	if gvk != nil {
+		log = log.WithValues("reconciler group", gvk.Group, "reconciler kind", gvk.Kind)
+	}
+
+	// Initialize random source, later used to generate reconcileIDs.
+	var rngSeed int64
+	if err := binary.Read(crand.Reader, binary.LittleEndian, &rngSeed); err != nil {
+		return nil, fmt.Errorf("could not read random bytes to seed random source for reconcileID generation: %v", err)
+	}
+	randSource := rand.New(rand.NewSource(rngSeed)) //nolint:gosec // math/rand is enough, we don't need crypto/rand.
```
vincepri, line 144: Is there any reason not to use crypto/rand anyway?

pkg/controller/controller.go
```diff
@@ -136,7 +153,9 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		CacheSyncTimeout:        options.CacheSyncTimeout,
 		SetFields:               mgr.SetFields,
 		Name:                    name,
-		Log:                     options.Log.WithName("controller").WithName(name).WithValues("controller", name),
+		GVK:                     gvk,
```
vincepri, line 156:
```suggestion
		GroupVersionKind:                     gvk,
```

pkg/internal/controller/controller.go
vincepri: We definitely can

pkg/controller/controller.go
sbueringer: Would be fine for me. I essentially copied this from the traceID/spanID generator from opentelemetry

Maybe it needs less "randomness"? :)

pkg/controller/controller.go
alvaroaleman: I don't think we should be doing this, this breaks the abstration and makes us hardcode the assumption "one controller has exactly one kubernetes type object it owns". So far, this assumpton is only in the builder and we should IMHO leave it like that. A controller is just a thing that has event sources, if they trigger it puts something in a q and eventually reconciles them.

Offhand examples where this new assumption wouldn't work out:
* Controller that reconciles many things based on a constant reconcile request (because easier than writing a controller per thing): https://github.com/openshift/hypershift/blob/eec759e5b6096ffc2ba823665bccce6e1c54f99a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L135
* Controller that reconciles based on keys that are not representing any resource that actually exists, just the abstraction at which the wokqeue is supposed to de-duplicate for us: https://github.com/openshift/ci-tools/blob/cd6d2f7d5ead0dbb72d75dc7a72062491ae17d6e/pkg/controller/promotionreconciler/prowjobreconciler/prowjobreconciler.go#L132

alvaroaleman: The builder already has this information, so by setting up logging for this in the builder we can both get the right logger and keep the abstraction intact.

sbueringer: I agree. Even having it in this func signature kind of breaks the abstraction.

My main concern was that GVK doesn't really seem to fit in the existing options. Especially I wouldn't know how we would end up with a clean semantic for this field, e.g. what happens if you set the GVK to another GVK that the one you're reconciling or as Alvaro mentioned what if you're not even reconciling a Kubernetes object (or multiple with the same reconciler).

The reason why I considered adding it was because I need the kind (or fallback "object") during a Reconcile to add a k/v pair which identifies the currently reconciled thing. It's not possible to only configure the key of a k/v pair on the logger.

What about adding a `LogObjectKey` (with default "object") to the options? 
Details:
* I would move adding the reconciler group / kind back into the builder
* The LogObjectKey would be set to the kind in the builder and then used during reconcile to log the k/v pair of the reconciled object.

pkg/internal/controller/controller.go
```diff
@@ -307,7 +318,24 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
 		return
 	}
 
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
+	// Add object to the logger.
+	var objectLogKey = "obj"
+	if c.GroupVersionKind != nil {
+		objectLogKey = strings.ToLower(c.GroupVersionKind.Kind)
```
sbueringer, line 324: Note to myself:
* We should follow the convention for keys too: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments (affects kind + at least reconciler group/kind)

pkg/controller/controller.go
alvaroaleman: How about passing a `logConstructor func(reconcile.Key) logr.Logr` rather than a literal `log` into the controller and making it call that with the key prior to passing it on to the reconciler?

pkg/internal/controller/controller.go
```diff
@@ -353,3 +370,12 @@ func (c *Controller) InjectFunc(f inject.Func) error {
 func (c *Controller) updateMetrics(reconcileTime time.Duration) {
 	ctrlmetrics.ReconcileTime.WithLabelValues(c.Name).Observe(reconcileTime.Seconds())
 }
+
+// generateReconcileID generates a reconcileID for logging.
+func (c *Controller) generateReconcileID() (string, error) {
+	id := [16]byte{}
+	if _, err := c.RandSource.Read(id[:]); err != nil {
+		return "", fmt.Errorf("failed to generate reconcileID: %v", err)
```
sbueringer, line 378: Probably this case never occurs, but I wonder if we should fallback to some constant reconcileID if this fails instead of failing the whole reconcile.

pkg/controller/controller.go
sbueringer: Good idea. Tried to implement accordingly. Please take a look if that's what you had in mind.

pkg/builder/controller.go
```diff
@@ -304,13 +308,25 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
 	}
 
+	controllerName := blder.getControllerName(gvk)
+
 	// Setup the logger.
-	if ctrlOptions.Log.GetSink() == nil {
-		ctrlOptions.Log = blder.mgr.GetLogger()
+	if ctrlOptions.LogConstructor == nil {
+		log := blder.mgr.GetLogger()
+		log = log.WithValues("reconcilerGroup", gvk.Group, "reconcilerKind", gvk.Kind)
```
sbueringer, line 316: Not sure if we need reconcilerGroup + reconcilerKind, but I'm fine with keeping them

pkg/controller/controller.go
sbueringer: Based on https://blog.gopheracademy.com/advent-2017/a-tale-of-two-rands/ I would prefer using math rand. crypto rand seems to be way slower. I think for us it's good enough to seed math rand with crypto and then just use the pseudo number generator provided by math rand.

pkg/builder/controller.go
```diff
@@ -289,6 +290,9 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+	if gvk.Kind == "" {
```
alvaroaleman, line 293: Is this possible? GVK should always include a Kind or not?

pkg/builder/controller.go
```diff
@@ -304,13 +308,27 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
+	controllerName := blder.getControllerName(gvk)
+
 	// Setup the logger.
-	if ctrlOptions.Log.GetSink() == nil {
-		ctrlOptions.Log = blder.mgr.GetLogger()
+	if ctrlOptions.LogConstructor == nil {
+		log = blder.mgr.GetLogger().WithValues(
+			"controller", controllerName,
+			"controllerGroup", gvk.Group,
+			"controllerKind", gvk.Kind,
```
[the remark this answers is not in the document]
alvaroaleman: Can we enumerate these changes in the description, so that ppl have an easy time updating their log query?

pkg/controller/controller.go
alvaroaleman: Yeah I think math rand should be enough as well, we don't expect cryptographic  security in these IDs.

But why not use `math/rand.Read` or maybe even `github.com/google/uuid`?

pkg/internal/controller/controller.go
```diff
@@ -83,8 +85,12 @@ type Controller struct {
 	// startWatches maintains a list of sources, handlers, and predicates to start when the controller is started.
 	startWatches []watchDescription
 
-	// Log is used to log messages to users during reconciliation, or for example when a watch is started.
-	Log logr.Logger
+	// LogConstructor is used to construct a logger to then log messages to users during reconciliation,
+	// or for example when a watch is started.
+	LogConstructor func(request *reconcile.Request) logr.Logger
```
alvaroaleman, line 90: Please document that it needs to handle being called with a nil `request`

pkg/internal/controller/controller.go
```diff
@@ -99,7 +105,6 @@ type watchDescription struct {
 
 // Reconcile implements reconcile.Reconciler.
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
-	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
```
alvaroaleman, line 102: Can we leave this? It is somewhat redundant, but it will keep existing queries working

pkg/internal/controller/controller.go
```diff
@@ -110,11 +115,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ re
 				return
 			}
 
+			log := logf.FromContext(ctx)
```
alvaroaleman, line 118: Why are we not using the `logConstructor` for this?

pkg/internal/controller/controller.go
```diff
@@ -83,8 +85,12 @@ type Controller struct {
...
 	startWatches []watchDescription
 
-	// Log is used to log messages to users during reconciliation, or for example when a watch is started.
-	Log logr.Logger
+	// LogConstructor is used to construct a logger to then log messages to users during reconciliation,
+	// or for example when a watch is started.
+	LogConstructor func(request *reconcile.Request) logr.Logger
+
+	// RandSource is used to generate reconcileIDs for logging.
+	RandSource *rand.Rand
```
alvaroaleman, line 93: Why is this configurable? Is there a valid usecase for wanting a custom rand source for the reconcile ID? And I really think the whole id generation would be easier if we used something like the uuid package

pkg/builder/controller.go
sbueringer: Exactly. I only added this to be 100% safe against panics in l.321 even if code elsewhere changes.

But I'm fine with dropping it if you prefer.

pkg/controller/controller.go
sbueringer: I'll check. I was mostly just reusing what they are doing in open telemetry to end up with the same format as traceIDs. But I don't mind diverging there is no strict reason why we have to use the same.

pkg/internal/controller/controller.go
sbueringer: I'll check the uuid package. It absolutely doesn't have to be configurable. I added it here because I didn't want to create a new RandSource for every reconcile. But when we move to uuid that's not necessary anymore.

pkg/internal/controller/controller.go
sbueringer: I'm already using it one layer above to generate a logger + add the reconcileID afterwards.

If I "re-generate" the logger the reconcileID gets lost

pkg/internal/controller/controller.go
sbueringer: I'm not sure if we want to have the redundancy going forward / forever.

WDYT about the following options:
* keeping it for now, but deprecating it (I'm not sure how) and then drop it with a later minor release (not the next one)
* providing a LogConstructor which uses either the old format or the new one + this k/v pair. In this case folks can still use the old format but would have to set that constructor explicitly

pkg/controller/controller.go
sbueringer: Let's use k8s.io/apimachinery/pkg/util/uuid (which is using the google lib under the hood)

pkg/internal/controller/controller.go
sbueringer: This code has been dropped

pkg/internal/controller/controller.go
sbueringer: Has been dropped

pkg/internal/controller/controller.go
sbueringer: Done

pkg/builder/controller.go
sbueringer: I've dropped it. We're checking that object is not nil one layer above + getGvk always returns either a gvk or an error. So should be safe enough.

pkg/builder/controller.go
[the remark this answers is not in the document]
sbueringer: Added the current state, will update accordingly when necessary later on.

sbueringer:
@alvaroaleman Findings should be addressed or answered, PTAL. Thank you very much!

pkg/internal/controller/controller.go
sbueringer: Apart from having it twice in the log which makes it slightly less readable, I assume there are no drawbacks.

I assume systems like Elastic and Loki are optimized enough that this doesn't really require more storage.

Then I'll keep it.

sbueringer: I added it in the log constructors to keep it close to where I already add the object. If someone has strong opinions on that they can pass in their own LogConstructor.

PTAL.

alvaroaleman (approved):
looks good to me, thanks for the work!
/hold
@joelanford @vincepri any concerns or objections?

vincepri:
+1 from me to merge

sbueringer:
Thanks everyone for the quick feedback and good discussions! :)

Issue #492, opened 2019-06-19
Better handling `kind is not registered in scheme` error


mengqiy:
When a kind is not registered in the scheme, it returns a typed error. We should handle it and return a more user friendly error message.

vincepri:
There is a log message now that explains this a bit more https://github.com/kubernetes-sigs/controller-runtime/blob/98e2435ba8888cb3332978e4aaa02b31e3e2e73a/pkg/internal/source/kind.go#L62-L66 in a few cases

/close

PR #1830, opened 2022-03-07
:sparkles: [WIP]  handle "no kind is registered for the type" errors

umangachapagain:
When a kind is not registered in the scheme, it returns a typed error.
This PR handles that error and returns a more user friendly error message.

Fixes: #492 

Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -182,11 +182,19 @@ func (c *client) RESTMapper() meta.RESTMapper {
 func (c *client) Create(ctx context.Context, obj Object, opts ...CreateOption) error {
 	switch obj.(type) {
 	case *unstructured.Unstructured:
-		return c.unstructuredClient.Create(ctx, obj, opts...)
+		err := c.unstructuredClient.Create(ctx, obj, opts...)
+		if err != nil && runtime.IsNotRegisteredError(err) {
+			return fmt.Errorf("custom IsNotRegisteredError:%w", err)
+		}
+		return err
 	case *metav1.PartialObjectMetadata:
 		return fmt.Errorf("cannot create using only metadata")
 	default:
-		return c.typedClient.Create(ctx, obj, opts...)
+		err := c.typedClient.Create(ctx, obj, opts...)
+		if err != nil && runtime.IsNotRegisteredError(err) {
+			return fmt.Errorf("custom IsNotRegisteredError:%w", err)
+		}
+		return err
 	}
 }
 


umangachapagain:
Is this the right place and approach to fix this issue for clients?
There are other places where that error is returned, at what level should we log this?

joelanford:
I'm not sure I understand the value proposition of the error wrapping in this particular case. This would add an extra conditional check to every `Create` call for all callers, and it would set the precedent that the myriad of other potential errors that could be returned by the underlying `Create` could/should have special handling in controller-runtime.

Can you provide an example of the before and after error messages based on this PR?

FillZpp:
Another question is, if you are going to handle this error, you should handle it for all methods of both Reader and Writer, because they will all return the error if the given object can not be found in scheme.

umangachapagain:
> Can you provide an example of the before and after error messages based on this PR?
>
Before: `no kind is registered for the type v1.Deployment in scheme "pkg/runtime/scheme.go:100"`.
After: `custom IsNotRegisteredError:no kind is registered for the type v1.Deployment in scheme "pkg/runtime/scheme.go:100"`
>
Plan is to add a user friendly message in place of "custom IsNotRegisteredError" for people who don't already know what this error means (like Kubebuilder users).
>
> I'm not sure I understand the value proposition of the error wrapping in this particular case. This would add an extra conditional check to every `Create` call for all callers, and it would set the precedent that the myriad of other potential errors that could be returned by the underlying `Create` could/should have special handling in controller-runtime.
> 
I didn't think of this earlier 😓.  I agree.
Since callers already receive original typed error, they can handle it in any way they want.

I looked at other places where this error is received via `scheme.ObjectKinds()` or `apiutil.GVKForObject()` calls.

Most of them just return the error as it is.
https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/controller/controllerutil/controllerutil.go#L70-L73
But, few log it with additional context  and return the original error
https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/handler/enqueue_owner.go#L101-L105
and few return a wrapped error
https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/builder/controller.go#L208-L211
or simply append it to another error
https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/envtest/webhook.go#L181-L184

Now I am wondering, how do we decide which approach to take? Should we update all of these to simply return the error?
And, if we don't want to handle this specific error here, can we close #492 with a comment on how to handle this error?

umangachapagain:
> Another question is, if you are going to handle this error, you should handle it for all methods of both Reader and Writer, because they will all return the error if the given object can not be found in scheme.
>
Yes, that is the plan. But, I wanted to start a discussion on how we do that before changing all of them.
Starting a conversation on the issue would've been better. My bad.

pkg/client/client.go
```diff
@@ -182,11 +182,19 @@ func (c *client) RESTMapper() meta.RESTMapper {
 func (c *client) Create(ctx context.Context, obj Object, opts ...CreateOption) error {
 	switch obj.(type) {
 	case *unstructured.Unstructured:
-		return c.unstructuredClient.Create(ctx, obj, opts...)
+		err := c.unstructuredClient.Create(ctx, obj, opts...)
```
alvaroaleman, line 185: How would that error ever occur here? The whole point of `unstructured` is to not need types.

pkg/client/client.go
```diff
@@ -182,11 +182,19 @@ func (c *client) RESTMapper() meta.RESTMapper {
...
+			return fmt.Errorf("custom IsNotRegisteredError:%w", err)
+		}
+		return err
 	case *metav1.PartialObjectMetadata:
 		return fmt.Errorf("cannot create using only metadata")
 	default:
-		return c.typedClient.Create(ctx, obj, opts...)
+		err := c.typedClient.Create(ctx, obj, opts...)
+		if err != nil && runtime.IsNotRegisteredError(err) {
+			return fmt.Errorf("custom IsNotRegisteredError:%w", err)
```
alvaroaleman, line 195: That is not a very user friendly error message. Also, please add a test for this

joelanford:
I actually had a case recently where I needed to introspect the error returned by the client to specifically check for `meta.NoKindMatchError`, and it turned out that the client actually wraps those errors (at least in some cases) in `errors.Aggregate`, which does not handle Go's error unwrapping correctly.

I ended up having to write a recursive helper function that specifically checked for `errors.Aggregate` layers so that I could loop through each of those errors to continue the search for `meta.NoKindMatchError`.

This problem isn't directly related to this PR, but it came as a surprise that I couldn't use `errors.As` to check for `meta.NoKindMatchError` when its embedded in an `errors.Aggregate`.

From errors.Aggregate GoDoc:
> Errors.As() is not supported, because the caller presumably cares about a specific error of potentially multiple that match the given type. 

:man_shrugging:

alvaroaleman:
> I actually had a case recently where I needed to introspect the err [...]

Right, but `errors.Is` works and should be fine for this use-case?

PR #1837, opened 2022-03-11
:book: Add documentation for zap-time-encoding flag

creydr:
In #1688 the `zap-time-encoding` flag was added. Unfortunately I missed to update the documentation of `Options.BindFlags`, so this flag is not really described in the doc.
This PR addresses it and adds the documenation for the `zap-time-encoding` flag so it shows up in the [go-doc](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags).


alvaroaleman (approved):
Thanks!

PR #1838, opened 2022-03-11
✨  Support multiple field selectors

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>

### DIFF
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -113,17 +113,55 @@ func (c *CacheReader) List(_ context.Context, out client.ObjectList, opts ...cli
 	listOpts.ApplyOptions(opts)
 
 	switch {
-	case listOpts.FieldSelector != nil:
+	case listOpts.FieldSelector != nil && !listOpts.FieldSelector.Empty():
 		// TODO(directxman12): support more complicated field selectors by
 		// combining multiple indices, GetIndexers, etc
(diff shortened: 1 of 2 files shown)


FillZpp (commented):
Emm.. I don't think we can support this by listing objects multiple times. You have to know that the indexer always keep changing, which means the objects might be modified between the multiple lists.

For example, when user list with field selectors `{"field1": "a", "field2": "b"}`, he/she wants to get objects that match the two requirements. But there may be an object that previously have `{"field1": "a", "field2": "y"}` so it is returned by your first list with `field1` index. And then someone updates it to be `{"field1": "x", "field2": "b"}`, so it is also returned by your second list with `field2` index.

In this way, you will return the object to user, but the problem is that it has never matched the two requirements at the same time.

tamalsaha:
Makes sense, @FillZpp !

tamalsaha:
Thanks @FillZpp ! I have added checks for generation. Now, we can be sure that the result for the all indices are for the same object.

weilaaa:
Hi, there. I wanna add a complicated index into thread safe store. If it will be accepted, it helps here.

Related [issues](https://github.com/kubernetes/kubernetes/issues/109329#issuecomment-1107688739) and [pr](https://github.com/kubernetes/kubernetes/pull/109334)

vincepri:
Closing due to inactivity, @tamalsaha please feel free to reopen once ready

/close

PR #1844, opened 2022-03-24
:bug: Let the envtest apiserver listen also on non-localhost

### LINKED ISSUE #1751: Setting APIServer.SecureServing.Address to anything other than 127.0.0.1 envtest.Start fails.
Repro code:

```go
	apiServer := envtest.APIServer{}
	apiServer.SecureServing.Address = "0.0.0.0"
	apiServer.SecureServing.Port = "6884"

	testEnv := &envtest.Environment{
		ControlPlane: envtest.ControlPlane{
			APIServer: &apiServer,
		},
	}

	//start testEnv
	cfg, err := testEnv.Start()
	if err != nil {
		log.Fatal(err)
	}
```

output:

```
2021/12/16 14:51:34 unable to install CRDs onto control plane: unable to create CRD instances: unable to create client: Get "https://0.0.0.0:6884/api?timeout=32s": x509: certificate is valid for ::1, 127.0.0.1, not 0.0.0.0
exit status 1
```

jkmw:
Currently, the certificate is only issued for localhost. With this PR, the IP from `APIServer.SecureServing.ListenAddr.Address` is also included, if it exists.

fixes #1751

### DIFF
--- a/pkg/internal/testing/controlplane/apiserver.go
+++ b/pkg/internal/testing/controlplane/apiserver.go
@@ -375,7 +375,11 @@ func (s *APIServer) populateAPIServerCerts() error {
 		return err
 	}
 
-	servingCerts, err := ca.NewServingCert()
+	certNames := []string{"localhost"}
+	if len(s.SecureServing.Address) > 0 {
+		certNames = append(certNames, s.SecureServing.Address)
+	}
+	servingCerts, err := ca.NewServingCert(certNames...)
 	if err != nil {
 		return err
 	}


zawachte:
ty!

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -375,7 +375,11 @@ func (s *APIServer) populateAPIServerCerts() error {
 		return err
 	}
 
-	servingCerts, err := ca.NewServingCert()
+	certNames := []string{"localhost"}
+	if len(s.SecureServing.Address) > 0 {
+		certNames = append(certNames, s.SecureServing.Address)
+	}
+	servingCerts, err := ca.NewServingCert(certNames...)
```
varshaprasad96, line 382: Only even when no certnames are passed to `ca.NewServingCert()`, `localhost` is getting added by [default](https://github.com/jkmw/controller-runtime/blob/616b89b5d49212fac50bb31cbdd6903eb4ed65a3/pkg/internal/testing/certs/tinyca.go#L160). Do we modify the condition in `pkg/internal/testing/certs/tinyca.go` instead of here?

jkmw: Would also be possible, of course. Should I implement this or is it a consideration for now?

vincepri:
Can we add a test?

Haegi:
Are there any updates on this? I would be also willing to help

PR #1846, opened 2022-03-24
✨ [WIP] Add BaseContext to manager Options for use with Runnables

### LINKED ISSUE #1817: Manager runnableGroup causes a disconnect between contexts
I was working on a Reconciler today and found that values I stored in the context passed to a Manager's `Start` method weren't available in the `Reconcile` context, so I started digging into it.

I ended up tracking it down to the `runnableGroup` struct, and specifically this line in the `newRunnableGroup` function: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/runnable_group.go#L110

It's creating an internal context for the group, but it's not being based on any other context. I looked through the code and couldn't find any code that references `r.ctx` or `r.cancel` before `Start` could be called, aside from the `StopAndWait` method where it's calling `r.cancel`.

If the `Start` method is updated to set `r.ctx` and `r.cancel` inside the `r.startOnce.Do` call with the context passed to `Start`, it seems like the context from the initial manager `Start` call is connected all the way to the `Reconcile` context. If `StopAndWait` is updated to check if `r.cancel` is `nil` or not before it tries calling it I think it's possible to implement it safely.

Is this intended functionality? It does seem like it works as expected the way the code is now, but is the intent to split the Manager's `Start` context from the ultimate `Reconcile` context?

I was going to submit a PR for this, but I saw you request an issue to be created first, so that's what this is!

aphistic:
This PR adds a new `BaseContext` option to `manager.Options`. It's designed to be used as the basis of any `Runnable`s created by a `Manager`, so there's a way to connect the contexts from the user's side to the `Runnable` side. Currently a new `context.Background()` is created so the `Runnable`s aren't canceled when a top-level context is, so this provides a way for a user to put info in their `Runnable` contexts.

There's a discussion about this in #1817, and this PR closes #1817.

I've marked this as a WIP because the functionality works fine through anecdotal usage, but I'm unable to run the tests locally using the recommended `make test`, so I can't add any tests for this at the moment. I brought it up in the #controller-runtime Slack channel but haven't seen any responses yet.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -239,6 +239,11 @@ type Options struct {
 	// use the cache for reads and the client for writes.
 	NewClient cluster.NewClientFunc
 
+	// BaseContext is the function that provides Context values to Runnables
+	// managed by the Manager. If a BaseContext function isn't provided, Runnables
+	// will receive a new Background Context instead.
+	BaseContext BaseContextFunc
+
 	// ClientDisableCacheFor tells the client that, if any cache is used, to bypass it
 	// for the given objects.
 	ClientDisableCacheFor []client.Object
@@ -277,6 +282,10 @@ type Options struct {
 	newMetricsListener     func(addr string) (net.Listener, error)
 	newHealthProbeListener func(addr string) (net.Listener, error)
 }
+
+// BaseContextFunc is a function used to provide a base Context to Runnables
+// managed by a Manager.
+type BaseContextFunc func() context.Context
 
 // Runnable allows a component to be started.
 // It's very important that Start blocks until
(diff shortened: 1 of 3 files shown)


vincepri (approved):
Hi @aphistic apologies for the delay in answering. The changes in this PR look good and I'm completely fine merging it in, it gives ultimate control to the caller and retains current functionality.

The main reason we need a context that's separate from the main start one is to control the stop procedure in a really specific order.

aphistic:
> Hi @aphistic apologies for the delay in answering. The changes in this PR look good and I'm completely fine merging it in, it gives ultimate control to the caller and retains current functionality.
> 
> The main reason we need a context that's separate from the main start one is to control the stop procedure in a really specific order.

Hi @vincepri! My main concern about it is that I wasn't able to add any tests for it to validate that the `BaseContext` is being connected through to the `Runnable`. I'm happy to see it merged, but I could add these tests if I could get `make test` to run locally for me. I keep running into linting errors that don't make any sense. =/

PR #1847, opened 2022-03-25
✨ [WIP] Improve performance of CreateOrPatch

bogdando:
When either of the before/after objects has no status data, the
reflect.DeepEqual call becomes redundant because comparing of
existing to non-existing always returns False.

Switch the logical OR condition to AND. That saves us a single
reflect.DeepEqual() call each time we end up with an object without
status after its mutation func was called (or before that).

Signed-off-by: Bogdan Dobrelya <bdobreli@redhat.com>

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -304,7 +304,7 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 		result = OperationResultUpdated
 	}
 
-	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
+	if hasBeforeStatus && hasAfterStatus && !reflect.DeepEqual(beforeStatus, afterStatus) {
 		// Only issue a Status Patch if the resource has a status and the beforeStatus
 		// and afterStatus copies differ
 		if result == OperationResultUpdated {


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -304,7 +304,7 @@ func CreateOrPatch(ctx context.Context, c client.Client, obj client.Object, f Mu
 		result = OperationResultUpdated
 	}
 
-	if (hasBeforeStatus || hasAfterStatus) && !reflect.DeepEqual(beforeStatus, afterStatus) {
+	if hasBeforeStatus && hasAfterStatus && !reflect.DeepEqual(beforeStatus, afterStatus) {
```
vincepri, line 307: Wouldn't this check now fail if the object didn't have status before the call, but it does have it afterwards?

bogdando: You're right. The 2nd case would yield unexpected result:
```
hasBeforeStatus	F	T/F	T
hasAfterStatus	F	F/T	T
not eq			F* 	T**	T/F
--------------------
result OR			F	T	T/F
result AND		F	F	T/F

*cannot be True
**cannot be False
```

bogdando: OK there is nothing to improve here, since ``(A || B) && ...`` would stop processing all the rightmost, if either A or B yielded True...

bogdando:
sorry for the noise!

PR #1849, opened 2022-03-26
🐛 Fallback on temp directory if user cache directory has issues

### LINKED ISSUE #1845: Panic creating kubebuilder-envtest cache directory
Encountered following error:
```
$ kubectl -n infra logs -f eks-provider-db6d54fd5-gzx7t
panic: mkdir /.cache: permission denied

goroutine 1 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/testing/addr.init.0()
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/testing/addr/manager.go:51 +0xda
```
controller-runtime version: `v0.9.6`
golang version: `go1.17.7`

The issue is related to the `init` function that sets up `kubebuilder-envtest` caching directory, see [code](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.6/pkg/internal/testing/addr/manager.go#L44-L53).

The [`UserCacheDir`](https://github.com/golang/go/blob/go1.17.7/src/os/file.go#L393-L442) is supposed to return the cache directory location. In the pod setup, `HOME` is set to `/` and `XDG_CACHE_HOME` is not set so `UserCacheDir` ends up returning `//.cache` without emitting error. However, the process doesn't have permission to create directory under that location and thus it paniced.

One walkaround is to manually set `XDG_CACHE_HOME` to `/tmp` in deployment spec. It would be great if we have logic that fall back to using `/tmp` on permission issues.

hanlins:
The original logic is if we can get the user cache directory(_i.e._ `os.UserCacheDir()`), then create the cache directory under that user cache directory. Otherwise, create the cache directory under the temp directory.

Basically `os.UserCacheDir` could return a path that the current user doesn't have permission for (typically in container environment). `os.UserCacheDir` typically returns the `$HOME`. If `$HOME` is set but the user doesn't have permission to create directories in `$HOME`, it will panic. This is somewhat confusing since it could also fallback on the temp directory.

The update logic will simply attempt to create directory under `os.UserCacheDir()`. If `os.UserCacheDir` returns error or user encounters error creating directories under it, it will fallback onto temp directory and retry.

Here's the manual verification:
```
➜  test cat main.go
package main

import (
	"fmt"
	"os"
	"path/filepath"
)

func main() {
	const (
		eventTestDir = "kubebuilder-envtest"
		permission   = 0750
	)
	if baseDir, err := os.UserCacheDir(); err == nil {
		cacheDir := filepath.Join(baseDir, eventTestDir)
		fmt.Println("#1", cacheDir)
		if err := os.MkdirAll(cacheDir, permission); err == nil {
			return
		}
	}
	cacheDir := filepath.Join(os.TempDir(), eventTestDir)
	fmt.Println("#2", cacheDir)
	if err := os.MkdirAll(cacheDir, permission); err != nil {
		panic(err)
	}

}
➜  test GO111MODULE=off go build .
➜  test # if HOME directory doesn't have permission issue
➜  test HOME=$(pwd)/somepath ./test
#1 /home/ubuntu/test/somepath/.cache/kubebuilder-envtest
➜  test # if HOME directory has permission issue
➜  test rm -rf somepath/.cache
➜  test sudo chown root:root somepath
➜  test HOME=$(pwd)/somepath ./test
#1 /home/ubuntu/test/somepath/.cache/kubebuilder-envtest
#2 /tmp/kubebuilder-envtest
```

Fixes #1845

### DIFF
--- a/pkg/internal/testing/addr/manager.go
+++ b/pkg/internal/testing/addr/manager.go
@@ -42,12 +42,18 @@ var (
 )
 
 func init() {
-	baseDir, err := os.UserCacheDir()
-	if err != nil {
-		baseDir = os.TempDir()
+	const (
+		envTestDir = "kubebuilder-envtest"
+		permission = 0750
+	)
+	if baseDir, err := os.UserCacheDir(); err == nil {
+		cacheDir = filepath.Join(baseDir, envTestDir)
+		if err := os.MkdirAll(cacheDir, permission); err == nil {
+			return
+		}
 	}
-	cacheDir = filepath.Join(baseDir, "kubebuilder-envtest")
-	if err := os.MkdirAll(cacheDir, 0750); err != nil {
+	cacheDir = filepath.Join(os.TempDir(), envTestDir)
+	if err := os.MkdirAll(cacheDir, permission); err != nil {
 		panic(err)
 	}
 }


vincepri:
@hanlins There is already a PR out for this issue, see #1800

PR #1851, opened 2022-03-29
🐛 returned objects of reference type should be unchangeable

mars1024:
Signed-off-by: Bruce Ma <brucema19901024@gmail.com>

The returned slice of `extractor` function is reference type, reusing it will bring unpredictable problems, it should be unchangeable.


pkg/cache/informer_cache.go
```diff
@@ -193,8 +193,8 @@ func indexByField(indexer Informer, field string, extractor client.IndexerFunc)
 		rawVals := extractor(obj)
 		var vals []string
 		if ns == "" {
-			// if we're not doubling the keys for the namespaced case, just re-use what was returned to us
-			vals = rawVals
+			// if we're not doubling the keys for the namespaced case, just create a new slice with same length
```
alvaroaleman, line 196: Sorry I don't understand. If you write a function that returns a pointer, you are not supposed to keep a reference to that pointer or its underlying value.

alvaroaleman: Can you clarify why you are doing that?

mars1024: The story is, I compute and put the indexer values into the object status `.status.indexerValues`, so the extractor will be
```
func(obj client.Object) []string {
    someObject, _ := obj.(*group.ObjectKind)
    return someObject.Status.IndexerValues
}
```
As we know, the object from client cache should be unchangeable.

Just think another scene,
```
values := []string{"a", "b", "c"}
do(values)
doAgain(values)
```
Do we expect `values` will be changed by `do`? Then `doAgain` will get unexpected values.

Though slice(extractor return) is a pointer type here, but it is used because it contains immutable values, so I don't think it should be changed when being used.

mars1024: Slice(extractor return) here is more like a container containing immutable values(indexer values).

mars1024: Of course, will do it ~

mars1024:
test case added, PTAL @alvaroaleman @FillZpp , thanks!

alvaroaleman (approved):
thanks!

PR #1858, opened 2022-04-07
🌱 Log all errors from cache.GetInformer call

devholic:
Currently, `kind.Start` prints the last error from the polling process, except for the NoKindMatchError. 

This behavior makes it users hard to debug the cause (e.g. forgot to add types to the scheme) of the `getInformer` call failure since it always returns the context timeout error, which is the last error while polling.

This commit logs all returned errors while calling `getInformer` in the polling process.

Signed-off-by: Sunghoon Kang <hoon@linecorp.com>

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
thanks!

PR #1867, opened 2022-04-13
✨ support for namespaced name predicate

### LINKED ISSUE #1866: Support for NamespacedName predicate
# Feature request

Would be cool if we could create a predicate to watch a specific resource, starting from a `NamespacedName` type.

prometherion:
Closes #1866.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -17,10 +17,12 @@ limitations under the License.
 package predicate
 
 import (
+	"fmt"
 	"reflect"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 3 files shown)


pkg/predicate/predicate.go
```diff
@@ -350,3 +352,15 @@ func LabelSelectorPredicate(s metav1.LabelSelector) (Predicate, error) {
...
 }
+
+// MatchingNamespacedNamePredicate constructs a Predicate to match a Namespaced object.
+// Only objects matching the Name and Namespace pair will be admitted.
+func MatchingNamespacedNamePredicate(namespacedName types.NamespacedName) (Predicate, error) {
+	if len(namespacedName.Name) == 0 {
+		return Funcs{}, fmt.Errorf("name cannot be empty")
+	}
+
+	return NewPredicateFuncs(func(object client.Object) bool {
```
alvaroaleman, line 363: I am not sure I see much value in this Predicate over just defining the `NewPredicateFuncs` inline. And the fact that it can error makes it quite inconvenient to use

FillZpp (commented):
Actually, I don't think this is a common predicate that only watches one certain object for a whole controller. Could you describe its usage? @prometherion 
We should not put every custom predicate into controller-runtime. Just keep it in the project where the Predicate is needed. WDYT?

PR #1877, opened 2022-04-26
:sparkles: certwatcher: add metrics to monitor certificate reads

isitinschi:
Helps to see how efficient is certwatcher and with debugging issues around reloading certificates from disk.


isitinschi:
@pwittrock @vincepri let me know if there is anything missing in my PR to get it reviewed 🙏

vincepri (approved):

PR #1889, opened 2022-05-06
:bug: Avoid nilref when copying leader election options from custom config

### LINKED ISSUE #1828: nilref panic when configuring leader election
With the unlucky combination of providing a config file with no config for leader elections (which should be OK, since the element [is optional](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/config/v1alpha1/types.go#L40) in the config type) and providing the `--leader-elect` CLI argument, you'll get a nilref panic in [`setLeaderElectionConfig`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L487-L517), since `obj.LeaderElection == nil`.

We could avoid this by just adding an early exit in this case, but I'm not sure if that will result in the intended behavior; perhaps it's better to do
```go
if obj.LeaderElection == nil {
    obj.LeaderElection = &configv1alpha1.LeaderElectionConfiguration{}
}
```
somewhere before entering this method?

tomasaschan:
This fixes #1828.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -494,6 +494,11 @@ func (o Options) AndFromOrDie(loader config.ControllerManagerConfiguration) Opti
 }
 
 func (o Options) setLeaderElectionConfig(obj v1alpha1.ControllerManagerConfigurationSpec) Options {
+	if obj.LeaderElection == nil {
+		// The source does not have any configuration; noop
+		return o
+	}
+
 	if !o.LeaderElection && obj.LeaderElection.LeaderElect != nil {
 		o.LeaderElection = *obj.LeaderElection.LeaderElect
 	}
--- a/pkg/manager/manager_options_test.go
+++ b/pkg/manager/manager_options_test.go
@@ -1,0 +1,54 @@
+package manager
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/controller-runtime/pkg/config"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	configv1alpha1 "sigs.k8s.io/controller-runtime/pkg/config/v1alpha1"
+)
+
+var _ = Describe("manager.Options", func() {
+	Describe("AndFrom", func() {
+		Describe("reading custom type using OfKind", func() {
+			var (
+				o   Options
+				c   customConfig
+				err error
+			)
+
+			JustBeforeEach(func() {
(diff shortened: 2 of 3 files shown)


pkg/manager/manager.go
```diff
@@ -494,6 +494,11 @@ func (o Options) AndFromOrDie(loader config.ControllerManagerConfiguration) Opti
 }
 
 func (o Options) setLeaderElectionConfig(obj v1alpha1.ControllerManagerConfigurationSpec) Options {
+	if obj.LeaderElection == nil {
```
alvaroaleman, line 497: There is a nilcheck right below this (`!o.LeaderElection && obj.LeaderElection.LeaderElect != nil`

Can you show a stacktrace of this panicing and please add a test? The current fix doesn't seem to me like it will do anything

FillZpp, line 497: Emm... I don't think we can return `o Options` directly if `obj.LeaderElection` is nil. The leaderelection fields in `o Options` should be set default values below.

tomasaschan: The nilcheck right below is the line that fails for me, since `obj.LeaderElection` is nil (and therefore `obj.LeaderElection.LeaderElect` panics). I'll see if I can figure out where to add test code that executes this path.

tomasaschan: What should those default values be, if `obj.LeaderElection` is not provided? There are no set operations on `o` that don't read from `obj.LeaderElection` in this function.

This code seems to assume that `obj.LeaderElection` will, at the very least, be initialized to a struct value with the defaults set; equivalent to something like
```
if obj.LeaderElection == nil {
    obj.LeaderElection = LeaderElectionConfiguration{}
}
```
However, if one does that, _none_ of the `if` statements below this will evaluate to true, which means the function will return `o` unmodified. This is exactly what my PR accomplishes as well.

tomasaschan: @alvaroaleman I pushed a test that panics without this check.

tomasaschan: @FillZpp I pushed a test that panics without this check. If this check is not the way that panic should be avoided, could you please outline what you think should be done instead?

FillZpp: Oh, apologies for my misunderstanding. You are right, thanks!

FillZpp: @alvaroaleman The `!o.LeaderElection && obj.LeaderElection.LeaderElect != nil` is ok, but the logic below like `if o.LeaderElectionResourceLock == "" && obj.LeaderElection.ResourceLock != "" {` will panic if `obj.LeaderElection` is nil.

alvaroaleman (approved):

PR #1891, opened 2022-05-09
🐛 Fix panic for lazy dynamicRESTMapper

### LINKED ISSUE #1712: Panic from dynamicRESTMapper when using lazy initialization
When `dynamicRESTMapper` lazy initialization fails, any subsequent calls to it will panic (`runtime error: invalid memory address or nil pointer dereference`) because `staticMapper` is uninitialized.

I ran into this when my cluster could not be reached, so I'll illustrate the problem this way, but any error returned by `setStaticMapper` will result in the same issue.

## How to reproduce
```golang
import (
)

func main() {
	// get a client config; this should succeed
	cfg := config.GetConfigOrDie()
	// to simulate a refused connection, set cfg.Host to some URL that won't answer
	cfg.Host = "https://127.0.0.1:12345"
	// create a new dynamicRESTMapper instance with lazy initialization; this should succeed
	mapper, _ := apiutil.NewDynamicRESTMapper(cfg, apiutil.WithLazyDiscovery)
	// call any method on mapper; this should result in an error
	_, err := mapper.KindFor(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "pods"})
	if err != nil {
		fmt.Println("failed to get kind for pods")
	}
	// later on, (or from another place/goroutine in your code) call any method on mapper; this should panic
	_, err := mapper.KindFor(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "pods"})
	if err != nil {
		fmt.Println("failed to get kind for pods")
	}
}
```

## Suggested fix
`sync.Once` is not suitable for guarding the lazy initialization code because the initialization should be retried until successful, not just tried once. I couldn't find an existing solution, so I propose to create a custom version of `sync.Once` that is _done_ only when the function given to `Do` returns with `true`, although, other alternative implementations might be fine too.

```golang
type Once struct {
	done uint32
	m    Mutex
}

func (o *Once) Do(f func() bool) {
	if atomic.LoadUint32(&o.done) == 0 {
		o.doSlow(f)
	}
}

func (o *Once) doSlow(f func() bool) {
	o.m.Lock()
	defer o.m.Unlock()
	if o.done == 0 {
		if f() {
			atomic.StoreUint32(&o.done, 1)
		}
	}
}
```
This should be used instead of `sync.Once`, updating the inline function passed to `Do` with a `return err == nil` statement.

FillZpp:

Fix panic for lazy dynamicRESTMapper to make `setStaticMapper` successfully once, which means if dynamicRESTMapper fails to `setStaticMapper`, it will do it once again in next `init`.

fixes #1712

### DIFF
--- a/pkg/client/apiutil/dynamicrestmapper_test.go
+++ b/pkg/client/apiutil/dynamicrestmapper_test.go
@@ -14,9 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package apiutil_test
+package apiutil
 
 import (
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -26,8 +27,6 @@ import (
 	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
-	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 var (
@@ -52,7 +51,7 @@ var _ = Describe("Dynamic REST Mapper", func() {
 		}
 
 		lim = rate.NewLimiter(rate.Limit(5), 5)
-		mapper, err = apiutil.NewDynamicRESTMapper(cfg, apiutil.WithLimiter(lim), apiutil.WithCustomMapper(func() (meta.RESTMapper, error) {
+		mapper, err = NewDynamicRESTMapper(cfg, WithLimiter(lim), WithCustomMapper(func() (meta.RESTMapper, error) {
 			baseMapper := meta.NewDefaultRESTMapper(nil)
 			addToMapper(baseMapper)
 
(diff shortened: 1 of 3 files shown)


pkg/client/apiutil/dynamicrestmapper_test.go
```diff
@@ -146,11 +145,28 @@ var _ = Describe("Dynamic REST Mapper", func() {
...
 
-	PIt("should lazily initialize if the lazy option is used", func() {
-
-	})
+		It("should lazily initialize if the lazy option is used", func() {
+			var err error
+			var failedOnce bool
+			mockErr := fmt.Errorf("mock failed once")
+			mapper, err = NewDynamicRESTMapper(cfg, WithLazyDiscovery, WithCustomMapper(func() (meta.RESTMapper, error) {
+				// Make newMapper failed once
```
siliconbrain, line 154:
```suggestion
				// Make newMapper fail once
```

siliconbrain (approved):
lgtm, thanks @FillZpp!

pkg/client/apiutil/dynamicrestmapper_test.go
FillZpp: Done, thanks.

alvaroaleman (approved):

Issue #1754, opened 2021-12-22
Allow configuring TLS ciphers for webhook server


akankshapanse:
With changes dony by https://github.com/kubernetes-sigs/controller-runtime/pull/1548, webhook server has TLSMinVersion configurable, while setting up webhook server with manager. However, there is no option to configure the TLSCiphers.

For TLSMinVersion set to 1.2, default TLS cipher suites, coming from Go crypto/tls library, include some insecure ciphers, which leave the system open to multiple vulnerabilities.

Similar issue raised earlier under https://github.com/kubernetes-sigs/controller-runtime/issues/1431, which needs to be addressed for TLS 1.2 version set.

ruchikaguptaa:
We are running into the same issue. Are there any updates on this?

PR #1897, opened 2022-05-11
:sparkles: Allow TLS to be entirely configured on webhook server

akalenyu:
Some operators might want to respect cluster-wide TLS ciphers for example,
which means that these will eventually have to be passed down to the webhook server.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1754

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -76,6 +76,9 @@ type Server struct {
 	// "", "1.0", "1.1", "1.2" and "1.3" only ("" is equivalent to "1.0" for backwards compatibility)
 	TLSMinVersion string
 
+	// TLSOpts is used to allow configuring the TLS config used for the server
+	TLSOpts []func(*tls.Config)
+
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
 
@@ -252,6 +255,11 @@ func (s *Server) Start(ctx context.Context) error {
 
 		cfg.ClientCAs = certPool
 		cfg.ClientAuth = tls.RequireAndVerifyClientCert
+	}
+
+	// fallback TLS config ready, will now mutate if passer wants full control over it
+	for _, op := range s.TLSOpts {
+		op(cfg)
 	}
 
 	listener, err := tls.Listen("tcp", net.JoinHostPort(s.Host, strconv.Itoa(s.Port)), cfg)
--- a/pkg/webhook/server_test.go
+++ b/pkg/webhook/server_test.go
@@ -18,6 +18,7 @@ package webhook_test
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net"
@@ -186,7 +187,7 @@ var _ = Describe("Webhook Server", func() {
 		})
 	})
 
-	It("should serve be able to serve in unmanaged mode", func() {
+	It("should be able to serve in unmanaged mode", func() {
 		server = &webhook.Server{
 			Host:    servingOpts.LocalServingHost,
 			Port:    servingOpts.LocalServingPort,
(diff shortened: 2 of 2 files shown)


pkg/webhook/server.go
```diff
@@ -76,6 +79,10 @@ type Server struct {
 	// "", "1.0", "1.1", "1.2" and "1.3" only ("" is equivalent to "1.0" for backwards compatibility)
 	TLSMinVersion string
 
+	// TLSCiphers is used to specify the cipher algorithms that are negotiated
+	// during the TLS handshake, refer to https://pkg.go.dev/crypto/tls#CipherSuites
+	TLSCiphers []string
```
alvaroaleman, line 84: Why not just make it `TLSOpts []func(*tls.Config)` to avoid needing a new setting everytime someone finds something new they would like to configure about TLS?

akalenyu: Thanks, that makes sense!
I have kept the previously-existing TLS fields to not break existing usages

pkg/webhook/server.go
```diff
@@ -41,6 +41,9 @@ import (
 // DefaultPort is the default port that the webhook server serves.
 var DefaultPort = 9443
 
+// ExportedTLSConfig is used in unit tests to ensure propagation of tls related configurables to server.
+var ExportedTLSConfig *tls.Config
```
alvaroaleman, line 45: Lets find a way to test this without globals

akalenyu: Done, hopefully better

akalenyu:
Can't find an open issue about this, should I open one?

> @akalenyu: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-controller-runtime-test-master	[187187e](https://github.com/kubernetes-sigs/controller-runtime/commit/187187eeddf649241369da9ac4c67e0aefcd7659)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1897/pull-controller-runtime-test-master/1526127109066461184)	true	`/test pull-controller-runtime-test-master`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=controller-runtime&pr=1897). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Aakalenyu). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/controller-runtime/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

alvaroaleman (approved):
thank you!

akalenyu:
@alvaroaleman @camilamacedo86 
Would you consider a backport of this to 0.12 so we can have it in a release?
We wanted to capitalize on this (not huge) feature but really shouldn't use unreleased/unpinned versions in our projects. Hopefully, that is ok

alvaroaleman:
sure
/cherrypick release-0.12

mythi:
Should this be possible via manager's `Options` too?

akalenyu:
> Should this be possible via manager's `Options` too?

Should be able to pass it in https://github.com/kubernetes-sigs/controller-runtime/blob/e4e13b981704e6f0bd87a26da4cdd556e2ec0a6a/pkg/manager/manager.go#L247

mythi:
> I might be missing something, but, the snippet above **is** from the `Manager.Options{}`, so you indeed should be able to do that?

Right, I'm all good passing `*webhook.Server`.

But when that is not passed, some elements of `webhook.Server` can still be set using `Manager.Options{}` with `Host/Port/CertDir` but that's just a subset. To set the new `TLS*` parameters  (that have been added in the past 12 months) users need to pass in `*webhook.Server` but they cannot be set using  `Manager.Options{}` directly.

srm09:
Are there any plans to backport this fix to the 1.11.x release line?

PR #1900, opened 2022-05-13
:sparkles: webhook: add an option to recover from panics in handler

isitinschi:
Currently, a panic occcurence in a webhook handler is not recovered and crashes the webhook server.

This change adds an option to Webhook to recover panics similar to how it is handled in Reconciler. It ensures that panics are converted to normal error response.

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -39,6 +39,7 @@ type WebhookBuilder struct {
 	gvk           schema.GroupVersionKind
 	mgr           manager.Manager
 	config        *rest.Config
+	recoverPanic  bool
 }
 
 // WebhookManagedBy allows inform its manager.Manager.
@@ -65,6 +66,12 @@ func (blder *WebhookBuilder) WithDefaulter(defaulter admission.CustomDefaulter) 
 // WithValidator takes a admission.WithValidator interface, a ValidatingWebhook will be wired for this type.
 func (blder *WebhookBuilder) WithValidator(validator admission.CustomValidator) *WebhookBuilder {
 	blder.withValidator = validator
+	return blder
+}
+
+// RecoverPanic indicates whether the panic caused by webhook should be recovered.
+func (blder *WebhookBuilder) RecoverPanic() *WebhookBuilder {
+	blder.recoverPanic = true
 	return blder
 }
 
@@ -124,10 +131,10 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
 
 func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
 	if defaulter := blder.withDefaulter; defaulter != nil {
-		return admission.WithCustomDefaulter(blder.apiType, defaulter)
+		return admission.WithCustomDefaulter(blder.apiType, defaulter).WithRecoverPanic(blder.recoverPanic)
 	}
 	if defaulter, ok := blder.apiType.(admission.Defaulter); ok {
-		return admission.DefaultingWebhookFor(defaulter)
+		return admission.DefaultingWebhookFor(defaulter).WithRecoverPanic(blder.recoverPanic)
 	}
 	log.Info(
 		"skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called",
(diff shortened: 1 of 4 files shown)


isitinschi:
@vincepri WDYT?

isitinschi:
@FillZpp thank you for the review. I've addressed all your suggestions. PTAL

pkg/builder/webhook.go
```diff
@@ -68,6 +69,12 @@ func (blder *WebhookBuilder) WithValidator(validator admission.CustomValidator)
 	return blder
 }
 
+// WithRecoverPanic takes a bool flag which indicates whether the panic caused by webhook should be recovered.
+func (blder *WebhookBuilder) WithRecoverPanic(recoverPanic bool) *WebhookBuilder {
```
FillZpp, line 73: How about just defined as `func (blder *WebhookBuilder) RecoverPanic() * WebhookBuilder ` so that `recoverPanic` should be `true` if it is called?

pkg/webhook/admission/defaulter.go
```diff
@@ -33,9 +33,10 @@ type Defaulter interface {
 }
 
 // DefaultingWebhookFor creates a new Webhook for Defaulting the provided type.
-func DefaultingWebhookFor(defaulter Defaulter) *Webhook {
+func DefaultingWebhookFor(defaulter Defaulter, recoverPanic bool) *Webhook {
```
FillZpp, line 36: I have a little concern about these public API change, not sure if there were some ppl directly use them.

So how about a new `func (wh *Webhook) WithRecoverPanic*(enable bool) *Webhook` and you can use it in builder/webhook like this `return admission.ValidatingWebhookFor(validator).WithRecoverPanic(blder. recoverPanic)`

isitinschi:
@FillZpp is it better like this?

isitinschi:
@alvaroaleman Does it look good to you to be merged?

pkg/webhook/admission/webhook.go
```diff
@@ -138,11 +143,32 @@ func (wh *Webhook) InjectLogger(l logr.Logger) error {
...
+
 // Handle processes AdmissionRequest.
 // If the webhook is mutating type, it delegates the AdmissionRequest to each handler and merge the patches.
 // If the webhook is validating type, it delegates the AdmissionRequest to each handler and
 // deny the request if anyone denies.
-func (wh *Webhook) Handle(ctx context.Context, req Request) Response {
+func (wh *Webhook) Handle(ctx context.Context, req Request) (response Response) {
+	defer func() {
+		if r := recover(); r != nil {
+			if wh.RecoverPanic {
```
vincepri, line 159: Shouldn't this `if` condition be before the `recover()`?

isitinschi: We won't be able to process the panic if we don't recover from it first. Panic recovery works the same way in controller.go: https://github.com/kubernetes-sigs/controller-runtime/blob/15154aaa67679df320008ed45534f83ff3d6922d/pkg/internal/controller/controller.go#L108

alvaroaleman: But we don't want to process it when RecoverPanic is false? What is done in controller.go seems quite pointless to me and IIRC the `r` won't contain the stacktrace?

isitinschi: Makes sense, let's make `RecoverPanic` flag more fair and really not process anything if it is set to false. Adjusted my PR accordingly 👍

isitinschi:
@vincepri @alvaroaleman does it look good to you?

pkg/webhook/admission/webhook.go
```diff
@@ -138,11 +143,29 @@ func (wh *Webhook) InjectLogger(l logr.Logger) error {
...
+}
+
 // Handle processes AdmissionRequest.
 // If the webhook is mutating type, it delegates the AdmissionRequest to each handler and merge the patches.
 // If the webhook is validating type, it delegates the AdmissionRequest to each handler and
 // deny the request if anyone denies.
-func (wh *Webhook) Handle(ctx context.Context, req Request) Response {
+func (wh *Webhook) Handle(ctx context.Context, req Request) (response Response) {
+	defer func() {
+		if wh.RecoverPanic {
```
alvaroaleman, line 158:
```suggestion
        if wh.RecoverPanic 
		defer func() {
```

alvaroaleman:
Thank you!
/lgtm
/approve

PR #1901, opened 2022-05-14
✨ Enable metrics of clientgo leader election

zqzten:
This PR enables metrics of clientgo leader election ( "k8s.io/client-go/tools/leaderelection" )

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -144,6 +144,9 @@ type controllerManager struct {
 	// webhookServer if unset, and Add() it to controllerManager.
 	webhookServerOnce sync.Once
 
+	// leaderElectionID is the name of the resource that leader election
+	// will use for holding the leader lock.
+	leaderElectionID string
 	// leaseDuration is the duration that non-leader candidates will
 	// wait to force acquire leadership.
 	leaseDuration time.Duration
@@ -637,6 +640,7 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 			},
 		},
 		ReleaseOnCancel: cm.leaderElectionReleaseOnCancel,
+		Name:            cm.leaderElectionID,
 	})
 	if err != nil {
 		return err
--- a/pkg/metrics/leaderelection.go
+++ b/pkg/metrics/leaderelection.go
@@ -1,0 +1,40 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/tools/leaderelection"
+)
+
+// This file is copied and adapted from k8s.io/component-base/metrics/prometheus/clientgo/leaderelection
+// which registers metrics to the k8s legacy Registry. We require very
+// similar functionality, but must register metrics to a different Registry.
+
+var (
+	leaderGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "leader_election_master_status",
+		Help: "Gauge of if the reporting system is master of the relevant lease, 0 indicates backup, 1 indicates master. 'name' is the string used to identify the lease. Please make sure to group by name.",
+	}, []string{"name"})
+)
+
+func init() {
+	Registry.MustRegister(leaderGauge)
+	leaderelection.SetProvider(leaderelectionMetricsProvider{})
+}
+
+type leaderelectionMetricsProvider struct{}
+
+func (leaderelectionMetricsProvider) NewLeaderMetric() leaderelection.SwitchMetric {
+	return &switchAdapter{gauge: leaderGauge}
+}
+
+type switchAdapter struct {
+	gauge *prometheus.GaugeVec
+}
+
+func (s *switchAdapter) On(name string) {
+	s.gauge.WithLabelValues(name).Set(1.0)
+}
+
+func (s *switchAdapter) Off(name string) {
+	s.gauge.WithLabelValues(name).Set(0.0)
+}
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -439,6 +439,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		certDir:                       options.CertDir,
 		tlsOpts:                       options.TLSOpts,
 		webhookServer:                 options.WebhookServer,
+		leaderElectionID:              options.LeaderElectionID,
 		leaseDuration:                 *options.LeaseDuration,
 		renewDeadline:                 *options.RenewDeadline,
 		retryPeriod:                   *options.RetryPeriod,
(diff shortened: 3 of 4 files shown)


pkg/metrics/leaderelection.go
```diff
@@ -0,0 +1,41 @@
...
+	"k8s.io/client-go/tools/leaderelection"
+)
+
+// This file is copied and adapted from k8s.io/component-base/metrics/prometheus/clientgo/leaderelection
+// which registers metrics to the k8s legacy Registry. We require very
+// similar functionality, but must register metrics to a different Registry.
+
+var (
+	leaderGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "leader_election_master_status",
```
camilamacedo86, line 14: What means master status?

pkg/metrics/client_go_adapter.go
```diff
@@ -152,20 +84,6 @@ func registerClientMetrics() {
 	})
 }
 
-// registerReflectorMetrics sets up reflector (reconcile) loop metrics.
```
camilamacedo86, line 155: Why are we removing these ones? Will we still export the same?
is not it a breaking change?

pkg/metrics/leaderelection.go
zqzten: It means whether this instance is master or backup, as described in the help field. The name is just grabbed from k8s.io/component-base which is also used by kube-controller-manager.

pkg/metrics/client_go_adapter.go
zqzten: These metrics have not been exported for a long time since cliengo has removed them, so the deletion of these code won't cause break changes on current metrics behavior.

camilamacedo86: Kubebuilder uses controller runtime, and the metrics exported by default are doc in: https://book.kubebuilder.io/reference/metrics-reference.html.

@varshaprasad96 wdyt?

zqzten: It seems that these reflector metrics are not doced as expected.

camilamacedo86: HI @zqzten, 

If we check it with client-go 0.24 and controller-runtime 0.12.1, we will no longer see these metrics exported?

zqzten: @camilamacedo86 For the first question, yes, they are no longer exported so we could no longer see these metrics. For the last question, yes, users can implement them as they want, but it's meaningless since cliengo is no longer using this metrics provider, the implementation would not be called.

pkg/manager/internal.go
```diff
@@ -633,6 +636,7 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 			},
 		},
 		ReleaseOnCancel: cm.leaderElectionReleaseOnCancel,
+		Name:            cm.leaderElectionID,
```
camilamacedo86, line 639: why we were not passing the id here before? 
were not it required before as well?
why it is required now and what is the relation of it with the proposed changes?

pkg/metrics/client_go_adapter.go
zqzten: There might be some misuderstanding, these metrics are no longer exported since a very early version (k8s 1.14), so this PR won't bring any changes to current exporting metrics.

Speaking of the NOTE, there might be an alternative that we just not remove these code and add a NOTE to say they are no longer exported, which do you prefer?

pkg/manager/internal.go
zqzten: Because the newly enabled clientgo leader election metrics needs the name of the leader election resource, which is the purpose of this parameter.

pkg/metrics/client_go_adapter.go
camilamacedo86: @zqzten,

You are saying:

> There might be some misuderstanding, these metrics are no longer exported since a very early version (k8s 1.14), so this PR won't bring any changes to current exporting metrics

So, how we can check these metrics exported in https://github.com/kubernetes-sigs/controller-runtime/pull/1901#discussion_r879200489?

assafad: just to make things clear, I could view the metrics in https://book.kubebuilder.io/reference/metrics-reference.html, but not the those which removed in https://github.com/kubernetes/kubernetes/pull/74636

camilamacedo86: HI @zqzten,

Sorry, after re-checking it better, I saw that the metrics we are removing here are not the same as those exported. My apologies for do not check this before. 

I think it is all fine and make more sense remove the unuseful code them keep it 👍

camilamacedo86 (approved):
@zqzten,

Thank you for the contribution 🥇 
I updated the first comment to try to make clear the questions made as well. I hope that you like it. However, please feel free to change.

Could you rebase the commits so we can move forward?
Otherwise, it shows for me

/approved

zqzten:
@camilamacedo86 Thanks for the clarification, it looks better👍. Code is rebased.

camilamacedo86 (commented):
/lgtm

zqzten:
Friendly ping @joelanford for approval or further comments

alvaroaleman:
@zqzten please make seperate PRs for removing the metrics and adding new ones so both the changes appear in the changelog

zqzten:
@alvaroaleman Okay, I've seperated the removal to #1946, PTAL.

zqzten:
@alvaroaleman Could you also help approve this PR? It seems that there're no response from other assignees for some time. Thanks!

zqzten:
/remove-lifecycle stale

@alvaroaleman @joelanford Need lgtm approve here, thanks!

pkg/metrics/leaderelection.go
```diff
@@ -0,0 +1,41 @@
...
+		Help: "Gauge of if the reporting system is master of the relevant lease, 0 indicates backup, 1 indicates master. 'name' is the string used to identify the lease. Please make sure to group by name.",
+	}, []string{"name"})
+)
+
+func init() {
+	Registry.MustRegister(leaderGauge)
+
```
alvaroaleman, line 21: Remove the newline please. Other than that, lgtm

pkg/manager/internal.go
alvaroaleman, line 639: TILd `Label values may contain any unicode character`, ref https://prometheus.io/docs/concepts/data_model/  so we don't have to sanitize this

pkg/metrics/leaderelection.go
zqzten: done

alvaroaleman (approved):
Thanks!

PR #1903, opened 2022-05-17
📖 Fix comment on LeaderElectionResourceLock

FillZpp:

Fix comment on LeaderElectionResourceLock

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -142,18 +142,36 @@ type Options struct {
 	LeaderElection bool
 
 	// LeaderElectionResourceLock determines which resource lock to use for leader election,
-	// defaults to "configmapsleases". Change this value only if you know what you are doing.
-	// Otherwise, users of your controller might end up with multiple running instances that
+	// defaults to "leases". Change this value only if you know what you are doing.
+	//
+	// If you are using `configmaps`/`endpoints` resource lock and want to migrate to "leases",
+	// you might do so by migrating to the respective multilock first ("configmapsleases" or "endpointsleases"),
+	// which will acquire a leader lock on both resources.
+	// After all your users have migrated to the multilock, you can go ahead and migrate to "leases".
+	// Please also keep in mind, that users might skip versions of your controller.
+	//
+	// Note: before controller-runtime version v0.7, it was set to "configmaps".
+	// And from v0.7 to v0.11, the default was "configmapsleases", which was
+	// used to migrate from configmaps to leases.
+	// Since the default was "configmapsleases" for over a year, spanning five minor releases,
+	// any actively maintained operators are very likely to have a released version that uses
+	// "configmapsleases". Therefore defaulting to "leases" should be safe since v0.12.
+	//
+	// So, what do you have to do when you are updating your controller-runtime dependency
+	// from a lower version to v0.12 or newer?
+	// - If your operator matches at least one of these conditions:
+	//   - the LeaderElectionResourceLock in your operator has already been explicitly set to "leases"
+	//   - the old controller-runtime version is between v0.7.0 and v0.11.x and the
+	//     LeaderElectionResourceLock wasn't set or was set to "leases"/"configmapsleases"/"endpointsleases"
+	//   feel free to update controller-runtime to v0.12 or newer.
+	// - Otherwise, you may have to take these steps:
+	//   1. update controller-runtime to v0.12 or newer in your go.mod
+	//   2. set LeaderElectionResourceLock to "configmapsleases" (or "endpointsleases")
+	//   3. package your operator and upgrade it in all your clusters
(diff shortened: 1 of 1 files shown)


pkg/manager/manager.go
```diff
@@ -142,18 +142,23 @@ type Options struct {
...
+	// which will acquire a leader lock on both resources.
+	// After all your users have migrated to the multilock, you can go ahead and migrate to "leases".
+	// Please also keep in mind, that users might skip versions of your controller.
+	//
+	// Note: before controller-runtime version v0.7, it was set to "configmaps".
+	// And from v0.7 to v0.11, the default was "configmapsleases", which was
+	// used to migrate from configmaps to leases.
+	// Since the default was "configmapsleases" for over a year, spanning five minor releases,
+	// any actively maintained operators are very likely to have a released version that uses
+	// "configmapsleases". Therefore defaulting to "leases" should be safe since v0.12.
```
camilamacedo86, line 161: I understand that it is safe only for new projects OR for those that properly migrated.
So, my concern here is that those who have operators and just bump C+R might not be aware and do not know the steps. In this way, wdyt about: 
- We add the steps to `migrating to the respective multilock first ("configmapsleases" or "endpointsleases"),`
- What exactly do users must do? Have we an example?

FillZpp: Aha, actually I just optimize the format for this paragraph, you can see the removed lines above. But sure, I'm going to add steps and examples to make the migration more clear.

FillZpp: Aha, actually this paragraph is just from the internal comment where the default value is set.

https://github.com/kubernetes-sigs/controller-runtime/blob/3f265c36d7bf597facc9b0363b48ff21a68cb782/pkg/leaderelection/leader_election.go#L60-L66

But sure, I'm going to add steps and examples to make the migration more clear.

camilamacedo86: I think after we have it in the docs we could highlight it in the release page: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.12.0

FillZpp: > I think after we have it in the docs we could highlight it in the release page: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.12.0

Yeah, makes sense. I have updated the docs again, please see if I have made it clear.

FillZpp:
@camilamacedo86 Hey, would you please take a look again?

pkg/manager/manager.go
```diff
@@ -142,18 +142,36 @@ type Options struct {
...
+	//
+	// So, what do you have to do when you are updating your controller-runtime dependency
+	// from a lower version to v0.12 or newer?
+	// - If your operator matches at least one of these conditions:
+	//   - the LeaderElectionResourceLock in your operator has already been explicitly set to "leases"
+	//   - the old controller-runtime version is between v0.7.0 and v0.11.x and the
```
camilamacedo86, line 167: If the project was scaffold with <, 0.12 does this not mean that the default value is configmapsleases?
Then, would not also be requires to follow the same steps?

FillZpp: Because "configmapsleases" can migrate to "leases" directly. It is designed as a middle state for projects migrating from "configmaps" to "leases".

camilamacedo86: Hi @FillZpp,

I think I properly understood now.
Operators using c+r> 0.7 and <= 0.11 can safely only bump the controller-runtime to 0.12 because their lock is configured with configmapsleases. Therefore from configmapsleases to leases by default not breaking change or bad side effects can be faced. 

Am I right?

WDYT about we clarifies also the issues/bad side effects if someone does not follow the guidance?

camilamacedo86 (approved):
It shows for me 🥇 
Just a nit. 
If possible would be nice if we also describe the bad side effects in case someone does not follows the guidance. Otherwise, it is: 

/approved

pkg/manager/manager.go
FillZpp: > Am I right?

Exactly.

> WDYT about we clarifies also the issues/bad side effects if someone does not follow the guidance?

Sure, I have moved the bad side effects under the steps.

camilamacedo86: thank you for the clarifications

FillZpp:
@camilamacedo86 Could you approve this again?

camilamacedo86 (approved):
/approved

camilamacedo86:
@FillZpp I have permissions to lgtm only. We will need a second reviewer here to add the approved label.

FillZpp:
> @FillZpp I have permissions to lgtm only. We will need a second reviewer here to add the approved label.

Okey. @alvaroaleman  Could you please take a look?

PR #1906, opened 2022-05-18
:sparkles: Add WebhookWrapper to enable wrapping the webhook HTTP handler

adammw:
This PR adds a WebhookWrapper to enable wrapping the webhook HTTP handler.

The use case is to enable tracing/middleware to be configured on the webhook server, for example:

```go
import (
  "net/http"
  "github.com/opentracing/opentracing-go"
  "github.com/opentracing-contrib/go-stdlib/nethttp"
  "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentracer"
  "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
)

func main() {
  t := opentracer.New(tracer.WithServiceName(api.OperatorName))
  defer tracer.Stop()
  opentracing.SetGlobalTracer(t)

  webhook := &webhook.Server{
    WebhookWrapper: func(handler http.Handler) (http.Handler, err) {
      return nethttp.Middleware(t, handler), nil
    }
  }
    WebhookServer: webhook
  })
}
```

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -82,6 +82,9 @@ type Server struct {
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
 
+	// WebhookWrapper is a custom wrapper func that enables adding middleware onto the webhook mux.
+	WebhookWrapper func(http.Handler) (http.Handler, error)
+
 	// webhooks keep track of all registered webhooks for dependency injection,
 	// and to provide better panic messages on duplicate webhook registration.
 	webhooks map[string]http.Handler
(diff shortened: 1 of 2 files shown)


FillZpp (commented):
Based on what I comment below, how about add a field like `WebhookWrapper` instead of another handler?

```golang
type Server struct {
    // ...
    WebhookWrapper func(http.Handler) http.Handler
}

func (s *Server) Start(ctx context.Context) error {
    // ...
    handler := s.WebhookMux
    if s.WebhookWrapper != nil {
        handler = s.WebhookWrapper(handler)
    }
    srv := httpserver.New(handler)
}
```

And it can also be `WebhookWrapper func(http.Handler) (http.Handler, error)` if the wrapper may get some errors.

adammw:
@FillZpp updated to your suggested implementation

pkg/webhook/server.go
```diff
@@ -82,6 +82,9 @@ type Server struct {
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
 
+	// WebhookWrapper is a custom wrapper func that enables adding middleware onto the webhook mux.
+	WebhookWrapper func(http.Handler) (http.Handler, error)
```
alvaroaleman, line 86: Why not name it like what it is, namely a `Middleware`?

Also, what is the reason for the error return here? I realize this could error during request handling, but what would be a reason where it could error here?

adammw: The error was just to make it flexible in case whatever setup function there is could fail and therefore bubble up the error to the Start() method. Note that this isn't error of request handling but error of setting up the request handler/middleware. 

Not sure if there's a common signature for middleware - hence why I went with wrapper but can change if there's agreement

PR #1910, opened 2022-05-18
:bug: Fix issue with starting multiple test envs

### LINKED ISSUE #1909: Starting multiple test environments can fail due to multiple trying to delete the same port file
#### Background

* When [starting a single  test environment](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/envtest/server.go#L197), the [control plane start-up logic](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/internal/testing/controlplane/plane.go#L49) launches an etcd and kube-apiserver process. 
* The start-up logic needs to bind these processes to multiple ports, and when none are configured in advance, the logic looks for unused ports, and then creates files in the user cache directory to reserve them ([example](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/internal/testing/controlplane/etcd.go#L120)).

#### Problem
* The same logic used to create files to reserve ports also has logic that [deletes files corresponding to "outdated" ports](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/internal/testing/addr/manager.go#L64).
* Therefore, if one were to start multiple test environments in parallel (e.g. in different processes or threads), then some may fail to start since they may end up trying to delete the same file that others have already deleted.
* In particular, they may fail to start due to the following error:

    ```
    unable to start control plane itself: failed to start the controlplane. retried 5 times: lstat /usr/local/home/jcanseco/.cache/kubebuilder-envtest/./port-42369: no such file or directory
    ```
    ...which comes from [this line](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/internal/testing/addr/manager.go#L74).
    ```
    unable to start control plane itself: failed to start the controlplane. retried 5 times: remove /usr/local/home/jcanseco/.cache/kubebuilder-envtest/port-36781: no such file or directory
    ```
    ...which comes from [this line](https://github.com/kubernetes-sigs/controller-runtime/blob/7fb8534b80339a5b8158cee3dc9e5f9439d679dc/pkg/internal/testing/addr/manager.go#L78).

#### Potential Solution

I have verified that this can be fixed by adding logic along the lines of "don't error out if trying to delete a file that no longer exists".

jcanseco:
Fixes #1909

### DIFF
--- a/pkg/internal/testing/addr/manager.go
+++ b/pkg/internal/testing/addr/manager.go
@@ -71,10 +71,20 @@ func (c *portCache) add(port int) (bool, error) {
 		}
 		info, err := d.Info()
 		if err != nil {
+			// No-op if file no longer exists; may have been deleted by another
+			// process/thread trying to allocate ports.
+			if errors.Is(err, fs.ErrNotExist) {
+				return nil
+			}
 			return err
 		}
 		if time.Since(info.ModTime()) > portReserveTime {
 			if err := os.Remove(filepath.Join(cacheDir, path)); err != nil {
+				// No-op if file no longer exists; may have been deleted by another
+				// process/thread trying to allocate ports.
+				if os.IsNotExist(err) {
+					return nil
+				}
 				return err
 			}
 		}


varshaprasad96:
/ok-to-test 
@jcanseco thanks for the PR, I think this does seem to be good.
/lgtm

alvaroaleman (approved):
@jcanseco thank you!

If possible, please add the commit description to the PR body next time.

jcanseco:
Thank you! Do we know when this PR will be released?

alvaroaleman:
I can cherrypick it onto the release branch for you, but we do not currently have a formalized release cadence

/cherrypick release-0.12

PR #1917, opened 2022-05-30
⚠ Add GetOptions as optional argument of client.Reader and all its implementation

### LINKED ISSUE #1915: Add GetOption in client get call
Add GetOption so client can work with `metav1.GetOptions`

FillZpp:

Add `GetOptions` as optional argument of client.Reader and all its implementation, so that ppl can set fields of `metav1.GetOptions` like `resourceVersion="0"` when they are directly getting from apiserver.

fixes #1915

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -50,7 +50,7 @@ type Reader interface {
 	// Get retrieves an obj for the given object key from the Kubernetes Cluster.
 	// obj must be a struct pointer so that obj can be updated with the response
 	// returned by the Server.
-	Get(ctx context.Context, key ObjectKey, obj Object) error
+	Get(ctx context.Context, key ObjectKey, obj Object, opts ...GetOption) error
 
 	// List retrieves list of objects for a given namespace and list options. On a
 	// successful call, Items field in the list will be populated with the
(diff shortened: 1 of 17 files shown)


pkg/client/interfaces.go
```diff
@@ -50,7 +50,7 @@ type Reader interface {
 	// Get retrieves an obj for the given object key from the Kubernetes Cluster.
 	// obj must be a struct pointer so that obj can be updated with the response
 	// returned by the Server.
-	Get(ctx context.Context, key ObjectKey, obj Object) error
+	Get(ctx context.Context, key ObjectKey, obj Object, opts ...GetOption) error
```
alvaroaleman, line 53: Generally the change looks good to me, but it needs to be marked as breaking since we change the interface here. It is very likely that ppl have fake implementations of the client that need changing to adhere to the new signature

FillZpp:
/retitle ⚠ Add GetOptions as optional argument of client.Reader and all its implementation

pkg/client/interfaces.go
FillZpp: Sure, I have updated the title. Do you think it is necessary to add the breaking warning into code comments?

alvaroaleman (approved):
thanks!

PR #1919, opened 2022-05-31
:warning: GIT-1918: enable leader election callback options

### LINKED ISSUE #1918: Enable additional options for Leader Election Callback
The current implementation of controller-runtime supports setting up a `OnStartLeading` callback in a not so direct way. (I am not complaining about this at all BTW). 

https://github.com/kubernetes-sigs/controller-runtime/blob/c7a98aa706379c4e5c79ea675c7f333192677971/pkg/manager/internal.go#L615-L621

This needs you to implement a `Runnable`

https://github.com/kubernetes-sigs/controller-runtime/blob/c7a98aa706379c4e5c79ea675c7f333192677971/pkg/manager/manager.go#L293-L298

And then https://github.com/kubernetes-sigs/controller-runtime/blob/c7a98aa706379c4e5c79ea675c7f333192677971/pkg/manager/manager.go#L311-L315. 

Then you can register a runnable under the https://github.com/kubernetes-sigs/controller-runtime/blob/b9219528d95974cb4f5b06f86c9b1c9b7d3045a5/pkg/manager/runnable_group.go#L53-L69. 

This works perfectly. However, this lacks a few other options provided by the `leaderelection` config

```go
type LeaderCallbacks struct {
	
	OnStartedLeading func(context.Context)
	
	OnStoppedLeading func()
	
	OnNewLeader func(identity string)
}
```

The handler for `OnStoppedLeading` exists but It is not entirely extendable to get a custom set of operations performed. 

https://github.com/kubernetes-sigs/controller-runtime/blob/196828e54e4210497438671b2b449522c004db5c/pkg/manager/internal.go#L622-L633

It can be super useful to provide a way to extend and hook into these events as well in the controller runtime like the current one provided.  This can be really useful in performing certain cleanup operations that were done when you start leading to make sure you make way for the next controller to process things. 

If this is an acceptable thing to add, Please let me know and I will be more than happy to open a PR with the changes.

harshanarayana:
Closes #1918 

This PR does the following. 

1. In order to retain the backward compatibility, this introduces a new interface named `ManagerWithLeaderCallbacks` that supports configuring the leader callbacks via the manager
2. A manager.Options that can be used to configure the global callbacks un-conditionally if required
3. Modify the leader election workflow to enable `OnNewLeader` callbacks
4. Introduce a new `NewManagerWithLeaderCallback` alias for creating a new Manager with callback support

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -106,6 +106,9 @@ var (
 	// NewManager returns a new Manager for creating Controllers.
 	NewManager = manager.New
 
+	// NewManagerWithLeaderCallback returns a new ManagerWithLeaderCallback for creating controllers with leader callbacks.
+	NewManagerWithLeaderCallback = manager.NewWithLeaderElectionCallback
+
 	// CreateOrUpdate creates or updates the given object obj in the Kubernetes
 	// cluster. The object's desired state should be reconciled with the existing
 	// state using the passed in ReconcileFn. obj must be a struct pointer so that
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -170,11 +170,32 @@ type controllerManager struct {
 	// internalProceduresStop channel is used internally to the manager when coordinating
 	// the proper shutdown of servers. This channel is also used for dependency injection.
 	internalProceduresStop chan struct{}
+
+	// onStoppedLeadingCallback provides a way to configure a set of helper functions
+	// that are invoked when the current leader stops leading by losing the election or failing
+	// to renew the lock.
+	onStoppedLeadingCallback []func() error
+
+	// onNewLeaderCallback provides a way to configure a set of handlers to be invoked when the
+	// current leader changes.
+	onNewLeaderCallback []func(string) error
 }
 
 type hasCache interface {
 	Runnable
 	GetCache() cache.Cache
+}
+
+// AddOnStoppedLeadingCallback provides a way to inject a callable function into the controllerManager that
+// will be triggered in case if the current leader stops leading.
+func (cm *controllerManager) AddOnStoppedLeadingCallback(callback func() error) {
+	cm.onStoppedLeadingCallback = append(cm.onStoppedLeadingCallback, callback)
+}
+
+// AddOnNewLeaderCallback provides a way to inject a callable function into the controllerManager that
+// will be triggered in case if the current leader changes and a new leader is elected with the identity.
+func (cm *controllerManager) AddOnNewLeaderCallback(callback func(identity string) error) {
+	cm.onNewLeaderCallback = append(cm.onNewLeaderCallback, callback)
 }
 
 // Add sets dependencies on i, and adds it to the list of Runnables to start.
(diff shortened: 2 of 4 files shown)


pkg/manager/manager.go
```diff
@@ -204,6 +212,15 @@ type Options struct {
...
 	RetryPeriod *time.Duration
 
+	// OnStoppedLeadingCallback provides a way to configure a set of helper functions
+	// that are invoked when the current leader stops leading by losing the election or failing
+	// to renew the lock.
+	OnStoppedLeadingCallback []func() error
+
+	// OnNewLeaderCallback provides a way to configure a set of handlers to be invoked when the
+	// current leader changes.
+	OnNewLeaderCallback []func(string) error
```
FillZpp, line 222: Why do you need the callback for OnNewLeader? Usually there is no such condition that one leader will be force migrated to another. Even after a leader lost its election, another one still has to wait for `LeaseDuration` to become a new leader.

pkg/manager/manager.go
```diff
@@ -204,6 +212,15 @@ type Options struct {
 	// between tries of actions. Default is 2 seconds.
 	RetryPeriod *time.Duration
 
+	// OnStoppedLeadingCallback provides a way to configure a set of helper functions
+	// that are invoked when the current leader stops leading by losing the election or failing
+	// to renew the lock.
+	OnStoppedLeadingCallback []func() error
```
FillZpp, line 218: Do we really need the function list? Maybe one callback for users can hold multiple code block or functions calling in it.

pkg/manager/internal.go
```diff
@@ -623,13 +644,34 @@ func (cm *controllerManager) startLeaderElection(ctx context.Context) (err error
 				if cm.onStoppedLeading != nil {
 					cm.onStoppedLeading()
 				}
+				var retErrors = []error{
+					errors.New("leader election lost"),
+				}
+
+				for _, cb := range cm.onStoppedLeadingCallback {
+					if err := cb(); err != nil {
+						retErrors = append(retErrors, err)
```
FillZpp, line 653: 1. What's this error for? Even we put it into `errChan`, it will only be logged for the Start error. Maybe just let the callback itself print the error log?
2. Can we just reuse `cm.onStoppedLeading` since it is only used for test now?

alias.go
```diff
@@ -106,6 +106,9 @@ var (
 	// NewManager returns a new Manager for creating Controllers.
 	NewManager = manager.New
 
+	// NewManagerWithLeaderCallback returns a new ManagerWithLeaderCallback for creating controllers with leader callbacks
+	NewManagerWithLeaderCallback = manager.NewWithLeaderElectionCallback
```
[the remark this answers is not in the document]
harshanarayana: The reason why I am not doing just `manager.Options` change alone is kind of based on how we use the runtime today. 

We have an entry point script that creates a `manager` instances and passes that around to a while bunch of other packages that contain Reconcilers. Each reconciler registers itself against the manager using `func (r *CRDReconciler) SetupWithManager(mgr ctrl.Manager) error {` handler. 

In different places on this registration workflow, we introduce certain leader elect helpers using `manager.Add` behaviour. This gets incredibly user friendly to programatically manage the components in package groups. 

Moving that to the `manager.Options` would mean you have to bring all of these things to the entry point where the `manager` instance is getting created. 

Would it not be better to provide a behavior similar to `manager.Add` for setting up the Leader election callbacks as well  in different places before the controller manager is actually started ?

pkg/manager/manager.go
harshanarayana: `OnNewLeader` is triggered when the observed state of the leader changes right ? This can handy in performing certain event driven workflow in multi node clusters. The node that becomes the leader will get a callback with the start leading callback. And when it loses, it will get an event for when the leader is lost. The `OnNewLeader` is invoked with following condition as well.
> This includes the first observed leader when the client starts.

So this can be handy to recover cases where pod crashed or restarted trying to address the Stopped leading event. It can use this call back to set the state of the world accordingly. 

Also, this provides a completeness against the Leader callback handlers.

pkg/manager/manager.go
harshanarayana: Keeping them as an array so that the registration for this callback can be done in more than one place. It is not always easy to aggregate all the callbacks in one uber callback from UT perspective right ? So, the provision to register multiple smaller callbacks programatically from different places in the code base before the `Manager` instance can be created.

pkg/manager/internal.go
harshanarayana: 1. //ACK. Sounds good to me. The idea however was to be able to aggregate and display all the errors in one place with regards to handling the leader election lost item. But I have no issue with callback handlers logging it and forgetting about it instead. 
2. Due to https://github.com/kubernetes-sigs/controller-runtime/pull/1919#discussion_r885550142 I decided to keep the callback for stopped leading in a different attribute.

Issue #452, opened 2019-05-29
[feature request] Support Pod expansion in Client


gaocegege:
Hi I am one of the maintainers @kubeflow 

We are using kubebuilder to implement some CRDs for Machine Learning/Deep Learning workloads. We meet some problems with the client in controller-runtime.

We need to get the logs of pods using the client, But It seems that controller-runtime client has a higher level abstraction on top of Kubernetes client. Do we have a plan to support expansion (e.g. https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod_expansion.go) for the client? Or is there any suggestion to do it with controller-runtime client?

Now we have to create a Kubernetes client and use this to get pod logs.

I'd appreciate it if anyone could help me.

Thanks :clinking_glasses: :beers:

DirectXMan12:
We had some plans for supporting subresources, but haven't gotten around to them yet.  If you put forth a proposed design, I'd be happy to take a look.  We'd have to add support for the wonky streaming that logs uses, since right now the client just deals with CRUD.

DirectXMan12:
So, we'll get generic subresource support eventually, but non-CRUD subresources need to be handled specially

DirectXMan12:
cc #172  (which is mainly for CRUD subresources ATM)

imilchev:
I am also interested in this

27149chen:
is there any workarouds?

coderanger:
> is there any workarouds?

You can use client-go instead.

gaocegege:
> You can use client-go instead.

It is hard to keep two clients, I think. It will be better if we support it in controller-runtime.

acechef:
Is there any progress on this issue?

benjaminjb:
👀

jasiu001:
Is it possible or still in the design phase?

sbueringer:
Nothing changed.

matheuscscp:
+1

songtao98:
+1

sbueringer:
Just fyi, the only thing that `+1` comments achieve is annoying everyone who is watching this issue for relevant updates :)
Please use the 👍 emoji on https://github.com/kubernetes-sigs/controller-runtime/issues/452#issue-449697241 instead.

PR #1922, opened 2022-06-02
:sparkles: Support subresource modification

g7r:
I need to add finalizers without granting access to the entire resource. It certainly is possible in Kubernetes:
```YAML
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: myrole
rules:
  - apiGroups:
      - apps
    resources:
      - deployments
    verbs:
      - get
      - watch
      - list

  - apiGroups:
      - apps
    resources:
      - deployments/finalizers
    verbs:
      - update
```

But I can't update finalizers using `client.Client`. I don't have a permission to modify entire deployment and `client.Client` doesn't provide an API to modify generic subresource, it only has special support for "status" subresource. In this PR I add support to other subresources.

I'm open to suggestions of course. Consider the initial PR version also as an invitation to discussion even if the implementation is totally wrong.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -289,40 +289,45 @@ func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) e
 }
 
 // Status implements client.StatusClient.
-func (c *client) Status() StatusWriter {
-	return &statusWriter{client: c}
+func (c *client) Status() SubResourceWriter {
+	return &subResourceWriter{client: c, subResource: "status"}
 }
 
-// statusWriter is client.StatusWriter that writes status subresource.
-type statusWriter struct {
-	client *client
+func (c *client) SubResource(subResource string) SubResourceWriter {
+	return &subResourceWriter{client: c, subResource: subResource}
 }
 
-// ensure statusWriter implements client.StatusWriter.
-var _ StatusWriter = &statusWriter{}
+// subResourceWriter is client.SubResourceWriter that writes to subresources.
+type subResourceWriter struct {
+	client      *client
+	subResource string
+}
 
-// Update implements client.StatusWriter.
-func (sw *statusWriter) Update(ctx context.Context, obj Object, opts ...UpdateOption) error {
+// ensure subResourceWriter implements client.SubResourceWriter.
+var _ SubResourceWriter = &subResourceWriter{}
+
+// Update implements client.SubResourceWriter.
+func (sw *subResourceWriter) Update(ctx context.Context, obj Object, opts ...UpdateOption) error {
 	defer sw.client.resetGroupVersionKind(obj, obj.GetObjectKind().GroupVersionKind())
 	switch obj.(type) {
 	case *unstructured.Unstructured:
-		return sw.client.unstructuredClient.UpdateStatus(ctx, obj, opts...)
+		return sw.client.unstructuredClient.UpdateSubResource(ctx, obj, sw.subResource, opts...)
 	case *metav1.PartialObjectMetadata:
 		return fmt.Errorf("cannot update status using only metadata -- did you mean to patch?")
 	default:
-		return sw.client.typedClient.UpdateStatus(ctx, obj, opts...)
+		return sw.client.typedClient.UpdateSubResource(ctx, obj, sw.subResource, opts...)
 	}
 }
 
-// Patch implements client.Client.
-func (sw *statusWriter) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error {
+// Patch implements client.SubResourceWriter.
+func (sw *subResourceWriter) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error {
 	defer sw.client.resetGroupVersionKind(obj, obj.GetObjectKind().GroupVersionKind())
 	switch obj.(type) {
 	case *unstructured.Unstructured:
-		return sw.client.unstructuredClient.PatchStatus(ctx, obj, patch, opts...)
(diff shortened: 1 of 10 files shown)


pkg/client/client.go
```diff
@@ -289,40 +289,45 @@ func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) e
 }
 
 // Status implements client.StatusClient.
-func (c *client) Status() StatusWriter {
-	return &statusWriter{client: c}
+func (c *client) Status() SubResourceWriter {
+	return &subResourceWriter{client: c, subResource: "status"}
```
joelanford, line 293: Does this work?
```suggestion
	return SubResource("status")
```

pkg/client/interfaces.go
```diff
@@ -81,11 +81,20 @@ type Writer interface {
...
 type StatusClient interface {
-	Status() StatusWriter
+	Status() SubResourceWriter
 }
 
-// StatusWriter knows how to update status subresource of a Kubernetes object.
-type StatusWriter interface {
+// SubResourceClient knows how to create a client which can update subresource
+// for kubernetes objects.
+type SubResourceClient interface {
```
alvaroaleman, line 89: One thing I am wondering is if this should maybe be called something like `SubResourceCRUDClient` or some such to clarify that this can not deal with other subresources like logs and differentiate it from a potential future non-crud subresource client?

I am not sure though if it is possible to come up with a sensible non-crud subresource client, since that could basically be anything? Maybe that one would just be a standard http client? Ref https://github.com/kubernetes-sigs/controller-runtime/issues/452

pkg/client/typed_client.go
```diff
@@ -159,8 +159,8 @@ func (c *typedClient) List(ctx context.Context, obj ObjectList, opts ...ListOpti
 		Into(obj)
 }
 
-// UpdateStatus used by StatusWriter to write status.
-func (c *typedClient) UpdateStatus(ctx context.Context, obj Object, opts ...UpdateOption) error {
+// UpdateSubResource used by SubResourceWriter to write status.
+func (c *typedClient) UpdateSubResource(ctx context.Context, obj Object, subResource string, opts ...UpdateOption) error {
```
FillZpp, line 163: IMHO, one `obj Object` in the parameters is not enough, because not all subresources (even CRUD subresources) take the original object as its request body. For subresources like `Status` and `Finalize`, yes. For subresources like `Scale`, no, which actually accept a `autoscaling.Scale` as the request body.

FillZpp (commented):
Some comments to this *CRUD* subresource client. It's nice that we can read/write more subresources in controller-runtime.

Also a non-CRUD subresource client will be great, but it is not easy to define a common interface for those subresources. It seems they are all defined for K8s built-in resources and don't support custom resources (I'm not sure). So maybe document that those could only be handled via client-go temporarily? WDYT @alvaroaleman

alvaroaleman:
@g7r do you have the cycles to address the feedback on this?

pkg/client/interfaces.go
alvaroaleman: Thinking about it it again, maybe we should instead just add more verb(s) once we get there?

pkg/client/interfaces.go
```diff
@@ -81,11 +81,20 @@ type Writer interface {
...
+	SubResource(subResource string) SubResourceWriter
+}
+
+// StatusWriter is kept for backward compatibility.
+type StatusWriter = SubResourceWriter
+
+// SubResourceWriter knows how to update subresource of a Kubernetes object.
+type SubResourceWriter interface {
```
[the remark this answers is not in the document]
alvaroaleman: We could probably in the future replace the currrent SubResourceWriter interface with a SubResourceClient and then add read functionality in there?

pkg/client/typed_client.go
alvaroaleman: Yeah, we definitely need that, apart from Scale there is also `authenticationv1.TokenRequest` in the case of serviceaccount tokens...Maybe we can have a `StatusUpdateOption` that has a `Body Object` field on top of the other `UpdateOptions`?

pkg/client/interfaces.go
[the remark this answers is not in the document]
FillZpp: Or we could have both SubResourceReader and SubResourceWriter, like the Reader and Writer?

And also the SubResourceWriter needs more methods other than Update and Patch, for example the subresource `pods/binding` has to be POST, which is actually Create.

pkg/client/typed_client.go
FillZpp: SGTM, or we can have two objects in the UpdateSubResource parameters like
`(ctx context.Context, obj Object, subObj Object, subResource string, opts ...UpdateOption)`

If the `subObj` is not nil, we should take it as the update body, otherwise we will use the `obj`.

pkg/client/interfaces.go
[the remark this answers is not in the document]
FillZpp: Or we could have both SubResourceReader and SubResourceWriter, like the Reader and Writer?

And also the SubResourceWriter needs more methods other than Update and Patch, for example the subresource pods/binding has to be POST, which is actually Create.

pkg/client/typed_client.go
alvaroaleman: Due to the fact that it is not always mandatory I'd prefer to make it part of the opts and we can then provide a convenient wrapper, so that you can do something like:

`c.UpdateSubresource(ctx, myObj, "my-subresource", client.WithSubResourceBody(mySubResource))`

PR #1935, opened 2022-06-20
✨  Bump k8s.io packages to v1.24.2

FriedrichWilken:
This bumps the patch level of the `k8s.io` packages `api`, `apiextensions-apiserver`, `apimachinery`, `client-go` and `component-base` to `v0.24.2`.

PR #1937, opened 2022-06-22
✨ Bump dependencies

FriedrichWilken:
This is a bump of all outdated dependencies but the k8s.io packages (there is [another PR](https://github.com/kubernetes-sigs/controller-runtime/pull/1935) for that).

The goal of this PR is to prevent potential security vulnerabilities and increase performance.


go.mod
```diff
@@ -3,19 +3,19 @@ module sigs.k8s.io/controller-runtime
 go 1.17
 
 require (
-	github.com/evanphx/json-patch v4.12.0+incompatible
-	github.com/fsnotify/fsnotify v1.5.1
-	github.com/go-logr/logr v1.2.0
-	github.com/go-logr/zapr v1.2.0
-	github.com/google/go-cmp v0.5.5
+	github.com/evanphx/json-patch v5.6.0+incompatible
```
sbueringer, line 6:
```suggestion
	github.com/evanphx/json-patch/v5 v5.6.0
```
I think you have to import it like this. They started using module path correctly with one of the latest releases IIRC (+ update the imports accordingly)

sbueringer (commented):
One nit

go.mod
FriedrichWilken: Thanks, I wasn't aware. I added a commit. :)

sbueringer: Thx!

sbueringer:
Thx!
not sure what the considerations in this repo are for squashing the commits (or if Prow is run in merge-mode), but apart from that
/lgtm
from my side

go.mod
sbueringer: I had this issue in CAPI when bumping. I think otherwise you will end up with the code in the root folder of the repo which is afaik some v4 version :)

alvaroaleman (approved):

PR #1939, opened 2022-06-23
📖 nit: fix in client package documentation

rpriyanshu9:
Just a minor nit fix in the documentation. 
This makes the documentation (https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client) more readable.

### DIFF
--- a/pkg/client/doc.go
+++ b/pkg/client/doc.go
@@ -25,7 +25,7 @@ limitations under the License.
 // The New function can be used to create a new client that talks directly
 // to the API server.
 //
-// A common pattern in Kubernetes to read from a cache and write to the API
+// It is a common pattern in Kubernetes to read from a cache and write to the API
 // server.  This pattern is covered by the DelegatingClient type, which can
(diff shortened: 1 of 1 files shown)


pkg/client/doc.go
```diff
@@ -25,7 +25,7 @@ limitations under the License.
 // The New function can be used to create a new client that talks directly
 // to the API server.
 //
-// A common pattern in Kubernetes to read from a cache and write to the API
+// A common pattern in Kubernetes is to read from a cache and write to the API
```
FillZpp, line 28: Emm.. Actually I don't think `is` should be added here.

FillZpp: Or it might be `It is a common pattern in Kubernetes to ...`

rpriyanshu9:
Hey, @sbueringer can you PTAL again? 🙂 I added another commit which seems to have removed the _lgtm_ label.

alvaroaleman (approved):

PR #1943, opened 2022-06-25
✨ Introduce pprof server to manager

### LINKED ISSUE #1779: [feature request] Add support for pprof
# Add support for serving pprof data

It would be nice to extend manager with option to enable pprof when requested, I found similar issue in kubebuilder suggesting it should be added in controller-runtime

Related issues :
 - https://github.com/kubernetes-sigs/kubebuilder/issues/1225

zqzten:
This PR introduces a configurable standalone server for serving pprof to manager, just like the health and metrics server.

fixes #1779 

Additional thoughts:
I've been aware of `metricsExtraHandlers` introduced in #824 which might be an alternative for end users to serve pprof, but I don't think it's quite appropriate to do so, mainly for these two reasons:
1. pprof is a language built-in feature which is widely used, so it's more appropriate to make its server built-in in the library rather than let end users construct and run themselves in every controller.
2. `metricsExtraHandlers` bind all handlers to metrics listener which is not that general and flexible, it cannot meet cases that user want to expose metrics and pprof by different listeners (common use case: a 0.0.0.0: listener for serving metrics and a 127.0.0.1: listener for serving pprof).

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -105,6 +106,9 @@ type controllerManager struct {
 
 	// Healthz probe handler
 	healthzHandler *healthz.Handler
+
+	// pprofListener is used to serve pprof
+	pprofListener net.Listener
 
 	// controllerConfig are the global controller options.
 	controllerConfig config.Controller
@@ -343,6 +347,24 @@ func (cm *controllerManager) serveHealthProbes() {
 	go cm.httpServe("health probe", cm.logger, server, cm.healthProbeListener)
 }
 
+func (cm *controllerManager) addPprofServer() error {
+	mux := http.NewServeMux()
+	srv := httpserver.New(mux)
+
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	return cm.add(&server{
+		Kind:     "pprof",
+		Log:      cm.logger,
+		Server:   srv,
+		Listener: cm.pprofListener,
+	})
+}
+
 func (cm *controllerManager) httpServe(kind string, log logr.Logger, server *http.Server, ln net.Listener) {
 	log = log.WithValues("kind", kind, "addr", ln.Addr())
 
@@ -438,6 +460,13 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 	// Serve health probes.
 	if cm.healthProbeListener != nil {
 		cm.serveHealthProbes()
+	}
+
+	// Add pprof server
+	if cm.pprofListener != nil {
+		if err := cm.addPprofServer(); err != nil {
+			return fmt.Errorf("failed to add pprof server: %w", err)
+		}
 	}
 
 	// First start any webhook servers, which includes conversion, validation, and defaulting
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -233,6 +233,13 @@ type Options struct {
 	// Liveness probe endpoint name, defaults to "healthz"
 	LivenessEndpointName string
 
+	// PprofBindAddress is the TCP address that the controller should bind to
+	// for serving pprof.
+	// It can be set to "" or "0" to disable the pprof serving.
+	// Since pprof may contain sensitive information, make sure to protect it
+	// before exposing it to public.
+	PprofBindAddress string
+
 	// Port is the port that the webhook server serves at.
 	// It is used to set webhook.Server.Port if WebhookServer is not set.
 	Port int
@@ -308,6 +315,7 @@ type Options struct {
 	newResourceLock        func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error)
 	newMetricsListener     func(addr string) (net.Listener, error)
 	newHealthProbeListener func(addr string) (net.Listener, error)
+	newPprofListener       func(addr string) (net.Listener, error)
 }
 
 // BaseContextFunc is a function used to provide a base Context to Runnables
@@ -417,6 +425,13 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		return nil, err
 	}
 
+	// Create pprof listener. This will throw an error if the bind
+	// address is invalid or already in use.
+	pprofListener, err := options.newPprofListener(options.PprofBindAddress)
+	if err != nil {
+		return nil, fmt.Errorf("failed to new pprof listener: %w", err)
+	}
+
 	errChan := make(chan error)
 	runnables := newRunnables(options.BaseContext, errChan)
 
@@ -444,6 +459,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		healthProbeListener:           healthProbeListener,
 		readinessEndpointName:         options.ReadinessEndpointName,
 		livenessEndpointName:          options.LivenessEndpointName,
+		pprofListener:                 pprofListener,
 		gracefulShutdownTimeout:       *options.GracefulShutdownTimeout,
 		internalProceduresStop:        make(chan struct{}),
 		leaderElectionStopped:         make(chan struct{}),
@@ -574,6 +590,19 @@ func defaultHealthProbeListener(addr string) (net.Listener, error) {
 	return ln, nil
 }
 
+// defaultPprofListener creates the default pprof listener bound to the given address.
+func defaultPprofListener(addr string) (net.Listener, error) {
+	if addr == "" || addr == "0" {
+		return nil, nil
+	}
+
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("error listening on %s: %w", addr, err)
+	}
+	return ln, nil
+}
+
 // defaultBaseContext is used as the BaseContext value in Options if one
 // has not already been set.
 func defaultBaseContext() context.Context {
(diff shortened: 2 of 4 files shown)


pkg/manager/internal.go
```diff
@@ -355,6 +359,19 @@ func (cm *controllerManager) serveHealthProbes() {
...
 
+func (cm *controllerManager) servePprof() {
+	mux := http.NewServeMux()
+	server := httpserver.New(mux)
+
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
varshaprasad96, line 370: This is really interesting and helpful. However, if we support this, how we allow users to expose custom profiles?

zqzten: Users can access their custom profiles through `/debug/pprof/*` which is already exposed by `mux.HandleFunc("/debug/pprof/", pprof.Index)`.

pkg/manager/manager.go
```diff
@@ -227,6 +227,10 @@ type Options struct {
 	// Liveness probe endpoint name, defaults to "healthz"
 	LivenessEndpointName string
 
+	// PprofBindAddress is the TCP address that the controller should bind to
+	// for serving pprof
+	PprofBindAddress string
```
FillZpp, line 232:
```golang
// Pprof is disabled if it is "" or "0".
```

AlmogBaku:
TBH, I'm against this one. This is very tempting to add functionality such as this to your solution, but enabling it on production might cause a *sensitive information leakage*.

I suggest adding a "hint" in your "DEVELOPMENT.MD" file as I did here:
https://github.com/natun-ai/natun/blob/master/CONTRIBUTING.md#profiling

FillZpp:
> This is very tempting to add functionality such as this to your solution, but enabling it on production might cause a sensitive information leakage.

That's true if we enable pprof and bind it to `0.0.0.0:<port>` on production. But I think this PR is ok, since it just provides an option which is disabled by default and we can add doc comments to recommend ppl only bind it to localhost.

zqzten:
@AlmogBaku Thanks for the security remind, but I don't think it as a blocking issue as we have ways to protect it:
* By binding its listener to `127.0.0.1:`, as @FillZpp suggested, which is a very common and naive solution. This is also why seperated listener is introduced instead of reusing `metricsExtraHandlers`.
* By protecting pprof server by [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy), which is also the default solution to [protect metrics server in kubebuilder](https://book.kubebuilder.io/reference/metrics.html). This is a sophisticated solution which is similarly used by kube-controller-manager (it does the rbac proxy by itself though).

Security is very important indeed, I'll add some doc comments on it, also we can put best practice usage in kubebuilder after this feature is added.

AlmogBaku:
what about adding a build tag as a solution?

zqzten:
> what about adding a build tag as a solution?

I'm not sure whether this solution is appropriate in this project. Let's just wait for opinions from other maintainers.

AlmogBaku:
Exposing this setting via build tag is like having a “remote debugger” - possible, but should be allowed implicitly and temporary on production mode

pkg/manager/internal.go
```diff
@@ -355,6 +359,19 @@ func (cm *controllerManager) serveHealthProbes() {
 	go cm.httpServe("health probe", cm.logger, server, cm.healthProbeListener)
 }
 
+func (cm *controllerManager) servePprof() {
```
alvaroaleman, line 362: This should be a normal `Runnable` and not a method on the controllerManager

AlmogBaku: also, don't forget to implement the runnable with `NeedLeaderElection() bool -> false` so it'll run also for non-leader instances

zqzten: > This should be a normal `Runnable` and not a method on the controllerManager

Sorry I don't quite get the point of it. As internal http servers, there're `serveMetrics` and `serveHealthProbes` methods in the manager, is there any difference between them and the pprof server?

zqzten: @alvaroaleman It would be grateful if you can give some further explanation on why it should be a normal `Runnable` rather than an internal http server like the metrics and health ones, thanks.

zqzten:
@alvaroaleman Mind take a look at the question in the review above? Thanks!

pkg/config/v1alpha1/types.go
```diff
@@ -68,6 +68,10 @@ type ControllerManagerConfigurationSpec struct {
 	// +optional
 	Health ControllerHealth `json:"health,omitempty"`
 
+	// Pprof contains the controller pprof configuration
+	// +optional
+	Pprof ControllerPprof `json:"pprof,omitempty"`
```
vincepri, line 73:
```suggestion
	// Profiling contains the controller profiling configuration
	// +optional
	Profiling ControllerProfiling `json:"profiling,omitempty"`
```

pkg/config/v1alpha1/types.go
```diff
@@ -121,6 +125,17 @@ type ControllerHealth struct {
 	LivenessEndpointName string `json:"livenessEndpointName,omitempty"`
 }
 
+// ControllerPprof defines the pprof configs.
+type ControllerPprof struct {
```
vincepri, line 129:
```suggestion
// ControllerProfiling defines profiling configuration for controllers.
type ControllerProfiling struct {
```

pkg/config/v1alpha1/types.go
```diff
@@ -121,6 +125,17 @@ type ControllerHealth struct {
...
+	// BindAddress is the TCP address that the controller should bind to
+	// for serving pprof.
+	// It can be set to "" or "0" to disable the pprof serving.
+	// Since pprof may contain sensitive information, make sure to protect it
+	// before exposing it to public.
+	// +optional
+	BindAddress string `json:"bindAddress,omitempty"`
```
vincepri, line 136:
```suggestion
	// PprofBindAddress is the TCP address that the controller should bind to
	// for serving pprof.
	// It can be set to "" or "0" to disable the pprof serving.
	// Since pprof may contain sensitive information, make sure to protect it
	// before exposing it to public.
	// +optional
	PprofBindAddress string `json:"pprofBindAddress,omitempty"`
```

pkg/manager/internal.go
vincepri: Runnables can be closed and their lifecycle is pretty clear. Anything else outside runnables has the lifecycle outside the manager

pkg/manager/manager.go
```diff
@@ -403,6 +411,13 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		return nil, err
 	}
 
+	// Create pprof listener. This will throw an error if the bind
+	// address is invalid or already in use.
+	pprofListener, err := options.newPprofListener(options.PprofBindAddress)
+	if err != nil {
+		return nil, err
```
vincepri, line 418: Wrap this error with a nice error message?

pkg/manager/internal.go
zqzten: @vincepri But the http server started by `func (cm *controllerManager) httpServe` can also be closed within the manager's lifecycle. If using `Runnable` is a recommended way, why don't we also switch health probes and metrics server to it?
In other words, I'm not object to the `Runnable` way, but to the disunity within profiling server and health/metrics ones.

alvaroaleman: The fact that they are not Runnables today is just historic, they should be (if you want to work on this, please make that a distinct, refactoring-only change)

zqzten: @alvaroaleman Thanks for the explanation, will do that after this PR get done.

fovgrubby128:
Hi team, could I know when this CL would be merged and released?

pkg/config/v1alpha1/types.go
```diff
@@ -134,6 +138,17 @@ type ControllerHealth struct {
...
+// ControllerProfiling defines profiling configuration for controllers.
+type ControllerProfiling struct {
+	// PprofBindAddress is the TCP address that the controller should bind to
+	// for serving pprof.
+	// It can be set to "" or "0" to disable the pprof serving.
+	// Since pprof may contain sensitive information, make sure to protect it
+	// before exposing it to public.
+	// +optional
+	PprofBindAddress string `json:"pprofBindAddress,omitempty"`
+}
```
vincepri, line 150: Given the component configuration types got deprecated, can we remove these?

pkg/manager/manager.go
```diff
@@ -574,6 +594,19 @@ func defaultHealthProbeListener(addr string) (net.Listener, error) {
 	return ln, nil
 }
 
+// defaultPprofListener creates the default pprof listener bound to the given address.
+func defaultPprofListener(addr string) (net.Listener, error) {
+	if addr == "" || addr == "0" {
```
vincepri, line 599: Why not just check for an empty string? Why allow `0` to be a valid value?

pkg/config/v1alpha1/types.go
zqzten: done

pkg/manager/server.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		close(serverShutdown)
+	}()
+
+	log.Info("starting server")
+	if err := s.Server.Serve(s.Listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		return err
+	}
+
+	<-serverShutdown
```
FillZpp, line 55: How about put `s.Server.Serve()` into goroutine, then wait `ctx.Done()` and shutdown the server outside? So that we don't need `serverShutdown`. Is it more intuitive?

zqzten: We should return error immediately if `s.Server.Serve()` failed, which can't be easily done (still needs a channel) if we put it into goroutine.

vincepri (approved):
/approve
/lgtm

sbueringer:
Yes, that's the plan

fovgrubby128:
> Yes, that's the plan

Could I know the next release date please since we want to introduce this change in our code for some deadline? thanks!

sbueringer:
There is no fixed date as far as I know. But I think we want to get https://github.com/kubernetes-sigs/controller-runtime/pull/2189#issuecomment-1512485271 in first

Issue #1928, opened 2022-06-09
Client creation sets a global warning handler by default


ash2k:
https://github.com/kubernetes-sigs/controller-runtime/pull/1468 introduced warning handling. Unfortunately, the code makes an assumption that it's the place in a program that has the authority to set global warning handlers. It's simply not true.
https://github.com/kubernetes-sigs/controller-runtime/blob/0c2effbc7eabc502eb79472b6c6ba8fbb6ec8b76/pkg/client/client.go#L91-L98 

The code above shouldn't mess with globals, it should be done by the program's author in `main()` or something like that. The above code should only set warning handler on the provided config. And even that is questionable - if the caller wants, they can just configure that by themself, right? In the config they are providing and/or via the mentioned global. controller-runtime simply doesn't need to do anything, this code should be removed.

In my concrete situation I set warning handler in `main()` and it's overwritten by controller-runtime somewhere deep in the call stack.

ash2k:
It's clearly documented with examples how to use warning handlers in the introductory blog post https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling. The blog was even linked in the original issue (https://github.com/kubernetes-sigs/controller-runtime/issues/1150), so really not clear why this code was introduced.

FillZpp:
> In my concrete situation I set warning handler in `main()` and it's overwritten by controller-runtime somewhere deep in the call stack.

Aha, that's true. This might be because the `cluster` always create the same `client.Options` so that users can not set the `WarningHandlerOptions` in it.

https://github.com/kubernetes-sigs/controller-runtime/blob/0c2effbc7eabc502eb79472b6c6ba8fbb6ec8b76/pkg/cluster/cluster.go#L170-L180

PR #1944, opened 2022-06-28
⚠️ Do not mutate the global warning handler

ash2k:
Only configure the local warning handler. Please see the issue for details.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/1928.


varshaprasad96 (approved):
Based on the discussion in #1928, the fix seems reasonable.
/lgtm

alvaroaleman:
@ash2k the change seems fine but can you please mark it as breaking, since that is what it is now, for better or worse?

ash2k:
@alvaroaleman Changed the title if that was what you meant (not sure).

Thanks for a quick review, appreciate it!

PR #1950, opened 2022-07-06
✨ Provide access to admission.Request in custom validator/defaulter

### LINKED ISSUE #1949: Provide access to the admission.Request in validating & defaulting webhooks
As a developer writing defaulting and validating webhooks I would like to have access to the `admission.Request` without having to write an entire `admission.Handler`.

Our specific use case in Cluster API is that we want to validate an object on update depending on if the admission request is a dryrun.

Other use cases:
* Leverage `request.UserInfo` to block requests

Possible solution:
* Expose `admission.Request` in the `context.Context` passed into the Default/Validate funcs in the `CustomDefaulter`/`CustomValidator` (and a corresponding util func to read it from there)
    * Idea is to avoid changing the Validator/CustomValidator Defaulter/CustomValidator again / introducing new ones

sbueringer:
This PR provides access to the admisison.Request in custom validator and defaulter.

Fixes #1949

### DIFF
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -253,3 +253,21 @@ func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, err
 	}
 	return metrics.InstrumentedHook(opts.MetricsPath, hook), nil
 }
+
+// requestContextKey is how we find the admission.Request in a context.Context.
+type requestContextKey struct{}
+
+// RequestFromContext returns an admission.Request from ctx.
+func RequestFromContext(ctx context.Context) (Request, error) {
+	if v, ok := ctx.Value(requestContextKey{}).(Request); ok {
+		return v, nil
+	}
+
+	return Request{}, errors.New("admission.Request not found in context")
+}
+
+// NewContextWithRequest returns a new Context, derived from ctx, which carries the
+// provided admission.Request.
+func NewContextWithRequest(ctx context.Context, req Request) context.Context {
+	return context.WithValue(ctx, requestContextKey{}, req)
+}
--- a/pkg/builder/webhook_test.go
+++ b/pkg/builder/webhook_test.go
@@ -539,11 +539,13 @@ func runTests(admissionReviewVersion string) {
 // TestDefaulter.
 var _ runtime.Object = &TestDefaulter{}
 
+const testDefaulterKind = "TestDefaulter"
+
 type TestDefaulter struct {
 	Replica int `json:"replica,omitempty"`
 }
 
-var testDefaulterGVK = schema.GroupVersionKind{Group: "foo.test.org", Version: "v1", Kind: "TestDefaulter"}
+var testDefaulterGVK = schema.GroupVersionKind{Group: "foo.test.org", Version: "v1", Kind: testDefaulterKind}
 
 func (d *TestDefaulter) GetObjectKind() schema.ObjectKind { return d }
 func (d *TestDefaulter) DeepCopyObject() runtime.Object {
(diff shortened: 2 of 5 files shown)


pkg/webhook/admission/webhook.go
```diff
@@ -253,3 +253,21 @@ func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, err
 	}
 	return metrics.InstrumentedHook(opts.MetricsPath, hook), nil
 }
+
+// requestContextKey is how we find the admission.Request in a context.Context.
+type requestContextKey struct{}
+
+// RequestFromContext returns an admission.Request from ctx.
+func RequestFromContext(ctx context.Context) (Request, error) {
```
sbueringer, line 261: It would be a nicer UX if we could omit the error return parameter. But returning an empty Request if it is not found feels wrong

pkg/webhook/admission/webhook.go
```diff
@@ -253,3 +253,21 @@ func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, err
...
+	if v, ok := ctx.Value(requestContextKey{}).(Request); ok {
+		return v, nil
+	}
+
+	return Request{}, errors.New("admission.Request not found in context")
+}
+
+// newContext returns a new Context, derived from ctx, which carries the
+// provided admission.Request.
+func newContext(ctx context.Context, req Request) context.Context {
```
FillZpp, line 271: `newContextWithRequest` should be better?

pkg/webhook/admission/webhook.go
```diff
@@ -253,3 +253,21 @@ func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, err
 	}
 	return metrics.InstrumentedHook(opts.MetricsPath, hook), nil
 }
+
+// requestContextKey is how we find the admission.Request in a context.Context.
+type requestContextKey struct{}
```
FillZpp, line 258: How about a const string like `"webhookRequest"`?

pkg/webhook/admission/webhook.go
FillZpp: +1, an error return parameter will be nicer, even if that should not happen for custom validator/defaulter.

pkg/webhook/admission/webhook.go
sbueringer: I think the idea behind using an unexported struct is to only make it possible to retrieve it with the exported `RequestFromContext` func.

(I took this patter from how go-logr stores a Logger in a context)

pkg/webhook/admission/webhook.go
```diff
@@ -253,3 +253,21 @@ func StandaloneWebhook(hook *Webhook, opts StandaloneOptions) (http.Handler, err
...
+	if v, ok := ctx.Value(requestContextKey{}).(Request); ok {
+		return v, nil
+	}
+
+	return Request{}, errors.New("admission.Request not found in context")
+}
+
+// newContextWithRequest returns a new Context, derived from ctx, which carries the
+// provided admission.Request.
+func newContextWithRequest(ctx context.Context, req Request) context.Context {
```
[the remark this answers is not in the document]
sbueringer: Good point

pkg/webhook/admission/webhook.go
vincepri: The error return message seems ok, better to have the clarity rather than checking nils?

vincepri:
LGTM pending tests

pkg/webhook/admission/webhook.go
sbueringer: Yup and a lot better then checking empty fields of an empty struct (as it's not a pointer)

sbueringer:
Would be nice if we can cherry-pick this to v0.12.x

/cherry-pick release-0.12

PR #1953, opened 2022-07-07
:sparkles: Allow `builder.WebhookBuilder` to be used without a Manager

unguiculus:
It can be beneficial to set up an external webhook server for a manager
or to even run the webhook server as separate deployment.
However, `builder.WebhookBuilder` is tied to a `manager.Manager`, which
you don't have when setting up webhook server of its own. With this
change, a server and a scheme can be supplied instead of a manager, so
the same functionality can be used without a manager while keeping
backwards compatibility.

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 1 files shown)


FillZpp:
> It can be beneficial to set up an external webhook server for a manager
> or to even run the webhook server as separate deployment.
> However, builder.WebhookBuilder is tied to a manager.Manager, which
> you don't have when setting up webhook server of its own.

If you want to run the webhook server as separate deployment, just new a Manager and only register the webhook without any controllers into it. Does that have any problems?

unguiculus:
> If you want to run the webhook server as separate deployment, just new a Manager and only register the webhook without any controllers into it. Does that have any problems?

I'm aware I could do that, but does it really make sense to set up a Manager when all I need is a webhook server? I've always considered a Manager to be the thing that sets up my controllers. Rethinking this, I guess Manager sounds pretty generic, so it could just manage whatever there is to manage, even if it is just the webhook server 🤷. If that's the general notion of a manager, I'd be happy to close this PR.

PR #1954, opened 2022-07-13
🐛 Fix broken kind logging

### LINKED ISSUE #1945: Broken resource kind logging
https://github.com/kubernetes-sigs/controller-runtime/pull/1827 introduced this funny bug that kind is logged in unexpected way.

More specifically, https://github.com/kubernetes-sigs/controller-runtime/blob/ffd9ec8768b7d2e82aee6b0c76d481a5bdda4075/pkg/builder/controller.go#L318

If gvk.Kind is "APIService", I expect "apiService", but the current code produces and logs "aPIService".

tatsuhiro-t:
If gvk.Kind is "APIService", I expect "apiService", but the current
code produces and logs "aPIService".  Because it is not easy to do
this kind of thing, it is better to log the kind as is.

Fixes #1945

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -315,13 +315,13 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 			"controllerKind", gvk.Kind,
 		)
 
-		lowerCamelCaseKind := strings.ToLower(gvk.Kind[:1]) + gvk.Kind[1:]
+		kind := gvk.Kind
 
 		ctrlOptions.LogConstructor = func(req *reconcile.Request) logr.Logger {
 			log := log
 			if req != nil {
 				log = log.WithValues(
-					lowerCamelCaseKind, klog.KRef(req.Namespace, req.Name),
+					kind, klog.KRef(req.Namespace, req.Name),
 					"namespace", req.Namespace, "name", req.Name,
 				)
 			}


pkg/builder/controller.go
```diff
@@ -315,13 +315,13 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
 
-		lowerCamelCaseKind := strings.ToLower(gvk.Kind[:1]) + gvk.Kind[1:]
+		kind := gvk.Kind
 
 		ctrlOptions.LogConstructor = func(req *reconcile.Request) logr.Logger {
 			log := log
 			if req != nil {
 				log = log.WithValues(
-					lowerCamelCaseKind, klog.KRef(req.Namespace, req.Name),
+					kind, klog.KRef(req.Namespace, req.Name),
```
FillZpp, line 324: So why not just use `gvk.Kind`?

tatsuhiro-t: Fixed via bb368904

pkg/builder/controller.go
```diff
@@ -315,13 +315,11 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
 		)
 
-		lowerCamelCaseKind := strings.ToLower(gvk.Kind[:1]) + gvk.Kind[1:]
-
 		ctrlOptions.LogConstructor = func(req *reconcile.Request) logr.Logger {
 			log := log
 			if req != nil {
 				log = log.WithValues(
-					lowerCamelCaseKind, klog.KRef(req.Namespace, req.Name),
+					gvk.Kind, klog.KRef(req.Namespace, req.Name),
```
alvaroaleman, line 322: This is a breaking change for anyone who filters their logs, please mark it as such. Please also describe how this now looks like rather than only how it does not look like anymore.

tatsuhiro-t: Changed the icon to ⚠️ to indicate this is a breaking change, and edited the PR description to include how it looks now.

sbueringer: I see your point that aPIService is definitely bad. On the other side this change is directly in conflict with the conventions used by k/k: https://www.google.com/url?q=https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md%23name-arguments&sa=D&source=docs&ust=1657779189364954&usg=AOvVaw3jZwaKa1EhWBt3vcpGUqBe

Specifically:
> Always use [lowerCamelCase](https://en.wiktionary.org/wiki/lowerCamelCase), for example use containerName and not container name or container_name.

I wonder if there is any alternative instead of using kind directly

sbueringer: Looking at the example of good keys I'm not sure if APIService should be logged as `APIService` or `apiService` (especially the `CIDR` and `PVC` examples).

If we assume the first, it could be good enough to just keep the current case when the second character is also uppercase. (I'm aware it's not a great solution)

FillZpp: My thoughts:

1. Things like `PVC` is just the short name of `PersistentVolumeClaim`, which is different with the real kind `APIService`.
2. IMHO, changing the first case to lower is a little weird. Kubernetes resources have kind like `PersistentVolumeClaim` and resource(plural) like `persistentvolumeclaims`. But things like `persistentVolumeClaim` are only used as golang variable names for they should have first case lower, why should we keep this format in log?

tatsuhiro-t: My first impression of the linked documentation is it does not consider the CRD names which can be arbitrarily chosen by third parties and is hard to convert to camelCase in some cases.

Alternatively, we may follow the statement under exception lists:

> When Kubernetes object kind is unknown without runtime checking we should use `object` key. To provide information about kind we should add separate `apiVersion` and `kind` fields.

so use `object` as a key to log name and namespace, and log kind as is under `kind` key.

```
"object"={"name":"foo","namespace":"bar"} "kind"="APIService"
```

sbueringer: I have no idea why the picked those conventions, but I agree it's definitely hard to do if can't just hard-code the key.

Given that I think i also prefer just using kind directly.

The problem with using object is that it makes it very very hard to correlate log entries via a log query across controllers.
E.g. the following example:
* 1st controller reconciles Clusters: k/v pairs:  "cluster": <custer>
* 2nd controller reconciles Machines: k/v pairs: "machine": <machine>, "cluster": <cluster>

By using kind it's trivial to match logs across controllers. If the reconciled object has the "object" key it requires to calculate the actual kind before correlation between logs of different controllers are possible.

fabriziopandini: +1 to use kind without transformations, it is simpler, less verbose than object and it read more naturally

tatsuhiro-t: Right, so if I understand it correctly, we have a consensus on what we have now in this PR, that is log kind as a key without transformation.

alvaroaleman (approved):

PR #1971, opened 2022-08-09
✨ logging: allow override default logger for compatibility

timonwong:
#1827 introduces `WithLogConstructor` and removed `WithLogger` but in practical it's very hard to use `WithLogConstructor` to inject fields like:
- "controller", controllerName
- "controllerGroup": gvk.Group
-  "controllerKind", gvk.Kind

This PR proposes we can keep `WithLogger` for backward-compat.


alvaroaleman:
> introduces WithLogConstructor and removed WithLogger but in practical it's very hard to use WithLogConstructor to inject fields like:

Can you elaborate why that is hard and easier with this? From what I can tell the only thing it does is putting a logger into a closure

timonwong:
> Can you elaborate why that is hard and easier with this? From what I can tell the only thing it does is putting a logger into a closure

Previously, we just construct multiple loggers (our logging facility is similar to istio's, which register each business "scope", and their levels can be changed at runtime separately), pseudo code :

```go
var genericControlPlaneLogger = logging.RegisterScope("control-plane")
var istioUpdateLogger = logging.RegisterScope("istio-update")
var domainLogger = logging.RegisterScope("domain")
var wasmLogger =  logging.RegisterScope("wasm")
// etc, etc

// and there are multiple controllers, each controller setup with the builder:

ctrl1.WithLogger(genericControlPlaneLogger).Build()
ctrl2.WithLogger(istioUpdateLogger).Build()
ctrl3.WithLogger(domainLogger).Build()
ctrl4.WithLogger(wasmLogger).Build()
// etc, etc
```

And previously, **logger keys "controller", "controllerGroup", "controllerKind"** are auto-generated (after #1827, when customized logger with `WithLogConstructor`, you can only set them manually):

```go
var genericControlPlaneLogger = logging.RegisterScope("control-plane")
var istioUpdateLogger = logging.RegisterScope("istio-update")
var domainLogger = logging.RegisterScope("domain")
var wasmLogger =  logging.RegisterScope("wasm")
// etc, etc

func logConstructorFunc(logger logr.Logger, controllerName string, gvk schema.GroupVersionKind) func(*reconcile.Request) logr.Logger {
  return func(req *reconcile.Request) logr.Logger {
    logger = logger.WithValues("controller", controller, "controllerGroup", gvk.Group, "controllerKind", gvk.Kind)
    if req != nil {
      logger = logger.WithValues(gvk.Kind, klog.KRef(req.Namespace, req.Name), "namespace", req.Namespace, "name", req.Name)
    }
    return logger
  }
}

ctrl1.WithLogConstructor(logConstructorFunc(genericControlPlaneLogger, ...)).Build()
ctrl2.WithLogConstructor(logConstructorFunc(istioUpdateLogger, ...).Build()
ctrl3.WithLogConstructor(logConstructorFunc(domainLogger, ...).Build()
ctrl4.WithLogConstructor(logConstructorFunc(wasmLogger, ...)).Build()
```

You can see it (func **logConstructorFunc**) requires more work to archive same functionality as `WithLogger`

pkg/builder/controller.go
```diff
@@ -309,7 +317,14 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
...
 	// Setup the logger.
 	if ctrlOptions.LogConstructor == nil {
-		log := blder.mgr.GetLogger().WithValues(
+		var log logr.Logger
+		if ctrlOptions.GetDefaultLogger != nil {
+			log = ctrlOptions.GetDefaultLogger()
+		} else {
+			log = blder.mgr.GetLogger()
+		}
+
```
camilamacedo86, line 326: IHMO we should have only one log approach and ideally follow the k8s standards.
Why? 

- That helps those that are looking for Oberservaility, many tools will try to check the logs centralized
- Maintainability. We do not increase the burn to have more implementation for logs 
- Encourage standards.

In this way, I'd like to raise a question: What is the k8s format? Is k8s doing the logs using the format of WithLogConstructor or of the removed WithLogger? 

/hold

timonwong: k8s uses [klog](https://github.com/kubernetes/klog), which is derived from good-old glog.
In that case, I think it's better to avoid the use of logr, which backend is configuratble, and we can stick to klog.

camilamacedo86: @timonwong,  

The most important I think is not the dependency used but the format of the logs. 

However, I think would be nice if we could check how the format is outputted by k8s as the dep (which you already checked) and if we can or do not follow the same standards. if not, why not? Also, have we any Kubernetes good practices definition about it? If so, I think we should also follow up on that. 

WDYT? If that also makes sense for you and you would like to work on this I think the next step would be an assessment and open an issue and/or PR with the proposal.

I will 
/hold cancel

to let the others also share their thoughts.

camilamacedo86: Hi @timonwong, 

(ihmo) The most important I think is not the dependency used but the output/format of the logs. According to the description of https://github.com/kubernetes-sigs/controller-runtime/pull/1827 the goal is to align with `align to Kubernetes structured logging, add reconcileID ` 

However, I think would be nice if we could check how the format is outputted by k8s as the dep (which you already checked) and if we can or do not follow the same standards. if not, why not? Also, have we any Kubernetes good practices definition about it? If so, I think we should also follow up on that. By last, I think also would be nice to understand why we are using `GetDefaultLogger ` and not klog. What was the motivation for this adoption instead of klog?  

(IMHO) the above assessment would need to be done before we propose changes on how it is implemented. If we figure out that we need to change then, maybe open an issue describing the proposal and motivations for that can help out in the discussion.  Also, a PR proposing the changes based on the motivations can be helpful for sure too. 

WDYT?

alvaroaleman:
@timonwong how about instead just adding the fields unconditionally, regardless of logConstructor being set or not?

The new api this introduces is IMHO not great, from a users POV it is very difficult to understand what the difference between `LogConstructor` and `GetDefaultLogger` is. Also I can not think of a reason for not wanting the `controller` etc fields on the logger. That will make it a breaking change though.

pkg/builder/controller.go
alvaroaleman: @camilamacedo86 I think a lot of what you are asking here is orthogonal to this change. This changes goal is to have the same fields that get set when a logger comes from the mgr when a custom logger is used.

I agree that it is not great to have to ways to set a custom logger that are different in a very subtle way.

PR #1972, opened 2022-08-10
✨ Pass webhook logger to handler via context

timebertt:
`admission.Webhook` now passes a request logger to the handler via the context.
It can be retrieved by handlers via `logf.FromContext(ctx)` and is setup with commonly interesting fields about the request.
The logger and it's fields can be customized by setting `Webhook.LogConstructor`.

This is very similar to how controllers are passing a reconciliation logger to reconcilers and allows for greater harmonization across the entire codebase, i.e. using contextual loggers everywhere instead of manually passing around half-cooked loggers.

### DIFF
--- a/pkg/webhook/admission/webhook.go
+++ b/pkg/webhook/admission/webhook.go
@@ -23,15 +23,17 @@ import (
 	"net/http"
 
 	"github.com/go-logr/logr"
-	jsonpatch "gomodules.xyz/jsonpatch/v2"
+	"gomodules.xyz/jsonpatch/v2"
 	admissionv1 "k8s.io/api/admission/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/json"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/klog/v2"
 
-	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
+	internallog "sigs.k8s.io/controller-runtime/pkg/internal/log"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics"
 )
@@ -131,6 +133,12 @@ type Webhook struct {
 	// headers thus allowing you to read them from within the handler
 	WithContextFunc func(context.Context, *http.Request) context.Context
 
+	// LogConstructor is used to construct a logger for logging messages during webhook calls
+	// based on the given base logger (which might carry more values like the webhook's path).
+	// Note: LogConstructor has to be able to handle nil requests as we are also using it
+	// outside the context of requests.
+	LogConstructor func(base logr.Logger, req *Request) logr.Logger
+
 	// decoder is constructed on receiving a scheme and passed down to then handler
 	decoder *Decoder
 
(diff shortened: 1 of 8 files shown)


timebertt:
Resolved the merge conflict.
@alvaroaleman could you kindly have a look?

pkg/webhook/admission/webhook.go
```diff
@@ -166,15 +174,38 @@ func (wh *Webhook) Handle(ctx context.Context, req Request) (response Response)
...
 	if err := resp.Complete(req); err != nil {
-		wh.log.Error(err, "unable to encode response")
+		reqLog.Error(err, "unable to encode response")
 		return Errored(http.StatusInternalServerError, errUnableToEncodeResponse)
 	}
 
 	return resp
 }
 
+// getLogger constructs a logger from the injected log and LogConstructor.
```
sbueringer, line 190: I wonder if we should implement this the same way as we did for the controller:
```go
	if options.LogConstructor == nil {
		log := mgr.GetLogger().WithValues(
			"controller", name,
		)
		options.LogConstructor = func(req *reconcile.Request) logr.Logger {
			log := log
			if req != nil {
				log = log.WithValues(
					"object", klog.KRef(req.Namespace, req.Name),
					"namespace", req.Namespace, "name", req.Name,
				)
			}
			return log
		}
	}
```

The delta is roughly:
* LogConstructor would have this signature: `LogConstructor func(req *Request) logr.Logger`
* The default log constructor would then use the base logger directly
* namespace & name are also added as top-level k/v pairs in addition to the nested values below object

e.g.:
```go
// getLogger constructs a logger from the injected log and LogConstructor.
func (wh *Webhook) getLogger(req *Request) logr.Logger {
	if wh.LogConstructor == nil {
		if req != nil {
			return wh.log.WithValues("object", klog.KRef(req.Namespace, req.Name),
				"namespace", req.Namespace, "name", req.Name,
				"resource", req.Resource, "user", req.UserInfo.Username,
			)
		}
		return wh.log
	}
	return wh.LogConstructor(req)
}
```

pkg/webhook/admission/webhook.go
```diff
@@ -131,6 +133,12 @@ type Webhook struct {
 	// headers thus allowing you to read them from within the handler
 	WithContextFunc func(context.Context, *http.Request) context.Context
 
+	// LogConstructor is used to construct a logger for logging messages during webhook calls
+	// based on the given base logger (which might carry more values like the webhook's path).
+	// Note: LogConstructor has to be able to handle nil requests as we are also using it
+	// outside the context of requests.
+	LogConstructor func(base logr.Logger, req *Request) logr.Logger
```
sbueringer, line 140: It would be nice if we can set the LogConstructor in the webhook builder like we do in the controller builder
cf. https://github.com/kubernetes-sigs/controller-runtime/blob/b93b5f92794b9383427995678d22ddac396dba13/pkg/builder/controller.go#L318-L327

(delta is that we know the kind there and also have access to the mgr)

sbueringer:
In general I like the idea to make the webhook logs consistent with the controller logs. This will also allow cross-referencing with webhook logs.

/ok-to-test

sbueringer:
cc @fabriziopandini (just fyi that there is work in this area)

pkg/webhook/admission/webhook.go
timebertt: > LogConstructor would have this signature: `LogConstructor func(req *Request) logr.Logger`

I added the `base logr.Logger` param to the `LogConstructor` to allow using the logger injected by the webhook server which includes the webhook path:
https://github.com/kubernetes-sigs/controller-runtime/blob/869888c3e70c8c87f33e3ca000aaef451c6c5cbb/pkg/webhook/server.go#L165

This is different from controllers, which use the manager's logger directly (which doesn't carry any additional information about the controller):
https://github.com/kubernetes-sigs/controller-runtime/blob/598978c592839b5a602af3706fce8751d2d2532c/pkg/controller/controller.go#L110

With this, one can add additional fields via a custom `LogConstructor` without adding the `webhook` field again.
If one wants to ignore the fields of the base logger, one can simply use any other logger.

I get that this is different from the controller loggers. If harmonization is more important here than preserving the injected fields, I will apply your suggestion. :)
@sbueringer WDYT?

timebertt:
@sbueringer I have pushed two more commits addressing parts of your suggestions. PTAL :)

pkg/webhook/admission/webhook.go
sbueringer: Looks good to me. Thank you very much!!

sbueringer:
Thank you!!

/lgtm

PR #1976, opened 2022-08-11
:running: Update golangci-lint to 1.47.3

alvaroaleman:
Updates golangci-lint in github actions

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -19,5 +19,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.45.2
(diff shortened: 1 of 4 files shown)


alvaroaleman:
Seems the latest version wants to gofmt in the (incompatible) go 1.19 style, I went back one version to not have go 1.19 support as the repo is tested with go 1.18, lets see how that goes

alvaroaleman:
The error assertions in the dynamic rest mappers checkAndReload never worked and fixing them makes the tests fail :/

alculquicondor:
/lgtm
/hold
in case you would like to have a review from an experienced contributor to this repo.

alvaroaleman:
I think this is fine, as can be seen by the prior failures this has a very good test coverage
/hold cancel

PR #1977, opened 2022-08-11
⚠ Migrate to ginkgo v2

schrej:
Revives #1792, thanks to @acumino for all your work on this.

Fixes: #1545 

Important Note:
-> Per node timeout has been removed to prevent the wrongly marking a test as failure which is caused by some other test. Read more about it [here](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing)
-> Custom reporters have been deprecated in ginkgo v2 and will be removed in future releases.

Note: This is a breaking change for two reasons:
- (we're not affected by this) ginkgo v1 and v2 have conflicting flags, so you can't import them both at the same time (see [here](https://github.com/onsi/ginkgo/issues/875#issuecomment-1036350659)). Thanks to @mboersma for pointing this out
- The PR removes two helpers in `pkg/envtest/printer` that were used with ginkgo v1. Since you can't use both versions at the same time, this isn't relevant though.

Upstream upgraded already: https://github.com/kubernetes/kubernetes/pull/109111

/cc @matteoolivi @camilamacedo86


schrej:
cc @matteoolivi @camilamacedo86

pkg/envtest/printer/prow.go
```diff
@@ -35,75 +32,22 @@ var (
...
-	if pr.junitReporter != nil {
-		pr.junitReporter.SpecDidComplete(specSummary)
-	}
-}
 
-// SpecSuiteDidEnd Prints a newline between "35 Passed | 0 Failed | 0 Pending | 0 Skipped" and "--- PASS:".
-func (pr *prowReporter) SpecSuiteDidEnd(summary *types.SuiteSummary) {
-	if pr.junitReporter != nil {
-		pr.junitReporter.SpecSuiteDidEnd(summary)
+	if os.Getenv("CI") != "" && artifactsDir != "" {
```
alvaroaleman, line 46: The only reason for this existing btw was that generating junit made ginkgo always exist with RC 0 before, even if there were failures. If there is an easier way to get it to both generate junit and exit with rc != 0 if there is a failure, that would be great.

alvaroaleman:
> ginkgo v1 and v2 have conflicting flags, so you can't import them both at the same time (see https://github.com/onsi/ginkgo/issues/875#issuecomment-1036350659). Thanks to @mboersma for pointing this out

But we only import it in _test.go files and in the printer package, so this shouldn't matter for consumers, right?

schrej:
> But we only import it in _test.go files and in the printer package, so this shouldn't matter for consumers, right?

I just tested with the cluster-api project, manually replacing controller-runtime with this branch there. Tests can still run, even though CAPI is still on ginkgo v1. So you're right, it shouldn't matter to consumers in our case.

I'll revert the removal of the two helpers in `pkg/envtest/printer`, which should make this a non-breaking change then.

pkg/envtest/printer/prow.go
schrej: I think we can run tests with the `ginkgo` cli instead of `go test`, which allows to configure junit report generation. Still evaluating if it works for everything.
It does run regular Go tests, but it does not include them in the junit report. But we don't have them in the report currently anyway, correct?

schrej: I've switched the PR to the ginkgo cli now to see if it works.
One major drawback I've seen is that `ginkgo` doesn't cache tests, so it runs the full suite every time. That might not be ideal and break/slow down workflows.

schrej:
Ok. This is now using the `ginkgo` cli, which allows to remove the ginkgo helper to create unit test reports.
The apidiff is failing because of that.

I've adapted the test scripts to install the `ginkgo` cli automatically and to copy the unit test report to the correct location for prow to detect and visualise it.

One major drawback of the `ginkgo` cli is that it does not cache tests, and always runs the full suite.

alvaroaleman:
@schrej the tests paniced but still passed: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/1977/pull-controller-runtime-test-master/1559847725640454144

pohly:
> The k8s.io/klog package logs panics it observes to stderr, and Ginkgo apparently prints that even when the test passes.

It's the runtime which calls klog.Error, which then seems to go to stderr.

Does the suite call `klog.SetOutput(ginkgo.GinkgoWriter)`? Then I think you can suppress that extra output to stderr by reconfiguring the "stderrthreshold". I have a pending commit which does that for the in-tree e2e.test:
https://github.com/kubernetes/kubernetes/pull/111023/commits/7d6af86168eede751865988664638a2037596e0d

pohly:
> I've adapted the test scripts to install the ginkgo cli automatically and to copy the unit test report to the correct location for prow

FYI, it is possible to enable JUnit output without using the ginkgo cli. You can either do `go test ./<test dir> -args -ginkgo.junit=<output file>` or add some code like https://github.com/kubernetes/kubernetes/blob/a1128e380c2cf1c2d7443694673d9f1dd63eb518/test/e2e/framework/test_context.go#L558-L607.

alvaroaleman:
> FYI, it is possible to enable JUnit output without using the ginkgo cli. You can either do go test ./<test dir> -args -ginkgo.junit=<output file> or add some code like

The issue with that was that it made the tests always exit with rc0 even when there are failures

pohly:
> The issue with that was that it made the tests always exit with rc0 even when there are failures

Which of the two options? `-ginkgo.junit` or adding some custom `GenerateJUnitReport` invocation? Either way, that sounds like a bug that should be reported.

pohly:
Hmm, both works for me (PMEM-CSI test/e2e, using ginkgo v2 v2.1.4).

FYI, I have to use this fork of the controller-runtime for that project because otherwise I would be stuck at ginkgo v1 and couldn't update to Kubernetes 1.25. I'll have to keep that in a WIP PR until this PR gets merged and released.

schrej:
> Why? Shouldn't we only depend on ginkgo in _test.go files? If anyhow possible, we shouldn't force users to use either version

Probably because we're currently importing ginkgo v1 in `pkg/envtest/printer`.

alvaroaleman:
> Probably because we're currently importing ginkgo v1 in pkg/envtest/printer.

But v2 is imported as `github.com/onsi/ginkgo/v2` or not?

schrej:
Yes, but importing both at the same time causes issues with flag parsing as mentioned in the PR description.

schrej:
All packages need to import `_ "github.com/onsi/ginkgo/v2"`, even if they don't contain ginkgo tests when we want to run tests with the `ginkgo.junit-report` flag. It's only one package though, I'd say it's fine.

pohly:
> > Which of the two options? -ginkgo.junit or adding some custom GenerateJUnitReport invocation? Either way, that sounds like a bug that should be reported.
> 
> It was reported here, maybe it doesn't happen anymore? [onsi/ginkgo#706](https://github.com/onsi/ginkgo/issues/706)

That looks like a different problem to me: exit code was zero when it shouldn't be because of a failed test, but it's not clear that this was caused by using the JUnit reporter.

pohly:
> All packages need to import _ "github.com/onsi/ginkgo/v2", even if they don't contain ginkgo tests when we want to run tests with the ginkgo.junit-report flag.

If a package doesn't contain ginkgo tests, then how can ginkgo gather results for the tests in the package? Presumably those are plain `func TestFoo(t *testing.T)` tests? Can you give an example?

Importing ginkgo into such a Go test package adds the ginkgo command line flags, but they don't have any effect at runtime.

schrej:
> If a package doesn't contain ginkgo tests, then how can ginkgo gather results for the tests in the package? Presumably those are plain `func TestFoo(t *testing.T)` tests? Can you give an example?
> 
> Importing ginkgo into such a Go test package adds the ginkgo command line flags, but they don't have any effect at runtime.

It can't. But `go test` will complain about the undefined command line arguments if it's not imported.

pohly:
> This is really, really bad. Is there any way to fix ginkgo, has this issue bee raised there?

This cannot be fixed. The Ginkgo API is designed around global variables and directly extending the default FlagSet, to simplify test startup. This could have been solved differently, but cannot be changed anymore without another breaking change.

Even if ginkgo v1 and v2 could co-exist, what would be the point? Tests registered with v1 Describe would be invisible to the test runner from ginkgo v2 and vice versa.

What you could do in controller-runtime is to limit the usage of ginkgo to _test.go files. Then consumers of the controller-runtime are free to choose whatever version of ginkgo they want to use in their tests. Actually, I am not sure anymore why controller-runtime seemed to pull in ginkgo v1 into PMEM-CSI. The only non-test files which use ginkgo are pkg/envtest/printer and those are not required. I've reverted the usage of this fork in PMEM-CSI and don't see any problems.

schrej:
> This is really, really bad. Is there any way to fix ginkgo, has this issue bee raised there?

That's outdated by now. I've tested it in the beginning and it was not necessary to upgrade everything to v2. As @pohly said, if we only import ginkgo in `_test.go` files there won't be a conflict, and we're currently doing that.

alvaroaleman:
Okay, that is awesome. I have a really busy week atm, will try to look at this next week. If someone else has time for a first pass, that would be great :)

hack/check-everything.sh
```diff
@@ -25,7 +25,7 @@ tmp_root=/tmp
 kb_root_dir=$tmp_root/kubebuilder
 
 # Run verification scripts.
-${hack_dir}/verify.sh
+# ${hack_dir}/verify.sh
```
vincepri, line 28: Leftover?

pkg/cache/cache_suite_test.go
```diff
@@ -19,20 +19,20 @@ package cache_test
...
 	. "github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+const suiteName = "Cache Suite"
```
vincepri, line 31: Why does this need to be a global constant?

pkg/cache/internal/internal_suite_test.go
```diff
@@ -19,13 +19,13 @@ package internal
 import (
 	"testing"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 )
 
+const suiteName = "Cache Internal Suite"
```
vincepri, line 26: Same here

pkg/certwatcher/certwatcher_suite_test.go
```diff
@@ -20,30 +20,29 @@ import (
...
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 var (
-	certPath = "testdata/tls.crt"
-	keyPath  = "testdata/tls.key"
+	certPath  = "testdata/tls.crt"
+	keyPath   = "testdata/tls.key"
+	suiteName = "CertWatcher Suite"
```
vincepri, line 32: Same here

pkg/internal/testing/controlplane/controlplane_suite_test.go
```diff
@@ -19,15 +19,14 @@ package controlplane_test
...
 	"testing"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 )
 
+const suiteName = "Control Plane Standup Unit Tests"
```
vincepri, line 26: Seems there are a lot of these, can we just inline them?

pkg/cache/cache_suite_test.go
schrej: In the past the suite name had to be passed twice. Since that's no longer the case we can remove it now.

hack/check-everything.sh
schrej: Yep, sorry, that sneaked in with the last amend :sweat_smile:

sbueringer:
> > This is really, really bad. Is there any way to fix ginkgo, has this issue bee raised there?
> 
> That's outdated by now. I've tested it [in the beginning](https://github.com/kubernetes-sigs/controller-runtime/pull/1977#issuecomment-1212892475) and it was not necessary to upgrade everything to v2. As @pohly said, if we only import ginkgo in `_test.go` files there won't be a conflict, and we're currently doing that.

+1. Just another data point (although in the other direction). We bumped to ginkgo v2 in ClusterAPI and are still using the regular controller-runtime v0.12.3 release without problems

sbueringer:
Should be the same issue as on https://github.com/kubernetes-sigs/controller-runtime/pull/1986

Essentially the next PR that goes in has to be a golangci-lint bump

(cause is probably that we bumped the jobs to use go 1.19 images)

sbueringer:
Fix PR: https://github.com/kubernetes-sigs/controller-runtime/pull/1987

sbueringer:
@schrej Looks like I introduced a conflict by bumping golangci-lint

PR #1980, opened 2022-08-16
⚠ cache.BuilderWithOptions inherit options from caller

joelanford:
using cache.BuilderWithOptions does not properly inherit all
options passed in from the caller:
- scheme was overridden instead of merged
- selectors were not inherited at all, even if specified options
  selectors remained unset
- transforms were not inherited at all, even if specified options
  transforms remained unset
- disable deep copy settings were not inherited at all, even if
  specified options for disabling deep copies remained unset

This commit resolves this issues by implementing merge logic for all
fields in the cache.Options struct:
- Schemes are merged
- RESTMapper is chosen by precedence only falling back to inherited
  options if left unset in specified options
- Resync is chosen by precedence only falling back to inherited
  options if left unset in specified options
- Namespace is chosen by precedence only falling back to inherited
  options if left unset in specified options
- Selectors are merged. If both inherited and specified Options defined
  selectors for a given type, those selectors are merged via logical
  AND.
- DisableDeepCopy is combined via precedence. Only if a value for a
  particular GVK is unset in the specified options will a value from the
  inherited options be used.
- Transform functions are combined via chaining. If both inherited and
  specified options define a transform function, the transform function
  from the inherited options will be called first, and the transform
  function from the specified options will be called second.

This is a breaking change if your code does all of the following:
1. Uses `cache.BuilderWithOptions`
2. Specifically depends on the fact that inherited options are not currently considered for selectors, disabling deep copies, and transform functions OR that the inherited scheme is ignored if it is defined in the options specified with `cache.BuilderWithOptions`.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -19,14 +19,18 @@ package cache
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	toolscache "k8s.io/client-go/tools/cache"
+
 	"sigs.k8s.io/controller-runtime/pkg/cache/internal"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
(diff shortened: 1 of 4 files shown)


varshaprasad96 (commented):
This fix does seem good to me. Except for my questions on having [`DisableDeepCopyByObject` ](https://github.com/kubernetes-sigs/controller-runtime/blob/cd0058ad295c268da1e7233e609a9a18dd60b5f6/pkg/cache/cache.go#L236) external. But that change isn't made here, and hence not concerning/blocking. Leaving it for others to review.

pkg/cache/cache.go
```diff
@@ -150,49 +154,242 @@ func New(config *rest.Config, opts Options) (Cache, error) {
...
+	if err != nil {
+		return nil, ObjectSelector{}, err
+	}
+	inheritedSelectorsByGVK, err := convertToByGVK(inherited.SelectorsByObject, inherited.DefaultSelector, inherited.Scheme)
+	if err != nil {
+		return nil, ObjectSelector{}, err
+	}
+
+	for gvk, selector := range inheritedSelectorsByGVK {
+		optionsSelectorsByGVK[gvk] = combineSelector(optionsSelectorsByGVK[gvk], selector)
```
sbueringer, line 285:
```suggestion
		optionsSelectorsByGVK[gvk] = combineSelector(selector, optionsSelectorsByGVK[gvk])
```
nit: I think it doesn't matter but should we also pass in the inherited selector here as first param like in all other combine/select funcs?

joelanford: Makes sense. Updated!

sbueringer:
Oh wasn't aware self-approve is enabled in this repo. Sorry if this wasn't intended to merge already

PR #1988, opened 2022-08-25
Bump golangci lint to v1.49.0

sbueringer:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,6 @@ linters:
   - goconst
   - gocritic
   - gocyclo
-  - godot
   - gofmt
   - goimports
   - goprintffuncname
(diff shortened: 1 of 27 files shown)


.golangci.yml
```diff
@@ -132,6 +131,9 @@ issues:
   - linters:
     - gosec
     text: "G304: Potential file inclusion via variable"
+  - linters:
+    - revive
+    text: "package-comments: should have a package comment"
```
sbueringer, line 136: If we want to add package comments everywhere I would suggest to do that in follow-up PRs (I can open an issue for it).

There are a lot of packages with missing package comments

alvaroaleman (approved):

PR #1992, opened 2022-09-02
📖 add comments for configuration

Icarus9913:
add configuration comments once use `NewManager` 

Signed-off-by: Icarus9913 <icaruswu66@qq.com>

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -219,6 +219,7 @@ type Options struct {
 
 	// HealthProbeBindAddress is the TCP address that the controller should bind to
 	// for serving health probes
+	// It can be set to "0" or "" to disable serving the health probe.
 	HealthProbeBindAddress string
 
 	// Readiness probe endpoint name, defaults to "readyz"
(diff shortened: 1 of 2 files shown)


Icarus9913:
reference : 
https://github.com/kubernetes-sigs/controller-runtime/blob/15b6689c55745514d6bade4c55efb3333c0da4d9/pkg/manager/manager.go#L552-L555

pkg/manager/manager.go
```diff
@@ -219,6 +219,7 @@ type Options struct {
 
 	// HealthProbeBindAddress is the TCP address that the controller should bind to
 	// for serving health probes
+	// It can be set to "0" or "" to disable the health serving.
```
sbueringer, line 222:
```suggestion
	// It can be set to "0" or "" to disable serving the health probe.
```
nit

Can you please add the same comment in: https://github.com/kubernetes-sigs/controller-runtime/blob/c48baad70c539a2efb8dfe8850434ecc721c1ee1/pkg/config/v1alpha1/types.go#L113

Icarus9913: already added

alvaroaleman (approved):

Issue #878, opened 2020-03-26
flag redefined: kubeconfig: allow double vendoring this library but still register flags on behalf of users


akoserwal:
Error:
```
 /var/folders/3d/_kgky9bj61g5144z004y51qr0000gp/T/go-build652443801/b001/e2e.test flag redefined: kubeconfig
panic: /var/folders/3d/_kgky9bj61g5144z004y51qr0000gp/T/go-build652443801/b001/e2e.test flag redefined: kubeconfig

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc0000d2120, 0x244c6a0, 0xc00004e250, 0x224908b, 0xa, 0x225fced, 0x12)
    /usr/local/Cellar/go/1.13.8/libexec/src/flag/flag.go:848 +0x4ae
flag.(*FlagSet).StringVar(...)
    /usr/local/Cellar/go/1.13.8/libexec/src/flag/flag.go:751
github.com/operator-framework/operator-sdk/pkg/test.(*frameworkOpts).addToFlagSet(0xc00004e240, 0xc0000d2120)
    /Users/akoserwa/go/pkg/mod/github.com/operator-framework/operator-sdk@v0.15.1/pkg/test/framework.go:102 +0x1bf
github.com/operator-framework/operator-sdk/pkg/test.MainEntry(0xc00040d280)
    /Users/akoserwa/go/pkg/mod/github.com/operator-framework/operator-sdk@v0.15.1/pkg/test/main_entry.go:27 +0x50
github.com/atlasmap/atlasmap-operator/test/e2e.TestMain(...)
       _testmain.go:42 +0x136
```
Using :- 
```
framework “github.com/operator-framework/operator-sdk/pkg/test” where it is setting
> flagset.StringVar(&opts.kubeconfigPath, KubeConfigFlag, “”, “path to kubeconfig”)
```

And in my test case, I am using a library which is also using controller-runtime/operator-sdk
```github.com/RHsyseng/operator-utils/pkg/utils/openshift```

 Method I am calling: `openshift.IsOpenShift(framework.KubeConfig)`

controller-runtime/pkg/client/config/config.go
```
// TODO: Fix this to allow double vendoring this library but still register flags on behalf of users
  flag.StringVar(&kubeconfig, “kubeconfig”, “”,
    “Paths to a kubeconfig. Only required if out-of-cluster.“)
```
 I suspect fixing TODO will resolve it? If yes, I would like to work on it.

[...]

PR #1999, opened 2022-09-13
:seedling: Register kubeconfig flag variable via RegisterFlags func

johannesfrey:
There have been reported issues (`flag redefined`) because CR registers the `kubeconfig` flag to the default command line FlagSet via init function. See:
* https://github.com/kubernetes-sigs/controller-runtime/issues/1396
* https://github.com/kubernetes-sigs/controller-runtime/issues/878

This PR provides an exported func to register flags that defensively tries to register the kubeconfig flag if not already defined. It does only change the current behaviour by previously checking if has already been defined.

As a follow up the registration via init should probably removed in conjunction with a :warning: release note, as it would be a breaking change for the user.

**NOTE**: I am not entirely sure if this really is the desired way to go and to start a discussion regarding the topic.

<sub>Johannes Frey <[johannes.frey@mercedes-benz.com](mailto:johannes.frey@mercedes-benz.com)>, Mercedes-Benz Tech Innovation GmbH ([Provider Information](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))</sub>

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -50,7 +50,10 @@ func RegisterFlags(fs *flag.FlagSet) {
 	if fs == nil {
 		fs = flag.CommandLine
 	}
-	if fs.Lookup(KubeconfigFlagName) == nil {
+	f := fs.Lookup(KubeconfigFlagName)
+	if f != nil {
+		kubeconfig = f.Value.String()
+	} else {
 		fs.StringVar(&kubeconfig, KubeconfigFlagName, "", "Paths to a kubeconfig. Only required if out-of-cluster.")
 	}
 }
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -29,15 +29,30 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
 
+// KubeconfigFlagName is the name of the kubeconfig flag
+const KubeconfigFlagName = "kubeconfig"
+
 var (
 	kubeconfig string
 	log        = logf.RuntimeLog.WithName("client").WithName("config")
 )
 
+// init registers the "kubeconfig" flag to the default command line FlagSet.
+// TODO: This should be removed, as it potentially leads to redefined flag errors for users, if they already
+// have registered the "kubeconfig" flag to the command line FlagSet in other parts of their code.
 func init() {
-	// TODO: Fix this to allow double vendoring this library but still register flags on behalf of users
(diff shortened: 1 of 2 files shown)


sbueringer:
> /label do-not-merge/work-in-progress

I think if you move [WIP] to be a prefix it should work automatically

sbueringer:
/retitle [WIP] :seedling: Register kubeconfig flag variable via RegisterFlags func

johannesfrey:
> > /label do-not-merge/work-in-progress
> 
> I think if you move [WIP] to be a prefix it should work automatically

Yeah, I had it like that but then workflow complains :sweat_smile:

pkg/client/config/config.go
```diff
@@ -29,15 +29,30 @@ import (
...
+}
+
+// RegisterFlags registers flag variables to the given FlagSet if not already registered.
+// It uses the default command line FlagSet, if none is provided. Currently, it only registers the kubeconfig flag.
+func RegisterFlags(fs *flag.FlagSet) {
+	if fs == nil {
+		fs = flag.CommandLine
+	}
+	if fs.Lookup(KubeconfigFlagName) == nil {
+		fs.StringVar(&kubeconfig, KubeconfigFlagName, "", "Paths to a kubeconfig. Only required if out-of-cluster.")
```
alvaroaleman, line 54: But if the flag is registered and ppl attempt to use it, it will just silently not work or not?

johannesfrey: You are right. In the case you describe the value is not stored in the `kubeconfig` string var. We could solve this by getting the value if it was already set:
```go
	f := fs.Lookup(KubeconfigFlagName)
	if f != nil {
		kubeconfig = f.Value.String()
	} else {
		fs.StringVar(&kubeconfig, KubeconfigFlagName, "", "Paths to a kubeconfig. Only required if out-of-cluster.")
	}
```

Or should we simply drop the lookup and expect that there is not flag yet provided if someone is about to call `RegisterFlags`?

alvaroaleman: Looking it up would be great if it already exists.

> Or should we simply drop the lookup and expect that there is not flag yet provided if someone is about to call RegisterFlags?

No super strong opinion here, but in case of someone explicitly calling `RegisterFlags` I would not expect it to be defined. In the case of the `init` though it might be

johannesfrey: Sorry for my late reply. I pushed a new commit containing the lookup.

alvaroaleman (approved):
thanks!

PR #2003, opened 2022-09-17
:warning: Inc Generation on fakeClient Update/Patch

### LINKED ISSUE #1857: Support incrementing Generation on fake client Update and Patch
When calling `Update` and `Patch` using the fake client, the `metadata.Generation` field is not incremented. This makes testing logic that depends on this field significantly harder (e.g. knowing if conditions are up to date by checking if the generation matches the observed generation).

I've taken a look at the code, and it seems this is a known limitation as documented in #1031.

[pkg/client/fake/doc.go](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.2/pkg/client/fake/doc.go#L30):
```
WARNING: ⚠️ Current Limitations / Known Issues with the fake Client ⚠️
- This client does not have a way to inject specific errors to test handled vs. unhandled errors.
- There is some support for sub resources which can cause issues with tests if you're trying to update
  e.g. metadata and status in the same reconcile.
- No OpeanAPI validation is performed when creating or updating objects.
- ObjectMeta's `Generation` and `ResourceVersion` don't behave properly, Patch or Update
operations that rely on these fields will fail, or give false positives.
```

I would like to contribute a fix for it and wanted to better understand the complexities behind implementing the fix. I also came across #553 that commented on this.

[pkg/predicate/predicate.go](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.2/pkg/predicate/predicate.go#L149):
```
// * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
// E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
// For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
```

It seems that the logic that should decide if the generation should be incremented is not straight forward. @jpatel-pivotal, @hasbro17 if any of  you (as the referenced CL authors) have any insights I would be grateful if you can share them. Meanwhile I'll try to look up answers myself.

My current intention is to add a conditional increment of the `metadata.Generation` field in [`versionedTracker.Update`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.2/pkg/client/fake/client.go#L255). This should cover both `Update` and `Patch` operations since `Patch` also calls `tracker.Update` under the hood if I read the code correctly.

danaradg:
Closes #1857.

This applies only to custom resources!
The behaviour of the API server regarding generation is polymorphic meaning that any resource type may behave differently. Custom resources have a uniform behaviour (increment on non-metadata change), and so this patch adds logic to `fakeClient` to handle this specific use case.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -385,9 +385,14 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	intResourceVersion++
 	accessor.SetResourceVersion(strconv.FormatUint(intResourceVersion, 10))
 
+	crRequiresGenerationUpdate, err := didNonMetadataFieldChange(oldObject, obj)
+	if err != nil {
+		return fmt.Errorf("failed to check if generation should be updated: %w", err)
+	}
+
 	// The behavior of the generation field is inconsistent across built-in k8s resources, but it's
 	// consistent for CRs. CRs have their generation increased on non-metadata changes only.
-	if isCustomResource(gvk) && didNonMetadataFieldChange(oldObject, obj) {
+	if isCustomResource(gvk) && crRequiresGenerationUpdate {
 		accessor.SetGeneration(oldAccessor.GetGeneration() + 1)
 	}
 
@@ -405,14 +410,21 @@ func isCustomResource(gvk schema.GroupVersionKind) bool {
 	return !scheme.Scheme.Recognizes(gvk)
 }
 
-func didNonMetadataFieldChange(old runtime.Object, new runtime.Object) bool {
-	oldUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
-	newUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(new)
+func didNonMetadataFieldChange(old runtime.Object, new runtime.Object) (bool, error) {
+	oldUnstructured, err := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
+	if err != nil {
+		return false, fmt.Errorf("can't convert %v to unstructued: %w", old, err)
+	}
+
+	newUnstructured, err := runtime.DefaultUnstructuredConverter.ToUnstructured(new)
+	if err != nil {
+		return false, fmt.Errorf("can't convert %v to unstructued: %w", new, err)
+	}
 
 	oldCopyContent := copyNonMetadata(oldUnstructured)
 	newCopyContent := copyNonMetadata(newUnstructured)
 
-	return !apiequality.Semantic.DeepEqual(oldCopyContent, newCopyContent)
+	return !apiequality.Semantic.DeepEqual(oldCopyContent, newCopyContent), nil
 }
 
 func copyNonMetadata(original map[string]interface{}) map[string]interface{} {
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -279,7 +279,19 @@ func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Ob
 		return apierrors.NewBadRequest("resourceVersion can not be set for Create requests")
 	}
 	accessor.SetResourceVersion("1")
-	accessor.SetGeneration(1)
+
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	if gvk.Empty() {
+		gvk, err = apiutil.GVKForObject(obj, t.scheme)
+		if err != nil {
+			return err
+		}
+	}
+
+	if isCustomResource(gvk) {
+		accessor.SetGeneration(1)
+	}
+
 	obj, err = convertFromUnstructuredIfNecessary(t.scheme, obj)
 	if err != nil {
 		return err
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -44,6 +44,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 )
 
 type versionedTracker struct {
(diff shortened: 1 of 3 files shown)


danaradg:
Tagging @FillZpp, @alvaroaleman as well as they corresponded with me on #1857.

pkg/client/fake/client.go
```diff
@@ -41,6 +41,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/internal/objectutil"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
```
FillZpp, line 45: Please sort the imports

pkg/client/fake/client.go
```diff
@@ -319,6 +321,13 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	}
 	intResourceVersion++
 	accessor.SetResourceVersion(strconv.FormatUint(intResourceVersion, 10))
+
+	// The behavior of the generation field is inconsistent across built-in k8s resources, but it's
+	// consistent for CRs. CRs have their generation increased on non-metadata changes only.
+	if isCustomResource(gvk) && didNonMetadataFieldChange(oldObject, obj) {
+		accessor.SetGeneration(oldAccessor.GetGeneration() + 1)
```
FillZpp, line 328: If we are going to enable `Generation` in fakeClient, you should also set it to `1` when object created.
See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L118

pkg/client/fake/client.go
danaradg: Done

pkg/client/fake/client.go
danaradg: Done

pkg/client/fake/client.go
```diff
@@ -329,6 +338,31 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
...
+
+func didNonMetadataFieldChange(old runtime.Object, new runtime.Object) bool {
+	oldUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
+	newUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(new)
+
+	oldCopyContent := copyNonMetadata(oldUnstructured)
+	newCopyContent := copyNonMetadata(newUnstructured)
+
+	return !apiequality.Semantic.DeepEqual(oldCopyContent, newCopyContent)
+}
```
alvaroaleman, line 353:
```suggestion
func didNonMetadataFieldChange(old runtime.Object, new runtime.Object) bool {
 oldCopy, newCopy := old.DeepCopy(), new.DeepCop()
 oldCopy.Metadata = metav1.ObjectMeta{}
 newCopy.Metadata = metav1.ObjectMeta{}
 return !reflect.DeepEqual(oldCopy, newCopy)
}
```

pkg/client/fake/client.go
```diff
@@ -227,6 +228,7 @@ func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Ob
 		return apierrors.NewBadRequest("resourceVersion can not be set for Create requests")
 	}
 	accessor.SetResourceVersion("1")
+	accessor.SetGeneration(1)
```
alvaroaleman, line 231: Why do we do this unconditional rather than checking if it is a CRD?

alvaroaleman (commented):
The change seems generally to be ok, but it is definitely breaking

pkg/client/fake/client.go
danaradg: This approach doesn't work, as you can't access the `Metadata` from the `runtime.Object` interface. Reverting the refactor.

FillZpp:
Generally LG, could you please add test cases for structured, unstructured and metadata objects?

danaradg:
> Generally LG, could you please add test cases for structured, unstructured and metadata objects?

Can you clarify? The `Update` method receives a `runtime.Object` and converts it to unstructured behind the scenes to check equality.
I don't understand the test cases that you want me to add.
Isn't structured object the same as what I'm already testing (a dummy custom resource)?
Unstructured objects are not `runtime.Object`s. I could make an object that implements both, but from the tested code's viewpoint it would be the same as passing any other `runtime.Object`.
What does a metadata object mean? `metav1.ObjectMeta`? It doesn't implement `runtime.Object`.

danaradg:
Hey @FillZpp. Stull waiting for an answer about the tests. Would very much like to merge the PR.

FillZpp:
> Can you clarify? The Update method receives a runtime.Object and converts it to unstructured behind the scenes to check equality.

Oh, I mean the runtime.Object could be a typed CR object or an unstructured CR object. The later one should also have a test case.

danaradg:
@FillZpp Done :)

danaradg:
Ping @alvaroaleman, @FillZpp. Can this be merged?

pkg/client/fake/client.go
```diff
@@ -380,6 +401,31 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	return t.ObjectTracker.Update(gvr, obj, ns)
 }
 
+func isCustomResource(gvk schema.GroupVersionKind) bool {
+	return !scheme.Scheme.Recognizes(gvk)
+}
+
+func didNonMetadataFieldChange(old runtime.Object, new runtime.Object) bool {
+	oldUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
+	newUnstructured, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(new)
```
alvaroaleman, line 410: Why are we ignoring errors here?

danaradg: My bad. Added error handling.

pkg/client/fake/client.go
```diff
@@ -370,6 +384,18 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	}
 	intResourceVersion++
 	accessor.SetResourceVersion(strconv.FormatUint(intResourceVersion, 10))
+
+	crRequiresGenerationUpdate, err := didNonMetadataFieldChange(oldObject, obj)
```
alvaroaleman, line 388: @danaradg we need the same for Patch or not?

sbueringer:
I'll close this PR given no recent activity.

@konkit If you're still interested, feel free to open a new PR to continue this work

/close

PR #2005, opened 2022-09-17
:seedling: add preWatch options for leader-elections runnables to reduce failover time when leader changes

nicktming:
### What does this do, and why do we need it? 
As we know leader-elections runnables only have one working instance, when the leader instance changes, the new leader should watch all resources; 

The problem is that watching resources takes long time in the big clusters, and this will have a affect for components who cares about the failover time。

This PR is used to provide a preWatch option for leader-election runnables to preWatch kubernetes resources (it does not handle resource events) for instances which does not acquire leader,  when the instance become leader, it starts to process events;

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -300,6 +300,9 @@ type Options struct {
 	newResourceLock        func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error)
 	newMetricsListener     func(addr string) (net.Listener, error)
 	newHealthProbeListener func(addr string) (net.Listener, error)
+
+	// preWatch resources for leader-election runnables
+	PreWatch bool
 }
 
 // BaseContextFunc is a function used to provide a base Context to Runnables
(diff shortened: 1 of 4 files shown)


nicktming:
Hey @mengqiy  @varshaprasad96, what do you think?  

thanks for your suggestions.

nicktming:
Hey @mengqiy  @varshaprasad96 , what do you think?

pkg/manager/manager.go
```diff
@@ -316,6 +319,12 @@ type Runnable interface {
 	Start(context.Context) error
 }
 
+// runnable with preWatch resources
+type RunnableWithPreWatch interface {
```
alvaroaleman, line 323: I don't really understand why this would be needed. We already have the `LeaderElectionRunnable` interace to start thigns before the leader lock was acquired: https://github.com/kubernetes-sigs/controller-runtime/blob/8ad090e0663e046c5a9dad68a572af0e30626c85/pkg/manager/manager.go#L330

Moreover, we start the cache before we start anything else: https://github.com/kubernetes-sigs/controller-runtime/blob/8ad090e0663e046c5a9dad68a572af0e30626c85/pkg/manager/internal.go#L466

We don't star the Source itself which might lead to us not getting an informer which then might mean we don't actually start a watch. But changing that is probably a breaking change, as it will lead to increased memory consumption on non-leader instances.

nicktming: @alvaroaleman  thanks for the discussion。 

it has the obvious problem that  it will lead to increased memory consumption on non-leader instances, so i provider one option for users who need this feature.  

in our env(big clusters) and when leader changes, it takes too long time to list resources from apiserver, so we want to save failover time by memory consumption.

PR #2017, opened 2022-10-04
📖 Advertise struct pointer requirement for Create

matteoolivi:
Make it explicit in the function comments that the "obj" argument of the Client interface's Create method must be a struct pointer, because implementers will store in it the object returned by the K8s API server.

The added statement was already present on `Update` and `Patch` methods, but not on `Create`.

**Note**:This comment will end up in the godoc of the package, but I couldn't find the godocs in the repo, so I don't know how to update the godocs.

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -60,7 +60,8 @@ type Reader interface {
 
 // Writer knows how to create, delete, and update Kubernetes objects.
 type Writer interface {
-	// Create saves the object obj in the Kubernetes cluster.
+	// Create saves the object obj in the Kubernetes cluster. obj must be a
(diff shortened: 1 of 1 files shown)


sbueringer:
> This comment will end up in the godoc of the package, but I couldn't find the godocs in the repo, so I don't know how to update the godocs.

The generated godocs pages are just generated by however hosts them (e.g. https://pkg.go.dev/), so they are not stored in the repo.


/lgtm
/ok-to-test

sbueringer:
No I think this should be fine. I'm just not a maintainer/approver so I can only lgtm

alvaroaleman (approved):

PR #2023, opened 2022-10-17
✨ Add tls options to manager.Options

### LINKED ISSUE #2020: Expose TLS configuration options for default webhook server creation
Similar to the flags for Webhook [Port and Host](https://github.com/kubernetes-sigs/controller-runtime/blob/2f77235e25b1e42d9e4957199f3cd0f2c3fb0d72/pkg/manager/manager.go#L212-L217), we should also expose flags for setting the minimum TLS version as well as the cipher suites to be used when creating the default webhook server.

Currently, to set the minimum TLS version, a hack would be to call `manager.GetWebhookServer().TLSMinVersion` before calling `manager.Start()`. To keep the behavior consistent, it makes sense to introduce new options in the `manager.Options` struct which allow doing the same.

sbueringer:
Fixes #2020

I hope I found all relevant places.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -18,6 +18,7 @@ package manager
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
@@ -242,6 +243,9 @@ type Options struct {
 	// It is used to set webhook.Server.CertDir if WebhookServer is not set.
 	CertDir string
 
+	// TLSOpts is used to allow configuring the TLS config used for the webhook server.
+	TLSOpts []func(*tls.Config)
+
 	// WebhookServer is an externally configured webhook.Server. By default,
 	// a Manager will create a default server using Port, Host, and CertDir;
 	// if this is set, the Manager will use this server instead.
@@ -422,6 +426,7 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		port:                          options.Port,
 		host:                          options.Host,
 		certDir:                       options.CertDir,
+		tlsOpts:                       options.TLSOpts,
 		webhookServer:                 options.WebhookServer,
 		leaseDuration:                 *options.LeaseDuration,
 		renewDeadline:                 *options.RenewDeadline,
--- a/pkg/envtest/webhook_test.go
+++ b/pkg/envtest/webhook_test.go
@@ -18,6 +18,7 @@ package envtest
 
 import (
 	"context"
+	"crypto/tls"
 	"path/filepath"
 	"time"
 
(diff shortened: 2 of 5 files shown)


pkg/manager/manager.go
```diff
@@ -490,6 +501,14 @@ func (o Options) AndFrom(loader config.ControllerManagerConfiguration) (Options,
 		o.CertDir = newObj.Webhook.CertDir
 	}
```
sbueringer, line 503: > pkg/manager/manager.go:453:1: cyclomatic complexity 31 of func `(Options).AndFrom` is high (> 30) (gocyclo)

Should I split up the func? Not sure if there's a good alternative to splitting up

pkg/config/v1alpha1/types.go
```diff
@@ -140,6 +140,10 @@ type ControllerWebhook struct {
 	// must be named tls.key and tls.crt, respectively.
 	// +optional
 	CertDir string `json:"certDir,omitempty"`
+
+	// TLSVersion is the minimum version of TLS supported. Accepts
+	// "", "1.0", "1.1", "1.2" and "1.3" only ("" is equivalent to "1.0" for backwards compatibility)
+	TLSMinVersion string
```
sbueringer, line 146: I couldn't add TLSOpts because of its type
```bash
hack/tools/bin/controller-gen object paths="./pkg/config/v1alpha1/...;./examples/configfile/custom/v1alpha1/..."
sigs.k8s.io/controller-runtime/pkg/config/v1alpha1:-: name requested for invalid type: func(*crypto/tls.Config)
sigs.k8s.io/controller-runtime/pkg/config/v1alpha1:-: invalid slice element type: func(*crypto/tls.Config)
137:27: missing ',' in parameter list
Error: not all generators ran successfully
```

Not sure if we want to add TLSMinVersion to this type (as I did). Opinions?

sbueringer:
/assign @vincepri 
(cc @alvaroaleman)

pkg/manager/manager.go
```diff
@@ -242,6 +244,13 @@ type Options struct {
 	// It is used to set webhook.Server.CertDir if WebhookServer is not set.
 	CertDir string
 
+	// TLSMinVersion is the minimum version of TLS supported by the webhook server.
+	// Accepts "", "1.0", "1.1", "1.2" and "1.3" only ("" is equivalent to "1.0" for backwards compatibility).
+	TLSMinVersion string
+
+	// TLSOpts is used to allow configuring the TLS config used for the webhook server.
+	TLSOpts []func(*tls.Config)
```
alvaroaleman, line 252: Why add both the `TLSOpts` and the `TLSMinVersion`?

sbueringer: Good point. I basically just added both because we have both on the Webhook Server.

But of course TLSOpts can be used to set the min version as well. For me TLSOpts would be enough.

Nevertheless, wondering if we want to have TLSMinVersion on `ControllerWebhook` (because there TLSOpts cannot be added).

vincepri: We should probably only keep a single field, in this case sounds like TLSOpts would be the way to go, given that it allows more customization?

alvaroaleman (approved):
thanks!

sbueringer:
@alvaroaleman Thx! 
 would it be okay to cherry pick this PR?

alvaroaleman:
> would it be okay to cherry pick this PR?
sgtm

Issue #1948, opened 2022-07-01
Add support for fieldSelector to fakeClient


grzesuav:
Currently it seems only `labelSelector` is supported - https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L423

matteoolivi:
If there's consensus from the project maintainers me and @pleic would like to open a PR to add this, since we have code that needs it. If there's not immediate consensus (e.g. because a real API server should be used with `envtest` rather than the fake client) we'll elaborate more on our use case and try to make a case for this.

Our rough idea to implement this was to:
1. add a `WithIndex(gvk schema.GroupVersionKind, IndexName, IndexFunc)` function to the [fake client builder](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L89).
2. add a field of type `map[schema.GroupVersionResource]map[string]IndexFunc` that stores all the Indexes that have been registered with the fake client on a per `GroupVersionResource` basis.
3. when a List requests specifies a field selector, we intercept the normal list results and filter it by lazily computing all the relevant indexes and dropping the list items that don't satisfy the selector.

cc @alvaroaleman @vincepri since you reviewed the previous attempt at this: https://github.com/kubernetes-sigs/controller-runtime/pull/800

alvaroaleman:
sounds good to me

matteoolivi:
PR opened: https://github.com/kubernetes-sigs/controller-runtime/pull/2025

alvaroaleman:
Fixed in https://github.com/kubernetes-sigs/controller-runtime/pull/2025

PR #2025, opened 2022-10-18
✨ Add indexes to fake client to mimic field selectors

matteoolivi:
Allow registration of indexes into the fake client to allow usage of field selectors when doing a List. The indexing is done lazily by doing a normal list and then filtering the results by computing the index value for each list item.

To enable the main change for this commit, some refactorings are performed: an internal and unexported function that checks whether a field selector is in the form key=val or key==val is made internal and exported to be shared between real and faked code to ensure loyalty. Unit tests for it are added.

Co-authored-by: Paul Eichler <peichler@anynines.com>

This PR addresses https://github.com/kubernetes-sigs/controller-runtime/issues/1948#issuecomment-1275389911 and is a reincarnation of https://github.com/kubernetes-sigs/controller-runtime/pull/800

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -56,9 +56,9 @@ type fakeClient struct {
 	scheme     *runtime.Scheme
 	restMapper meta.RESTMapper
 
-	// indexes maps each GroupVersionResource (GVR) to the indexes registered for that GVR.
+	// indexes maps each GroupVersionKind (GVK) to the indexes registered for that GVK.
 	// The inner map maps from index name to IndexerFunc.
-	indexes map[schema.GroupVersionResource]map[string]client.IndexerFunc
+	indexes map[schema.GroupVersionKind]map[string]client.IndexerFunc
 
 	schemeWriteLock sync.Mutex
 }
@@ -102,9 +102,9 @@ type ClientBuilder struct {
 	initRuntimeObjects []runtime.Object
 	objectTracker      testing.ObjectTracker
 
-	// indexes maps each GroupVersionResource (GVR) to the indexes registered for that GVR.
+	// indexes maps each GroupVersionKind (GVK) to the indexes registered for that GVK.
 	// The inner map maps from index name to IndexerFunc.
-	indexes map[schema.GroupVersionResource]map[string]client.IndexerFunc
+	indexes map[schema.GroupVersionKind]map[string]client.IndexerFunc
 }
 
 // WithScheme sets this builder's internal scheme.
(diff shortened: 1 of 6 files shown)


pkg/client/fake/client.go
```diff
@@ -420,21 +458,92 @@ func (c *fakeClient) List(ctx context.Context, obj client.ObjectList, opts ...cl
...
+	// behavior of the cache reader (which we're faking here).
+	fieldKey, fieldVal, requiresExact := selector.RequiresExactMatch(fs)
+	if !requiresExact {
+		return nil, fmt.Errorf("field selector %s is not in one of the two supported forms \"key==val\" or \"key=val\"",
+			fs)
+	}
+
+	// Field selection is mimicked via indexes, so there's no sane answer this function can give
+	// if there are no indexes registered for the GroupVersionResource of the objects in the list.
+	indexes, listGVRHasIndexes := c.indexes[gvr]
```
alvaroaleman, line 514: I think we should return the same error for this and the second error condition, for the user of this it is the same mistake so getting one or the other might be confusing because what they have to change is the same in both cases
```suggestion
	if indexes  := c.indexes[gvr]; indexers == nil || indexers[fieldKey] == nil {
	     return nil, fmt.Errorf(...
	}
```

pkg/client/fake/client_test.go
```diff
@@ -86,7 +100,7 @@ var _ = Describe("Fake client", func() {
 		}
 	})
 
-	AssertClientBehavior := func() {
+	AssertClientWOIndexBehavior := func() {
```
alvaroaleman, line 103:
```suggestion
	AssertClientWithoutIndexBehavior := func() {
```

alvaroaleman (commented):
/ok-to-test

pkg/client/fake/client_test.go
matteoolivi: Done in 5db9d19

pkg/client/fake/client.go
matteoolivi: Done in https://github.com/kubernetes-sigs/controller-runtime/commit/5db9d19a2a67ad34eace05ff974653f377410d19

pkg/client/fake/client.go
```diff
@@ -135,6 +147,31 @@ func (f *ClientBuilder) WithObjectTracker(ot testing.ObjectTracker) *ClientBuild
 	return f
 }
 
+// WithIndex can be optionally used to register an index with name `name` and indexer `indexer` for
+// API objects of GroupVersionResource `gvr` in the fake client.
+// It can be invoked multiple times, both with different GroupVersionResource or the same one.
+// Invoking WithIndex twice with the same `name` and `gvr` will panic.
+func (f *ClientBuilder) WithIndex(gvr schema.GroupVersionResource, name string, indexer client.IndexerFunc) *ClientBuilder {
```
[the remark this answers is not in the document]
matteoolivi: Done in https://github.com/kubernetes-sigs/controller-runtime/commit/5db9d19a2a67ad34eace05ff974653f377410d19

matteoolivi:
@alvaroaleman I pushed a separate commit to address your change requests.

alvaroaleman (approved):
thanks!

alvaroaleman:
looks like we have some flaky tests

/retest

Issue #2026, opened 2022-10-19
Allow to pass a custom resourcelock.Interface for leader election


johscheuer:
The current controller-runtime supports some customisation for the leader election by providing/setting the according `LeaderElection..` options. It would be nice if a user could provide their own `resourcelock.Interface` to be more flexible and allow the user to create different customisations of `resourcelock.Interface` that are currently not supported.

alvaroaleman:
Could you expand on what your use case is/What you want to to configure that you can't and why?

johscheuer:
Sure the https://pkg.go.dev/k8s.io/client-go/tools/leaderelection/resourcelock#MultiLock supports to provide two Interfaces and it's mostly used for a transition path e.g. when using configmaps and leases at the same time but you can also use that mechanism to get a lease in two different Kubernetes clusters e.g. if you run an operator across multiple Kubernetes clusters. Another benefit would be that this change allows to implement a custom Interface that e.g. uses something else than Kubernetes as a lease mechanism.

PR #2027, opened 2022-10-19
:sparkles: Allow to provide a custom lock interface to manager

johscheuer:
Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/2026 

This allows a more flexible configuration of the controller-runtime leader election if needed. Those changes allow to provide a custom `resourcelock.Interface` that will be used for leader election.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -193,6 +193,12 @@ type Options struct {
 	// LeaseDuration time first.
 	LeaderElectionReleaseOnCancel bool
 
+	// LeaderElectionResourceLockInterface allows to provide a custom resourcelock.Interface that was created outside
+	// of the controller-runtime. If this value is set the options LeaderElectionID, LeaderElectionNamespace,
+	// LeaderElectionResourceLock, LeaseDuration, RenewDeadline and RetryPeriod will be ignored. This can be useful if you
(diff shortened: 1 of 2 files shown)


pkg/manager/manager.go
```diff
@@ -377,14 +383,19 @@ func New(config *rest.Config, options Options) (Manager, error) {
...
-		LeaderElectionResourceLock: options.LeaderElectionResourceLock,
-		LeaderElectionID:           options.LeaderElectionID,
-		LeaderElectionNamespace:    options.LeaderElectionNamespace,
-	})
-	if err != nil {
-		return nil, err
+	var resourceLock resourcelock.Interface
+	if options.LeaderElectionResourceLockInterface != nil && options.LeaderElection {
+		resourceLock = options.LeaderElectionResourceLockInterface
+	} else {
```
alvaroaleman, line 389: If `options.LeaderElectionResourceLockInterface` is nil we will go into the `else` clause even if `options.LeaderElection` is false

alvaroaleman (commented):
/ok-to-test

pkg/manager/manager.go
alvaroaleman: ah nvm, missed the fact that we only construct it her and whether or not we use it is determined elsewhere

alvaroaleman (approved):

PR #2031, opened 2022-10-24
:sparkles: feat: add NOT predicate

erikgb:
This PR adds a new boolean NOT predicate function (`predicate.Not()`) to complement `predicate.And()` and `predicate.Or()` added in https://github.com/kubernetes-sigs/controller-runtime/pull/1043.

While not extremely useful, I think it can make code for bootstrapping controllers easier to read. And avoid having to create a dedicated predicate just for selecting the complement set selected by another predicate.

I have also seen a few examples of this implemented as a utility in publicly available controllers.


alvaroaleman (approved):
thanks!

PR #2036, opened 2022-11-03
✨Feat implement new NamespacePredicate NamespaceNamePredicate

### LINKED ISSUE #1866: Support for NamespacedName predicate

halfcrazy:
This pr adds new predicates `NamespacePredicate` and `NamespaceNamePredicate`.  
Instead of filtering in the custom cache, it provides a new way to do so in the reconciler.

Fix #1866

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -48,6 +48,8 @@ var _ Predicate = Funcs{}
 var _ Predicate = ResourceVersionChangedPredicate{}
 var _ Predicate = GenerationChangedPredicate{}
 var _ Predicate = AnnotationChangedPredicate{}
+var _ Predicate = NamespacePredicate{}
+var _ Predicate = NamespaceNamePredicate{}
 var _ Predicate = or{}
 var _ Predicate = and{}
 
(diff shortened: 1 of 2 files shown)


pkg/predicate/predicate.go
```diff
@@ -48,6 +48,8 @@ var _ Predicate = Funcs{}
 var _ Predicate = ResourceVersionChangedPredicate{}
 var _ Predicate = GenerationChangedPredicate{}
 var _ Predicate = AnnotationChangedPredicate{}
+var _ Predicate = NamespacePredicate{}
```
alvaroaleman, line 51: You can achieve the same with 

```
NewPredicateFuncs(func(o client.Object) bool {
   return o.GetNamespace() == "my-namespace" 
}
```

Rather than adding so many predicates I think it would be preferrable to encourage ppl to write their own.

alvaroaleman: is `NewPredicateFuncs` not well documented or hard to find or what prompted you to implement this?

halfcrazy: Oh， I come from `kubebuilder`. This method is not mentioned in `kubebuilder` docs, so I may miss it.

PR #2039, opened 2022-11-08
🐛 fix: improve semantics of combining cache selectorsByObject

joelanford:
This PR follows up on #1980, which introduced a regression in the way `cache.BuilderWithOptions` deals with scheme selection when handling selectors.

When handling inherited selectors, we should only use the inherited scheme. However, when handling the selectors in the cache builder options, we should use the combined scheme. This ensures that callers are not forced to pass their parent `runtime.Scheme` down to every `cache.BuilderWithOptions` call.

#1980 has not yet been released, so this bug fix does not require any backport.


joelanford:
Failure seems like a flake, unrelated to this PR...

alvaroaleman (approved):

Issue #1884, opened 2022-05-03
Add and remove watches at runtime


jcanseco:
Opening a new issue to raise visibility of [my question here](https://github.com/kubernetes-sigs/controller-runtime/issues/246#issuecomment-965988777).

We have a very similar use-case as that linked issue, and we'd like to know how one can actually remove watches at runtime.

FillZpp:
@jcanseco Could you please describe your use-case that needs to remove watches(informers) dynamically?

jcanseco:
@FillZpp sure.

Suppose you have a Resource A that depends on Resource B. We'd like to immediately reconcile Resource A once Resource B is Ready.

We were hoping to do that by having the Resource A Controller create a Watch on Resource B, so that we can enqueue Resource A for immediate reconciliation once Resource B is ready. Once done, we'd like to remove the Watch since it's no longer needed.

Please let us know if there is a better way to resolve this problem as well.

FillZpp:
Are A and B the resource kinds or the object names?

I suppose they are the resource kinds. So you firstly watch two kinds, but you only care about **ONE object** of the kind B. After watching the only object of kind B, your controller triggers to reconcile kind A and no longer needed to watch kind B.

Do I understand correctly?

jcanseco:
Ah yes apologies for not being precise enough. That is correct.

To be particularly precise: Resource A and B are two particular objects (i.e. uniquely identified by the combination of their GVK, Namespace, and Name).

Typically in our case, the GVKs of Resource A and B are different.

FillZpp:
Understand. Another question is, is there only one particular object of Resource B? Or there are multiple objects of Resource B, but you only need to watch the particular one of them?

How about this:
- If there is only one object of Resource B, I think you don't have to stop and remove its informer.
- If there are multiple objects, you could use [SelectorByObject](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/cache.go#L120) to make sure your cache only watch the particular object and then don't have to stop and remove it.

jcanseco:
By Resource B, are you referring to just the GVK? If so, yes, there can be multiple objects with GVK B, but we only need to watch one particular object (i.e. a particular GVK + Namespace + Name combination).

I am not sure I understand the suggestion -- can you elaborate?

[...]

PR #2043, opened 2022-11-11
⚠ Bump K8s v0.26.0-alpha.3

FillZpp:

Bump K8s v0.26.0-alpha.3

Note that because of the new removal of event handler in K8s 1.26 https://github.com/kubernetes/kubernetes/pull/111122, this PR is a breaking change.

After it merged, I will post a new PR to support removal of watches/controllers for https://github.com/kubernetes-sigs/controller-runtime/issues/1884.


pkg/cache/cache.go
```diff
@@ -78,11 +78,19 @@ type Informer interface {
 	// AddEventHandler adds an event handler to the shared informer using the shared informer's resync
 	// period.  Events to a single handler are delivered sequentially, but there is no coordination
 	// between different handlers.
-	AddEventHandler(handler toolscache.ResourceEventHandler)
+	// It returns a registration handle for the handler that can be used to remove
+	// the handler again.
+	AddEventHandler(handler toolscache.ResourceEventHandler) (toolscache.ResourceEventHandlerRegistration, error)
```
alvaroaleman, line 83: @FillZpp can we maybe have to distinct PRs, one that just bumps the deps and a follow-up that changes the interfaces so that our release note tooling will be able to properly call out the EventHandler change?

FillZpp: Sure, I can do this, but the first one will fail to run golangci-lint and some other tests. Is it ok?

alvaroaleman: Maybe just use a tiny wrapper around the upstream `AddEventHandler` that just does `_, _ = AddEventHandler(handler)`?

alvaroaleman (approved):
thanks!

PR #2045, opened 2022-11-15
✨ add EnableRateLimiting to EnqueueRequestForObject handler

jakobmoellersap:
This allows controllers using EnqueueRequestForObject to enable rate-limited adding of a resource via EnableRateLimiting flag. By default it is set to off, resulting in no compatibility changes.


alvaroaleman:
@jakobmoellersap its the first time I see someone asking for this. Can you please implement an eventhandler downstream instead?

PR #2051, opened 2022-11-21
:bug: Missing error log for in-cluster config

maciej-karas-form3:
Signed-off-by: Maciej Karaś [maciej.karas@form3.tech](mailto:maciej.karas@form3.tech)

closes https://github.com/kubernetes-sigs/controller-runtime/issues/2053

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -113,7 +113,7 @@ func GetConfigWithContext(context string) (*rest.Config, error) {
 var loadInClusterConfig = rest.InClusterConfig
 
 // loadConfig loads a REST Config as per the rules specified in GetConfig.
-func loadConfig(context string) (*rest.Config, error) {
+func loadConfig(context string) (config *rest.Config, configErr error) {
 	// If a flag is specified with the config location, use that
 	if len(kubeconfig) > 0 {
 		return loadConfigWithContext("", &clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfig}, context)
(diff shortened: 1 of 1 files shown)


pkg/client/config/config.go
```diff
@@ -125,6 +125,8 @@ func loadConfig(context string) (*rest.Config, error) {
 	if len(kubeconfigPath) == 0 {
 		if c, err := loadInClusterConfig(); err == nil {
 			return c, nil
+		} else {
+			log.Error(err, "unable to load in-cluster config")
```
FillZpp, line 129: It's just a try, for example it may return `ErrNotInCluster` if it is not running in a cluster, so we'd better not log it as an error.

maciej-karas-form3: @FillZpp Your'e right. I don't know the conventions for logging warnings in this codebase. Are we using WARN or just [INFO,ERROR]? I can change it to info if that's the case. Will this work?
```go
log.Info("unable to load in-cluster config: %s", err.String())
```

alvaroaleman: I would probably use a `defer` that logs this only if `loadConfig` returns an error

maciej-karas-form3: @alvaroaleman @FillZpp updated the PR accordingly. Now it only logs error when the `loadConfig` method returns an error

alvaroaleman (approved):
Thanks!

maciej-karas-form3:
@alvaroaleman The build failed with some lint errors. I've fixed them but the `lgmt` tag was removed

pkg/client/config/config.go
```diff
@@ -123,9 +123,16 @@ func loadConfig(context string) (*rest.Config, error) {
...
 	if len(kubeconfigPath) == 0 {
-		if c, err := loadInClusterConfig(); err == nil {
+		c, err := loadInClusterConfig()
+		if err == nil {
 			return c, nil
 		}
+
+		defer func() {
+			if configErr != nil {
+				log.Error(err, "unable to load in-cluster config")
```
sbueringer, line 133: @alvaroaleman 
Q: Does this really work as expected?

I think it could potentially pick up errors produced later in the func in l. 150 or l.155 

Wondering if we could simply use `if err != nil` to log this error instead

PR #2054, opened 2022-11-21
:sparkles: Add cluster.NewClientFunc with options

erikgb:
This adds a new `cluster.ClientBuilderWithOptions` alongside the `cluster.DefaultNewClient`, which will allow users to easily get a client that caches `Unstructured` - which the default client intentionally doesn't, ref https://github.com/kubernetes-sigs/controller-runtime/issues/615#issuecomment-536769176.

Almost all the operators I work on seem to benefit from enabling caching of `Unstructured`. While not "that" common, I have also seen users asking for this on Slack. 😉 Adding this new func to controller-runtime will avoid having to implement the workaround explained in [this blog](https://ymmt2005.hatenablog.com/entry/2021/07/25/Caching_Unstructured_Objects_using_controller-runtime).

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -112,6 +112,7 @@ type Options struct {
 	// NewClient is the func that creates the client to be used by the manager.
 	// If not set this will create the default DelegatingClient that will
 	// use the cache for reads and the client for writes.
+	// NOTE: The default client will not cache Unstructured.
 	NewClient NewClientFunc
 
 	// ClientDisableCacheFor tells the client that, if any cache is used, to bypass it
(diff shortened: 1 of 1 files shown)


varshaprasad96 (commented):
I understand the use-case of caching unstructured objects especially since most of them are CRs in controllers. But going through the discussion in the [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/615#issuecomment-536769176), we would still be hitting the same issue of caching all the arbitrary objects accidentally. The evolution of uncached client was that we initially used to hit the API server to read all the unstructured objects (https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.5/pkg/client/split.go#L45-L61) which then turned out to be expensive. Which is why the option of specifying uncached objects was introduced so that we could hit the api server only when needed (https://github.com/kubernetes-sigs/controller-runtime/blob/accd2623e14e6d4c4861ff76e7af40001e640f6b/pkg/client/split.go#L115). IMO I would still not want to cache all the unstructured objects by default (accidentally caching un-needed objects) or turn that option "on" for users who aren't aware of what happens inside the hood. From a users perspective, I feel its easier to figure out and understand that there is throttling happening in API server's end (and fix it) than to figure out it's a caching problem while debugging performance issues.

erikgb:
But why are more concerned about this for `Unstructured` than for any other type? I've seen many questions on Slack asking why a controller needs cluster-wide RBAC to resources just because the cache/watch is auto-created for any read of in-tree resource types like secrets. I have no problem with the current default (not cache `Unstructured`), but IMO it should be easier for users to enable caching of `Unstructured`, and would like it to surface somehow. It's considered best practice to grant a controller/operator least-privilege permissions, so caching all resources in the cluster is unlikely IMO.

pkg/cluster/cluster.go
```diff
@@ -255,16 +256,28 @@ func setOptionsDefaults(options Options) Options {
 // NewClientFunc allows a user to define how to create a client.
 type NewClientFunc func(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error)
 
-// DefaultNewClient creates the default caching client.
+// DefaultNewClient creates the default caching client, that will not cache Unstructured.
```
alvaroaleman, line 259:
```suggestion
// DefaultNewClient creates the default caching client, that will never cache Unstructured.
```

pkg/cluster/cluster.go
```diff
@@ -255,16 +256,28 @@ func setOptionsDefaults(options Options) Options {
...
 func DefaultNewClient(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
+	return newClient(cache, config, options, false, uncachedObjects)
+}
+
+// CacheUnstructuredNewClient creates a caching client that will cache Unstructured.
+func CacheUnstructuredNewClient(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (client.Client, error) {
```
alvaroaleman, line 265: How about we do this similiar to what we do in the cache, where we have a `func BuilderWithOptions(options Options) NewCacheFunc {` to accomodate the case that ppl might want additional options in the future?

alvaroaleman: For example, we have for a long time discussed that the default behavior of the cache-backed client of automatically creating caches is not great, but it is extremely hard to change due to compatibility converns (ppl might only discover this change at runtime).

So having an option to allow ppl to opt into this would be nice

erikgb: @alvaroaleman Thanks a lot for the initial review and your suggestion. I have made an attempt to implement the suggested design with an extensible `cluster.ClientOptions` struct. Please take another look, and tell me what you think.

pkg/cluster/cluster.go
```diff
@@ -255,16 +256,35 @@ func setOptionsDefaults(options Options) Options {
...
-		Client:          c,
-		UncachedObjects: uncachedObjects,
-	})
+// ClientBuilderWithOptions returns a Client constructor that will build a client
+// honoring the options argument
+func ClientBuilderWithOptions(options ClientOptions) NewClientFunc {
+	return func(cache cache.Cache, config *rest.Config, clientOpts client.Options, uncachedObjects ...client.Object) (client.Client, error) {
+		if uncachedObjects != nil {
+			options.UncachedObjects = append(options.UncachedObjects, uncachedObjects...)
+		}
```
alvaroaleman, line 276: No reason for the conditional:

```suggestion
		options.UncachedObjects = append(options.UncachedObjects, uncachedObjects...)
```

erikgb: Thx! 😊

alvaroaleman (approved):

alvaroaleman:
thanks!

erikgb:
Thanks for the review @alvaroaleman. I really like your suggestion, which makes this change extensible for future improvements! 👍

PR #2065, opened 2022-11-25
:seedling: seedling to dependabot

kannon92:
I notice that the dependabot PRs were manually be changed to add :seedling: tag.  We can add that to the commit message.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,5 +23,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: ":seedling:"
     labels:
       - "ok-to-test"


alvaroaleman (approved):
@kannon92 thanks. It is also possible to exclude k8s.io/* dependencies? We manually bump those all together whenever upstream does releases.

kannon92:
> @kannon92 thanks. It is also possible to exclude k8s.io/* dependencies? We manually bump those all together whenever upstream does releases.

Yea, do you want me to add it as part of this PR?



I can add 

```
    ignore:
        # Avoid K8
      - dependency-name: "k8s.io/*"

I'll have to add the explict name but it is possible.

alvaroaleman:
@kannon92 that would be great, a follow-up is fine

Issue #172, opened 2018-10-11
Support for CRUD sub resources in the Client


shawn-hurley:
Currently, I do not see a way to use subresources from the client. 

I suggest that we add a variadic argument for subresources. example: https://github.com/kubernetes/client-go/blob/master/dynamic/interface.go#L32

Or we may need to add the subresources to the Options for functions that already have a variadic?

coderanger:
There is limited support for the status subresource via https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L143

liyinan926:
I have a similar use case here that needs to evict pods instead of simply deleting them, using the eviction subresource of Pod.

DirectXMan12:
/kind feature

We had some initial ideas for how to design this, but I think the interface needs potential a bit more work.  For instance, should we do `client.Subresource("name").Update(obj)`?  Can we discover `name` from the object?  (maybe?).  Thoughts and usecases always welcome :-)

shawn-hurley:
So your thinking that we have a `subresoureces` Client, that has an interface containing `Create`, `Update`, `Delete` and `Get` that look exactly the same as the current interface?

I don't understand how we could determine the subresources for the object, could you explain what you had in mind would love to look into this.

I am assuming that we would want to keep the StatusWriter because it is so common that the helper makes sense?

DirectXMan12:
Depending on the subresource, we could use a RESTMapper to convert a Kind to the corresponding subresource entry in discovery, IIRC, but I haven't confirmed that -- it's just an initial thought off the top of my head.  This doesn't work with subresources like status, though (even though we deal with status separately).  I'd be curious to see what seemed more readable or ergonomic.

jrthrawny:
Is this issue looking for help? I have a need for doing the client supporting subresources such as `get pods/log`

DirectXMan12:
yeah.  In general, anything in CR that doesn't have someone working on it is open for help -- I tend to apply the help-wanted tag a bit less in cases where the design might be a bit thorny or that might require serious apimachinery spelunking, but if you're up for some back and forth on design, I'm happy to have someone working on it :-)

cadedaniel:
[...]

PR #2072, opened 2022-11-28
:warning: Add client.SubResourceWriter

alvaroaleman:
This change adds a SubResourceWriter to the client package that can be used to Create, Update or Patch arbitrary subresources.

This fixes the majority of  https://github.com/kubernetes-sigs/controller-runtime/issues/172, the only thing missing for it to close would be to add subresource Get, as I think there is no existing usage of subresource Delete.

Relative to https://github.com/kubernetes-sigs/controller-runtime/pull/1922 the main changes are:
* Use distinct options for subresource Update and Patch which makes this a breaking change
* Add subresource Create, needed for example for ServiceAccount tokens

Supersedes and thereby closes https://github.com/kubernetes-sigs/controller-runtime/pull/1922
Supersedes and thereby closes https://github.com/kubernetes-sigs/controller-runtime/pull/2015

/cc @fillzpp @joelanford @sbueringer

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -82,20 +82,64 @@ type Writer interface {
 // StatusClient knows how to create a client which can update status subresource
 // for kubernetes objects.
 type StatusClient interface {
-	Status() StatusWriter
+	Status() SubResourceWriter
 }
 
-// StatusWriter knows how to update status subresource of a Kubernetes object.
-type StatusWriter interface {
+// SubResourceClient knows how to create a client which can update subresource
+// for kubernetes objects.
+type SubResourceClient interface {
+	// SubResource returns a subresource client for the named subResource. Known
+	// upstream subResources are:
+	// - ServiceAccount tokens:
+	//     sa := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "bar"}}
+	//     token := &authenticationv1.TokenRequest{}
+	//     c.SubResourceClient("token").Create(ctx, sa, token)
+	//
+	// - Pod evictions:
+	//     pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "bar"}}
+	//     c.SubResourceClient("eviction").Create(ctx, pod, &policyv1.Eviction{})
+	//
+	// - Pod bindings:
+	//     pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "bar"}}
+	//     binding := &corev1.Binding{Target: corev1.ObjectReference{Name: "my-node"}}
+	//     c.SubResourceClient("binding").Create(ctx, pod, binding)
+	//
+	// - CertificateSigningRequest approval:
+	//     csr := &certificatesv1.CertificateSigningRequest{
+	//	     ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "bar"},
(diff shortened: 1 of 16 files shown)


pkg/client/options.go
```diff
@@ -67,6 +67,24 @@ type DeleteAllOfOption interface {
...
 
+// SubResourceUpdateOption is some configuration that modifies options for a update request.
+type SubResourceUpdateOption interface {
+	// ApplyToUpdate applies this configuration to the given update options.
+	ApplyToSubResourceUpdate(*SubResourceUpdateOptions)
+}
+
+// SubResourceCreateOption is some configuration that modifies options for a create request.
+type SubResourceCreateOption interface {
+	// ApplyToCreate applies this configuration to the given create options.
```
padlar, line 78: fun name is not updated in the comment

pkg/client/options.go
```diff
@@ -67,6 +67,24 @@ type DeleteAllOfOption interface {
 	ApplyToDeleteAllOf(*DeleteAllOfOptions)
 }
 
+// SubResourceUpdateOption is some configuration that modifies options for a update request.
+type SubResourceUpdateOption interface {
+	// ApplyToUpdate applies this configuration to the given update options.
```
padlar, line 72: fun name not updated in comment

pkg/client/typed_client.go
```diff
@@ -168,8 +167,32 @@ func (c *typedClient) List(ctx context.Context, obj ObjectList, opts ...ListOpti
 		Into(obj)
 }
 
-// UpdateStatus used by StatusWriter to write status.
-func (c *typedClient) UpdateStatus(ctx context.Context, obj Object, opts ...UpdateOption) error {
+func (c *typedClient) CreateSubResource(ctx context.Context, obj Object, subResource Object, subResourceName string, opts ...SubResourceCreateOption) error {
```
FillZpp, line 170: `subResource` is an `Object` in `CreateSubResource` but a `string` in `UpdateSubResource` and `PatchSubResource`.
How about name it `subResourceObj Object` so that `subResource` can all be the string of name?

pkg/client/interfaces.go
```diff
@@ -82,27 +82,72 @@ type Writer interface {
...
 type StatusClient interface {
-	Status() StatusWriter
+	Status() SubResourceWriter
 }
 
-// StatusWriter knows how to update status subresource of a Kubernetes object.
-type StatusWriter interface {
+// SubResourceClient knows how to create a client which can update subresource
+// for kubernetes objects.
+type SubResourceClient interface {
```
FillZpp, line 90: Should we add a TODO comment for SubResourceReader and those non-CRUD subresources?

alvaroaleman: Not sure, I think by then we would rename the `SubResourceWriter`

FillZpp:
Whoops, didn't notice it is approved, maybe I should have added a /hold. Not sure if someone else has any comment.

andriisoldatenko:
In the original PR https://github.com/kubernetes-sigs/controller-runtime/pull/1922, there was a problem with updating the finalizer. 

Can someone please explain how to implement it with the new `client.SubResourceWriter`, assuming we want to update ONLY the finalizer of deployment and don't have permission to update the entire object?

Thanks in advance, and sorry for commenting merged PR.

sbueringer:
I dont' know how to do it, but I would recommend opening a new issue. If that is already possible directly with client-go we might be able to infer from that how it could be done with the controller-runtime client.

I think in any case it would be good to have an example / test covering that scenario

PR #2076, opened 2022-12-05
feat: supprot disable deepcopy on list funcion

qiankunli:
related issue: https://github.com/kubernetes-sigs/controller-runtime/pull/1274

If the all object of a specific GVK supports DisableDeepCopy option, the risk is too great. it may be a better way to specify DisableDeepCopy option every time list function is called


pkg/client/options.go
```diff
@@ -417,6 +417,11 @@ type ListOptions struct {
 	// it has expired. This field is not supported if watch is true in the Raw ListOptions.
 	Continue string
 
+	// disableDeepCopy indicates not to deep copy objects during list objects.
+	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
+	// otherwise you will mutate the object in the cache.
+	DisableDeepCopy bool
```
alvaroaleman, line 423:
```suggestion
	UnsafeDisableDeepCopy bool
```

pkg/cache/internal/cache_reader.go
```diff
@@ -161,7 +161,7 @@ func (c *CacheReader) List(_ context.Context, out client.ObjectList, opts ...cli
 		}
 
 		var outObj runtime.Object
-		if c.disableDeepCopy {
+		if c.disableDeepCopy || listOpts.DisableDeepCopy {
```
alvaroaleman, line 164: Please add a test for this, maybe by checking that the pointer for the obj is the same as for the one in the cache

alvaroaleman (commented):
/ok-to-test

pkg/cache/internal/cache_reader.go
qiankunli: > Please add a test for this, maybe by checking that the pointer for the obj is the same as for the one in the cache

done

vincepri (approved):
/lgtm

Issue #2083, opened 2022-12-07
manager never sets started boolean


terinjokes:
If the manager is started multiple times in error, subsequent calls to `Start` are supposed to return a friendly error message.

https://github.com/kubernetes-sigs/controller-runtime/blob/44c5d5029cc3c19bf6e7df3f5c5943977a39637c/pkg/manager/internal.go#L403-L408

There are also several other checks against `started`, such as preventing modifications to metrics, health checks, and readiness check endpoints while the HTTP server is doing concurrent reads.

https://github.com/kubernetes-sigs/controller-runtime/blob/44c5d5029cc3c19bf6e7df3f5c5943977a39637c/pkg/manager/internal.go#L221-L223
https://github.com/kubernetes-sigs/controller-runtime/blob/44c5d5029cc3c19bf6e7df3f5c5943977a39637c/pkg/manager/internal.go#L243-L245
https://github.com/kubernetes-sigs/controller-runtime/blob/44c5d5029cc3c19bf6e7df3f5c5943977a39637c/pkg/manager/internal.go#L260-L262

It looks like the manager's internals were refactored in 612e9b28858354c69696a91b0778f3440ce4687c, and in the process the code to set `started` was removed.

<details><summary>Example Program</summary><p>

```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"sigs.k8s.io/controller-runtime/pkg/client/config"
	"sigs.k8s.io/controller-runtime/pkg/manager"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	controller := manager.RunnableFunc(func(ctx context.Context) error {
		fmt.Println("controller started")
		<-ctx.Done()
		return nil
	})

	cfg := config.GetConfigOrDie()

	m, err := manager.New(cfg, manager.Options{})
	if err != nil {
		fmt.Printf("error creating manager: %s", err)
		os.Exit(1)
	}

	if err := m.Add(controller); err != nil {
		fmt.Printf("error add controller: %s", err)
		os.Exit(1)
	}

        // mimic a user accidentally calling Start on the manager twice.
	go func() {
		err := m.Start(ctx)
		fmt.Printf("err: %s\n", err)
	}()

	go func() {
		err := m.Start(ctx)
		fmt.Printf("err: %s\n", err)
	}()

	<-ctx.Done()
	fmt.Println("Done")
}
```

</p></details>
[...]

PR #2090, opened 2022-12-10
:bug: Prevent manager from getting started a second time

alvaroaleman:
Currently it is possible to start the mgr multiple times, which will cause all sorts of issues. Prevent that from happening.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2083


shafeeqes:
Currently we set `cm.started=true` at the very beginning? Is this intended? Shouldn't we rather do this after starting the webhooks, cache and all the runnables?
Context: We are facing some issues calling `AddReadyzCheck` inside a runnable `Start()` because, by the time the runnable is started by the manager, the `cm.started` boolean is already set to true, and it errors.

Issue #2052, opened 2022-11-21
Is there a way to add controller without For?


Noksa:
Hello!

This library is really great. Thanks for all what you guys are doing.

I have a question.

My case is that I don't want to watch a `kind` type. I just want to watch a channel (with ticks, for example).
Currently, it looks like we can't do it without a workaround.

Working code:
```
...
return ctrl.NewControllerManagedBy(mgr).
  For(&v1.ResourceQuota{}, predicates.IgnoreAllEvents()).
  Watches(&source.Channel{Source: r.tickChan}, handler.EnqueueRequestsFromMapFunc(tickRequest())). 
  WithEventFilter(p()).
  Complete(r)
...

func IgnoreAllEvents() builder.Predicates {
	return builder.WithPredicates(predicate.Funcs{
		CreateFunc: func(createEvent event.CreateEvent) bool {
			return false
		},
		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
			return false
		},
		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
			return false
		},
		GenericFunc: func(genericEvent event.GenericEvent) bool {
			return false
		},
	})
}
```

So I had to add `For(&v1.ResourceQuota{}, predicates.IgnoreAllEvents())` (I chose `ResourceQuota` because usually no one uses it, to ensure that there will be small count of events with this kind)

If I remove For (or replace it with `Watches`, I got an error that controller can't be created:
```
2022-11-21T23:06:30.552+0300    ERROR   setup   unable to create controller     {"controllers": "Garbage", "error": "must provide an object for reconciliation"}
main.main
```

It would be cool if we could run controller without having to check 'fake' kind resource type.

Thanks.

alvaroaleman:
I agree requiring `For` is an annoyance and sometimes undesired. Removing that requirement if `len(append(blder.ownsInput, blder.watchesInput...) > 0` sounds good to me.

karlhjm:
> I agree requiring `For` is an annoyance and sometimes undesired. Removing that requirement if `len(append(blder.ownsInput, blder.watchesInput...) > 0` sounds good to me.

Hello @alvaroaleman , it seems that this has been modified in 0.14. That's great.

But will this restriction be modified in the previous version? Due to the dependency version of golang, we are limited to using 0.11, so we cannot create the controller without calling the For() func.

PR #2091, opened 2022-12-11
✨Builder: Do not require For

alvaroaleman:
Requiring For does not make sense for all controllers as it is possible that their primary event triggering is not an object in the current cluster but for example an object in a different cluster or a source.Channel.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2052


vincepri (approved):
/approve
/lgtm

Issue #2001, opened 2022-09-15
support configuring `RecoverPanic` globally


mars1024:
Now `RecoverPanic` is only a field in `controller.Options`, if we want to enable this for all controllers, each controller initialization should be modified.

Since  `GroupKindConcurrency` and `CacheSyncTimeout` can both be set through global `v1alpha1.ControllerConfigurationSpec`, IMO, `RecoverPanic` can do the same.

The only left problem is that `RecoverPanic` is bool type which has no zero-value, so maybe it has to be changed to pointer type, but i'm afraid this will be a break change.

PR #2093, opened 2022-12-11
:warning: Allow configuring RecoverPanic for controllers globally

alvaroaleman:
This change allows configuring the RecoverPanic setting for controllers globally. It is a breaking change as it requires changing the field type of the setting to *bool from bool in order to be able to check if it has been set already.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2001

### DIFF
--- a/pkg/internal/controller/controller_test.go
+++ b/pkg/internal/controller/controller_test.go
@@ -33,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/pointer"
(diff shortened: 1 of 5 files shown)


pkg/internal/controller/controller_test.go
```diff
@@ -114,7 +114,7 @@ var _ = Describe("controller", func() {
 			defer func() {
 				Expect(recover()).To(BeNil())
 			}()
-			ctrl.RecoverPanic = true
+			ctrl.RecoverPanic = ptr(true)
```
vincepri, line 117:
```suggestion
			ctrl.RecoverPanic = pointer.Bool(true)
```
? 
Just so we keep using the same utilities all around

alvaroaleman: Updated, ptal

vincepri (approved):
/approve
/lgtm

PR #2099, opened 2022-12-14
⚠ Support shutdown controllers and watches dynamically

### LINKED ISSUE #1884: Add and remove watches at runtime

FillZpp:

Support shutdown controllers and watches dynamically.

API changes:
- An optional `ControllerCtx` added into `controller.Options` to let developer stop a specific controller and its watches.
- A `Stop()` method added into some stoppable sources, e.g., `Kind`, `KindWithCache`, `Informer`, to let developer only stop a specific watch.

fixes #1884

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -86,6 +86,10 @@ type Controller interface {
 
 	// GetLogger returns this controller logger prefilled with basic information.
 	GetLogger() logr.Logger
+
+	// Stop stops the controller and all its watches dynamically.
+	// Note that it will only trigger the stop but will not wait for them all stopped.
+	Stop() error
 }
 
 // New returns a new Controller registered with the Manager.  The Manager will ensure that shared Caches have
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -58,11 +57,17 @@ type Controller struct {
 	// the Queue for processing
 	Queue workqueue.RateLimitingInterface
 
+	// startedSources maintains a list of sources that have already started.
+	startedSources []source.Source
+
 	// mu is used to synchronize Controller setup
 	mu sync.Mutex
 
 	// Started is true if the Controller has been Started
 	Started bool
+
+	// Stopped is true if the Controller has been Stopped
+	Stopped bool
 
 	// ctx is the context that was passed to Start() and used when starting watches.
 	//
@@ -70,6 +75,9 @@ type Controller struct {
 	// while we usually always strive to follow best practices, we consider this a legacy case and it should
 	// undergo a major refactoring and redesign to allow for context to not be stored in a struct.
 	ctx context.Context
+
+	// cancel is the CancelFunc of the ctx, to stop the controller and its watches dynamically.
+	cancel context.CancelFunc
 
 	// CacheSyncTimeout refers to the time limit set on waiting for cache to sync
 	// Defaults to 2 minutes if not set.
@@ -123,6 +131,10 @@ func (c *Controller) Watch(src source.Source, evthdler handler.EventHandler, prc
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	if c.Stopped {
+		return fmt.Errorf("can not start watch in a stopped controller")
+	}
+
 	// Controller hasn't started yet, store the watches locally and return.
 	//
 	// These watches are going to be held on the controller struct until the manager or user calls Start(...).
@@ -132,7 +144,11 @@ func (c *Controller) Watch(src source.Source, evthdler handler.EventHandler, prc
 	}
 
 	c.LogConstructor(nil).Info("Starting EventSource", "source", src)
-	return src.Start(c.ctx, evthdler, c.Queue, prct...)
+	err := src.Start(c.ctx, evthdler, c.Queue, prct...)
+	if err == nil {
+		c.startedSources = append(c.startedSources, src)
+	}
+	return err
 }
 
 // NeedLeaderElection implements the manager.LeaderElectionRunnable interface.
@@ -148,23 +164,21 @@ func (c *Controller) Start(ctx context.Context) error {
 	// use an IIFE to get proper lock handling
 	// but lock outside to get proper handling of the queue shutdown
 	c.mu.Lock()
+	if c.Stopped {
+		return fmt.Errorf("can not restart a stopped controller, you should create a new one")
+	}
 	if c.Started {
-		return errors.New("controller was started more than once. This is likely to be caused by being added to a manager multiple times")
+		return fmt.Errorf("controller was started more than once. This is likely to be caused by being added to a manager multiple times")
 	}
 
 	c.initMetrics()
 
 	// Set the internal context.
-	c.ctx = ctx
+	c.ctx, c.cancel = context.WithCancel(ctx)
 
 	c.Queue = c.MakeQueue()
-	go func() {
-		<-ctx.Done()
-		c.Queue.ShutDown()
-	}()
-
 	wg := &sync.WaitGroup{}
-	err := func() error {
+	startErr := func() error {
 		defer c.mu.Unlock()
 
 		// TODO(pwittrock): Reconsider HandleCrash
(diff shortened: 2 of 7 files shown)


FillZpp:
K8s 1.26 is released and it adds removal of event handler https://github.com/kubernetes/kubernetes/pull/111122 , so that c-r had better support this in the next v0.14 release.

I'm still adding tests. But one thing I'm not sure is that whether we should add a `ControllerCtx` into `controller.Options` or add a `Stop` method for the Controller? Which one is better to let developers stop a running controller?

Any suggestions? @alvaroaleman @vincepri @joelanford

pkg/internal/controller/controller.go
```diff
@@ -240,8 +264,25 @@ func (c *Controller) Start(ctx context.Context) error {
...
-		return err
+
+	defer func() {
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		c.Stopped = true
+
+		c.Queue.ShutDown()
+		for _, src := range c.startedSources {
+			if stoppableSrc, ok := src.(source.StoppableSource); ok {
```
alvaroaleman, line 275: Maybe log a warning here if that is not the case?

pkg/source/source.go
```diff
@@ -359,7 +415,14 @@ func (is *Informer) Start(ctx context.Context, handler handler.EventHandler, que
 		return fmt.Errorf("must specify Informer.Informer")
 	}
 
-	_, err := is.Informer.AddEventHandler(internal.EventHandler{Queue: queue, EventHandler: handler, Predicates: prct})
+	is.mu.Lock()
+	defer is.mu.Unlock()
+	if is.canceled {
+		return nil
```
alvaroaleman, line 421: Shouldn't we error rather than silently doing nothing?

pkg/controller/controller.go
```diff
@@ -58,6 +58,10 @@ type Options struct {
 	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
 	// Defaults to the Controller.RecoverPanic setting from the Manager if unset.
 	RecoverPanic *bool
+
+	// ControllerCtx is the optional context for only this Controller. If it is set and been canceled,
+	// this controller and its watches will be stopped dynamically.
+	ControllerCtx context.Context
```
alvaroaleman, line 64: Why not add a `Stop` method instead?

pkg/source/source.go
```diff
@@ -370,6 +433,18 @@ func (is *Informer) String() string {
...
 }
 
+// Stop implements StoppableSource to stop it dynamically.
+func (is *Informer) Stop() error {
+	is.mu.Lock()
+	defer is.mu.Unlock()
+	is.canceled = true
+
+	if is.eventHandlerRegistration != nil {
+		return is.Informer.RemoveEventHandler(is.eventHandlerRegistration)
```
alvaroaleman, line 443: None of this actually stops the informer which I'd argue is the more important part - Is that a folllow-up?

Admittedly not easy, we will likely need to do some kind of refcounting

pkg/internal/controller/controller.go
FillZpp: added

pkg/source/source.go
FillZpp: I'm not sure, maybe we have to prevent the source been stopped multiple times by controller or user?
For example, someone may create a controller with three watches, then stop one watch manually, and finally stop the whole controller. At the time, controller will stop the stopped watch source and it will get the error.

pkg/controller/controller.go
FillZpp: Yeah, as I asked in https://github.com/kubernetes-sigs/controller-runtime/pull/2099#issuecomment-1350370602 .
Now I have changed the `ControllerCtx` to a `Stop` method.

pkg/source/source.go
FillZpp: Good question. It's hard to determine when should we stop the informer. Not only the informers in cache are not only created by source watches, but also triggered by user's `Get`, `List` calls to DelegatingClient. So we can't manage the lifecycle of informer according to the reference count of how many active watches in it.

joelanford: Thinking out loud here a few options:
1. Ref count for watches + some sort of LRU or timed cache for gets/lists that don't already have informers?
2. Ref count for watches + if a get/list requires a new informer (i.e. a watch didn't already start one), that informer is never removed?

maxsmythe: Hey! Just found this PR.

Gatekeeper is doing something similar, forking controller-runtime's informer cache to add a function that can remove informers. Would it be possible to have this available before solving the ref count issue? That would at least give users who are comfortable handling that complexity the ability to do so, but probably wouldn't catch unwary users.

Here is the key function we rely on:

https://github.com/open-policy-agent/gatekeeper/blob/8b426fb55da22abc0fe9bc925a3ca1ed08df50fe/third_party/sigs.k8s.io/controller-runtime/pkg/dynamiccache/informer_cache.go#L242-L251

We currently use it by maintaining a separate cache for dynamic watches:

https://github.com/open-policy-agent/gatekeeper/blob/8b426fb55da22abc0fe9bc925a3ca1ed08df50fe/pkg/watch/manager.go#L64-L91

That exports "registrars" to managing controllers to add or remove watches.

https://github.com/open-policy-agent/gatekeeper/blob/8b426fb55da22abc0fe9bc925a3ca1ed08df50fe/pkg/watch/registrar.go#L213-L276

https://github.com/open-policy-agent/gatekeeper/blob/8b426fb55da22abc0fe9bc925a3ca1ed08df50fe/pkg/controller/constrainttemplate/constrainttemplate_controller.go#L570-L582

Static watches (i.e. old-school controller.Watch() and watches initiated by client.Get()) use a separate cache.

Happy to talk more about this model, if interested, or help implementing parts if it means we can stop maintaining a fork!

vincepri: This is great background! I might rebase and continue on the work on this PR (unless @FillZpp has time) to get it to the finish line before 0.15 is released. Feel free to reach out on slack if we want to chat more about it and brainstorm.

pkg/controller/controller.go
```diff
@@ -86,6 +86,10 @@ type Controller interface {
 
 	// GetLogger returns this controller logger prefilled with basic information.
 	GetLogger() logr.Logger
+
+	// Stop stops the controller and all its watches dynamically.
+	// Note that it will only trigger the stop but will not wait for them all stopped.
+	Stop() error
```
inteon, line 92: Why do we add a `Stop` function here instead of canceling the context that was passed to `Start`?

pkg/source/source.go
maxsmythe: Happy to brainstorm! (sorry, it's been a busy week, so haven't reached out yet)

FillZpp: Sorry for this late reply (I'm on vacation last two weeks). I'll continue on this and get it into 0.15.

About the removal of informer, it seems we could have several ways:

1. Keep ref counts to remove informers automatically, like @joelanford suggested.

> Thinking out loud here a few options:
> 
> 1. Ref count for watches + some sort of LRU or timed cache for gets/lists that don't already have informers?
> 2. Ref count for watches + if a get/list requires a new informer (i.e. a watch didn't already start one), that informer is never removed?

I prefer this way, but I'm not sure will this make users confused? Most of them don't know when and why an informer been removed or not, and the sequence of doing watch and get/list also affect whether the informer will be removed or not...

2. Expose a method to let users manually remove a informer, as @maxsmythe suggested, IIUC.

3. Maybe 1+2 both provided, if they are all necessary and needed.

WDYT @vincepri @alvaroaleman @joelanford @sbueringer

pkg/controller/controller.go
FillZpp: 1. Not all of the implementations of `Source` can stop, e.g., `Func` or some custom types.
2. Users have no way to cancel the context passed from internal controller to source, to only stop the single watch.

pkg/source/source.go
maxsmythe: In addition, because Gatekeeper uses generic controllers (we have one controller that listens for all constraint kinds), (2) would be a better fit for our model than governing watch livelihood at the controller granularity.

Because dynamic watches are probably best for generic-type controllers (if you have the type hard-coded, why the need for dynamic watches?), managing the watches directly may be a better fit than managing controllers. I think both are workable, but (2) would definitely be less of a reach for us to implement.

FillZpp:
How about get this PR merged first, and I will post new PRs to support removal of informers (automatically & manually)? Otherwise it will become a huge PR that have too many API changes to review.
/cc @vincepri @alvaroaleman

inteon:
@FillZpp I created another PR that tries to solve the same problem as your PR.
PTAL, feel free to accept the approach in that PR or to copy (some of) the code from that PR to this PR instead.

FillZpp:
Thanks @inteon , I close this PR and help out on the new one.

/close

PR #2102, opened 2022-12-16
WIP: DNR: Manager multi-cluster support

vincepri:


pkg/cluster/context.go
```diff
@@ -0,0 +1,20 @@
...
+	if v, ok := ctx.Value(contextKey{}).(string); ok {
+		return v
+	}
+	return ""
+}
+
+// IntoContext returns a new Context, derived from ctx, which carries the
+// provided cluster identifier.
+func IntoContext(ctx context.Context, cluster Cluster) context.Context {
+	return context.WithValue(ctx, contextKey{}, cluster.ID())
```
alvaroaleman, line 19: This implicitly assumes that a controller will always interact with exactly one cluster, which often isn't true

pkg/cluster/cluster.go
```diff
@@ -37,6 +37,9 @@ import (
 
 // Cluster provides various methods to interact with a cluster.
 type Cluster interface {
+	// ID returns a unique identifier for the current cluster.
+	ID() string
```
alvaroaleman, line 41: It should be possible to also set this, what we do currently seems ok as default but whenever there is a controller that interacts with more than one cluster, something more human-readable seems useful

pkg/cache/cache.go
```diff
@@ -156,10 +156,33 @@ type Options struct {
 
 var defaultResyncTime = 10 * time.Hour
 
+// Default defaults the options internals.
+func (options *Options) Default(config *rest.Config) error {
```
alvaroaleman, line 160: Why export this?

pkg/cluster/context.go
vincepri: The idea behind the context key is to have the request cluster information always available, Clients would know which client to use based on this information.

We can, however, decide to go to a complete different path and rework our clients to be cluster-scoped, and in turn have the request coming into a Reconcile loop to be enhanced with other means to get clients/mappers/rest/etc

pkg/cache/cache.go
vincepri: I think this was a leftover, probably can remove

pkg/cluster/cluster.go
vincepri: Agreed! Could be an option in the builder?

PR #2103, opened 2022-12-16
🌱 Refactor internal cache/informers map impl

vincepri:
This is a first of a series of cleanups of our internal libraries, which has been long overdue. There are two commits in here, which we can also separate in different PRs if needed. The first pass simplifies w/ adding options to the implementation, the second refactors the map by removing the delegating map logic, which had lots of repetition and created confusion in the past.


pkg/cache/internal/informers_map.go
```diff
@@ -175,27 +189,43 @@ func (ip *specificInformersMap) waitForStarted(ctx context.Context) bool {
...
 	return syncedFuncs
 }
 
+// WaitForCacheSync waits until all the caches have been started and synced.
+func (ip *InformersMap) WaitForCacheSync(ctx context.Context) bool {
+	syncedFuncs := append([]cache.InformerSynced(nil), ip.HasSyncedFuncs()...)
+	if !ip.waitForStarted(ctx) {
+		return false
+	}
+	return cache.WaitForCacheSync(ctx.Done(), syncedFuncs...)
```
alvaroaleman, line 214:
```suggestion
	if !ip.waitForStarted(ctx) {
		return false
	}
	return cache.WaitForCacheSync(ctx.Done(), ip.HasSyncedFuncs()...)
```

pkg/cache/internal/informers_map.go
```diff
@@ -264,161 +314,141 @@ func (ip *specificInformersMap) addInformerToMap(gvk schema.GroupVersionKind, ob
...
-	//  pass in their own contexts instead of relying on this fixed one here.
-	ctx := context.TODO()
-
-	// create the relevant listwatch
-	return &cache.ListWatch{
-		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
-			ip.selectors(gvk).ApplyToList(&opts)
+	switch obj.(type) {
+	//
+	// Unstructured
```
alvaroaleman, line 337: These comments seem somewhat redundant

vincepri: Yeah they are, it's more for visibility than anything else

pkg/cache/internal/informers_map.go
```diff
@@ -34,45 +35,50 @@ import (
...
+	ResyncPeriod time.Duration
+	Namespace    string
+	ByGVK        InformersMapOptionsByGVK
+}
+
+// InformersMapOptionsByGVK configured additional by group version kind (or object)
+// in an InformerMap.
+type InformersMapOptionsByGVK struct {
+	Selectors       SelectorsByGVK
+	Transformers    TransformFuncByObject
```
sbueringer, line 58: No strong objection but it seems a bit odd to have "ByGVK" and then have `TransformFuncByObject`.

Looking at `TransformFuncByObject` it seems to essentially also use GVK
```go
	Set(runtime.Object, *runtime.Scheme, cache.TransformFunc) error
	Get(schema.GroupVersionKind) cache.TransformFunc
```

Get directly works with GVK and only Set takes an Object but only uses it to infer the GVK.

Wondering if we could/should rename `TransformFuncByObject` to `TransformFuncByGVK` especially as I would say the main functionality is mapping from "something" to `TransformFunc` in `Get` and that is done by GVK

pkg/cache/internal/informers_map.go
```diff
@@ -145,7 +152,7 @@ type specificInformersMap struct {
 
 // Start calls Run on each of the informers and sets started to true.  Blocks on the context.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
-func (ip *specificInformersMap) Start(ctx context.Context) {
+func (ip *InformersMap) Start(ctx context.Context) error {
```
sbueringer, line 155: Q: Was the error return parameter just added in case we want to return errors in the future? (as of now we only return nil)

pkg/cache/internal/informers_map.go
```diff
@@ -175,27 +189,42 @@ func (ip *specificInformersMap) waitForStarted(ctx context.Context) bool {
...
 
 // HasSyncedFuncs returns all the HasSynced functions for the informers in this map.
-func (ip *specificInformersMap) HasSyncedFuncs() []cache.InformerSynced {
+func (ip *InformersMap) HasSyncedFuncs() []cache.InformerSynced {
 	ip.mu.RLock()
 	defer ip.mu.RUnlock()
-	syncedFuncs := make([]cache.InformerSynced, 0, len(ip.informersByGVK))
-	for _, informer := range ip.informersByGVK {
-		syncedFuncs = append(syncedFuncs, informer.Informer.HasSynced)
+	syncedFuncs := make([]cache.InformerSynced, 0, len(ip.informers.Structured))
```
sbueringer, line 195: nit: should we use len(ip.informers.Structured) + len(ip.informers.Unstructured) + len (ip.informers.Metadata) instead?

pkg/cache/internal/informers_map.go
```diff
@@ -213,24 +242,44 @@ func (ip *specificInformersMap) Get(ctx context.Context, gvk schema.GroupVersion
...
+	ni := cache.NewSharedIndexInformer(&cache.ListWatch{
+		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
+			ip.selectors(gvk).ApplyToList(&opts)
+			return lw.ListFunc(opts)
+		},
+		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
+			ip.selectors(gvk).ApplyToList(&opts)
+			opts.Watch = true // Watch needs to be set to true separately
+			return lw.WatchFunc(opts)
+		},
```
sbueringer, line 281: Q: would it make sense to do this in `makeListWatcher` so that `makeListWatcher` alredy creates the final `ListWatch` vs having "layered" ListWatches?

pkg/cache/internal/informers_map.go
```diff
@@ -264,161 +313,141 @@ func (ip *specificInformersMap) addInformerToMap(gvk schema.GroupVersionKind, ob
...
+				}
+				return list, err
+			},
+			// Setup the watch function
+			WatchFunc: func(opts metav1.ListOptions) (watcher watch.Interface, err error) {
+				defer func() {
+					if watcher != nil {
+						watcher = newGVKFixupWatcher(gvk, watcher)
+					}
+				}()
```
sbueringer, line 403: nit: If not necessary, I would prefer the previous variant without defer as it's just slightly easier to read, but up to you

sbueringer (commented):
Just some nits, overall lgtm

pkg/cache/internal/informers_map.go
vincepri: Originally it was in each list watcher creation function, although the code was duplicate between each watch/listfunc, which was a bit repetitive. Do you see another way to do this that doesn't end up in repetition?

pkg/cache/internal/informers_map.go
vincepri: It was added to satisfy the `Runnable` interface, the same (nil error returned in any case) was before in the deleg_map file

pkg/cache/internal/informers_map.go
vincepri: done ptal

pkg/cache/internal/informers_map.go
vincepri: Renamed, it was all internal functions fortunately

pkg/cache/internal/informers_map.go
sbueringer: Makes sense, thx for the info!

pkg/cache/internal/informers_map.go
sbueringer: I see the following options:
1. write the listFunc and watchFunc in makeListWatcher inside the switch in variables and then build the ListWatch like it is done here at the end of the makeListWatcher func
1. writing the ListWatch'es in makeListWatcher in a var and then add this code at the end of the makeListWatcher func
1. defining inline modifyListOptions modifyWatchOptions func in makeListWatcher and calling them in each case

I would probably go with option 1, but also fine to just keep it as is if you want to

PR #2114, opened 2022-12-27
🐛 initial version of nondestructive hasLabels

### LINKED ISSUE #2098: ListOptions step on each other, HasLabels and MatchingLabels compete
When trying to get a List using the variadic ListOptions argument, I ran to unexpected behavior if applying both HasLabels and MatchingLabels.

```
List(ctx, podList, client.InNamespace(namespace)
    client.MatchingLabels{
        utils.ClusterLabelName: clusterName,
    },
    client.HasLabels{"role"}, // this ensures we are getting instance pods only
)
```

When watching the code in action, more pods than expected were being returned.
It seemed as if the ListOptions were being OR'ed rather than AND'ed.

Looking into the implementation of `MatchingLabels` and `HasLabels`, in their `ApplyToList(opts *ListOptions)`,
both will overwrite whatever values the `LabelSelector` had before.

```
opts.LabelSelector = sel
```

This means, likely, that if we apply both HasLabels and MatchingLabels, the last one applied will "win".
This is rather un-intuitive behavior.
If it is so by design, I think it should be documented.
If possible, I would think a logical AND would be the intuitive expectation.

jsilvela:
WIP - I still have stuff I want to do on this, but this initial version is non-destructive.
Todo:
- make the unit tests clearly order-consistent or order-independent
- get `HasLabels` to not silently drop invalid requirements
- follow guidance by Alvaro Aleman on making flow clearer

Closes #2098 

Signed-off-by: Jaime Silvela <jaime.silvela@enterprisedb.com>

Fixes the problem that `HasLabels` and `MatchingLabels`, applied as ListOptions, are destructive.
Each will overwrite the LabelSelector with its own arguments, and the last applied wins.
This code attempts to make `HasLabels` and `MatchingLabels` add as an AND, matching intuitive
expectations when applying filters

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -44,6 +44,7 @@ import (
 var _ = Describe("Fake client", func() {
 	var dep *appsv1.Deployment
 	var dep2 *appsv1.Deployment
+	var dep3 *appsv1.Deployment
 	var cm *corev1.ConfigMap
 	var cl client.WithWatch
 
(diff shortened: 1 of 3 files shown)


jsilvela:
Closing, as issue was fixed by https://github.com/kubernetes-sigs/controller-runtime/pull/2363

PR #2117, opened 2023-01-03
:sparkles: Add controller option for setting leader election

jonstacks:
## Background

Creating a controller that is not leader elected is seems to require a bit of indirection. You currently have to create a new type that embeds the `Controller` interface and adds a `NeedLeaderElection` method as shown below so that it implements the `manager.LeaderElectionRunnable` interface.

```go
type MyReconciler struct {
	client.Client
	Log      logr.Logger
	Scheme   *runtime.Scheme
	Recorder record.EventRecorder
}

func (rec *MyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	return ctrl.Result{}, nil
}

func (rec *MyReconciler) SetupWithManager(mgr ctrl.Manager) error {
	cont, err := controller.NewUnmanaged("my-controller", mgr, controller.Options{
		Reconciler: rec,
	})
	if err != nil {
		return err
	}
	cont = MyController{cont}
	cont.Watch(&source.Kind{Type: &netv1.Ingress{}}, &handler.EnqueueRequestForObject{}, commonPredicateFilters)
	return mgr.Add(cont)
}

type MyController struct {
	controller.Controller
}

func (c *MyController) NeedLeaderElection() bool {
	return false
}
```

## What

This PR adds a new controller option, `LeaderElected`, that can be used to disable leader election for the controller, without having to create a new type.

## How

It does this by making `controller.Controller` implement the `manager.LeaderElectionRunnable` interface.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -59,9 +59,9 @@ type Options struct {
 	// Defaults to the Controller.RecoverPanic setting from the Manager if unset.
 	RecoverPanic *bool
 
-	// LeaderElected indicates whether the controller needs to use leader election.
+	// NeedLeaderElection indicates whether the controller needs to use leader election.
 	// Defaults to true, which means the controller will use leader election.
-	LeaderElected *bool
+	NeedLeaderElection *bool
 }
 
 // Controller implements a Kubernetes API.  A Controller manages a work queue fed reconcile.Requests
@@ -160,7 +160,7 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		Name:                    name,
 		LogConstructor:          options.LogConstructor,
 		RecoverPanic:            options.RecoverPanic,
-		LeaderElected:           options.LeaderElected,
+		LeaderElected:           options.NeedLeaderElection,
 	}, nil
 }
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -58,6 +58,10 @@ type Options struct {
 	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
 	// Defaults to the Controller.RecoverPanic setting from the Manager if unset.
 	RecoverPanic *bool
+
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -58,6 +58,10 @@ type Options struct {
 	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
 	// Defaults to the Controller.RecoverPanic setting from the Manager if unset.
 	RecoverPanic *bool
+
+	// LeaderElected indicates whether the controller needs to use leader election.
+	// Defaults to true, which means the controller will use leader election.
+	LeaderElected *bool
```
alvaroaleman, line 64: IMHO the method name is clearer in terms of clarifying what this is for
```suggestion
	NeedLeaderElection *bool
```

jonstacks: @alvaroaleman, I agree that the method name is clearer.

I can't use `NeedLeaderElection` for the field though since the method needs that name to implement `manager.LeaderElectionRunnable` and making the field match leads to a [DuplicateFieldAndMethod](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#DuplicateFieldAndMethod) error. Do you think any of these field names would be clearer?

* `NeedsLeaderElection` <- Just adds a `s` at the end of `Need` to prevent duplicate field
* `RequiresLeaderElection`
* `WantsLeaderElection`
* `UsesLeaderElection`

I'm open to any suggestions to make it clearer

alvaroaleman: The field name here can be anything, this is controller.Options not controller.Controller

jonstacks: I'm sorry I missed that :man_facepalming: . Thank you for pointing that out! Should be fixed in https://github.com/kubernetes-sigs/controller-runtime/pull/2117/commits/8988b10b2131766f2ea401a8687c77d7d4298453

alvaroaleman (approved):
thanks!

PR #2119, opened 2023-01-05
⚠ Refactor controllers to require a cluster on watches, remove dep inject

vincepri:

Marking this as ⚠ given that a public interface is now changed and it removes a public field in the Controller struct.

Previously, a Controller exposed a SetFields field, which was used to perform dependency injection on sources. A controller is always associated with a Manager when created, which then calls `SetFields()` on it. This PR removes a layer of indirection which allows using SetFields directly, and instead it gets the required dependencies from the Cluster object.

Further cleanup is needed although that's going to be slated future PRs.

By passing in a Cluster as the Watch parameter, in the future we should be able to add watches for different clusters based on the given source and handlers.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -74,7 +75,7 @@ type Controller interface {
 	// Watch may be provided one or more Predicates to filter events before
 	// they are given to the EventHandler.  Events will be passed to the
 	// EventHandler if all provided Predicates evaluate to true.
-	Watch(src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
+	Watch(cluster cluster.Cluster, src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
 
 	// Start starts the controller.  Start blocks until the context is closed or a
 	// controller has an error starting.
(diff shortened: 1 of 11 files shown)


pkg/controller/controller.go
```diff
@@ -74,7 +75,7 @@ type Controller interface {
 	// Watch may be provided one or more Predicates to filter events before
 	// they are given to the EventHandler.  Events will be passed to the
 	// EventHandler if all provided Predicates evaluate to true.
-	Watch(src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
+	Watch(cluster cluster.Cluster, src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
```
alvaroaleman, line 78: This seems like a stopgap. The controller doesn't need this at all, nor do all source.Source and handler.EventHandler implementations. I'd rather unexport the ones that need it and export a new constructor that takes a cluster.Cluster to end up with one breaking change only

vincepri: That's true, the controller doesn't, but the Source and EventHandler need the fields on the Cluster/Manager to set their caches, and so on. Do you have a better alternative in mind?

Removing the dependency injection from the Source is going to probably take a major effort, which not sure if we should tackle just yet, but open to better alternatives. The tricky part is going to be later on to make sure Sources, specifically the source.Kind one, and other most used ones, can be used w/ specific clusters.

alvaroaleman: Why do you think it will be a major effort? There is exactly one source that uses this, `source.Kind`. We can unexport the type and make `Kind` a func that matches what we have in `NewKindWithCache`. 

Same story for the handler, the only one that actually needs this is `EnqueueRequestForOwner`. Lets unexport the type and instead make `*EnqueueRequestForOwner` a constructor func. Other than that there is only `*enqueueRequestsFromMapFunc` which just passes on `Inject.Funcs` but doens't use that itself.


> The tricky part is going to be later on to make sure Sources, specifically the source.Kind one, and other most used ones, can be used w/ specific clusters.

This is already possible today, using `source.NewKindWithCache`.

PR #2120, opened 2023-01-06
⚠️ Refactor source/handler/predicate packages to remove dep injection

vincepri:


hack/test-all.sh
```diff
@@ -25,7 +25,7 @@ if [[ -n ${ARTIFACTS:-} ]]; then
 fi
 
 result=0
-go test -race ${P_FLAG} ${MOD_OPT} ./... ${GINKGO_ARGS} || result=$?
+go test -v -race ${P_FLAG} ${MOD_OPT} ./... --ginkgo.fail-fast ${GINKGO_ARGS} || result=$?
```
vincepri, line 28: These were actually useful to catch errors right away, and have a bit more verbosity when tests are running, we should probably keep them

pkg/builder/controller_test.go
```diff
@@ -157,7 +157,7 @@ var _ = Describe("application", func() {
 
 			instance, err := ControllerManagedBy(m).
 				Named("my_controller").
-				Watches(&source.Kind{Type: &appsv1.ReplicaSet{}}, &handler.EnqueueRequestForObject{}).
+				Watches(source.Kind(m.GetCache(), &appsv1.ReplicaSet{}), &handler.EnqueueRequestForObject{}).
```
vincepri, line 160: This is now the new experience on how to create a controller in a manager. Over time, the builder has increased the options we're passing in, which is ok, but we've also had to deal with all of the dependency injection stuff which isn't pretty at all.

This UX might be a stopgap, but it's actually nicer to be explicit where a source is getting its cache or any other function. In a follow-up PR we might want to explore easier ways to plow through this information when we're clearly building a controller in a Manager, and make the source aware that it can access whatever it needs through the manager.

For now, we can keep this as a first step breaking change and iterate on it as we see fit. The clarity might come handy in the future when we reason through the codebase.

pkg/builder/controller.go
```diff
@@ -259,12 +255,12 @@ func (blder *Builder) doWatch() error {
 		allPredicates = append(allPredicates, w.predicates...)
 
 		// If the source of this watch is of type *source.Kind, project it.
-		if srckind, ok := w.src.(*source.Kind); ok {
-			typeForSrc, err := blder.project(srckind.Type, w.objectProjection)
-			if err != nil {
-				return err
+		if srckind, ok := w.src.(source.KindSource); ok {
```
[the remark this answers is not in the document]
vincepri: I'm with you that this isn't ideal, although the UX for building sources with PartialObjectMetadata is pretty poor if we don't do this for our users. That would basically mean creating an PartialObjectMetadata, figure out the GVK, potentially using the scheme directly, which can cause errors. We should probably keep this as-is for now, and figure out a different way later

pkg/source/source.go
```diff
@@ -67,34 +67,21 @@ type SyncingSource interface {
...
 
-// NewKindWithCache creates a Source without InjectCache, so that it is assured that the given cache is used
-// and not overwritten. It can be used to watch objects in a different cluster by passing the cache
-// from that other cluster.
-func NewKindWithCache(object client.Object, cache cache.Cache) SyncingSource {
-	return &kindWithCache{kind: Kind{Type: object, cache: cache}}
+// KindSource is a syncing source that targets a specific object type.
+type KindSource interface {
+	SyncingSource
+	AsPartialMetadata(*runtime.Scheme) error
```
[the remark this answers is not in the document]
vincepri: done, ptal

pkg/source/source.go
```diff
@@ -67,34 +67,21 @@ type SyncingSource interface {
...
 }
 
-type kindWithCache struct {
-	kind Kind
+// Kind creates a KindSource with the given cache provider.
+func Kind(cache cache.Cache, object client.Object) KindSource {
```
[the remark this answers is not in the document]
vincepri: We can but the EnqueueRequestForOwner counterpart for example takes any required parameters before taking the object, wdyt?

pkg/predicate/predicate.go
```diff
@@ -242,15 +241,6 @@ type and struct {
 	predicates []Predicate
 }
 
-func (a and) InjectFunc(f inject.Func) error {
```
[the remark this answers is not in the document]
vincepri: I went ahead and removed most of the injection code except for the ones that touches webhooks, which we should handle in a separate PR

pkg/builder/controller.go
```diff
@@ -259,12 +255,12 @@ func (blder *Builder) doWatch() error {
 		allPredicates = append(allPredicates, w.predicates...)
 
 		// If the source of this watch is of type *source.Kind, project it.
```
sbueringer, line 257: nit: Comment seems to be outdated

sbueringer: Q: is it correct to check for SyncingSource and then pass the src into KindAsPartialMetadata where we check for kind?

Could it be a SyncingSource which is not kind? ("Users may build their own Source implementations.")

pkg/cluster/cluster_test.go
```diff
@@ -243,34 +169,15 @@ var _ = Describe("cluster.Cluster", func() {
 	})
 })
 
-var _ inject.Cache = &injectable{}
-var _ inject.Client = &injectable{}
 var _ inject.Scheme = &injectable{}
-var _ inject.Config = &injectable{}
 var _ inject.Logger = &injectable{}
 
 type injectable struct {
```
sbueringer, line 175: Do we still need the injectable? Seems unused (except l.172+173)

pkg/handler/enqueue_owner.go
```diff
@@ -42,13 +45,34 @@ var log = logf.RuntimeLog.WithName("eventhandler").WithName("EnqueueRequestForOw
 //
 // - a source.Kind Source with Type of Pod.
 //
-// - a handler.EnqueueRequestForOwner EventHandler with an OwnerType of ReplicaSet and IsController set to true.
-type EnqueueRequestForOwner struct {
-	// OwnerType is the type of the Owner object to look for in OwnerReferences.  Only Group and Kind are compared.
-	OwnerType runtime.Object
+// - a handler.enqueueRequestForOwner EventHandler with an OwnerType of ReplicaSet and OnlyControllerOwner set to true.
+func EnqueueRequestForOwner(scheme *runtime.Scheme, mapper meta.RESTMapper, owner client.Object, opts ...OwnerOption) EventHandler {
```
sbueringer, line 49:
```suggestion
func EnqueueRequestForOwner(scheme *runtime.Scheme, mapper meta.RESTMapper, ownerType client.Object, opts ...OwnerOption) EventHandler {
```

Just wondering if the parameter should be called ownerType as well

pkg/source/source.go
```diff
@@ -67,34 +67,33 @@ type SyncingSource interface {
...
 }
 
-// Kind is used to provide a source of events originating inside the cluster from Watches (e.g. Pod Create).
-type Kind struct {
-	// Type is the type of object to watch.  e.g. &v1.Pod{}
-	Type client.Object
+// kind is used to provide a source of events originating inside the cluster from Watches (e.g. Pod Create).
+type kind struct {
+	// obj is the type of object to watch.  e.g. &v1.Pod{}
+	obj client.Object
```
sbueringer, line 96: Should this be objType? (just wondering why obj here vs ownerType elsewhere, fells like we could keep it consistent)

pkg/source/source.go
```diff
@@ -219,7 +204,7 @@ type Channel struct {
 	// Source is the source channel to fetch GenericEvents
 	Source <-chan event.GenericEvent
 
-	// stop is to end ongoing goroutine, and close the channels
+	// Stop is to end ongoing goroutine, and close the channels
```
sbueringer, line 207:
```suggestion
	// stop is to end ongoing goroutine, and close the channels
```
nit: given the field name is lower case?

pkg/source/source_test.go
```diff
@@ -188,25 +178,23 @@ var _ = Describe("Source", func() {
 			})
 		})
 
-		It("should return an error from Start if informers were not injected", func() {
-			instance := source.Kind{Type: &corev1.Pod{}}
+		It("should return an error from Start if informers were provided", func() {
```
sbueringer, line 181:
```suggestion
		It("should return an error from Start if cache was not provided", func() {
```
?

pkg/builder/controller.go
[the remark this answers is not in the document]
sbueringer: I have a related comment below. not sure if we can actually do this as is https://github.com/kubernetes-sigs/controller-runtime/pull/2120#discussion_r1068235965

pkg/builder/controller.go
vincepri: I agree it's not ideal. Originally I had a `KindSource` instead with `AsPartialMetadata()` method required. These builders need a lot of rework honestly, the UX also isn't ideal with modifiers in place. I'm happy to go back to the interface if we prefer that.

pkg/cluster/cluster_test.go
vincepri: Thanks! Leftovers :)

pkg/builder/controller.go
vincepri: PTAL, added a new interface called `TypedSource` which lets users provide their own if needed

sbueringer: lgtm

sbueringer:
lgtm pending squash from my side

vincepri:
@sbueringer Let's keep the commits separate for ease of history later, wdyt?

sbueringer:
> @sbueringer Let's keep the commits separate for ease of history later, wdyt?

Oh sure. Didn't look if they were intentionally separated to be honest 😀

/lgtm

/hold
So Alvaro can take another look as well

pkg/builder/controller.go
```diff
@@ -197,18 +196,16 @@ func (blder *Builder) Build(r reconcile.Reconciler) (controller.Controller, erro
...
 	switch proj {
 	case projectAsNormal:
-		return obj, nil
+		return src, nil
 	case projectAsMetadata:
-		metaObj := &metav1.PartialObjectMetadata{}
-		gvk, err := getGvk(obj, blder.mgr.GetScheme())
-		if err != nil {
-			return nil, fmt.Errorf("unable to determine GVK of %T for a metadata-only watch: %w", obj, err)
+		if err := source.KindAsPartialMetadata(src, blder.mgr.GetScheme()); err != nil {
```
alvaroaleman, line 205: What I meant with the other comment is that it would be nicer to have a `source.NewMetadataKind(obj client.Object, scheme *runtime.scheme)`, right now it requires to first build a `kind` source and then convert it, which is a bit awkward IMHO

pkg/controller/controller.go
```diff
@@ -139,11 +139,6 @@ func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller
 		options.RateLimiter = workqueue.DefaultControllerRateLimiter()
 	}
 
-	// Inject dependencies into Reconciler
-	if err := mgr.SetFields(options.Reconciler); err != nil {
```
alvaroaleman, line 143: Pretty sure this will break people, as Kubebuilder at least used to scaffold in a way that this ended up being done, not sure if that is still the case today. You sure you want to just remove this instead of warn for one release if any of the things end up getting injected?

pkg/source/source.go
```diff
@@ -67,34 +67,35 @@ type SyncingSource interface {
 	WaitForSync(ctx context.Context) error
 }
 
-// NewKindWithCache creates a Source without InjectCache, so that it is assured that the given cache is used
-// and not overwritten. It can be used to watch objects in a different cluster by passing the cache
-// from that other cluster.
-func NewKindWithCache(object client.Object, cache cache.Cache) SyncingSource {
-	return &kindWithCache{kind: Kind{Type: object, cache: cache}}
+// TypedSource contains methods to set the Source's watched object type.
+type TypedSource interface {
```
alvaroaleman, line 71: Maybe `KindSource` to make clear that the two are related?

pkg/source/source.go
```diff
@@ -174,16 +172,28 @@ func (ks *Kind) Start(ctx context.Context, handler handler.EventHandler, queue w
...
 
-func (ks *Kind) String() string {
-	if ks.Type != nil {
-		return fmt.Sprintf("kind source: %T", ks.Type)
+func (ks *kind) GetType() client.Object {
+	return ks.objType
+}
+
+func (ks *kind) SetType(objType client.Object) error {
+	if ks.started != nil {
```
alvaroaleman, line 180: Do you think `ks.started` is worth protecting with a mutex?

alvaroaleman (commented):
Sorry it took so long, just some nits.

pkg/controller/controller.go
vincepri: Yes, these fields have been deprecated since 0.10; it's better to not linger the support for dependency injection longer. We can do a couple of things to avoid breaking, including giving a heads up to folks in the mailing list.

pkg/builder/controller.go
vincepri: Agreed, we need more thinking on how we handle the Metadata projection stuff, reverted some changes and instead moved the Kind object in an internal package

alvaroaleman (approved):

alvaroaleman (approved):

PR #2121, opened 2023-01-08
✨Use correct context to cancel "list and watch" & wait for all informers to complete

inteon:
This PR fixes this TODO item:
https://github.com/kubernetes-sigs/controller-runtime/blob/b7561613a41a7759668636e9a69fa377d7989f84/pkg/cache/internal/informers_map.go#L320-L322

It also make sure that all informers finish (using a `sync.WaitGroup`) before returning from the `Start` function.

### DIFF
--- a/pkg/cache/internal/informers_map.go
+++ b/pkg/cache/internal/informers_map.go
@@ -117,9 +117,6 @@ type InformersMap struct {
 	// paramCodec is used by list and watch
 	paramCodec runtime.ParameterCodec
 
-	// stop is the stop channel to stop informers
-	stop <-chan struct{}
-
 	// resync is the base frequency the informers are resynced
 	// a 10 percent jitter will be added to the resync period between informers
 	// so that all informers will not send list requests simultaneously.
@@ -128,12 +125,21 @@ type InformersMap struct {
 	// mu guards access to the map
 	mu sync.RWMutex
 
-	// start is true if the informers have been started
+	// started is true if the informers have been started
 	started bool
 
 	// startWait is a channel that is closed after the
 	// informer has been started.
 	startWait chan struct{}
+
+	// waitGroup is the wait group that is used to wait for all informers to stop
+	waitGroup sync.WaitGroup
+
+	// stopped is true if the informers have been stopped
+	stopped bool
+
+	// ctx is the context to stop informers
+	ctx context.Context
 
 	// namespace is the namespace that all ListWatches are restricted to
 	// default or empty string means all namespaces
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/hold

Looks good to me, thanks!

/cc @vincepri for a second set of eyes

pkg/cache/internal/informers_map.go
```diff
@@ -158,24 +164,42 @@ func (ip *InformersMap) Start(ctx context.Context) error {
 		defer ip.mu.Unlock()
 
 		// Set the stop channel so it can be passed to informers that are added later
```
vincepri, line 166: Change this comment?

pkg/cache/internal/informers_map.go
```diff
@@ -310,17 +334,17 @@ func (ip *InformersMap) addInformerToMap(gvk schema.GroupVersionKind, obj runtim
 	// Start the Informer if need by
 	// TODO(seans): write thorough tests and document what happens here - can you add indexers?
 	// can you add eventhandlers?
-	if ip.started {
-		go i.Informer.Run(ip.stop)
+	if ip.started && !ip.stopped {
```
vincepri, line 337: Add a comment on why this is needed?

pkg/cache/internal/informers_map.go
```diff
@@ -158,24 +164,42 @@ func (ip *InformersMap) Start(ctx context.Context) error {
...
+
+		ip.waitGroup.Add(len(ip.informers.Structured) + len(ip.informers.Unstructured) + len(ip.informers.Metadata))
 
 		// Start each informer
 		for _, i := range ip.informers.Structured {
-			go i.Informer.Run(ctx.Done())
+			i := i
+			go func() {
+				defer ip.waitGroup.Done()
+				i.Informer.Run(ctx.Done())
```
vincepri, line 176: Can we make this a standalone method that takes an informer instead?

pkg/cache/internal/informers_map.go
```diff
@@ -158,24 +164,42 @@ func (ip *InformersMap) Start(ctx context.Context) error {
 		defer ip.mu.Unlock()
 
 		// Set the stop channel so it can be passed to informers that are added later
-		ip.stop = ctx.Done()
+		ip.ctx = ctx
```
vincepri, line 167: Should we create a sub-context here with a cancel function?

pkg/cache/internal/informers_map.go
```diff
@@ -310,17 +334,17 @@ func (ip *InformersMap) addInformerToMap(gvk schema.GroupVersionKind, obj runtim
 	// Start the Informer if need by
```
vincepri, line 334: At the top of this function, shouldn't we check first if we're stopped already before getting all the way down here and return an error?

If I got the flow right, this function would still return correctly even if the informermap has been stopped

pkg/cache/internal/informers_map.go
inteon: When would we cancel this sub-context?

pkg/cache/internal/informers_map.go
inteon: Indeed, I still return correctly. I did this to make sure the behavior remained identical to before the change.

pkg/cache/internal/informers_map.go
inteon: Done

pkg/cache/internal/informers_map.go
inteon: Done

pkg/cache/internal/informers_map.go
inteon: It won't fail silently, it will error here (identical to current behavior): https://github.com/kubernetes-sigs/controller-runtime/blob/45e7aa7ab1fa4e8675848e2bcd9058c14cb0131a/pkg/cache/internal/informers_map.go#L266

I think that in order to prevent making a breaking change, we can keep the code as-is or throw the same timeout error from the addInformerToMap function.

vincepri: That makes sense, although it sounds to me it might be a user error to add after Stop has been called. Could we mark this as a breaking change and return that error?

inteon: Ok, I'll need some more time to figure out how to reliably return an error. I agree that it would be useful to make a breaking change here.

vincepri (approved):
/approve
/lgtm
/hold cancel

Merging this now, the current set of changes look good. Let's follow-up for the next iteration to return an error

PR #2122, opened 2023-01-08
⚠️ make *http.Client configurable and use/ share the same http.Client in the default configuration

inteon:
client-go has all these fancy `NewForConfigAndClient` constructor functions.
Let's expose this functionality to the controller-runtime user too & reduce the number of different http.Clients that we use.

In https://github.com/kubernetes/kubernetes/pull/105490, the decision for making http.Client configurable is explained: "when using a custom transport, this has the consequence that each group/version had an independent TCP connection to the API server"
By re-using the same http client for all the clients, we can share customized transports, resulting in no additional TCP connections (see https://github.com/kubernetes/kubernetes/blob/5681b0da14f11165fc53ad82a3400fd027979fc4/test/integration/apiserver/apiserver_test.go#L2517-L2574 for an example)

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -19,6 +19,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
 	"time"
 
@@ -108,6 +109,9 @@ type SelectorsByObject map[client.Object]ObjectSelector
 
 // Options are the optional arguments for creating a new InformersMap object.
 type Options struct {
+	// HTTPClient is the http client to use for the REST client
+	HTTPClient *http.Client
+
 	// Scheme is the scheme to use for mapping objects to GroupVersionKinds
 	Scheme *runtime.Scheme
 
@@ -184,6 +188,7 @@ func New(config *rest.Config, opts Options) (Cache, error) {
 	return &informerCache{
 		scheme: opts.Scheme,
 		Informers: internal.NewInformers(config, &internal.InformersOpts{
+			HTTPClient:   opts.HTTPClient,
 			Scheme:       opts.Scheme,
 			Mapper:       opts.Mapper,
 			ResyncPeriod: *opts.Resync,
@@ -414,6 +419,18 @@ func combineTransform(inherited, current toolscache.TransformFunc) toolscache.Tr
 }
 
 func defaultOpts(config *rest.Config, opts Options) (Options, error) {
+	logger := log.WithName("setup")
+
+	// Use the rest HTTP client for the provided config if unset
+	if opts.HTTPClient == nil {
+		var err error
+		opts.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			logger.Error(err, "Failed to get HTTP client")
+			return opts, fmt.Errorf("could not create HTTP client from config")
+		}
+	}
+
 	// Use the default Kubernetes Scheme if unset
 	if opts.Scheme == nil {
 		opts.Scheme = scheme.Scheme
@@ -422,9 +439,9 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 	// Construct a new Mapper if unset
 	if opts.Mapper == nil {
 		var err error
-		opts.Mapper, err = apiutil.NewDiscoveryRESTMapper(config)
+		opts.Mapper, err = apiutil.NewDiscoveryRESTMapper(config, opts.HTTPClient)
 		if err != nil {
-			log.WithName("setup").Error(err, "Failed to get API Group-Resources")
+			logger.Error(err, "Failed to get API Group-Resources")
 			return opts, fmt.Errorf("could not create RESTMapper from config")
 		}
 	}
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -22,6 +22,7 @@ package apiutil
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"reflect"
 	"sync"
 
@@ -61,9 +62,13 @@ func AddToProtobufScheme(addToScheme func(*runtime.Scheme) error) error {
 
 // NewDiscoveryRESTMapper constructs a new RESTMapper based on discovery
 // information fetched by a new client with the given config.
-func NewDiscoveryRESTMapper(c *rest.Config) (meta.RESTMapper, error) {
+func NewDiscoveryRESTMapper(c *rest.Config, httpClient *http.Client) (meta.RESTMapper, error) {
+	if httpClient == nil {
+		panic("httpClient must not be nil")
+	}
+
 	// Get a mapper
-	dc, err := discovery.NewDiscoveryClientForConfig(c)
+	dc, err := discovery.NewDiscoveryClientForConfigAndClient(c, httpClient)
 	if err != nil {
 		return nil, err
 	}
@@ -150,8 +155,11 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 // RESTClientForGVK constructs a new rest.Interface capable of accessing the resource associated
 // with the given GroupVersionKind. The REST client will be configured to use the negotiated serializer from
 // baseConfig, if set, otherwise a default serializer will be set.
-func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory) (rest.Interface, error) {
-	return rest.RESTClientFor(createRestConfig(gvk, isUnstructured, baseConfig, codecs))
+func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory, httpClient *http.Client) (rest.Interface, error) {
+	if httpClient == nil {
+		panic("httpClient must not be nil")
+	}
+	return rest.RESTClientForConfigAndClient(createRestConfig(gvk, isUnstructured, baseConfig, codecs), httpClient)
 }
 
 // createRestConfig copies the base config and updates needed fields for a new rest config.
--- a/pkg/client/apiutil/dynamicrestmapper.go
+++ b/pkg/client/apiutil/dynamicrestmapper.go
@@ -17,6 +17,7 @@ limitations under the License.
 package apiutil
 
 import (
+	"net/http"
 	"sync"
 	"sync/atomic"
 
@@ -75,8 +76,12 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
 // NewDynamicRESTMapper returns a dynamic RESTMapper for cfg. The dynamic
 // RESTMapper dynamically discovers resource types at runtime. opts
 // configure the RESTMapper.
-func NewDynamicRESTMapper(cfg *rest.Config, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
-	client, err := discovery.NewDiscoveryClientForConfig(cfg)
+func NewDynamicRESTMapper(cfg *rest.Config, httpClient *http.Client, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
+	if httpClient == nil {
+		panic("httpClient must not be nil")
+	}
+
+	client, err := discovery.NewDiscoveryClientForConfigAndClient(cfg, httpClient)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -38,6 +39,9 @@ import (
 
 // Options are creation options for a Client.
 type Options struct {
+	// HTTPClient is the HTTP client to use for requests.
+	HTTPClient *http.Client
+
 	// Scheme, if provided, will be used to map go structs to GroupVersionKinds
 	Scheme *runtime.Scheme
 
@@ -116,6 +120,15 @@ func newClient(config *rest.Config, options Options) (*client, error) {
 		)
 	}
 
+	// Use the rest HTTP client for the provided config if unset
+	if options.HTTPClient == nil {
+		var err error
+		options.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// Init a scheme if none provided
 	if options.Scheme == nil {
 		options.Scheme = scheme.Scheme
@@ -124,23 +137,24 @@ func newClient(config *rest.Config, options Options) (*client, error) {
 	// Init a Mapper if none provided
 	if options.Mapper == nil {
 		var err error
-		options.Mapper, err = apiutil.NewDynamicRESTMapper(config)
+		options.Mapper, err = apiutil.NewDynamicRESTMapper(config, options.HTTPClient)
 		if err != nil {
 			return nil, err
 		}
 	}
 
 	resources := &clientRestResources{
-		config: config,
-		scheme: options.Scheme,
-		mapper: options.Mapper,
-		codecs: serializer.NewCodecFactory(options.Scheme),
+		httpClient: options.HTTPClient,
+		config:     config,
+		scheme:     options.Scheme,
+		mapper:     options.Mapper,
+		codecs:     serializer.NewCodecFactory(options.Scheme),
 
 		structuredResourceByType:   make(map[schema.GroupVersionKind]*resourceMeta),
 		unstructuredResourceByType: make(map[schema.GroupVersionKind]*resourceMeta),
 	}
 
-	rawMetaClient, err := metadata.NewForConfig(config)
+	rawMetaClient, err := metadata.NewForConfigAndClient(config, options.HTTPClient)
 	if err != nil {
 		return nil, fmt.Errorf("unable to construct metadata-only client for use as part of client: %w", err)
 	}
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -19,6 +19,7 @@ package cluster
 import (
 	"context"
 	"errors"
+	"net/http"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -37,6 +38,9 @@ import (
 
 // Cluster provides various methods to interact with a cluster.
 type Cluster interface {
+	// GetHTTPClient returns an HTTP client that can be used to talk to the apiserver
+	GetHTTPClient() *http.Client
+
 	// GetConfig returns an initialized Config
 	GetConfig() *rest.Config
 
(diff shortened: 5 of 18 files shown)


pkg/cache/cache.go
```diff
@@ -158,6 +159,15 @@ var defaultResyncTime = 10 * time.Hour
 
 // New initializes and returns a new Cache.
 func New(config *rest.Config, opts Options) (Cache, error) {
+	httpClient, err := rest.HTTPClientFor(config)
```
alvaroaleman, line 162: What is your use-case for this that can not be satisfied by setting a custom `Transport` in the config?

inteon: Here https://github.com/kubernetes/kubernetes/pull/105490 you can find more info about the upstream decision.
This is one of the problem statements that decision was based on: https://github.com/kubernetes/kubernetes/issues/103225.

pkg/cache/internal/informers_map.go
```diff
@@ -60,11 +61,33 @@ type InformersMapOptionsByGVK struct {
 }
 
 // NewInformersMap creates a new InformersMap that can create informers under the hood.
+// Deprecated: use NewInformersMapForConfig instead.
```
vincepri, line 64: This is only used internally (internal package). We should just switch it

pkg/client/client.go
```diff
@@ -74,14 +75,37 @@ type Options struct {
...
+// as well as unstructured types.
+//
+// In the case of normal types, the scheme will be used to look up the
+// corresponding group, version, and kind for the given type.  In the
+// case of unstructured types, the group, version, and kind will be extracted
+// from the corresponding fields on the object.
+// Unlike New, this function allows you to provide your own http.Client.
+func NewForConfigAndClient(config *rest.Config, httpClient *http.Client, options Options) (Client, error) {
+	return newClient(config, httpClient, options)
+}
```
vincepri, line 98: Make this private, and modify the `New` only?

pkg/client/watch.go
```diff
@@ -29,11 +30,21 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-	dynamicClient, err := dynamic.NewForConfig(config)
+	return NewWithWatchForConfigAndClient(config, httpClient, options)
+}
+
+// NewWithWatchForConfigAndClient is a client that can watch objects.
+// It differs from NewForConfigAndClient in that it returns a client that can watch objects.
+func NewWithWatchForConfigAndClient(config *rest.Config, httpClient *http.Client, options Options) (WithWatch, error) {
```
vincepri, line 42: Same as above

pkg/internal/recorder/recorder.go
```diff
@@ -111,7 +112,17 @@ func (p *Provider) getBroadcaster() record.EventBroadcaster {
...
+	httpClient, err := rest.HTTPClientFor(config)
+	if err != nil {
+		return nil, err
+	}
+	return NewProviderForConfigAndClient(config, httpClient, scheme, logger, makeBroadcaster)
+}
+
+// NewProviderForConfigAndClient create a new Provider instance.
+// It differs from NewProvider in that it allows the caller to specify a custom http.Client.
+func NewProviderForConfigAndClient(config *rest.Config, httpClient *http.Client, scheme *runtime.Scheme, logger logr.Logger, makeBroadcaster EventBroadcasterProducer) (*Provider, error) {
```
vincepri, line 124: same as above

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -76,7 +77,19 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
...
+	}
+	return NewDynamicRESTMapperForConfigAndClient(cfg, httpClient, opts...)
+}
+
+// NewDynamicRESTMapperForConfigAndClient returns a dynamic RESTMapper for cfg. The dynamic
+// RESTMapper dynamically discovers resource types at runtime. opts
+// configure the RESTMapper.
+// Unlike NewDynamicRESTMapper, this function allows you to specify a custom http client.
+func NewDynamicRESTMapperForConfigAndClient(cfg *rest.Config, httpClient *http.Client, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
+	client, err := discovery.NewDiscoveryClientForConfigAndClient(cfg, httpClient)
```
vincepri, line 92: same, make this private and let's only have one way to create things

pkg/client/apiutil/apimachinery.go
```diff
@@ -122,6 +123,14 @@ func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConf
 	return rest.RESTClientFor(createRestConfig(gvk, isUnstructured, baseConfig, codecs))
 }
 
+// RESTClientForGVKAndClient constructs a new rest.Interface capable of accessing the resource associated
+// with the given GroupVersionKind. The REST client will be configured to use the negotiated serializer from
+// baseConfig, if set, otherwise a default serializer will be set.
+// Unlike RESTClientForGVK, this function allows the caller to specify a custom http.Client.
+func RESTClientForGVKAndClient(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory, httpClient *http.Client) (rest.Interface, error) {
+	return rest.RESTClientForConfigAndClient(createRestConfig(gvk, isUnstructured, baseConfig, codecs), httpClient)
+}
```
vincepri, line 132: Make the breaking change here? Let's convert the parameters to a options struct and validate them before passing them on

pkg/cache/cache.go
```diff
@@ -158,6 +159,15 @@ var defaultResyncTime = 10 * time.Hour
...
+	if err != nil {
+		return nil, err
+	}
+	return NewForConfigAndClient(config, httpClient, opts)
+}
+
+// NewForConfigAndClient initializes and returns a new Cache.
+func NewForConfigAndClient(config *rest.Config, httpClient *http.Client, opts Options) (Cache, error) {
```
vincepri, line 170: Make this private

pkg/client/client.go
inteon: @vincepri could you expand on what you mean here? Do I leave the New function as-is or do I add the `httpClient *http.Client` to the New function?

vincepri: httpClient can be just an `Options` and default to what `rest.HTTPClientFor` returns

pkg/cache/cache.go
```diff
@@ -414,6 +419,16 @@ func combineTransform(inherited, current toolscache.TransformFunc) toolscache.Tr
 }
 
 func defaultOpts(config *rest.Config, opts Options) (Options, error) {
+	// Use the rest HTTP client for the provided config if unset
+	if opts.HTTPClient == nil {
+		var err error
+		opts.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			log.WithName("setup").Error(err, "Failed to get HTTP client")
```
vincepri, line 427: Could we setup this logger once outside and reuse it?

pkg/client/apiutil/apimachinery.go
```diff
@@ -146,11 +151,29 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 	return gvks[0], nil
 }
 
+// DefaultHTTPClient is a helper function that sets the HTTP client based on a rest config if it is not already set.
+func DefaultHTTPClient(config *rest.Config, httpClient **http.Client) error {
```
vincepri, line 155: Instead of using a double pointer, could we just return a new client?

pkg/client/apiutil/apimachinery.go
```diff
@@ -146,11 +151,29 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
...
+		panic("httpClient must not be nil")
+	}
+
+	if *httpClient != nil {
+		return nil
+	}
+
+	var err error
+	*httpClient, err = rest.HTTPClientFor(config)
+	return err
```
vincepri, line 166: If we return a client and error here, instead of modifying the caller's pointer, we could reuse this function in a lot of other places where the defaulting logic is duplicated

pkg/cluster/cluster.go
```diff
@@ -100,6 +104,11 @@ type Options struct {
 	// will only hold objects from the desired namespace.
 	Namespace string
 
+	// HTTPClient is the http client that will be used to create the default
+	// Cache and Client. If not set the rest.HTTPClientFor function will be used
+	// to create the http client.
+	HTTPClient *http.Client
```
vincepri, line 110: Does this need to be exporter here? Why?

inteon: @vincepri I have to be able to access the client here: https://github.com/kubernetes-sigs/controller-runtime/pull/2122/files#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR368-R382

pkg/client/watch.go
```diff
@@ -29,11 +29,21 @@ import (
 
 // NewWithWatch returns a new WithWatch.
 func NewWithWatch(config *rest.Config, options Options) (WithWatch, error) {
+	// Use the rest HTTP client for the provided config if unset
+	if options.HTTPClient == nil {
+		var err error
+		options.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			return nil, err
+		}
```
vincepri, line 38: Could we use apiutil.DefaultHTTPClient?

pkg/cluster/cluster.go
```diff
@@ -201,15 +215,23 @@ func New(config *rest.Config, opts ...Option) (Cluster, error) {
...
 
 // setOptionsDefaults set default values for Options fields.
-func setOptionsDefaults(options Options) Options {
+func setOptionsDefaults(options Options, config *rest.Config) (Options, error) {
+	if options.HTTPClient == nil {
+		var err error
+		options.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			return options, err
+		}
```
vincepri, line 224: Could we use apiutil.DefaultHTTPClient?

pkg/cache/cache.go
```diff
@@ -414,6 +419,18 @@ func combineTransform(inherited, current toolscache.TransformFunc) toolscache.Tr
...
+	logger := log.WithName("setup")
+
+	// Use the rest HTTP client for the provided config if unset
+	if opts.HTTPClient == nil {
+		var err error
+		opts.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			logger.Error(err, "Failed to get HTTP client")
+			return opts, fmt.Errorf("could not create HTTP client from config")
+		}
```
vincepri, line 431: Could we use apiutil.DefaultHTTPClient?

pkg/client/apiutil/apimachinery.go
```diff
@@ -146,11 +152,24 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
...
+
+	return rest.HTTPClientFor(config)
+}
+
 // RESTClientForGVK constructs a new rest.Interface capable of accessing the resource associated
 // with the given GroupVersionKind. The REST client will be configured to use the negotiated serializer from
 // baseConfig, if set, otherwise a default serializer will be set.
-func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory) (rest.Interface, error) {
-	return rest.RESTClientFor(createRestConfig(gvk, isUnstructured, baseConfig, codecs))
+func RESTClientForGVK(gvk schema.GroupVersionKind, isUnstructured bool, baseConfig *rest.Config, codecs serializer.CodecFactory, httpClient *http.Client) (rest.Interface, error) {
```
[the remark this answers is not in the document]
inteon: Indeed, we use this function internally, but the httpClient parameter will always be provided:
- https://github.com/kubernetes-sigs/controller-runtime/pull/2122/files#diff-a2af599574fb354cc98a1fb082e380b56c886d048c92d6c1596a1e20cd6e9ab3R66
- https://github.com/kubernetes-sigs/controller-runtime/pull/2122/files#diff-75eedff1c8b986479db1b4e5a4cd577484737bff44772c54b587f535473b783dR444

pkg/client/apiutil/dynamicrestmapper.go
```diff
@@ -75,8 +76,13 @@ func WithCustomMapper(newMapper func() (meta.RESTMapper, error)) DynamicRESTMapp
 // NewDynamicRESTMapper returns a dynamic RESTMapper for cfg. The dynamic
 // RESTMapper dynamically discovers resource types at runtime. opts
 // configure the RESTMapper.
-func NewDynamicRESTMapper(cfg *rest.Config, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
-	client, err := discovery.NewDiscoveryClientForConfig(cfg)
+func NewDynamicRESTMapper(cfg *rest.Config, httpClient *http.Client, opts ...DynamicRESTMapperOption) (meta.RESTMapper, error) {
```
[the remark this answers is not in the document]
inteon: We need the parameter to make sure we respect the HTTP client that was passed by the user via eg. options.
- https://github.com/kubernetes-sigs/controller-runtime/pull/2122/files#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R122

pkg/client/apiutil/apimachinery.go
```diff
@@ -146,11 +152,24 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 	return gvks[0], nil
 }
 
+// DefaultHTTPClient is a helper function that sets the HTTP client based on a rest config if it is not already set.
+func DefaultHTTPClient(config *rest.Config, httpClient *http.Client) (*http.Client, error) {
+	if httpClient != nil {
+		return httpClient, nil
+	}
```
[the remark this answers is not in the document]
vincepri: That's probably fine, we have "defaulting" functions in lots of places around the codebase. The main thing I would like to avoid is to pass `nil` parameters around

pkg/cache/cache.go
```diff
@@ -414,6 +419,18 @@ func combineTransform(inherited, current toolscache.TransformFunc) toolscache.Tr
...
 
 func defaultOpts(config *rest.Config, opts Options) (Options, error) {
+	logger := log.WithName("setup")
+
+	// Use the rest HTTP client for the provided config if unset
+	if opts.HTTPClient == nil {
+		var err error
+		opts.HTTPClient, err = rest.HTTPClientFor(config)
+		if err != nil {
+			logger.Error(err, "Failed to get HTTP client")
```
sbueringer, line 429: This log + return error pattern seems strange. Does anyone know why we're doing this in a bunch of places?

inteon:
> @inteon
> 
> What problem are we trying to solve with this PR?
> 
> Is it about that controller-runtime should use the same http.Client everywhere?
> 
> This seems reasonable, but I'm not sure if that necessarily means that we have to expose http.Client as a parameter everywhere. What is the use case for that?

I'm trying to solve both.
I think we should allow configuring the HTTP client so we can reuse the client when manually creating controller-runtime clients or when you want to use the HTTP client when you are creating clients using the client-go API.

sbueringer:
> > @inteon
> > What problem are we trying to solve with this PR?
> > Is it about that controller-runtime should use the same http.Client everywhere?
> > This seems reasonable, but I'm not sure if that necessarily means that we have to expose http.Client as a parameter everywhere. What is the use case for that?
> 
> I'm trying to solve both. I think we should allow configuring the HTTP client so we can reuse the client when manually creating controller-runtime clients or when you want to use the HTTP client when you are creating clients using the client-go API.

I'm mostly asking because we already have a lot of options and with this implementation we now pass a lot of nils around and have a lot of defaulting everywhere.

If the goal is to be able to share a single http.Client between client-go and clients created via controller-runtime we could instead also use a similar pattern that we already have for the logger.

Essentially there we have a `ctrl.SetLogger` func which allos setting the logger and then we can use it from there whenever we need a logger. With http client it cloud looks like this:
* ctrl.SetHttpClient stores the http client
* in all places where we want to pass the http client "down" to client-go we can get the http client from there and implement the defaulting once in that get func

inteon:
I updated the PR and reduced the number of places where we set a default value for the httpClient.
Now, the httpClient value is defaulted only when we receive `Options` objects.
All non-user-facing functions now just panic if the httpClient == nil.

alvaroaleman:
@inteon so just for my understanding, the reason for wanting this is not for testing/mocking purposes it is to reduce the number of TCP connections we end up creating? If so, please update the description of the PR to reflect that.

vincepri (commented):
/lgtm

inteon:
@alvaroaleman @vincepri I added some additional info to the PR description.

sbueringer:
Thx!

/lgtm

vincepri (approved):
/approve

pkg/cache/cache.go
```diff
@@ -414,6 +419,18 @@ func combineTransform(inherited, current toolscache.TransformFunc) toolscache.Tr
...
+			return opts, fmt.Errorf("could not create HTTP client from config")
```
alvaroaleman, line 430: return the original error

inteon: @alvaroaleman @vincepri this comment was posted just a bit too late, I think it is a valid remark. Do I create a new PR to fix this issue?

sbueringer: Yup just create a follow-up PR that's fine

inteon: see https://github.com/kubernetes-sigs/controller-runtime/pull/2161

pkg/cluster/cluster.go
```diff
@@ -148,24 +157,29 @@ func New(config *rest.Config, opts ...Option) (Cluster, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	writeObj, err := options.NewClient(config, client.Options{
-		Scheme: options.Scheme,
-		Mapper: mapper,
+		HTTPClient: options.HTTPClient,
+		Scheme:     options.Scheme,
+		Mapper:     mapper,
```
alvaroaleman, line 182: @inteon I can't comment on that but can we also use the httpClient when constructing the apireader on line 197?

pkg/client/apiutil/apimachinery.go
```diff
@@ -61,9 +62,13 @@ func AddToProtobufScheme(addToScheme func(*runtime.Scheme) error) error {
 
 // NewDiscoveryRESTMapper constructs a new RESTMapper based on discovery
 // information fetched by a new client with the given config.
-func NewDiscoveryRESTMapper(c *rest.Config) (meta.RESTMapper, error) {
+func NewDiscoveryRESTMapper(c *rest.Config, httpClient *http.Client) (meta.RESTMapper, error) {
+	if httpClient == nil {
```
alvaroaleman, line 66: the apiutil package is not actually internal but exported, can we return an error if this is nil here and in `RESTClientForGVK`?

I just find it somewhat debatable that we make this a mandatory arg now, since this can not be just any http client but must be constructed from the rest config.

inteon: I made the parameter mandatory because we wanted to not default in multiple places (see comments above).
If we expect users to directly use this function, I would argue that defaulting wouldn't be a bad idea?
I'll create a PR to change the panic to an error & let's continue the discussion there.

sbueringer: @inteon Looks like this one still resturns a panic. Was that intended?

sbueringer: Sorry not this one. `NewDynamicRESTMapper` in dynamicrestmapper.go

inteon: https://github.com/kubernetes-sigs/controller-runtime/pull/2164

inteon: @sbueringer this PR fixes the issue for `NewDynamicRESTMapper` too: https://github.com/kubernetes-sigs/controller-runtime/pull/2178

sbueringer: Thx!

PR #2131, opened 2023-01-17
🌱 Simplify cache.objectTypeForListObject

vincepri:
This changeset streamlines and simplifies the logic inside objectTypeForListObject. This method is used to retrieve from a List object, its respective actual resource GVK and a new instance of the object.

In the previous iteration, the use of reflection was complicated to follow and looking at the fields in the struct passed in, instead of using the information we already have in the Scheme to create a new instance of the expected object and return.

In addition to all of the above, we're also now handling PartialObjectMetadataList(s).

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -19,10 +19,10 @@ package cache
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


vincepri (commented):
/assign @sbueringer

sbueringer:
Looks good

/lgtm

Sometimes wonder if a bit more error wrapping could be beneficial (but the code fits into the current style that we have in this regard)

PR #2134, opened 2023-01-18
:warning: Remove pkg/inject

vincepri:
This changeset brings in a few more breaking changes to remove the rest of the dependency injection code.

The webhook code is now a bit simplified and more explicit in accepting a *runtime.Scheme. The logger is now also not configured in the Standalone mode. We should revisit in a future PR the webhook standalone configuration, and potentially always require a manager. From a quick look in GitHub projects, it seems most folks don't actually use these standalone webhooks, which makes sense because the Manager makes things easier, although there might be private code that's using them.

The config loader was using InjectScheme before, although it really didn't need to.


sbueringer:
> We should revisit in a future PR the webhook standalone configuration, and potentially always require a manager. From a quick look in GitHub projects, it seems most folks don't actually use these standalone webhooks, which makes sense because the Manager makes things easier, although there might be private code that's using them.

We are using a standalone webhook server at the moment in CAPI Runtime Extensions: https://github.com/kubernetes-sigs/cluster-api/blob/b6e824c92c4c0d85050b9dc09ba99ca23b238f7a/test/extension/main.go#L141

But we probably should switch to using a manager anyway as it gives us a few more useful features.
(cc @fabriziopandini @killianmuldoon)

pkg/webhook/conversion/decoder.go
```diff
@@ -31,6 +31,9 @@ type Decoder struct {
 
 // NewDecoder creates a Decoder given the runtime.Scheme
 func NewDecoder(scheme *runtime.Scheme) (*Decoder, error) {
```
sbueringer, line 33: Would it make sense to also drop the error return parameter here?

pkg/webhook/server.go
```diff
@@ -141,51 +134,11 @@ func (s *Server) Register(path string, hook http.Handler) {
...
-			regLog.Error(err, "unable to inject fields into webhook during registration")
-		}
-
-		baseHookLog := log.WithName("webhooks")
-
-		// NB(directxman12): we don't propagate this further by wrapping setFields because it's
-		// unclear if this is how we want to deal with log propagation.  In this specific instance,
-		// we want to be able to pass a logger to webhooks because they don't know their own path.
-		if _, err := inject.LoggerInto(baseHookLog.WithValues("webhook", path), hook); err != nil {
-			regLog.Error(err, "unable to logger into webhook during registration")
```
sbueringer, line 166: If I see correctly, we're loosing a few k/v pairs by dropping the log injection here and replacing it by just using the global logger.

Would be good to cleanup the logging story for webhooks eventually. 

There's also this PR: https://github.com/kubernetes-sigs/controller-runtime/pull/1972 which changes the logging for webhooks.

Should we consider not dropping the logger here and passing it through explicitly? (if that's possible, I'm lacking a bit overview to be honest)

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -46,15 +46,11 @@ type defaulterForType struct {
...
-var _ DecoderInjector = &defaulterForType{}
-
-func (h *defaulterForType) InjectDecoder(d *Decoder) error {
-	h.decoder = d
-	return nil
-}
-
 // Handle handles admission requests.
 func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
+	if h.decoder == nil {
```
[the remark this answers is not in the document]
vincepri: done!

examples/builtins/mutatingwebhook.go
```diff
@@ -60,12 +60,3 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
-
-// podAnnotator implements admission.DecoderInjector.
-// A decoder will be automatically injected.
-
-// InjectDecoder injects the decoder.
-func (a *podAnnotator) InjectDecoder(d *admission.Decoder) error {
```
alvaroaleman, line 68: Can we update the `main.go` file to set the decoder for the two webhooks so the example actually works?

examples/tokenreview/tokenreview.go
```diff
@@ -29,7 +29,7 @@ type authenticator struct {
 }
 
 // authenticator admits a request by the token.
-func (a *authenticator) Handle(ctx context.Context, req authentication.Request) authentication.Response {
+func (a *authenticator) Handle(ctx context.Context, req *authentication.Request) authentication.Response {
```
alvaroaleman, line 32: What is the reason to change the signature here?

pkg/builder/example_test.go
```diff
@@ -155,8 +155,3 @@ func (a *ReplicaSetReconciler) Reconcile(ctx context.Context, req reconcile.Requ
 
 	return reconcile.Result{}, nil
 }
-
-func (a *ReplicaSetReconciler) InjectClient(c client.Client) error {
```
alvaroaleman, line 159: Same here, please update to set the client so the example will actually work (Unrelated to this change, it would be good if we tested that)

pkg/webhook/conversion/decoder.go
```diff
@@ -30,8 +30,11 @@ type Decoder struct {
 }
 
 // NewDecoder creates a Decoder given the runtime.Scheme
-func NewDecoder(scheme *runtime.Scheme) (*Decoder, error) {
-	return &Decoder{codecs: serializer.NewCodecFactory(scheme)}, nil
+func NewDecoder(scheme *runtime.Scheme) *Decoder {
+	if scheme == nil {
+		panic("scheme should never be nil")
```
alvaroaleman, line 35: lol at the old code. But why panic rather than returning an err and keeping the signature?

pkg/webhook/example_test.go
```diff
@@ -80,75 +76,3 @@ func Example() {
...
 	}
 }
-
-// This example creates a webhook server that can be
-// ran without a controller manager.
-//
-// Note that this assumes and requires a valid TLS
-// cert and key at the default locations
-// tls.crt and tls.key.
-func ExampleServer_StartStandalone() {
```
alvaroaleman, line 90: What is the reason for removing this?

pkg/webhook/conversion/decoder.go
vincepri: Given that NewDecoder should only be called internally, I opted for a panic (it also looks like there are other places that we check fields and panic)

I'm also fine just removing every panic honestly, if that's what we prefer?

pkg/webhook/example_test.go
vincepri: StandaloneServer here makes it more difficult to support the Webhook code long term, looking at `grep.app` this code is not used on publicly available code at least. The Standalone seems it was added to use webhooks outside of a Manager, but really lots of things from the manager are then required (like scheme, or client).

Going forward, we should try to make the Manager as slim as possible, and require it explicitly

examples/tokenreview/tokenreview.go
vincepri: I can revert if we want, it was to avoid copying around the Request object

alvaroaleman: I don't have strong feelings either way, its a slight optimization that likely rarely matter. But if we do it, can we do it in a different PR so this gets a distinct release note? It seems unrelated to the rest

pkg/webhook/conversion/decoder.go
alvaroaleman: If it should only be called internally, it shouldn't be exported IMHO

vincepri: Well, it looks like our examples are using it, even though there the builder would achieve the same thing?

pkg/webhook/example_test.go
alvaroaleman: grep.app only indexes the N most popluar repos, so it is somewhat useless when finding usages in potentially non-popular repos. Trough cs.github.com I can find for example these usages:
* https://cs.github.com/wking/ci-tools/blob/64bde4a5f06d600725ddcc59f8d88410ec4fbe35/cmd/pod-scaler/admission.go?q=StartStandalone+language%3AGo#L39
* https://cs.github.com/openshift/cluster-api/blob/1101b984bc6d4e0c81dcb6d6e4651dd6dfbdbec8/exp/runtime/server/server.go?q=StartStandalone+language%3AGo#L239
* https://cs.github.com/varshavaradarajan/webhook-cr-poc/blob/7da2130be1d295be447196b77d5f4fbebe3af652/cmd/webhook-cr-poc/main.go?q=StartStandalone+language%3AGo#L40

So it does seem to be used.

> but really lots of things from the manager are then required (like scheme, or client).

Both of these can be constructed just fine without a manager.

pkg/webhook/conversion/decoder.go
alvaroaleman: Maybe change them to either deal with the erroring version or make them use the builder?

pkg/webhook/example_test.go
vincepri: reverted, thanks!

pkg/webhook/conversion/decoder.go
vincepri: Let's do this as a follow-up? A panic here is probably ok, given that scheme is a parameter, not sure what one would expect if you do NewDecoder(nil) 🤷

vincepri: Fixed the examples/builtin one, there are still ways that folks might want to use a decoder, let's keep it exported for now. We might want to revisit most of the webhook package at some point, it's really cumbersome and there is lots of repetition between admission/conversion/authentication which can be avoided. It could also be good to see if we can do something with generics that's a bit more elegant

examples/builtins/main.go
```diff
@@ -71,13 +71,14 @@ func main() {
...
-	// Setup webhooks
-	entryLog.Info("setting up webhook server")
-	hookServer := mgr.GetWebhookServer()
-
-	entryLog.Info("registering webhooks to the webhook server")
-	hookServer.Register("/mutate-v1-pod", &webhook.Admission{Handler: &podAnnotator{Client: mgr.GetClient()}})
-	hookServer.Register("/validate-v1-pod", &webhook.Admission{Handler: &podValidator{Client: mgr.GetClient()}})
+	if err := builder.WebhookManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithDefaulter(&podAnnotator{}).
```
alvaroaleman, line 76: Can we update the annotator and defaulter to remote their now-unused client and scheme attribute?

pkg/webhook/example_test.go
alvaroaleman: You commented that you reverted this but its still being removed?

vincepri: The `StartStandalone` part isn't necessary now, the scheme is neither used or injected, `webhook.Start` should work here

examples/builtins/main.go
vincepri: Done!

alvaroaleman (approved):

terinjokes:
It seems like this will be a breaking change for implementers of Reconciler without a deprecation window. The documentation previously stated the injection package was deprecated, not the injection mechanism.

vincepri:
@terinjokes These packages have been deprecated since 0.10 and should have been removed but never did. The release notes will reflect these changes.

terinjokes:
Just providing feedback that deprecation notices in the future should be very clear about what they apply to. The packages are deprecated with a notice, sure, but that could have been intepreted as "don't import this package for these interfaces, just implement them" rather then "the entire documented injection system is going away".

PR #2136, opened 2023-01-19
⚠️ Add client.{GroupVersionKindFor, IsObjectNamespaced}

vincepri:
Add commonly used utilities that can be nicely wrapped in the built-in clients, in more details GroupVersionKindFor(obj) can be used to retrieve the GVK for a given object, while IsObjectNamespace can be used to determine if an object is global, or namespace scoped.

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -20,6 +20,7 @@ limitations under the License.
 package apiutil
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"sync"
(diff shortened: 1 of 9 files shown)


aliok:
LGTM

I required something like this in more than 1 place and it is definitely useful to see it provided by controller-runtime.

pkg/client/apiutil/apimachinery.go
```diff
@@ -72,6 +73,36 @@ func NewDiscoveryRESTMapper(c *rest.Config) (meta.RESTMapper, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	return IsAPINamespacedWithGVK(gvk, scheme, restmapper)
+}
+
+// IsAPINamespacedWithGVK returns true if the object having the provided
+// GVK is namespace scoped.
+func IsAPINamespacedWithGVK(gk schema.GroupVersionKind, scheme *runtime.Scheme, restmapper meta.RESTMapper) (bool, error) {
```
sbueringer, line 89:
```suggestion
func IsAPINamespacedWithGVK(gvk schema.GroupVersionKind, scheme *runtime.Scheme, restmapper meta.RESTMapper) (bool, error) {
```
nit

pkg/client/client.go
```diff
@@ -168,6 +168,16 @@ func (c *client) resetGroupVersionKind(obj runtime.Object, gvk schema.GroupVersi
 	}
 }
 
+// GroupVersionKindFor returns the GroupVersionKind for the given object.
+func (c *client) GroupVersionKindFor(obj Object) (schema.GroupVersionKind, error) {
+	return apiutil.GVKForObject(obj, c.scheme)
+}
+
+// IsObjectNamespaced returns true if the GroupVersionKind of the object is namespaced.
+func (c *client) IsObjectNamespaced(ctx context.Context, obj Object) (bool, error) {
```
sbueringer, line 177: Why do we take a ctx here?

sbueringer: I think it's unused, but maybe I missed an implementation

pkg/client/apiutil/apimachinery.go
```diff
@@ -72,6 +73,36 @@ func NewDiscoveryRESTMapper(c *rest.Config) (meta.RESTMapper, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	return IsGVKNamespaced(gvk, scheme, restmapper)
+}
+
+// IsGVKNamespaced returns true if the object having the provided
+// GVK is namespace scoped.
+func IsGVKNamespaced(gvk schema.GroupVersionKind, scheme *runtime.Scheme, restmapper meta.RESTMapper) (bool, error) {
```
sbueringer, line 89: scheme is unused, maybe drop?

vincepri: done, thanks!

PR #2142, opened 2023-01-26
:seedling:fix default value for cfg.Burst

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

fix config default value.
when the cfg.Burst!=0 , can not change it value.

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -98,12 +98,12 @@ func GetConfigWithContext(context string) (*rest.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-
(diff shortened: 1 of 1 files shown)


sbueringer:
Seems reasonable to me. client-go also has separate defaults (even though different ones than the ones we have)
/lgtm

vincepri:
What/Where are the default client-go values that you were looking at?

sbueringer:
I was looking here: https://github.com/kubernetes/client-go/blob/master/rest/config.go#L43-L46

But it's probably too much to actually call them "client go defaults".

vincepri (approved):
/approve

PR #2150, opened 2023-01-27
⚠ Remove DelegatedClient, move Options in client.New

vincepri:


vincepri:
/retest 

flake

pkg/manager/manager.go
```diff
@@ -264,7 +264,7 @@ type Options struct {
 	// NewClient is the func that creates the client to be used by the manager.
 	// If not set this will create the default DelegatingClient that will
 	// use the cache for reads and the client for writes.
-	NewClient cluster.NewClientFunc
+	NewClient client.NewClientFunc
```
FillZpp, line 267: The comment that says `DelegatingClient` should be updated.

pkg/cluster/cluster.go
```diff
@@ -108,7 +108,7 @@ type Options struct {
 	// If not set this will create the default DelegatingClient that will
 	// use the cache for reads and the client for writes.
 	// NOTE: The default client will not cache Unstructured.
-	NewClient NewClientFunc
+	NewClient client.NewClientFunc
```
FillZpp, line 111: The comment that says `DelegatingClient` should be updated.

FillZpp (commented):
Nice change, I have seen lots of ppl confused by the hidden DelegatedClient.

pkg/cluster/cluster.go
vincepri: fixed, thanks!

pkg/manager/manager.go
vincepri: fixed, thanks!

PR #2157, opened 2023-01-30
⚠ Refactor cache.Options, deprecate MultiNamespacedCacheBuilder

vincepri:
This changeset refactors the entire cache.Options struct to provide a legible and clear way to configure parameters when a cache is created. In addition, it handles under the hood the automatic multi-namespace cache support we currently offer through the
MultiNamespacedCacheBuilder, which is now deprecated.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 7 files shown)


pkg/cache/internal/informers.go
```diff
@@ -500,15 +500,13 @@ func newGVKFixupWatcher(gvk schema.GroupVersionKind, watcher watch.Interface) wa
 	)
 }
 
-// resyncPeriod returns a function which generates a duration each time it is
+// calculateResyncPeriod returns a function which generates a duration each time it is
```
sbueringer, line 503: nit: Doesn't return a func anymore

pkg/cache/cache.go
```diff
@@ -114,44 +115,59 @@ type Options struct {
...
+// to the desired parameter.
+type ViewOptions struct {
+	// Namespaces restricts the cache's ListWatch to the desired namespaces
+	// Default watches all namespaces
+	Namespaces []string
 
 	// DefaultSelector will be used as selectors for all object types
-	// that do not have a selector in SelectorsByObject defined.
+	// unless they have a more specific selector set in ByObject.
 	DefaultSelector ObjectSelector
```
sbueringer, line 139: I wonder if it would make sense to move DefaultSelector and DefaultTransform together down into a `Default` field?

(fine for me either way)

pkg/cache/cache.go
```diff
@@ -277,8 +301,8 @@ func selectResync(def, override *time.Duration) *time.Duration {
 	return def
 }
 
-func selectNamespace(def, override string) string {
-	if override != "" {
+func selectNamespaces(def, override []string) []string {
+	if len(override) > 0 {
```
sbueringer, line 305: Do we have to check for nil here instead? Otherwise you can't overwrite with an empty slice. Or is that not a relevant use case?

pkg/cache/cache_unit_test.go
```diff
@@ -93,73 +93,73 @@ var _ = Describe("cache.inheritFrom", func() {
...
-			specified.Resync = pointer.Duration(time.Second)
-			inherited.Resync = pointer.Duration(time.Minute)
-			Expect(checkError(specified.inheritFrom(inherited)).Resync).To(Equal(specified.Resync))
+			specified.ResyncEvery = pointer.Duration(time.Second)
+			inherited.ResyncEvery = pointer.Duration(time.Minute)
+			Expect(checkError(specified.inheritFrom(inherited)).ResyncEvery).To(Equal(specified.ResyncEvery))
 		})
 	})
 	Context("Namespace", func() {
 		It("is NamespaceAll when specified and inherited are unset", func() {
```
sbueringer, line 113: Either the title or the validation is not correct anymore

sbueringer (commented):
Just a few smaller nits/findings

pkg/cache/cache.go
vincepri: I tried both `Global.Selector`, or `GlobalSelector`, or `Default.Selector`; in code it's easy to mistake between the options, so figured to keep the one word ones `DefaultSelector`

pkg/cache/cache_unit_test.go
vincepri: fixed, thanks!

pkg/cache/cache.go
sbueringer: Sounds good

pkg/cache/cache.go
sbueringer: The last time I checked this it was a difference. Let me double-check

sbueringer: https://go.dev/play/p/jF2-tgk93OB

vincepri: I _think_ that works as I would expect, the default is a nil slice, which should never overwrite anything. Only slices with at least one member should, so it works like before

sbueringer: Yup that's why I asked. Do we ever want to overwrite an existing value with an empty slice (which would then lead to using all namespaces).

But I guess in that case folks can also just overwrite it with a slice with the All value

sbueringer:
/lgtm

We can follow-up if someone has a different opinion on https://github.com/kubernetes-sigs/controller-runtime/pull/2157#discussion_r1092308960

pkg/cache/cache.go
```diff
@@ -118,44 +119,59 @@ type Options struct {
...
-	// list objects per GVK at the specified object.
-	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
-	// otherwise you will mutate the object in the cache.
-	UnsafeDisableDeepCopyByObject DisableDeepCopyByObject
+	// DefaultTransform will be used as transform for all object types
+	// unless they have a more specific transform set in ByObject.
+	DefaultTransform toolscache.TransformFunc
+
+	// ByObject restricts the cache's ListWatch to the desired fields per GVK at the specified object.
+	ByObject ViewByObject
```
alvaroaleman, line 150: I don't really understand how it improves anything to move this into a substruct, IMHO this just makes it harder to find what options exist.

What might have improved things would be if this was the object -> object-level options map rather than a struct that embedds a bunch of maps.

pkg/cache/cache.go
```diff
@@ -118,44 +119,59 @@ type Options struct {
...
+	Transform TransformByObject
 
-	// DefaultTransform is the transform used for all GVKs which do
-	// not have an explicit transform func set in TransformByObject
-	DefaultTransform toolscache.TransformFunc
+	// UnsafeDisableDeepCopy indicates not to deep copy objects during get or
+	// list objects per GVK at the specified object.
+	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
+	// otherwise you will mutate the object in the cache.
+	UnsafeDisableDeepCopy DisableDeepCopyByObject
```
alvaroaleman, line 174: This doesn't really have anything to do with `viewing` objects

pkg/cache/cache.go
```diff
@@ -118,44 +119,59 @@ type Options struct {
...
 	// So that all informers will not send list requests simultaneously.
-	Resync *time.Duration
+	ResyncEvery *time.Duration
 
-	// Namespace restricts the cache's ListWatch to the desired namespace
-	// Default watches all namespaces
-	Namespace string
+	// View restricts the cache's ListWatch to the desired fields per GVK
+	// Default watches all fields and all namespaces.
+	View ViewOptions
```
alvaroaleman, line 130: I am really, really not a fan of moving thins into substructs. It is much easier to find what options exist if they are on the top level.

alvaroaleman (commented):
While its nice to see some work getting done in here I would have much preferred if we had started off with some kind of design doc that includes what options we want to end up supporting with the cache and how we configure them. Right now we are going back to changing in to whatever seems best at the moment without any clear vision on how this should look in the end.

I also feel we are getting a bit too careless with doing breaking changes, they have a cost for many of our users and we should avoid doing them unless they clearly and obviously improve things.

pkg/cache/cache.go
vincepri: I'll follow-up and move this one out

pkg/cache/cache.go
vincepri: Do you have a better name? Having all of the selectors and transformers at the top level IMO it's super confusing for new users. It's also not clear at all that all Get() and List() requests would be limited to those for example

PR #2158, opened 2023-01-31
🐛 Don't share error concurrently in test

inteon:
Two goroutines were writing to the same error variable.


vincepri (approved):
/ok-to-test
/approve
/lgtm

PR #2159, opened 2023-01-31
⚠ Support shutdown watches dynamically (v2)

inteon:
⚠️ depends on https://github.com/kubernetes-sigs/controller-runtime/pull/2315

PTAL at the original PR: https://github.com/kubernetes-sigs/controller-runtime/pull/2099

I started making some small changes to the original PR, thinking there would be an easy fix. However, it turned out that there were still a lot of fundamental questions requiring more time and effort to get fixed.

The current state of the PR is best described by https://github.com/kubernetes-sigs/controller-runtime/pull/2159#issuecomment-1605271769.

(4/4) PR chain:
- https://github.com/kubernetes-sigs/controller-runtime/pull/2685
- https://github.com/kubernetes-sigs/controller-runtime/pull/2686
- https://github.com/kubernetes-sigs/controller-runtime/pull/2315
- https://github.com/kubernetes-sigs/controller-runtime/pull/2159


FillZpp (commented):
Thanks @inteon , nice work!

It seems you have removed the `Stop()` from Controller, so how should users stop one or more specific controller?

inteon:
> Thanks @inteon , nice work!
> 
> It seems you have removed the `Stop()` from Controller, so how should users stop one or more specific controller?

@FillZpp what do you think about a more generic solution for canceling runnables:
```go
// NewCancelRunnable returns a new Runnable and a function to cancel that runnable.
func NewCancelRunnable(r Runnable) (Runnable, context.CancelFunc) {
	mu := sync.Mutex{}
	cancel := (func())(nil)

	return RunnableFunc(func(ctx context.Context) error {
			mu.Lock()
			ctx, cancel = context.WithCancel(ctx)
			mu.Unlock()

			return r.Start(ctx)
		}), func() {
			mu.Lock()
			defer mu.Unlock()
			if cancel != nil {
				cancel()
			}
		}
}
```

FillZpp:
> > Thanks @inteon , nice work!
> > It seems you have removed the `Stop()` from Controller, so how should users stop one or more specific controller?
> 
> @FillZpp what do you think about a more generic solution for canceling runnables:
> 
> ```go
> // NewCancelRunnable returns a new Runnable and a function to cancel that runnable.
> func NewCancelRunnable(r Runnable) (Runnable, context.CancelFunc) {
> 	mu := sync.Mutex{}
> 	cancel := (func())(nil)
> 	canceled := false
> 
> 	return RunnableFunc(func(ctx context.Context) error {
> 			mu.Lock()
> 			ctx, cancel = context.WithCancel(ctx)
> 			if canceled {
> 				cancel()
> 			}
> 			mu.Unlock()
> 
> 			return r.Start(ctx)
> 		}), func() {
> 			mu.Lock()
> 			defer mu.Unlock()
> 			canceled = true
> 			if cancel != nil {
> 				cancel()
> 			}
> 		}
> }
> ```

Emm, I think it's alright. It may not be intuitive as the Stop method, but more generic as you said.

pkg/internal/source/kind.go
```diff
@@ -104,14 +97,10 @@ func (ks *Kind) String() string {
...
-			return nil
-		}
-		return errors.New("timed out waiting for cache to be synced")
-	}
+	return ks.informer.WaitForSync(ctx)
+}
+
+// Stop implements StoppableSource to stop it dynamically.
+func (ks *Kind) Stop() error {
+	return ks.informer.Stop()
```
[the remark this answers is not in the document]
inteon: Thanks, I fixed the issue in https://github.com/kubernetes-sigs/controller-runtime/pull/2159/commits/8de47097d6ff16395d719d712b3fcafb74943c3e.

FillZpp:
fixes #1884

jnan806:
> > Thanks @inteon , nice work!
> > It seems you have removed the `Stop()` from Controller, so how should users stop one or more specific controller?
> 
> @FillZpp what do you think about a more generic solution for canceling runnables:
> 
> ```go
> // NewCancelRunnable returns a new Runnable and a function to cancel that runnable.
> func NewCancelRunnable(r Runnable) (Runnable, context.CancelFunc) {
> 	mu := sync.Mutex{}
> 	cancel := (func())(nil)
> 	canceled := false
> 
> 	return RunnableFunc(func(ctx context.Context) error {
> 			mu.Lock()
> 			ctx, cancel = context.WithCancel(ctx)
> 			if canceled {
> 				cancel()
> 			}
> 			mu.Unlock()
> 
> 			return r.Start(ctx)
> 		}), func() {
> 			mu.Lock()
> 			defer mu.Unlock()
> 			canceled = true
> 			if cancel != nil {
> 				cancel()
> 			}
> 		}
> }
> ```

@inteon   As a user, how can i use this func to stop controller ?

pkg/manager/manager.go
```diff
@@ -341,6 +342,34 @@ type LeaderElectionRunnable interface {
...
+		}), func() {
+			mu.Lock()
+			defer mu.Unlock()
+			canceled = true
+			for _, cancel := range cancels {
+				cancel()
+			}
+		}
+}
+
```
jnan806, line 372: @inteon  As a user, how can i use this func to stop controller ?

pkg/internal/source/informer.go
```diff
@@ -0,0 +1,176 @@
...
+
+		is.mu.Unlock()
+		// Wait for starting abort
+		<-is.started
+		is.mu.Lock()
+	}
+
+	if is.eventHandlerRegistration != nil {
+		is.eventHandlerCancel()
+		return is.informer.RemoveEventHandler(is.eventHandlerRegistration)
```
maxsmythe, line 173: To re-raise a thread from another pr, it looks like this doesn't tear down the informer if there are no longer any event handlers:

https://github.com/kubernetes-sigs/controller-runtime/pull/2099#discussion_r1051660956

IIRC the comment on the old PR was cache cleanup would be a follow-on change. Is that the same idea here?

inteon: Yes, I think the informer has to be responsible for dynamically starting and shutting down when there are no EventHandlers registered anymore.

maxsmythe: I'm okay with dynamic, but having a manual method as an MVP may be a lower bar.

pkg/manager/manager.go
```diff
@@ -341,6 +342,34 @@ type LeaderElectionRunnable interface {
 	NeedLeaderElection() bool
 }
 
+// NewCancelRunnable returns a new Runnable and a function to cancel that runnable.
+func NewCancelRunnable(r Runnable) (Runnable, context.CancelFunc) {
```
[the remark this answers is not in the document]
vincepri: To resemble context.WithCancel

pkg/internal/controller/controller.go
```diff
@@ -58,12 +57,18 @@ type Controller struct {
 	// the Queue for processing
 	Queue workqueue.RateLimitingInterface
 
+	// startedSources maintains a list of sources that have already started.
+	startedSources []source.Source
```
varshaprasad96, line 61: Just curious, would it be possible to not stop the controller but selectively remove watches for certain GVKs? That was been one of the common asks by the operator authors.

pkg/source/source.go
```diff
@@ -46,8 +47,12 @@ const (
 // Users may build their own Source implementations.
 type Source interface {
 	// Start is internal and should be called only by the Controller to register an EventHandler with the Informer
-	// to enqueue reconcile.Requests.
+	// to enqueue reconcile.Requests. It should NOT block, instead, it should start a goroutine and return immediately.
```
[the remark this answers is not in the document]
inteon: Yes, a source's `Start` function does not block currently.

pkg/manager/manager.go
inteon: I added a test here: https://github.com/kubernetes-sigs/controller-runtime/pull/2159/files#diff-21b67ad2c12fb31030a5f2e93e1f1e1593b7375402163a660bc056e126cb0cfbR349-R372

pkg/internal/controller/controller.go
inteon: It should be possible to keep a reference to the source before watching it in your controller, you can then call the `Stop` function on this source reference. I'm not sure if there is a better/ more elegant way to do this.

varshaprasad96: Thanks @inteon. I'll test it out!

inteon:
@vincepri are these breaking changes ok?: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-runtime/2159/pull-controller-runtime-apidiff-master/1625461909174095872

vincepri (changes requested):

pkg/source/source.go
```diff
@@ -46,8 +47,12 @@ const (
...
 	Start(context.Context, handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
+
+	// Stop is internal and should be called only by the Controller to stop the Source. It should block until the
+	// Source has stopped.
+	Stop() error
```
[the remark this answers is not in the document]
inteon: We need the `Stop` function so we can wait for the source to stop, since the `Start` function does not block until the source completes.

pkg/manager/manager.go
```diff
@@ -341,6 +342,34 @@ type LeaderElectionRunnable interface {
 	NeedLeaderElection() bool
 }
 
+// RunnableWithCancel returns a new Runnable and a function to cancel that runnable.
+func RunnableWithCancel(r Runnable) (Runnable, context.CancelFunc) {
```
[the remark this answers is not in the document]
inteon: removed this from this PR

pkg/manager/manager.go
```diff
@@ -341,6 +342,34 @@ type LeaderElectionRunnable interface {
...
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithCancel(ctx)
+			if canceled {
+				cancel()
+			} else {
+				cancels = append(cancels, cancel)
+			}
+			mu.Unlock()
+
+			return r.Start(ctx)
```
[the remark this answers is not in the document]
inteon: removed this from this PR

pkg/source/source.go
```diff
@@ -133,99 +165,100 @@ func (cs *Channel) Start(
...
-func (is *Informer) String() string {
-	return fmt.Sprintf("informer source: %p", is.Informer)
+	groupCancel()
+	return groupWait()
 }
 
 var _ Source = Func(nil)
 
 // Func is a function that implements Source.
+// Deprecated: use manully implemented Source instead.
```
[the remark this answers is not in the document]
inteon: I updated the comment

pkg/internal/source/kind.go
```diff
@@ -26,70 +28,95 @@ type Kind struct {
 
 	// started may contain an error if one was encountered during startup. If its closed and does not
 	// contain an error, startup and syncing finished.
-	started     chan error
-	startCancel func()
+	mu                 sync.Mutex
+	stopped            bool
```
sbueringer, line 32: nit: let's please add a bit more godoc for the other fields (the godoc for started seems to be in the wrong place now)

pkg/internal/source/kind.go
```diff
@@ -104,14 +131,61 @@ func (ks *Kind) String() string {
...
 	select {
 	case err := <-ks.started:
+		if err != nil {
+			// Try to stop the source to clean up any resources
+			// and collect any errors.
+			if stopErr := ks.Stop(); stopErr != nil {
+				err = kerrors.NewAggregate([]error{err, fmt.Errorf("failed to stop source: %w", stopErr)})
+			}
+		}
 		return err
```
sbueringer, line 152: nit: can we maybe return nil here and return err at the end of the if branch? Just to make it slightly more obvious that this here is also a non-error case

pkg/internal/source/kind.go
```diff
@@ -104,14 +131,61 @@ func (ks *Kind) String() string {
...
 		}
 		return errors.New("timed out waiting for cache to be synced")
 	}
 }
+
+// Stop is internal and should be called only by the Controller to stop the Source. It should block until the
+// Source has stopped.
+func (ks *Kind) Stop() error {
+	// First check that the started channel is populated
+	// by holding the lock.
```
sbueringer, line 165: nit: comment seems slightly off. We hold the lock while checking but we don't really check by holding? (I guess)

pkg/source/source.go
```diff
@@ -68,127 +77,184 @@ var _ Source = &Channel{}
...
-func (cs *Channel) Start(
-	ctx context.Context,
-	handler handler.EventHandler,
-	queue workqueue.RateLimitingInterface,
-	prct ...predicate.Predicate) error {
+// Start is internal and should be called only by the Controller to register an EventHandler with the Informer
+// to enqueue reconcile.Requests. It should NOT block, instead, it should start a goroutine and return immediately.
+// The context passed to Start can be used to cancel the blocking operations in the Start method. To cancel the
+// goroutine/ shutdown the source a user should call Stop.
+func (cs *Channel) Start(ctx context.Context, handler handler.EventHandler, queue workqueue.RateLimitingInterface,
```
sbueringer, line 106: Nit: can we document for both sources how often / when start/stop can be called?

If I see correctly the difference is mainly that for source.Channel Start can be called multiple times to add multiple event handlers and on source.Kind Start can be called only once (and thus Kind allows only adding one event handler?)

pkg/internal/source/kind.go
```diff
@@ -26,70 +28,95 @@ type Kind struct {
...
 	}
 
-	// cache.GetInformer will block until its context is cancelled if the cache was already started and it can not
-	// sync that informer (most commonly due to RBAC issues).
-	ctx, ks.startCancel = context.WithCancel(ctx)
-	ks.started = make(chan error)
+	ks.mu.Lock()
+	defer ks.mu.Unlock()
+	if ks.stopped {
+		return nil
```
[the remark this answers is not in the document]
inteon: I think we should allow Start to be called after Stop, eg. when we shutdown the controller before Start was able to obtain the mutex lock.

pkg/internal/source/kind.go
```diff
@@ -26,70 +28,95 @@ type Kind struct {
...
+	)
+
+	// Tries to get an informer until it returns true,
+	// an error or the specified context is cancelled or expired.
+	if err := wait.PollImmediateUntilWithContext(ctx, 10*time.Second, func(pollCtx context.Context) (bool, error) {
+		// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
+		i, lastErr = ks.Cache.GetInformer(pollCtx, ks.Type)
+		if lastErr != nil {
+			kindMatchErr := &meta.NoKindMatchError{}
+			switch {
```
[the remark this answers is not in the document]
inteon: I just copied this logic from the original code. However, I think this is actually not true. We should probably only retry in the first case (NoKindMatchError).

pkg/internal/source/kind.go
```diff
@@ -104,14 +131,61 @@ func (ks *Kind) String() string {
...
+// Stop is internal and should be called only by the Controller to stop the Source. It should block until the
+// Source has stopped.
+func (ks *Kind) Stop() error {
+	// First check that the started channel is populated
+	// by holding the lock.
+	if func() bool {
+		ks.mu.Lock()
+		defer ks.mu.Unlock()
+
+		ks.stopped = true
```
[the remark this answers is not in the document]
inteon: I renamed stopped to canceled, which better explains the meaning of this boolean.

pkg/internal/source/kind.go
```diff
@@ -104,14 +131,61 @@ func (ks *Kind) String() string {
...
+
+	// Now that we know the started channel is populated,
+	// wait for started to be closed, and then cancel the
+	// context and unregister the handler.
+	var errs []error
+	if err := <-ks.started; err != nil {
+		errs = append(errs, err)
+	}
+	ks.cancel()
+	if ks.removeEventHandler != nil {
```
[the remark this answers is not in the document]
inteon: It seems to me that this is not safe indeed. I liked it better when I only had to solve this once because Kind and Informer were sharing the same code.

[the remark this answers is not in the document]
inteon: I'll just fix this bug here for now.

pkg/source/source.go
```diff
@@ -68,127 +77,184 @@ var _ Source = &Channel{}
...
-		// Distribute GenericEvents to all EventHandler / Queue pairs Watching this source
-		go cs.syncLoop(ctx)
-	})
+	// Create a destination channel for the event handler
+	// and add it to the list of destinations
+	destination := make(chan event.GenericEvent, cs.DestBufferSize)
+	cs.destinations = append(cs.destinations, destination)
+
+	// Initialize the group and event loop if this is the first time.
+	if cs.group == nil {
```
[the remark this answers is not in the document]
inteon: We already have a lock, so Once is probably not necessary?

vincepri:
What's the status of this PR?

vincepri:
@alvaroaleman @sbueringer if you can review

sbueringer:
Realistically not before next week

pkg/source/source.go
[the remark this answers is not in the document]
sbueringer: Yup should work, just thought it might be easier to read

pkg/source/source_integration_test.go
```diff
@@ -66,15 +60,19 @@ var _ = Describe("Source", func() {
...
-		close(c1)
-		close(c2)
 	})
 
 	Describe("Kind", func() {
 		Context("for a Deployment resource", func() {
 			obj = &appsv1.Deployment{}
 
 			It("should provide Deployment Events", func() {
+				q := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "test")
```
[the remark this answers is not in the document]
sbueringer: @inteon @vincepri _ping_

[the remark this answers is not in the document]
inteon: I reverted the change, it did not seem to be necessary anymore.

sbueringer:
@inteon Are you currently working on further changes? (just asking because I was about to review the PR)

pkg/internal/source/kind.go
```diff
@@ -24,72 +27,142 @@ type Kind struct {
...
+				// This is a programming error, so we should stop retrying and return the error.
+				return true, fmt.Errorf("kind must be registered to the Scheme: %w", lastErr)
+			default:
+				// We got an error that is not a NoKindMatchError or IsNotRegisteredError, so we should
+				// stop retrying and return the error.
+				return true, fmt.Errorf("failed to get informer from cache: %w", lastErr)
 			}
-			ks.started <- err
-			return
+			return false, nil // Retry.
```
sbueringer, line 113: nit: I would suggest to move this line after l.103. Just slightly more readable vs the fallthrough behavior

pkg/internal/source/kind.go
```diff
@@ -104,14 +177,69 @@ func (ks *Kind) String() string {
...
+	ks.mu.Lock()
+
+	// Remove all event handlers.
+	if ks.registration != nil {
+		if err := ks.informer.RemoveEventHandler(ks.registration); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
```
sbueringer, line 244:
```suggestion
	// Remove event handler.
	if ks.registration != nil {
		if err := ks.informer.RemoveEventHandler(ks.registration); err != nil {
			return err
		}
	}

	return nil
```

pkg/source/source.go
```diff
@@ -68,164 +77,249 @@ var _ Source = &Channel{}
...
 	return fmt.Sprintf("channel source: %p", cs)
 }
 
-// Start implements Source and should only be called by the Controller.
-func (cs *Channel) Start(
-	ctx context.Context,
-	handler handler.EventHandler,
-	queue workqueue.RateLimitingInterface,
-	prct ...predicate.Predicate) error {
+// Start is internal and should be called only by the Controller to register an EventHandler with the Informer
```
sbueringer, line 103: nit: Informer doesn't seem to make sense here (similar in the interface)

pkg/source/source.go
```diff
@@ -68,164 +77,249 @@ var _ Source = &Channel{}
...
 		return fmt.Errorf("must specify Channel.Source")
 	}
 
-	// set the stop channel to be the context.
-	cs.stop = ctx.Done()
-
 	// use default value if DestBufferSize not specified
 	if cs.DestBufferSize == 0 {
 		cs.DestBufferSize = defaultBufferSize
 	}
```
sbueringer, line 118: Should we move this inside the mutex?

Not sure if it's important but seems to be the same as https://github.com/kubernetes-sigs/controller-runtime/pull/2286

pkg/source/source.go
```diff
@@ -68,164 +77,249 @@ var _ Source = &Channel{}
...
-			for _, p := range prct {
-				if !p.Generic(evt) {
-					shouldHandle = false
-					break
+	// Start a goroutine to handle events for the event handler.
+	cs.group.Go(func() error {
+	eventloop:
+		for {
+			select {
+			case <-ctx.Done():
```
sbueringer, line 150:
```suggestion
			case <-cs.ctx.Done():
```
I think we should use cs.ctx here just like in the main event loop

pkg/source/source_test.go
```diff
@@ -226,6 +226,27 @@ var _ = Describe("Source", func() {
 			Expect(err.Error()).To(Equal("cache did not sync"))
 
 		})
```
sbueringer, line 228: Looks like we have 0 (also before this PR) tests for source.Informer.

If it's relatively simple can we have those start/stop tests for the informer source as well?

sbueringer (commented):
Can you please update the PR description, I think it's outdated.

pkg/internal/source/kind.go
inteon: The comment is indeed wrong, the `errs = append(errs, err)` however is used started above this section.

inteon:
@sbueringer should be all resolved.

pkg/internal/source/kind.go
sbueringer: Ups, sounds fine 😀

sbueringer:
Thx, will take another look probably tomorrow. Overall looks good to me just want to check the locking again

pkg/controller/controllertest/util.go
```diff
@@ -135,8 +147,25 @@ func (f *FakeInformer) AddEventHandlerWithResyncPeriod(handler cache.ResourceEve
 	return nil, nil
 }
 
-// RemoveEventHandler does nothing.  TODO(community): Implement this.
+// RemoveEventHandler does nothing.
```
sbueringer, line 150:
```suggestion
// RemoveEventHandler removes an EventHandler to the fake Informers.
```
nit: sorry missed it the last time

pkg/source/source_integration_test.go
```diff
@@ -185,6 +185,126 @@ var _ = Describe("Source", func() {
...
+				evt = <-c2
+				createEvt, ok = evt.(event.CreateEvent)
+				Expect(ok).To(BeTrue(), fmt.Sprintf("expect %T to be %T", evt, event.CreateEvent{}))
+				Expect(createEvt.Object).To(Equal(created))
+
+				By("Stop the second kind source")
+				err = instance2.Stop()
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Updating a Deployment and expecting the UpdateEvent.")
```
sbueringer, line 267:
```suggestion
				By("Updating a Deployment and expecting the UpdateEvent only via first kind source.")
```
nit

pkg/source/source_integration_test.go
```diff
@@ -185,6 +185,126 @@ var _ = Describe("Source", func() {
...
+				case <-c2:
+					Fail("kind2 is expected to be stopped")
+				case <-waitCtx1.Done():
+				}
+
+				By("Stop the first kind source")
+				err = instance1.Stop()
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Deleting a Deployment and expecting the Delete.")
```
sbueringer, line 295:
```suggestion
				By("Deleting a Deployment and expecting no DeleteEvent via the first kind source.")
```
nit

pkg/internal/source/kind.go
```diff
@@ -24,72 +27,142 @@ type Kind struct {
...
-		close(ks.started)
+		ks.registration = registration
+
+		return nil
 	}()
+	if err != nil {
+		return err
+	}
+
+	// Wait for the cache to sync before adding the EventHandlers.
```
sbueringer, line 158:
```suggestion
	// Wait for the cache to sync.
```
nit: we already add the event handler before

pkg/internal/source/kind.go
```diff
@@ -24,72 +27,142 @@ type Kind struct {
...
+func getInformer(ctx context.Context, informerCache cache.Cache, resourceType client.Object) (cache.Informer, error) {
+	var (
+		i       cache.Informer
+		lastErr error
+	)
+
+	// Tries to get an informer until it returns true,
+	// an error or the specified context is cancelled or expired.
+	if err := wait.PollImmediateUntilWithContext(ctx, 10*time.Second, func(pollCtx context.Context) (bool, error) {
+		// Lookup the Informer from the Cache and add an EventHandler which populates the Queue
```
sbueringer, line 95:
```suggestion
		// Lookup the Informer from the Cache.
```
nit

pkg/internal/source/kind.go
```diff
@@ -104,14 +177,69 @@ func (ks *Kind) String() string {
...
+			return err
+		}
+
+		return nil
 	case <-ctx.Done():
-		ks.startCancel()
 		if errors.Is(ctx.Err(), context.Canceled) {
 			return nil
 		}
 		return errors.New("timed out waiting for cache to be synced")
```
sbueringer, line 206: I think this error is misleading, cache sync is just the last thing that occurs, but we're waiting for more

pkg/internal/source/kind.go
```diff
@@ -104,14 +177,69 @@ func (ks *Kind) String() string {
...
 	}
 }
+
+// Stop is internal and should be called only by the Controller to stop the Source. It should block until the
+// Source has stopped.
+// Stop can be called multiple times, and should be idempotent. If the Source is not running, Stop will prevent
+// the Source from starting in the future without error.
```
sbueringer, line 213:
```suggestion
// Stop can be called multiple times, and should be idempotent. If the Source is not running, Stop will prevent
// the Source from starting in the future.
```

nit: otherwise it's misleading as it will just prevent the source from starting but not from "starting without error"

pkg/source/source.go
```diff
@@ -68,164 +77,249 @@ var _ Source = &Channel{}
...
-			for _, p := range prct {
-				if !p.Generic(evt) {
-					shouldHandle = false
-					break
+	// Start a goroutine to handle events for the event handler.
+	cs.group.Go(func() error {
+	eventloop:
+		for {
+			select {
+			case <-cs.ctx.Done():
```
sbueringer, line 150: (I know I just had a comment to change it to this yesterday)

I'm a bit confused about the cancel semantics we want to have. With the current code only the ctx of the first Start() call is actually used. So if Start() is called a second time canceling that context won't do anything.

Let's figure out what we want and then modify the code accordingly.

Following example:
.Start(ctx1, ...)
.Start(ctx2, ...)

What should happen if ctx1 or ctx2 is canceled? Is the idea that each of them cancels the entire source? Or is the idea that they only cancel their own go routine?

I assume the later makes more sense.

In this case we would go back to using `ctx.Done()` here, but I gues in addition we have to check for `cs.ctx.Done()` (aka if the group is canceled)?

But when we do this we have to ensure that canceling ctx1 doesn't cancel the entire group.

I think we have a similar problem when using cs.ctx.Done() in other places, e.g. in `loop()`.

Hope this makes sense :)

sbueringer: cc @alvaroaleman @vincepri

pkg/source/source.go
```diff
@@ -68,164 +77,249 @@ var _ Source = &Channel{}
...
 	}
 
-	_, err := is.Informer.AddEventHandler(internal.NewEventHandler(ctx, queue, handler, prct).HandlerFuncs())
+	is.mu.Lock()
+	defer is.mu.Unlock()
+	if is.canceled {
+		return nil
+	}
+	if is.ctx == nil {
+		is.ctx, is.cancel = context.WithCancel(ctx)
```
sbueringer, line 267: Same problem as above with Channel, only the ctx of the first Start() call is used

vincepri:
/hold

For now until we clarify if we want to support `Stop()` method or other means to cancel the watches. After testing the PR locally, it looks like there are still a few tests failing, and a deadlock in the Channel source during stop

pkg/source/source.go
inteon: I think the solution is to only allow each source to be started once.
For Kind and Informer sources this can be added without removing functionality for c/r users. For the Channel source however, we need to add some logic that allows us to create multiple sources from the same channel while receiving each event in each source.

Additionally, I think the ctx passed to Start should be the only way to shutdown the source. A Shutdown function can be added to the source, but this function will just wait for the source to shutdown and will signal  that new calls to Start should not do anything (this is based in the Shutdown function in the DynamicSharedInformerFactory https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/staging/src/k8s.io/client-go/dynamic/dynamicinformer/interface.go#L38-L48)

inteon:
@vincepri sorry for changing the PR again. However, I think I found solutions to the remaining questions:
1. I realised that starting the same Source object multiple times is very messy (eg. for determining what context does what) and can easily cause a lot of issues. Instead I changed all sources so they can only be started once. For Kind and Informer, you can easily construct a new instance if you want two sources that listen to the same kind/ informer. For the Channel Source, it is not easy to create two sources that read from the same channel while making sure each source receives all events on that channel. To fix this issue, I created a Broadcaster wrapper for the channel that has to be passed to the Channel source, allowing you to construct multiple Channel sources with the same channel.

2. I realised that, like you said before, having two ways to stop a Source is non-ideal. Instead of having the Stop function cancel the source, I replaced the function with a Shutdown function which only waits for the source to shutdown (based on what exists for DynamicSharedInformerFactory https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/staging/src/k8s.io/client-go/dynamic/dynamicinformer/interface.go#L38-L48)

rnsv:
@inteon After calling ctrlruntime.NewControllerManagedBy(...).Complete(reconciler), how would I initiate a Shutdown()

inteon:
xref: https://github.com/kubernetes-sigs/controller-runtime/issues/2176

inteon:
@lou-lan I'm not sure what is blocking this PR.
I resolved the outstanding issues and challenges.
Looking for a final review from @vincepri @sbueringer or @FillZpp

alvaroaleman:
Sorry, I don't understand this change. Contrary to what the title claims, it never stops any watch, it stops the source. How does that help? Please provide a real-world use-case.

If this is supposed to work towards allowing to stop watches by refcounting if they have eventhandlers or are otherwise used, I'd like to see a description somewhere as to how exactly this will work once all pieces are done.

inteon:
> Sorry, I don't understand this change. Contrary to what the title claims, it never stops any watch, it stops the source. How does that help? Please provide a real-world use-case.
> 
> If this is supposed to work towards allowing to stop watches by refcounting if they have eventhandlers or are otherwise used, I'd like to see a description somewhere as to how exactly this will work once all pieces are done.

@alvaroaleman I just added a `StopWatch` function to expose this functionality via the Controller API (see https://github.com/kubernetes-sigs/controller-runtime/pull/2159/commits/6c16925df7905f625d3cb54803b12e432e76fc2a).
This integration test should be a good example on how this PR can be used:
https://github.com/kubernetes-sigs/controller-runtime/pull/2159/commits/6c16925df7905f625d3cb54803b12e432e76fc2a#diff-c1c23edcf40f177710028c953468fd5bf7aa0f32d67780354c609e127603bb04

inteon:
I'm discontinuing this effort. Feel free to pick this up and continue this work in a new PR.

PR #2161, opened 2023-01-31
🌱 Return the error from rest.HTTPClientFor instead of hiding it

inteon:
Follow-up PR for https://github.com/kubernetes-sigs/controller-runtime/pull/2122 (see https://github.com/kubernetes-sigs/controller-runtime/pull/2122#discussion_r1092334261)


sbueringer:
/retitle 🌱 Return the error from rest.HTTPClientFor instead of hiding it

(I think we don't have to flag it as breaking)

alvaroaleman (approved):

PR #2164, opened 2023-02-01
🌱Return an error if no httpClient is provided

inteon:
As discussed here: https://github.com/kubernetes-sigs/controller-runtime/pull/2122#discussion_r1092736800, these `NewDiscoveryRESTMapper` and `RESTClientForGVK` are actually public functions.
For this reason, we want to return an error instead of panicking.

TODO: decide whether we want to just create a default httpClient if the parameter is nil

@alvaroaleman PTAL

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -64,7 +64,7 @@ func AddToProtobufScheme(addToScheme func(*runtime.Scheme) error) error {
 // information fetched by a new client with the given config.
 func NewDiscoveryRESTMapper(c *rest.Config, httpClient *http.Client) (meta.RESTMapper, error) {
 	if httpClient == nil {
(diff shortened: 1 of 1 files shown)


pkg/client/apiutil/apimachinery.go
```diff
@@ -64,7 +64,7 @@ func AddToProtobufScheme(addToScheme func(*runtime.Scheme) error) error {
 // information fetched by a new client with the given config.
 func NewDiscoveryRESTMapper(c *rest.Config, httpClient *http.Client) (meta.RESTMapper, error) {
 	if httpClient == nil {
-		panic("httpClient must not be nil")
+		return nil, fmt.Errorf("httpClient must not be nil")
```
alvaroaleman, line 67: Maybe include the info into the error how ppl can construct one from a rest config?

vincepri (approved):

PR #2165, opened 2023-02-01
🌱 [0.14] Deprecate component configuration package

vincepri:
This provides an early warning for users on 0.14 that the pkg/config types are going to be removed in a future release and they should migrate off.

/assign @sbueringer

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -21,7 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	cfg "sigs.k8s.io/controller-runtime/pkg/config"
+	cfg "sigs.k8s.io/controller-runtime/pkg/config" //nolint:staticcheck
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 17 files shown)


sbueringer:
/lgtm

nit: I wonder if we should go with the standard cherry-pick bot PR titles: `[release-0.14] 🌱 Deprecate component configuration package`

qbarrand:
Hello! I am looking for more context or documentation about this change. Both [examples](https://github.com/kubernetes-sigs/controller-runtime/tree/master/examples/configfile) in this repo are still using the deprecated methods.
What should we migrate to?

tyxeron:
> Hello! I am looking for more context or documentation about this change. Both [examples](https://github.com/kubernetes-sigs/controller-runtime/tree/master/examples/configfile) in this repo are still using the deprecated methods. What should we migrate to?

I am also wondering what the migration path looks like, since our static checker does not really like the deprecation warnings.

qbarrand:
@vincepri @sbueringer could you please help us here?

errordeveloper:
Per commit message in 63e5fef661f3dbc74f779101339a3ce1cda642a0:

> This provides an early warning for users on 0.14 that the pkg/config
> types are going to be removed in a future release and they should
> migrate off.

What is missing is any info as to what package users should be looking to migrate to?

vincepri:
👋 

The codebase around configuration package is missing maintainers and the general integration between the configuration types and the manager options or controllers can be very confusing to use. 

There is currently no alternative implementation, folks could use their own json/yaml unmarshalling and configure manager options that way.

Happy to revisit if we have folks stepping up to review, cleanup, support, and maintain the codebase.

PR #2166, opened 2023-02-01
⚠ Cache's ByObject is now a map Object -> options

vincepri:
This is a breaking change and revisits some of the previous changes we've made. It brings the options ByObject to a top level map, which now contains all the options the internal informer can be configured with. In addition, it removes the ObjectAll object, which was super confusing and only used to disable the deep copy for every object; instead it's now possible to disable deep copy through a top level option.


pkg/cache/cache.go
```diff
@@ -147,31 +134,36 @@ type ViewOptions struct {
...
 	// and for updated objects.
-	Transform TransformByObject
+	Transform toolscache.TransformFunc
 
 	// UnsafeDisableDeepCopy indicates not to deep copy objects during get or
 	// list objects per GVK at the specified object.
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
-	UnsafeDisableDeepCopy DisableDeepCopyByObject
+	UnsafeDisableDeepCopy bool
```
varshaprasad96, line 166: @vincepri just curious, is it intentional to not overwrite `UnsafeDisableDeepCopy` by the top level `opts. UnsafeDisableDeepCopy` if it's not specified per object? (I understand its safe to default it to false, but just asking if it would be confusing to the user). Also is there a reason to have it as a pointer in top level and not in byObject?

vincepri: The thinking behind this was to retain some of the features we have today, and make it a bit safer as you said. 

- If a user wants to disable deep copy globally, we should be able to override it for specific objects.
- If users want to keep it to false globally (default), we should be able to override it to true for specific objects.

The reason for the top level pointer is for merging options together, there is a need to know if the option was left unset, or was false, which is the default of a boolean field. The option inheritance logic is hard to follow (spent a good deal of time on it today 😄), but it basically requires the pointer to know if we should inherit or not.

Happy to chat about it more if we feel like this should be made safer or easier to use. I would expect though that most users shouldn't disable deep copy unless they absolutely know the repercussions of doing so, which is an advanced use case

pkg/cache/cache.go
```diff
@@ -147,31 +134,36 @@ type ViewOptions struct {
 	DefaultTransform toolscache.TransformFunc
 
 	// ByObject restricts the cache's ListWatch to the desired fields per GVK at the specified object.
-	ByObject ViewByObject
+	ByObject map[client.Object]ByObject
+
+	// UnsafeDisableDeepCopy indicates not to deep copy objects during get or
+	// list objects for EVERY object.
+	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
+	// otherwise you will mutate the object in the cache.
```
alvaroaleman, line 142: Nit: mention that this can be overrriden through `ByObject` per GVK?

alvaroaleman (commented):
On

pkg/cache/cache.go
varshaprasad96: Ah, I see, got it! Thanks for explaining @vincepri :)

pkg/cache/cache.go
```diff
@@ -147,31 +134,36 @@ type ViewOptions struct {
...
-	// Selectors restricts the cache's ListWatch to the desired
-	// fields per GVK at the specified object, the map's value must implement
-	// Selectors [1] using for example a Set [2]
-	// [1] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selectors
-	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
-	Selectors SelectorsByObject
+// ByObject offers more fine-grained control over the cache's ListWatch by object.
+type ByObject struct {
+	// Label represents a label selector for the object.
+	Label labels.Selector
```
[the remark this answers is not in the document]
vincepri: This was a suggestion by @alvaroaleman, I'm happy to do either way, although to reduce verbosity of the ByObject struct, might make sense to keep it this way

pkg/cache/cache.go
```diff
@@ -147,31 +134,36 @@ type ViewOptions struct {
...
-// ViewByObject offers more fine-grained control over the cache's ListWatch by object.
-type ViewByObject struct {
-	// Selectors restricts the cache's ListWatch to the desired
-	// fields per GVK at the specified object, the map's value must implement
-	// Selectors [1] using for example a Set [2]
-	// [1] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selectors
-	// [2] https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Set
-	Selectors SelectorsByObject
+// ByObject offers more fine-grained control over the cache's ListWatch by object.
+type ByObject struct {
```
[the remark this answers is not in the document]
vincepri: Can rename if we want to, the Struct is only used within the options, and was trying to reduce overall verbosity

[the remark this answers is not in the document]
alvaroaleman: Maybe `ObjectOptions`? But no strong opining either way

pkg/cache/cache.go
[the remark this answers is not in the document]
alvaroaleman: yeah IMHO this is easier to use when keeping them separate. I guess we could also do the same for the `DefaultSelector`

pkg/cache/cache.go
vincepri: fixed, thanks

pkg/cache/cache.go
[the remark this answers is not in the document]
vincepri: As in, separate `DefaultLabelSelector` and `DefaultFieldSelector`?

pkg/cache/cache.go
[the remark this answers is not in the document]
vincepri:
```go
cache.Options{
  ByObject: map[client.Object]cache.ByObject{
    &corev1.Pod{}: {}
  }
}
```

IMO, this reads well and looks nice. @FillZpp if you feel that we should change it, happy to do it though

pkg/cache/cache_unit_test.go
```diff
@@ -110,120 +109,131 @@ var _ = Describe("cache.inheritFrom", func() {
...
+			specified.ByObject = map[client.Object]ByObject{
+				&corev1.Pod{}: {},
+			}
+			Expect(checkError(specified.inheritFrom(inherited)).ByObject).To(HaveLen(1))
 		})
 		It("is inherited when only inherited is set", func() {
 			inherited.Scheme = coreScheme
-			inherited.View.ByObject.Selectors = map[client.Object]ObjectSelector{&corev1.ConfigMap{}: {}}
-			Expect(checkError(specified.inheritFrom(inherited)).View.ByObject.Selectors).To(HaveLen(1))
+			specified.ByObject = map[client.Object]ByObject{
```
sbueringer, line 142: should be inherited

pkg/cache/cache_unit_test.go
```diff
@@ -110,120 +109,131 @@ var _ = Describe("cache.inheritFrom", func() {
...
+			inherited.ByObject = map[client.Object]ByObject{
+				&corev1.ConfigMap{}: {},
+			}
+			Expect(checkError(specified.inheritFrom(inherited)).ByObject).To(HaveLen(1))
 		})
 		It("uses inherited scheme for specified selectors", func() {
 			inherited.Scheme = coreScheme
-			specified.View.ByObject.Selectors = map[client.Object]ObjectSelector{&corev1.ConfigMap{}: {}}
-			Expect(checkError(specified.inheritFrom(inherited)).View.ByObject.Selectors).To(HaveLen(1))
+			inherited.ByObject = map[client.Object]ByObject{
```
sbueringer, line 200: should be specified

pkg/cache/cache_unit_test.go
```diff
@@ -110,120 +109,131 @@ var _ = Describe("cache.inheritFrom", func() {
...
+				&corev1.ConfigMap{}: {},
+			}
+			Expect(checkError(specified.inheritFrom(inherited)).ByObject).To(HaveLen(1))
 		})
 		It("uses specified scheme for specified selectors", func() {
 			specified.Scheme = coreScheme
-			specified.View.ByObject.Selectors = map[client.Object]ObjectSelector{&corev1.ConfigMap{}: {}}
-			Expect(checkError(specified.inheritFrom(inherited)).View.ByObject.Selectors).To(HaveLen(1))
+			inherited.ByObject = map[client.Object]ByObject{
```
sbueringer, line 207: same

pkg/cache/cache.go
```diff
@@ -147,31 +134,38 @@ type ViewOptions struct {
...
 	// and for updated objects.
-	Transform TransformByObject
+	Transform toolscache.TransformFunc
 
 	// UnsafeDisableDeepCopy indicates not to deep copy objects during get or
 	// list objects per GVK at the specified object.
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
-	UnsafeDisableDeepCopy DisableDeepCopyByObject
+	UnsafeDisableDeepCopy bool
```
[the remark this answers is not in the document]
vincepri: I was thinking the latter, if we're overriding something I _guess_ that would be the source of truth. That said, the bool's default to false comes into play into unexpected UX. Should we convert to pointer here as well?

pkg/cache/cache.go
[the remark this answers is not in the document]
alvaroaleman: Yeah

pkg/cache/cache.go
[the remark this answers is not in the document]
sbueringer: I think pointer gives the better UX

pkg/cache/cache.go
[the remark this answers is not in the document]
vincepri: done, ptal

pkg/cache/cache_test.go
```diff
@@ -1822,12 +1810,11 @@ func isKubeService(svc metav1.Object) bool {
...
 
 func isPodDisableDeepCopy(opts cache.Options) bool {
-	if d, ok := opts.View.ByObject.UnsafeDisableDeepCopy[&corev1.Pod{}]; ok {
-		return d
-	} else if d, ok = opts.View.ByObject.UnsafeDisableDeepCopy[cache.ObjectAll{}]; ok {
-		return d
-	} else if d, ok = opts.View.ByObject.UnsafeDisableDeepCopy[&cache.ObjectAll{}]; ok {
-		return d
+	if opts.ByObject[&corev1.Pod{}].UnsafeDisableDeepCopy != nil && *opts.ByObject[&corev1.Pod{}].UnsafeDisableDeepCopy {
+		return true
```
sbueringer, line 1814: Shouldn't this return the value of UnsafeDisableDeepCopy?

(same below)

sbueringer: I.e. if it's set to false on the object shouldn't it be false?

vincepri: Changed it, there shouldn't be any logical difference, but I agree it might be more clear

vincepri: Actually I stand corrected, we want to always return the most specific value now, given that's a pointer

sbueringer: Yup, that's what I thought :)

sbueringer:
lgtm from my side

alvaroaleman (approved):

PR #2168, opened 2023-02-02
🌱 Use runtime.Unstructured interface instead of Unstructured struct

### LINKED ISSUE #415: unstructured client should key off of the unstructured interface

vincepri:

Fixes #415

I'm indifferent if we actually want to do this or not, if we decide not to, we should also close the related issue as well


FillZpp:
> I'm indifferent if we actually want to do this or not, if we decide not to, we should also close the related issue as well

I'm in favor of this. The code looks more clear and we don't have to check both Unstructured and UnstructuredList in some of the places.

alvaroaleman (approved):
Looks like a clear simplification to me and doesn't result in any public api changes or test changes

PR #2169, opened 2023-02-02
🐛 Allow to set GracefulShutdownTimeout to -1, disabling timeouts

### LINKED ISSUE #2110: Question: Handling graceful shutdown without timeout
I would like to get some clarification regarding the graceful shutdown implementation in the manager. In the [documentation](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.14.1/pkg/manager/manager.go#L293), it seems that there is support for timeout-less graceful shutdown, however, [this feature has been removed](https://github.com/kubernetes-sigs/controller-runtime/commit/612e9b28858354c69696a91b0778f3440ce4687c#diff-517abdd2f47b744e11990598de732426f33aba68600eaa9600d2ee20bc0b25dcL513-L525). It is not clear if this was removed intentionally or by mistake. If this was indeed removed intentionally then it should be removed from the documentation. Otherwise, I think it would be great if this feature could be restored as it can be helpful for situations when we want to delegate force termination to Kubernetes via `terminationGracePeriodSeconds`.

@vincepri maybe you could help clarify?

vincepri:

This is already specified in the documentation https://github.com/kubernetes-sigs/controller-runtime/blob/v0.14.1/pkg/manager/manager.go#L293 

Fixes #2110

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -509,7 +509,12 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 	//
 	// The shutdown context immediately expires if the gracefulShutdownTimeout is not set.
 	var shutdownCancel context.CancelFunc
-	cm.shutdownCtx, shutdownCancel = context.WithTimeout(context.Background(), cm.gracefulShutdownTimeout)
+	if cm.gracefulShutdownTimeout < 0 {
+		// We want to wait forever for the runnables to stop.
+		cm.shutdownCtx, shutdownCancel = context.WithCancel(context.Background())
+	} else {
+		cm.shutdownCtx, shutdownCancel = context.WithTimeout(context.Background(), cm.gracefulShutdownTimeout)
+	}
 	defer shutdownCancel()
 
 	// Start draining the errors before acquiring the lock to make sure we don't deadlock
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -936,6 +936,50 @@ var _ = Describe("manger.Manager", func() {
 				<-runnableStopped
 			})
 
+			It("should wait forever for runnables if gracefulShutdownTimeout is <0 (-1)", func() {
+				m, err := New(cfg, options)
+				Expect(err).NotTo(HaveOccurred())
+				for _, cb := range callbacks {
+					cb(m)
+				}
+				m.(*controllerManager).gracefulShutdownTimeout = time.Duration(-1)
+
+				Expect(m.Add(RunnableFunc(func(ctx context.Context) error {
+					<-ctx.Done()
+					time.Sleep(100 * time.Millisecond)
+					return nil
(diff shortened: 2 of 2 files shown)


pkg/manager/manager_test.go
```diff
@@ -936,6 +936,50 @@ var _ = Describe("manger.Manager", func() {
...
+					Expect(m.Start(ctx)).NotTo(HaveOccurred())
+					close(managerStopDone)
+				}()
+				<-m.Elected()
+				cancel()
+
+				beforeDone := time.Now()
+				<-managerStopDone
+				Expect(time.Since(beforeDone)).To(BeNumerically(">=", 1500*time.Millisecond))
+			})
```
vincepri, line 981: Is there a better way to test this? Here I'm just spinning up N goroutines, and making sure we've waited at least the biggest sleep, which should make sense 🤔  although wondering if there is a clearer way that doesn't necessarily rely on timing

relu: Here's an idea, what if it just evaluates the manager shutdownCtx? You would be able to simplify the test case as well as the runnable can be skipped.
```
...
<-managerStopDone
_, ok := m.(*controllerManager).shutdownCtx.Deadline()
Expect(ok).To(BeFalse())
```
According to the docs:
> Deadline returns the time when work done on behalf of this context should be canceled. Deadline returns ok==false when no deadline is set. Successive calls to Deadline return the same results.

pkg/manager/internal.go
```diff
@@ -509,7 +509,12 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 	//
 	// The shutdown context immediately expires if the gracefulShutdownTimeout is not set.
 	var shutdownCancel context.CancelFunc
-	cm.shutdownCtx, shutdownCancel = context.WithTimeout(context.Background(), cm.gracefulShutdownTimeout)
+	if cm.gracefulShutdownTimeout < 0 {
+		// We want to wait forever for the runnables to stop.
+		cm.shutdownCtx, shutdownCancel = context.WithCancel(context.Background())
+	} else {
+		cm.shutdownCtx, shutdownCancel = context.WithTimeout(context.Background(), cm.gracefulShutdownTimeout)
```
FillZpp, line 516: I'm not sure, is it meaningful to create a ctx with timeout 0? Or we can disable timeouts when gracefulShutdownTimeout is 0?

vincepri: No, by the documentation, `-1` disables the timeout, `0` is basically immediate, `>0` is a waiting

pkg/manager/manager_test.go
vincepri: Might be good as a follow-up?

relu: Up to you! :smile:

sbueringer:
Looks good!

/lgtm

PR #2177, opened 2023-02-03
✨ Expose Cache/Client options on Cluster

vincepri:

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 
@@ -82,6 +83,8 @@ type Options struct {
 	Scheme *runtime.Scheme
 
 	// MapperProvider provides the rest mapper used to map go types to Kubernetes APIs
+	//
+	// Deprecated: Set Cache.Mapper and Client.Mapper directly instead.
 	MapperProvider func(c *rest.Config, httpClient *http.Client) (meta.RESTMapper, error)
 
 	// Logger is the logger that should be used by this Cluster.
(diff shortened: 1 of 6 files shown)


vincepri:
/hold

for tests

vincepri:
@alvaroaleman @sbueringer thoughts on the approach?

pkg/cluster/cluster.go
```diff
@@ -171,36 +196,63 @@ func New(config *rest.Config, opts ...Option) (Cluster, error) {
...
+		}
+		if clientOpts.HTTPClient == nil {
+			clientOpts.HTTPClient = options.HTTPClient
+		}
+		if clientOpts.Cache == nil {
+			clientOpts.Cache.Reader = cache
+			clientOpts.Cache.DisableFor = options.ClientDisableCacheFor
+			clientOpts.Cache.Unstructured = false
+		}
+		if !clientOpts.DryRun && options.DryRunClient {
```
alvaroaleman, line 239: maybe make `clientOpts.DryRun` a pointer to be able to differentiate between unset and set here?

pkg/cluster/cluster.go
```diff
@@ -171,36 +196,63 @@ func New(config *rest.Config, opts ...Option) (Cluster, error) {
...
-		},
-	})
+	// Create the client, and default its options.
+	clientOpts := options.Client
+	{
+		if clientOpts.Scheme == nil {
+			clientOpts.Scheme = options.Scheme
+		}
+		if clientOpts.Mapper == nil {
+			clientOpts.Mapper = mapper
```
alvaroaleman, line 229: Nit: Maybe also deprecate the `mapperProvider` and let the client itself default the `Mapper` if not passed?

alvaroaleman (commented):
I like it a lot, much cleaner :)

pkg/cluster/cluster.go
vincepri: Good idea, done!

pkg/cluster/cluster.go
vincepri: Done!

vincepri: Both the cache and the client already default it, so we should be good there once we remove the mapperProvider

sbueringer:
Looks good to me

/assign @alvaroaleman

alvaroaleman (approved):

PR #2180, opened 2023-02-06
:seedling: code clean for pkg/envtest/server.go

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

/kind cleanup

Code clean for evntest.server. 
Because the variable ```AttachControlPlaneOutput``` is only used in the ```os.Getenv(envAttachOutput) ``` method..
Maybe we should clean the code for that condition.

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -230,17 +230,19 @@ func (te *Environment) Start() (*rest.Config, error) {
 		if os.Getenv(envAttachOutput) == "true" {
 			te.AttachControlPlaneOutput = true
 		}
-		if apiServer.Out == nil && te.AttachControlPlaneOutput {
-			apiServer.Out = os.Stdout
-		}
-		if apiServer.Err == nil && te.AttachControlPlaneOutput {
-			apiServer.Err = os.Stderr
-		}
-		if te.ControlPlane.Etcd.Out == nil && te.AttachControlPlaneOutput {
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -229,18 +229,18 @@ func (te *Environment) Start() (*rest.Config, error) {
 
 		if os.Getenv(envAttachOutput) == "true" {
 			te.AttachControlPlaneOutput = true
```
inteon, line 231: @yanggangtony `AttachControlPlaneOutput` is set to `true` if `os.Getenv(envAttachOutput) == "true"`, however it can also be true if `os.Getenv(envAttachOutput) != "true"` and the `AttachControlPlaneOutput` value was set to true.
So, please create a new `if te.AttachControlPlaneOutput {` check that contains all the other checks.

inteon (changes requested):

yanggangtony:
@inteon 
Thanks for review.
Had rebased..

inteon:
@yanggangtony can you fix the typo in this PR, replace `evntest` with `envtest`?

inteon (approved):
Super small code-style change
/approve
/lgtm

yanggangtony:
thanks for review..

FillZpp (approved):
Thanks!
/approve

PR #2096, opened 2022-12-12
🌱 Generate files and update modules

lucacome:
Ran `make generate` and `make modules`

PR #2186, opened 2023-02-12
🌱 verify.sh: verify generate & modules (in CI)

sbueringer:
Follow-up to https://github.com/kubernetes-sigs/controller-runtime/pull/2096 to ensure we always verify that go modules and generated files are up-to-date.

(verify.sh is run by our presubmit test job)

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -24,12 +24,22 @@ cd "${REPO_ROOT}"
 header_text "running generate"
 make generate
 
-header_text "running golangci-lint"
-make lint
+# Only run verify-generate in CI, otherwise running generate
+# locally (which is a valid operation) causes `make test` to fail.
+if [[ -n ${CI} ]]; then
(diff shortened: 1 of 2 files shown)


hack/verify.sh
```diff
@@ -24,12 +24,22 @@ cd "${REPO_ROOT}"
...
-make lint
+# Only run verify-generate in CI, otherwise running generate
+# locally (which is a valid operation) causes `make test` to fail.
+if [[ -n ${CI} ]]; then
+    header_text "verifying generate"
+    make verify-generate
+fi
 
-# Only run module verification in CI, otherwise updating
+header_text "running modules"
```
sbueringer, line 34: Flipped the order here so we first run go mod tidy before running the linter.

Depending on the state of the go.mod files the linter could fail because they are out-of-date.

vincepri (approved):
/lgtm

PR #2199, opened 2023-02-15
⚠ Expose Manager.Cache/Client options, deprecate older opts

vincepri:

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -100,10 +100,44 @@ type Options struct {
 	// Scheme is the scheme used to resolve runtime.Objects to GroupVersionKinds / Resources.
 	// Defaults to the kubernetes/client-go scheme.Scheme, but it's almost always better
 	// to pass your own scheme in. See the documentation in pkg/scheme for more information.
+	//
+	// If set, the Scheme will be used to create the default Client and Cache.
 	Scheme *runtime.Scheme
 
-	// MapperProvider provides the rest mapper used to map go types to Kubernetes APIs
+	// MapperProvider provides the rest mapper used to map go types to Kubernetes APIs.
+	//
+	// If set, the RESTMapper returned by this function is used to create the RESTMapper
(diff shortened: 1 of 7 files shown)


pkg/manager/manager.go
```diff
@@ -130,6 +164,8 @@ type Options struct {
 	// is "done" with an object, and would otherwise not requeue it, i.e., we
 	// recommend the `Reconcile` function return `reconcile.Result{RequeueAfter: t}`,
 	// instead of `reconcile.Result{}`.
+	//
+	// Deprecated: Use Cache.SyncPeriod instead.
```
sbueringer, line 168: Should we replicate the godoc of this field over at the Cache.SyncPeriod?

pkg/manager/manager.go
```diff
@@ -488,16 +525,27 @@ func (o Options) AndFrom(loader config.ControllerManagerConfiguration) (Options,
...
 	}
 
-	if o.Port == 0 && newObj.Webhook.Port != nil {
-		o.Port = *newObj.Webhook.Port
+	webhookServer := &webhook.Server{}
+	if newObj.Webhook.Port != nil {
+		if o.Port == 0 {
+			o.Port = *newObj.Webhook.Port
+		}
+		webhookServer.Port = *newObj.Webhook.Port
```
sbueringer, line 533: I'm probably missing something. Shouldn't webhookServer.Port be o.Port if o.Port is != 0?

sbueringer: Hm or is the idea that .Webhook.Port has the higher priority? It's just a bit confusing as in all the other places above we only set anything on o Options if the field there is not set

vincepri: `webhookServer.Port` is set regardless because it's really only set `if options.WebhookServer == nil {`

vincepri: And yeah this inheritance logic is very confusing to say the least, hence the deprecation :D

PR #2200, opened 2023-02-15
🌱 Provide more information on deprecating ComponentConfig

vincepri:


FillZpp:
/lgtm
/hold
for other reviews

When do we plan to remove the ControllerManagerConfiguration?

vincepri:
@FillZpp We could aim for N+2 versions after the deprecation goes out?

FillZpp:
> @FillZpp We could aim for N+2 versions after the deprecation goes out?

Alright, sounds good!
/hold cancel

PR #2208, opened 2023-02-22
🐛 Allow lazy restmapper to work with CRDs created at runtime

Fedosin:
Now lazy restmapper fetches all API resources once at start and then caches them. It prevents it from discovery of new CRDs created after the controller has started.
This PR allows lazy restmapper to work with such CRDs.

### DIFF
--- a/pkg/client/apiutil/lazyrestmapper.go
+++ b/pkg/client/apiutil/lazyrestmapper.go
@@ -33,7 +33,7 @@ type lazyRESTMapper struct {
 	mapper      meta.RESTMapper
 	client      *discovery.DiscoveryClient
 	knownGroups map[string]*restmapper.APIGroupResources
-	apiGroups   *metav1.APIGroupList
+	apiGroups   []metav1.APIGroup
 
 	// mutex to provide thread-safe mapper reloading.
 	mu sync.Mutex
@@ -45,6 +45,7 @@ func newLazyRESTMapperWithClient(discoveryClient *discovery.DiscoveryClient) (me
 		mapper:      restmapper.NewDiscoveryRESTMapper([]*restmapper.APIGroupResources{}),
 		client:      discoveryClient,
 		knownGroups: map[string]*restmapper.APIGroupResources{},
+		apiGroups:   []metav1.APIGroup{},
 	}, nil
 }
 
@@ -147,7 +148,7 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
 	// This operation requires 2 requests: /api and /apis, but only once. For all subsequent calls
 	// this data will be taken from cache.
 	if len(versions) == 0 {
-		apiGroup, err := m.findAPIGroupByName(groupName)
+		apiGroup, err := m.findAPIGroupByNameLocked(groupName)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/client/apiutil/lazyrestmapper_test.go
```diff
@@ -401,4 +403,81 @@ func TestLazyRestMapperProvider(t *testing.T) {
 		g.Expect(err).To(gmg.HaveOccurred())
 		g.Expect(crt.GetRequestCount()).To(gmg.Equal(6))
 	})
+
+	t.Run("LazyRESTMapper can fetch CRDs if they were created in runtime", func(t *testing.T) {
```
alvaroaleman, line 407:
```suggestion
	t.Run("LazyRESTMapper can fetch CRDs if they were created at runtime", func(t *testing.T) {
```

The title actually confused me because I thought you mean't "Created in controller-runtime" :P

pkg/client/apiutil/lazyrestmapper.go
```diff
@@ -200,22 +210,18 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
...
-		if err != nil {
-			return metav1.APIGroup{}, fmt.Errorf("failed to get server groups: %w", err)
-		}
-		if len(apiGroups.Groups) == 0 {
-			return metav1.APIGroup{}, fmt.Errorf("received an empty API groups list")
-		}
-
-		m.apiGroups = apiGroups
+	// It will make 2 API calls to /api and /apis respectively.
+	apiGroups, err := m.client.ServerGroups()
```
alvaroaleman, line 214: Shouldn't we cache this and only re-fetch if the cached serverGroups don't have the requested groupName?

Fedosin: Yeah, I returned caching here again. Now we just update it when nothing was found.

pkg/client/apiutil/lazyrestmapper.go
```diff
@@ -199,26 +211,32 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
...
-	if m.apiGroups == nil {
-		apiGroups, err := m.client.ServerGroups()
-		if err != nil {
-			return metav1.APIGroup{}, fmt.Errorf("failed to get server groups: %w", err)
-		}
-		if len(apiGroups.Groups) == 0 {
-			return metav1.APIGroup{}, fmt.Errorf("received an empty API groups list")
+	// Looking in the cache first.
+	for i := range m.apiGroups {
+		if groupName == (&m.apiGroups[i]).Name {
```
vincepri, line 216: Could we assign `m.apiGroups[i]` to a local variable?

pkg/client/apiutil/lazyrestmapper.go
```diff
@@ -199,26 +211,32 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
 
 // findAPIGroupByName returns API group by its name.
 func (m *lazyRESTMapper) findAPIGroupByName(groupName string) (metav1.APIGroup, error) {
```
vincepri, line 213: Could we rename this to add `findAPIGroupByName_locked` or similar to make sure callers always hold the lock?

pkg/client/apiutil/lazyrestmapper.go
Fedosin: Yeah, totally makes sense. Thanks for the suggestion!

sbueringer:
Thx for the fix!

/approve

PR #2212, opened 2023-02-27
:bug: fix: set DeleteStateUnknown when obj is DeletedFinalStateUnknown

zwtop:
As the document says, DeleteStateUnknown should be true if the object was deleted but delete event was missed.

https://github.com/kubernetes-sigs/controller-runtime/blob/b6fc422accf925890b9bd8832f99ecd0eba40c6c/pkg/event/event.go#L44-L46


FillZpp:
/lgtm
Good catch!

PR #2216, opened 2023-02-28
:sparkles: Guess GVK for unstructured list items not in scheme

### LINKED ISSUE #2171: Unstructured client's `List` does not propagate GVK to items
The `List` method does prepare the `gvk` to an "unsafe guess that doen't actually work, but matches ~99% of the cases out there", but this is never actually being put to use: https://github.com/kubernetes-sigs/controller-runtime/blob/10a02729220c108cb112962b9f227229c708c052/pkg/client/unstructured_client.go#L207-L208

Which does seem to be a left-over from changes in #812, which also mentions:

> [...] feels like I probably need to fix up some GVKs somewhere in the lists...

However, given that for `metav1.PartialObjectMetadataList` the GVK is later restored in the wrapping client: https://github.com/kubernetes-sigs/controller-runtime/blob/4fd4f6e7116453ea4e07a2cdd8ba0d38ad64680f/pkg/client/client.go#L372-L383

I am wondering if the same approach could be taken for the `unstructured.UnstructuredList`, as this would ease consumption in ~99% of the cases. If agreed on, I am happy to submit a PR.

hiddeco:
Fixes #2171

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -363,7 +363,31 @@ func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) e
 
 	switch x := obj.(type) {
 	case runtime.Unstructured:
-		return c.unstructuredClient.List(ctx, obj, opts...)
+		// Call the list client.
+		if err := c.unstructuredClient.List(ctx, obj, opts...); err != nil {
+			return err
+		}
+
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		if !c.scheme.Recognizes(gvk) {
+			// If the scheme doesn't recognize the GVK, it will not be set on
+			// the list items. We can however produce an unsafe guess which
+			// will work for most cases, and eases consumption.
+			itemGVK := schema.GroupVersionKind{
+				Group:   gvk.Group,
+				Version: gvk.Version,
+				// TODO: this is producing unsafe guesses that don't actually work,
+				// but it matches ~99% of the cases out there.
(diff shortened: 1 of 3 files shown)


hiddeco:
Interesting enough 89c0643 now seems to pass without fixes compared to original state before I rebased against `main`, which does not seem to happen for https://github.com/hiddeco/controller-runtime/commit/3391a83e3d1af62a5027976bdf6888aad185ea2f against `release-0.14`.

pkg/client/client.go
```diff
@@ -363,7 +363,31 @@ func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) e
...
+		if err := c.unstructuredClient.List(ctx, obj, opts...); err != nil {
+			return err
+		}
+
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		if !c.scheme.Recognizes(gvk) {
+			// If the scheme doesn't recognize the GVK, it will not be set on
+			// the list items. We can however produce an unsafe guess which
+			// will work for most cases, and eases consumption.
+			itemGVK := schema.GroupVersionKind{
```
alvaroaleman, line 376: Are you sure that this is set in the structured case? I thought this is never the case due to https://github.com/kubernetes/kubernetes/issues/80609

FillZpp:
> Guess GVK for unstructured list items not in scheme

I remember the return items of unstructured list should already have GVK, no matter the type is in scheme or not.

hiddeco:
Closing this until I have more time to get to the bottom of it, thanks both.

PR #2217, opened 2023-03-01
✨ Add rest client request duration and size metrics

sbueringer:

This PR adds the following metrics:
* rest_client_request_duration_seconds
* rest_client_request_size_bytes
* rest_client_response_size_bytes

As they are defined in component-base upstream: https://github.com/kubernetes/component-base/blob/master/metrics/prometheus/restclient/metrics.go

I think the cardinality should be fine as the `host` label intentionally only contains the "template url".

Example:

> rest_client_request_duration_seconds_sum{host="https://172.18.0.3:6443/api/v1/namespaces/%7Bnamespace%7D/pods/%7Bname%7D?timeout=10s",verb="GET"} 0.002900856
> rest_client_request_duration_seconds_count{host="https://172.18.0.3:6443/api/v1/namespaces/%7Bnamespace%7D/pods/%7Bname%7D?timeout=10s",verb="GET"} 1

More examples can be found under: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api/8207/pull-cluster-api-e2e-main/1630798093588893696/artifacts/clusters/bootstrap/metrics/capi-system/capi-controller-manager/capi-controller-manager-9db79449b-vnfh8/metrics.txt
(please note that this file contains more rest_client metrics than I added in this PR)

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -62,11 +62,44 @@ var (
 		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 10),
 	}, []string{"verb", "url"})
 
-	requestResult = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Subsystem: RestClientSubsystem,
-		Name:      ResultKey,
-		Help:      "Number of HTTP requests, partitioned by status code, method, and host.",
-	}, []string{"code", "method", "host"})
+	// requestLatency is a Prometheus Histogram metric type partitioned by
+	// "verb", and "host" labels. It is used for the rest client latency metrics.
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb, and host.",
+			Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
+		},
+		[]string{"verb", "host"},
+	)
+
+	requestSize = prometheus.NewHistogramVec(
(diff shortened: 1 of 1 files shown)


pkg/metrics/client_go_adapter.go
```diff
@@ -62,11 +61,44 @@ var (
...
+			Help: "Response size in bytes. Broken down by verb and host.",
+			// 64 bytes to 16MB
+			Buckets: []float64{64, 256, 512, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216},
+		},
+		[]string{"verb", "host"},
+	)
+
+	requestResult = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "rest_client_requests_total",
```
sbueringer, line 97: Just aligned this to the way the metrics are defined just with name instaed of subsystem + name here: https://github.com/kubernetes/component-base/blob/master/metrics/prometheus/restclient/metrics.go#L78

pkg/metrics/client_go_adapter.go
```diff
@@ -62,11 +61,44 @@ var (
 		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 10),
 	}, []string{"verb", "url"})
```
sbueringer, line 62: I wonder if we should drop RequestLatency in l.41-62 (in a separate PR) given: 
```go
	// Deprecated: This metric is deprecated for removal in a future release: using the URL as a
	// dimension results in cardinality explosion for some consumers. It was deprecated upstream
	// in k8s v1.14 and hidden in v1.17 via https://github.com/kubernetes/kubernetes/pull/83836.
```

sbueringer:
@vincepri Would it be possible to cherry-pick this PR into release-0.14?

No problem if we don't want to do it, would be just nice to be able to pick it up earlier

Issue #2222, opened 2023-03-03
Adapt to breaking changes in 1.27 client-go


howardjohn:
controller-runtime does not compile with client-go 1.27. We should make it compatible. I believe this is possible to do in a way that works with both 1.27+ *and* 1.26-

/assign

I will send a PR for this

vincepri:
See https://github.com/kubernetes-sigs/controller-runtime/pull/2189/files

PR #2223, opened 2023-03-03
🐛 Add cross-version compatibility with client-go 1.27

howardjohn:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2222

I have manually verified this compiles with client-go 1.26 and 1.27, but I am not sure how to get test coverage of this; likely we will not be able to.

### DIFF
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -305,42 +305,63 @@ type multiNamespaceInformer struct {
 	namespaceToInformer map[string]Informer
 }
 
+type handlerRegistration struct {
+	handles map[string]toolscache.ResourceEventHandlerRegistration
+}
+
+type syncer interface {
+	HasSynced() bool
+}
+
+// HasSynced asserts that the handler has been called for the full initial state of the informer.
+// This uses syncer to be compatible between client-go 1.27+ and older versions when the interface changed.
+func (h handlerRegistration) HasSynced() bool {
+	for _, reg := range h.handles {
+		if s, ok := reg.(syncer); ok {
+			if !s.HasSynced() {
+				return false
+			}
+		}
+	}
+	return true
+}
+
 var _ Informer = &multiNamespaceInformer{}
 
 // AddEventHandler adds the handler to each namespaced informer.
 func (i *multiNamespaceInformer) AddEventHandler(handler toolscache.ResourceEventHandler) (toolscache.ResourceEventHandlerRegistration, error) {
-	handles := make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))
+	handles := handlerRegistration{handles: make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))}
 	for ns, informer := range i.namespaceToInformer {
 		registration, err := informer.AddEventHandler(handler)
 		if err != nil {
 			return nil, err
 		}
-		handles[ns] = registration
+		handles.handles[ns] = registration
 	}
 	return handles, nil
 }
 
 // AddEventHandlerWithResyncPeriod adds the handler with a resync period to each namespaced informer.
 func (i *multiNamespaceInformer) AddEventHandlerWithResyncPeriod(handler toolscache.ResourceEventHandler, resyncPeriod time.Duration) (toolscache.ResourceEventHandlerRegistration, error) {
-	handles := make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))
+	handles := handlerRegistration{handles: make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))}
 	for ns, informer := range i.namespaceToInformer {
 		registration, err := informer.AddEventHandlerWithResyncPeriod(handler, resyncPeriod)
 		if err != nil {
 			return nil, err
 		}
-		handles[ns] = registration
+		handles.handles[ns] = registration
 	}
 	return handles, nil
 }
 
 // RemoveEventHandler removes a formerly added event handler given by its registration handle.
 func (i *multiNamespaceInformer) RemoveEventHandler(h toolscache.ResourceEventHandlerRegistration) error {
-	handles, ok := h.(map[string]toolscache.ResourceEventHandlerRegistration)
+	handles, ok := h.(handlerRegistration)
 	if !ok {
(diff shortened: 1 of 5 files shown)


pkg/cache/multi_namespace_cache.go
```diff
@@ -305,42 +305,63 @@ type multiNamespaceInformer struct {
...
+// This uses syncer to be compatible between client-go 1.27+ and older versions when the interface changed.
+func (h handlerRegistration) HasSynced() bool {
+	for _, reg := range h.handles {
+		if s, ok := reg.(syncer); ok {
+			if !s.HasSynced() {
+				return false
+			}
+		}
+	}
+	return true
```
lavalamp, line 326: seems like this will give a problematic answer with the old client?

howardjohn: I would expect an old client can never call this since they would see a `type ResourceEventHandlerRegistration interface{}` as the return type. Of course they could attempt to type-cast that and would always get `true`, but that seems pretty obscure?

pkg/cache/multi_namespace_cache.go
```diff
@@ -305,42 +305,63 @@ type multiNamespaceInformer struct {
...
 func (i *multiNamespaceInformer) AddEventHandler(handler toolscache.ResourceEventHandler) (toolscache.ResourceEventHandlerRegistration, error) {
-	handles := make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))
+	handles := handlerRegistration{handles: make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))}
 	for ns, informer := range i.namespaceToInformer {
 		registration, err := informer.AddEventHandler(handler)
 		if err != nil {
 			return nil, err
 		}
-		handles[ns] = registration
+		handles.handles[ns] = registration
```
lavalamp, line 339: won't the multiple informers collide and overwrite the prior one's handle?

pkg/cache/multi_namespace_cache.go
```diff
@@ -305,42 +305,63 @@ type multiNamespaceInformer struct {
...
 func (i *multiNamespaceInformer) AddEventHandlerWithResyncPeriod(handler toolscache.ResourceEventHandler, resyncPeriod time.Duration) (toolscache.ResourceEventHandlerRegistration, error) {
-	handles := make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))
+	handles := handlerRegistration{handles: make(map[string]toolscache.ResourceEventHandlerRegistration, len(i.namespaceToInformer))}
 	for ns, informer := range i.namespaceToInformer {
 		registration, err := informer.AddEventHandlerWithResyncPeriod(handler, resyncPeriod)
 		if err != nil {
 			return nil, err
 		}
-		handles[ns] = registration
+		handles.handles[ns] = registration
```
lavalamp, line 352: same Q

pkg/cache/multi_namespace_cache.go
lavalamp: wait, there's a different informer per namespace? ok nvm, sorry.

pkg/controller/controllertest/util.go
```diff
@@ -33,7 +33,49 @@ type FakeInformer struct {
...
+}
+
+// eventHandlerWrapper wraps a ResourceEventHandler in a manner that is compatible with client-go 1.27+ and older.
+// The interface was changed in these versions.
+type eventHandlerWrapper struct {
+	handler any
+}
+
+func (e eventHandlerWrapper) OnAdd(obj interface{}) {
+	if m, ok := e.handler.(modernResourceEventHandler); ok {
```
lavalamp, line 58: clever, lol

pkg/internal/source/event_handler.go
```diff
@@ -55,6 +53,15 @@ type EventHandler struct {
 	predicates []predicate.Predicate
 }
 
+// HandlerFuncs converts EventHandler to a ResourceEventHandlerFuncs
+func (e *EventHandler) HandlerFuncs() cache.ResourceEventHandlerFuncs {
+	return cache.ResourceEventHandlerFuncs{
```
lavalamp, line 58: Maybe add a TODO to just switch to the new system in a few releases?

pkg/cache/multi_namespace_cache.go
lavalamp: I see, gotcha

lavalamp:
LGTM but I'm not a reviewer here :)

pkg/controller/controllertest/util.go
```diff
@@ -33,7 +33,49 @@ type FakeInformer struct {
...
+type legacyResourceEventHandler interface {
+	OnAdd(obj interface{})
+	OnUpdate(oldObj, newObj interface{})
+	OnDelete(obj interface{})
+}
+
+// eventHandlerWrapper wraps a ResourceEventHandler in a manner that is compatible with client-go 1.27+ and older.
+// The interface was changed in these versions.
+type eventHandlerWrapper struct {
+	handler any
```
alvaroaleman, line 54: Can we use generics to type this to one of the two interfaces?

pkg/cache/multi_namespace_cache.go
```diff
@@ -305,42 +305,63 @@ type multiNamespaceInformer struct {
...
+	handles map[string]toolscache.ResourceEventHandlerRegistration
+}
+
+type syncer interface {
+	HasSynced() bool
+}
+
+// HasSynced asserts that the handler has been called for the full initial state of the informer.
+// This uses syncer to be compatible between client-go 1.27+ and older versions when the interface changed.
+func (h handlerRegistration) HasSynced() bool {
```
alvaroaleman, line 318: I think I am blind, where do we call this?

howardjohn: We don't call this, if allows the interface to be implemented on 0.27

howardjohn: Since they added this function

pkg/controller/controllertest/util.go
howardjohn: I don't think so, or at least do not know how to. See https://github.com/golang/go/issues/49054

alvaroaleman (approved):

pkg/controller/controllertest/util.go
caozhuozi: This remind me of the http handlerfunc tech in the std library.

PR #2237, opened 2023-03-14
:bug: Avoid creating a new scheme object when combining the options

ccojocar:
This keeps the inherited scheme object created by the cluster when combining the options instead of creating a new one.
That scheme object can be used later on to register schemes. 

The inherited scheme was already configured in the cluster object, by creating a new scheme object 
for the cache object when combining the schemes will not allow the schemes registered afterwards into the
cluster scheme object to be visible into the cache, and it will result in an error when the controller is started (see below).

Configure the controller as follows:

```
	ctrlOpts := ctrl.Options{
		NewCache: cache.BuilderWithOptions(cache.Options{
			SelectorsByObject: cache.SelectorsByObject{
				&corev1.Pod{}: {
					Label: labels.SelectorFromSet(labels.Set{
						bindata.EnableRecordingLabel: "true",
					}),
				},
			},
		}),
	}

	mgr, err := ctrl.NewManager(cfg, ctrlOpts)
	if err != nil {
		return fmt.Errorf("create manager: %w", err)
	}

        if err := spodv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
		return fmt.Errorf("add SPOD config API to scheme: %w", err)
	}

        if err := mgr.Start(sigHandler); err != nil {
		return fmt.Errorf("SPOd error: %w", err)
	}
```

This fix will allow the controller to start properly instead of getting this error that the scheme is not registered:

```
E0314 10:44:24.541365 1302480 source.go:146] controller-runtime/source "msg"="kind must be registered to the Scheme" 
"error"="no kind is registered for the type v1alpha1.SecurityProfilesOperatorDaemon in scheme \"pkg/runtime/scheme.go:100"  
```

You can see more details in https://github.com/kubernetes-sigs/security-profiles-operator/pull/1543 where we encountered this issue.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -288,6 +288,12 @@ func (options Options) inheritFrom(inherited Options) (*Options, error) {
 }
 
 func combineScheme(inherited *runtime.Scheme, new *runtime.Scheme) *runtime.Scheme {
+	if inherited == nil {
+		return new
+	}
+	if new == nil {
+		return inherited
+	}
 	for gvk, t := range new.AllKnownTypes() {
 		inherited.AddKnownTypeWithName(gvk, reflect.New(t).Interface().(runtime.Object))
 	}
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -287,20 +287,11 @@ func (options Options) inheritFrom(inherited Options) (*Options, error) {
 	return &combined, nil
 }
 
-func combineScheme(schemes ...*runtime.Scheme) *runtime.Scheme {
-	var out *runtime.Scheme
-	for _, sch := range schemes {
-		if sch == nil {
-			continue
-		}
-		for gvk, t := range sch.AllKnownTypes() {
-			if out == nil {
-				out = runtime.NewScheme()
-			}
-			out.AddKnownTypeWithName(gvk, reflect.New(t).Interface().(runtime.Object))
-		}
+func combineScheme(inherited *runtime.Scheme, new *runtime.Scheme) *runtime.Scheme {
+	for gvk, t := range new.AllKnownTypes() {
+		inherited.AddKnownTypeWithName(gvk, reflect.New(t).Interface().(runtime.Object))
 	}
-	return out
+	return inherited
 }
 
 func selectMapper(def, override meta.RESTMapper) meta.RESTMapper {
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -38,11 +38,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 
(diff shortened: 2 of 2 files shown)


alvaroaleman:
Please provide a testcase for the underlying bug

ccojocar:
> Please provide a testcase for the underlying bug

I'll try to add one.

ccojocar:
@alvaroaleman I added a test case in manager. I hope that's the right place for this.

pkg/manager/manager_test.go
```diff
@@ -1156,6 +1158,45 @@ var _ = Describe("manger.Manager", func() {
 			)
 		})
 
+		Context("with custom object selector", func() {
```
alvaroaleman, line 1161: I don't understand what exactly this is supposed to test and how this relates to the problem you described in the PR body? It also passes without your codechanges btw. It might be easier to unittest `combineScheme` directly

ccojocar: I am trying to test the scenario mentioned above in the description. 

The main issue is when a scheme is combined using the BuilderWithOptions function, a new scheme object pointer is created and that pointer is configured into the cache - https://github.com/kubernetes-sigs/controller-runtime/blob/c3c1f058a9a080581e8fe99c004fcc792b2aff07/pkg/cache/cache.go#L300. 

The issue is that the original scheme object pointer is configured into the manager/controller and not the newly created one by combineScheme, so this means all the schemes added to `mgr.GetScheme()`, object are no available in the cache. As a result, the controller will fail to start, returning the error `"msg"="kind must be registered to the Scheme"`.

I think just unit-testing the `combineScheme` doesn't cover this scenario end to end. Also the `combineScheme` has unittest which already covers this entirely.

pkg/cache/cache.go
```diff
@@ -287,20 +287,17 @@ func (options Options) inheritFrom(inherited Options) (*Options, error) {
...
+		return new
+	}
+	if new == nil {
+		return inherited
+	}
+	for gvk, t := range new.AllKnownTypes() {
+		inherited.AddKnownTypeWithName(gvk, reflect.New(t).Interface().(runtime.Object))
 	}
-	return out
+	return inherited
```
joelanford, line 300: Won't this leak `new` scheme types into the inherited scheme? What happens here if (somehow) a conflicting scheme is required by a different cache, and that cache shares the same inherited scheme? Will we change the scheme out from under the first cache?

joelanford:
This may be an area of the design where we didn't really consider the possibility of schemes changing dynamically.

Should adding more types to `mgr.GetScheme()` mean anything that inherited that scheme in the past pick up those changes? What if the inheritee has already added that type in its sub-scheme?

This seems like it could absolutely be an issue in the case of a multi-cluster controller where the schemes for different clusters may need to have conflicting registrations. I can't personally think of a valid case where the same GVK in two different clusters (or maybe even different contexts within the same cluster?) would need a different go type associated with them, but it doesn't seem like we should preclude that possibility.

vincepri:
> This seems like it could absolutely be an issue in the case of a multi-cluster controller where the schemes for different clusters may need to have conflicting registrations. I can't personally think of a valid case where the same GVK in two different clusters (or maybe even different contexts within the same cluster?) would need a different go type associated with them, but it doesn't seem like we should preclude that possibility.

It should probably be a hard limitation that the scheme for the time being has to be global.

In terms of this PR, I'm ok allowing the Scheme to be shared, given that's one of those things that's already passed around

alvaroaleman:
The corresponding core was removed in https://github.com/kubernetes-sigs/controller-runtime/pull/2300, so this isn't needed anymore.

PR #2240, opened 2023-03-14
`:warning:`  Remove global metrics registration in init, add registration func

### LINKED ISSUE #2238: Don't set global workqueue MetricsProvider
Currently, controller-runtime [sets the global workqueue MetricsProvider](https://github.com/kubernetes-sigs/controller-runtime/blob/e2d8821b277fb57f3a951f430bb1746d8297b853/pkg/metrics/workqueue.go#L99) in an init method which makes it impossible for users of the library to configure their own.

As noted in:
* https://github.com/kubernetes/client-go/issues/1193
* https://github.com/kubernetes/kubernetes/pull/114242#issuecomment-1345613392
* SIG API meeting: https://docs.google.com/document/d/1x9RNaaysyO0gXHIr1y50QFbiL1x8OWnk2v3XnrdkT5Y/edit?pli=1#heading=h.41qdxxsr3sdz

austince:
This PR removes setting the workqueue global metrics provider in an `init` method and instead provides a way for consumers of this lib to set it themselves. Otherwise, there is no way to allow consumers to use a different metrics provider.

Also proposed allowing per-queue metrics providers, which would in effect also solve this problem, but the jury's out on it: https://github.com/kubernetes/kubernetes/pull/114242

Closes #2238


austince:
cc @alvaroaleman

pkg/metrics/workqueue.go
```diff
@@ -95,8 +95,18 @@ func init() {
...
 	Registry.MustRegister(retries)
+}
+
+// SetDefaultGlobalWorkqueueMetricsProvider registers the GlobalWorkqueueMetricsProvider
+// as the global workqueue.MetricsProvider.
+func SetDefaultGlobalWorkqueueMetricsProvider() {
+	workqueue.SetProvider(GlobalWorkqueueMetricsProvider())
+}
 
-	workqueue.SetProvider(workqueueMetricsProvider{})
```
alvaroaleman, line 99: As mentioned in the k/k pr, we wont merge this. This breaks the majority use-case (`workqueue metrics just work without doing anything`) in order to support a minority use-case (`set custom provider`).

/hold

austince: Is this not what you were suggesting here? https://github.com/kubernetes/kubernetes/pull/114242#issuecomment-1468802756

This is marked as a breaking change, and allows consumers to set it. Unsure what the path forward is if it is neither of the options I've proposed.

alvaroaleman: No, my suggestion was to change the upstream workqueue to not use a sync.Once so what is set by controller-runtime can be overridden later if desired.

austince: Alright, I'll send a PR for that -- seems to paper over the fact that this library is mutating global state and the "metrics just work without doing anything" could be addressed somewhere else (like kubebuilder, which sets up a `main`), but if this gets it in 🤷🏼 

I'll leave this open in the meantime if others want to weigh in.

sbueringer: I think changing kubebuilder does not really mitigate the situation. I don't have a good overview over the ecosystem but the controllers that I'm aware of don't regualarly regenerate their main file with kubebuilder.

austince: Okie, I'll close this then -- I've also got a PR for what Alvaro suggested here, if you guys want to drive that approach: https://github.com/kubernetes/kubernetes/pull/116616

PR #2246, opened 2023-03-23
🐛 Preserve unstructured object GVKs in cache.Options.ByObject

joelanford:
This patch enables unstructured objects that have populated GVKs to be used as keys when configuring `ByObject` cache options.

Prior to this patch, a cache setup like the following causes `manager.New` to return an error with "unstructured object has no kind":
```go
obj := &unstructured.Unstructured{}
obj.SetGroupVersionKind(schema.GroupVersionKind{"example.com", "v1", "MyKind"})
_, err := manager.New(cfg, manager.Options{
	NewCache: cache.BuilderWithOptions(cache.Options {
		ByObject: map[client.Object]cache.ByObject{
			obj: cache.ByObject{}
		},
	})
})
```


inteon (approved):
LGTM

PR #2247, opened 2023-03-25
✨ Add Source field to reconcile.Request

inteon:
When you use the `Owns` watch for your controller, it is not possible to deduce in the reconciliation loop what resource was actually changed. This can be a cause of a lot of unnecessary updates (because of a partially-outdated cache).

See the following example (based on https://book-v1.book.kubebuilder.io/basics/what_is_a_controller.html):
image

^^ In this example, "Pod 3" is created 2 times because the "Pod 3 Running Event" was not received in time.
More pods and a slower connections can result in more unnecessary create requests

This PR adds the source Name + Namespace to reconcile.Request resource.
I do now realised that this is NOT a fault-less solution, because it breaks workqueue deduplication.
I added an example in the `./examples/crd-multi-target/` folder, hopefully that further clarifies how information about the source could be used to prevent unnecessary updates.


Makefile
```diff
@@ -102,7 +102,25 @@ modules: ## Runs go mod to ensure modules are up to date.
 
 .PHONY: generate
 generate: $(CONTROLLER_GEN) ## Runs controller-gen for internal types for config file
```
inteon, line 104: I updated the makefile to simplify testing the new example.

alvaroaleman:
This is an antipattern. Your reconciliation should work correctly and efficiently regardless of what triggered it.

/hold

inteon:
@alvaroaleman 

> This is an antipattern. Your reconciliation should work correctly and efficiently regardless of what triggered it.

I'm missing some info to do the reconciliation efficiently, I do not know what owned resources are stale and what resources are not. So I have to update all out-of-date child resources every time a parent reconciliation is triggered.

> Additionally, the workqueue uses the request for deduplication, which is why the example as you described it is incorrect and why we can't merge something like this, as it would break that.

I agree, this approach is probably not the best approach. However, I still think the idea of being able to access the all the source GVK + Name & Namespace are crucial for an efficient reconciliation.

alvaroaleman:
> I'm missing some info to do the reconciliation efficiently, I do not know what owned resources are stale and what resources are not. So I have to update all out-of-date child resources every time a parent reconciliation is triggered.
I updated the PR description so it better explains the problem and how some source context can be useful. 

Yes, which is why you need to find this out during your reconciliation what is stale and what is not. The cache lag is an orthogonal issue you have to deal with either way. Possibilities include creating owned resources with a deterministic name so creating them a second time will error or blocking reconciliation from happening until new changes are visible in the cache. The upstream replicaset controller does the latter through [expectations](https://github.com/kubernetes/kubernetes/blob/8f15859afc9cfaeb05d4915ffa204d84da512094/pkg/controller/replicaset/replica_set.go#L99)

> I agree, this approach is probably not the best approach. However, I still think the idea of being able to access the all the source GVK + Name & Namespace are crucial for an efficient reconciliation.

No, in the contrary. Any relevant event creates an entry in the workqueue and entries in the workqueue get deduplicated. With your approach, you can not let this de-duplication happen ever which means you must reconcile once per event. You also have to write code for each specific event rather than coding once how to check if the desired state is present and if not, how to get there, resulting in more complicated logic.

alvaroaleman:
In general, I would recommend that you try to find some existing controllers that you think would profit from this and try to understand how they work :) Upstream Kube doesn't use controller-runtime but the basic principles on what to [not-] do when writing a controller apply regardless of using c-r or not.

inteon:
Thank you for helping me. I'll close the PR, but will continue experimenting with [expectations](https://github.com/kubernetes/kubernetes/blob/8f15859afc9cfaeb05d4915ffa204d84da512094/pkg/controller/replicaset/replica_set.go#L99) or other methods to determine if a resource is stale.
I feel like I'm slowly starting to grasp how to approach these kind of problems.

PR #2248, opened 2023-03-27
✨ Added methods to intercept calls to a fake k8s client

ludydoo:
This PR adds some methods to allow consumers of the fake client to intercept kubernetes api calls to inject custom logic, which is helpful while testing e.g. expected behavior upon certain types of failures.


ludydoo:
Hi @alvaroaleman , thanks a lot for your quick review! What's your opinion about: would it make more sense to put this into the `client` package? Personally I don't see much use-case for a non-testing scenario, so the `fake` package makes sensse, but in terms of usage, perhaps it's clearer if consumers will use some `client.Interceptor` rather than some `fake.Interceptor` ?

pkg/client/fake/intercept.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (c interceptor) DeleteAllOf(ctx context.Context, obj client.Object, opts ...client.DeleteAllOfOption) error {
+	if c.fns.DeleteAllOf != nil {
+		return c.fns.DeleteAllOf(ctx, c.client, obj, opts...)
+	}
+	return c.client.DeleteAllOf(ctx, obj, opts...)
+}
+
+func (c interceptor) Status() client.SubResourceWriter {
```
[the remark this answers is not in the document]
ludydoo: @inteon good point, no benefit at all. I'll remove that.

ludydoo:
@alvaroaleman @inteon please let me know if there is anything I should tweak or modify

pkg/client/fake/intercept.go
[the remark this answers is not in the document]
inteon: we then probably want to call `c.SubResource("status")` here instead of `c.client.Status()`?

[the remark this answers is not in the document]
ludydoo: @inteon fixed and added tests for that.

alvaroaleman:
@ludydoo we had a lot of discussion about fakeclient changes like this, which is why I can't just merge without some of the other maintainers approving and two of them are on PTO unfortunately. So this will take a bit of time, but there is nothing expected from you at this point. Sorry about that.

vincepri (commented):
This looks like a good way to allow some mocking to happen in the fake client, I'm personally +1, but would like to hear @sbueringer's opinion as well.

sbueringer:
Just back from PTO, will take a look ASAP, not sure if it will be before KubeCon

sbueringer:
Looks good to me as well. Looks like it needs a rebase to pick up the change to the golangci-lint action.

One thought (similar to: https://github.com/kubernetes-sigs/controller-runtime/pull/2248#issuecomment-1486245876): as this functionality is probably often used together with fake client, but it can also be used with a regular client, should we move it out of the fake package?

I was thinking about maybe a new package like pkg/client/intercepter. Has the benefit to keep it clearly separate from the fake client and if we rename the func we could make it e.g. interceptor.NewClient instead of fake.Interceptor.

alvaroaleman:
If we move this into a distinct package, can we make this discoverable by integrating with the fake client builder by adding a With method for each of the interceptor methods that sets them?

ludydoo:
> If we move this into a distinct package, can we make this discoverable by integrating with the fake client builder by adding a With method for each of the interceptor methods that sets them?

I'm not sure how to resolve the circular dependency in this case, since the `interceptor_test` import the `fake` package.

pkg/client/interceptor/intercept.go
```diff
@@ -0,0 +1,197 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type (
+	// GetFn is a function that can be used to intercept GET calls.
+	GetFn func(ctx context.Context, client client.WithWatch, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error
```
sbueringer, line 15: Was it necessary to have the funcs as separate types? (vs just inline in the struct as before)

pkg/client/interceptor/intercept.go
```diff
@@ -0,0 +1,197 @@
...
+	SubResourceGetFn func(ctx context.Context, client client.SubResourceClient, obj client.Object, subResource client.Object, opts ...client.SubResourceGetOption) error
+	// SubResourceCreateFn is a function that can be used to intercept SubResource CREATE calls.
+	SubResourceCreateFn func(ctx context.Context, client client.SubResourceClient, obj client.Object, subResource client.Object, opts ...client.SubResourceCreateOption) error
+	// SubResourceUpdateFn is a function that can be used to intercept SubResource UPDATE calls.
+	SubResourceUpdateFn func(ctx context.Context, client client.SubResourceClient, obj client.Object, opts ...client.SubResourceUpdateOption) error
+	// SubResourcePatchFn is a function that can be used to intercept SubResource PATCH calls.
+	SubResourcePatchFn func(ctx context.Context, client client.SubResourceClient, obj client.Object, patch client.Patch, opts ...client.SubResourcePatchOption) error
+
+	// Fns contains functions that are called instead of the underlying client's methods.
+	Fns struct {
```
sbueringer, line 43: Sorry missed this before. Can we maybe replace all Fn with Func and Fns with Funcs? (in all files)

Just reads a lot better in my opinion

sbueringer:
@ludydoo Can you please drop the merge commits & rebase to main?

(I know you probably kept them for easier reviews, but the delta of this PR would be easier to see without the merge commits)

pkg/client/interceptor/intercept.go
ludydoo: No, I thought it was cleaner this way, but I removed it as it is not necessary

pkg/client/interceptor/intercept.go
ludydoo: Done

ludydoo:
> @ludydoo Can you please drop the merge commits, squash and rebase to main?
> 
> (I assume you probably kept them for easier reviews, but the delta of this PR would be easier to see without the merge commits)

@sbueringer just addressed your comments, and squashed & rebased to main. Please let me know if you think there should be some more changes.

sbueringer:
@ludydoo Thank you very much!!

/lgtm

alvaroaleman (approved):
Thanks!

alvaroaleman:
Thanks a lot for this @ludydoo ! This makes injecting custom beheavior into the fakeclient for tests much easier :)

ludydoo:
Thank you for your reviews and advices @alvaroaleman @sbueringer @inteon @vincepri !

PR #2252, opened 2023-03-27
:bug: fix a bug in multinamespaced cache implementation

varshaprasad96:
This commit fixes a bug that was brought in long ago in #1520. When the object's scope is not deterministic from the RESTMapper it should return an error instead of ignoring it.


varshaprasad96:
cc: @alebedev87

inteon:
Thank you for fixing the bug
/lgtm

alvaroaleman (approved):

PR #2256, opened 2023-03-29
:bug: type print will always <nil>

qmloong:
This log will always print <nil> when indexer abnormal.


FillZpp:
/ok-to-test
/lgtm
thanks @qmloong

PR #2257, opened 2023-03-31
🐛  ForceOwnership should work with subresources

### LINKED ISSUE #2125: 0.14.1 regression: cannot use server-side apply
Hi,

After upgrading from 0.13.1 to 0.14.1, I can no longer use server-side apply.

In 0.13.1 I used to be able to do this:
`err = r.Status().Patch(ctx, obj, client.Apply, client.FieldOwner("whatever"), client.ForceOwnership)`

In 0.14.1 this results in:
```
Cannot use 'client.FieldOwner("whatever")' (type client.FieldOwner) as the type SubResourcePatchOptionType does not implement 'SubResourcePatchOption' as some methods are missing:ApplyToSubResourcePatch(*SubResourcePatchOptions)
Cannot use 'client.ForceOwnership' (type forceOwnership) as the type SubResourcePatchOptionType does not implement 'SubResourcePatchOption' as some methods are missing:ApplyToSubResourcePatch(*SubResourcePatchOptions)
```

I assume this is due to the breaking change introduced in #2072. It seems that client.FieldOwner and client.ForceOwnership were not updated to use the new SubResourcePatchOption interface instead of the existing PatchOption interface.

If this was intended, I would appreciate any insight about what the line of code should look like in 0.14.1.

Thanks!

justinsb:
Ensure that we can force when applying to subresources (in particular status)

FIxes #2125

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -788,6 +788,11 @@ func (forceOwnership) ApplyToPatch(opts *PatchOptions) {
 	opts.Force = &definitelyTrue
 }
 
+func (forceOwnership) ApplyToSubResourcePatch(opts *SubResourcePatchOptions) {
+	definitelyTrue := true
+	opts.Force = &definitelyTrue
+}
+
 // }}}
 
 // {{{ DeleteAllOf Options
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
/cherrypick release-0.14

Issue #2230, opened 2023-03-09
Fake Client .Update updates both spec and status?


ZakMiller:
The fake client's `Update` call updates both the spec and status of the object.

This is different from how the real client behaves. This can lead to situations where someone writes a unit test, the unit test passes, but the behavior will be different when dealing with a real client.

Is this known/intended? It seems like the fake client should update the spec and replace the status with whatever the value is on the "server".

fovgrubby128:
Hey @vincepri @alvaroaleman, sorry to ping you here. I also found the same weird behavior from `fakeClient`. Could you help to have a look on this issue? thanks!

alvaroaleman:
It depends if resources have or do not have a status subresource and today there is no way in the fake client to figure that out. Never updating the status would be just as wrong and also breaking.

Ideas on how to improve this are always welcome :)

sbueringer:
[Only for CRDs] I wonder if it's an option to change behavior if the corresponding CRD has also been added to the fake client. With that information the fake client could handle the update correctly

Thinking about this a bit more, it's a bit strange and a difference to Kubernetes / envtest that the fake client does not require to add the CRDs before CRs can be added. But I think changing the fake client to require CRDs would break to many users of fake client.

ZakMiller:
I'm not sure I understand.

We're using the fake client like this:

```
	fakeClient := fake.NewClientBuilder().
		WithScheme(s).
		Build()
```

If I do a `fakeClient.Create` call with a resource not included in that scheme, then I'll get back an error " no kind is registered for the type ..."

I guess there's some other way of interacting with it where you don't have to register the type?

Regardless, for our case, we're primarily concerned with the behavior for registered CRDs. If we changed the fake client `Update` to not update known subresources because the CRD is registered, would that work?

alvaroaleman:
> [Only for CRDs] I wonder if it's an option to change behavior if the corresponding CRD has also been added to the fake client

I think what @sbueringer was talking about is if the CRD object has been added to the fakeclient rather than the type registered in the scheme, as the CRD has the setting if the `status` subresource is or isn't enabled

sbueringer:
Yes exactly!

If you write your unit tests with env test instead of fake client or move unit tests from fake client to envtest the first thing you have to do is actually deploy your CRDs :)

ZakMiller:
Okay cool. That all makes sense to me.

[...]

PR #2259, opened 2023-04-01
:warning: Fakeclient: Fix status handling

alvaroaleman:
This change fixes the status handling of the fake client by

* Adding a new `WithStatusSubresource` option and pre-poluating it with
  all in-tree resources that have a status subresource
* Making its `Update` and `Patch` not change the status of any such
  resource
* Making its status client `Update` and `Patch` only change the status
  for any such resources

Since this was completely broken before in that both non-status and
status Update/Patch would always update everything and the status
resources get pre-populated, this is a breaking change. Any test that
relied on the previous behavior would pass incorrectly though, as it
didn't match what the Kubernetes API does.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2230

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -17,11 +17,13 @@ limitations under the License.
 package fake
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
+	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync"
@@ -35,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/hold
I realized there is one piece missing: We need to error if status is used on objects that do not have the status subresource

alvaroaleman:
> I realized there is one piece missing: We need to error if status is used on objects that do not have the status subresource

Added that.

/hold cancel

pkg/client/fake/client.go
```diff
@@ -350,6 +384,20 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
...
+			if err := copyNonStatusFrom(oldObject, obj); err != nil {
+				return fmt.Errorf("failed to copy non-status field for object with status subresouce: %w", err)
+			}
+		} else { // copy status from original object
+			if err := copyStatusFrom(oldObject, obj); err != nil {
+				return fmt.Errorf("failed to copy the status for object with status subresource: %w", err)
+			}
+		}
+	} else if isStatus {
+		return apierrors.NewNotFound(gvr.GroupResource(), accessor.GetName())
```
sbueringer, line 398: @alvaroaleman just fyi

Regarding:
> Any test that relied on the previous behavior would pass incorrectly though, as it
didn't match what the Kubernetes API does.

In my experience this change is relatively safe as all Update/Patch calls will fail with this error hard

sbueringer: Ah that is assuming that folks where using the status client correctly. Of course doesn't apply if they used the regular Patch/Update calls

PR #2261, opened 2023-04-01
:book: Add a design for cache options configuration

alvaroaleman:
Prior art
* https://github.com/kubernetes-sigs/controller-runtime/pull/1962

/cc @joelanford @vincepri @sbueringer @FillZpp

### DIFF
--- a/designs/cache_options.md
+++ b/designs/cache_options.md
@@ -1,0 +1,226 @@
+Cache Options
+===================
+
+This document describes how we imagine the cache options to look in
+the future.
+
+## Goals
+
+* Align everyone on what settings on the cache we want to support and
+  their configuration surface
+* Ensure that we support both complicated cache setups and provide an
+  intuitive configuration UX
+
+## Non-Goals
+
+* Describe the design and implementation of the cache itself.
+  The assumption is that the most granular level we will end up with is
+  "per-object multiple namespaces with distinct selectors" and that this
+  can be implemented using a "meta cache" that delegates per object and by
+  extending the current multi-namespace cache
+* Outline any kind of timeline for when these settings will be implemented.
+  Implementation will happen gradually over time whenever someone steps up
+  to do the actual work
+
+## Proposal
+
+
+```
+const (
+   AllNamespaces = corev1.NamespaceAll
+)
+
+type Config struct {
+  // LabelSelector specifies a label selector. A nil value allows to
+  // default this.
+  LabelSelector labels.Selector
+
+  // FieldSelector specifics a field selector. A nil value allows to
+  // default this.
+  FieldSelector fields.Selector
+
+  // Transform specifies a transform func. A nil value allows to default
+  // this.
+  Transform     toolscache.TransformFunc
+
+  // UnsafeDisableDeepCopy specifies if List and Get requests against the
+  // cache should not DeepCopy. A nil value allows to default this.
+  UnsafeDisableDeepCopy *bool
+}
+
+
+type ByObject struct {
+  // Namespaces maps a namespace name to cache setting. If set, only the
+  // namespaces in this map will be cached.
+  //
+  // Settings in the map value that are unset because either the value as a
+  // whole is nil or because the specific setting is nil will be defaulted.
+  // Use an empty value for the specific setting to prevent that.
+  //
+  // It is possible to have specific Config for just some namespaces
+  // but cache all namespaces by using the AllNamespaces const as the map key.
+  // This wil then include all namespaces that do not have a more specific
+  // setting.
+  //
+  // A nil map allows to default this to the cache's DefaultNamespaces setting.
+  // An empty map prevents this.
+  //
+  // This must be unset for cluster-scoped objects.
+  Namespaces map[string]*Config
+
+  // Config will be used for cluster-scoped objects and to default
+  // Config in the Namespaces field.
+  //
+  // It gets defaulted from the cache'sDefaultLabelSelector, DefaultFieldSelector,
+  // DefaultUnsafeDisableDeepCopy and DefaultTransform.
+  Config *Config
+}
+
+type Options struct {
+  // ByObject specifies per-object cache settings. If unset for a given
+  // object, this will fall through to Default* settings.
+  ByObject map[client.Object]*ByObject
+
+  // DefaultNamespaces maps namespace names to cache settings. If set, it
+  // will be used for all objects that have a nil Namespaces setting.
+  //
+  // It is possible to have a specific Config for just some namespaces
+  // but cache all namespaces by using the `AllNamespaces` const as the map
+  // key. This wil then include all namespaces that do not have a more
+  // specific setting.
+  //
+  // The options in the Config that are nil will be defaulted from
+  // the respective Default* settings.
+  DefaultNamespaces map[string]*Config
+
+  // DefaultLabelSelector is the label selector that will be used as
+  // the default field label selector for everything that doesn't
+  // have one configured.
+  DefaultLabelSelector labels.Selector
+
+  // DefaultFieldSelector is the field selector that will be used as
+  // the default field selector for everything that doesn't have
+  // one configured.
+  DefaultFieldSelector fields.Selector
+
+  // DefaultUnsafeDisableDeepCopy is the default for UnsafeDisableDeepCopy
+  // for everything that doesn't specify this.
+  DefaultUnsafeDisableDeepCopy *bool
+
+  // DefaultTransform will be used as transform for all object types
+  // unless they have a more specific transform set in ByObject.
+  DefaultTransform toolscache.TransformFunc
+
+  // HTTPClient is the http client to use for the REST client
+  HTTPClient *http.Client
+
+  // Scheme is the scheme to use for mapping objects to GroupVersionKinds
+  Scheme *runtime.Scheme
+
+  // Mapper is the RESTMapper to use for mapping GroupVersionKinds to Resources
+  Mapper meta.RESTMapper
+
+  // SyncPeriod determines the minimum frequency at which watched resources are
+  // reconciled. A lower period will correct entropy more quickly, but reduce
+  // responsiveness to change if there are many watched resources. Change this
+  // value only if you know what you are doing. Defaults to 10 hours if unset.
+  // there will a 10 percent jitter between the SyncPeriod of all controllers
+  // so that all controllers will not send list requests simultaneously.
+  //
+  // This applies to all controllers.
+  //
+  // A period sync happens for two reasons:
+  // 1. To insure against a bug in the controller that causes an object to not
+  // be requeued, when it otherwise should be requeued.
+  // 2. To insure against an unknown bug in controller-runtime, or its dependencies,
+  // that causes an object to not be requeued, when it otherwise should be
+  // requeued, or to be removed from the queue, when it otherwise should not
+  // be removed.
+  //
+  // If you want
+  // 1. to insure against missed watch events, or
+  // 2. to poll services that cannot be watched,
+  // then we recommend that, instead of changing the default period, the
+  // controller requeue, with a constant duration `t`, whenever the controller
+  // is "done" with an object, and would otherwise not requeue it, i.e., we
+  // recommend the `Reconcile` function return `reconcile.Result{RequeueAfter: t}`,
+  // instead of `reconcile.Result{}`.
+  SyncPeriod *time.Duration
+
+}
+```
+
+
+## Example usages
+
+### Cache ConfigMaps in the `public` and `kube-system` namespaces and Secrets in the `operator` Namespace
+
+
+```
+cache.Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.ConfigMap{}: {
+      Namespaces: map[string]*cache.Config{
+        "public":      {},
+        "kube-system": {},
+      },
+    },
+    &corev1.Secret{}: {Namespaces: map[string]*Config{
+        "operator": {},
+    }},
+  },
+}
+```
+
+### Cache ConfigMaps in all namespaces without selector, but have a selector for the `operator` Namespace
+
+```
+cache.Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.ConfigMap{}: {
+      Namespaces: map[string]*cache.Config{
+        cache.AllNamespaces: nil,                   // No selector for all namespaces...
+        "operator": {LabelSelector: labelSelector}, // except for the operator namespace
+      },
+    },
+  },
+}
+```
+
+
+### Only cache the `operator` namespace for namespaced objects and all namespaces for Deployments
+
+```
+cache.Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &appsv1.Deployment: {Namespaces: map[string]*cache.Config{
+       cache.AllNamespaces: nil,
+    }},
+  },
+  DefaultNamespaces: map[string]*cache.Config{
+      "operator": nil,
+  },
+}
+```
+
+### Use a LabelSelector for everything except Nodes
+
+```
+cache.Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.Node: {LabelSelector: labels.Everything()},
+  },
+  DefaultLabelSelector: myLabelSelector,
+}
+```
+
+### Only cache namespaced objects in the `foo` and `bar` namespace
+
+```
+cache.Options{
+  DefaultNamespaces: map[string]*cache.Config{
+    "foo": nil,
+    "bar": nil,
+  }
+}
+```


designs/cache_options.md
```diff
@@ -0,0 +1,104 @@
...
+  extending the current multi-namespace cache
+* Outline any kind of timeline for when these settings will be implemented.
+  Implementation will happen gradually over time whenever someone steps up
+  to do the actual work
+
+## Proposal
+
+
+```
+type CacheSetting struct {
```
varshaprasad96, line 29: It would be super helpful to have implementation examples (with `CacheSetting`) of different scenarios in the proposal.

designs/cache_options.md
```diff
@@ -0,0 +1,104 @@
...
+  // but cache all namespaces by using the empty string as map key for
+  // "all namespaces". This wil then include all namespaces that do not have
+  // a more specific setting.
+  //
+  // A nil map allows to default this to the caches DefaultNamespaces setting.
+  // An empty map prevents this.
+  Namespaces map[string]*CacheSetting
+
+  // CacheSettings will be used if Namespaces is empty or for entries in
+  // Namespaces that have a nil value.
```
varshaprasad96, line 66: This is for cluster scoped objects? How would it work if we would like to cache specific cluster scoped objects and all namespaces scoped objects (with specific CacheSettings)?

alvaroaleman: I also added a node to ByObject.Namespaces that that setting is invalid for cluster-scoped objects

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+  // default this.
+  FieldSelector fields.Selector
+
+  // Transform specifies a transform func. A nil value allows to default
+  // this.
+  Transform     toolscache.TransformFunc
+
+  // UnsafeDisableDeepCopy specifies if List and Get requests against the
+  // cache should not DeepCopy. A nil value allows to default this.
+  UnsafeDisableDeepCopy *bool
```
seh, line 44: The double negative here is confusing to read. It's not safe, and it disables behavior, without committing to what the default value would be.

Alternatives to consider:
- _CopyDeeply_
- _CopyDefensively_
- _CopyOnRead_
- _ReturnDefensiveCopies_

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+  // A nil value in this map means "fall through to the ByObject CacheSetting"
+  //
+  // An empty value in this map means "No Selectors".
+  //
+  // It is possible to have specific CacheSettings for just some namespaces
+  // but cache all namespaces by using the empty string as map key for
+  // "all namespaces". This wil then include all namespaces that do not have
+  // a more specific setting.
+  //
+  // A nil map allows to default this to the caches DefaultNamespaces setting.
```
seh, line 61: I think you meant for "caches" to mean "possessed by the cache."
```suggestion
  // A nil map allows to default this to the cache's DefaultNamespaces setting.
```

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+}
+
+
+type ByObject struct {
+  // Namespaces maps a namespace name to cache setting. If set, only the
+  // namespaces in this map will be cached.
+  //
+  // A nil value in this map means "fall through to the ByObject CacheSetting"
+  //
+  // An empty value in this map means "No Selectors".
```
seh, line 54: How do you detect an empty value? Does this involve comparing all of a `CacheSetting`'s against the zero values?

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+  // An empty map prevents this.
+  //
+  // This must be unset for cluster-scoped objects.
+  Namespaces map[string]*CacheSetting
+
+  // CacheSettings will be used if Namespaces is empty or for entries in
+  // Namespaces that have a nil value.
+  //
+  // If nil, this will be defaulted to the caches DefaultLabelSelector,
+  // DefaultFieldSelector and DefaultTransform.
```
seh, line 71:
```suggestion
  // DefaultFieldSelector, and DefaultTransform.
```

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+
+  // DefaultNamespaces maps namespace names to cache settings. If set, it
+  // will be usd for all objects that have a nil Namespaces setting.
+  //
+  // It is possible to have specific CacheSettings for just some namespaces
+  // but cache all namespaces by using the empty string as map key for
+  // "all namespaces". This wil then include all namespaces that do not have
+  // a more specific setting.
+  //
+  // If CacheSetting is nil, DefaultLabelSelector, DefaultFieldSelector
```
seh, line 88:
```suggestion
  // If CacheSetting is nil, DefaultLabelSelector, DefaultFieldSelector,
```

designs/cache_options.md
seh: You want that to be in the field name, and not just in the field's accompanying documentation?

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+
+  // CacheSettings will be used if Namespaces is empty or for entries in
+  // Namespaces that have a nil value.
+  //
+  // If nil, this will be defaulted to the cache's DefaultLabelSelector,
+  // DefaultFieldSelector, and DefaultTransform.
+  CacheSettings *CacheSetting
+}
+
+type Options struct {
```
FillZpp, line 75: 1. So where is the other fields like `Scheme`, `SyncPeriod`? They should remain in Options, right?
2. Will the `Namespaces []string` be replaced by the new `DefaultNamespaces map[string]*CacheSetting`?
3. Is the `DefaultTransform` field removed?

designs/cache_options.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Cache ConfigMaps in all namespaces without selector, but have a selector for the `operator` Namespace
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.ConfgMap{}: {
+      Namespaces: map[string]*cache.CacheSetting{
+        "": {},                                    // No selector for all namespaces...
+        "operator" {LabelSelector: labelSelector}, // except for the operator namespace
```
umangachapagain, line 138:
```suggestion
        "operator": {LabelSelector: labelSelector}, // except for the operator namespace
```

designs/cache_options.md
alvaroaleman: 1. Yes, I guess I didn't include them because they are global options, will add
2. Yes
3. No, that is an oversight, will add

designs/cache_options.md
alvaroaleman: I think this comment is in the wrong place but having a Policy that could only possibly ever have two values just to avoid the negative doesn't seem great and I also think that ppl who want this option will have a hard time finding it if its hidden like that

designs/cache_options.md
alvaroaleman: Added the missing options, ptal

FillZpp:
/lgtm
/hold
see if there is more discussion @vincepri @joelanford @sbueringer

designs/cache_options.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+
+### Only cache the `operator` namespace except for Deployments
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &appsv1.Deployment: {Namespaces: map[string]*cacheSetting:
+       "": {},
```
vincepri, line 193: This isn't super clear that's caching all namespaces, can we have a constant that's a catch all instead?

designs/cache_options.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+### Use a LabelSelector for everything except Nodes
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.Node: {},
+  },
+  DefaultLabelSelector: myLabelSelector,
```
vincepri, line 209: Should we capture the "global selectors and options" under a different field for clarity?

designs/cache_options.md
```diff
@@ -0,0 +1,211 @@
...
+  },
+}
+```
+
+### Cache ConfigMaps in all namespaces without selector, but have a selector for the `operator` Namespace
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.ConfgMap{}: {
```
vincepri, line 176:
```suggestion
    &corev1.ConfigMap{}: {
```

designs/cache_options.md
alvaroaleman: You mean a constant for the key?

designs/cache_options.md
```diff
@@ -0,0 +1,211 @@
...
+### Only cache the `operator` namespace except for Deployments
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &appsv1.Deployment: {Namespaces: map[string]*cacheSetting:
+       "": {},
+    },
+  },
+  DefaultNamespaces: map[string]*cache.CacheSetting{
```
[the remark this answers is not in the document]
alvaroaleman: The reason I used `DefaultNamespaces` is to clarify that this will only be used if `ByObject` has no namespace settings

designs/cache_options.md
alvaroaleman: Well, its a global default. How would you structure them instead?

designs/cache_options.md
vincepri: Yes, something like `cache.AllNamespaces` or something like that

designs/cache_options.md
```diff
@@ -0,0 +1,211 @@
...
+  extending the current multi-namespace cache
+* Outline any kind of timeline for when these settings will be implemented.
+  Implementation will happen gradually over time whenever someone steps up
+  to do the actual work
+
+## Proposal
+
+
+```
+type CacheSetting struct {
```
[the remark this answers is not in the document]
alvaroaleman: In the `Cache` package. Maybe simply stick with the existing `Options` or WDYT?

designs/cache_options.md
JoelSpeed: Given that will be a string of some kind, what prevents that from clashing with someones namespace name? You'd have to pick a value that is not a valid kubernetes object name right?

JoelSpeed: Perhaps `cache.AllNamespace = corev1.NamespaceAll` would be the way to go here, you can put the constant in to make it obvious what you mean, but underneath, it still uses the empty string to avoid any name clashing complications?

vincepri: That sounds good to me, what I was looking for was mostly clarity in code through a constant

alvaroaleman: Added

designs/cache_options.md
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+type ByObject struct {
+  // Namespaces maps a namespace name to cache setting. If set, only the
+  // namespaces in this map will be cached.
+  //
+  // A nil value in this map means "fall through to the ByObject CacheSetting"
+  //
+  // An empty value in this map means "No Selectors".
```
sbueringer, line 59: Does it also mean no transform?

designs/cache_options.md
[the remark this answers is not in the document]
sbueringer: Config?

designs/cache_options.md
```diff
@@ -0,0 +1,216 @@
...
+type Options struct {
+  // ByObject specifies per-object cache settings. If unset for a given
+  // object, this will fall through to Default* settings.
+  ByObject map[client.Object]*ByObject
+
+  // DefaultNamespaces maps namespace names to cache settings. If set, it
+  // will be usd for all objects that have a nil Namespaces setting.
+  //
+  // It is possible to have specific CacheSettings for just some namespaces
+  // but cache all namespaces by using the empty string as the map key for
```
sbueringer, line 89: Should this also be the `AllNamespaces` const instead of the empty string

designs/cache_options.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+### Cache ConfigMaps in all namespaces without selector, but have a selector for the `operator` Namespace
+
+```
+Options{
+  ByObject: map[client.Object]*cache.ByObject{
+    &corev1.ConfigMap{}: {
+      Namespaces: map[string]*cache.CacheSetting{
+        "": {},                                    // No selector for all namespaces...
```
sbueringer, line 183: nit: If my understanding is correct we would use the AllNamespaces const now here, right?

designs/cache_options.md
```diff
@@ -0,0 +1,216 @@
...
+  // will be usd for all objects that have a nil Namespaces setting.
+  //
+  // It is possible to have specific CacheSettings for just some namespaces
+  // but cache all namespaces by using the empty string as the map key for
+  // "all namespaces". This wil then include all namespaces that do not have
+  // a more specific setting.
+  //
+  // If CacheSetting is nil, DefaultLabelSelector, DefaultFieldSelector,
+  // and DefaultTransform will be used if set.
+  DefaultNamespaces map[string]*CacheSetting
```
tsaarni, line 95: Is the intention that one will set `Options.DefaultNamespaces = {"foo": nil, "bar": nil}` if one would like to watch only namespaces foo and bar, or would this be achieved by some other means, outside of cache options?

tsaarni: Great! If possible, it would be nice to have small example config for this too in the design doc. It might not be immediately clear that this can be achieved by `DefaultNamespaces`.

alvaroaleman: Makes sense, added one

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+        "operator": {},
+    }},
+  },
+}
+```
+
+### Cache ConfigMaps in all namespaces without selector, but have a selector for the `operator` Namespace
+
+```
+Config{
```
sbueringer, line 178:
```suggestion
cache.Options{
```
nit (similar below)

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+        cache.AllNamespaces: nil,                   // No selector for all namespaces...
+        "operator": {LabelSelector: labelSelector}, // except for the operator namespace
+      },
+    },
+  },
+}
+```
+
+
+### Only cache the `operator` namespace except for Deployments
```
sbueringer, line 191: Q: This caches the operator namespace + Deployments in all namespaces, correct?

Can we maybe slightly modify the title to reflect this better. I thought that we don't cache Deployments

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  // one configured.
+  DefaultFieldSelector fields.Selector
+
+  // DefaultUnsafeDisableDeepCopy is the default for UnsafeDisableDeepCopy
+  // for everything that doesn't specify this.
+  DefaultUnsafeDisableDeepCopy *bool
+
+  // DefaultTransform will be used as transform for all object types
+  // unless they have a more specific transform set in ByObject.
+  DefaultTransform toolscache.TransformFunc
```
sbueringer, line 112: Apologies if we discussed this before (and no strong opinion). Would it make sense to combine those 4 fields in a `Default *Config` field?

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  // This wil then include all namespaces that do not have a more specific
+  // setting.
+  //
+  // A nil map allows to default this to the cache's DefaultNamespaces setting.
+  // An empty map prevents this.
+  //
+  // This must be unset for cluster-scoped objects.
+  Namespaces map[string]*CacheSetting
+
+  // CacheSettings will be used for cluster-scoped objects and to default
```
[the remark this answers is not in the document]
sbueringer: Trying to summarize some of my comments as concise as possible
```go
type Config struct {
  LabelSelector labels.Selector
  FieldSelector fields.Selector
  Transform     toolscache.TransformFunc
  UnsafeDisableDeepCopy *bool
}
type ByObject struct {
  Namespaces map[string]*Config
  ClusterWide *Config
  Default *Config
}
type Options struct {
  ByObject map[client.Object]*ByObject
  DefaultNamespaces map[string]*Config
  DefaultClusterWide *Config  
  Default *Config
  ...
}
```
As suggested defaulting would work "bottom up" from the most precise config:
* Namespaced object:
  * Options.ByObject[obj].Namespaces[ns].LabelSelector
  * Options.ByObject[obj].Default.LabelSelector
  * Options.DefaultNamespaces[ns].LabelSelector
  * Options.Default.LabelSelector
* Cluster-wide object:
  * Options.ByObject[obj].ClusterWide.LabelSelector
  * Options.ByObject[obj].Default.LabelSelector
  * Options.DefaultClusterWide.LabelSelector
  * Options.Default.LabelSelector

While `nil` always means default to the upper layers and empty means the "zero value we define" (e.g. for Label Selector no selectors aka match everything)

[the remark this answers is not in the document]
sbueringer: So basically the comments above can be resolved.

Just to confirm, this is correct, right?

> While nil always means default to the upper layers and empty means the "default value of the field itself" (e.g. for Label Selector no selectors aka match everything)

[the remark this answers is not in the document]
alvaroaleman: > Just to confirm, this is correct, right?

Yes

In general, the defaulting order for namespaced objects is byObject.Namespaces[namespace or namespace all] -> byObject.CacheSettings -> DefaultNamespaces[namespace or namespace all] -> DefaultFieldSelector/LabelSelector etc

For non-namespaced, the Namespaces part is omitted so byObject.CacheSettings -> DefaultFieldSelector/LabelSelector etc

designs/cache_options.md
alvaroaleman: Updated, WDYT?

designs/cache_options.md
alvaroaleman: There was a bit of discussion around that, but we decided its more ergonomic to have a top-level field (`cache.Options{DefaultLabelSelector: myLabelSelector}` vs `cache.Options{Defaults: &cache.CacheSettings{LabelSelector: myLabelSelector}}`)

designs/cache_options.md
sbueringer: Sounds good. Out of scope of this PR but might be useful to provide a no-op transform for simplicity

designs/cache_options.md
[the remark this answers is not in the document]
sbueringer: Sounds good!

designs/cache_options.md
sbueringer: Alright!

designs/cache_options.md
sbueringer: Perfect!

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+
+type ByObject struct {
+  // Namespaces maps a namespace name to cache setting. If set, only the
+  // namespaces in this map will be cached.
+  //
+  // Settings in the map value that are unset because either the value as a
+  // whole is nil or because the specific setting is nil will be defaulted.
+  // Use an empty value for the specific setting to prevent that.
+  //
+  // It is possible to have specific CacheSettings for just some namespaces
```
sbueringer, line 60: I think with the change to the Config struct every other occurence of "CacheSettings" in this doc has to be changed as well

alvaroaleman: Updated, PTAL

sbueringer:
Thank you very much!

/lgtm

/hold
in case you want to wait for more feedback

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  // It is possible to have specific Config for just some namespaces
+  // but cache all namespaces by using the AllNamespaces const as the map key.
+  // This wil then include all namespaces that do not have a more specific
+  // setting.
+  //
+  // A nil map allows to default this to the cache's DefaultNamespaces setting.
+  // An empty map prevents this.
+  //
+  // This must be unset for cluster-scoped objects.
+  Namespaces map[string]*Config
```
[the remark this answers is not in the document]
irbekrm: > What is cached is what is being watched, so both.

At the moment the calls to the kube apiserver to LIST resources are different (this might be in client-go, not sure) depending on whether the whole cache is scoped to a namespace or whether the namespace selector is applied as a field selector.
Because of that (I think) a namespace-scoped cache allows us to also scope RBAC to that namespace whereas the selector does not.

We have a use case where we want to watch Secrets in namespace X and Configmaps in all namespaces. We create two caches for this because although we could use the existing cache options to apply namespace filter to Secrets, only namespaced cache allows us to scope RBAC down so that users only need to give our controller permissions to list Secrets in that one namespace.

For us, it would be quite useful if we were able to use this `Namespaces` option in such a way that the filter is also reflected in RBAC.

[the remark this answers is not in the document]
alvaroaleman: Yes, it will be. Something like your use-case is what we want to support. Your config would look something like:
```
cache.Options{
ByObject: map[client.Object]*cache.ByObject{
  &corev1.Secrets{}: *cache.ByObject{
    Namespaces: map[string]*cache.Config{"the-namespace-for-secrets": nil},
  }
}
```

The above config would watch secrets in one namespace and everything else in all namespaces. That also applies to rbac.

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  // one configured.
+  DefaultFieldSelector fields.Selector
+
+  // DefaultUnsafeDisableDeepCopy is the default for UnsafeDisableDeepCopy
+  // for everything that doesn't specify this.
+  DefaultUnsafeDisableDeepCopy *bool
+
+  // DefaultTransform will be used as transform for all object types
+  // unless they have a more specific transform set in ByObject.
+  DefaultTransform toolscache.TransformFunc
```
vincepri, line 112: Would it be better or worse to drop the `Default` prefix? It seems redundant, given that the `ByObject` seems clear that it overrides these imo

alvaroaleman: I like it because it clarifies that this is the last layer of defaulting

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  //
+  // It gets defaulted from the cache'sDefaultLabelSelector, DefaultFieldSelector,
+  // DefaultUnsafeDisableDeepCopy and DefaultTransform.
+  Config *Config
+}
+
+type Options struct {
+  // ByObject specifies per-object cache settings. If unset for a given
+  // object, this will fall through to Default* settings.
+  ByObject map[client.Object]*ByObject
```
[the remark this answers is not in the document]
alvaroaleman: Not a super strong opinion here, the reason I chose the pointer type is in some contexts an empty thing has a meaning (for example an empty labelSelector is a match-everything selector) so I thought by not requireding ppl to do a `map[string]ByObject{"namespace":{}}` we can avoid a bit of confusion

designs/cache_options.md
```diff
@@ -0,0 +1,226 @@
...
+  //
+  // This must be unset for cluster-scoped objects.
+  Namespaces map[string]*Config
+
+  // Config will be used for cluster-scoped objects and to default
+  // Config in the Namespaces field.
+  //
+  // It gets defaulted from the cache'sDefaultLabelSelector, DefaultFieldSelector,
+  // DefaultUnsafeDisableDeepCopy and DefaultTransform.
+  Config *Config
```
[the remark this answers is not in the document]
alvaroaleman: I think that might be confusing because with an override, ppl expect the value in there to be the final setting but:
* Its not the top-most setting for namespaced resources
* It gets defaulted itself

PR #2265, opened 2023-04-07
:sparkles: Use generics for predicates

### LINKED ISSUE #2214: Evaluate using generics for source/predicate/handler chain
Predicates currently take a `client.Object` and return a `bool`. With generics, we might be able to have typed generics, which would let ppl avoid a type assertion inside their predicate when they want to access anything other than metadata.

We currently type to `client.Object`, but this is both too much in the case of a `source.Channel` that could take literally everything that might not actually be a `client.Object` and too little in the case of everything else, as we lose the type information.

nikola-jokic:
Fixes #2214


nikola-jokic:
Hey @alvaroaleman :wave: 

I hope this is what you meant by using generics :relaxed:

pkg/predicate/predicate.go
```diff
@@ -98,22 +98,29 @@ func (p Funcs) Generic(e event.GenericEvent) bool {
...
+
 // NewPredicateFuncs returns a predicate funcs that applies the given filter function
 // on CREATE, UPDATE, DELETE and GENERIC events. For UPDATE events, the filter is applied
 // to the new object.
-func NewPredicateFuncs(filter func(object client.Object) bool) Funcs {
+func NewPredicateFuncs[T client.Object](filter FilterFunc[T]) Funcs {
 	return Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return filter(e.Object)
+			v := e.Object.(T)
```
alvaroaleman, line 110: We still have the type assertion here and its possible to misuse this by constructing a predicate for the wrong type. What I'd want is the whole chain of source/predicate/handler using generics. I'll update the issue to clarify that

nikola-jokic: On it :smile:

nikola-jokic: #### Update

The only thing I'm having trouble with is builder
Right now, builder can watch multiple resources and I can't really instantiate many watches with different type parameters using generics without providing a top level function and potentially changing the interface of the `Watch` method
I'm looking into ways of doing it, so the change is as minimal as possible

alvaroaleman: I looked into this some time ago. Golang doesn't support generic methods, which is why we can not keep assembling everything in the controller.Watch but instead have to provide predicate+handler in a constructor to the source

pkg/builder/controller.go
```diff
@@ -54,55 +59,246 @@ const (
...
+	watchesInput     []watcher
 	mgr              manager.Manager
-	globalPredicates []predicate.Predicate
+	globalPredicates []predicate.Predicate[client.Object]
 	ctrl             controller.Controller
 	ctrlOptions      controller.Options
 	name             string
 }
 
+func newObjectPredicate[T client.ObjectConstraint](p predicate.Predicate[T]) predicate.Predicate[client.Object] {
```
nikola-jokic, line 73: This is heavily work in progress, but I wanted to push changes to make sure this approach is okay
Since controller is able to watch on multiple resources, the only way I could think of (since we can't use generic methods) is to create a generic function that will convert this generic type to a `client.Object`
That way, the complexity is internal, and we use `client.Object` interface internally. The caller still can leverage generics to use those types

The only thing is that the project needs to be go version 1.20, because they added interfaces to be comparable in that version

nikola-jokic: This does not work now, there are many tests where I'd have to provide types and similar
The controller watch is also something that needs to be changed or provided another method like `controller.WatchType(WithWatch()` that can be a generic function returning wrapped objects

nikola-jokic: Since many types can handle multiple types, wrapping them in interfaces was the only thing that I could think of to provide a nice API... Some types can be improved, but the only thing I wanted to ask is, is there a different approach I should take?

alvaroaleman: This is the issue I was referring to in https://github.com/kubernetes-sigs/controller-runtime/pull/2265#discussion_r1162854237. We have to change how this works in order to accomodate generics. The source needs to get the handler and predicates at its construction time rather than at `Start` time, because using `Start` entails passing it through the `controller` and that would entail that we can only support one type per controller.

nikola-jokic: I see, I'll work on it

PR #2266, opened 2023-04-07
🐛 Support get config inside snap with SNAP_REAL_HOME

HollowMan6:
We can't get the config file from $HOME when running as a snap app, as $HOME is remapped in such case. https://github.com/snapcore/snapd/pull/9189

This PR checks for $SNAP_REAL_HOME when the path contains a `.kube\config` file, and add it to loadingRule path when it is available.

Alternative PR:

I've also created a PR in the upstream, it will work fine if either or both of these two PR gets merged.
- https://github.com/kubernetes/kubernetes/pull/117165

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -152,6 +152,15 @@ func loadConfig(context string) (config *rest.Config, configErr error) {
 		loadingRules.Precedence = append(loadingRules.Precedence, filepath.Join(u.HomeDir, clientcmd.RecommendedHomeDir, clientcmd.RecommendedFileName))
 	}
 
+	// Add $SNAP_REAL_HOME when in snap and the path contains a `.kube\config` file
+	// https://github.com/snapcore/snapd/pull/9189
+	if value, ok := os.LookupEnv("SNAP_REAL_HOME"); ok {
+		location := filepath.Join(value, clientcmd.RecommendedHomeDir, clientcmd.RecommendedFileName)
+		if _, err := os.Stat(location); err == nil {
+			loadingRules.Precedence = append(loadingRules.Precedence, location)
+		}
+	}
+
 	return loadConfigWithContext("", loadingRules, context)
 }
 


alvaroaleman:
Uhm how about snap just using the correct env vars? Patching every possible project with a non-standard env var doesn't really seem like a great thing to do

HollowMan6:
Hi! It's not something related to snap's bug. In a snap package, `$HOME` is re-written by snapd so that each snap appears to have a dedicated home directory for some sort of isolation (e.g. /home/$USER/snap/<some_app>/$SNAP_VERSION), but the kubeconfig file is still in the real home directory, and the real home directory is "$SNAP_REAL_HOME/.kube/config" in this case.

I think this PR would be particularly useful for those possible project that wants to get packed as a snap package and be available in the snap store: https://snapcraft.io/store

alvaroaleman:
But shouldn't snap then find some way to overlay the re-written home with the real home or vice versa? It seems a much better use of efforts to try to solve this in snap instead of adjusting literally every possible project that might access something from $HOME

HollowMan6:
I'm not a maintainer of snapscaft, and I don't know why they don't do it like you said but choose remapping instead, maybe they have some security concerns, as they are actually limiting the permissions for using files under home dir https://snapcraft.io/docs/home-interface.

I'm just working on the kubescape k8s-interface and they rely on getting config files from your SDK https://github.com/kubescape/k8s-interface/issues/46 Now I'm replying a patch to get the kubescape snap package working https://github.com/HollowMan6/packaging/blob/main/snap_homedir.patch

If you don't like this PR anyway, we can also wait to see if your upstream will accept this PR:
- https://github.com/kubernetes/kubernetes/pull/117165

Actually some SDK has already support the SNAP_REAL_HOME long time ago, like the openstacksdk https://opendev.org/openstack/openstacksdk/commit/78dd1a73d8ff063612560a9882b26bffa5f9e51a?style=unified&whitespace=

alvaroaleman:
Yeah, please don't get me wrong, I realize that you are only the messenger here.

It doesn't make sense to me why exposing the same file under a different location would be any different from a security standpoint, that is at most making it more obscure but not more secure. I just think that patching literally any project that has config under $HOME to fall back to a non-standard snap-only dir is not a good use of ppls time and that snap should instead find a way to solve this.

HollowMan6:
Okay, I will open a issue at their side for them to discuss about it.

HollowMan6:
Issue opened at https://forum.snapcraft.io/t/feature-request-stop-using-snap-real-home-to-visit-real-home-files/34683

alvaroaleman:
Following up on upstream

PR #2275, opened 2023-04-14
🌱 Make metrics and health probe servers be Runnables

zqzten:
Because they should be. [Backgrounds here](https://github.com/kubernetes-sigs/controller-runtime/pull/1943#discussion_r1038896709).


alvaroaleman (approved):
Thanks!

PR #2278, opened 2023-04-15
✨ envtest: Add Environment.KubeConfig field

kaovilai:
Pulling function from internal https://github.com/kubernetes-sigs/controller-runtime/blob/b826f390522092f54fbea9ff18a97ddde4a828b5/pkg/internal/testing/controlplane/kubectl.go#L42-L81

This is useful for creating kubeconfig file in order to pass filename to KUBECONFIG env which can be helpful in writing [unit tests](https://github.com/vmware-tanzu/velero/blob/a51cc56b43b34ac379a2a5d9908b0e0427c25a34/pkg/cmd/cli/install/install_test.go#L84-L97) without having to break apart some of the current client code to accept a rest.Config

Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

Used in
- https://github.com/vmware-tanzu/velero/pull/6152
- https://github.com/vmware-tanzu/velero/pull/6155

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -126,6 +126,10 @@ type Environment struct {
 	// loading.
 	Config *rest.Config
 
+	// KubeConfig provides []byte of a kubeconfig file to talk to the apiserver
+	// It's automatically populated if not set based on the `Config`
+	KubeConfig []byte
+
 	// CRDInstallOptions are the options for installing CRDs.
 	CRDInstallOptions CRDInstallOptions
 
@@ -289,6 +293,14 @@ func (te *Environment) Start() (*rest.Config, error) {
 			return te.Config, fmt.Errorf("unable to provision admin user: %w", err)
 		}
 		te.Config = adminUser.Config()
+	}
+
+	if len(te.KubeConfig) == 0 {
+		var err error
+		te.KubeConfig, err = controlplane.KubeConfigFromREST(te.Config)
+		if err != nil {
+			return nil, fmt.Errorf("unable to set KubeConfig field: %w", err)
+		}
 	}
 
 	// Set the default scheme if nil.
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -28,7 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
-
+	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -680,6 +680,14 @@ var _ = Describe("Test", func() {
 			)
 			Expect(err).NotTo(HaveOccurred())
 		})
+	})
+
+	It("should set a working KubeConfig", func() {
+		kubeconfigRESTConfig, err := clientcmd.RESTConfigFromKubeConfig(env.KubeConfig)
+		Expect(err).ToNot(HaveOccurred())
+		kubeconfigClient, err := client.New(kubeconfigRESTConfig, client.Options{Scheme: s})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(kubeconfigClient.List(context.Background(), &apiextensionsv1.CustomResourceDefinitionList{})).To(Succeed())
 	})
 
 	It("should update CRDs if already present in the cluster", func() {


kaovilai:
Need to `gofmt -s`

kaovilai:
bump

pkg/envtest/client/client.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	envtestName = "envtest"
+)
+
+// KubeConfigFromREST reverse-engineers a kubeconfig file from a rest.Config.
+// The options are tailored towards the rest.Configs we generate, so they're
+// not broadly applicable.
+//
+// This is not intended to be exposed outside envtest for the above reasons.
```
sbueringer, line 20: Question: The idea is to use this function only to convert envtest.Environment.Config to a kubeconfig, correct?

Could we instead just write a envtest.Environment.KubeConfig field when we set envtest.Environment.Config?

kaovilai: Yes

kaovilai: Yes

sbueringer: Then let's do that. Then folks can simply consume env.KubeConfig and we don't have to expose a func to rest.Config to kubeconfig

kaovilai: Done.

pkg/envtest/server.go
```diff
@@ -126,6 +126,12 @@ type Environment struct {
 	// loading.
```
sbueringer, line 126: Let's update the PR title & commit message

pkg/envtest/server.go
```diff
@@ -126,6 +126,12 @@ type Environment struct {
...
 	Config *rest.Config
 
+	// KubeConfig provides []byte of a kubeconfig file to talk to the apiserver
+	// It's automatically
+	// populated if not set using the standard controller-runtime config
+	// loading.
```
sbueringer, line 132:
```suggestion
	// KubeConfig provides []byte of a kubeconfig file to talk to the apiserver
	// It's automatically populated if not set using the standard controller-runtime
	// config loading.
```

pkg/envtest/server.go
```diff
@@ -221,6 +227,10 @@ func (te *Environment) Start() (*rest.Config, error) {
 			if err != nil {
 				return nil, fmt.Errorf("unable to get configuration for existing cluster: %w", err)
 			}
+			te.KubeConfig, err = controlplane.KubeConfigFromREST(te.Config)
+			if err != nil {
+				return nil, fmt.Errorf("unable to convert configuration to kubeconfig: %w", err)
+			}
```
sbueringer, line 233:
```suggestion

			if len(te.KubeConfig) == 0 {
				te.KubeConfig, err = controlplane.KubeConfigFromREST(te.Config)
				if err != nil {
					return nil, fmt.Errorf("unable to convert configuration to kubeconfig: %w", err)
				}
			}
```
The godoc says this is only populated if not set (which I think makes sense). Let's also implement it that way :)

sbueringer: Let's move this down below the if (~ l.305). So that we also get the KubeConfig in the case where we don't use an existing cluster

pkg/envtest/server.go
```diff
@@ -221,6 +227,10 @@ func (te *Environment) Start() (*rest.Config, error) {
...
 		}
 	} else {
 		apiServer := te.ControlPlane.GetAPIServer()
```
sbueringer, line 236: Let's add some test coverage. I would suggest envtest_test.go. Maybe simply at the end of BeforeEach (~l.55).

Let's take env.KubeConfig convert it via `clientcmd.RESTConfigFromKubeConfig` create an additional client from it and run a simple request (e.g. list CRD) to ensure the client works

kaovilai: Makes sense.. this was old PR and no longer priority but may circle back . Anyone feel free to take over / push to this branch etc.

sbueringer: No problem. I'll do that

pkg/envtest/server.go
```diff
@@ -126,6 +126,11 @@ type Environment struct {
 	// loading.
 	Config *rest.Config
 
+	// KubeConfig provides []byte of a kubeconfig file to talk to the apiserver
+	// It's automatically populated if not set using the standard controller-runtime
+	// config loading.
```
alvaroaleman, line 131:
```suggestion
	// KubeConfig provides []byte of a kubeconfig file to talk to the apiserver
	// It's automatically populated if not set based on the `Config`
```

alvaroaleman (approved):

PR #2189, opened 2023-02-13
:sparkles: upgrade to k8s 0.27.1

mythi:
This PR moves k8s dependencies to 0.27.1 and reworks things to make the tests/linter happy.

k8s.io/apimachinery deprecated legacy Poll methods and this
triggers golangi-lint staticheck errors.
    
Rework all the poll logic to move to the preferred methods. 

Details: https://github.com/kubernetes/apimachinery/commit/831cf05a7c4f839939a6c5e1e8bee238ddc6baac

Additionally, NamespacedName type now implements MarshalLog() which
replaces the functionality in kube_helpers.go. See k/k [#117238.](https://github.com/kubernetes/kubernetes/pull/117238).

PR #2281, opened 2023-04-18
✨  Upgrade k8s dependency from 1.26 to 1.27.1

camilamacedo86:
## Description

Upgrade k8s versions from  1.26 to 1.27.1 and add TODO to fix deprecations in a follow-up (https://github.com/kubernetes-sigs/controller-runtime/pull/2189).


vincepri:
This is already in progress https://github.com/kubernetes-sigs/controller-runtime/pull/2189

/hold

camilamacedo86:
HI @vincepri, 

> This is already in progress https://github.com/kubernetes-sigs/controller-runtime/pull/2189
 
Thank you for sharing it. I notice that the above PR is been open since Feb and it still not passing yet in the tests. We are blocked from moving forward without a new controller-runtime release using k8s 1.27

Therefore, I found it easier to indeed share what is missing with @mythi just by creating a new PR with a proposed solution where we just bump k8s and do not try to sort out the deprecations. So, @mythi  can still be working in its PR just to address the deprecations. Also, note that I added a co-author to provide the credit. So we can move forward within and be unblocked. 

@mythi are you OK with that? 
Could we move forward with this one to unblock us?

camilamacedo86:
Closing in favour of https://github.com/kubernetes-sigs/controller-runtime/pull/2189

We cannot change the tests here. The fact that name/namespace is not using a lower case is a bug, and we will need the bug fixes of https://github.com/kubernetes/kubernetes/pull/117238 released first. 👍

PR #2285, opened 2023-04-22
🌱 Proposal for dynamic informer cache

maxsmythe:
This PR shows how Gatekeeper has forked controller runtime to support the dynamic addition/removal of informers.

Happy to flesh this out if people are interested. Not sure what the correct licensing actions are for moving code across CNCF projects.

This is related to #2159 and #1884

Basically, something like this PR will be necessary to clean up informers once no more controllers are using them.

In the interim, having this code would be helpful to Gatekeeper by eliminating our need to maintain a fork, which has been fairly labor intensive. It also may give other members of the community a way to meet their needs for dynamic watches while waiting for the dynamic controller/reference counting approach to be implemented.

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -190,6 +190,17 @@ func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.Grou
 	return ic.Informers.Get(ctx, gvk, obj, &internal.GetOptions{})
 }
 
+// RemoveInformer deactivates and removes the informer from the cache.
+func (ic *informerCache) RemoveInformer(_ context.Context, obj client.Object) error {
+	gvk, err := apiutil.GVKForObject(obj, ic.scheme)
+	if err != nil {
+		return err
+	}
+
+	ic.Informers.Remove(gvk, obj)
+	return nil
+}
+
 // NeedLeaderElection implements the LeaderElectionRunnable interface
 // to indicate that this can be started without requiring the leader lock.
 func (ic *informerCache) NeedLeaderElection() bool {
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/internal/syncs"
 )
 
 // InformersOpts configures an InformerMap.
@@ -86,6 +87,20 @@ type Cache struct {
 
 	// CacheReader wraps Informer and implements the CacheReader interface for a single type
 	Reader CacheReader
+
+	// Stop can be used to stop this individual informer.
+	stop chan struct{}
+}
+
+// Start starts the informer managed by a MapEntry.
+// Blocks until the informer stops. The informer can be stopped
+// either individually (via the entry's stop channel) or globally
+// via the provided stop argument.
+func (c *Cache) Start(stop <-chan struct{}) {
+	// Stop on either the whole map stopping or just this informer being removed.
+	internalStop, cancel := syncs.MergeChans(stop, c.stop)
+	defer cancel()
+	c.Informer.Run(internalStop)
 }
 
 type tracker struct {
@@ -173,13 +188,13 @@ func (ip *Informers) Start(ctx context.Context) error {
 
 		// Start each informer
 		for _, i := range ip.tracker.Structured {
-			ip.startInformerLocked(i.Informer)
+			ip.startInformerLocked(i)
 		}
 		for _, i := range ip.tracker.Unstructured {
-			ip.startInformerLocked(i.Informer)
+			ip.startInformerLocked(i)
 		}
 		for _, i := range ip.tracker.Metadata {
-			ip.startInformerLocked(i.Informer)
+			ip.startInformerLocked(i)
 		}
 
 		// Set started to true so we immediately start any informers added later.
@@ -194,7 +209,7 @@ func (ip *Informers) Start(ctx context.Context) error {
 	return nil
 }
 
-func (ip *Informers) startInformerLocked(informer cache.SharedIndexInformer) {
+func (ip *Informers) startInformerLocked(cacheEntry *Cache) {
 	// Don't start the informer in case we are already waiting for the items in
 	// the waitGroup to finish, since waitGroups don't support waiting and adding
 	// at the same time.
@@ -205,7 +220,7 @@ func (ip *Informers) startInformerLocked(informer cache.SharedIndexInformer) {
 	ip.waitGroup.Add(1)
 	go func() {
 		defer ip.waitGroup.Done()
-		informer.Run(ip.ctx.Done())
+		cacheEntry.Start(ip.ctx.Done())
 	}()
 }
 
@@ -281,6 +296,21 @@ func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj r
 	return started, i, nil
 }
 
+// Remove removes an informer entry and stops it if it was running.
+func (ip *Informers) Remove(gvk schema.GroupVersionKind, obj runtime.Object) {
+	ip.mu.Lock()
+	defer ip.mu.Unlock()
+
+	informerMap := ip.informersByType(obj)
+
+	entry, ok := informerMap[gvk]
+	if !ok {
+		return
+	}
+	close(entry.stop)
+	delete(informerMap, gvk)
+}
+
 func (ip *Informers) informersByType(obj runtime.Object) map[schema.GroupVersionKind]*Cache {
 	switch obj.(type) {
 	case runtime.Unstructured:
@@ -342,13 +372,14 @@ func (ip *Informers) addInformerToMap(gvk schema.GroupVersionKind, obj runtime.O
 			scopeName:        mapping.Scope.Name(),
 			disableDeepCopy:  ip.unsafeDisableDeepCopy,
 		},
+		stop: make(chan struct{}),
 	}
 	ip.informersByType(obj)[gvk] = i
 
 	// Start the informer in case the InformersMap has started, otherwise it will be
 	// started when the InformersMap starts.
 	if ip.started {
-		ip.startInformerLocked(i.Informer)
+		ip.startInformerLocked(i)
 	}
 	return i, ip.started, nil
 }
--- a/pkg/internal/syncs/syncs.go
+++ b/pkg/internal/syncs/syncs.go
@@ -1,0 +1,38 @@
+package syncs
+
+import (
+	"context"
+	"reflect"
+	"sync"
+)
+
+// MergeChans returns a channel that is closed when any of the input channels are signaled.
+// The caller must call the returned CancelFunc to ensure no resources are leaked.
+func MergeChans[T any](chans ...<-chan T) (<-chan T, context.CancelFunc) {
+	var once sync.Once
+	out := make(chan T)
+	cancel := make(chan T)
+	cancelFunc := func() {
+		once.Do(func() {
+			close(cancel)
+		})
+		<-out
+	}
+	cases := make([]reflect.SelectCase, len(chans)+1)
+	for i := range chans {
+		cases[i] = reflect.SelectCase{
+			Dir:  reflect.SelectRecv,
+			Chan: reflect.ValueOf(chans[i]),
+		}
+	}
+	cases[len(cases)-1] = reflect.SelectCase{
+		Dir:  reflect.SelectRecv,
+		Chan: reflect.ValueOf(cancel),
+	}
+	go func() {
+		defer close(out)
+		_, _, _ = reflect.Select(cases)
+	}()
+
+	return out, cancelFunc
+}
--- a/pkg/internal/syncs/syncs_test.go
+++ b/pkg/internal/syncs/syncs_test.go
@@ -1,0 +1,107 @@
+package syncs
+
+import (
+	"testing"
+	"time"
+
+	// This appears to be needed so that the prow test runner won't fail.
+	_ "github.com/onsi/ginkgo/v2"
+	_ "github.com/onsi/gomega"
+)
+
+func TestMergeChans(t *testing.T) {
+	tests := []struct {
+		name   string
+		count  int
+		signal int
+	}{
+		{
+			name:   "single channel, close 0",
+			count:  1,
+			signal: 0,
+		},
+		{
+			name:   "double channel, close 0",
+			count:  2,
+			signal: 0,
+		},
+		{
+			name:   "five channel, close 0",
+			count:  5,
+			signal: 0,
+		},
+		{
+			name:   "five channel, close 1",
+			count:  5,
+			signal: 1,
+		},
+		{
+			name:   "five channel, close 2",
+			count:  5,
+			signal: 2,
+		},
+		{
+			name:   "five channel, close 3",
+			count:  5,
+			signal: 3,
+		},
+		{
+			name:   "five channel, close 4",
+			count:  5,
+			signal: 4,
+		},
+		{
+			name:   "single channel, cancel",
+			count:  1,
+			signal: -1,
+		},
+		{
+			name:   "double channel, cancel",
+			count:  2,
+			signal: -1,
+		},
+		{
+			name:   "five channel, cancel",
+			count:  5,
+			signal: -1,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if callAndClose(test.count, test.signal, 1) {
+				t.Error("timeout before merged channel closed")
+			}
+		})
+	}
+}
+
+func callAndClose(numChans, signalChan, timeoutSeconds int) bool {
+	chans := make([]chan struct{}, numChans)
+	readOnlyChans := make([]<-chan struct{}, numChans)
+	for i := range chans {
+		chans[i] = make(chan struct{})
+		readOnlyChans[i] = chans[i]
+	}
+	defer func() {
+		for i := range chans {
+			close(chans[i])
+		}
+	}()
+
+	merged, cancel := MergeChans(readOnlyChans...)
+	defer cancel()
+
+	timer := time.NewTimer(time.Duration(timeoutSeconds) * time.Second)
+
+	if signalChan >= 0 {
+		chans[signalChan] <- struct{}{}
+	} else {
+		cancel()
+	}
+	select {
+	case <-merged:
+		return false
+	case <-timer.C:
+		return true
+	}
+}
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -81,6 +81,9 @@ type Informers interface {
 	// of the underlying object.
 	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, opts ...InformerGetOption) (Informer, error)
 
+	// RemoveInformer removes an informer entry and stops it if it was running.
+	RemoveInformer(ctx context.Context, obj client.Object) error
+
 	// Start runs all the informers known to this cache until the context is closed.
 	// It blocks.
 	Start(ctx context.Context) error
@@ -119,6 +122,8 @@ type Informer interface {
 
 	// HasSynced return true if the informers underlying store has synced.
 	HasSynced() bool
+	// IsStopped returns true if the informer has been stopped.
+	IsStopped() bool
 }
 
 // Options are the optional arguments for creating a new Cache object.
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -1758,6 +1758,42 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					By("verifying the object is received on the channel")
 					Eventually(out).Should(Receive(Equal(pod)))
 				})
+				It("should be able to stop and restart informers", func() {
+					By("getting a shared index informer for a pod")
+					pod := &corev1.Pod{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "informer-obj",
+							Namespace: "default",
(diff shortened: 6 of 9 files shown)


ritazh:
@vincepri can you please help take a look at this?

braghettos:
We have also a use case where a feature like this one could be really helpful: we are going to generate at runtime CRDs and we would like to watch all the CRs based on the created CRDs.

vincepri:
In the queue, should be done reviewing by EOW tomorrow

vincepri:
@maxsmythe We should probably add some tests before merging as well

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,24 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
 	return i.Informer, err
 }
 
+// GetInformerNonBlocking returns the informer for the obj without waiting for its cache to sync.
+func (ic *informerCache) GetInformerNonBlocking(ctx context.Context, obj client.Object) (Informer, error) {
```
vincepri, line 157: I might be missing something, where is the new code going to be used and how?

pkg/cache/internal/informers.go
```diff
@@ -558,3 +592,50 @@ func restrictNamespaceBySelector(namespaceOpt string, s Selector) string {
 	}
 	return ""
 }
+
+// eitherChan returns a channel that is closed when either of the input channels are signaled.
+// The caller must call the returned CancelFunc to ensure no resources are leaked.
+func eitherChan(a, b <-chan struct{}) (<-chan struct{}, context.CancelFunc) {
```
vincepri, line 598: These utilities might be useful elsewhere as well, could we:
- Move them under a new package like `pkg/internal/syncs`
- Use generics so we can reuse for other types of channels as well
- Add some tests

pkg/cache/internal/informers.go
```diff
@@ -558,3 +592,50 @@ func restrictNamespaceBySelector(namespaceOpt string, s Selector) string {
...
+		case <-cancel:
+		}
+	}()
+
+	return out, cancelFunc
+}
+
+// mergeChan returns a channel that is closed when any of the input channels are signaled.
+// The caller must call the returned CancelFunc to ensure no resources are leaked.
+func mergeChan(a, b, c <-chan struct{}) (<-chan struct{}, context.CancelFunc) {
```
vincepri, line 621: Same as above

pkg/cache/internal/informers.go
maxsmythe: Absolutely!

maxsmythe:
Thanks for the review!

Regarding tests, I was holding off until I got signal that there was interest in the idea. It sounds like that hurdle is cleared?

pkg/cache/informer_cache.go
vincepri: Got it, thanks for the example. I think overall the approach seems good, there might be more feedback in terms of API changes and whatnot.

I assume to actually use these methods we'll require changes to the exposed interface?

For the non-blocking scenario, could we opt into an option instead of adding a new method?

vincepri: For the controller runtime project specifically, it might be worth discussing how we'll use these facilities to shutdown informers at runtime

cc @alvaroaleman

maxsmythe: Totally cool with adding an option. Are you thinking of:

* adding functional options (e.g. `GetInformer(ctx, obj, Blocking(false), SomeOtherOption())`)
* Adding a straight boolean: `GetInformer(ctx, obj, false)`
* An options object: `GetInformer(ctx, obj, Options{Blocking: false})`

?

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,24 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
...
+	gvk, err := apiutil.GVKForObject(obj, ic.scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use a canceled context to signal non-blocking
+	canceledCtx, cancel := context.WithCancel(context.Background())
+	cancel()
```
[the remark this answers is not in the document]
maxsmythe: +1, waiting for feedback on the shape that should take in a different comment:

https://github.com/kubernetes-sigs/controller-runtime/pull/2285#discussion_r1186564862

maxsmythe:
@vincepri Any feedback on the correct thing to do vis-a-vis licensing, since this code is copied from the Gatekeeper project (which is Apache licensed and CNCF-owned)?

Not sure if this requires a NOTICE file, or a citation in the comment or similar.

pkg/cache/informer_cache.go
vincepri: Probably the first option would be more in line with what we have elsewhere, I was thinking something pretty simple like:

```golang
// Informers knows how to create or fetch informers for different
// group-version-kinds, and add indices to those informers.  It's safe to call
// GetInformer from multiple threads.
type Informers interface {
	// GetInformer fetches or constructs an informer for the given object that corresponds to a single
	// API kind and resource.
	GetInformer(ctx context.Context, obj client.Object, opts ...InformerOption) (Informer, error)

	// GetInformerForKind is similar to GetInformer, except that it takes a group-version-kind, instead
	// of the underlying object.
	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, opts ...InformerOption) (Informer, error)
[...]
}
```

And ultimately define the option such as `NonBlocking()` or similar

vincepri: The above is a breaking change, but it provides callers with a backward compatible approach, and allows us to extend the option further in the future if needed

vincepri:
Where is the code copied from? In general, I have no idea how that works, but I'd expect code for this specific feature to go be net-new in controller runtime, especially with the changes proposed above

pkg/cache/informer_cache.go
maxsmythe: Functional options it is!

braghettos:
@maxsmythe @vincepri any updates on the proposed PR? This looks so useful for many use cases!

maxsmythe:
@braghettos Sorry, I got distracted. Just pushed a first pass responding to feedback now.

@vincepri I took a stab at addressing your comments. LMK what you think. Also let me know what, if any, additional tests you'd like to see.

Tests pass locally, not sure why there is a failure currently.

maxsmythe:
Unit tests passed on retest. Still haven't been able to get them to fail locally. Rare flake?

braghettos:
Any updates on this PR?

stevekuznetsov:
Kubernetes upstream added support for dynamic informer lifecycle handling for the shared informer factory in 1.26. Any chances that we could lean on that?

maxsmythe:
@stevekuznetsov This is a step towards being able to lean on that.

For context, here is DynamicInformerFactory's interface:

https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/staging/src/k8s.io/client-go/dynamic/dynamicinformer/interface.go#L25-L49

Which returns a GenericInformer:

https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/staging/src/k8s.io/client-go/informers/generic.go#L74-L79

Which returns a SharedIndexInformer:

https://github.com/kubernetes/client-go/blob/2a5f18df73b70cb85c26a3785b06162f3d513cf5/tools/cache/shared_informer.go#L161-L164

Which allows for the dynamic addition/removal of listeners and whose Run() method accepts a channel to signal the watch should be torn down. I believe this is the dynamic lifecycle handling you're talking about? Let me know if not.

What this PR does is provide a mechanism for sending that teardown signal, but takes no stance on how listeners should be added/removed.

The idea is that, eventually, controller-runtime will leverage the lifecycle methods for adding/removing listeners. Once it does, it will need some kind of method for tearing down watches to avoid memory leaks (I believe reference counting was an idea that had been floated). This PR adds the building block for tearing down watches that will eventually be used.

In the interim, users of controller-runtime who need dynamic watches and are willing to handle engaging with dynamic watches directly (Gatekeeper is one of these users) can use controller runtime without forking it while the community finds the best way forward for a more managed approach.

@vincepri LMK if you have any feedback on the PR. I'm blocked on feedback. Also, WRT the license issue, LMK if you'd like me to reach out to CNCF for this. I believe they have people to give clarity for these kinds of questions.

stevekuznetsov:
Awesome, sorry I did not dig in deeply enough to catch that. That approach sounds awesome :)

maxsmythe:
Glad to add clarity!

vincepri:
In queue, need some more time to look through the changes early next week 😄

pkg/internal/syncs/syncs_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func TestMergeChans(t *testing.T) {
+	tests := []struct {
+		name   string
+		count  int
+		signal int
+	}{
+		{
+			name:   "single channel",
```
sbueringer, line 19:
```suggestion
			name:   "single channel, close 0",
```
nit

Same below

sbueringer:
Overall looks good to me +/- the linter/compile issues

pkg/cache/internal/informers.go
```diff
@@ -300,6 +339,21 @@ func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj r
 	return started, i, nil
 }
 
+// Remove removes an informer entry and stops it if it was running.
+func (ip *Informers) Remove(gvk schema.GroupVersionKind, obj runtime.Object) {
```
sbueringer, line 343: Who  can use this func?

pkg/cache/internal/informers.go
```diff
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
```
sbueringer, line 37: I think unit tests on some level would be good to get coverage for the new functionality

pkg/internal/syncs/syncs.go
```diff
@@ -0,0 +1,37 @@
...
+
+// MergeChans returns a channel that is closed when any of the input channels are signaled.
+// The caller must call the returned CancelFunc to ensure no resources are leaked.
+func MergeChans[T any](chans ...<-chan T) (<-chan T, context.CancelFunc) {
+	var once sync.Once
+	out := make(chan T)
+	cancel := make(chan T)
+	cancelFunc := func() {
+		once.Do(func() {
+			close(cancel)
```
inteon, line 17: I think we have to wait for out to be canceled here.
Currently `cancel()` can complete without `chan out` being closed.

pkg/cache/internal/informers.go
```diff
@@ -92,6 +107,25 @@ type tracker struct {
 	Metadata     map[schema.GroupVersionKind]*Cache
 }
 
+// GetOptions provides configuration to customize the behavior when
+// getting an informer.
+type GetOptions struct {
+	// Async instructs Get() to return the informer immediately,
+	// without waiting for its cache to sync.
+	Async bool
```
inteon, line 115:
```suggestion
	WaitForInformerSynced bool
```
Its not very clear to me what `Async` means, maybe its just me.

pkg/cache/informer_cache.go
inteon: @vincepri @maxsmythe I don't think the "Blocking" option will be compatible with reference counting the number of calls to AddEventHandler and RemoveEventHandler. If no calls to AddEventHandler have been made yet, we won't start the informer yet. Only once we have one or more `ResourceEventHandlerRegistration`s, we start the informer. So, I would instead opt for removing the `Blocking` option.
I think DynamicSharedInformerFactory took the same approach (since no ctx is passed to ForResource (see https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/staging/src/k8s.io/client-go/dynamic/dynamicinformer/interface.go#L32).

pkg/cache/internal/informers.go
```diff
@@ -92,6 +107,25 @@ type tracker struct {
 	Metadata     map[schema.GroupVersionKind]*Cache
 }
 
+// GetOptions provides configuration to customize the behavior when
+// getting an informer.
+type GetOptions struct {
```
sbueringer, line 112: (Just trying to understand how it fits together)

How is this option related to allowing to stop informers?

Is it that in your scenario the informers are started without waiting for the cache and if they don't come up you call the stop func?

maxsmythe: This bit is more about the utility of and usage patterns for stoppable informers.

In the original code, when a user requests an informer, controller-runtime blocks until that informer's cache is populated. This makes sense for imperative requests, like retrieving a specific object, since the cache needs to be read from for the request to be satisfied. This is not strictly necessary for registering a listener with an informer, since that listener will simply be dormant until events are fed to it.

Gatekeeper has a two-controller model for dynamic informers. There is the main controller and the sub controller. The main controller knows how to derive what GVKs the sub controller should be receiving events for. It then registers a listener with that GVK's associated informer (constructing it, if necessary). This listener passes events to the sub controller's work queue.

I should note I'm being a bit hand-wavy here as Gatekeeper actually uses a centralized structure to keep track of "interested" controllers (essentially a home-grown version of the reference counting solution controller-runtime is considering). We call it the "watch manager".

By allowing this centralized reference counter the ability to get the informer without blocking on it being ready, the watch manager can add/remove listeners without waiting for the informer to bootstrap (it's preferable to have initializing a watch be a synchronous operation so that the managing controller's reconciler can return whether or not a retry is warranted). This reduces the amount of time that main controllers can block each other.

Of course, sub controllers will eventually need to run a `Get()` request in order to run their reconcile loop, and that will be blocked on bootstrapping but:

1) if the sub controller is receiving events, that means the initial `List()` operation of the informer has been completed and therefore bootstrapping is likely also complete, or nearly complete.

2) there are fewer goroutines involved. A blocked sub controller only blocks itself. Main controllers can block each other.

TL;DR: for efficiency's sake, it'd be nice if adding/removing listeners didn't block on a bootstrapping process which they are not dependent on.

pkg/cache/internal/informers.go
maxsmythe: It's intended to be used by whatever is responsible for knowing when an informer is no longer needed.

In the short term, this responsibility is owned by individual users. Longer-term, it sounds like controller-runtime wants to have some kind of solution for this (reference counting has been floated).

I think it's useful to have this function public as:

1) Otherwise, users of controller-runtime will be blocked from using dynamic informers until a solution is agreed on/implemented (original issue, #1884, is over a year old without a definitive path forward, so this is definitely not a trivial task). Without this, they have no way of cleaning up unused state.

2) Having this method public will provide an escape hatch should controller-runtime's chosen model of managing informers be incompatible with a given project.

pkg/cache/internal/informers.go
alvaroaleman: That makes sense but seems unrelated to the rest of this PR, can this be it's own change? We seem to be unclear about what the best name is for this but other than that this bit seems uncontroversial

pkg/internal/syncs/syncs.go
maxsmythe: Are you saying that you'd like to see `cancelFunc()` block until `out` is definitely closed?

This is different than how context.WithCancel() [works](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/context/context.go;l=471), but I don't see that as proof-positive it shouldn't be done how you're suggesting.

What is the use case for blocking-until-closed? That won't necessarily guarantee that any goroutines that are expected to respond to `out` have done so (e.g. maybe they are running an infinite for-loop, occasionally checking to see if `out` is closed). To guarantee that, each goroutine would need to send a `done` signal, confirming they have shut down.

Or do you see a bug where `out` may not successfully be closed?

pkg/cache/internal/informers.go
maxsmythe: I have no objections to splitting non-blocking-get into its own PR. I'll wait a day or so to open up a new PR to give people time to object.

pkg/cache/informer_cache.go
maxsmythe: @inteon blocking is actually the current behavior. This option allows us to get an informer without blocking-for-sync. As such, I think adding this option is moving things in the direction you're suggesting.

blocking is still necessary for supporting `client.Get()`, which is synchronous and assumes the informer cache is ready to be read from.

Does that allay your concern?

braghettos:
Hi everyone! Any possibility that this issue moves forward?

maxsmythe:
Hey, sorry, I got distracted by integrating the latest release of controller-runtime into Gatekeeper. I hope to re-pick-up this thread this week.

pkg/internal/syncs/syncs.go
maxsmythe: I'm not sure what an example of code that relies on cancel blocking on channel close would look like? Generally cancel() signals child goroutines that they should exit.

To avoid arguing, I've added it in. It does mean that future coding errors have a higher chance of causing deadlocks, but I think is fine as-written.

pkg/cache/internal/informers.go
maxsmythe: This is moved to https://github.com/kubernetes-sigs/controller-runtime/pull/2371

I'll add the suggestion in there.

pkg/cache/internal/informers.go
maxsmythe: Split into https://github.com/kubernetes-sigs/controller-runtime/pull/2371

pkg/cache/internal/informers.go
maxsmythe: Wait... Async was a separate suggestion. Leaving `DoNotBlockUntilSynced` for now, we can debate naming on the new PR.

One thing to note is that `false` should mean "block until ready" in order to maintain backwards compatible behavior when no args are applied.

maxsmythe:
Split the get options into a separate PR. Tests will have to wait until next week, sadly.

pkg/cache/internal/informers.go
maxsmythe: Added unit tests

pkg/internal/syncs/syncs_test.go
maxsmythe: fixed

maxsmythe:
ping

maxsmythe:
Waiting for feedback.

pkg/cache/informer_cache.go
maxsmythe: closing this, since non-blocking get has been moved to #2371

pkg/cache/informer_cache.go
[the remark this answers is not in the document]
maxsmythe: Closing this, since non-blocking was moved to #2371

pkg/cache/internal/informers.go
maxsmythe: moved

pkg/cache/internal/informers.go
maxsmythe: Closing this, since non-blocking was moved to #2371

sozercan:
@sbueringer @vincepri can you ptal again when you get a chance please?

vincepri:
changes generally lgtm, can you rebase?

maxsmythe:
@vincepri Thanks for looking! Rebased!

Any concerns WRT attribution, since this code is adapted from Gatekeeper's fork? At this point, the only un-modified code is living in `internal/informers.go`.

The `MergeChans` function is IMO new, since it has been reworked to be entirely generic (previously had hardcoded arity and types). Here is a link to the old Gatekeeper code for reference: https://github.com/open-policy-agent/gatekeeper/blob/694e771143fdbefb1630609eea9443e2c0ca7742/third_party/sigs.k8s.io/controller-runtime/pkg/dynamiccache/internal/informers_map.go#L427-L472

LMK if you want me to see if there is anyone we should be asking for advice.

maxsmythe:
@vincepri Any updates? Barring licensing questions, code should be good to merge.

maxsmythe:
@vincepri ping

vincepri:
On my queue for Monday, apologies I’ve just got back

maxsmythe:
No worries!

vincepri (approved):

vincepri:
@maxsmythe Chatting with @alvaroaleman it'd be good going forward to have a small design that talks on how we can get to the larger state of dynamic watchers.

alvaroaleman:
@maxsmythe given that we do not have anything to dynamically manage watches on controllers, would you be ok if we change this to `StopInformer` but not remove it from the map? It seems that would still solve your problem but not result in confusing behavior if the informer is then later re-created (existing event sources do not get event)

maxsmythe:
Thanks for merging!!

@vincepri @alvaroaleman Definitely interested in making this more user-friendly. Lemme know how I can help.

@alvaroaleman The one concern I would have is memory leaks. The main reason to remove the informer is to also remove the cache data, cleaning stale state.

Is there some way to remove the cache but keep a record of interested sources (this could still be a memory leak, but a smaller one)? Is there a way to have two methods (stop informer and remove informer)? Maybe figuring out the edge cases is part of the larger "dynamic watcher" story, that this is more a stopgap on the way to?

alvaroaleman:
@maxsmythe memory leak because there is still data in the informer (the state of the world before it was stopped) or because of the informer itself?

alvaroaleman:
> Mostly concerned about the state of the world, as that's the larger dataset (there can be a LOT of very large config maps in a cluster, for instance).

Yeah, maybe there is a way to clean up the store in the informer?

> AFAIK the first attempt is https://github.com/kubernetes-sigs/controller-runtime/pull/2159, which appears to be focused on starting/stopping controllers ad-hoc. Is there any other state I should know about? I can sketch some rough thoughts on a Google doc to start with, unless you have a more preferred way of handling these things?

Only the linked issues to my knowledge. And starting to sketch something would be really helpful :) The PRs make it too easy to miss the forest from all the trees because these changes are big.

maxsmythe:
Still working on the stories. @alvaroaleman Is there a timeline yet for when a release may be cut that has this?

PR #2289, opened 2023-04-26
:seedling: envtest: set default webhook options for polling

mythi:
Similar to how it's done for the CRD installation, set default webhook options for timeout/poll interval values.

This avoids getting MaxTime=0 which changes its behaviour when moving from k8s deprecated wait.PollImmediate() to wait.PollUntilContextTimeout().

Spun off from: #2189


sbueringer:
Makes sense to wait for webhooks the same way as we wait for CRDs

/ok-to-test
/lgtm
/approve

PR #2293, opened 2023-05-01
⚠️ Allow passing a custom webhook server

alvaroaleman:
Currently it is impossible to pass a custom webhook server, because we reference the concrete type rather than an interface. Change this to an interface.


pkg/manager/internal.go
```diff
@@ -130,7 +130,7 @@ type controllerManager struct {
 	// election was configured.
 	elected chan struct{}
 
-	webhookServer *webhook.Server
+	webhookServer webhook.Server
```
vincepri, line 133: What are some use cases that require an interface instead?

alvaroaleman: Basically anything where the default implementation isn't working. The case I have at hand is that the same binary also generates the certificates, so `Start` has to wait until they exist.

pkg/webhook/server.go
```diff
@@ -79,6 +101,18 @@ type Server struct {
 
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
+}
+
+// New constructs a new Server from the provided options.
+func New(o Options) Server {
```
sbueringer, line 107: Should this be called NewServer to end up with webhook.NewServer?

pkg/webhook/server.go
```diff
@@ -79,6 +101,18 @@ type Server struct {
...
+	return &DefaultServer{
+		Options: o,
+	}
+}
+
+// DefaultServer is the default implementation used for Server.
+type DefaultServer struct {
+	Options Options
```
sbueringer, line 115: Wondering if Options should be private given that we also have a constructor (but no strong opinion)

alvaroaleman: The reason I left it public is that there are a bunch of tests in the `Manager` that check some values in there

pkg/webhook/server.go
alvaroaleman: Good idea, renamed it

sbueringer:
/lgtm

/hold
not sure if Vince wants to take another look

PR #2296, opened 2023-05-02
:warning: Use and provide a single dynamic rest mapper

vincepri:


sbueringer:
Didn't see the test failure before but looks unrelated

/test pull-controller-runtime-test

pkg/client/apiutil/restmapper.go
```diff
@@ -18,30 +18,46 @@ package apiutil
...
 	// mutex to provide thread-safe mapper reloading.
-	mu sync.Mutex
+	mu sync.RWMutex
 }
 
-// newLazyRESTMapperWithClient initializes a LazyRESTMapper with a custom discovery client.
-func newLazyRESTMapperWithClient(discoveryClient *discovery.DiscoveryClient) (meta.RESTMapper, error) {
-	return &lazyRESTMapper{
+// newMapperWithClient initializes a LazyRESTMapper with a custom discovery client.
+func newMapperWithClient(discoveryClient *discovery.DiscoveryClient) (meta.RESTMapper, error) {
```
sbueringer, line 59:
```suggestion
func newMapper(discoveryClient *discovery.DiscoveryClient) (meta.RESTMapper, error) {
```
nit

sbueringer (commented):
one nit, up to you

pkg/client/apiutil/restmapper.go
```diff
@@ -162,9 +175,11 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
 		Group:              metav1.APIGroup{Name: groupName},
 		VersionedResources: make(map[string][]metav1.APIResource),
 	}
+	m.mu.Lock()
```
sbueringer, line 178: Should this be a read lock?

pkg/client/apiutil/restmapper.go
```diff
@@ -196,52 +211,64 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
 	}
 
 	// Update data in the cache.
+	m.mu.Lock()
 	m.knownGroups[groupName] = groupResources
```
sbueringer, line 215: I assume it's not a problem if we have concurrent calls for the same groupName as they will end up with the same groupResources?

pkg/client/apiutil/restmapper.go
```diff
@@ -162,9 +175,11 @@ func (m *lazyRESTMapper) addKnownGroupAndReload(groupName string, versions ...st
...
 		VersionedResources: make(map[string][]metav1.APIResource),
 	}
+	m.mu.Lock()
 	if _, ok := m.knownGroups[groupName]; ok {
 		groupResources = m.knownGroups[groupName]
 	}
+	m.mu.Unlock()
 
 	// Update information for group resources about versioned resources.
 	// The number of API calls is equal to the number of versions: /apis/<group>/<version>.
```
sbueringer, line 185: Could it be that two go routines concurrently modify groupResources?

RainbowMango:
@vincepri May I ask the background of this PR?

Why removed the following options: WithCustomMapper, WithExperimentalLazyMapper, WithLazyDiscovery, and WithLimiter?

vincepri:
@RainbowMango The options were either outdated, or provided no meaningful additions. Folks can always use their own RESTMapper implementation if they need. For the majority of controller-runtime users, having a single implementation is less error prone.

RainbowMango:
Hi @vincepri, 
We built a [cached restmapper](https://github.com/karmada-io/karmada/blob/2a314eb46ec3cc1d18d1fd326559442515afbc19/pkg/util/restmapper/restmapper.go#L23-L28) based on the `DynamicRESTMapper` of contriller-runtime@v0.14, the `cached restmapper` simply maintains a map which records the recent match result, and the map will [be reset](https://github.com/karmada-io/karmada/blob/2a314eb46ec3cc1d18d1fd326559442515afbc19/pkg/util/restmapper/restmapper.go#L92-L100) if the `DiscoveryRESTMapper` needs a reload. 
We did that by the `WithCustomMapper` before, but during the bump to v0.15, we can't use it anymore.

Do you have any suggestions about how we can adopt the breaking change?
Or, is it possible to add the `WithCustomMapper` back to the upcoming release?

PS:
The background why we need the `cahced restmapper` is all about the performance, the underlying Kubernetes `DiscoveryRESTMapper` caches the resource [with a slice](https://github.com/kubernetes/kubernetes/blob/9946ea9fd851fb1cd2148b746077d9f89c27e074/staging/src/k8s.io/client-go/restmapper/discovery.go#L44), which performance is not good comparing with map. 
We did a benchmark after that:
```
# go test -bench=. 
goos: linux
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/restmapper
cpu: Intel(R) Xeon(R) Gold 6278C CPU @ 2.60GHz
BenchmarkGetGroupVersionResource-4            	   52779	     20799 ns/op
BenchmarkGetGroupVersionResourceWithCache-4   	  245550	      4754 ns/op   // cached restmapper
PASS
ok  	github.com/karmada-io/karmada/pkg/util/restmapper	2.574s
```
The performance of `cached restmapper` had improved significantly.

vincepri:
@RainbowMango Overriding the RESTMapper implementation in the manager is still possible with https://github.com/kubernetes-sigs/controller-runtime/blob/56a973c46efe463ef7b5fbecd8495e8378b50d20/pkg/manager/manager.go#L108-L112

PR #2298, opened 2023-05-04
⚠ Remove metrics with high cardinality

sbueringer:
This PR basically reverts the metrics introduced in #2217. In practice the cardinality of these metrics is too high when CR is used to communicate with a lot of different clusters (host x verb x histogram). So we shouldn't register them per default.

This PR also drops `RequestLatency` as:
1. The documented way to register it doesn't actually work
2. 1.17 was a while ago

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,8 +18,6 @@ package metrics
 
 import (
 	"context"
-	"net/url"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	clientmetrics "k8s.io/client-go/tools/metrics"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

jonathan-innis:
@sbueringer I was just looking around at this. Though this was dropped from the default registration flow, I wonder if it still makes sense to have a mechanism for registering these metrics if you are using controller-runtime. At present, anyone who wants to get these kinds of metrics has to go hunt around for how to do it in the `client-go` code and then discover that controller-runtime _used_ to implement this but doesn't anymore.

sbueringer:
@jonathan-innis In general it would be great to have a mechanism for this. Unfortunately, today we have to register these metrics in an `init` func. (the register func in client-go uses once).

So I think if we want to make this configurable in controller-runtime (e.g. via manager options) we have to make changes to the Register func in client-go first

jonathan-innis:
> So I think if we want to make this configurable in controller-runtime (e.g. via manager options) we have to make changes to the Register func in client-go first

I could see us just adding the `RegisterClientMetrics()` call back. I'll put a PR up upstream. We are already thinking about this in https://github.com/awslabs/operatorpkg. I also think it would be really nice if we just parsed out all of URL rather than just showing the raw URL data (getting something like "group", "kind", "version", etc. in a structured format would be really nice)

PR #2300, opened 2023-05-04
:warning: Remove cache.BuilderWithOptions, and inheritance funcs

vincepri:
The function was previously created and used to override the Manager's NewCache function. Now that the Manager embeds cache options we can remove this function and all of the inheritance logic which is highly error prone.


vincepri:
Supersedes https://github.com/kubernetes-sigs/controller-runtime/pull/2237

alvaroaleman (approved):

PR #2305, opened 2023-05-07
:sparkles: Fakeclient: Add support for evictions

alvaroaleman:
This change adds support for evictions into the fakeclient. 
Prior to this, it would just error on any subresource creation.

/cc @vincepri @sbueringer

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -30,6 +30,9 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -969,7 +972,15 @@ func (sw *fakeSubResourceClient) Get(ctx context.Context, obj, subResource clien
 }
 
 func (sw *fakeSubResourceClient) Create(ctx context.Context, obj client.Object, subResource client.Object, opts ...client.SubResourceCreateOption) error {
-	panic("fakeSubResourceWriter does not support create")
+	switch subResource.(type) {
```
sbueringer, line 975: Would it make sense to store "subResource string" of the SubResource call (l.936) in the fakeSubResourceClient struct and then only implement this logic here if the subResource (string) is indeed "eviction"?

alvaroaleman: Good point, updated, PTAL

sbueringer:
Thx!

/lgtm
/approve

PR #2306, opened 2023-05-07
:warning: Inteceptor: Unify Client and Subresource Client

alvaroaleman:
At the moment, the inteceptor has two different constructors for a client interceptor and a subresource interceptor. This means that in order to intercept subresource requests, one has to set up a normal interceptor, set the `SubResource` func to return a subresource interceptor and then configure the desired behavior on the subresource interceptor.

This is needlessly complex, unify the two and add the parameters for the subresource funcs into the top-level `Funcs` struct.

/cc @vincepri @sbueringer @ludydoo


ludydoo:
That looks much simpler indeed

PR #2314, opened 2023-05-10
:bug: builder: handle core Group correctly in webhook paths

mythi:
Webhooks for the legacy/core v1 Kinds registered using the builder gives paths like "/mutate--v1-pod" which may not be very intuitive to use in the markers.

To follow the schema.GroupVersion().String() convention, just leave empty Group out of the webhook path to get "/mutate-v1-pod" (also as it was in the examples/builtin originally).

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -240,11 +240,17 @@ func (blder *WebhookBuilder) isAlreadyHandled(path string) bool {
 }
 
 func generateMutatePath(gvk schema.GroupVersionKind) string {
-	return "/mutate-" + strings.ReplaceAll(gvk.Group, ".", "-") + "-" +
-		gvk.Version + "-" + strings.ToLower(gvk.Kind)
+	return strings.Join([]string{
+		"/mutate",
+		strings.ReplaceAll(strings.Replace(gvk.GroupVersion().String(), "/", "-", 1), ".", "-"),
+		strings.ToLower(gvk.Kind),
+	}, "-")
 }
 
 func generateValidatePath(gvk schema.GroupVersionKind) string {
-	return "/validate-" + strings.ReplaceAll(gvk.Group, ".", "-") + "-" +
-		gvk.Version + "-" + strings.ToLower(gvk.Kind)
+	return strings.Join([]string{
+		"/validate",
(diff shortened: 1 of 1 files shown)


sbueringer:
I"m not sure if it's more intuitive to not have a group at all in the path for the core group vs having the empty string (which is used in a bunch of places to refer to the core group).

This is also a breaking change for every user of controller-runtime that has a webhook on a core API.

mythi:
> Given that, I'm not sure if this is something we should change.

I'm fine dropping this. Are everyone using the double hyphen then?

mythi:
I pushed an alternative "fix" to: https://github.com/mythi/controller-runtime/commit/696c28ca72bc4dc08942fe16c7a2e0dd2c38f341

mythi:
> > > Given that, I'm not sure if this is something we should change.
> > 
> > 
> > I'm fine dropping this. Are everyone using the double hyphen then?
> 
> Everyone who used the `WebhookBuilder` had to use the double hyphen otherwise they would have ended up with 404 errors.

True, however, the [documentation](https://book.kubebuilder.io/reference/webhook-for-core-types.html) for core types is suggesting another approach and that's going to break with 0.15 so users will see breakages anyway.

But I'm not that deep in this space so I'm fine any approach and I've already fixed my webhooks for 0.15 after first debugging why I need to add `--` ;-)

alvaroaleman:
> Seems to me like a documentation error in kubebuilder. Basically this never worked in CR (not only v0.15 as far as I know). I think we shouldn't break CR users because the kubebuilder documentation doesn't match the CR implementation.


Yeah, agreed. Also agreed on the point that this isn't worth breaking users for.

mythi:
> > Seems to me like a documentation error in kubebuilder. Basically this never worked in CR (not only v0.15 as far as I know). I think we shouldn't break CR users because the kubebuilder documentation doesn't match the CR implementation.
> 
> Yeah, agreed. Also agreed on the point that this isn't worth breaking users for.

ack. CR's examples/builtin used the same approach as documented by kubebuilder so I'll just have the examples fixed then via https://github.com/mythi/controller-runtime/commit/696c28ca72bc4dc08942fe16c7a2e0dd2c38f341

/close

sbueringer:
Sounds good. Thx!

Issue #2184, opened 2023-02-07
Fake client does not delete objects containing deletion timestamp


lleshchi:
The fake client Get(), List(), Patch() methods do not account for objects that have been marked for deletion, allowing for viewing of and changes to objects that are expected to be deleted. The Update() method does [delete the object](https://github.com/kubernetes-sigs/controller-runtime/blob/f127c11669de63e97a0e568b53b09362c8735154/pkg/client/fake/client.go#L375-L377), but I believe it should return an error, since the update of a deleted object has failed.

This is a separate issue from https://github.com/kubernetes-sigs/controller-runtime/issues/138

2uasimojo:
> The Update() method does [delete the object](https://github.com/kubernetes-sigs/controller-runtime/blob/f127c11669de63e97a0e568b53b09362c8735154/pkg/client/fake/client.go#L375-L377), but I believe it should return an error, since the update of a deleted object has failed.

So actually there are two possibilities here:
- If this Update() removed the last finalizer(s), then the current behavior is correct. However...
- If there were no finalizers *before* this Update() was called, we should simulate gc having happened prior to the Update() and error NotFound.

Glancing at the code, it should be easy enough to make this distinction:
```go
if len(oldAccessor.GetFinalizers()) > 0 && len(accessor.GetFinalizers()) == 0 {
```

alvaroaleman:
How did you get an object with a non-nil deletionTimestamp into the fakeclient in the first place?

lleshchi:
Hi @alvaroaleman, we use `WithRuntimeObjects()` [here](https://github.com/openshift/hive/blob/a04d71c0205fc279c281152c3f5106d6590c267c/pkg/controller/hibernation/hibernation_controller_test.go#L991) to add a DeletionTimestamp. Since the k8s api does not allow a user to add a deletion timestamp when creating an object, and ignores attempts to add/edit the timestamp, perhaps the fake client should handle deletion timestamps similarly, in order to prevent tests such as the one above.

2uasimojo:
Right, so to expand on this:

In general we want to be able to test reconciler behavior on deleted objects. With the fake client as it stands, we've been able to write test cases where we just init the fake client `With[Runtime]Objects()` that have `deletionTimestamp` set, but no finalizers -- which is not a valid state in real k8s (ignoring delete propagation, which we're not looking to solve here).

[...]

PR #2316, opened 2023-05-10
⚠ Update fake client deletionTimestamp behavior

lleshchi:
Closes https://github.com/kubernetes-sigs/controller-runtime/issues/2184

The current fake client methods diverge from the behavior of real k8s. This results in scenarios where an object can be created or modified to have a deletionTimestamp and no finalizers, with a delay in garbage collection, causing inaccurate or ambiguous behavior in  test cases that use the fake client.

Bring fake client behavior closer in line with real k8s behavior by modifying the following:
- Update() and Patch() will reject changes that set a previously-unset deletionTimestamp
- Create() will ignore a deletionTimestamp on an object
- Build() will error if initialized with an object that has a deletionTimestamp but no finalizers; but will accept a deletionTimestamp if it has finalizers.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/controller-runtime
 go 1.20
 
 require (
+	github.com/evanphx/json-patch v4.12.0+incompatible
 	github.com/evanphx/json-patch/v5 v5.6.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/go-logr/logr v1.2.4
@@ -31,7 +32,6 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/emicklei/go-restful/v3 v3.9.0 // indirect
-	github.com/evanphx/json-patch v4.12.0+incompatible // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
 	github.com/go-openapi/jsonreference v0.20.1 // indirect
 	github.com/go-openapi/swag v0.22.3 // indirect
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -27,7 +27,10 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
+	// Using v4 to match upstream
+	jsonpatch "github.com/evanphx/json-patch"
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
 	corev1 "k8s.io/api/core/v1"
@@ -41,8 +44,10 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -285,6 +290,9 @@ func (t versionedTracker) Add(obj runtime.Object) error {
 		if err != nil {
 			return fmt.Errorf("failed to get accessor for object: %w", err)
 		}
+		if accessor.GetDeletionTimestamp() != nil && len(accessor.GetFinalizers()) == 0 {
+			return fmt.Errorf("refusing to create obj %s with metadata.deletionTimestamp but no finalizers", accessor.GetName())
+		}
 		if accessor.GetResourceVersion() == "" {
 			// We use a "magic" value of 999 here because this field
 			// is parsed as uint and and 0 is already used in Update.
@@ -368,10 +376,10 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	if bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch")) {
 		isStatus = true
 	}
-	return t.update(gvr, obj, ns, isStatus)
+	return t.update(gvr, obj, ns, isStatus, false)
 }
 
-func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Object, ns string, isStatus bool) error {
+func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Object, ns string, isStatus bool, deleting bool) error {
 	accessor, err := meta.Accessor(obj)
 	if err != nil {
 		return fmt.Errorf("failed to get accessor for object: %w", err)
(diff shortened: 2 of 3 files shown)


pkg/client/fake/client.go
```diff
@@ -835,6 +882,80 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
...
+		return newTime == oldTime
+	}
+	return newTime.Time.Sub(oldTime.Time).Abs() < time.Second
+}
+
+// The behavior of applying the patch is pulled out into dryPatch(),
+// which applies the patch and returns an object, but does not Update() the object.
+// This function returns a patched runtime object that may then be validated before a call to Update() is executed.
+// This results in some code duplication, but was found to be a cleaner alternative than unmarshalling the data
+// or updating the k8s client-go methods directly.
```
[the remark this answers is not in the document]
2uasimojo: Also suggest referencing the upstream code you copied, linking a specific commit level or version, so your posterity can easily identify drift. E.g. current master is https://github.com/kubernetes/client-go/blob/783d0d33626e59d55d52bfd7696b775851f92107/testing/fixture.go#L146-L194

2uasimojo:
This is looking solid to me. One minor query inline. Also, I'm not digging deep into the test cases -- leaving that for the experts :)

go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/controller-runtime
 go 1.20
 
 require (
+	github.com/evanphx/json-patch v4.12.0+incompatible
```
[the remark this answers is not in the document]
lleshchi: It was `make modules`

pkg/client/fake/client_test.go
```diff
@@ -726,6 +726,54 @@ var _ = Describe("Fake client", func() {
 			Expect(apierrors.IsNotFound(err)).To(BeTrue())
```
vincepri, line 726: Any reason file permissions have changed?

pkg/client/fake/client.go
```diff
@@ -438,7 +445,12 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
 	}
 	intResourceVersion++
 	accessor.SetResourceVersion(strconv.FormatUint(intResourceVersion, 10))
-	if !accessor.GetDeletionTimestamp().IsZero() && len(accessor.GetFinalizers()) == 0 {
+
+	if !deleting && !validateTimestampDifference(accessor, oldAccessor) {
+		return fmt.Errorf("error: Unable to edit %s: metadata.deletionTimestamp field is immutable", accessor.GetName())
+	}
+
+	if !oldAccessor.GetDeletionTimestamp().IsZero() && len(accessor.GetFinalizers()) == 0 {
```
sbueringer, line 453:
```suggestion
	if !accessor.GetDeletionTimestamp().IsZero() && len(accessor.GetFinalizers()) == 0 {
```
nit: we can keep accessor here right?

pkg/client/fake/client.go
```diff
@@ -813,8 +829,39 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 		return err
 	}
 
+	o, err := c.tracker.Get(gvr, accessor.GetNamespace(), accessor.GetName())
```
sbueringer, line 832:
```suggestion
	oldObj, err := c.tracker.Get(gvr, accessor.GetNamespace(), accessor.GetName())
```

pkg/client/fake/client.go
```diff
@@ -813,8 +829,39 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
...
+	if err != nil {
+		return err
+	}
+	oldObj, err := meta.Accessor(o)
```
sbueringer, line 836:
```suggestion
	oldAccessor, err := meta.Accessor(o)
```
nit: for consistency

pkg/client/fake/client.go
```diff
@@ -838,6 +885,81 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 	return err
 }
 
+// Applying a patch results in a deletionTimestamp that is truncated to the nearest second.
+// Check that the diff between a new and old deletion timestamp is within a reasonable threshold
+// to be considered unchanged.
+func validateTimestampDifference(newObj metav1.Object, obj metav1.Object) bool {
```
sbueringer, line 891:
```suggestion
func deletionTimestampEqual(newObj metav1.Object, obj metav1.Object) bool {
```
nit. I think this would read better where we use it

pkg/client/fake/client.go
```diff
@@ -838,6 +885,81 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
...
+	newTime := newObj.GetDeletionTimestamp()
+	oldTime := obj.GetDeletionTimestamp()
+
+	if newTime == nil || oldTime == nil {
+		return newTime == oldTime
+	}
+	return newTime.Time.Sub(oldTime.Time).Abs() < time.Second
```
sbueringer, line 898: Why are we allowing any difference at all?

pkg/client/fake/client.go
```diff
@@ -27,7 +27,9 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
+	jsonpatch "github.com/evanphx/json-patch"
```
sbueringer, line 32: Let's add a comment that we are intentionally not using /v5 to use the same dependency as upstream

pkg/client/fake/client_test.go
```diff
@@ -927,6 +1046,70 @@ var _ = Describe("Fake client", func() {
...
+			Expect(err).NotTo(BeNil())
+
+		})
+
+		It("should handle finalizers on Patch", func() {
+			namespacedName := types.NamespacedName{
+				Name:      "test-cm",
+				Namespace: "delete-with-finalizers",
+			}
+			By("Updating a new object")
```
sbueringer, line 1087:
```suggestion
			By("Creating a new object")
```

pkg/client/fake/client_test.go
```diff
@@ -927,6 +1046,70 @@ var _ = Describe("Fake client", func() {
...
+
+			By("Removing the finalizer")
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       namespacedName.Name,
+					Namespace:  namespacedName.Namespace,
+					Finalizers: []string{},
+				},
+			}
 			obj.Finalizers = []string{}
```
sbueringer, line 1113:
```suggestion
```

go.mod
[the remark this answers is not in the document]
sbueringer: This was moved to a direct dependency because we are now also using the v4 version of json-patch directly

pkg/client/fake/client.go
lleshchi: Patch seems to truncate the timestamps at the second mark, so an `Equal()` comparison fails:
`s: "Rejected patch, metadata.deletionTimestamp immutable. Diff: new: 2023-05-09 11:33:21 -0400 EDT old: 2023-05-09 11:33:21.749195866 -0400 EDT m=+0.479131449",`

This function was meant to add some tolerance, but it's far from a clean solution

pkg/client/fake/client_test.go
lleshchi: Could you clarify what you mean by this? My understanding was that by removing the object's finalizer, the resource is garbage collected, resulting in a 404 error when calling Get()

pkg/client/fake/client.go
sbueringer: Ah sorry, you wrote it in the godoc. Sounds fine!

sbueringer:
@lleshchi Could it be that you forgot to push the new version? :)

vincepri:
Which one? :)

pkg/client/fake/client_test.go
lleshchi: @vincepri Could you clarify what you mean here? My understanding is that `Expect(apierrors.IsNotFound(err)).To(BeTrue())` is a check to ensure that once the object's finalizer is removed, it is garbage collected, and a call to `Get()` returns a 404 error.

vincepri: The diff was showing the file (client_test.go) permissions changed (to 777), it can be resolved now, seems ok

lleshchi:
@vincepri https://github.com/kubernetes-sigs/controller-runtime/pull/2316#discussion_r1194071114

Apologies, I reach out in slack about this. Should have posted here as well

pkg/client/fake/client_test.go
lleshchi: Ah, I see. It's currently 755, should I change it back to 644?

sbueringer:
Thank you!

/lgtm

PR #2321, opened 2023-05-11
🐛 Revert: move health probes to runnable

sbueringer:
This PR reverts part of: https://github.com/kubernetes-sigs/controller-runtime/pull/2275

On main we have the following sequence in controllerManager.Start():
* ...
* add health probe server
* ...
* Start webhook runnables
* Start cache runnables
* Start other runnables (including health probe server)

Usually this works except under the following circumstances (concrete example):
* The controller implements a reconciler for a CRD called MachineDeployment
* MachineDeployment has multiple apiVersions and thus the controller also implements a conversion webhook
* There are MachineDeployments in the older apiVersion stored in etcd

Now we have the following problem:
* Start cache runnables fails because caches are not getting ready
* Caches are not getting ready because MachineDeployment list does not work
* MachineDeployment list does not work because the conversion webhook does not work ("Post \"https://capi-webhook-service.capi-system.svc:443/convert?timeout=30s\": dial tcp 10.142.135.97:443: connect: connection refused")
* conversion webhook does not work because the service/pod does not get ready
* Pod does not get ready because health probes are not started
* health probes are not started because cache does not get ready

tl;dr cyclic dependency between start caches and start runnables.

This PR solves this by immediately starting the health probes again.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -317,9 +317,9 @@ func (cm *controllerManager) addMetricsServer() error {
 	})
 }
 
-func (cm *controllerManager) addHealthProbeServer() error {
+func (cm *controllerManager) serveHealthProbes() {
 	mux := http.NewServeMux()
-	srv := httpserver.New(mux)
+	server := httpserver.New(mux)
 
 	if cm.readyzHandler != nil {
 		mux.Handle(cm.readinessEndpointName, http.StripPrefix(cm.readinessEndpointName, cm.readyzHandler))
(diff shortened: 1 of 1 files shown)


sbueringer:

/hold
I'll verify with Cluster API if the deadlock is gone before merge

sbueringer:
/hold cancel

Ran the Cluster API test with the CR version from this PR locally and everything works as expected again.

sbueringer:
(cc just fyi @zqzten, we're reverting this part for now to ensure CR main works again, we can work on a new implementation independent of that)

vincepri (approved):

zqzten:
I think we can introduce a new "basics" runnable group which includes the previous internal servers and start it before any other runnable groups, this seems to solve the dead lock issue here. wdyt @sbueringer

sbueringer:
> I think we can introduce a new "basics" runnable group which includes the previous internal servers and start it before any other runnable groups, this seems to solve the dead lock issue here. wdyt @sbueringer

Yup, something like this would work.

PR #2334, opened 2023-05-19
⚠ Change webhook request received / response written logs to log level 4

sbueringer:
Tested a bit with the CR beta and the logs are very verbose.

I"m not sure if that was also the case with v0.14 but 4 seems more reasonable.

xref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -93,7 +93,7 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		wh.writeResponse(w, reviewResponse)
 		return
 	}
-	wh.getLogger(&req).V(1).Info("received request")
+	wh.getLogger(&req).V(4).Info("received request")
 
 	reviewResponse = wh.Handle(ctx, req)
 	wh.writeResponseTyped(w, reviewResponse, actualAdmRevGVK)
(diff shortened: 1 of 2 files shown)


pkg/webhook/admission/http.go
```diff
@@ -136,11 +136,11 @@ func (wh *Webhook) writeAdmissionResponse(w io.Writer, ar v1.AdmissionReview) {
 		}
 	} else {
 		res := ar.Response
-		if log := wh.getLogger(nil); log.V(1).Enabled() {
+		if log := wh.getLogger(nil); log.V(4).Enabled() {
```
alvaroaleman, line 139: It is technically a breaking change in that the logs may disappear when running on v1, right?

sbueringer: Changed the PR title

fabriziopandini:
Thanks for this PR, those logs are really noisy and they are making it harder to read controller logs.

pkg/webhook/admission/http.go
fabriziopandini: I don't have full context and I defer the decision to the team, but as far as I remember in previous versions of controller runtime those logs lines were not generated, so it could also be considered a fix...

alvaroaleman: Oh okay, is his new relative to 0.14?

sbueringer: Just checked. The behavior was the same with 0.14. We just didn't notice it before as we are running our controllers per default with log-level 0.

The recommended production log level is 2 though IIRC. I couldn't think a quote for that.

alvaroaleman (approved):

PR #2337, opened 2023-05-22
🌱 Introduce a new runnable group for basic servers of the manager

zqzten:
This PR introduces a new runnable group for basic servers (health probes, metrics and pprof server) of the manager and make it start before any other runnables. This can resolve the [cyclic dependency issue](https://github.com/kubernetes-sigs/controller-runtime/pull/2321) after we changing these servers to runnables.

### DIFF
--- a/pkg/manager/runnable_group.go
+++ b/pkg/manager/runnable_group.go
@@ -28,6 +28,7 @@ type runnableCheck func(ctx context.Context) bool
 // runnables handles all the runnables for a manager by grouping them accordingly to their
 // type (webhooks, caches etc.).
 type runnables struct {
+	HTTPServers    *runnableGroup
 	Webhooks       *runnableGroup
 	Caches         *runnableGroup
 	LeaderElection *runnableGroup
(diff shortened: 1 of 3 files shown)


zqzten:
cc @sbueringer plz double check if this works under your case, thanks

sbueringer:
I'll take a look soon, but I would delay merging it until after the v0.15 release.

vincepri:
Is there a specific problem we're trying to solve to have the servers in a runnable group?

zqzten:
> Is there a specific problem we're trying to solve to have the servers in a runnable group?

Yep, if we don't have them in a specific runnable group, they would be added to the `Others` runnable group by default, which would be started after the caches are ready. This can lead to this [cyclic dependency issue](https://github.com/kubernetes-sigs/controller-runtime/pull/2321).

In a word, we need to have the servers in a specific runnable group which would be started first (at least before the `Caches` runnable group) in order to keep the same start order as which before they become Runnables.

pkg/manager/runnable_group.go
```diff
@@ -37,6 +38,7 @@ type runnables struct {
 // newRunnables creates a new runnables object.
 func newRunnables(baseContext BaseContextFunc, errChan chan error) *runnables {
 	return &runnables{
+		BasicServers:   newRunnableGroup(baseContext, errChan),
```
vincepri, line 41:
```suggestion
		HTTPServers:   newRunnableGroup(baseContext, errChan),
```

?

sbueringer:
Took me a while to find some time to work on the unit test. But there's now a PR up to ensure we don't hit deadlocks during manager.Start: https://github.com/kubernetes-sigs/controller-runtime/pull/2418

We should probably merge #2418 before this PR

vincepri (approved):

PR #2338, opened 2023-05-22
:seedling: bump jsonpatch to v2.3.0

howardjohn:
This pulls in
https://github.com/gomodules/jsonpatch/releases/tag/v2.3.0. This patch fixes performance issues in some cases (large arrays of objects). This is a fairly common case in Kubernetes objects -- we saw in Pods with 1000 env vars, a trivial patch can take 2s of 100% CPU, and the scaling is n^2. With the patch, its ~instant.


sbueringer:
Thx!

/lgtm
/approve

PR #2343, opened 2023-05-24
✨Fix logs in unstructured client

Poor12:
When I use the client of controller-runtime to construct a `subResourceClient` by passing in unstructured obj, I found a log which made me confused, ""unstructured client did not understand object: *unstructured.Unstructured".

I went through the codebase and found that logs are wrongly printed.


inteon (approved):
LGTM
/approve
/lgtm

Issue #2349, opened 2023-05-25
LazyRESTMapper KindsFor does not return all versions when single version is previously queried


tenstad:
## Problem

`addKnownGroupAndReload` is only triggered when no matches are found. If priming the mapper with a single version and then requesting all versions (not specifying version), a match is found and only a single version returned (not multiple versions).

```go
func (m *mapper) KindsFor(resource schema.GroupVersionResource) ([]schema.GroupVersionKind, error) {
	res, err := m.getMapper().KindsFor(resource)
	if meta.IsNoMatchError(err) {
		if err := m.addKnownGroupAndReload(resource.Group, resource.Version); err != nil {
```

## Test Case

The test passes if `lazyRestMapper.KindsFor(schema.GroupVersionResource{Group: "autoscaling", Version: "v1", Resource: "horizontalpodautoscaler"})` is removed, thus making the other `KindsFor` trigger a `addKnownGroupAndReload`.

```go
t.Run("LazyRESTMapper should be able to fetch all versions after fetching single version", func(t *testing.T) {
	g := gmg.NewWithT(t)

	httpClient, err := rest.HTTPClientFor(restCfg)
	g.Expect(err).NotTo(gmg.HaveOccurred())

	lazyRestMapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
	g.Expect(err).NotTo(gmg.HaveOccurred())

	_, err = lazyRestMapper.KindsFor(schema.GroupVersionResource{Group: "autoscaling", Version: "v1", Resource: "horizontalpodautoscaler"})
	g.Expect(err).NotTo(gmg.HaveOccurred())

	kinds, err := lazyRestMapper.KindsFor(schema.GroupVersionResource{Group: "autoscaling", Resource: "horizontalpodautoscaler"})
	g.Expect(err).NotTo(gmg.HaveOccurred())
	g.Expect(len(kinds)).To(gmg.BeNumerically(">", 1))
})
```

## Suggestion

Trigger addKnownGroupAndReload when `schema.GroupVersionResource` input is partial, and some fields are unspecified.

sbueringer:
We had the same issue. I framed it slightly differently: https://github.com/kubernetes-sigs/cluster-api/issues/10032#issuecomment-1906287932

[...]

PR #2350, opened 2023-05-25
🐛 Fix fetching all LazyRESTMapper versions after fetching single version

tenstad:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/2349

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -75,6 +75,12 @@ func (m *mapper) KindFor(resource schema.GroupVersionResource) (schema.GroupVers
 
 // KindsFor implements Mapper.KindsFor.
 func (m *mapper) KindsFor(resource schema.GroupVersionResource) ([]schema.GroupVersionKind, error) {
+	if resource.Version == "" {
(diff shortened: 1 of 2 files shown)


pkg/client/apiutil/restmapper.go
```diff
@@ -75,6 +75,12 @@ func (m *mapper) KindFor(resource schema.GroupVersionResource) (schema.GroupVers
 
 // KindsFor implements Mapper.KindsFor.
 func (m *mapper) KindsFor(resource schema.GroupVersionResource) ([]schema.GroupVersionKind, error) {
+	if resource.Version == "" {
+		if err := m.addKnownGroupAndReload(resource.Group, resource.Version); err != nil {
```
alvaroaleman, line 79: Isn't this a bit over eager and removing the "lazy" aspect if we always reload as soon as we get a request without version?

tenstad: Yeah, can be a bit over eager, but is it possible to fetch less and at the same time return the complete list?

Or is this by design, and https://github.com/kubernetes-sigs/controller-runtime/issues/2349 not actually a bug?

PR #2363, opened 2023-05-31
🐛 hasLabels and matchingLabels step on each other

shanshanying:
This pr is try to fix issue #2098, where `HasLabels` and `MatchingLables` will step on each other (or overwrites each other)

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -513,8 +513,15 @@ type MatchingLabels map[string]string
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingLabels) ApplyToList(opts *ListOptions) {
 	// TODO(directxman12): can we avoid reserializing this over and over?
-	sel := labels.SelectorFromValidatedSet(map[string]string(m))
-	opts.LabelSelector = sel
+	if opts.LabelSelector == nil {
+		opts.LabelSelector = labels.NewSelector()
+	}
+	// If there's already a selector, we need to AND the two together.
+	noValidSel := labels.SelectorFromValidatedSet(map[string]string(m))
+	reqs, _ := noValidSel.Requirements()
+	for _, req := range reqs {
+		opts.LabelSelector = opts.LabelSelector.Add(req)
+	}
 }
 
 // ApplyToDeleteAllOf applies this configuration to the given an List options.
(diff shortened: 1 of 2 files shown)


pkg/client/options.go
```diff
@@ -513,7 +513,18 @@ type MatchingLabels map[string]string
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingLabels) ApplyToList(opts *ListOptions) {
 	// TODO(directxman12): can we avoid reserializing this over and over?
-	sel := labels.SelectorFromValidatedSet(map[string]string(m))
+	var sel labels.Selector
+	if opts.LabelSelector != nil {
+		sel = opts.LabelSelector
+	} else {
+		sel = labels.NewSelector()
+	}
```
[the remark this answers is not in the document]
shanshanying: Got it. Refactored the writing.

shanshanying:
@alvaroaleman PTAL  thanks

pkg/client/options_test.go
```diff
@@ -292,3 +305,45 @@ var _ = Describe("ForceOwnership", func() {
...
+
+	It("Should produce hasLabels in given order", func() {
+		listOpts := &client.ListOptions{}
+
+		hasLabels := client.HasLabels([]string{"labelApe", "labelFox"})
+		hasLabels.ApplyToList(listOpts)
+		Expect(listOpts.LabelSelector.String()).To(Equal("labelApe,labelFox"))
+	})
+
+	It("Should produce an empty selector when given invalid input", func() {
```
alvaroaleman, line 320: This is a bug IMHO. Given invalid input we should produce an invalid selector so this errors down the line rather than just silently dropping the selector.

Its fine not to fix that bug in here but please do not add a testcase for the presence of a bug :)

alvaroaleman (commented):
One comment, other than that lgtm, thanks!

/ok-to-test

pkg/client/options_test.go
shanshanying: I will remove the case as a quick fix.  
(It did puzzled for a while where the behavior of `HasLable.ApplyToList()` differs from that of `MatchingLabels`. Both of them are expected to produce an invalid selector on errors)

shanshanying:
Hi, @alvaroaleman Thanks for your feedback.  I revised the test case as suggested.  PTAL

alvaroaleman (approved):

PR #2365, opened 2023-05-31
🐛 fakeClient.Status().Update(...) cannot recognize resource version conflicts

### LINKED ISSUE #2362: FakeClient update on status subresource not returning error despite conflicting resource version since v0.15.0
After upgrading to controller-runtime 0.15.0 one of our unit tests started failing.
The tested controller performs a `client.Status().Update(...)` on Pod resources. The now failing unit test aims to test the controller behavior in case of a resource conflict (i.e. The fake client has a pod object, the controller performs a status update on the same pod object, but with a different resource version -> expected behavior: Conflict error, actual behavior since upgrading to 0.15.0: no error).

This can be reproduced with the following testcase:  (controller-runtime 0.15.0, k8s api 0.27.2)
```go
func Test_FakeStatusUpdate(t *testing.T) {
	pod := &corev1.Pod{
			Namespace:       "test",
			Name:            "test",
			ResourceVersion: "1",
		},
	}

	c := fake.NewClientBuilder().
		WithRuntimeObjects(pod).
		Build()

	pod.Status.Phase = "testphase"
	pod.ResourceVersion = "2"
	err := c.Status().Update(context.Background(), pod)
	if err == nil {
		t.Fatal("Expected conflict error, but got nil")
	}
}
```

I think it happens because the result of [SetResourceVersion](https://github.com/kubernetes-sigs/controller-runtime/blob/30eae58f1b984c1b8139dd9b9f68dd2d530ed429/pkg/client/fake/client.go#L995) is immediately overwritten again by [fromMapStringAny](https://github.com/kubernetes-sigs/controller-runtime/blob/30eae58f1b984c1b8139dd9b9f68dd2d530ed429/pkg/client/fake/client.go#L997).

There appears to be a unit test for this case, but it doesn't fail because it uses `client.Update(...)` instead of `client.Status().Update(...)`. https://github.com/kubernetes-sigs/controller-runtime/blob/30eae58f1b984c1b8139dd9b9f68dd2d530ed429/pkg/client/fake/client_test.go#LL1441C2-L1441C2

Related PR: #2259

iiiceoo:
The fake client of subresource is unable to correctly handle the case of resource version conflict when updating. The phenomenon is that it did not return a 409 status error.

Ref:
#2362 

Signed-off-by: iiiceoo <iiiceoo@foxmail.com>

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -992,11 +992,11 @@ func copyNonStatusFrom(old, new runtime.Object) error {
 		}
 	}
 
-	newClientObject.SetResourceVersion(rv)
-
 	if err := fromMapStringAny(newMapStringAny, new); err != nil {
 		return fmt.Errorf("failed to convert back from map[string]any: %w", err)
 	}
+	newClientObject.SetResourceVersion(rv)
+
 	return nil
 }
 
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -1431,14 +1431,15 @@ var _ = Describe("Fake client", func() {
 	It("should return a conflict error when an incorrect RV is used on status update", func() {
 		obj := &corev1.Node{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "node",
+				Name:            "node",
+				ResourceVersion: trackerAddResourceVersion,
 			},
 		}
 		cl := NewClientBuilder().WithStatusSubresource(obj).WithObjects(obj).Build()
 
 		obj.Status.Phase = corev1.NodeRunning
 		obj.ResourceVersion = "invalid"
-		err := cl.Update(context.Background(), obj)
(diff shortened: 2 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -992,11 +992,11 @@ func copyNonStatusFrom(old, new runtime.Object) error {
 		}
 	}
 
-	newClientObject.SetResourceVersion(rv)
-
 	if err := fromMapStringAny(newMapStringAny, new); err != nil {
 		return fmt.Errorf("failed to convert back from map[string]any: %w", err)
 	}
+	newClientObject.SetResourceVersion(rv)
```
alvaroaleman, line 998: Please add a testcase

iiiceoo: Sorry, I just noticed this message and I will complete it as soon as possible.

alvaroaleman (approved):
thanks!

PR #2371, opened 2023-06-10
✨ Allow non-blocking retrieval of informers

maxsmythe:
This PR is split off from #2285 

It allows users to retrieve an informer before an informer has finished bootstrapping. This is valuable as it allows users to add/remove listeners without waiting for the informer to be ready. Since a listener is a passive observer of events, it does not require the informer to be fully initialized. Not waiting for full initialization can improve bootstrapping times and enable more flexibility with how dynamic watches are handled.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -44,13 +44,15 @@ var (
 	defaultSyncPeriod = 10 * time.Hour
 )
 
+type InformerGetOptions internal.GetOptions
+
 // InformerGetOption defines an option that alters the behavior of how informers are retrieved.
-type InformerGetOption func(*internal.GetOptions)
+type InformerGetOption func(*InformerGetOptions)
 
 // BlockUntilSynced determines whether a get request for an informer should block
 // until the informer's cache has synced.
 func BlockUntilSynced(shouldBlock bool) InformerGetOption {
-	return func(opts *internal.GetOptions) {
+	return func(opts *InformerGetOptions) {
 		opts.BlockUntilSynced = &shouldBlock
 	}
 }
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -44,6 +44,17 @@ var (
 	defaultSyncPeriod = 10 * time.Hour
 )
 
+// InformerGetOption defines an option that alters the behavior of how informers are retrieved.
+type InformerGetOption func(*internal.GetOptions)
+
+// BlockUntilSynced determines whether a get request for an informer should block
+// until the informer's cache has synced.
+func BlockUntilSynced(shouldBlock bool) InformerGetOption {
+	return func(opts *internal.GetOptions) {
+		opts.BlockUntilSynced = &shouldBlock
+	}
+}
+
 // Cache knows how to load Kubernetes objects, fetch informers to request
 // to receive events for Kubernetes objects (at a low-level),
 // and add indices to fields on the objects stored in the cache.
@@ -61,11 +72,11 @@ type Cache interface {
 type Informers interface {
 	// GetInformer fetches or constructs an informer for the given object that corresponds to a single
 	// API kind and resource.
-	GetInformer(ctx context.Context, obj client.Object) (Informer, error)
+	GetInformer(ctx context.Context, obj client.Object, opts ...InformerGetOption) (Informer, error)
 
 	// GetInformerForKind is similar to GetInformer, except that it takes a group-version-kind, instead
 	// of the underlying object.
-	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind) (Informer, error)
+	GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, opts ...InformerGetOption) (Informer, error)
 
 	// Start runs all the informers known to this cache until the context is closed.
 	// It blocks.
@@ -187,6 +198,9 @@ type Options struct {
 	// ByObject restricts the cache's ListWatch to the desired fields per GVK at the specified object.
 	// object, this will fall through to Default* settings.
 	ByObject map[client.Object]ByObject
+
+	// NewInformer allows overriding of NewSharedIndexInformer for testing.
+	NewInformer func(toolscache.ListerWatcher, runtime.Object, time.Duration, toolscache.Indexers) toolscache.SharedIndexInformer
 }
 
 // ByObject offers more fine-grained control over the cache's ListWatch by object.
@@ -337,6 +351,7 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 				},
 				Transform:             config.Transform,
 				UnsafeDisableDeepCopy: pointer.BoolDeref(config.UnsafeDisableDeepCopy, false),
+				NewInformer:           opts.NewInformer,
 			}),
 			readerFailOnMissingInformer: opts.ReaderFailOnMissingInformer,
 		}
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -44,6 +44,7 @@ var (
 	defaultSyncPeriod = 10 * time.Hour
 )
 
+// InformerGetOptions defines the behavior of how informers are retrieved.
 type InformerGetOptions internal.GetOptions
 
 // InformerGetOption defines an option that alters the behavior of how informers are retrieved.
@@ -201,8 +202,8 @@ type Options struct {
 	// object, this will fall through to Default* settings.
 	ByObject map[client.Object]ByObject
 
-	// NewInformer allows overriding of NewSharedIndexInformer for testing.
-	NewInformer func(toolscache.ListerWatcher, runtime.Object, time.Duration, toolscache.Indexers) toolscache.SharedIndexInformer
+	// newInformer allows overriding of NewSharedIndexInformer for testing.
+	newInformer *func(toolscache.ListerWatcher, runtime.Object, time.Duration, toolscache.Indexers) toolscache.SharedIndexInformer
 }
 
 // ByObject offers more fine-grained control over the cache's ListWatch by object.
@@ -353,7 +354,7 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 				},
 				Transform:             config.Transform,
 				UnsafeDisableDeepCopy: pointer.BoolDeref(config.UnsafeDisableDeepCopy, false),
-				NewInformer:           opts.NewInformer,
+				NewInformer:           opts.newInformer,
 			}),
 			readerFailOnMissingInformer: opts.ReaderFailOnMissingInformer,
 		}
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -45,7 +45,7 @@ type InformersOpts struct {
 	Mapper                meta.RESTMapper
 	ResyncPeriod          time.Duration
 	Namespace             string
-	NewInformer           func(cache.ListerWatcher, runtime.Object, time.Duration, cache.Indexers) cache.SharedIndexInformer
+	NewInformer           *func(cache.ListerWatcher, runtime.Object, time.Duration, cache.Indexers) cache.SharedIndexInformer
 	Selector              Selector
 	Transform             cache.TransformFunc
 	UnsafeDisableDeepCopy bool
@@ -55,7 +55,7 @@ type InformersOpts struct {
 func NewInformers(config *rest.Config, options *InformersOpts) *Informers {
 	newInformer := cache.NewSharedIndexInformer
 	if options.NewInformer != nil {
-		newInformer = options.NewInformer
+		newInformer = *options.NewInformer
 	}
 	return &Informers{
 		config:     config,
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -256,7 +256,7 @@ func (ip *Informers) Peek(gvk schema.GroupVersionKind, obj runtime.Object) (res 
 
 // Get will create a new Informer and add it to the map of specificInformersMap if none exists. Returns
 // the Informer from the map.
-func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, cfg *GetOptions) (bool, *Cache, error) {
+func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, opts *GetOptions) (bool, *Cache, error) {
 	// Return the informer if it is found
 	i, started, ok := ip.Peek(gvk, obj)
 	if !ok {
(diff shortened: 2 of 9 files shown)


pkg/cache/cache.go
```diff
@@ -41,8 +41,23 @@ import (
 var (
 	log               = logf.RuntimeLog.WithName("object-cache")
 	defaultSyncPeriod = 10 * time.Hour
+
+	// BlockUntilSynced determines whether a get request for an informer should block
+	// until the informer's cache has synced.
+	BlockUntilSynced = InformerGetOption(internal.BlockUntilSynced)
```
alvaroaleman, line 47: Can we make this a `func BlockUntilSynced() internal.InformerGetOption { return internal.BlockUntilSynced(true) }`

to
* Avoid the package-level var
* The casting back-and-forth shebang?

pkg/cache/internal/informers.go
```diff
@@ -92,6 +93,25 @@ type tracker struct {
 	Metadata     map[schema.GroupVersionKind]*Cache
 }
 
+// GetOptions provides configuration to customize the behavior when
+// getting an informer.
+type GetOptions struct {
+	// DoNotBlockUntilSynced tells Get() to return the informer immediately,
+	// without waiting for its cache to sync.
+	DoNotBlockUntilSynced bool
```
alvaroaleman, line 101: Lets avoid the negation to simplify this:
```suggestion
        // BlockUntilSynced controls if the informer retrieval will block until the informer is synced. Defaults to `true`
	BlockUntilSynced *bool
```

pkg/cache/internal/informers.go
```diff
@@ -92,6 +93,25 @@ type tracker struct {
 	Metadata     map[schema.GroupVersionKind]*Cache
 }
 
+// GetOptions provides configuration to customize the behavior when
+// getting an informer.
+type GetOptions struct {
```
alvaroaleman, line 98: Why is this type `internal`? What can be configured ends up being part of the public API so this seems to make things needlessly complicated to me. You can use aliases to make to avoid the import loop. This applies to `type InformerGetOption func(*GetOptions)` and `func BlockUntilSynced` as well.

pkg/cache/internal/informers.go
```diff
@@ -290,7 +310,12 @@ func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj r
 		}
 	}
 
-	if started && !i.Informer.HasSynced() {
+	cfg := &GetOptions{}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	if started && !i.Informer.HasSynced() && !cfg.DoNotBlockUntilSynced {
```
alvaroaleman, line 318: Can we add a test for this, please?

maxsmythe: Yeah, to avoid duplicate work, I'm waiting for agreement on form before dedicating time to tests.

pkg/cache/cache.go
maxsmythe: Done, though preserving the boolean arg (since the main concern is to be able to turn off blocking)

pkg/cache/internal/informers.go
maxsmythe: I'm slightly confused by this comment. How can aliasing avoid an import loop if internal needs to see the options?

I agree perpetuating the functional options is a bit unwieldy.

I see 3 alternatives:

* split this out into a shared package to avoid the import cycle (we will probably still need type aliasing on the public package for simplicity)

* Leave the GetOptions struct internal, but do not have the internal package accept functional options (this avoids growth in the arity of the Get() function if more options are added)

* Add a boolean parameter for blocking/non-blocking (probably fine, but might not scale well if more parameters are added later)

I'll do the second option for now, LMK what you think.

pkg/cache/internal/informers.go
maxsmythe: Done

pkg/cache/internal/informers.go
maxsmythe: @alvaroaleman Does this approach LGTY? If so, I can add some unit tests

pkg/cache/internal/informers.go
maxsmythe: Added unit tests

maxsmythe:
@alvaroaleman comments should be addressed

pkg/cache/cache.go
```diff
@@ -166,6 +177,9 @@ type Options struct {
 	//
 	// This is a global setting for all objects, and can be overridden by the ByObject setting.
 	UnsafeDisableDeepCopy *bool
+
+	// NewInformer allows overriding of NewSharedIndexInformer for testing.
+	NewInformer func(toolscache.ListerWatcher, runtime.Object, time.Duration, toolscache.Indexers) toolscache.SharedIndexInformer
```
alvaroaleman, line 182: Can we remove this so we don't have to extend our public API for tests?

vincepri: To clarify, you'd like to use informertest.FakeInformer in the cache package? 

> Drop the type assertion from the FakeInformer package, removing the need for the import but reducing likelihood of drift detection

Could we move the type assertion in the cache package test?

maxsmythe: > To clarify, you'd like to use informertest.FakeInformer in the cache package?

Correct. The test currently uses the fake informer to emulate an informer that never syncs:

```golang
	NonBlockingGetTest(cache.MultiNamespacedCacheBuilder([]string{testNamespaceOne, testNamespaceTwo, "default"}), cache.Options{
		NewInformer: func(lw kcache.ListerWatcher, o runtime.Object, d time.Duration, i kcache.Indexers) kcache.SharedIndexInformer {
			return &controllertest.FakeInformer{Synced: false}
		},
	})
```

With the public interface change in this PR, I can write a test like the above but still have a separate test package (`cache_test`). If we were to make this override ability private, the `cache_test` package would become part of `cache`, so it could access the private methods.

> Could we move the type assertion in the cache package test?

The circular import problem comes from the import chain:

```
 "cache_test" => "sigs.k8s.io/controller-runtime/pkg/controller/controllertest"
 "sigs.k8s.io/controller-runtime/pkg/controller/controllertest" => "cache"
```

If `cache_test` becomes `cache`, the import is circular.

I was wrong when I said it was just a type assertion. It looks like controllertest depends heavily on the "cache" package. Example:

https://github.com/kubernetes-sigs/controller-runtime/blob/a23f3603e1dc14e1d14620b814662f55d1987b02/pkg/controller/controllertest/util.go#L86-L94

maxsmythe: Any feedback as to which approach people would like to see?

maxsmythe:
rebased.

@alvaroaleman @vincepri, let me know what approach you'd like to see for the unit tests, per this thread:

https://github.com/kubernetes-sigs/controller-runtime/pull/2371#discussion_r1252489000

pkg/cache/cache.go
vincepri: @alvaroaleman I'll leave this up to you, I'm ok either way as long as we have tests

alvaroaleman: Allright, lets leave it in

maxsmythe:
Rebased again. @alvaroaleman @vincepri Any further feedback? AFACT all threads are closed and/or awaiting feedback.

ritazh:
_bump_ @vincepri do you mind reviewing this again? Thanks in advance!

pkg/cache/internal/informers.go
maxsmythe: To be clear, I've done the second option and have added unit tests.

pkg/cache/internal/informers.go
```diff
@@ -240,7 +256,7 @@ func (ip *Informers) Peek(gvk schema.GroupVersionKind, obj runtime.Object) (res
 
 // Get will create a new Informer and add it to the map of specificInformersMap if none exists. Returns
 // the Informer from the map.
-func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *Cache, error) {
+func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, cfg *GetOptions) (bool, *Cache, error) {
```
sbueringer, line 259:
```suggestion
func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, opts *GetOptions) (bool, *Cache, error) {
```

pkg/cache/internal/informers.go
```diff
@@ -250,7 +266,12 @@ func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj r
 		}
 	}
 
-	if started && !i.Informer.HasSynced() {
+	shouldBlock := true
+	if cfg.BlockUntilSynced != nil {
+		shouldBlock = *cfg.BlockUntilSynced
+	}
+
+	if started && !i.Informer.HasSynced() && shouldBlock {
```
sbueringer, line 274:
```suggestion
	if shouldBlock && started && !i.Informer.HasSynced() {
```
nit: minor optimization but should avoid the call to HasSynced if shouldBlock is false

pkg/cache/cache.go
```diff
@@ -44,6 +44,17 @@ var (
 	defaultSyncPeriod = 10 * time.Hour
 )
 
+// InformerGetOption defines an option that alters the behavior of how informers are retrieved.
+type InformerGetOption func(*internal.GetOptions)
```
sbueringer, line 48: Is it really okay to have an internal type as part of an exported function?

pkg/cache/cache.go
```diff
@@ -187,6 +198,9 @@ type Options struct {
 	// ByObject restricts the cache's ListWatch to the desired fields per GVK at the specified object.
 	// object, this will fall through to Default* settings.
 	ByObject map[client.Object]ByObject
+
+	// NewInformer allows overriding of NewSharedIndexInformer for testing.
```
sbueringer, line 202: I see two options to avoid exposing this option:

1. Reflection
```go
			v := reflect.ValueOf(informerCache).Elem()
			informersField := v.FieldByName("Informers").Elem()
			newInformerField := informersField.FieldByName("newInformer")
			reflect.NewAt(newInformerField.Type(), unsafe.Pointer(newInformerField.UnsafeAddr())).
				Elem().
				Set(reflect.ValueOf(func(lw kcache.ListerWatcher, o runtime.Object, d time.Duration, i kcache.Indexers) kcache.SharedIndexInformer {
					return &controllertest.FakeInformer{Synced: false}
				}))
```

2. Add an exported func to informerCache to set the NewInformer func. Define an interface with this func in the test, cast to this interface and use the func

sbueringer: I personally prefer hacks in unit tests over exporting funcs only for unit tests.

pkg/cache/cache.go
maxsmythe: I didn't find anything directly on-point, but exporting public constants that use private classes is a best practice for context keys: https://forum.golangbridge.org/t/the-way-to-set-a-context-key-and-value/16311

Not directly on point, but seems related (since I'm guessing we want to control which option functions can be emitted).

Also, the original idea for functional options from rpike has them returning private types:

https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

Using an internal type does have the benefit of preventing callers from defining their own options.

However, better safe than sorry, so added aliasing.

pkg/cache/cache.go
maxsmythe: @sbueringer suggested using reflect in the unit tests. Trying that on. LMK if it LGTY.

maxsmythe: Linking to reflect comment for reference: https://github.com/kubernetes-sigs/controller-runtime/pull/2371/files/35b504b14c79305b5fee50ea039d514b34c39d62#r1297426919

pkg/cache/cache.go
maxsmythe: Ack, used the reflect hack and moved discussion to the related comment thread below ( https://github.com/kubernetes-sigs/controller-runtime/pull/2371/files/35b504b14c79305b5fee50ea039d514b34c39d62#r1252489000 )

LMK what you think

sbueringer: Thx will review in a bit and resolve here

sbueringer:
@maxsmythe looks like there is something slightly of with the unit test.

maxsmythe:
@sbueringer It looks like there are multiple possible structures for the underlying object:

* delegating.defaultCache.Informers (delegating cache wraps normal cache)
* delegating.defaultCache.Informers + delegating.caches[GVK].Informers (delegating + gvk cache)
* delegating.defaultCache.clusterCache.Informers + delegating.defaultCache[NAMESPACE].Informers (multi-namespaced cache)
* the above + delegating.caches[GVK].Informers (multi-NS + gvk cache)

I'm inferring this from https://github.com/kubernetes-sigs/controller-runtime/blob/304027bcbe4b3f6d582180aec5759eb4db3f17fd/pkg/cache/cache.go#L267-L301

We could make `newInformer` a private member of the `cache.Options` struct. It would need to be "public" for the `internal` package, but that's an `internal` package. That may make things not-settable by users but less brittle to implementation changes (and statically analyzable). WDYT?

sbueringer:
Ah yup makes sense. I probably didn't test all variants somehow.

> We could make newInformer a private member of the cache.Options struct. It would need to be "public" for the internal package, but that's an internal package. That may make things not-settable by users but less brittle to implementation changes (and statically analyzable). WDYT?

+1 Sounds like the best option

pkg/cache/cache.go
sbueringer: Sounds good. Thank you for the additional context!

sbueringer:
@maxsmythe Do you have time to address the issue? (context: would be nice to get this PR into the upcoming v0.16 release)

maxsmythe:
Yep! Sorry, was out Friday, aiming to push something in the next ~hour

sbueringer:
> Yep! Sorry, was out Friday, aiming to push something in the next ~hour

No worries, thank you!

maxsmythe:
For sure! Took a bit longer than an hour... still needed to get around golang's visibility restrictions, and was unable to do that without turning `newInformer` into a pointer-to-function. Trying to use a function-type variable resulted in a segfault.

maxsmythe:
@sbueringer unit tests pass

sbueringer:
Perfect. Thank you very much!

/lgtm

wyike:
pull-controller-runtime-apidiff  job seems still failing

sbueringer:
> pull-controller-runtime-apidiff job seems still failing

This is expected behavior. Although we should probably change this to a breaking change, given the change in API

wyike:
> > pull-controller-runtime-apidiff job seems still failing
> 
> This is expected behavior. Not sure if we should flag this PR as a breaking change as we're just adding a vararg so it doesn't break anyone using the changed funcs.

I see. In this case, apidiff job failure is acceptable.

alvaroaleman (approved):

alvaroaleman:
Thanks for sticking to it and getting this done, @maxsmythe !

PR #2387, opened 2023-06-23
✨ fake: simply WithStatusSubresource() signature

nicks:
simply WithStatusSubresource() signature

runtime.Object is a subinterface of client.Object,
and is all that's needed here
    
you really need a client to get a client.Object,
so it doesn't really make sense to use a client.Object
to initialize a client.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -115,7 +115,7 @@ type ClientBuilder struct {
 	initObject            []client.Object
 	initLists             []client.ObjectList
 	initRuntimeObjects    []runtime.Object
-	withStatusSubresource []client.Object
+	withStatusSubresource []runtime.Object
 	objectTracker         testing.ObjectTracker
 	interceptorFuncs      *interceptor.Funcs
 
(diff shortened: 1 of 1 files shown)


pkg/client/fake/client.go
```diff
@@ -115,7 +115,7 @@ type ClientBuilder struct {
 	initObject            []client.Object
 	initLists             []client.ObjectList
 	initRuntimeObjects    []runtime.Object
-	withStatusSubresource []client.Object
+	withStatusSubresource []runtime.Object
```
alvaroaleman, line 118: Sorry, I do not understand what problem you are solving with this. Could you please give a real-world example where the current typing is an issue?

nicks: `runtime.Object` tends to be more broadly supported in the apimachinery, e.g. `https://pkg.go.dev/k8s.io/apimachinery@v0.27.3/pkg/runtime#Scheme.New

alvaroaleman: That is not an example of a real world use case. All types I've seen so far people want to use the client for implement client.Object.

nicks: re: "That is not an example of a real world use case. "

seems likely that your definition of "real world use case" is different than mine. can you elaborate on what you're looking for?

nicks: sure, i added a test! the downstream code is obviously a bit more complicated, but i think the test captures the essence of it.

alvaroaleman: Ah. Well, you can type assert it.

The reason we use `client.Object` over `runtime.Object` is that the latter is fullfilled by both lists and single objects (e.G. a `&corev1.PodList` is a valid `runtime.Object`). As a result, if we were to use `runtime.Object` instead of the current `client.Object` and `client.ObjectList`, the compiler would allow passing lists where we expect a single object and vice versa and people would only notice the issue at runtime. 


Additionally, `client.Object` and `client.ObjectList` preseve additional type information.

alvaroaleman: /hold

nicks: ok! thanks for considering

PR #2393, opened 2023-07-01
⚠ Bump to k8s.io/ v0.28.0-alpha.3

sbueringer:
Let's already bump to v0.28 alpha3 to get some experience with 0.28 to find issues early


pkg/manager/manager_test.go
```diff
@@ -500,7 +501,7 @@ var _ = Describe("manger.Manager", func() {
 			It("should use the specified ResourceLock", func() {
 				m, err := New(cfg, Options{
 					LeaderElection:             true,
-					LeaderElectionResourceLock: resourcelock.ConfigMapsLeasesResourceLock,
+					LeaderElectionResourceLock: resourcelock.LeasesResourceLock,
```
sbueringer, line 504: Upstream now only supports the Leases lock: https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/interface.go#L166-L189

@alvaroaleman @vincepri I wonder if we should just drop the LeaderElectionResourceLock field. Doesn't make much sense to keep it if it can only be set to `LeasesResourceLock`. WDYT?

alvaroaleman (approved):

PR #2406, opened 2023-07-13
✨: pkg/cache: add options for cache miss policy

### LINKED ISSUE #2397: Proposal: Allow Setting Cache Miss Policy in Cache Options
Today, when a user gets a `client.Client` from a `Manager` (with `mgr.GetClient()`), they get a cache-backed client that has some surprising behavior - when a cache miss occurs, an informer is spun up to feed data into the cache. This means that one `client.Get()` call for one resource will, by default, start a cluster-scoped watch on all objects of that type and keep them in memory forever. While this is an understandable way for the cache to work, I've seen it bite folks who did not expect this behavior. While I don't think that this should be the default, at a minimum I'd love to see the cache miss policy be configurable at creation time so that folks can opt out of this behavior. Today it is possible to opt out of caching particular resources, but that's done with a hard-coded deny-list, which means if you ever forget to update that list, you get bitten.

I'd like to propose adding a new field in the `cache.Options`:

```go
// pkg/cache/cache.go

// Options are the optional arguments for creating a new InformersMap object.
type Options struct {
    // ...
    
    // MissPolicy determines how the cache should behave when a Get finds no
    // entry in the cache or a List captures no items. See the CacheMissPolicies for
    // documentation for each policy.
    MissPolicy CacheMissPolicy
}

type CacheMissPolicy string

const (
    // Forward configures the cache to forward a cache miss to the client, unchanged,
    // meaning that the client gets an errors.NotFound for a Get and an empty response
    // for a List.
    Forward CacheMissPolicy = "forward"

    // Backfill configures the cache to spin up an informer for a resource when the first
    // request for that resource comes in. This means that a Get or a List may take longer
    // than normal to succeed on the first invocation as the cache waits until it's fully back-
    // filled.
    Backfill CacheMissPolicy = "backfill"

    // LiveLookup configures the cache to issue a live client call to the server for requests
    // that do not correspond to resources that have been explicitly configured to be cached.
    LiveLookup CacheMissPolicy = "liveLookup"
)

/cc @alvaroaleman @vincepri

### PR DESCRIPTION
**Pull request by stevekuznetsov**
This commit allows users to opt out of the "start informers in the background" behavior that the current cache implementation uses. Additionally, when opting out of this behavior, the client can be configured to do a live lookup on a cache miss. The default behaviors are:

  pkg/cache: backfill data on a miss (today's default, unchanged)
  pkg/client: live lookup when cache is configured to miss

Fixes #2397

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -145,6 +145,15 @@ type Options struct {
 	// instead of `reconcile.Result{}`.
 	SyncPeriod *time.Duration
 
+	// ReaderFailOnMissingInformer configures the cache to return a ErrResourceNotCached error when a user
+	// requests, using Get() and List(), a resource the cache does not already have an informer for.
+	//
+	// This error is distinct from an errors.NotFound.
+	//
+	// Defaults to false, which means that the cache will start a new informer
+	// for every new requested resource.
+	ReaderFailOnMissingInformer bool
+
 	// DefaultNamespaces maps namespace names to cache configs. If set, only
 	// the namespaces in here will be watched and it will by used to default
 	// ByObject.Namespaces for all objects if that is nil.
@@ -329,6 +338,7 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 				Transform:             config.Transform,
 				UnsafeDisableDeepCopy: pointer.BoolDeref(config.UnsafeDisableDeepCopy, false),
 			}),
+			readerFailOnMissingInformer: opts.ReaderFailOnMissingInformer,
 		}
 	}
 }
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -46,11 +46,28 @@ func (*ErrCacheNotStarted) Error() string {
 	return "the cache is not started, can not read objects"
 }
 
+var _ error = (*ErrCacheNotStarted)(nil)
+
+// ErrResourceNotCached indicates that the resource type
+// the client asked the cache for is not cached, i.e. the
+// corresponding informer does not exist yet.
+type ErrResourceNotCached struct {
+	GVK schema.GroupVersionKind
+}
+
+// Error returns the error
+func (r ErrResourceNotCached) Error() string {
+	return fmt.Sprintf("%s is not cached", r.GVK.String())
+}
+
+var _ error = (*ErrResourceNotCached)(nil)
+
 // informerCache is a Kubernetes Object cache populated from internal.Informers.
 // informerCache wraps internal.Informers.
 type informerCache struct {
 	scheme *runtime.Scheme
 	*internal.Informers
+	readerFailOnMissingInformer bool
 }
 
 // Get implements Reader.
@@ -60,7 +77,7 @@ func (ic *informerCache) Get(ctx context.Context, key client.ObjectKey, out clie
 		return err
 	}
 
-	started, cache, err := ic.Informers.Get(ctx, gvk, out)
+	started, cache, err := ic.getInformerForKind(ctx, gvk, out)
 	if err != nil {
 		return err
 	}
@@ -78,7 +95,7 @@ func (ic *informerCache) List(ctx context.Context, out client.ObjectList, opts .
 		return err
 	}
 
-	started, cache, err := ic.Informers.Get(ctx, *gvk, cacheTypeObj)
+	started, cache, err := ic.getInformerForKind(ctx, *gvk, cacheTypeObj)
 	if err != nil {
 		return err
 	}
@@ -124,7 +141,7 @@ func (ic *informerCache) objectTypeForListObject(list client.ObjectList) (*schem
 	return &gvk, cacheTypeObj, nil
 }
 
-// GetInformerForKind returns the informer for the GroupVersionKind.
+// GetInformerForKind returns the informer for the GroupVersionKind. If no informer exists, one will be started.
 func (ic *informerCache) GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind) (Informer, error) {
 	// Map the gvk to an object
 	obj, err := ic.scheme.New(gvk)
@@ -139,7 +156,7 @@ func (ic *informerCache) GetInformerForKind(ctx context.Context, gvk schema.Grou
 	return i.Informer, nil
 }
 
-// GetInformer returns the informer for the obj.
+// GetInformer returns the informer for the obj. If no informer exists, one will be started.
 func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (Informer, error) {
 	gvk, err := apiutil.GVKForObject(obj, ic.scheme)
 	if err != nil {
@@ -151,6 +168,18 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
 		return nil, err
 	}
 	return i.Informer, nil
+}
+
+func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
+	if ic.readerFailOnMissingInformer {
+		cache, started, ok := ic.Informers.Peek(gvk, obj)
+		if !ok {
+			return false, nil, &ErrResourceNotCached{GVK: gvk}
+		}
+		return started, cache, nil
+	}
+
+	return ic.Informers.Get(ctx, gvk, obj)
 }
 
 // NeedLeaderElection implements the LeaderElectionRunnable interface
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -230,7 +230,8 @@ func (ip *Informers) WaitForCacheSync(ctx context.Context) bool {
 	return cache.WaitForCacheSync(ctx.Done(), ip.getHasSyncedFuncs()...)
 }
 
-func (ip *Informers) get(gvk schema.GroupVersionKind, obj runtime.Object) (res *Cache, started bool, ok bool) {
+// Peek attempts to get the informer for the GVK, but does not start one if one does not exist.
+func (ip *Informers) Peek(gvk schema.GroupVersionKind, obj runtime.Object) (res *Cache, started bool, ok bool) {
 	ip.mu.RLock()
 	defer ip.mu.RUnlock()
 	i, ok := ip.informersByType(obj)[gvk]
@@ -241,7 +242,7 @@ func (ip *Informers) get(gvk schema.GroupVersionKind, obj runtime.Object) (res *
 // the Informer from the map.
 func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *Cache, error) {
 	// Return the informer if it is found
-	i, started, ok := ip.get(gvk, obj)
+	i, started, ok := ip.Peek(gvk, obj)
 	if !ok {
 		var err error
 		if i, started, err = ip.addInformerToMap(gvk, obj); err != nil {
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -77,10 +77,12 @@ type CacheOptions struct {
 	// Reader is a cache-backed reader that will be used to read objects from the cache.
 	// +required
 	Reader Reader
-	// DisableFor is a list of objects that should not be read from the cache.
+	// DisableFor is a list of objects that should never be read from the cache.
+	// Objects configured here always result in a live lookup.
 	DisableFor []Object
 	// Unstructured is a flag that indicates whether the cache-backed client should
 	// read unstructured objects or lists from the cache.
+	// If false, unstructured objects will always result in a live lookup.
 	Unstructured bool
 }
 
@@ -342,9 +344,11 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
 	if isUncached, err := c.shouldBypassCache(obj); err != nil {
 		return err
 	} else if !isUncached {
+		// Attempt to get from the cache.
 		return c.cache.Get(ctx, key, obj, opts...)
 	}
 
+	// Perform a live lookup.
 	switch obj.(type) {
 	case runtime.Unstructured:
 		return c.unstructuredClient.Get(ctx, key, obj, opts...)
@@ -362,9 +366,11 @@ func (c *client) List(ctx context.Context, obj ObjectList, opts ...ListOption) e
 	if isUncached, err := c.shouldBypassCache(obj); err != nil {
 		return err
 	} else if !isUncached {
+		// Attempt to get from the cache.
 		return c.cache.List(ctx, obj, opts...)
 	}
 
+	// Perform a live lookup.
 	switch x := obj.(type) {
 	case runtime.Unstructured:
 		return c.unstructuredClient.List(ctx, obj, opts...)
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,7 @@ generate: $(CONTROLLER_GEN) ## Runs controller-gen for internal types for config
 
 .PHONY: clean
 clean: ## Cleanup.
+	$(GOLANGCI_LINT) cache clean
 	$(MAKE) clean-bin
 
 .PHONY: clean-bin
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -18,6 +18,7 @@ package cache_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -117,6 +118,11 @@ func deletePod(pod client.Object) {
 var _ = Describe("Informer Cache", func() {
 	CacheTest(cache.New, cache.Options{})
 })
+
+var _ = Describe("Informer Cache with ReaderFailOnMissingInformer", func() {
+	CacheTestReaderFailOnMissingInformer(cache.New, cache.Options{ReaderFailOnMissingInformer: true})
+})
+
 var _ = Describe("Multi-Namespace Informer Cache", func() {
 	CacheTest(cache.New, cache.Options{
 		DefaultNamespaces: map[string]cache.Config{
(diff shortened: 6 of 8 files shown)

### CODE REVIEW

File: pkg/cache/cache.go [hunk 1]
```diff
@@ -166,8 +166,38 @@ type Options struct {
...
+	MissPolicy CacheMissPolicy
 }
 
+type CacheMissPolicy string
+
+const (
+	// CacheMissPolicyFail configures the cache to return a ErrResourceNotCached error when a user
+	// requests a resource the cache is not configured to hold. This error is distinct from an
+	// errors.NotFound.
+	CacheMissPolicyFail CacheMissPolicy = "fail"
```
vincepri, line 182: Can we call this `Strict` and call out in the comment that's the default?

pkg/cache/informer_cache.go
```diff
@@ -145,13 +145,32 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
...
+	case CacheMissPolicyFail:
+		var ok bool
+		cache, started, ok = ic.Informers.Peek(gvk, obj)
+		if !ok {
+			err = &ErrResourceNotCached{gvk: gvk}
+		}
+	case CacheMissPolicyBackfill:
+		fallthrough
+	default:
+		started, cache, err = ic.Informers.Get(ctx, gvk, obj)
```
vincepri, line 169: Shouldn't the `Get` go into backfill, and we error out in any other case?

stevekuznetsov: Go's type system is not expressive enough to capture that there are only two valid options for the miss policy here. I wanted to write this in a way such that the reviewers would be more confident that the default behavior did not change from the previous behavior. No strong opinions here but not having a valid miss policy here is a programmer error / panic and should not be possible at runtime.

vincepri: Let's not have two options that do the same thing, if we have a default == empty which is to backfill we can try to express that through an empty string option maybe, or just offer `Fail` for now as a non-empty option?

pkg/client/client.go
```diff
@@ -82,8 +83,26 @@ type CacheOptions struct {
...
+type CacheMissPolicy string
+
+const (
+	// CacheMissPolicyFail configures the client to return an errors.NotFound error when a user
+	// requests a resource the cache is not configured to hold.
+	CacheMissPolicyFail CacheMissPolicy = "fail"
+
+	// CacheMissPolicyLiveLookup configures the client to issue a live client lookup to get data
+	// for a resource that is not cached.
+	CacheMissPolicyLiveLookup CacheMissPolicy = "lookup"
```
vincepri, line 103:
```suggestion
	CacheMissPolicyLiveLookup CacheMissPolicy = "live-lookup"
```

pkg/client/client.go
```diff
@@ -353,12 +382,22 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
 	}
 }
 
+// shouldReturnCacheErr determines if we should return the error we got from the cache, which is always except
```
vincepri, line 385: which is always?

pkg/client/client.go
```diff
@@ -353,12 +382,22 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
...
+// for the case that we're configured to do live lookups and got a cache.ErrResourceNotCached
```
vincepri, line 386: If we're doing live lookups, why would we get a `cache.ErrResourceNotCached`?

pkg/client/client.go
vincepri: ?

pkg/client/client.go
```diff
@@ -82,8 +83,26 @@ type CacheOptions struct {
 	// Unstructured is a flag that indicates whether the cache-backed client should
 	// read unstructured objects or lists from the cache.
 	Unstructured bool
+	// MissPolicy determines how the client should behave when the underlying cache is
+	// configured to fail on a miss. See the CacheMissPolicies for documentation for each
+	// policy.
+	MissPolicy CacheMissPolicy
```
vincepri, line 89: Hm, does this only apply when the underlying cache is configured in fail mode?

pkg/cache/informer_cache.go
stevekuznetsov: There are no two options that do the same thing. You just can't express that the type has two valid enum cases, and not the empty string.

pkg/client/client.go
stevekuznetsov: Will reword, is not clear

pkg/client/client.go
stevekuznetsov: Will reword, is not clear

pkg/client/client.go
stevekuznetsov: Yes, the GoDoc mentions this. There is no possibility of a cache miss in the backfill case.

pkg/cache/cache.go
```diff
@@ -166,8 +166,31 @@ type Options struct {
 	//
 	// This is a global setting for all objects, and can be overridden by the ByObject setting.
 	UnsafeDisableDeepCopy *bool
+
+	// MissPolicy determines how the cache should behave when no informers are started for the
+	// resource type a client is asking for in a Get or a List. See the CacheMissPolicies for
+	// documentation for each policy.
+	MissPolicy MissPolicy
```
alvaroaleman, line 173: Lets godoc the default

pkg/client/client.go
```diff
@@ -185,9 +205,15 @@ func newClient(config *rest.Config, options Options) (*client, error) {
 		return c, nil
 	}
 
+	// Default to looking up items when missing in the cache
+	if options.Cache.MissPolicy == "" {
+		options.Cache.MissPolicy = CacheMissPolicyLiveLookup
```
alvaroaleman, line 210: I thought we wanted to default to fail?

pkg/client/client.go
```diff
@@ -82,8 +83,27 @@ type CacheOptions struct {
 	// Unstructured is a flag that indicates whether the cache-backed client should
 	// read unstructured objects or lists from the cache.
 	Unstructured bool
+	// MissPolicy determines how the client should behave when the underlying cache is
+	// configured to fail on a miss. See the CacheMissPolicies for documentation for each
+	// policy.
+	MissPolicy CacheMissPolicy
```
alvaroaleman, line 89: Same here: Lets godoc the default

pkg/client/client.go
stevekuznetsov: I considered this but it seems like the more surprising / breaking change. Brings it more inline with k8s lister / live client split, though - is that why you think fail should be default?

pkg/cache/internal/informers.go
```diff
@@ -282,7 +283,7 @@ func (ip *Informers) get(gvk schema.GroupVersionKind, obj runtime.Object) (res *
 // the Informer from the map.
 func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *Cache, error) {
 	// Return the informer if it is found
-	i, started, ok := ip.get(gvk, obj)
+	i, started, ok := ip.Peek(gvk, obj)
```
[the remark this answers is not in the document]
stevekuznetsov: No, this is written explicitly for the case where the user has not made that option. In the general case it's not reasonable to ask someone _a priori_ to list every possible GVK they do not want to cache. Much easier to ask the cache only to do things it's asked to do.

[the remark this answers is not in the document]
stevekuznetsov: That does seem right, let me check to see why tests aren't failing on this.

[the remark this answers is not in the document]
stevekuznetsov: Ah - last commit should do it. The `GetInformer*` methods, since they're used to indicate "start caching", will always start if not found.

pkg/client/client.go
alvaroaleman: I think falling back might lead to questions, because based on if there is a cache or not the request will get a fast or slow response that is incoherent or not - That is less likely to happen if we only do that when ppl explicitly opted into this

stevekuznetsov:
Seeing two failures in test that must be unrelated, kind of confused. Would appreciate knowing if these are known flakes?

pkg/cache/informer_cache.go
joelanford: > the value of the miss policy may only ever be CacheMissPolicyBackfill or CacheMissPolicyFail.

For this reason, I don't see what's wrong with a `panic` in the case that the value is not one of those two options.

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,25 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
...
+	case MissPolicyFail:
+		var ok bool
+		cache, started, ok = ic.Informers.Peek(gvk, obj)
+		if !ok {
+			err = &errors.ErrResourceNotCached{GVK: gvk}
+		}
+	case MissPolicyBackfill:
+		fallthrough
+	default:
+		started, cache, err = ic.Informers.Get(ctx, gvk, obj)
```
vincepri, line 170:
```suggestion
	case MissPolicyBackfill:
		return ic.Informers.Get(ctx, gvk, obj)
```

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,25 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
...
 }
 
+func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
+	var cache *internal.Cache
+	var started bool
+	var err error
+	switch ic.missPolicy {
+	case MissPolicyFail:
+		var ok bool
+		cache, started, ok = ic.Informers.Peek(gvk, obj)
```
vincepri, line 163:
```suggestion
		if cache, started, ok := ic.Informers.Peek(gvk, obj); !ok {
```

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,25 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
...
+		if !ok {
+			err = &errors.ErrResourceNotCached{GVK: gvk}
```
vincepri, line 165: Just return here?

pkg/client/client.go
vincepri: Sec, what's the current behavior?

pkg/client/client.go
```diff
@@ -338,7 +365,10 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
 	if isUncached, err := c.shouldBypassCache(obj); err != nil {
 		return err
 	} else if !isUncached {
-		return c.cache.Get(ctx, key, obj, opts...)
+		err := c.cache.Get(ctx, key, obj, opts...)
+		if c.shouldReturnCacheErr(err) {
```
vincepri, line 369:
```suggestion
		if err := c.cache.Get(ctx, key, obj, opts...); c.shouldReturnCacheErr(err) {
```

pkg/client/client.go
```diff
@@ -353,12 +383,22 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
 	}
 }
 
+// shouldReturnCacheErr determines if we should return the error we got from the cache, which we do in every case
+// except for when we're configured to do live lookups and the cache returns cache.ErrResourceNotCached
+func (c *client) shouldReturnCacheErr(err error) bool {
+	var errNotCached *cacheerrors.ErrResourceNotCached
+	return c.cacheMissPolicy != CacheMissPolicyLiveLookup || !errors.As(err, &errNotCached)
```
vincepri, line 390: Two comments:

- Can we use https://pkg.go.dev/errors#example-Is ?
- The comment explains this condition as an AND, although we're doing OR with negatives, can we flip this?

pkg/cache/cache.go
vincepri: The failure explained here is `ErrResourceNotCached`, which is a soft error given we can always do a live lookup manually from a user side, or rework the cache to cache the new resource. IMO a Strict mode signals that the cache will never return something that it's not configured to hold and we can expand this mode later if we want to. The word Fail makes me think about a terminal failure, or panic, which isn't the case.

@sbueringer @alvaroaleman ?

pkg/cache/informer_cache.go
stevekuznetsov: This won't set the `var cache, started` above though?

pkg/cache/informer_cache.go
stevekuznetsov: Yeah, didn't know what was easier to read, one return / control flow or inline returns

pkg/client/client.go
stevekuznetsov: There is no possibility of this happening currently, as the cache always backfills.

pkg/client/client.go
stevekuznetsov: No, you cannot use `errors.Is` as that checks for pointer equality (for e.g. a sentinel/global error value).

Yes, I can flip the boolean logic for readability.

pkg/cache/informer_cache.go
stevekuznetsov: I mean, what's the _value_ in a panic?

pkg/cache/cache.go
stevekuznetsov: As far as the cache is concerned, it does fail. Up to the clients to do what they want with it. I don't have a strong opinion, as long as the options read like English and are parallel to each other. `"fail"` and `"backfill"` are verbs and parallel. What would you like to pair with `"strict"`?

pkg/client/client.go
stevekuznetsov: Perhaps we should re-name this from "miss" policy in both cases, since it's not truly a cache miss in the 404 sense, but a "cache not configured to hold this resource type" miss.

pkg/cache/informer_cache.go
stevekuznetsov:
```
 pkg/cache/informer_cache.go:161:10: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (revive)
		} else {
			return started, cache, nil
		} 
```

your linter does not like this

pkg/client/client.go
stevekuznetsov: Refactored to make this wording more clear.

pkg/client/client.go
```diff
@@ -82,8 +83,27 @@ type CacheOptions struct {
 	// Unstructured is a flag that indicates whether the cache-backed client should
 	// read unstructured objects or lists from the cache.
 	Unstructured bool
+	// UncachedResourcePolicy determines how the client should behave when the underlying cache is
+	// configured to fail when encountering an unknown resource type. See the CacheMissPolicies for
```
ncdc, line 87: Should `CacheMissPolicies` be `UncachedResourcePolicies`?

pkg/client/client.go
```diff
@@ -353,12 +382,21 @@ func (c *client) Get(ctx context.Context, key ObjectKey, obj Object, opts ...Get
 	}
 }
 
+// shouldReturnCacheErr determines if we should return the error we got from the cache, which we do in every case
+// except for when we're configured to do live lookups and the cache returns cache.ErrResourceNotCached
+func (c *client) shouldReturnCacheErr(err error) bool {
+	var errNotCached *cacheerrors.ErrResourceNotCached
+	return !(c.cacheMissPolicy == UncachedResourcePolicyLiveLookup && errors.As(err, &errNotCached))
```
ncdc, line 389: This hurts my brain 😄.

joelanford:
Am I reading correctly that the default policy in `cache.go` is [Backfill](https://github.com/kubernetes-sigs/controller-runtime/blob/a4f7424dd767555eac0897017f2f26ff4ce55b4d/pkg/cache/cache.go#L296-L298) and the default policy in `client.go` is [Fail](https://github.com/kubernetes-sigs/controller-runtime/blob/a4f7424dd767555eac0897017f2f26ff4ce55b4d/pkg/client/client.go#L208-L211)?

If so, a few questions:
1. Why are they different?
2. Why do we need to set the default in two different places?

If we're changing the default to Fail, we should update the PR emoji to ⚠ to reflect that this is a breaking change.

alvaroaleman:
> If we're changing the default to Fail, we should update the PR emoji to ⚠ to reflect that this is a breaking change.

This is confusing: The setting in the client is for when the cache tells it that it doesn't have this resource and won't set up an informer. If that client setting is not fail, then it will read from the api directly instead of the cache. The overall default behavior should be unchanged.


Do you think there is a way we could better clarify the above?

pkg/client/client.go
stevekuznetsov: I had it the other way around but that broke Vince's brain, not sure if there's a good middle ground :|

stevekuznetsov:
@joelanford this PR adds configuration to both the client and the cache. Regardless of what you configure the cache to do, the client acts independently. We're not "defining the default in two places" since we're configuring two different bits entirely.

The matrix looks like:

<table>
<tbody>
<tr>
<td>&nbsp;</td>
<td colspan="2">Client <code>UncachedResourcePolicy</code></td>
</tr>
<tr>
<td>Cache <code>UnknownResourcePolicy</code></td>
<td><code>"fail"</code></td>
<td><code>"live-lookup"</code></td>
</tr>
<tr>
<td align="right"><code>"backfill"</code></td>
<td align="center"> 1 </td>
<td align="center"> 2 </td>
</tr>
<tr>
<td align="right"><code>"fail"</code></td>
<td align="center"> 3 </td>
<td align="center"> 4 </td>
</tr>
</tbody>
</table>

The states are:

1. This is the default behavior. The cache always spins up new informers (like today) and the client, while configured to fail in case of an uncached resource, will never encounter that.
2. The cache always spins up new informers and the client's configuration is similarly uninteresting.
3. The cache does not spin up new informers, and the client passes along the error. This is equivalent to using an upstream `Lister`.
4. The cache does not spin up new informers, but the client does a live lookup. Might be more ergonomic.

My intent here is to allow using option 3 and `APIReader()` to get the equivalent of upstream's `Lister` + `Clientset`.

joelanford:
Oh and a second question. In the case that `cache.UnknownResourcePolicy` is backfill, that makes `client.UncachedResourcePolicy` moot (unless it also applies to `DisableFor` objects; see previous comment). IMO, its a poor UX for two configurations to interact in a way where one configuration causes the other configuration to be ignored. Is there a way to avoid that?

stevekuznetsov:
@joelanford 

> It looks like the behavior of client.CacheOptions.DisableFor before and after this change is to go to the direct client,

Good catch. I can update those and add to the docs for `DisableFor`. Seems like we need the live lookup client policy to be default.

> IMO, its a poor UX for two configurations to interact in a way where one configuration causes the other configuration to be ignored. Is there a way to avoid that?

I'm open to suggestions here.

vincepri:
> IMO, its a poor UX for two configurations to interact in a way where one configuration causes the other configuration to be ignored. Is there a way to avoid that?

In which case the configuration would be ignored? Maybe it's a matter of documentation, but `DisableFor` is very specific and should always take precedence.

stevekuznetsov:
@vincepri I understood @joelanford to mean that if you configure `client.Options.Cache.UncachedResourcePolicy` but do *not* set `cache.Options.UnknownResourcePolicy=fail` then the client's configuration is obviated since it's configuring for a case that never occurs.

vincepri:
🤔 Yeah that needs to be fixed probably, the client's cache configuration portion should inform the internal cache and those option need to be related to each other.

vincepri:
Could we return an error in case the options don't match?

stevekuznetsov:
@vincepri from where? AFAICT it's nontrivial to get at both of those option sets and, again, the function constructors are opaque, so we couldn't do it every time - is that ok?

stevekuznetsov:
@joelanford updated the client options to affect `client.Options.Cache.DisableFor`
@vincepri added some best-effort validation to the `manager.New()` code

pkg/cache/cache.go
```diff
@@ -166,6 +166,14 @@ type Options struct {
 	//
 	// This is a global setting for all objects, and can be overridden by the ByObject setting.
 	UnsafeDisableDeepCopy *bool
+
+	// FailOnUnknownResource configures the cache to return a ErrResourceNotCached error when a user
+	// requests a resource the cache is not configured to hold. This error is distinct from an
+	// errors.NotFound.
+	//
+	// Defaults to false, which means that the cache will start a new informer
+	// for every new requested resource.
```
stevekuznetsov, line 175: Might be worth it to note that by default this creates a cluster-wide cache to help folks understand why they might choose to set it to `true` if they don't have the background we do

pkg/cache/informer_cache.go
```diff
@@ -152,6 +153,19 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object) (In
 	return i.Informer, err
 }
 
+func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
+	switch {
+	case ic.failOnUnknown:
```
stevekuznetsov, line 158: nit: now that this is a bool, seems like `if` would be clearer than `switch` but not a big deal

stevekuznetsov (commented):
Tiny comments, but LGTM

pkg/cache/errors/errors.go
```diff
@@ -0,0 +1,20 @@
+package errors
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// ErrResourceNotCached indicates that the resource type
```
alvaroaleman, line 9: I know I told you something else over Slack, but I realized there is another case of a cache error which is not in a subpackage: https://github.com/kubernetes-sigs/controller-runtime/blob/b1d6919d3e12fa85a119dd9792bdfdc17bdf8c3b/pkg/cache/informer_cache.go#L42

Lets maybe move this into the cache package, too?

alvaroaleman:
/retitle ✨: pkg/cache: add options for cache miss policy

alvaroaleman (commented):
One small comment, other than that LGTM

stevekuznetsov:
@vincepri did we want to rebase this?

sbueringer:
I think apart from a rebase and the nit above we're ready to merge

sbueringer:
Hey folks,
we're getting closer to the Kubernetes 1.28 / controller-runtime v0.16. 
Would be really nice to get this merged before :)

sbueringer:
@alvaroaleman @vincepri Rebased, should be ready now

pkg/cache/cache.go
```diff
@@ -145,6 +145,15 @@ type Options struct {
...
 	SyncPeriod *time.Duration
 
+	// FailOnUnknownResource configures the cache to return a ErrResourceNotCached error when a user
+	// requests, using Get() and List(), a resource the cache does not already have an informer for.
+	//
+	// This error is distinct from an errors.NotFound.
+	//
+	// Defaults to false, which means that the cache will start a new informer
+	// for every new requested resource.
+	FailOnUnknownResource bool
```
inteon, line 155: For me, it's not directly clear what an `UnknownResource` is.
Could we maybe rename this to `FailReadWithoutRegisteredInformer`?

alvaroaleman: Maybe `ReaderFailOnMissingInformer`?

sbueringer: Sounds good! Will update

sbueringer: Done

stevekuznetsov: Does the GoDoc here clear it up sufficiently? We've gone back and forth on the names here a couple of times and I'm not sure we will find something perfect. If the doc makes it sufficiently clear I think it's probably a good stopping place.

stevekuznetsov: (disregard, thanks @sbueringer )

PR #2412, opened 2023-07-19
⚠ fake client: use correct RBAC apiGroup name when deciding if to allow unconditional updates

jaideepr97:
The fake client logic runs through a switch case, switching on resource group to decide if unconditional update should be allowed for a particular resource or not
However, for rbac resources it checks against the string "rbac" as the group name, as opposed to "rbac.authorization.k8s.io",  which is the actual group name for rbac resources

This issue currently results in an "object modified" error when the fake client tries to update any rbac resource, as it fails the `allowsUnconditionalUpdate` check

This PR fixes that by updating the rbac group name to the correct value

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1177,6 +1177,8 @@ func allowsUnconditionalUpdate(gvk schema.GroupVersionKind) bool {
 		case "PodSecurityPolicy":
 			return true
 		}
+	case "rbac":
+		fallthrough
 	case "rbac.authorization.k8s.io":
 		switch gvk.Kind {
 		case "ClusterRole", "ClusterRoleBinding", "Role", "RoleBinding":
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1177,7 +1177,7 @@ func allowsUnconditionalUpdate(gvk schema.GroupVersionKind) bool {
 		case "PodSecurityPolicy":
 			return true
 		}
-	case "rbac":
+	case "rbac.authorization.k8s.io":
 		switch gvk.Kind {
 		case "ClusterRole", "ClusterRoleBinding", "Role", "RoleBinding":
 			return true
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1177,8 +1177,6 @@ func allowsUnconditionalUpdate(gvk schema.GroupVersionKind) bool {
 		case "PodSecurityPolicy":
 			return true
 		}
-	case "rbac":
-		fallthrough
 	case "rbac.authorization.k8s.io":
 		switch gvk.Kind {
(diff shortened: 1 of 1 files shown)


pkg/client/fake/client.go
```diff
@@ -1177,7 +1177,7 @@ func allowsUnconditionalUpdate(gvk schema.GroupVersionKind) bool {
 		case "PodSecurityPolicy":
 			return true
 		}
-	case "rbac":
+	case "rbac.authorization.k8s.io":
```
vincepri, line 1180: Add both?

sbueringer: Why both?

jaideepr97: @vincepri you mean for backward compatibility?

vincepri: Yes

jaideepr97: @vincepri updated, thanks!

pkg/client/fake/client.go
```diff
@@ -1178,6 +1178,8 @@ func allowsUnconditionalUpdate(gvk schema.GroupVersionKind) bool {
 			return true
 		}
 	case "rbac":
```
sbueringer, line 1180: @vincepri The part I don't understand is why do we have to support both. There is no "rbac" API group in Kubernetes.

Is this to not break folks who are using this with Unstructured and are using a non-existing "rbac" api group for ClusterRole, etc.?

I would argue it's better to not continue to support a non-existing api group.

No objection though if you think it's the right thing to do. Would be good to add a godoc though why we are doing it

jaideepr97: I would agree
I don't know why it was "rbac" in the first place for this long, and how it has not been corrected before this

jaideepr97: I would just like to see this fixed either way

sbueringer: Yup, no worries. Let's see what Vince thinks and then we get it merged one way or the other :)

vincepri: We can remove it, although we might want to mark this as a breaking change and add some release notes

sbueringer: I have a slight tendency to remove "rbac" as it should be dead code and if it didn't than folks are doing pretty invalid stuff

sbueringer:
/retitle ⚠ fake client: use correct RBAC apiGroup name when deciding if to allow unconditional updates

pkg/client/fake/client.go
jaideepr97: Thanks @sbueringer @vincepri,I've updated the PR

sbueringer:
@jaideepr97

Please squash the commits, then we can merge. thx!

/lgtm

PR #2414, opened 2023-07-22
🌱 Minor improvements to godoc, code style in builder pkg

sbueringer:

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -41,14 +41,14 @@ import (
 var newController = controller.New
 var getGvk = apiutil.GVKForObject
 
-// project represents other forms that the we can use to
+// project represents other forms that we can use to
 // send/receive a given resource (metadata-only, unstructured, etc).
 type objectProjection int
 
 const (
 	// projectAsNormal doesn't change the object from the form given.
 	projectAsNormal objectProjection = iota
-	// projectAsMetadata turns this into an metadata-only watch.
+	// projectAsMetadata turns this into a metadata-only watch.
 	projectAsMetadata
 )
 
@@ -69,7 +69,7 @@ func ControllerManagedBy(m manager.Manager) *Builder {
 	return &Builder{mgr: m}
 }
 
-// ForInput represents the information set by For method.
+// ForInput represents the information set by the For method.
 type ForInput struct {
 	object           client.Object
 	predicates       []predicate.Predicate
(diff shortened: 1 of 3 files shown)


pkg/builder/controller.go
```diff
@@ -344,12 +340,13 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	globalOpts := blder.mgr.GetControllerOptions()
 
 	ctrlOptions := blder.ctrlOptions
-	if ctrlOptions.Reconciler == nil {
-		ctrlOptions.Reconciler = r
+	if ctrlOptions.Reconciler != nil {
+		return errors.New("reconciler was already set via WithOptions()")
```
sbueringer, line 344: I think it makes sense to return an error when two reconcilers are set via the buidler and not just silently overwrite

pkg/builder/controller.go
```diff
@@ -344,12 +340,15 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	globalOpts := blder.mgr.GetControllerOptions()
 
 	ctrlOptions := blder.ctrlOptions
+	if ctrlOptions.Reconciler != nil && r != nil {
```
sbueringer, line 343: I think it makes sense to return an error when two reconcilers are set via the buidler and not just silently pick one

(I can move this to a separate PR if preferred)

pkg/builder/controller.go
```diff
@@ -311,19 +310,16 @@ func (blder *Builder) doWatch() error {
 		return errors.New("there are no watches configured, controller will never get triggered. Use For(), Owns() or Watches() to set them up")
 	}
 	for _, w := range blder.watchesInput {
-		allPredicates := append([]predicate.Predicate(nil), blder.globalPredicates...)
```
inteon, line 314: I liked this better, it explicitly makes sure we are not editing the `globalPredicates` value.

sbueringer: My bad. You're right

sbueringer: If I got it right we should also do this in l.277

sbueringer: Thx for catching this!

pkg/builder/controller.go
```diff
@@ -274,7 +274,8 @@ func (blder *Builder) doWatch() error {
 		}
 		src := source.Kind(blder.mgr.GetCache(), obj)
 		hdler := &handler.EnqueueRequestForObject{}
-		allPredicates := append(blder.globalPredicates, blder.forInput.predicates...)
+		allPredicates := append([]predicate.Predicate(nil), blder.globalPredicates...)
```
sbueringer, line 277: Just aligning to the way we do it below. I played a bit around with it, I don't think it makes a difference

alvaroaleman (approved):

sbueringer:
Should be an independent flake

/retest

Issue #1021, opened 2020-06-29
ControllerManagedBy builder unconditionally overrides options.Reconciler value


Danil-Grigorev:
/kind bug

https://github.com/kubernetes-sigs/controller-runtime/blob/7e2196f8357ca2ddbf0ab94318e1dedbeff1d303/pkg/builder/controller.go#L248

is always overriding `options.Reconciler` value, explicitly set in https://github.com/kubernetes-sigs/controller-runtime/blob/7e2196f8357ca2ddbf0ab94318e1dedbeff1d303/pkg/builder/controller.go#L142-L146

This complicates exclusive unit testing for `SetupWithManager` method, and forces to use older `controller.Watch` implementation to pass a mocked `Reconcile` method, or surfacing unfinished `builder.Builder` structure to achieve the same goal.

PR #2415, opened 2023-07-22
⚠ builder: return error when multiple reconcilers are set

sbueringer:
Just saw this and was confused why we would want this behavior. I also looked at https://github.com/kubernetes-sigs/controller-runtime/issues/1021 but I don't understand the reasoning.

If we want a silent overwrite behavior I would have expected Build/Complete to overwrite the reconciler from the previous set ctrlOptions not the other way around

I think this implementation still covers the original ask, folks can pass in a `nil` reconciler to Build/Complete

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -344,6 +344,9 @@ func (blder *Builder) doController(r reconcile.Reconciler) error {
 	globalOpts := blder.mgr.GetControllerOptions()
 
 	ctrlOptions := blder.ctrlOptions
+	if ctrlOptions.Reconciler != nil && r != nil {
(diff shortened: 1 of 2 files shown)


FillZpp:
Sounds good, but not sure if someones have already relied on the behavior that c-r will use the reconciler in options first with no error?

sbueringer:
> Sounds good, but not sure if someones have already relied on the behavior that c-r will use the reconciler in options first with no error?

I would assume that if they do they would get an error now and then can decide which reconciler they want to use. If they want to keep using the one from options it should be trivial to pass in "nil" into Build/Complete

alvaroaleman (approved):

PR #2417, opened 2023-07-22
✨ Avoid contacting kube apiserver is patch is empty

tamalsaha:
If the generated patch is empty, we can avoid calling kube apiserver.


timebertt (commented):
This is a breaking change in behavior as people currently rely on the client to send a patch every time that they call `Patch()`.
Though, I don't think we can do this kind of optimization under the hood. Controllers can already check if they mutated an object and skip calling `Patch()` if not. With this change however, controllers won't be able to send empty patches.

We have multiple places where we explicitly send empty patches. E.g., for triggering webhooks that might just have been installed, or for triggering API server's encryption at rest after rotating the encryption key.

vincepri (commented):
/hold

These generics clients are meant to suffice a number of use cases, like @timebertt pointed out. If you're looking for higher level optimizations I'd suggest to take a look / use the Cluster API patch client https://github.com/kubernetes-sigs/cluster-api/blob/main/util/patch/patch.go which includes optimizations to only send patches when changes are detected.

There were some discussions about upstreaming the Cluster API patch helper in controller-runtime eventually, although I personally haven't had time to look at it yet.

tamalsaha:
Thanks for enlightening me. I was not familiar with the empty patch call type cases.

I think the cluster-api patch helper goes further than what we do for Patch calls. We make separate api calls for Patch and PatchStatus and don't want to change that for potentially unintended consequences.

> Controllers can already check if they mutated an object and skip calling Patch() if not.
How do Controllers check if they have mutated an object or not in Go? We use the empty patch to check for this.


We have a wrapper method to avoid patch calls for empty patches. But this is suboptimal because we have to compute the patch (`patch.Data(obj)`) twice where the patch is not empty (which is the common scenario). Do you think it will be reasonable to add a `PatchOptions {SkipCallForEmptyPatch bool}` and use that do this if user wants?

vincepri:
>  We make separate api calls for Patch and PatchStatus and don't want to change that for potentially unintended consequences.

The Cluster API patch helper does the same

inteon:
For SSA patches, the meaning of `{}` is different, it means that you un-own all the owned fields.
(however, I don't think `{}` would work because it does not contain the metadata of the resource)

oreissig:
> We have a wrapper method to avoid patch calls for empty patches. But this is suboptimal because we have to compute the patch (`patch.Data(obj)`) twice where the patch is not empty (which is the common scenario).

You don't have to. You can calculate the patch once, check if it is `{}` and if not you can submit the `[]byte` slice already at hand using `client.RawPatch()` instead of `client.MergeFrom()`

Issue #2254, opened 2023-03-28
Cache: Support multiple namespaces for just one type


alvaroaleman:
Today, the cache supports multiple namespaces but they then get set for all namespaced objects. It is not possible to have a per-object multi-namespace config.

PR #2421, opened 2023-07-24
⚠️ Allow configuring more granular cache filtering

alvaroaleman:
This change implements most of the [cache options design][0], in particular it is now possible to:
* Configure Namespaces per type, fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2254
* Configure filtering per namespace

What is still missing is to allow configuring namespace-level default settings that will be used for all namespaces not explicitly configured. There is nothing in the way of doing that as a follow-up.

The implementation slightly derivates from the design document in that the filter settings in `ByGVK` do not use the `Config` struct but instead are directly embedded. This allows us to not break compatibility and is more ergonomic to use.

The implementation is based on a new and internal per-type delegating "meta cache".

[0]: https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/cache_options.md


pkg/cache/internal/informers.go
```diff
@@ -35,23 +35,17 @@ import (
...
-// in an InformerMap.
-type InformersOptsByGVK struct {
+	HTTPClient            *http.Client
+	Scheme                *runtime.Scheme
+	Mapper                meta.RESTMapper
+	ResyncPeriod          time.Duration
+	Namespace             string
 	Selector              Selector
 	Transform             cache.TransformFunc
 	UnsafeDisableDeepCopy *bool
```
sbueringer, line 51: nit: If we use false as a default (l.73) we can probably simply make this a bool instead of a bool pointer

(as unset doesn't have a semantic on this level here)

pkg/cache/cache.go
```diff
@@ -144,9 +144,13 @@ type Options struct {
 	// instead of `reconcile.Result{}`.
 	SyncPeriod *time.Duration
 
-	// Namespaces restricts the cache's ListWatch to the desired namespaces
-	// Per default ListWatch watches all namespaces
-	Namespaces []string
+	// DefaultNamespaces maps namespace names to cache settings. If set, only
```
sbueringer, line 147:
```suggestion
	// DefaultNamespaces maps namespace names to cache configs. If set, only
```
nit

(same for other occurences of "cache settings" below)

pkg/cache/cache.go
```diff
@@ -160,20 +164,40 @@ type Options struct {
...
+	// namespaces in this map will be cached.
+	//
+	// Settings in the map value that are unset because either the value as a
+	// whole is nil or because the specific setting is nil will be defaulted.
+	// Use an empty value for the specific setting to prevent that.
+	//
+	// It is possible to have specific Config for just some namespaces
+	// but cache all namespaces by using the AllNamespaces const as the map key.
+	// This wil then include all namespaces that do not have a more specific
+	// setting.
```
sbueringer, line 193: This should also work via DefaultNamespaces right? (i.e. DefaultNamespaces should also handle AllNamespaces)

pkg/cache/cache.go
```diff
@@ -160,20 +164,40 @@ type Options struct {
...
+	//
+	// A nil map allows to default this to the cache's DefaultNamespaces setting.
+	// An empty map prevents this.
```
sbueringer, line 196: What would the behavior be in the case of an empty map? A namespaced resource that we cache for no namespaces? :)

pkg/cache/cache.go
```diff
@@ -160,20 +164,40 @@ type Options struct {
...
+	//
+	// This must be unset for cluster-scoped objects.
+	Namespaces map[string]Config
+
 	// Label represents a label selector for the object.
 	Label labels.Selector
```
sbueringer, line 202: I think we should rename this to LabelSelector and Field to FieldSelector.

Seems inconsistent to only call these fields different in this struct

pkg/cache/cache.go
```diff
@@ -194,48 +218,111 @@ type ByObject struct {
...
+		defaultCache: defaultCache,
 	}
-	// Set the default selector and transform.
-	byGVK[schema.GroupVersionKind{}] = internal.InformersOptsByGVK{
-		Selector: internal.Selector{
-			Label: opts.DefaultLabelSelector,
-			Field: opts.DefaultFieldSelector,
-		},
+
+	for tp, config := range opts.ByObject {
```
sbueringer, line 270:
```suggestion
	for obj, config := range opts.ByObject {
```
nit

pkg/cache/cache.go
```diff
@@ -194,48 +218,111 @@ type ByObject struct {
...
-			Scheme:       opts.Scheme,
-			Mapper:       opts.Mapper,
-			ResyncPeriod: *opts.SyncPeriod,
-			Namespace:    opts.Namespaces[0],
-			ByGVK:        byGVK,
-		}),
-	}, nil
+type newCacheFunc func(config Config, namespace string) Cache
+
+func newCache(cfg *rest.Config, opts Options) newCacheFunc {
```
sbueringer, line 307:
```suggestion
func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
```
nit: just a detail but makes it easier to distinguish from config

pkg/cache/cache.go
```diff
@@ -262,31 +349,70 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 		}
 	}
 
+	for namespace, cfg := range opts.DefaultNamespaces {
+		cfg = defaultConfig(cfg, optionDefaultsToConfig(&opts))
+		opts.DefaultNamespaces[namespace] = cfg
+	}
+
+	for tp, cfg := range opts.ByObject {
```
sbueringer, line 357:
```suggestion
	for obj, cfg := range opts.ByObject {
```
nit

sbueringer (commented):
First quick round of review

pkg/cache/internal/informers.go
alvaroaleman: Done

pkg/cache/cache.go
alvaroaleman: Done

pkg/cache/cache.go
alvaroaleman: That section was actually incorrectly copied from the doc. The `AllNamespaces` part is not implemented. I've removed this section.

pkg/cache/cache.go
alvaroaleman: Empty map means all namespaces. I've updated the godoc to clarify that.

pkg/cache/cache.go
alvaroaleman: That is a breaking change and we already had a lot of breaking changes so I think it would be nice to avoid this

pkg/cache/cache.go
alvaroaleman: Done

pkg/cache/cache.go
alvaroaleman: Done

pkg/cache/cache.go
alvaroaleman: Done

pkg/cache/cache.go
sbueringer: Sounds okay. Maybe small follow-up after 0.16?

alvaroaleman: Yeah, thats an option

sbueringer: Let's resolve for now

pkg/cache/delegate_by_type.go
```diff
@@ -0,0 +1,127 @@
...
+	"golang.org/x/exp/maps"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// delegatingByTypeCache delegates to a type-specific cache if present
+// and uses the defaultCache otherwise.
+type delegatingByTypeCache struct {
```
sbueringer, line 33:
```suggestion
type delegatingByGVKCache struct {
```
nit: just feels slightly more precise, but no strong opinion

pkg/cache/multi_namespace_cache.go
```diff
@@ -146,11 +142,14 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
 
 func (c *multiNamespaceCache) Start(ctx context.Context) error {
 	// start global cache
-	go func() {
-		if err := c.clusterCache.Start(ctx); err != nil {
-			log.Error(err, "multi-namespace cache failed to start cluster scoped cache")
-		}
-	}()
+	if c.clusterCache != nil {
```
sbueringer, line 145: Do we have to check this in more places to avoid panics if folks misconfigure the cache?

If I see correctly we should never hit those cases with the way we delegate / fallback, right?

pkg/cache/defaulting_test.go
```diff
@@ -0,0 +1,290 @@
...
+		in   Options
+
+		verification func(Options) string
+	}{
+		{
+			name: "ByObject.Namespaces gets defaulted from ByObject",
+			in: Options{
+				ByObject: map[client.Object]ByObject{pod: {
+					Namespaces: map[string]Config{
+						"default": {},
```
sbueringer, line 52: Would it make sense to have cases where Config exists but its fields are nil?

sbueringer (commented):
Very nice, just nits from my side

pkg/cache/cache.go
```diff
@@ -194,48 +214,111 @@ type ByObject struct {
 	UnsafeDisableDeepCopy *bool
 }
 
+// Config describes all potential options for a given watch.
```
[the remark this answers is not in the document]
alvaroaleman: well, informer maps 1:1 to a watch, so either that or watch would work but I think odds are higher that ppl know what a watch is than that they know what an informer is.

Cache is IMHO not good, because it encompasses the whole thing with potentially many watches/informers.

pkg/cache/defaulting_test.go
alvaroaleman: This is a testcase where config exists but its fields are nil :)

pkg/cache/multi_namespace_cache.go
alvaroaleman: Yeah, there is no way to access the multiNamespace cache directly as it is not exported and we do have testcases in cache_test.go where this is used only for a namespaces type and thus doesn't have a clusterCache. That is how I found out that these are needed :)

pkg/cache/cache.go
```diff
@@ -160,20 +165,35 @@ type Options struct {
...
 	ByObject map[client.Object]ByObject
 }
 
 // ByObject offers more fine-grained control over the cache's ListWatch by object.
 type ByObject struct {
+	// Namespaces maps a namespace name to cache configs. If set, only the
+	// namespaces in this map will be cached.
+	//
+	// Settings in the map value that are unset because either the value as a
+	// whole is nil or because the specific setting is nil will be defaulted.
```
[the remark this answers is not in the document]
alvaroaleman: Added some more details, WDYT?

pkg/cache/cache.go
```diff
@@ -144,9 +145,13 @@ type Options struct {
...
+	// DefaultNamespaces maps namespace names to cache configs. If set, only
+	// the namespaces in here will be watched and it will by used to default
+	// ByObject.Namespaces for all objects if that is nil.
+	//
+	// The options in the Config that are nil will be defaulted from
+	// the respective Default* settings.
+	DefaultNamespaces map[string]Config
 
 	// DefaultLabelSelector will be used as a label selector for all object types
 	// unless they have a more specific selector set in ByObject.
```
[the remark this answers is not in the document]
alvaroaleman: Added something, WDYT?

pkg/cache/cache.go
```diff
@@ -194,48 +214,111 @@ type ByObject struct {
...
+	// default this.
+	LabelSelector labels.Selector
+
+	// FieldSelector specifics a field selector. A nil value allows to
+	// default this.
+	FieldSelector fields.Selector
+
+	// Transform specifies a transform func. A nil value allows to default
+	// this.
+	Transform toolscache.TransformFunc
```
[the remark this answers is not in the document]
alvaroaleman: Added that, WDYT?

pkg/cache/cache.go
```diff
@@ -262,31 +345,70 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 		}
 	}
 
+	for namespace, cfg := range opts.DefaultNamespaces {
+		cfg = defaultConfig(cfg, optionDefaultsToConfig(&opts))
+		opts.DefaultNamespaces[namespace] = cfg
+	}
+
+	for obj, cfg := range opts.ByObject {
```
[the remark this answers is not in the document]
alvaroaleman: Done

pkg/cache/cache.go
```diff
@@ -194,48 +214,111 @@ type ByObject struct {
...
-		opts.Namespaces = []string{metav1.NamespaceAll}
+func New(cfg *rest.Config, opts Options) (Cache, error) {
+	opts, err := defaultOpts(cfg, opts)
+	if err != nil {
+		return nil, err
 	}
-	if len(opts.Namespaces) > 1 {
-		return newMultiNamespaceCache(config, opts)
+
+	newCache := newCache(cfg, opts)
```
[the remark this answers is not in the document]
alvaroaleman: Renamed it. Either works and either shadows something 🙃

pkg/cache/delegate_by_type.go
```diff
@@ -0,0 +1,127 @@
+/*
```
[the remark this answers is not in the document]
alvaroaleman: Renamed it to `delegating_by_gvk_cache.go`, WDYT?

pkg/cache/delegate_by_type.go
```diff
@@ -0,0 +1,127 @@
...
+		wg.Wait()
+		return nil
+	}
+}
+
+func (dbt *delegatingByTypeCache) WaitForCacheSync(ctx context.Context) bool {
+	synced := true
+	for _, cache := range append(maps.Values(dbt.caches), dbt.defaultCache) {
+		if !cache.WaitForCacheSync(ctx) {
+			synced = false
```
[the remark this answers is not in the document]
alvaroaleman: The reason it is the way that it is is that the interface is documented as "Blocks until caches are synced". If we return early, not all caches might be synced. I don't know if anyone would continue their binary if this returns `false` but odds are, someone somewhere will

pkg/cache/delegate_by_type.go
alvaroaleman: Renamed it

pkg/cache/cache.go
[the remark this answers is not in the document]
sbueringer: Sounds good!

pkg/cache/cache.go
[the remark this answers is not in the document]
sbueringer: Sounds good

pkg/cache/delegate_by_type.go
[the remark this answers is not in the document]
sbueringer: Yeah let's better not change. Was just thinking if we're hitten the case where all caches are definitely not synced we might as well return earlier. But no real reason to make a change

pkg/cache/multi_namespace_cache.go
sbueringer: Perfect!

pkg/cache/defaulting_test.go
sbueringer: ... ups :) 

What I was thinking was basically did we test "both cases" based on your previous godoc comment:
> 	// Settings in the map value that are unset because either the value as a
>	// whole is nil or because the specific setting is nil will be defaulted.

But if I see correctly (now) "the value as a whole is nil" is not possible as our map values are all not pointers, right?

alvaroaleman: Yeah, exactly. I did notice that discrepancy between godocs and code to and updated the godoc

sbueringer:
Thx

/lgtm

Issue #1186, opened 2020-09-28
Log events that trigger reconciles


asmacdo:
As an controller author/user, it would be helpful to know which event has triggered a reconcile, especially in situations where a reconcile can be triggered by changes to multiple resources. 

What I have in mind is to log the event in the EnqueueRequestForObject event handlers: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/handler/enqueue.go#L37-L91

asmacdo:
I brought this up at the sig meeting, and I got the go-ahead to implement this. It was requested that this info be logged at a very high level.

camilamacedo86:
Based on the comment `I brought this up at the sig meeting, and I got the go-ahead to implement this. It was requested that this info be logged at a very high level.` I am labelling it as help-wanted and triage/accepted.

coderanger:
On my list to shift upstream is this code https://github.com/coderanger/controller-utils/blob/e26c5734ecc941a03e9bf8798886a419e5202126/predicates/update_debug.go

Would probably answer this need?

shashankram:
@DirectXMan12 @camilamacedo86 is there a way to enable event trigger logs in controller-runtime?

sbueringer:
There is no correlation between an event and a single reconcile request. Accordingly there are also no logs which surface which event triggered which reconcile.

You can add predicates to log all incoming events if you want though

PR #2427, opened 2023-07-27
:sparkles: add logging to Enqueue_owner event handler

varshaprasad96:
This PR adds logging to Enqueue owner event handler, that is helpful in logging the GVK of the owner that
caused reconciliation.

Fixes: #1186 

In Operator SDK we have a `LoggingEnqueueRequestForOwnerHandler` ([ref](https://github.com/operator-framework/operator-sdk/blob/bd0ccd22e9daa9935770445b0f11c5190042481a/internal/ansible/handler/logging_enqueue_owner.go#L29C1-L31)) that used to add logging to the existing `EnqueueRequestForOwnerHandler`. In C-R 0.15.0 this was made internal, making it difficult to create a wrapper around it. However, adding logging to the existing handler available in c-r would be useful to other consumers also in identifying the cause of reconcile.

### DIFF
--- a/pkg/handler/enqueue_owner.go
+++ b/pkg/handler/enqueue_owner.go
@@ -84,6 +84,7 @@ type enqueueRequestForOwner struct {
 
 // Create implements EventHandler.
 func (e *enqueueRequestForOwner) Create(ctx context.Context, evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	e.logEvent("Create", evt.Object, nil)
 	reqs := map[reconcile.Request]empty{}
 	e.getOwnerReconcileRequest(evt.Object, reqs)
 	for req := range reqs {
@@ -93,6 +94,7 @@ func (e *enqueueRequestForOwner) Create(ctx context.Context, evt event.CreateEve
 
 // Update implements EventHandler.
 func (e *enqueueRequestForOwner) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	e.logEvent("Update", evt.ObjectOld, evt.ObjectNew)
 	reqs := map[reconcile.Request]empty{}
 	e.getOwnerReconcileRequest(evt.ObjectOld, reqs)
 	e.getOwnerReconcileRequest(evt.ObjectNew, reqs)
@@ -103,6 +105,7 @@ func (e *enqueueRequestForOwner) Update(ctx context.Context, evt event.UpdateEve
 
 // Delete implements EventHandler.
 func (e *enqueueRequestForOwner) Delete(ctx context.Context, evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+	e.logEvent("Delete", evt.Object, nil)
 	reqs := map[reconcile.Request]empty{}
 	e.getOwnerReconcileRequest(evt.Object, reqs)
 	for req := range reqs {
@@ -112,6 +115,7 @@ func (e *enqueueRequestForOwner) Delete(ctx context.Context, evt event.DeleteEve
 
 // Generic implements EventHandler.
 func (e *enqueueRequestForOwner) Generic(ctx context.Context, evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	e.logEvent("Generic", evt.Object, nil)
 	reqs := map[reconcile.Request]empty{}
 	e.getOwnerReconcileRequest(evt.Object, reqs)
 	for req := range reqs {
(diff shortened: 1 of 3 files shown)


pkg/handler/eventhandler_test.go
```diff
@@ -467,7 +491,7 @@ var _ = Describe("Eventhandler", func() {
 				{
 					Name:       "foo-parent",
 					Kind:       "HorizontalPodAutoscaler",
-					APIVersion: "autoscaling/v2beta1",
+					APIVersion: "autoscaling/v2",
```
varshaprasad96, line 494: Bumping this since v2beta1 was deprecated since k8s 1.22

pkg/handler/enqueue_owner.go
```diff
@@ -197,3 +201,49 @@ func (e *enqueueRequestForOwner) getOwnersReferences(object metav1.Object) []met
...
+		if objectNs := object.GetNamespace(); objectNs != "" {
+			kvs = append(kvs, "Namespace", objectNs)
+		}
+		kvs = append(kvs,
+			"Owner APIVersion", ownerReference.APIVersion,
+			"Owner Kind", ownerReference.Kind,
+			"Owner Name", ownerReference.Name,
+		)
+
+		log.Info("OwnerReference handler event", kvs...)
```
everettraven, line 231: What "level" does this get logged at? Is there a way for me to specify that I only want this logged at a certain verbosity level?

erikgb:
```suggestion
		log.V(1).Info("OwnerReference handler event", kvs...)
```

I don't think this should be logged at the normal Info level, as it will create too chatty logs. See https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md

pkg/handler/enqueue_owner.go
```diff
@@ -197,3 +201,49 @@ func (e *enqueueRequestForOwner) getOwnersReferences(object metav1.Object) []met
 	// No Controller OwnerReference found
 	return nil
 }
+
+// logEvent logs an event with the version, kind and name of the object and its owner.
+func (e *enqueueRequestForOwner) logEvent(eventType string, object, newObject client.Object) {
```
erikgb, line 206: I am personally not a big fan of func signatures like this one - with positional parameters. Could using the builder pattern be an option here?

pkg/handler/eventhandler_suite_test.go
```diff
@@ -34,9 +35,10 @@ func TestEventhandler(t *testing.T) {
 
 var testenv *envtest.Environment
 var cfg *rest.Config
+var logBuffer bytes.Buffer
```
erikgb, line 38: Could this global variable cause issues since it's mutated (`Reset`) in the test specs? How about unit testing the logging in `enqueueRequestForOwner` instead of in the e2e-tests?

varshaprasad96: > How about unit testing the logging in enqueueRequestForOwner instead of in the e2e-tests?

The logger is currently getting tested while unit testing for specific events in `enqueueRequestForOwner`. Could you point to the alternate option?

> Could this global variable cause issues since it's mutated (Reset) in the test specs?

Would prefer to keep this at the beginning of test suite, because it needs to be set as the [default logger](https://github.com/varshaprasad96/controller-runtime/blob/0edc5c2afb51816e2834bec25dc4838ef5fe67c5/pkg/handler/eventhandler_suite_test.go#L41) before starting the tests. `BeforeSuite` anyways gets executed only once, though we are resetting the content quite a few times in the test spec for specific tests.

pkg/handler/enqueue_owner.go
```diff
@@ -112,6 +115,7 @@ func (e *enqueueRequestForOwner) Delete(ctx context.Context, evt event.DeleteEve
 
 // Generic implements EventHandler.
 func (e *enqueueRequestForOwner) Generic(ctx context.Context, evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	e.logEvent("Generic", evt.Object, nil)
```
fgiloux, line 118: It adds some amount of processing to a code path, which is frequently called. I am wondering whether it would not be possible to leverage what is already returned by `getOwnerReconcileRequest` a couple of lines below to get information on the owner instead of what is made in `logEvent` and `extractTypedOwnerReference`?
Also wouldn't we want to provide information on the object itself?

pkg/handler/enqueue_owner.go
varshaprasad96: Agreed. But a builder pattern just for `logEvent` could be an overkill just for one helper. I'll rework on this bit, to fetch the owner references from both the objects and then pass it to logevent so that its a bit more cleaner.

pkg/handler/enqueue_owner.go
alvaroaleman: Yeah, this is defininitely debug level, it will cause a huge amount of logs for a controller that is even slightly busy

erikgb:
I agree with @alvaroaleman, and to get a bit more familiar with the c-r codebase, I drafted something that might allow users to wrap event handlers. Feel free to use it in a new PR if you like the idea: https://github.com/kubernetes-sigs/controller-runtime/commit/2f012137f407fe9a3d8055d889e79a3d80f2ca92

varshaprasad96:
> Your wrapper can and should wrap the handler.Handler interface, not the concrete implementation.

@alvaroaleman Agreed that we need to implement the handler interface. However, `enqueueRequestForOwner` is a commonly used handler, which is what we are looking to re-use, just with additional logging on who the owner was that caused reconcile.

> How does the implementation being internal prevent you from adding logging?
Based on the current state, if we want to customize and reuse a particular implementation of handler we would have to re-write the same piece of code that implements the interface. 

I would like to explore the options before closing the PR here:
1. Agreed on @erikgb's implementation, implementing a logging handler and providing an option to wrap it with existing one's is useful.
2. Injecting a logger into the handler implementation (similar to how it was done in previous implementations with `inject` pkg), where if provided we would log the details on the events. Or have the current implementation and allow users to enable/disable it using Options. (Con: we cannot control the log message for each event).
3. Provide an option to enable passing of multiple event handlers. Each event handler would create a listener (https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L556), which would then be enabled [here](https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L962-L987).

I'm not super familiar with the client-go side of code, but the question which I have with (1) and (3) is while wrapping or enabling multiple handlers, there is no guarantee on the order in which [listeners](https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L797) will be distributed. Which is something we would want to maintain at least while logging for a particular event?

alvaroaleman:
> Agreed that we need to implement the handler interface. However, enqueueRequestForOwner is a commonly used handler, which is what we are looking to re-use, just with additional logging on who the owner was that caused reconcile.

Yeah, you can do that as a wraper.

> Injecting a logger into the handler implementation (similar to how it was done in previous implementations with inject pkg), where if provided we would log the details on the events. Or have the current implementation and allow users to enable/disable it using Options. (Con: we cannot control the log message for each event).

The approach of injecting things is deprecated. You can get a logger from the `manager` or anywhere else.

> Provide an option to enable passing of multiple event handlers. Each event handler would create a listener [..]

I don't understand this part. The source registers the eventhandler in the cache and then passed the event onto the predicate.predicate and finally handler.Handler (the package here). Why do you want multiple handlers?

sbueringer:
I'm inclined to close this PR given no recent activity and the alternative suggested by @alvaroaleman. @varshaprasad96 WDYT?

vincepri:
/close

per comment above

PR #2432, opened 2023-08-02
optimize PatchResponseFromRaw

### LINKED ISSUE #2431: Reduce cpu usage for controller-runtime based webhook
The reason for high CPU consumption is that the mutating webhook performs a reflect.DeepEqual operation on the fields of both oldObj and newObj when generating the jsonpatch array for the diff between them, regardless of whether they are equal or not. This reflect.DeepEqual operation consumes a lot of CPU time.

while most of the time, oldObj and newObj are the same. So comparing the bytes array of oldObj and newObj before generating jsonpatch array can save a lot cpu cycles. 

a sample profile:

image

d3c3mber:
fixes #2431

### DIFF
--- a/pkg/webhook/admission/defaulter_test.go
+++ b/pkg/webhook/admission/defaulter_test.go
@@ -2,12 +2,18 @@ package admission
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"net/http"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	admissionv1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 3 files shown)


pkg/webhook/admission/defaulter_test.go
```diff
@@ -66,3 +72,124 @@ func (d *TestDefaulter) Default() {
...
+
+AFTER (with isByteArrayEqual)
+goos: darwin
+goarch: arm64
+pkg: sigs.k8s.io/controller-runtime/pkg/webhook/admission
+BenchmarkHandle
+BenchmarkHandle/benchmark_handle_function
+BenchmarkHandle/benchmark_handle_function-10         	   44620	     26178 ns/op
+*/
+func BenchmarkHandle(b *testing.B) {
```
alvaroaleman, line 94: These two benchmarks are not representative, as you are only benchmarking the "objects are equal case" and not the "objects are not equal", in which case your change worsens performance

d3c3mber: Thanks for your prompt reply @alvaroaleman 
I've add more benchmark tests regarding equal case and not equal case.
```
BenchmarkPatchResponseFromRawAndIsByteArrayEqual/benchmark_PatchResponseFromRaw_function-10         	   52774	     22749 ns/op
BenchmarkPatchResponseFromRawAndIsByteArrayEqual/benchmark_isByteArrayEqual_function-10             	 5252599	       230.7 ns/op
```
To add more details to this:
1) By benchmarking, comparing two byte arrays takes far less time than the recursive reflect.DeepEqual, which is about 1/100 according to my above benchmark
2) Most of the Apiserver's patch/update requests will still get same object after going through all mutating webhooks. as mutating webhooks most of the time take effect at the beginning of a Pod's lifecycle (or other resources lifecycle)
3) In our production cluster, this change "reducing recursive reflect.DeepEqual" can save more than 40% cpu time for webhook.

pkg/webhook/admission/response.go
```diff
@@ -86,6 +86,17 @@ func ValidationResponse(allowed bool, message string) Response {
 // The original object should be passed in as raw bytes to avoid the roundtripping problem
 // described in https://github.com/kubernetes-sigs/kubebuilder/issues/510.
 func PatchResponseFromRaw(original, current []byte) Response {
+	if isByteArrayEqual(original, current) {
```
inteon, line 89:
```suggestion
	if bytes.Equal(original, current) {
```

PR #2434, opened 2023-08-03
✨ Use object's TypeMeta as GVK fallback

scothis:
GVKForObject looks up the GVK for an Object via a Scheme, unless the object is one of a few generic types. When handling multiple similar resources, it's connivent to coerce those resource to a common static type (often a subset of the actual object). These types cannot be registered in the scheme as they do not map to a single GVK.

I would like to be able to resolve a GVK from a non-scheme registered object type so that these types can be used in other useful methods that internally call GVKForObject, like SetControllerReference. To do this, if the lookup from the scheme fails, it will use the TypeMeta fields on the object and return those values as the GVK. This behavior requires the APIVersion and Kind fields are set, otherwise it will fall back to returning the scheme error.

### DIFF
--- a/pkg/client/apiutil/apimachinery.go
+++ b/pkg/client/apiutil/apimachinery.go
@@ -135,6 +135,13 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 	// Use the given scheme to retrieve all the GVKs for the object.
 	gvks, isUnversioned, err := scheme.ObjectKinds(obj)
 	if err != nil {
+		if runtime.IsNotRegisteredError(err) && obj.GetObjectKind() != nil {
+			// Use object's current GVK as a last ditch attempt for a type not registered with the scheme.
+			currentGVK := obj.GetObjectKind().GroupVersionKind()
+			if !currentGVK.Empty() {
+				return currentGVK, nil
+			}
+		}
 		return schema.GroupVersionKind{}, err
 	}
 	if isUnversioned {
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -131,6 +131,28 @@ var _ = Describe("Controllerutil", func() {
 			Expect(controllerutil.SetControllerReference(dep, rs, runtime.NewScheme())).To(HaveOccurred())
 		})
 
+		It("should set the OwnerReference if it can find the group version kind, even if not registered in the scheme", func() {
+			rs := &appsv1.ReplicaSet{}
+			dep := &extensionsv1beta1.Deployment{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "extensions/v1beta1",
+					Kind:       "Deployment",
+				},
+				ObjectMeta: metav1.ObjectMeta{Name: "foo", UID: "foo-uid"},
+			}
+
+			Expect(controllerutil.SetControllerReference(dep, rs, runtime.NewScheme())).NotTo(HaveOccurred())
+			t := true
+			Expect(rs.OwnerReferences).To(ConsistOf(metav1.OwnerReference{
+				Name:               "foo",
+				Kind:               "Deployment",
+				APIVersion:         "extensions/v1beta1",
+				UID:                "foo-uid",
+				Controller:         &t,
+				BlockOwnerDeletion: &t,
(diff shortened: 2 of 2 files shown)


alvaroaleman (approved):
/hold

Seems ok to me. @vincepri any concerns?

sbueringer:
Sounds reasonable to me. Should be also safe because the behavior just changes for cases where before we returned an error

/lgtm

pkg/client/apiutil/apimachinery.go
```diff
@@ -135,6 +135,13 @@ func GVKForObject(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersi
 	// Use the given scheme to retrieve all the GVKs for the object.
 	gvks, isUnversioned, err := scheme.ObjectKinds(obj)
 	if err != nil {
+		if runtime.IsNotRegisteredError(err) && obj.GetObjectKind() != nil {
+			// Use object's current GVK as a last ditch attempt for a type not registered with the scheme.
+			currentGVK := obj.GetObjectKind().GroupVersionKind()
+			if !currentGVK.Empty() {
+				return currentGVK, nil
+			}
+		}
```
vincepri, line 144: 🤔 Reading through the use case, it makes sense why one would want to reuse a type across many types and ultimately make use of the facilities already in place.

This code IIRC is used in a number of places, including the client itself, by allowing non-registered types to pass through, aren't now allowing types to go through caches/clients even when they're not supported?

scothis: > This code IIRC is used in a number of places, including the client itself, by allowing non-registered types to pass through, aren't now allowing types to go through caches/clients even when they're not supported?

My desire is to make those types behave largely as if they are supported. To do so, I currently wrap the manager returned client adding a check if the obj is registered with the scheme, and if not converts it to an unstructured before calling the upstream client, and then reflects any updated values on the unstructured object back. (This behavior is largely working in the latest release of [reconciler-runtime](https://github.com/vmware-labs/reconciler-runtime))

Updating `GVKForObject` will help avoid conversions to/from unstructured in other parts of the codebase that are harder to wrap and user code, like `SetControllerReference`.

> Example, when creating a controller `GVKForObject` is used to determine the GVK, and we pass in the scheme given that we're expecting to use the scheme to create new instances of that object, if that object isn't registered, it won't fail until much later?

I've never seen someone use a non-empty object as the type, but it's certainly possible. If you like, I can add a defensive check in the builder that will fail if the object is not unstructured or registered in the scheme.

alvaroaleman: I don't think as of today we ever assume that a type that is not unstructured/PartialObjectMeta can pass this check and not be registered in the scheme. So this definitely has potential to cause issues both for us and existing users of the downstream code.

What is the reason to not simply do this downstream through a slim wrapper around this?

scothis: > I don't think as of today we ever assume that a type that is not unstructured/PartialObjectMeta can pass this check and not be registered in the scheme.

Correct, it always errs today.

>What is the reason to not simply do this downstream through a slim wrapper around this?

It's not easy to wrap bare functions, unlike an interface that is expected to be injected (e.g. `Client`). I'd have to provide alternate implementation and then tell users to call my custom api, or resort to nasty hacks.

alvaroaleman: Not sure I follow. So you can tell your users to use a different type but not to use a different way of inferring the GVK from that type?

> It's not easy to wrap bare functions, 
It is:
```
func wrapping(f foo, b bar)(gvk, err)
  upstreamResult, err := upstream(f, b)
  if err != nil {
     return doSomethingElse(f, b)
  }
  return upstreamResult, nil
}
```

scothis: @alvaroaleman yea, that's the easy part. The hard part is switching every caller to use the wrapper function.

vincepri: A random thought about this: maybe we could do something like Go has done for errors, with Wrap/Unwrap, and have a fallback to check a well-defined interface to Unwrap the true type of a struct. This would give us the ability to keep the current behavior, while introducing this new functionality of common types.

scothis:
I'm no longer employed by VMware, so closing this PR. Will incorporate the feedback into a new PR.

PR #2435, opened 2023-08-03
:bug: Fix Defaulting of the User Agent

alvaroaleman:
This broke when we added the HTTP client, because the user-agent gets set by a roundtripper that is constructed within `rest.HTTPClientFor`. As a result, we have to default it before we do that. Currently, it ends up being defaulted to `Go-http-client` which is not very useful.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -334,6 +334,11 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 }
 
 func defaultOpts(config *rest.Config, opts Options) (Options, error) {
+	config = rest.CopyConfig(config)
+	if config.UserAgent == "" {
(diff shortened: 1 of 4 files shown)


FillZpp:
LGTM. Nit: the `rest.DefaultKubernetesUserAgent()` only contains the version of `client-go`, should we have a default user-agent that also show the version of controller-runtime?

alvaroaleman:
> LGTM. Nit: the rest.DefaultKubernetesUserAgent() only contains the version of client-go, should we have a default user-agent that also shows the version of controller-runtime?

No real strong opinion either way, but I don't think that version is currently available to us?

vincepri (approved):
/lgtm

Merging this right now, given that it fixes the underlying bug. @FillZpp let's follow-up on having a more useful user agent with controller-runtime in mind

Issue #2430, opened 2023-08-01
TerminalError(nil).Error() panics


sheidkamp:
In v0.15.0
Code to trigger:
```
package main

import (
	"fmt"
	"sigs.k8s.io/controller-runtime/pkg/reconcile"
)

func main() {
	te_nil := reconcile.TerminalError(nil)
	// te_nil = reconcile.TerminalError(fmt.Errorf("nil"))
	fmt.Print(te_nil.Error())
}
```

`TerminalError(nil)` is being used to create errors here: https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/internal/controller/controller.go#L319

This specifically causes problems when used with the Eris error logger:
```
package main

import (
	"errors"
	"fmt"

	"github.com/rotisserie/eris"
	"sigs.k8s.io/controller-runtime/pkg/reconcile"
)

func main() {
	err := eris.New("some error")

	if errors.Is(err, reconcile.TerminalError(nil)) {
		fmt.Print("Error is nil")
	} else {
		fmt.Print("Error is not nil")
	}
}
```
results in 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x48cafb]

goroutine 1 [running]:
sigs.k8s.io/controller-runtime/pkg/reconcile.(*terminalError).Error(0x49c0e0?)
	/tmp/gopath2655630741/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/reconcile/reconcile.go:120 +0x1b
github.com/rotisserie/eris.(*rootError).Is(0xc0000161e0, {0x4c57e8?, 0xc000014280?})
	/tmp/gopath2655630741/pkg/mod/github.com/rotisserie/eris@v0.5.4/eris.go:216 +0x73
errors.Is({0x4c5748, 0xc0000161e0}, {0x4c57e8, 0xc000014280})
	/usr/local/go-faketime/src/errors/wrap.go:53 +0x156
main.main()
	/tmp/sandbox73597216/prog.go:17 +0x5c
```

Proposed change:
Code to check for nil in https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/reconcile/reconcile.go#L119

along the lines of:
```
func (te *terminalError) Error() string {
[...]

PR #2438, opened 2023-08-04
:bug:  Fix TerminalError(nil).Error() panic

sheidkamp:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2430 by checking it `err` is nil before calling `Error()` on it.

### DIFF
--- a/pkg/reconcile/reconcile.go
+++ b/pkg/reconcile/reconcile.go
@@ -112,11 +112,15 @@ type terminalError struct {
 	err error
 }
 
+// This function will return nil if te.err is nil.
 func (te *terminalError) Unwrap() error {
 	return te.err
 }
 
 func (te *terminalError) Error() string {
+	if te.err == nil {
+		return "nil terminal error"
+	}
 	return "terminal error: " + te.err.Error()
 }
 
--- a/pkg/reconcile/reconcile_test.go
+++ b/pkg/reconcile/reconcile_test.go
@@ -96,5 +96,10 @@ var _ = Describe("reconcile", func() {
 
 			Expect(apierrors.IsGone(terminalError)).To(BeTrue())
 		})
+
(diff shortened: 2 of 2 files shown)


pkg/reconcile/reconcile.go
```diff
@@ -112,11 +112,15 @@ type terminalError struct {
...
 
+// This function will return nil if te.err is nil.
 func (te *terminalError) Unwrap() error {
 	return te.err
 }
 
 func (te *terminalError) Error() string {
+	if te.err == nil {
+		return "nil terminal error"
+	}
```
vincepri, line 123: Is this a common pattern? We're at this point creating an error with a nil cause/inner error which might throw off the reconciliation loop.

sheidkamp: I only see it used here: https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/internal/controller/controller.go#L319
which doesn't totally make sense because we just checked for `err==nil`

An alternate solution would be to modify the controller to create that error with a non-nil value (empty string?) and maybe add protection in the [TerminalError()](https://github.com/kubernetes-sigs/controller-runtime/blob/dbcf380a24f5048925a3981905ded66a72c2a35f/pkg/reconcile/reconcile.go#L107) function to not allow nil values.

vincepri: cc @alvaroaleman

alvaroaleman: The nil terminal error there is only used for the `errors.Is` comparison, in that context, the `Error()` method will never be called.

It isn't really expected to construct a terminal error with a nil error when returning it. When did you run into this panicing?

alvaroaleman (approved):

PR #2439, opened 2023-08-05
:warning: Return an error if the continue list option is set for the cache reader

shuheiktgw:
Hi, currently the cache reader ignores the continue option. This behavior could lead to a hard-to-find bug when a user mistakenly tries to paginate with the limit and continue option by the cache client. I faced the problem, and that was quite hard to debug since the limit option works correctly and I always received the first page only. I believe it is more user-friendly if we just return an error when a user mistakenly set the option instead of ignoring it silently. Thanks for your review!

### DIFF
--- a/pkg/cache/internal/cache_reader.go
+++ b/pkg/cache/internal/cache_reader.go
@@ -111,6 +111,10 @@ func (c *CacheReader) List(_ context.Context, out client.ObjectList, opts ...cli
 	listOpts := client.ListOptions{}
 	listOpts.ApplyOptions(opts)
 
+	if listOpts.Continue != "" {
+		return fmt.Errorf("continue list option is not supported by the cache")
+	}
+
 	switch {
 	case listOpts.FieldSelector != nil:
 		// TODO(directxman12): support more complicated field selectors by
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -688,6 +688,14 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 					Expect(informerCache.List(context.Background(), listObj, labelOpt, limitOpt)).To(Succeed())
 					Expect(listObj.Items).Should(HaveLen(1))
 				})
+
+				It("should return an error if the continue list options is set", func() {
+					listObj := &corev1.PodList{}
+					continueOpt := client.Continue("token")
+					By("verifying that an error is returned")
+					err := informerCache.List(context.Background(), listObj, continueOpt)
+					Expect(err).To(HaveOccurred())
+				})
 			})
 
 			Context("with unstructured objects", func() {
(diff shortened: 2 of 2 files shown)


gjkim42:
> This behavior could lead to a hard-to-find bug when a user mistakenly tries to paginate with the limit and continue option by the cache client. I faced the problem, and that was quite hard to debug since the limit option works correctly and I always received the first page only.

I believe that I faced the same issue [here](https://github.com/Kong/kubernetes-ingress-controller/issues/4456).

czeslavo:
I have one concern: a typical flow of using pagination is you set `Limit` to your page size and leave `Continue` empty in your first `List` call. This fix wouldn't catch that someone is trying to use `CacheReader` with pagination as `Continue` would be empty anyway.

As `CacheReader.List` accepts `ListOptions` with both `Limit` and `Continue`, claiming de facto it implements the `client.Reader` interface, wouldn't it make sense to simulate the continuation on its side? If we don't do that, it's always on the caller's side to know what kind of a client is being used and account for continuations not working for some of them, which is not a perfect situation. I'd expect the interface to be implemented consistently despite the implementation details.

alvaroaleman:
> As CacheReader.List accepts ListOptions with both Limit and Continue, claiming de facto it implements the client.Reader interface, wouldn't it make sense to simulate the continuation on its side? 

Thats on the callers side anyways, because the two are not identical, they just expose an identical interface. You for example have to know that if you get data from cache it might be stale and deal with that properly.

pkg/cache/internal/cache_reader.go
```diff
@@ -111,6 +111,10 @@ func (c *CacheReader) List(_ context.Context, out client.ObjectList, opts ...cli
 	listOpts := client.ListOptions{}
 	listOpts.ApplyOptions(opts)
 
+	if listOpts.Continue != "" {
```
alvaroaleman, line 114: I agree with this change, however it is technically breaking because using this before would not do anything, now it returns an error. Could you update the title accordingly?

czeslavo:
> Thats on the callers side anyways, because the two are not identical, they just expose an identical interface. You for example have to know that if you get data from cache it might be stale and deal with that properly.

If that's expected to not implement some options it technically accepts, could we make sure that these intentional gaps are properly documented?

pkg/cache/internal/cache_reader.go
shuheiktgw: Thank you! I've updated the title 🙂

alvaroaleman (approved):

multi-io:
As long as only continue is rejected but limit is still accepted (but not implemented), this still leads to hard to find bugs, so I don't see how this PR alone changes much.

multi-io:
As long as only continue is rejected but limit is still accepted, this still leads to hard to find bugs if the user tries to use continue (rather than just calling List() once with a fixed limit, ignoring the continue flag). As long as the cache doesn't support continue, List() calls should probably just return a random non-empy continue string so that if the client copies into into the next request, they'll fail-fast into this error rather than just getting a single truncated list with no error.

PR #2451, opened 2023-08-17
🌱 log warning if error and non-zero Result returned

sbueringer:


pkg/internal/controller/controller.go
```diff
@@ -323,6 +323,9 @@ func (c *Controller) reconcileHandler(ctx context.Context, obj interface{}) {
 		}
 		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
 		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelError).Inc()
+		if !result.IsZero() {
+			log.Info("Warning: error and non-zero Result returned, ignoring Result")
```
alvaroaleman, line 327:
```suggestion
			log.Warn("Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff")
```

Issue #2381, opened 2023-06-15
SetControllerReference should allow overwriting the previous controller


therealak12:
I want to change the controller of an object. The current implementation of `SetControllerReference` returns an error if the new owner is different than the previous one.

Adding a new method `SetControllerReferenceOverwrite` can be a solution.

vincepri:
Changing the controller owner of an object is a very rare operation, which can also impact deletion flows in case of forward collection. I'd expect users to remove the older reference first, then set a new one in that case?

therealak12:
Agree it won't be used that much, but for doing so I needed to duplicate some controller-runtime methods such as `referSameObject` and `validateOwner` into our code which wasn't so pleasant!

troy0820:
/kind support 

Would creating a function to remove the controller reference (`RemoveControllerReference`) be something as a compromise?

troy0820:
PR merged for function `RemoveOwnerReference` in controllerutil that should help this use case.

therealak12:
> PR merged for function `RemoveOwnerReference` in controllerutil that should help this use case.

Thanks!

PR #2455, opened 2023-08-24
🌱 Add function to controllerutil `RemoveControllerReference` to remove controller references from objects

troy0820:
When using the function `SetControllerReference` it will error when trying to overwrite the controller reference.  In issue #2381 the method used to circumvent the errors is to remove the controller reference and then reapply the new controller reference.  There is no exposed method to do this.  This mimics the behavior of removing a finalizer 

- [x] Add function with description 
- [x] Add test with coverage
 

Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/2381

Release note:
```release-note 
Add function RemoveControllerReference to pkg/controller/controllerutil 
to allow the removal of the owner reference of the controlled object.
```

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -91,6 +91,27 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
+// RemoveControllerReference is a helper method to make sure the given object removes an controller reference to the object provided.
+// This allows you to remove the owner to establish a new owner of the object in a subsequent call.
+func RemoveControllerReference(owner, controlled metav1.Object) error {
+	owners := controlled.GetOwnerReferences()
+	length := len(owners)
+	if length < 1 {
+		return fmt.Errorf("%T does not have any owner references", controlled)
+	}
+	index := 0
+	for i := 0; i < length; i++ {
+		if owners[i].Name == owner.GetName() {
+			owners = append(owners[:index], owners[index+1:]...)
+		}
+		index++
+	}
+	if length == len(owners) {
+		return fmt.Errorf("%T does not have an owner reference for %T", controlled, owner)
+	}
+	return nil
+}
+
 // SetOwnerReference is a helper method to make sure the given object contains an object reference to the object provided.
 // This allows you to declare that owner has a dependency on the object without specifying it as a controller.
 // If a reference to the same object already exists, it'll be overwritten with the newly provided version.
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -91,24 +91,25 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
-// RemoveControllerReference is a helper method to make sure the given object removes an controller reference to the object provided.
+// RemoveControllerReference is a helper method to make sure the given object removes a controller reference to the object provided.
 // This allows you to remove the owner to establish a new owner of the object in a subsequent call.
-func RemoveControllerReference(owner, controlled metav1.Object) error {
-	owners := controlled.GetOwnerReferences()
+func RemoveControllerReference(owner, object metav1.Object) error {
+	owners := object.GetOwnerReferences()
 	length := len(owners)
+	result := []metav1.OwnerReference{}
 	if length < 1 {
-		return fmt.Errorf("%T does not have any owner references", controlled)
+		return fmt.Errorf("%T does not have any owner references", object)
 	}
-	index := 0
-	for i := 0; i < length; i++ {
-		if owners[i].Name == owner.GetName() {
-			owners = append(owners[:index], owners[index+1:]...)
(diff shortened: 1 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -91,6 +91,29 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
+// RemoveControllerReference is a helper method to make sure the given object removes an controller reference to the object provided.
+// This allows you to remove the owner to establish a new owner of the object in a subsequent call.
+func RemoveControllerReference(owner, controlled metav1.Object) error {
```
vincepri, line 96:
```suggestion
func RemoveControllerReference(owner, object metav1.Object) error {
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -91,6 +91,29 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	return nil
 }
 
+// RemoveControllerReference is a helper method to make sure the given object removes an controller reference to the object provided.
```
vincepri, line 94:
```suggestion
// RemoveControllerReference is a helper method to make sure the given object removes a controller reference to the object provided.
```

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -91,6 +91,29 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
...
+	result := []metav1.OwnerReference{}
+	if length < 1 {
+		return fmt.Errorf("%T does not have any owner references", controlled)
+	}
+	for _, ownerref := range owners {
+		if ownerref.Name == owner.GetName() {
+			continue
+		}
+		result = append(result, ownerref)
+	}
```
vincepri, line 108: Could we use `indexOwnerRef` here, find the index (if any) and cut the slice instead of creating a brand new one?

troy0820: I was trying to do that but was failing miserably as my slice foo isn't there.

Also, I didn't want to stomp on situations where we would try to re-slice an index of 1 and be out of bounds by `[index+1]`.  It seemed premature to handle a specific case where we would re-slice for length of 1 instead of allocating a new slice and re-set.  

I can use this method to see if I  can do that then what I'm currently doing.

troy0820: Couple of questions: @vincepri 
1. Re-slicing, I still have to `SetControllerReference` once it's done, do we need to keep the same slice? 
2. This causes me to make more method(s) to get the GVK of the object which doesn't have any exposed method to do so.  The method you are suggesting gets the owner ref from the ownerReferences to do this where that information is available. The `ref.Name == owner.GetName()` isn't available in that function for me to migrate this over to that.

pkg/controller/controllerutil/controllerutil.go
troy0820: Resolving as I updated this on the latest push.

pkg/controller/controllerutil/controllerutil.go
troy0820: Resolving because I updated this on the latest push.

pkg/controller/controllerutil/controllerutil.go
vincepri: Similarly how the other functions handle it sounds reasonable, https://github.com/kubernetes-sigs/controller-runtime/pull/2455/files#diff-049ddcd31ccf7b24e1e222739d4e5129f1ea5583be912bacc6b0fb5251695b12R120-R133

troy0820: Thanks for the suggestion @vincepri  I updated the logic to use that function and updated the unit tests as well.

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -256,6 +255,98 @@ var _ = Describe("Controllerutil", func() {
...
+		})
+		It("should only fail because the object is not a runtime.Object", func() {
+			var obj metav1.Object
+			rs := &appsv1.ReplicaSet{}
+			dep := &extensionsv1beta1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "foo", UID: "foo-uid"},
+			}
+			Expect(controllerutil.SetControllerReference(dep, rs, scheme.Scheme)).NotTo(HaveOccurred())
+			Expect(len(rs.GetOwnerReferences())).To(BeEquivalentTo(1))
+			Expect(controllerutil.RemoveControllerReference(obj, rs, runtime.NewScheme())).To(HaveOccurred())
```
troy0820, line 346: I am going to change this to use the actual scheme and not the new scheme that is blank.

troy0820: Resolving, updated on latest push with the scheme adjusted to show that the error is coming from the object and not the scheme in function.

troy0820:
/assign @vincepri 

Could you take another look at this @vincepri, not sure what else we need with this.

PR #2460, opened 2023-08-29
🐛 Refactor tests to drop hard otel dependency

howardjohn:
Since v0.16.0 (#2407), there is an import for all users of various opentelemetry libraries. This is caused by manager test -> pkg/metrics/filters -> k8s.io/apiserver -> otel-go.

The issue is API server users a otel-go library from >1 year ago. This makes it impossible to import a modern otel-go library and a modern controller-runtime library together.

Go is unable to prune the dependency *even though its only used in tests* (TBH, this surprised me!). By moving the tests that use `filters` under the `filters` package, though, this drops the required dependency on otel-go. Users that import the `filters` package will use it, of course, but everyone does not need to.

So basically this refactors tests, but has user facing changes -- fewer dependencies are required to import controller-runtime core, allowing using the newer controller-runtime with newer otel-go.

In parallel I will attempt to get k/k to update their otel-go version, but this will take months to ship at the earliest. IMO, this is worth fixing in the meantime as a v0.16.x patch.


pkg/metrics/filters/filters_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+			BeforeEach(func() {
+				srv = nil
+				opts = manager.Options{
+					Metrics: metricsserver.Options{
+						BindAddress:    ":0",
+						SecureServing:  true,
+						FilterProvider: WithAuthenticationAndAuthorization,
+					},
+					NewMetricsServer: func(options metricsserver.Options, config *rest.Config, httpClient *http.Client) (metricsserver.Server, error) {
```
sbueringer, line 43: I know it's a bit hacky, but I would prefer setting this via reflection in this test instead of exposing the function.

We did something similar here: https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/cache/cache_test.go#L546-L553

I think it's just better to have slightly hacky unit tests vs. exposing an option to CR users that we might not want to expose

pkg/metrics/filters/filters_test.go
```diff
@@ -0,0 +1,253 @@
...
+						}
+						return srv, err
+					},
+				}
+				httpClient = &http.Client{Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+				}}
+			})
+
+			It("should serve metrics in its registry", func() {
```
sbueringer, line 57: Looks like we lost some test cases. Was this intentional?

pkg/metrics/filters/filter_suite_test.go
```diff
@@ -0,0 +1,64 @@
+package filters
```
sbueringer, line 1: These files probably should have the copyrights header

Not good that CI is not capturing it, but that's a separate issue :)

sbueringer:
> Go is unable to prune the dependency even though its only used in tests (TBH, this surprised me!)

Very interesting. I also didn't expect that and tried to implement the feature in a way that exactly something like this does not happen.

> In parallel I will attempt to get k/k to update their otel-go version, but this will take months to ship at the earliest. IMO, this is worth fixing in the meantime as a v0.16.x patch.

That would be great, would be nice if you can link the issue so we can follow along.

> IMO, this is worth fixing in the meantime as a v0.16.x patch.

Absolutely agree. Thx for reporting and fixing this so quickly. Let's get this PR merged, cherry-picked and released asap

pkg/metrics/filters/filters_test.go
howardjohn: I had thought I captured the authz specific ones here and the general metrics ones left in manager/. but maybe not, let me double check

sbueringer: Just compared and saw that we lost a few. But they were the easiest ones. Not sure how much additional code paths they cover. If it's nothing new definitely fine to not add them again

howardjohn:
Somehow https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.17.0 magically fixes this. I have no clue how, TBH. This is probably still desired anyways but not as important

sbueringer:
> Somehow https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.17.0 magically fixes this. I have no clue how, TBH. This is probably still desired anyways but not as important

I'm fine with going ahead with this PR regardless

sbueringer:
Thank you!
/lgtm
/approve

sbueringer:
@howardjohn Would be nice if you can open a cherry-pick for release-0.16. I'll merge & release after.

howardjohn:
@sbueringer opened https://github.com/kubernetes-sigs/controller-runtime/pull/2465, thanks for the review+help

sbueringer:
You're welcome!

PR #2462, opened 2023-08-30
🐛Create `RemoveOwnerReference` function in controllerutil

troy0820:
Not sure if this is useful but when trying to remove an owner reference, there is no util function that is available to do this.  This is also the case with `RemoveControllerReference` #2455.  

Creating this function allows a user to remove the owner reference of an object.  
Also helps with https://github.com/kubernetes-sigs/controller-runtime/issues/2381

- [x] Create function that removes owner reference from object _(similar to remove finalizer)_
- [x] Create unit test for function

```release-note
Add function RemoveOwnerReference to pkg/controller/controllerutil 
to allow the removal of the owner reference of the controlled object.```


sbueringer:
lgtm pending the linter findings

sbueringer:
Thank you!

/lgtm
/approve

Issue #2454, opened 2023-08-22
Leader election events are recorded with the wrong HTTP client


rubofelipe:
The version v0.15.0 introduced an issue for controllers that use a separated kubeconfig for leader election:
```
mgr, err := ctrl.NewManager(configA, manager.Options{
    ...
    LeaderElectionConfig: configB,
})
```
The lease object is correctly created in the cluster pointed by the `configB` and the controller is able to acquire the leadership. However, an error is returned on creation of the Event object associated to the coordination object (e.g Lease):

```
E0821 20:38:02.494486       1 event.go:289] Unable to write event: '&v1.Event{*****}': 'Post "https://api.****:443/api/v1/namespaces/my-namespace/events": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "ca")'(may retry after sleeping)
```

The kubernetes API server is not trusting the certificate authority presented by the HTTP client.

The origin of this problem is the commit https://github.com/kubernetes-sigs/controller-runtime/commit/4fd4f6e7116453ea4e07a2cdd8ba0d38ad64680f

A new an unique HTTP client was created into `Options` struct:
https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/cluster/cluster.go#L264-L272

This HTTP client is being copied to the Cluster struct and being passed as argument to the function that creates the leaderRecorderProvider:
https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/manager/manager.go#L351-L360

So the `options.newRecorderProvider` is receiving the LeaderElectionConfig (i.e. `configB`) and an HTTPClient based on `configA`. The `options.newRecorderProvider` is a function pointer that is pointing to `NewProvider` function:

https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/internal/recorder/recorder.go#L114-L126

The documentation of NewForConfigAndClient function gives us an important info:
```
// NewForConfigAndClient creates a new CoreV1Client for the given config and http client.
// Note the http client provided takes precedence over the configured transport values.
func NewForConfigAndClient(c *rest.Config, h *http.Client) (*CoreV1Client, error) {
```
[...]

PR #2464, opened 2023-08-31
🐛 Use http client from leaderElectionConfig

troy0820:
During creation of the manager, the default `cluster.HTTPClient` is being used during leader election causing cert issues and not responding to events because of the conflict of the kubeconfig it uses for the manager.  It never uses the kubeconfig from the leaderElectionConfig when being set.  

This PR sets that value and creates a httpClient from the restConfig passed in on the `options.LeaderElectionConfig`

- [x] Use the http client from the leaderElectionConfig from options in manager 

Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/2454

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -364,7 +364,11 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
Thanks!

PR #2465, opened 2023-08-31
[release-0.16] 🐛 Refactor tests to drop hard otel dependency

howardjohn:
Since v0.16.0 (#2407), there is an import for all users of various opentelemetry libraries. This is caused by manager test -> pkg/metrics/filters -> k8s.io/apiserver -> otel-go.

The issue is API server users a otel-go library from >1 year ago. This makes it impossible to import a modern otel-go library and a modern controller-runtime library together.

Go is unable to prune the dependency *even though its only used in tests* (TBH, this surprised me!). By moving the tests that use `filters` under the `filters` package, though, this drops the required dependency on otel-go. Users that import the `filters` package will use it, of course, but everyone does not need to.

So basically this refactors tests, but has user facing changes -- fewer dependencies are required to import controller-runtime core, allowing using the newer controller-runtime with newer otel-go.

In parallel I will attempt to get k/k to update their otel-go version, but this will take months to ship at the earliest. IMO, this is worth fixing in the meantime as a v0.16.x patch.

(cherry picked from commit d17751f5663e32c1afa19c368ef2509b96e2a1f2)

(Manual cherry-pick from: #2460)


sbueringer:
Thank you!!!

/lgtm
/approve

PR #2471, opened 2023-09-05
🌱  Implement missing client methods for komega interface

Danil-Grigorev:
This PR adds some missing interface members for komega, making it more suitable for general use in tests.

### DIFF
--- a/pkg/envtest/komega/default.go
+++ b/pkg/envtest/komega/default.go
@@ -79,6 +79,30 @@ func UpdateStatus(obj client.Object, f func(), opts ...client.SubResourceUpdateO
 	return defaultK.UpdateStatus(obj, f, opts...)
 }
 
+// Delete returns a function that deletes a resource and returns the occurring error.
+// It can be used with gomega.Eventually() like this:
+//
+//	deployment := appsv1.Deployment{ ... }
+//	gomega.Eventually(k.Delete(&deployment).To(gomega.Succeed())
+//
+// By calling the returned function directly it can also be used as gomega.Expect(k.Delete(...)()).To(...)
+func Delete(obj client.Object, opts ...client.DeleteOption) func() error {
+	checkDefaultClient()
+	return defaultK.Delete(obj, opts...)
+}
+
+// DeleteAllOf returns a function that deletes a list of resources and returns the occurring error.
+// It can be used with gomega.Eventually() like this:
+//
+//	deployments := appsv1.Deployment{ ... }
+//	gomega.Eventually(k.DeleteAllOf(&deployments, client.InNamespace("default")).To(gomega.Succeed())
+//
+// By calling the returned function directly it can also be used as gomega.Expect(k.DeleteAllOf(...)()).To(...)
+func DeleteAllOf(obj client.Object, opts ...client.DeleteAllOfOption) func() error {
+	checkDefaultClient()
+	return defaultK.DeleteAllOf(obj, opts...)
+}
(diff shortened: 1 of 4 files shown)


schrej:
The idea behind these functions was to help with handling eventual consistency in the apiserver. After an object is created, it can take multiple GET calls until it's actually returned. In addition, when you test controllers, it can take a while until some modification is performed.
When running actions like create or delete it's usually not necessary to retry them, since they are very likely to  be successful in a test setup. Instead, you're much more likely to wait until a deleted object is actually gone, which can be done using `Eventually(Get()).Should(NotSucceed())`.
For the same reason there are no Create functions.
Update is a little out of line - I think the main motivation was the Get part in the beginning of it.

If you have a good use-case that requires these methods, we can add them. In that case I'd go all the way though, and also add a method to create objects (and add package level functions for all of them).

Danil-Grigorev:
I think komega in our case is used to wait for some resource to reach expected state during e2e run. Our setup requires a controller to be started, which in turn will create a resource, and then we want to delete it. So something like
```go
		By("Removing cluster record, and waiting for it to be removed")
		Eventually(func() error {
			return bootstrapClusterProxy.GetClient().Delete(ctx, cluster)
		}, e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(Succeed())
		Eventually(Get(rancherCluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(MatchError(ContainSubstring(…..
```
could be simplified to
```go
		By("Removing rancher cluster record, and waiting for it to be removed")
		Eventually(Delete(cluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(Succeed())
		Eventually(Get(cluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(MatchError(ContainSubstring( …..
```
which improves readability without sacrificing functionality. It additionally helps to rule out some transient errors, like 408 from client opened to a distant and resource constraint k8s cluster. This scenario could be handled with an additional `get` call above, but I’d prefer to have options.

As for create scenario, I agree, this should be added, if you think this is a sufficient use-case to get this in. Could help in situations when a webhook configuration is slow to come up in a freshly deployed controller instance, and we immediately want to apply some custom resource it should process.

schrej:
Sorry, I lost track of this PR. That's a fair point. My question would be: why not wait until the object exists before trying to delete it? So instead of your example, it would look like this:
```
By("Removin cluster record, and waiting for it to be removed")
Eventually(Get(cluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(MatchError(ContainSubstring( …..
Expect(Delete(cluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).To(Succeed())
Eventually(Get(cluster), e2eConfig.GetIntervals(bootstrapClusterProxy.GetName(), "wait-controllers")...).Should(MatchError(ContainSubstring( …..
```
Don't get me wrong, I'm not fundamentally against adding more functions, I just don't see the point of waiting for deletion or creation to succeed. In a working test setup I'd just expect those calls to work. If they don't work every time, the test setup is broken.
For calls like Get and the other helpers in this package, it's expected that they do not succeed immediately, even if the test setup is working perfectly fine.

Issue #2354, opened 2023-05-26
Usage of client-go's `discovery.ErrGroupDiscoveryFailed` breaks apimachinery `errors.IsNotFound`


saschagrunert:
The introduction of the error conversion in controller-runtime v0.15.0:

https://github.com/kubernetes-sigs/controller-runtime/blob/30eae58f1b984c1b8139dd9b9f68dd2d530ed429/pkg/client/apiutil/restmapper.go#L275-L290

breaks the use case of checking available APIs via `k8s.io/apimachinery/pkg/api/errors.IsNotFound`:

https://github.com/kubernetes/kubernetes/blob/2e306e9632b13a3c1530a7e7ddf52a50f4741523/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L525-L536

Reason is the error conversion, which now needs a workaround like this:
https://github.com/kubernetes-sigs/security-profiles-operator/pull/1710/files#diff-fb01936f787b03e615f35a185799e91ec41896c1fcae0991db3a24137f21cf54R78-R87

Do we want to provide an additional API or something similar to make availability checks simpler?

alvaroaleman:
@Fedosin What is the reason for using `discovery.ErrGroupDiscoveryFailed` rather than just passing on the error we got from `ServerResourcesForGroupVersion`?

ebensh:
> @Fedosin What is the reason for using `discovery.ErrGroupDiscoveryFailed` rather than just passing on the error we got from `ServerResourcesForGroupVersion`?

Is this an approach we can take as a fix / patch? I understand that the v0.15.0 contained several changes that had the potential to break us, but detecting whether resources exist is a common operation for controllers to perform, and breaking the pattern of detecting whether a resource exists or not is a problem.

dmitri-d:
Just encountered this issue as well.

porridge:
Note that this also breaks use of `k8s.io/apimachinery/pkg/api/meta.IsNoMatchError` on the resulting error.

I wonder if this will work:
/kind bug
/kind regression

porridge:
I've taken a stab at fixing this in https://github.com/kubernetes-sigs/controller-runtime/pull/2472
Please take a look @alvaroaleman @Fedosin

PR #2472, opened 2023-09-06
:bug: Return NoResourceMatchError when appropriate for backwards compatibility.

porridge:
Fixes: #2354

### DIFF
--- a/pkg/client/apiutil/errors.go
+++ b/pkg/client/apiutil/errors.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apiutil
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// ErrResourceDiscoveryFailed is returned if the RESTMapper cannot discover supported resources for some GroupVersions.
+// It wraps the errors encountered, except "NotFound" errors are replaced with meta.NoResourceMatchError, for
+// backwards compatibility with code that uses meta.IsNoMatchError() to check for unsupported APIs.
+type ErrResourceDiscoveryFailed map[schema.GroupVersion]error
+
+// Error implements the error interface.
+func (e *ErrResourceDiscoveryFailed) Error() string {
+	subErrors := []string{}
+	for k, v := range *e {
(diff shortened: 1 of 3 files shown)


pkg/client/apiutil/errors.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	sort.Strings(subErrors)
+	return fmt.Sprintf("unable to retrieve the complete list of server APIs: %s", strings.Join(subErrors, ", "))
+}
+
+func (e *ErrResourceDiscoveryFailed) Unwrap() []error {
+	subErrors := []error{}
+	for gv, err := range *e {
+		if apierrors.IsNotFound(err) {
+			err = &meta.NoResourceMatchError{PartialResource: gv.WithResource("")}
```
alvaroaleman, line 49: This confuses me - In the initial issue description, @saschagrunert mentions that they expected to match through IsNotFound: https://github.com/kubernetes-sigs/controller-runtime/issues/2354#issue-1727232531 - Which should this be?

porridge:
@alvaroaleman good point, let's see.

I created [a tiny repro program](https://github.com/porridge/controller-runtime/blob/check-on-v0.14.6/t/retval_test.go) to investigate the behaviour of an actual **client** that uses the mapper, and ran it on a few versions:
* pre-v0.15
   ```
    deleg.go:124: INFO ret-error-demo: Kind is present but object was not found kind="ReplicaSet"
    deleg.go:124: INFO ret-error-demo: Kind does not exist kind="Badger"
    deleg.go:124: INFO ret-error-demo: Kind does not exist kind="Gherkin"
   ```
* curent tip of `main` ([commit](https://github.com/porridge/controller-runtime/commit/89ae07ce10bad2f4d1c73f03df025c59438208b2)):
   ```
    deleg.go:127: I0911 12:22:01.453070] ret-error-demo: Kind is present but object was not found kind="ReplicaSet"
    deleg.go:127: I0911 12:22:01.453660] ret-error-demo: Kind does not exist kind="Badger"
    deleg.go:142: E0911 12:22:01.454161] ret-error-demo: Unexpected error err="failed to get API group resources: unable to retrieve the complete list of server APIs: inexistent.group.com/v1: the server could not find the requested resource"
   ```
* on top of this PR ([commit](https://github.com/porridge/controller-runtime/commit/fea1836291f7605b6e5d3a3b49cd6610527aa6a2)):
   ```
    deleg.go:127: I0911 12:24:06.089190] ret-error-demo: Kind is present but object was not found kind="ReplicaSet"
    deleg.go:127: I0911 12:24:06.089777] ret-error-demo: Kind does not exist kind="Badger"
    deleg.go:127: I0911 12:24:06.090234] ret-error-demo: Kind does not exist kind="Gherkin"
   ```

BTW, this also shows that the regression is only for the case of missing group(version) - error returned on a missing kind in an existing group+version is ok w.r.t. backwards compatibility. Unfortunately the missing group/version case is the most interesting one, as it's used to discover capabilities supported by the cluster.

My theory on why @saschagrunert mentioned `IsNotFound` is that it was a mistake or simply lack of precision. At least judging by the code he mentions, it contains a [single method that ORs together checks](https://github.com/kubernetes-sigs/security-profiles-operator/pull/1710/files#diff-fb01936f787b03e615f35a185799e91ec41896c1fcae0991db3a24137f21cf54R73-R74) done by `IsNoMatchError` and `IsNotFound`.

The conditional branches in the demo program I created mimics [actual the code in my operator](https://github.com/stackrox/stackrox/blob/de940b3d011d5bcb9a2519f916664ff4ce25588f/operator/pkg/securedcluster/scanner/auto_sense.go#L72-L81) which is careful to distinguish between errors that satisfy `IsNoMatchError` vs `IsNotFound`.

alvaroaleman (approved):
Thanks for fixing this!

PR #2473, opened 2023-09-07
✨ Export HTTP server manager runnable implementation

joelanford:
In operator-framework, we have run into a few situations where we need to run a custom HTTP server inside a controller process. In order to decouple from assumptions made by the metrics server (which allows extra handlers to be added), we  need a fully separate server. Since controller-runtime already has a generic HTTP server runnable for use with health probes and pprof handlers, it seems useful to export this implementation to users as well so that they can run their own HTTP servers without having to worry about implementing the `manager.Runnable` interface correctly.

### DIFF
--- a/pkg/manager/server.go
+++ b/pkg/manager/server.go
@@ -21,34 +21,67 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"time"
 
-	"github.com/go-logr/logr"
+	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// server is a general purpose HTTP server Runnable for a manager
-// to serve some internal handlers such as health probes, metrics and profiling.
-type server struct {
-	Kind     string
-	Log      logr.Logger
-	Server   *http.Server
+var (
+	_ Runnable               = (*Server)(nil)
+	_ LeaderElectionRunnable = (*Server)(nil)
+)
+
+// Server is a general purpose HTTP server Runnable for a manager.
+// It is used to serve some internal handlers for health probes and profiling,
+// but it can also be used to run custom servers.
+type Server struct {
+	// Name is an optional string that describes the purpose of the server. It is used in logs to distinguish
+	// among multiple servers.
+	Name string
+
+	// Server is the HTTP server to run. It is required.
+	Server *http.Server
+
+	// Listener is an optional listener to use. If not set, the server start a listener using the server.Addr.
+	// Using a listener is useful when the port reservation needs to happen in advance of this runnable starting.
 	Listener net.Listener
+
+	// OnlyServeWhenLeader is an optional bool that indicates that the server should only be started when the manager is the leader.
+	OnlyServeWhenLeader bool
+
+	// ShutdownTimeout is an optional duration that indicates how long to wait for the server to shutdown gracefully. If not set,
+	// the server will wait indefinitely for all connections to close.
+	ShutdownTimeout *time.Duration
 }
 
-func (s *server) Start(ctx context.Context) error {
-	log := s.Log.WithValues("kind", s.Kind, "addr", s.Listener.Addr())
+// Start starts the server. It will block until the server is stopped or an error occurs.
+func (s *Server) Start(ctx context.Context) error {
+	log := crlog.FromContext(ctx)
+	if s.Name != "" {
+		log = log.WithValues("name", s.Name)
+	}
+	log = log.WithValues("addr", s.addr())
 
 	serverShutdown := make(chan struct{})
 	go func() {
 		<-ctx.Done()
 		log.Info("shutting down server")
-		if err := s.Server.Shutdown(context.Background()); err != nil {
+
+		shutdownCtx := context.Background()
+		if s.ShutdownTimeout != nil {
+			var shutdownCancel context.CancelFunc
+			shutdownCtx, shutdownCancel = context.WithTimeout(context.Background(), *s.ShutdownTimeout)
+			defer shutdownCancel()
+		}
+
+		if err := s.Server.Shutdown(shutdownCtx); err != nil {
 			log.Error(err, "error shutting down server")
 		}
 		close(serverShutdown)
 	}()
 
 	log.Info("starting server")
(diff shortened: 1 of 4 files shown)


pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
+	// the server will wait indefinitely for all connections to close.
+	ShutdownTimeout *time.Duration
 }
 
-func (s *server) Start(ctx context.Context) error {
-	log := s.Log.WithValues("kind", s.Kind, "addr", s.Listener.Addr())
+// Start starts the server. It will block until the server is stopped or an error occurs.
+func (s *Server) Start(ctx context.Context) error {
+	log := s.Log
+	if log.GetSink() == nil {
```
stevekuznetsov, line 65: Can `s.Log` itself be `nil` here?

pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
-	Server   *http.Server
+// Server is a general purpose HTTP server Runnable for a manager.
+// It is used to serve some internal handlers for health probes and profiling,
+// but it can also be used to run custom servers.
+type Server struct {
+	// Kind is an optional string that describes the purpose of the server. It is used in logs to distinguish
+	// among multiple servers.
+	Kind string
+
+	// Log is the logger used by the server. If not set, a logger will be derived from the context passed to Start.
```
stevekuznetsov, line 44: Why not only have one way to pass it in?

pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
+
+	// Server is the HTTP server to run. It is required.
+	Server *http.Server
+
+	// Listener is an optional listener to use. If not set, the server start a listener using the server.Addr.
+	// Using a listener is useful when the port reservation needs to happen in advance of this runnable starting.
 	Listener net.Listener
+
+	// OnlyLeader is an optional bool that indicates that the server should only be started when the manager is the leader.
+	OnlyLeader bool
```
stevekuznetsov, line 55: nit: might prefer something more descriptive like `OnlyServeWhenLeader`

pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
 	"net/http"
+	"time"
 
 	"github.com/go-logr/logr"
+
+	crlog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var (
+	_ Runnable               = &Server{}
```
stevekuznetsov, line 32: nit: I think the other invocation keeps you from allocating `Server`s

```
var (
    _ Runnable = (*Server)(nil)
    _ LeaderElectionRunnable = (*Server)(nil)
)
```

stevekuznetsov (commented):
Looks good, some minor comments

pkg/manager/server.go
joelanford: No, `s.Log` is a non-pointer concrete struct. This is a fairly standard check: https://github.com/search?q=org%3Akubernetes-sigs+GetSink%28%29+%3D%3D+nil&type=code

pkg/manager/server.go
joelanford: I didn't want to require a `Server` struct to be setup with a logger in advance. I don't have a strong opinion here though.

pkg/manager/server.go
joelanford: Doesn't the compiler optimize out the allocation because of the `_` variable assignment?

joelanford: Either way, official docs use your example. I'll switch over. 

https://go.dev/doc/effective_go#blank_implements

pkg/manager/server.go
vincepri: Could we remove the Log here, and just inherit the logger from Start?

pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
-	Kind     string
-	Log      logr.Logger
-	Server   *http.Server
+// Server is a general purpose HTTP server Runnable for a manager.
+// It is used to serve some internal handlers for health probes and profiling,
+// but it can also be used to run custom servers.
+type Server struct {
+	// Kind is an optional string that describes the purpose of the server. It is used in logs to distinguish
+	// among multiple servers.
+	Kind string
```
vincepri, line 42: 🤔 Maybe `Name` or similar? `Kind` is an overloaded word, and I immediately thought it was related to GVK

pkg/manager/server.go
```diff
@@ -21,41 +21,97 @@ import (
...
 	<-serverShutdown
 	return nil
 }
 
-func (s *server) NeedLeaderElection() bool {
-	return false
+// NeedLeaderElection returns true if the server should only be started when the manager is the leader.
+func (s *Server) NeedLeaderElection() bool {
+	return s.OnlyServeWhenLeader
+}
```
vincepri, line 103: Is there a use case for this?

sbueringer:
@joelanford Are you still interested in this PR?

pkg/manager/server.go
joelanford: The use case I have in mind is when there is a strong coupling between what is being reconciled and what is being served. One example might be an API that represents OCI images, where creating an instance of that API makes the OCI image contents available as a tar archive via the HTTP server.

pkg/manager/server.go
joelanford: This was already present in the unexported `server` struct, so I wasn't going to change it.

There is existing code in the pprof and health probe server setup that explicitly sets the servers' logger to the controller manager's logger (which comes from manager.Options.Logger). 

There's a bunch of context plumbing that makes it difficult to tell which contexts are used where, but I can't find anywhere that injects the controllerManager's logger into the context that is used to start the http servers. 

We could potentially make a custom context with the logger injected and use that context _just_ with the HTTPServers group to preserve existing behavior.

WDYT?

sbueringer (commented):
Just one nit, otherwise lgtm

sbueringer:
/lgtm

Will fix the nit in a follow up

PR #2476, opened 2023-09-08
Fake client.Update() should not store omit empty Status fields

berlin-ab:
[x] - solve for typed object Update()
[ ] - solve for unstructured object Update()


troy0820:
/hold 

I know this is a draft to show that the test fails with the logic you expect, however adding hold to help with triage and not a work in progress PR.

berlin-ab:
> I know this is a draft to show that the test fails with the logic you expect, however adding hold to help with triage and not a work in progress PR.

Yeah, thanks for clarifying. This is not an in-progress PR, but more of a proof of concept/bug.

Issue #2474, opened 2023-09-07
The fake client Status().Update() should not store object meta


berlin-ab:
```
obj.SetAnnotations(map[string]string{
	"foo": "bar",
})
			
err = testClient.Status().Update(ctx, obj)
require.NoError(t, err)

_ = testClient.Get(ctx, k8sclient.ObjectKeyFromObject(obj), obj)
require.Equal(t, cluster.Annotations["foo"], "bar")
```
			
This test fails when using a real client, but succeeds when using the fake client.

berlin-ab:
The same is true of labels.

troy0820:
/kind support 
What version of controller runtime is this? 
Also if you put that `Get` in a new “obj” bar and not shadow it, does this happen?

troy0820:
```
 _ = testClient.Get(ctx, k8sclient.ObjectKeyFromObject(obj), obj)
require.Equal(t, cluster.Annotations["foo"], "bar")
```
This also doesn’t seem to test `obj` but asserts that the cluster.Annotations[“foo”] is equal to bar. Is this `obj.Annotations[“foo”]`?

berlin-ab:
> This also doesn’t seem to test `obj` but asserts that the cluster.Annotations[“foo”] is equal to bar. Is this `obj.Annotations[“foo”]`?

Sorry, I was changing some variable names to make it less specific to my situation and missed one.

berlin-ab:
> What version of controller runtime is this?

v0.15.2

berlin-ab:
I've modified the fake client test suite to show an example:

https://github.com/kubernetes-sigs/controller-runtime/pull/2475/files

berlin-ab:
```
Fake client [It] should not change non-status field of typed objects that have a status subresource on status update
/Users/aberlin/workspace/controller-runtime/pkg/client/fake/client_test.go:1445

  [FAILED] Expected
      <string>:   &v1.Node{
        	TypeMeta: {Kind: "Node", APIVersion: "v1"},
        	ObjectMeta: v1.ObjectMeta{
        		... // 8 identical fields
        		DeletionTimestamp:          nil,
        		DeletionGracePeriodSeconds: nil,
      - 		Labels:                     nil,
      + 		Labels:                     map[string]string{"some-label-key": "some-label-value"},
      - 		Annotations:                nil,
      + 		Annotations:                map[string]string{"some-annotation-key": "some-annotation-value"},
[...]

PR #2479, opened 2023-09-08
🐛 Do not update anything but status when using subresource client 

troy0820:
During updating with the status client, it is shown that the metadata/labels will come with the subresource client.  To mitigate this, updating the method to only use the old object in the tracker, will allow the status to be copied from the new object and put in the update chain to the fake client tracker. 

Resolves: https://github.com/kubernetes-sigs/controller-runtime/issues/2474

- [x] Update fake client subreresource client method update to only copy status before passing to tracker.
- [x] Include test from issue to show this is working properly  cc @berlin-ab _(cherry-picked from draft PR)_
- [x] Create test case to show other status fields do not get written over.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1108,12 +1108,30 @@ func (sw *fakeSubResourceClient) Create(ctx context.Context, obj client.Object, 
 func (sw *fakeSubResourceClient) Update(ctx context.Context, obj client.Object, opts ...client.SubResourceUpdateOption) error {
 	updateOptions := client.SubResourceUpdateOptions{}
 	updateOptions.ApplyOptions(opts)
-
-	body := obj
-	if updateOptions.SubResourceBody != nil {
-		body = updateOptions.SubResourceBody
+	gvr, err := getGVRFromObject(obj, sw.client.scheme)
+	if err != nil {
+		return err
 	}
-	return sw.client.update(body, true, &updateOptions.UpdateOptions)
+	o, err := sw.client.tracker.Get(gvr, obj.GetNamespace(), obj.GetName())
+	if err != nil {
+		return err
+	}
+	gvk, err := apiutil.GVKForObject(obj, sw.client.scheme)
+	if err != nil {
+		return err
+	}
+	body := o
+	if sw.client.tracker.withStatusSubresource.Has(gvk) {
+		err := copyStatusFrom(obj, body)
+		if err != nil {
+			return err
+		}
+	}
+	bodyObj := body.(client.Object)
+	if bodyObj.GetResourceVersion() != obj.GetResourceVersion() {
+		return apierrors.NewConflict(gvr.GroupResource(), obj.GetName(), fmt.Errorf("resource version conflict"))
+	}
+	return sw.client.update(bodyObj, true, &updateOptions.UpdateOptions)
 }
 
 func (sw *fakeSubResourceClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.SubResourcePatchOption) error {
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -950,46 +950,6 @@ func dryPatch(action testing.PatchActionImpl, tracker testing.ObjectTracker) (ru
 	return obj, nil
 }
 
-//lint:ignore U1000 function is causing errors with status updates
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -1108,12 +1108,30 @@ func (sw *fakeSubResourceClient) Create(ctx context.Context, obj client.Object,
 func (sw *fakeSubResourceClient) Update(ctx context.Context, obj client.Object, opts ...client.SubResourceUpdateOption) error {
 	updateOptions := client.SubResourceUpdateOptions{}
 	updateOptions.ApplyOptions(opts)
-
-	body := obj
-	if updateOptions.SubResourceBody != nil {
-		body = updateOptions.SubResourceBody
+	gvr, err := getGVRFromObject(obj, sw.client.scheme)
```
alvaroaleman, line 1111: Hey @troy0820 , thanks so much for jumping into this and providing a fix this quickly!

The root of the problem seems to be that `copyNonStatusFrom` is basically broken, because it is not going to clear fields that exist on the status object but not on the main object. Rather than working around that like you suggested, I'd prefer it if we kept the status handling in one place and avoid the code duplication, something like this:
```diff
index 18dd639d..d7b8427a 100644
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -399,10 +399,11 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
        }

        if t.withStatusSubresource.Has(gvk) {
-               if isStatus { // copy everything but status and metadata.ResourceVersion from original object
-                       if err := copyNonStatusFrom(oldObject, obj); err != nil {
-                               return fmt.Errorf("failed to copy non-status field for object with status subresouce: %w", err)
+               if isStatus { // copy status from object into oldObject, then set object to a copy of oldObject
+                       if err := copyStatusFrom(obj, oldObject); err != nil {
+                               return fmt.Errorf("failed to copy the status for object with status subresource: %w", err)
                        }
+                       obj = oldObject.DeepCopyObject().(client.Object)
                } else { // copy status from original object
                        if err := copyStatusFrom(oldObject, obj); err != nil {
                                return fmt.Errorf("failed to copy the status for object with status subresource: %w", err)
```

(and delete `copyNonStatusFrom`)

troy0820: Yeah, I can do that @alvaroaleman.  I fear we may have this issue with client.Update() using the method to include status with the update there in another issue.  I can fix this and push up the changes and we can discuss on the other issue.

troy0820: This is going to fail lint because the function is not used.  I can add the lint override to not flag this if that is what we want to do.

alvaroaleman: Why not remove the func?

pkg/client/fake/client_test.go
```diff
@@ -1472,6 +1479,36 @@ var _ = Describe("Fake client", func() {
 		objOriginal.Status.NodeInfo.MachineID = "machine-id-from-status-update"
 		Expect(cmp.Diff(objOriginal, actual)).To(BeEmpty())
 	})
+	It("should not overwrite status fields of typed objects that have a status subresource on status update", func() {
```
alvaroaleman, line 1482: I think the tittle is incorrect and the number of negations makes me 🤯 . Did you mean to say `should not overwrite non-status fields of typed objects that have a status subresource on status update`? If yes, maybe express it as `Should only override status fields of typed objects that have a status subresource on status update`?

pkg/client/fake/client.go
troy0820: I can but didn't want to arbitrarily start deleting code.  I'll delete it.

pkg/client/fake/client_test.go
troy0820: I was trying to capture the impact that doing the workaround (which doesn't exist anymore) would only update the status fields that we bring to the update and not overwrite the status fields that exists already with blank fields, etc.  _(still a useful test)_

Changing it to what you suggested provides a clearer understanding though of what the test provides.  Negating the negations always  🤯

The inference of the update should include those fields not changing but asserting them in this test demonstrates that update only touching the fields it needs to write to.

alvaroaleman (approved):
Thanks for fixing this!

troy0820:
No problem, does this need to be cherry-picked? 

/cherry-pick release-0.16

PR #2481, opened 2023-09-09
🌱 Add request latency, rate limiter latency and request retry metrics

sbueringer:

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,24 +18,46 @@ package metrics
 
 import (
 	"context"
+	"net/url"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	clientmetrics "k8s.io/client-go/tools/metrics"
 )
 
-// this file contains setup logic to initialize the myriad of places
-// that client-go registers metrics.  We copy the names and formats
-// from Kubernetes so that we match the core controllers.
-
 var (
-	// client metrics.
-
 	requestResult = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "rest_client_requests_total",
 			Help: "Number of HTTP requests, partitioned by status code, method, and host.",
 		},
 		[]string{"code", "method", "host"},
+	)
+
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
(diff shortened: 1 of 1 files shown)


sbueringer:
/hold 
Let's discuss first.

@alvaroaleman I thought about this a bit more and the high cardinality problem that we had before was only caused by the "host" label. I think these metrics are valuable even without this label. WDYT about adding the metrics but without the hots label?

sbueringer:
cc @fabriziopandini (fyi)

pkg/metrics/client_go_adapter.go
```diff
@@ -18,25 +18,47 @@ package metrics
...
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb. Intentionally not by host to avoid high cardinality.",
+			Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
+		},
+		[]string{"verb"},
+	)
+
+	rateLimiterLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_rate_limiter_duration_seconds",
```
alvaroaleman, line 48: sig apimachineries guidance the last time around I asked was not to use client-side ratelimiting but rely on APF instead - Maybe skip this one?

pkg/metrics/client_go_adapter.go
```diff
@@ -18,25 +18,47 @@ package metrics
...
+			Help:    "Client side rate limiter latency in seconds. Broken down by verb. Intentionally not by host to avoid high cardinality.",
+			Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
+		},
+		[]string{"verb"},
+	)
+
+	requestRetry = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "rest_client_request_retries_total",
```
alvaroaleman, line 57: For my understanding, this will only ever be populated for watch requests by the informer, because in all other cases, the error gets bubbled up to the application and the application then has to decide if it wants to retry which entails the underlying machinery doesn't know its a retry or am I wrong?

alvaroaleman (commented):
This mostly seems ok. What I am starting to heavily dislike is the fact that its impossible to configure this. Can you think of any way we could make this configurable?

sbueringer:
/close 
for now

Issue #2478, opened 2023-09-08
The fake client.Update() updates the status fields


ahmetb:
I'm observing that using the non-subresource `fake.Client` on both in-tree types (like corev1.Node or corev1.Pod) as well as Custom Resources actually save modifications on `status` fields to the resource (and are reflected on subsequent `Get` calls).

A minimal repro: https://gist.github.com/ahmetb/50784526a54244d45022a7ed8d556a62
This is the case even if I do `fake.NewClientBuilder().WithStatusSubresource(&corev1.Pod{})`

Expected behavior is to be the same as a non-fake client where if you modify a `status` field and use `client.Update(ctx, obj)` the modifications on the `status` field are discarded.

/kind bug

troy0820:
This seems to replicate the behavior in issue https://github.com/kubernetes-sigs/controller-runtime/issues/2474 but the other way around.  It seems to me that the fake client is copying over everything and not discarding the status when going through the update chain.  PR I have for that https://github.com/kubernetes-sigs/controller-runtime/pull/2479 solves that case however long term we may need to decouple the methods that the fakeclient uses.

ahmetb:
For whatever reason, the test case that's supposed to catch this acts differently on different fields of corev1.Node.Status. I'll debug further, but this patch to the test shows the bug:

```
diff --git a/pkg/client/fake/client_test.go b/pkg/client/fake/client_test.go
index 1350fa2..c823282 100644
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -1420,6 +1420,7 @@ var _ = Describe("Fake client", func() {
 		cl := NewClientBuilder().WithStatusSubresource(obj).WithObjects(obj).Build()
 
 		obj.Status.NodeInfo.MachineID = "updated-machine-id"
+		obj.Status.Phase = corev1.NodeRunning
 		Expect(cl.Update(context.Background(), obj)).To(Succeed())
 
 		Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(obj), obj)).To(Succeed())
```

ahmetb:
I think here's where the problem is: fake client uses json marshaling/unmarshaling under the covers to assign the `.status` field from old object to new.

[...]

PR #2484, opened 2023-09-09
:bug: Fix status subresource getting updated on Update when it is empty

alvaroaleman:
Before this, Update and Patch requests will override to-level status fields for objects that have a status resource if the status was initially empty.

This happens because the code to set the status to what we already have in the unupdated object converts the object into a `map[string]any` using json serialization/deserialization, copies the original `status` field, then converts back. The json deserializer unfortunatelly does not properly clear the target, allowing data to remain that is absent in the json serialization.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2478

/cherrypick release-0.16

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -995,6 +995,7 @@ func fromMapStringAny(u map[string]any, target runtime.Object) error {
 		return fmt.Errorf("failed to serialize: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


alvaroaleman:
Interesting, the bot doesn't pick up cherrypick commands from the PR description
/cherrypick release-0.16

PR #2486, opened 2023-09-10
Draft: Test that an object is updatable after updating its status.

berlin-ab:
Test fails at:

```
  [FAILED] Unexpected error:
      <*errors.StatusError | 0xc0005ac000>: 
      Operation cannot be fulfilled on nodes "node": object was modified
      {
          ErrStatus: {
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Operation cannot be fulfilled on nodes \"node\": object was modified",
              Reason: "Conflict",
              Details: {Name: "node", Group: "", Kind: "nodes", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 409,
          },
      }
      
  occurred
  In [It] at: /Users/aberlin/workspace/controller-runtime/pkg/client/fake/client_test.go:1506 @ 09/09/23 20:52:18.056
```

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -403,6 +403,7 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
 			if err := copyStatusFrom(obj, oldObject); err != nil {
 				return fmt.Errorf("failed to copy non-status field for object with status subresouce: %w", err)
 			}
+
 			obj = oldObject.DeepCopyObject().(client.Object)
 		} else { // copy status from original object
 			if err := copyStatusFrom(oldObject, obj); err != nil {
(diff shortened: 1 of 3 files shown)


berlin-ab:
/hold

This is just a proof of concept test.

pkg/client/fake/client_test.go
```diff
@@ -1476,6 +1476,84 @@ var _ = Describe("Fake client", func() {
...
+		obj.Status.NodeInfo.MachineID = "machine-id-from-status-update"
+		Expect(cl.Status().Update(context.Background(), obj)).NotTo(HaveOccurred())
+
+		obj.Annotations = map[string]string{
+			"some-annotation-key": "some",
+		}
+		expectedObj.Annotations = map[string]string{
+			"some-annotation-key": "some",
+		}
+		Expect(cl.Update(context.Background(), obj)).NotTo(HaveOccurred())
```
berlin-ab, line 1506: This test fails. The test below was double checking that the reverse order also works.

pkg/client/fake/client.go
```diff
@@ -436,6 +437,14 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
 	intResourceVersion++
 	accessor.SetResourceVersion(strconv.FormatUint(intResourceVersion, 10))
 
+	// obtain the current obj accessor's pointer,
```
berlin-ab, line 440: I was looking for a cleaner way to do this, so maybe you can think of one. I don't like the distance between when we make the deep copy and when we need to handle the repercussions of doing so.

berlin-ab:
I've pulled in the latest fixes for the fake client to test against my personal project. That how I ran against this current bug. After applying the fix in this PR, my local tests work as expected.

This PR is a proof of concept test and fix. Please rework it as you need to in order to incorporate it into the project.

PR #2488, opened 2023-09-10
✨Replace k8s.io/utils/pointer with k8s.io/utils/ptr

inteon:
Replaced k8s.io/utils/pointer with k8s.io/utils/ptr to fix the golint deprecation errors


inteon:
@vincepri I re-scoped this PR, it should be ready to be merged.

alvaroaleman (approved):

Issue #2487, opened 2023-09-10
The fake client should be able to `cl.Update(...)` an object after calling `cl.Status().Update(...)`


berlin-ab:
Here's a Draft PR that tests the behavior:

https://github.com/kubernetes-sigs/controller-runtime/pull/2486

This is using the main branch.

troy0820:
/kind support 

Wouldn't what you propose, make the object re-queue after the first update?  Meaning, it wouldn't even get to the status update.

If you want to update the status, then you could update that first and not requeue by writing a predicate for that so it won't requeue on status updates. I'm trying to understand this case where it wouldn't requeue as an event for update to allow you to get to the status update.

berlin-ab:
> If you want to update the status, then you could update that first and not requeue by writing a predicate for that so it won't requeue on status updates. I'm trying to understand this case where it wouldn't requeue as an event for update to allow you to get to the status update.

While this is a good suggestion, I think we should still have the fake client match the behavior of the real client for ResourceVersion during `cl.Status().Update()`.

troy0820:
The behavior from my understanding doesn't allow this.  If you update an object that changes the resource version of that object and trying to update it again (any client, status included) without getting the latest resource would result in a conflict.  Patching doesn't do this but with update it respects the resource version as part of its check to allow it to go through.  We have a check that errors if the resource version isn't the same to mimic this behavior of what the apiserver does. 

@alvaroaleman thoughts?

berlin-ab:
Using my example CR `api.MyDatabase` against Minikube and a real c-r client:
 
```
	t.Run("test multiple updates succeed", func(t *testing.T) {
		scheme := runtime.NewScheme()
		_ = client.AddMyCRDTo(scheme)
		
		ctx := context.Background()
		testClient, err := k8sclient.New(config.GetConfigOrDie(), k8sclient.Options{
			Scheme: scheme,
		})
		require.NoError(t, err)

		db := &api.MyDatabase{}
		db.SetName("my-name")
		db.SetNamespace("default")
		_ = testClient.Delete(ctx, db)

		db.Spec.Instances = ptr.To(1)
		err = testClient.Create(ctx, db)
		require.NoError(t, err)
		t.Logf("resource version: %v", db.ResourceVersion)

		db.Status.IsUp = ptr.To(true)
		err = testClient.Status().Update(ctx, db)
		require.NoError(t, err)
		t.Logf("resource version: %v", db.ResourceVersion)

[...]

PR #2489, opened 2023-09-10
:bug: Fix returning object after status update

alvaroaleman:
Prior to this, we overwrote the passed in pointer to `obj` on status update, which means that the caller can never see the updated object after the call. This changes makes us instead overwrite the value behind the pointer, so that it gets propagated back to the caller.

@berlin-ab I like this approach more, as it seems a bit simpler and keeps the changed contained to the place that causes the need for them. WDYT?

/cc @berlin-ab @troy0820 

Superseeds https://github.com/kubernetes-sigs/controller-runtime/pull/2486
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2487


troy0820:
@alvaroaleman this makes sense and also behaves as expected. 
Testing errors but otherwise,
/lgtm

berlin-ab:
Very nice. Thanks.

Again, thanks for maintaining this tool that I use heavily daily.

troy0820 (approved):

PR #2494, opened 2023-09-14
✨ SetErrorWatchHandler on SharedIndexInformer 

### LINKED ISSUE #2395: should support SetWatchErrorHandler
client-go support SetWatchErrorHandler
controller-runtime should support SetWatchErrorHandler, dev can do something when watch failed

troy0820:
- [x] Adds `WatchErrorHandler` to InformersOpts struct 
- [x] Adds DefaultWatchErrorHandler to Options struct 
- [x] Adds watchErrorHandler type to Informers struct 
    - [x] _defaults to cache.DefaultWatchErrorHandler if not passed through options_
- [x] `SetErrorWatchErrorHandler` on sharedIndexInformer that we create when creating a `Cache`

This would resolve #2395 

/hold

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -188,6 +188,12 @@ type Options struct {
 	// unless there is already one set in ByObject or DefaultNamespaces.
 	DefaultTransform toolscache.TransformFunc
 
+	// DefaultWatchErrorHandler will be used to the WatchErrorHandler which is called
+	// whenever ListAndWatch drops the connection with an error.
+	//
+	// After calling this handler, the informer will backoff and retry.
+	DefaultWatchErrorHandler toolscache.WatchErrorHandler
+
 	// DefaultUnsafeDisableDeepCopy is the default for UnsafeDisableDeepCopy
 	// for everything that doesn't specify this.
 	//
@@ -353,6 +359,7 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 					Field: config.FieldSelector,
 				},
 				Transform:             config.Transform,
+				WatchErrorHandler:     opts.DefaultWatchErrorHandler,
 				UnsafeDisableDeepCopy: pointer.BoolDeref(config.UnsafeDisableDeepCopy, false),
 				NewInformer:           opts.newInformer,
 			}),
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -49,13 +49,18 @@ type InformersOpts struct {
 	Selector              Selector
 	Transform             cache.TransformFunc
 	UnsafeDisableDeepCopy bool
+	WatchErrorHandler     cache.WatchErrorHandler
 }
 
 // NewInformers creates a new InformersMap that can create informers under the hood.
 func NewInformers(config *rest.Config, options *InformersOpts) *Informers {
 	newInformer := cache.NewSharedIndexInformer
+	watchErrorHandler := cache.DefaultWatchErrorHandler
 	if options.NewInformer != nil {
 		newInformer = *options.NewInformer
+	}
+	if options.WatchErrorHandler != nil {
+		watchErrorHandler = options.WatchErrorHandler
 	}
 	return &Informers{
 		config:     config,
@@ -76,6 +81,7 @@ func NewInformers(config *rest.Config, options *InformersOpts) *Informers {
 		transform:             options.Transform,
 		unsafeDisableDeepCopy: options.UnsafeDisableDeepCopy,
 		newInformer:           newInformer,
+		watchErrorHandler:     watchErrorHandler,
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/cache/internal/informers.go
```diff
@@ -49,14 +49,19 @@ type InformersOpts struct {
 	Selector              Selector
 	Transform             cache.TransformFunc
 	UnsafeDisableDeepCopy bool
+	WatchErrorHandler     cache.WatchErrorHandler
```
FillZpp, line 52: Question is how could users set this flied in InformersOpts?

troy0820: The user can set this the same way we allow them to pass in transform in the InformersOpts. They’d pass the function in and it’ll be set when the mapping function sets the `WatchErrorHandler` using that method after the SharedIndexInformer is created.

troy0820: Very good point.  I can put this in the `Options` struct and allow this to be overridden within the instantiation of the cache.   Being that this only has one, it doesn't need to be done `byObject` so I don't have to put it there.  The resultant handler will be the one passed in or the `toolscache.DefaultWatchErrorHandler`.  I will make the changes. 

```
// There's only one handler, so if you call this multiple times, last one
// wins; calling after the informer has been started returns an error.
```

pkg/cache/internal/informers.go
```diff
@@ -159,6 +161,8 @@ type Informers struct {
 
 	// NewInformer allows overriding of the shared index informer constructor for testing.
 	newInformer func(cache.ListerWatcher, runtime.Object, time.Duration, cache.Indexers) cache.SharedIndexInformer
+
+	watchErrorHandler cache.WatchErrorHandler
```
troy0820, line 165: I can add a comment here.

pkg/cache/cache.go
```diff
@@ -188,6 +188,11 @@ type Options struct {
 	// unless there is already one set in ByObject or DefaultNamespaces.
 	DefaultTransform toolscache.TransformFunc
 
+	// DefaultWatchErrorHandler will be used to the WatchErrorHandler which is called
+	// whenever ListAndWatch drops the connection with an error. After calling this handler,
```
troy0820, line 192: I can split that comment line here.

pkg/cache/cache.go
```diff
@@ -381,6 +388,9 @@ func defaultOpts(config *rest.Config, opts Options) (Options, error) {
 		opts.Scheme = scheme.Scheme
 	}
 
+	if opts.DefaultWatchErrorHandler == nil {
```
FillZpp, line 391: How about keep this field defaults to be nil, not to set the `toolscache.DefaultWatchErrorHandler`.

We can call `sharedIndexInformer.SetWatchErrorHandler()` only if it is not nil, otherwise let the internal informer/reflector use default handler.

troy0820: ~~So if it is nil, set this to the `toolscache.DefaultWatchErrorHandler`?~~

I will move this but only set it when this is not nil.  The default handler you are suggesting is set when invoking the creation of the sharedindexinformer.

alvaroaleman (commented):
Thanks for looking into this, this was in the back of my mind as something we should do for a long time

pkg/cache/cache.go
```diff
@@ -188,6 +188,12 @@ type Options struct {
 	// unless there is already one set in ByObject or DefaultNamespaces.
 	DefaultTransform toolscache.TransformFunc
 
+	// DefaultWatchErrorHandler will be used to the WatchErrorHandler which is called
+	// whenever ListAndWatch drops the connection with an error.
+	//
+	// After calling this handler, the informer will backoff and retry.
+	DefaultWatchErrorHandler toolscache.WatchErrorHandler
```
alvaroaleman, line 195: Two questions:
1. What does this default to inside the `SharedIndexInformer`, a logger?
2. The infomers relist once in a while. Does this get called on that part? The reason I am asking is that I am wondering if it would be ok to have an implementation here that exits the binary with an error or if that would break things.

troy0820: > What does this default to inside the SharedIndexInformer, a logger?

From my understanding it seems to just log even in the case where `utilruntime.HandleError` is the default case e.g io.EOF, etc

> The infomers relist once in a while. Does this get called on that part? The reason I am asking is that I am wondering if it would be ok to have an implementation here that exits the binary with an error or if that would break things.

From what I understand, the relist calls this method during the relist but because it exited normally, it wouldn't error but log to notify.  

It seems like we were already using this default behavior but not exposing a way to `SetErrorWatchHandler` for people to set on the SharedIndexInformer.  Initially I was setting this to `DefaultWatchErrorHandler` but because the behavior of the SharedIndexInformer is to use this if not set, feedback was presented to not set this as it was redundant.

alvaroaleman (approved):

PR #2495, opened 2023-09-14
🐛 Handle unstructured status update with fake client

### LINKED ISSUE #2493: Panic in fake client when updating status with Unstructured resource
There is a regression in 0.16.2 in the fake client when attempting to update a resource's status with an unstructured object.

In this case, the GVK has a type registered in the scheme, but the client is invoked using an unstructured object. The initial object was also defined as unstructured and [converted to the typed resource inside the tracker](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.2/pkg/client/fake/client.go#L292).

The panic is triggered at
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.2/pkg/client/fake/client.go#L407

- `obj` is the unstructured value passed to `client.Status().Update()`
- `oldObject` is a typed value managed by the tracker

The two types are incompatible and cannot be reflectively assigned, at least not naively.

```
panic: reflect.Set: value of type v1.MyResource is not assignable to type unstructured.Unstructured [recovered]
	panic: reflect.Set: value of type v1.MyResource is not assignable to type unstructured.Unstructured

goroutine 12 [running]:
testing.tRunner.func1.2({0x105d91e20, 0x140004cc010})
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/testing/testing.go:1545 +0x370
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/testing/testing.go:1548 +0x528
panic({0x105d91e20?, 0x140004cc010?})
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/panic.go:920 +0x254
reflect.Value.assignTo({0x105f00420, 0x14000100900, 0x199}, {0x105b0d7e6, 0xb}, 0x105e6d040, 0x0)
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/reflect/value.go:3307 +0x340
reflect.Value.Set({0x105e6d040, 0x14000683588, 0x199}, {0x105f00420, 0x14000100900, 0x199})
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/reflect/value.go:2260 +0xa8
sigs.k8s.io/controller-runtime/pkg/client/fake.versionedTracker.update({{0x1060581a0, 0x140000a0370}, 0x140002885b0, 0x1400007fe30}, {{0x1400069dde0, 0x1a}, {0x1400069ddfb, 0x2}, {0x140001ae220, 0xd}}, ...)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/client/fake/client.go:407 +0x938
sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeClient).update(0x140000a03c0, {0x10605dd18, 0x14000683588}, 0x1, {0x14000356dc0, 0x1, 0x1})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/client/fake/client.go:789 +0x46c
sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Update(0x14000516f18, {0x106054c50, 0x14000024c30}, {0x10605dd18, 0x14000683588}, {0x0, 0x0, 0x0})
```

scothis:
In order to prevent unintentional mutations outside of the status during a status update, the non-status fields are copied back onto the passed object.

This operation now gracefully handles both unstructured and typed objects. Previously, it would panic if an unstructured object was passed for a GVK known to the scheme, as internally the object within the tracker is converted to the typed equivalent. The two types cannot be directly assigned to each other and instead must be copied.

Adds a failing test to demonstrate the bug, and the fix to make the test pass.

Resolves #2493

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -404,7 +404,9 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
 				return fmt.Errorf("failed to copy non-status field for object with status subresouce: %w", err)
 			}
 			passedRV := accessor.GetResourceVersion()
-			reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(oldObject.DeepCopyObject()).Elem())
+			if err := copyFrom(oldObject, obj); err != nil {
+				return fmt.Errorf("failed to restore non-status fields: %w", err)
+			}
 			accessor.SetResourceVersion(passedRV)
 		} else { // copy status from original object
 			if err := copyStatusFrom(oldObject, obj); err != nil {
@@ -966,6 +968,19 @@ func copyStatusFrom(old, new runtime.Object) error {
 	newMapStringAny["status"] = oldMapStringAny["status"]
 
 	if err := fromMapStringAny(newMapStringAny, new); err != nil {
+		return fmt.Errorf("failed to convert back from map[string]any: %w", err)
+	}
+
+	return nil
+}
+
+// copyFrom copies from old into new
+func copyFrom(old, new runtime.Object) error {
+	oldMapStringAny, err := toMapStringAny(old)
+	if err != nil {
+		return fmt.Errorf("failed to convert old to *unstructured.Unstructured: %w", err)
+	}
+	if err := fromMapStringAny(oldMapStringAny, new); err != nil {
 		return fmt.Errorf("failed to convert back from map[string]any: %w", err)
 	}
 
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -1677,6 +1677,52 @@ var _ = Describe("Fake client", func() {
 		Expect(obj.Object["spec"]).To(BeEquivalentTo("original"))
 	})
 
+	It("should not change the status of known unstructured objects that have a status subresource on update", func() {
+		obj := &unstructured.Unstructured{}
+		obj.SetAPIVersion("v1")
+		obj.SetKind("Pod")
+		obj.SetName("pod")
+		err := unstructured.SetNestedField(obj.Object, "Always", "spec", "restartPolicy")
+		Expect(err).NotTo(HaveOccurred())
+		err = unstructured.SetNestedField(obj.Object, "Pending", "status", "phase")
+		Expect(err).NotTo(HaveOccurred())
+		cl := NewClientBuilder().WithStatusSubresource(obj).WithObjects(obj).Build()
+
+		err = unstructured.SetNestedField(obj.Object, "Never", "spec", "restartPolicy")
+		Expect(err).NotTo(HaveOccurred())
+		err = unstructured.SetNestedField(obj.Object, "Running", "status", "phase")
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(cl.Update(context.Background(), obj)).To(Succeed())
+		Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(obj), obj)).To(Succeed())
+
+		Expect(obj.Object["spec"].(map[string]any)["restartPolicy"]).To(Equal("Never"))
+		Expect(obj.Object["status"].(map[string]any)["phase"]).To(Equal("Pending"))
+	})
+
+	It("should not change non-status field of known unstructured objects that have a status subresource on status update", func() {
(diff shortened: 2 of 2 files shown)


scothis:
/retest

looks like a test flake

troy0820:
/retest 

Did this not get retriggered? Looks like a golangci-lint error. Interesting.

scothis:
> Did this not get retriggered?

It did. the first run's flake was a timeout in the leader election code.

troy0820:
/assign @alvaroaleman 

Assigning @alvaroaleman  as he made the initial change.  I'd look to him to give the lgtm. But this looks good to me. 

Thanks for the contribution @scothis

pkg/client/fake/client.go
```diff
@@ -972,6 +974,19 @@ func copyStatusFrom(old, new runtime.Object) error {
 	return nil
 }
 
+// copyFrom copies from old into new
+func copyFrom(old, new runtime.Object) error {
+	oldMapStringAny, err := toMapStringAny(old)
```
alvaroaleman, line 979: The original reason for not doing this anymore is that it might leave stale data in the target. I am confused as to why this isn't a problem anymore. Maybe we just didn't add sufficient test coverage?

scothis: This isn't changing existing behavior. No tests were modified or removed. What it does is handle a mix of  types and unstructured objects gracefully, instead of panicing.

What stale data is left behind?

troy0820: What @alvaroaleman is referring to I believe is that before this function was added, we removed a function similar to this where we were doing a `copyFrom` because of the issues with that function leaving stale data around. 
https://github.com/kubernetes-sigs/controller-runtime/pull/2479

https://github.com/kubernetes-sigs/controller-runtime/pull/2479#discussion_r1320627434

If this being added doesn't impact what is already there (covers previous tests and your new test case) and handles the new case to fix the regression, the new test coverage should be sufficient to have this fix be in and resolve the issue with unstructured stuff.  

I believe we have unstructured.Unstructured tests cases and if not being caught in testing, it begs the question how did those tests pass, if this problem still presented itself. 

_IANAM, (I am not a maintainer)_

scothis: the previous tests and code handled both `typed->typed` and `unstructured->unstructured`, but triggered a panic with `typed->unstructured`. The mix can happen when the GVK is registered in the scheme, so the resource in the tracker is always converted to the typed form, and then the user performs a status update using an unstructured object.

It's a bit esoteric as most users will favor a typed resource when available, but we hit it.

alvaroaleman: Ah, we also added code to [zero the target](https://github.com/kubernetes-sigs/controller-runtime/blob/6860cfb3d210ac028dc5d744e7824a99c8841f07/pkg/client/fake/client.go#L1015), that is why it works. Thanks for the patch and sorry for the review delay

pkg/client/fake/client_test.go
```diff
@@ -1677,6 +1677,52 @@ var _ = Describe("Fake client", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+		cl := NewClientBuilder().WithStatusSubresource(obj).WithObjects(obj).Build()
+
+		err = unstructured.SetNestedField(obj.Object, "Never", "spec", "restartPolicy")
+		Expect(err).NotTo(HaveOccurred())
+		err = unstructured.SetNestedField(obj.Object, "Running", "status", "phase")
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(cl.Update(context.Background(), obj)).To(Succeed())
+		Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(obj), obj)).To(Succeed())
```
alvaroaleman, line 1697: Would you actually mind to do the retrieval and verification using `corev1.Pod` instead of `unstructured.Unstructured`? That I believe would simplify the code a bit but more importantly clarify the intention of the testcase. I had to read the title three times to understand it and that is with the context from the PR. Same for the other testcase.

scothis: Updated the tests. Sorry for the delay, I was traveling last week.

alvaroaleman (approved):
thanks!

PR #2498, opened 2023-09-18
:book: Add more examples in documentation

aerfio:
- Adds an example showing how to use [Server Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) with controller-runtime client and structs from _k8s.io/client-go/applyconfigurations/*_
- Adds an example how to watch any resource and map its changes to main reconciled resource

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -24,8 +24,14 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	// since we invoke tests with -ginkgo.junit-report we need to import ginkgo.
 	_ "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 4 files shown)


example_test.go
```diff
@@ -62,6 +68,80 @@ func Example() {
...
+
+// DeepCopyObject implements client.ObjectList.
+func (*ExampleCRDWithConfigMapRefList) DeepCopyObject() runtime.Object {
+	panic("unimplemented")
+}
+
+// This example creates a simple application Controller that is configured for ExampleCRDWithConfigMapRef CRD.
+// Any change in the configMap referenced in this Custom Resource will cause the re-reconcile of the parent ExampleCRDWithConfigMapRef
+// due to the implementation of the .Watches method of "sigs.k8s.io/controller-runtime/pkg/builder".Builder.
+func Example_watches() {
```
alvaroaleman, line 96: I don't think the fact that this has a watch is the special bit about this example. And I'd rather not have an example that doesn't implement `DeepCopy` but just panics. Maybe do something like a special configmap key that references something else or such?
```suggestion
func ExampleCustomHandler() {
```

pkg/client/example_test.go
```diff
@@ -188,6 +191,16 @@ func ExampleClient_patch() {
 	}, client.RawPatch(types.StrategicMergePatchType, patch))
 }
 
+// This example shows how to use the client with unstructured objects to create/patch objects using Server Side Apply,
```
alvaroaleman, line 194: Please mention why the dance of converting to unstructured is needed

alvaroaleman (commented):
/ok-to-test

example_test.go
aerfio: > I don't think the fact that this has a watch is the special bit about this example. 

That was my intention, to have an easy to grasp example for "Watches" 😢 

> Maybe do something like a special configmap key that references something else or such?

I didn't want to complicate this example too much, I thought about switching from LocalObjectReference to e.g [ConfigMapKeySelector](https://github.com/kubernetes/api/blob/v0.28.2/core/v1/types.go#L2137), but the code inside Watches would a too complicated for an example imho.

> And I'd rather not have an example that doesn't implement DeepCopy but just panics. 

Sure, I've changed it to json.Marshal + json.Unmarshal with panic on error, is that enough? I could use `controller-gen` for that, but that would result in new file in the root of the repo + some changes in makefile, which I wanted to omit

aerfio:
@alvaroaleman done

alvaroaleman (approved):

Issue #1215, opened 2020-10-12
Enable a user-friendly means of the controller-manager self-identifying with a distinct ID with HTTP calls


akutz:
# Summary

This story tracks enabling a user-friendly option for telling controller-runtime to self-identify with a distinct ID when making HTTP calls to the Kubernetes API server.

# Context

The actor that causes a raw Kubernetes event to occur is identified by a field named `manager`, for example:

```json
{
    "type": "MODIFIED",
    "object": {
        "apiVersion": "vmoperator.vmware.com/v1alpha1",
        "kind": "VirtualMachine",
        "metadata": {
            "annotations": {
                "vsphere-cluster-module-group": "tkg7-cluster7-workers-0",
                "vsphere-tag": "WorkerVmVmAATag"
            },
            "creationTimestamp": "2020-10-08T11:41:50Z",
            "finalizers": [
                "virtualmachine.vmoperator.vmware.com"
            ],
            "generation": 19,
            "labels": {
                "capw.vmware.com/cluster.name": "tkg7-cluster7",
                "capw.vmware.com/cluster.role": "node"
            },
            "managedFields": [
                {
                    "apiVersion": "vmoperator.vmware.com/v1alpha1",
                    "fieldsType": "FieldsV1",
                    "fieldsV1": { ... }
                    "manager": "manager",
                    "operation": "Update",
                    "time": "2020-10-08T14:11:49Z"
                }
            ],
            "name": "tkg7-cluster7-workers-whph9-6ff8cdb6fb-fq2b6",
            "namespace": "tkg7",
            ...
        },
        ...
}
```

These raw, Kubernetes events are useful for root cause analysis (RCA). The field `object.metadata.managedFields[0].manager` has a value of simply `manager`. This is the identifier for the actor that caused this `MODIFIED` event to occur, but `manager` is not exactly helpful in identifying the component in question.

The reason, if the reader will permit this author to make a broad guess, is that the majority of projects based on controller-runtime and [client-go](https://github.com/kubernetes/client-go) have adopted the pattern of naming the manager binary simply manager for a few reasons:

* It simplifies Makefiles
* It simplifies CI
* It enables a standard pattern for managing/working on these projects 

So, good reasons. Unfortunately it has the side effect that all the components send the same user agent to the API server when updating/patching resources. This happens because:

[...]

PR #2506, opened 2023-09-22
✨ Add UserAgentID option to the manager

m-messiah:
The PR is intended to solve issue #1215, allowing passing the new `Option` to the controller-manager constructor to override the first part of the UserAgent (as it was suggested in https://github.com/kubernetes-sigs/controller-runtime/issues/1215#issuecomment-707181082)

The solution is pretty safe as it does not change the `UserAgent` if it was already set in the `rest.Config` provided, but override it only when `rest.DefaultKubernetesUserAgent()` is used (which is always `manager/<version> ...`)

Manipulation of UserAgent broke the test, initially created as part of https://github.com/kubernetes-sigs/controller-runtime/pull/1401, but the test does not make sense anymore as:
1. LeaderElectionConfig is created as the copy of `rest.Config`
2. We manipulate UserAgent always if it is empty, it was done after the original config was assigned to the `cluster.New()`.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"net/http"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -119,6 +120,11 @@ type Options struct {
 	// NOTE: LOW LEVEL PRIMITIVE!
 	// Only use a custom NewCache if you know what you are doing.
 	NewCache cache.NewCacheFunc
+
+	// UserAgentID is the name of the manager to represent itself in the client requests.
+	//
+	// If set, the UserAgentID will be used to override the first part of rest.DefaultKubernetesUserAgent()
+	UserAgentID string
 
 	// Client is the client.Options that will be used to create the default Client.
 	// By default, the client will use the cache for reads and direct calls for writes.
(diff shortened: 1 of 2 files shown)


pkg/manager/manager.go
```diff
@@ -120,6 +121,11 @@ type Options struct {
 	// Only use a custom NewCache if you know what you are doing.
 	NewCache cache.NewCacheFunc
 
+	// UserAgentID is the name of the manager to represent itself in the client requests.
+	//
+	// If set, the UserAgentID will be used to override the first part of rest.DefaultKubernetesUserAgent()
+	UserAgentID string
```
alvaroaleman, line 127: What is the advantage of this over setting it in the rest config?

alvaroaleman: > It re-uses DefaultKubernetesUserAgent, having the same structure and versioning as it was before, but with fixed os.Args[0] part.

I don't know what that means because I don't know how the default looks like. Why does it matter to re-use the same structure and versioning rather than setting the entire user-agent?

m-messiah: Currently, DefaultKubernetesUserAgent is the useragent like: `manager/v1.0.1 (linux/amd64) kubernetes/12312afdb`, where we have os.Args[0] + version + runtime.Os/arch + git hash. 

So in the way of keep version and runtime the solution is to cut off the first part and replace it with correct name, rather than requiring operator to re-construct the whole UserAgent string with version and runtime lookups.

The better way to achieve it is to have a name param to client-go DefaultKubernetesUserAgent (or have the similar method) and set the UserAgent explicitely, but it is out of scope of controller-runtime.. But yes, I think I'd better do a PR to client-go, as it makes more sense there

Issue #2503, opened 2023-09-19
FakeClient patch on any subresource other than status is not working


acumino:
FakeClient patch on any subresource other than status is not working.

For identifying if the last patch call was to update the status sub resource [here](https://github.com/kubernetes-sigs/controller-runtime/blob/cda3b29e29b8759e6c4507329082777257b36c0b/pkg/client/fake/client.go#L361-L367) we check if the stack trace has `sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch`,  and if true only allow to update status. 
This is wrong as  `sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch` will be present in the stack trace if a patch is done on any subresource. 

For ex, Patching `scale` subresource in deployment will also have `sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch` in the stack trace because of which the code above will identify it as a status subresource update call and will not allow spec(replica) change in deployment.

Issues introduced with https://github.com/kubernetes-sigs/controller-runtime/pull/2259
controller-runtime version - v0.16.2

alvaroaleman:
Feel free to take a stab at it

PR #2508, opened 2023-09-22
 🐛 Correctly identify if patch call was made on status

acumino:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2503

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -361,7 +361,7 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	isStatus := false
 	// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
 	// that reaction, we use the callstack to figure out if this originated from the status client.
-	if bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch")) {
+	if bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).statusPatch")) {
 		isStatus = true
 	}
 	return t.update(gvr, obj, ns, isStatus, false)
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -361,7 +361,7 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 	isStatus := false
 	// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
 	// that reaction, we use the callstack to figure out if this originated from the status client.
-	if bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch")) {
+	if bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).statusPatch")) {
```
alvaroaleman, line 364: Please provide a testcase that starts passing with this change

acumino: There are multiple tests for the status subresource, which passes after the changes made in this PR also, I added a test to verify that subresource other than status can also be patched. PTAL

pkg/client/fake/client_test.go
```diff
@@ -1590,6 +1591,28 @@ var _ = Describe("Fake client", func() {
 		Expect(objOriginal.Status.Phase).ToNot(Equal(actual.Status.Phase))
 	})
 
+	It("should be able to change typed objects that have a scale subresource on patch", func() {
```
alvaroaleman, line 1594: I have to say that I am amazed this works at all. It ends up patching the original object, not some subresource which happens to work if the original object has a `.spec.replicas` field, which doesn't have to be the case. I'll accept this patch, but I will also note that the `scale` subresource isn't really supported in the fake client and just coincidentally works under some specific coincidences

alvaroaleman (approved):

pkg/client/fake/client_test.go
acumino: > It ends up patching the original object, not some subresource which happens to work if the original object has a .spec.replicas field, which doesn't have to be the case. I'll accept this patch, but I will also note that the scale subresource isn't really supported in the fake client and just coincidentally works under some specific coincidences

This is true for any subresource other than status.

acumino:
@alvaroaleman Can we cherry-pick this PR in v0.16?

PR #2509, opened 2023-09-23
🌱 Add function RemoveControllerReference and HasControllerReference 

troy0820:
- [x] Add function `RemoveControllerReference` to controllerutil
- [x] Add function `HasControllerReference` to controllerutil 
- [x] Add tests for functions 

When creating https://github.com/kubernetes-sigs/controller-runtime/pull/2462 it was discussed that to achieve the ask of the issue #2381 two functions can help remove ControllerReferences to get around the errors of it not being able to be overwritten. 

These two functions  _(and not to be confused with the original PR that allows removal of any owner reference)_, allows the caller to see if the owner reference they want to remove in fact is a "ControllerReference" _(controller = true)_

This functionality is present with finalizers and will be helpful when trying to mitigate issues with trying to overwrite the owner reference on an object.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -152,6 +152,28 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
 	return nil
 }
 
+// HasControllerReference returns true if the object
+// has an owner ref with controller equal to true
+func HasControllerReference(object metav1.Object) bool {
+	owners := object.GetOwnerReferences()
+	for _, owner := range owners {
+		isTrue := owner.Controller
+		if owner.Controller != nil && *isTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// RemoveControllerReference removes an owner reference where the controller
+// equals true
+func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
thanks!

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -152,6 +152,28 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
...
+		if owner.Controller != nil && *isTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// RemoveControllerReference removes an owner reference where the controller
+// equals true
+func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
```
sbueringer, line 170: @troy0820 Sorry for the late review :).

I'm not sure if this func works as intended. It first checks if there is a controller ownerRef on the object and then it removes the passed in owner. It doesn't check that the passed in owner is actually the controller (which I think is the goal of this func)

troy0820: Now looking at it, I’m not sure how I missed that. I’ll update the func to look if the owner is the controlled and remove that reference, otherwise it misses the goal as you said. 

I’ll create an issue and assign it to me.

PR #2518, opened 2023-09-27
✨ Add NewCache function to cache config

timuthy:
Similar to [NewCache](https://github.com/kubernetes-sigs/controller-runtime/blob/c93e2abcb28eb71fccad7dc565f0547cc07e5566/pkg/cluster/cluster.go#L118) in the `Cluster` options, this PR allows to configure a custom cache creation function per object GVK. Using a custom cache is ideally not only restricted to a global level, and using the lately introduced `ByObject` seems to be a perfect fit.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -251,6 +251,8 @@ type ByObject struct {
 	UnsafeDisableDeepCopy *bool
 
 	// NewCache allows to use a custom cache creation function for the object.
+	// The Options object passed to the NewCache function is pruned to only contain
+	// the object's relevant ByObject configuration.
 	//
 	// NOTE: LOW LEVEL PRIMITIVE!
 	// Only use a custom NewCache if you know what you are doing.
@@ -322,7 +324,8 @@ func New(cfg *rest.Config, opts Options) (Cache, error) {
 
 		switch {
 		case config.NewCache != nil:
-			cache, err = config.NewCache(cfg, opts)
+			object := obj
+			cache, err = config.NewCache(cfg, byObjectOptions(object, opts, config))
 			if err != nil {
 				return nil, err
 			}
@@ -353,6 +356,13 @@ func byObjectToConfig(byObject ByObject) Config {
 		Transform:             byObject.Transform,
 		UnsafeDisableDeepCopy: byObject.UnsafeDisableDeepCopy,
 	}
+}
+
+func byObjectOptions(obj client.Object, opts Options, byObject ByObject) Options {
+	opts.ByObject = map[client.Object]ByObject{
+		obj: byObject,
+	}
+	return opts
 }
 
 type newCacheFunc func(config Config, namespace string) Cache
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -249,6 +249,12 @@ type ByObject struct {
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopy *bool
+
+	// NewCache allows to use a custom cache creation function for the object.
+	//
+	// NOTE: LOW LEVEL PRIMITIVE!
+	// Only use a custom NewCache if you know what you are doing.
+	NewCache NewCacheFunc
 }
 
 // Config describes all potential options for a given watch.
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -313,9 +319,16 @@ func New(cfg *rest.Config, opts Options) (Cache, error) {
 			return nil, fmt.Errorf("failed to get GVK for type %T: %w", obj, err)
 		}
 		var cache Cache
-		if len(config.Namespaces) > 0 {
+
+		switch {
+		case config.NewCache != nil:
```
troy0820, line 324: Could you add a test case with this?

pkg/cache/cache.go
```diff
@@ -249,6 +249,12 @@ type ByObject struct {
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopy *bool
+
+	// NewCache allows to use a custom cache creation function for the object.
+	//
+	// NOTE: LOW LEVEL PRIMITIVE!
+	// Only use a custom NewCache if you know what you are doing.
+	NewCache NewCacheFunc
```
acumino, line 257: Few point:
- Setting the cache using `NewCache` means all comments about other values in ByObject will become irrelevant. Please adapt those doc strings. 
- If `NewCache` is there we should not allow setting other fields of `ByObject`.
- [Defaulting](https://github.com/kubernetes-sigs/controller-runtime/blob/c93e2abcb28eb71fccad7dc565f0547cc07e5566/pkg/cache/cache.go#L405-L442) should not be done in case of `NewCache` is set for `ByObject`.

timuthy: I wonder about all of the above points. Shouldn't we rather create a copy of `Options` and remove all `ByObject` items that are not relevant for this object type? We don't now what information `NewCache` uses.

timuthy: Please see [cc3b976](https://github.com/kubernetes-sigs/controller-runtime/pull/2518/commits/cc3b9767fa293af93f6349779b7cfac20a878477) and tell me what you think (cc @troy0820) 🙂

pkg/cache/cache.go
```diff
@@ -249,6 +249,14 @@ type ByObject struct {
...
 	// otherwise you will mutate the object in the cache.
 	UnsafeDisableDeepCopy *bool
+
+	// NewCache allows to use a custom cache creation function for the object.
+	// The Options object passed to the NewCache function is pruned to only contain
+	// the object's relevant ByObject configuration.
+	//
+	// NOTE: LOW LEVEL PRIMITIVE!
+	// Only use a custom NewCache if you know what you are doing.
+	NewCache NewCacheFunc
```
alvaroaleman, line 259: Sorry, but no. The `NewCache` func gets the cache.Options already. You can then inside your custom `NewCache` func decide what to do on a per-object basis.

/hold

timuthy: You have a point here. 
It would be nice to combine both, the controller-runtime's cache implementation (to be used as a default cache) and a custom implementation on a per object basis (via `delegatingByGVKCache`). 
Is there any acceptable way to achieve this?

timuthy: Would you be open to have any custom cache function on an per object basis at all?
It's definitely a use-case for us and we'd be happy to reuse the `delegatingByGVKCache` implementation. I guess making this implementation public isn't an option either?!

Please let me know what you think.

timuthy: What we try to accomplish is to use a custom cache implementation for some kinds, while falling back to the general cache for others.

Specifically, we use a single-object cache implementation due to security restrictions. It's similar to `kubelet` which also uses such a cache ([ref](https://github.com/kubernetes/kubernetes/blob/v1.26.0/pkg/kubelet/util/manager/watch_based_manager.go)) to play well with the [NodeRestriction](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction) admission plugin.

vincepri: That should work well with what @alvaroaleman suggested? Seems like a layer on top, of this implementation, rather than something controller runtime should handle

alvaroaleman: > It's definitely a use-case for us and we'd be happy to reuse the delegatingByGVKCache implementati

Copy it? The issue I have with making it public or adding the option you suggest here is that sooner or later someone will come up with a use-case where neither of the two fits and they want to have another layer at which they want to configure things.

IMHO what we currently have covers 98% of use-cases and leaves the low-level escape hatch of using the `NewCacheFunc`. Our goal IMHO is not to make 100% of use-cases easy because the more esoteric ones will actually make the more standard one confusing but to make the majority of use-cases easy and all possible, which I think we do through exposing the `NewCache` func.

timuthy: Sure, it works with our implementation or if we copy the code. Still thought it was worth a shot to contribute our use-case back to the community. Thanks for your feedback 🙂

timuthy:
Closing this PR due to lack of interest (see https://github.com/kubernetes-sigs/controller-runtime/pull/2518#discussion_r1341777562)

PR #2519, opened 2023-09-27
✨ Add ListenerConfig for metrics server options

tomelliot16:
The reason for this change is having the ability to set a custom ListenConfig on the metric server listener.

The reason I need (my company needs) it is that it will allow us to set unix.SO_REUSEPORT flag which works around an edge case when a port is ever leaked and you are using the host network for your controller.

### DIFF
--- a/pkg/metrics/server/server.go
+++ b/pkg/metrics/server/server.go
@@ -103,7 +103,7 @@ type Options struct {
 	TLSOpts []func(*tls.Config)
 
 	// ListenConfig contains options for listening to an address on the metric server.
-	ListenConfig *net.ListenConfig
+	ListenConfig net.ListenConfig
 }
 
 // Filter is a func that is added around metrics and extra handlers on the metrics server.
@@ -252,11 +252,7 @@ func (s *defaultServer) Start(ctx context.Context) error {
 
 func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
 	if !s.options.SecureServing {
-		if s.options.ListenConfig == nil {
-			return net.Listen("tcp", s.options.BindAddress)
-		} else {
-			return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
-		}
+		return s.options.ListenConfig.Listen(ctx, "tcp", s.options.BindAddress)
 	}
 
 	cfg := &tls.Config{ //nolint:gosec
@@ -309,14 +305,7 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
 		cfg.Certificates = []tls.Certificate{keyPair}
 	}
 
-	var l net.Listener
-	var err error
-
-	if s.options.ListenConfig == nil {
-		l, err = net.Listen("tcp", s.options.BindAddress)
-	} else {
-		l, err = s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
-	}
+	l, err := s.options.ListenConfig.Listen(ctx, "tcp", s.options.BindAddress)
 
 	return tls.NewListener(l, cfg), err
 }
--- a/pkg/metrics/server/server.go
+++ b/pkg/metrics/server/server.go
@@ -310,7 +310,7 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
 		return nil, err
 	}
 
-	return tls.NewListener(l, cfg), err
+	return tls.NewListener(l, cfg), nil
 }
 
 func (s *defaultServer) GetBindAddr() string {
--- a/pkg/metrics/server/server.go
+++ b/pkg/metrics/server/server.go
@@ -103,8 +103,7 @@ type Options struct {
 	TLSOpts []func(*tls.Config)
 
 	// ListenConfig contains options for listening to an address on the metric server.
-	// Note: This does not work with SecureServing=true
-	ListenConfig net.ListenConfig
+	ListenConfig *net.ListenConfig
 }
 
 // Filter is a func that is added around metrics and extra handlers on the metrics server.
@@ -253,7 +252,11 @@ func (s *defaultServer) Start(ctx context.Context) error {
 
 func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
 	if !s.options.SecureServing {
-		return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
+		if s.options.ListenConfig == nil {
+			return net.Listen("tcp", s.options.BindAddress)
+		} else {
+			return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
+		}
 	}
 
 	cfg := &tls.Config{ //nolint:gosec
@@ -306,7 +309,16 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
 		cfg.Certificates = []tls.Certificate{keyPair}
 	}
 
-	return tls.Listen("tcp", s.options.BindAddress, cfg)
+	var l net.Listener
+	var err error
+
+	if s.options.ListenConfig == nil {
+		l, err = net.Listen("tcp", s.options.BindAddress)
+	} else {
+		l, err = s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
+	}
+
+	return tls.NewListener(l, cfg), err
 }
 
 func (s *defaultServer) GetBindAddr() string {
--- a/pkg/metrics/server/server.go
+++ b/pkg/metrics/server/server.go
@@ -306,6 +306,9 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
 	}
 
 	l, err := s.options.ListenConfig.Listen(ctx, "tcp", s.options.BindAddress)
+	if err != nil {
+		return nil, err
+	}
 
 	return tls.NewListener(l, cfg), err
 }


pkg/metrics/server/server.go
```diff
@@ -101,6 +101,10 @@ type Options struct {
 	// TLSOpts is used to allow configuring the TLS config used for the server.
 	// This also allows providing a certificate via GetCertificate.
 	TLSOpts []func(*tls.Config)
+
+	// ListenConfig contains options for listening to an address on the metric server.
+	// Note: This does not work with SecureServing=true
+	ListenConfig net.ListenConfig
```
troy0820, line 107: If this doesn't work for SecureServing=true do we want this to be an option?

pkg/metrics/server/server.go
```diff
@@ -249,7 +253,7 @@ func (s *defaultServer) Start(ctx context.Context) error {
 
 func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
 	if !s.options.SecureServing {
-		return net.Listen("tcp", s.options.BindAddress)
+		return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
```
troy0820, line 256: Should we use the context that is passed in?

troy0820: Also, this will always be the listener returned from the `options.ListenConfig`.  Did we not want to fall back on `net.Listen` if that isn't set?

tomelliot16: @troy0820 To both of your points 

net.Listen definition looks like
```
package net
...
func Listen(network, address string) (Listener, error) {
	var lc ListenConfig
	return lc.Listen(context.Background(), network, address)
}
```

1. I was not trying to change current behavior as you can see net package function Listen just uses Background Context so I assumed the behavior we would want to continue
2. This by design auto falls back I believe but could be wrong.

pkg/metrics/server/server.go
tomelliot16: yeah I was concerned about this. I debated adding more complicity in the function to fail if both were set but decided I would allow that to be decided here. The reason I chose not to just support secure as well is that in that code path for it harder to inject.
```
...
tls.Listen("tcp", s.options.BindAddress, cfg)
...
```
Is used instead of net.Listen
```
package tls
...
func Listen(network, laddr string, config *Config) (net.Listener, error) {
	if config == nil || len(config.Certificates) == 0 &&
		config.GetCertificate == nil && config.GetConfigForClient == nil {
		return nil, errors.New("tls: neither Certificates, GetCertificate, nor GetConfigForClient set in Config")
	}
	l, err := net.Listen(network, laddr)
	if err != nil {
		return nil, err
	}
	return NewListener(l, config), nil
}
```
which means we probably would need this to do some of the responsibilities as it looks like it just uses net.Listen within it with a lot of other domain specific things. I couldn't see an easy way to inject the ListenConfig but maybe someone does.

pkg/metrics/server/server.go
troy0820: What I mean is that if the option doesn't exist, there is not listener to do this with.  There is no default `net.Listen(...)` because we have to use the `options.ListenConfig` that isn't there if it is not passed through.  Before with the old implementation, we didn't use an option.  If you check that the option exists, you could utilize the pattern you wish.

pkg/metrics/server/server.go
troy0820: `tls.NewListener` you can pass a listener in (which can be a ListenConfig) right? 

Something like this: 

```go
l, err := s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
     if err != nil {
       return nil, err
   }
     return tls.NewListener(l, cfg), nil
```

I'm also not sure what this provides beyond the means you express. We may need a maintainer to discuss as well.  

However, I still believe we shouldn't break existing functionality so guarding across the `net.Listen` should happen if the option doesn't exist.

troy0820 (changes requested):
This should work if SecureServing=true with the feedback presented as well as guarding against the options if they do not exist.

pkg/metrics/server/server.go
tomelliot16: @troy0820 I'm a little confused are you concerned about scope of `var lc ListenConfig` being instantiated inside the built in net.Listen function to have true backwards compatibility? If so I can add an if statement.

tomelliot16: I was trying to not make s.options.ListenConfig a pointer and without it being a pointer its hard to know if its been set because it is always set. Are you suggesting that we make it a pointer?

tomelliot16: There are two ways we can solve it being an if statement.

tomelliot16: I chose option 2 because there are other config the consumer might want to change within ListenConfig

pkg/metrics/server/server.go
tomelliot16: Addressed

troy0820: > I'm also not sure what this provides beyond the means you express. We may need a maintainer to discuss as well.

Looking into this a little bit more, I know this behavior allows the sharing of the port with other "things" and can't confidently say this is something that controller-runtime is willing to support.  I am not a maintainer and not quite familiar with what the intended use case for this expanding beyond what is already there.  

I understand your use case but want to see what other implications this has with the metrics server and what fallout can exist with allowing the new configuration in.

troy0820: /cc @alvaroaleman

pkg/metrics/server/server.go
```diff
@@ -249,7 +252,11 @@ func (s *defaultServer) Start(ctx context.Context) error {
 
 func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
 	if !s.options.SecureServing {
-		return net.Listen("tcp", s.options.BindAddress)
+		if s.options.ListenConfig == nil {
+			return net.Listen("tcp", s.options.BindAddress)
+		} else {
+			return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
+		}
```
troy0820, line 259: nit: ```suggestion
		if s.options.ListenConfig != nil {
			return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
	}
			return net.Listen("tcp", s.options.BindAddress)
}
```

File: pkg/metrics/server/server.go [hunk 4]
```diff
@@ -249,7 +252,11 @@ func (s *defaultServer) Start(ctx context.Context) error {
 
 func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (net.Listener, error) {
 	if !s.options.SecureServing {
-		return net.Listen("tcp", s.options.BindAddress)
+		if s.options.ListenConfig == nil {
```
vincepri, line 255: This should be defaulted, and we should always call it; instead of using the if with net.Listen which is basically just an empty ListenConfig https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/net/dial.go;l=804

pkg/metrics/server/server.go
```diff
@@ -249,7 +252,11 @@ func (s *defaultServer) Start(ctx context.Context) error {
...
+			return net.Listen("tcp", s.options.BindAddress)
+		} else {
+			return s.options.ListenConfig.Listen(context.Background(), "tcp", s.options.BindAddress)
```
vincepri, line 258: ctx is available, don't use background

pkg/metrics/server/server.go
tomelliot16: That's what I had originally but @troy0820  had concerns.

pkg/metrics/server/server.go
tomelliot16: Yeah I was just trying not to change any behavior for the caller but if you think that is better I will change.

tomelliot16: https://github.com/kubernetes-sigs/controller-runtime/pull/2519#discussion_r1349817547

pkg/metrics/server/server.go
troy0820: My concern was that this default thing @vincepri is asking you to do would have been nil if not set.  So not passing it in the options would have resulted in a nil listenconfig.  

What @vincepri is asking you to do is if this is not set, is to make the options.ListenConfig != nil (creating a default listenconfig) so the listenconfig will be available if not passed through the options.

tomelliot16: options.ListenConfig cannot be nil is my understanding because its by value

tomelliot16: I ran this just to prove my theory
```
package main

import (
	"fmt"
	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
)

func main() {
	opt := server.Options{}
	fmt.Println(opt.ListenConfig.KeepAlive)
}
```

running that prints
```
0s
```

troy0820: You are correct.  This works because it's not a pointer, so it's nil value is a listenConfig. You never made it a pointer, and this will default to a correct listenConfig.  We can resolve this.

pkg/metrics/server/server.go
tomelliot16: I have addressed this it should support it for all cases now

tomelliot16:
@sbueringer Did I miss anything in the process or anything I can do to help?

tomelliot16:
Any update here?

pkg/metrics/server/server.go
```diff
@@ -302,7 +305,9 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
 		cfg.Certificates = []tls.Certificate{keyPair}
 	}
 
-	return tls.Listen("tcp", s.options.BindAddress, cfg)
+	l, err := s.options.ListenConfig.Listen(ctx, "tcp", s.options.BindAddress)
```
sbueringer, line 308: Should we return the error here if listen already fails?

Not sure if we should call tls.NewListener if Listen failed

sbueringer: @tomelliot16  ^^

troy0820: This was mentioned in this comment.  https://github.com/kubernetes-sigs/controller-runtime/pull/2519#discussion_r1340705060

/remove lgtm

sbueringer:
@tomelliot16 Do you have time to address the remaining finding?

(if it gets fixed quickly we can include it into v0.17)

tomelliot16:
@sbueringer just saw this going to address comments.

tomelliot16:
@sbueringer I returned the error but didn't know if we wanted to make a special error wrapping with some context.

pkg/metrics/server/server.go
```diff
@@ -302,7 +305,12 @@ func (s *defaultServer) createListener(ctx context.Context, log logr.Logger) (ne
...
+	if err != nil {
+		return nil, err
+	}
+
+	return tls.NewListener(l, cfg), err
```
sbueringer, line 313:
```suggestion
	return tls.NewListener(l, cfg), nil
```

sbueringer:
> @sbueringer I returned the error but didn't know if we wanted to make a special error wrapping with some context.

It's fine as is. One last nit

sbueringer:
Thx!

/approve

troy0820:
@sbueringer this was slated for v0.17, did we want to cherry pick to that?

sbueringer:
> @sbueringer this was slated for v0.17, did we want to cherry pick to that?

Good catch. No strong opinion from my side either way. I mostly flagged it for v0.17 because at some point I thought it will make it into the release and I wanted to track it as part of the milestone

PR #2528, opened 2023-10-08
:sparkles: Cache: Allow defining options that apply to all namespaces that themselves have no explicit config

alvaroaleman:
This change allows to define a cache selector config that applies to all namespaces that themselves do not have an explicit config. An example would be "Cache all namespaces without selector, except for namespace foo, there use labelSelector bar=baz".

More as a side effect than intentionally, this also makes it valid to use the empty string as a key in the `Namespaces` and `byObject.Namespaces` config of the cache. This is very useful to for example have a `namespace` CLI flag that might be empty.

This change is the last missing bit to finish the implementation of the [cache options design doc](./designs/cache_options.md).

Sidenote: If someone can come up with a less wordy title that describes what this does, that would be nice :)

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -22,8 +22,10 @@ import (
 	"net/http"
 	"time"
 
+	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


pkg/cache/cache.go
```diff
@@ -172,6 +178,11 @@ type Options struct {
 	// the namespaces in here will be watched and it will by used to default
 	// ByObject.Namespaces for all objects if that is nil.
 	//
+	// It is possible to have specific Config for just some namespaces
+	// but cache all namespaces by using the AllNamespaces const as the map key.
+	// This wil then include all namespaces that do not have a more specific
```
troy0820, line 183: nit
```suggestion
	// This will then include all namespaces that do not have a more specific
```

pkg/cache/cache.go
```diff
@@ -220,6 +231,11 @@ type ByObject struct {
 	// Settings in the map value that are unset will be defaulted.
 	// Use an empty value for the specific setting to prevent that.
 	//
+	// It is possible to have specific Config for just some namespaces
+	// but cache all namespaces by using the AllNamespaces const as the map key.
+	// This wil then include all namespaces that do not have a more specific
```
troy0820, line 236: nit
```suggestion
	// This will then include all namespaces that do not have a more specific
```

Issue #2522, opened 2023-10-03
Possible performance regression when upgrading to 0.15.2


timflannagan:
We recently bumped our controller runtime dependency to the 0.15.2 release. After running our internal performance suite against a build with the 0.15.2 c-r version, we noticed roughly a ~10-15% performance regression compared to previous builds that used the 0.15.0 c-r version. When gathering and analyzing pprof profiles from those affected builds, our controller was spending a lot more time in the MatchingLabels ApplyToList client method for that type compared to 0.15.0 builds.

Looking at the changelog diff between 0.15.0 and 0.15.2, we can make the assumption that the culprit for this regression at scale was https://github.com/kubernetes-sigs/controller-runtime/issues/2098 due to the changes made to the client package. For additional context, our controller extensively uses the client.MatchingLabels list option when making client list calls.

alvaroaleman:
The previous behavior was buggy. If you have ideas on how to speed up the non-buggy implementation, feel free to suggest them or raise a PR.

timflannagan:
@alvaroaleman Thanks for the quick reply - I'll see if I can shore up some cycles to look into this further. I'm guessing the issue is we're constantly re-shuffling the underlying set when adding new label selector requirements.

PR #2529, opened 2023-10-09
:running: Make client.MatchingLabels faster

alvaroaleman:
The 99% use-case of this is to set a selector, not to adjust an existing one. This change introduces a fastpath that does that with half the allocations and in a bit less than half the time.

The reason slowpath is slow is that for each label a requirement has to be constructed that is then appended to a slice, both of which cause allocations.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2522

@timflannagan can you confirm that this fixes your issue? It brings the performance for the case that the labelselector is only set through `MatchingLabels` back to what it originally was.

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -514,7 +514,8 @@ type MatchingLabels map[string]string
 func (m MatchingLabels) ApplyToList(opts *ListOptions) {
 	// TODO(directxman12): can we avoid reserializing this over and over?
 	if opts.LabelSelector == nil {
-		opts.LabelSelector = labels.NewSelector()
+		opts.LabelSelector = labels.SelectorFromValidatedSet(map[string]string(m))
+		return
 	}
 	// If there's already a selector, we need to AND the two together.
 	noValidSel := labels.SelectorFromValidatedSet(map[string]string(m))


alvaroaleman:
I also have to say that while yes, this got slower, I find it surprising this is noticeable relative to DeepCopying in the cache or doing an api request. Would you mind describing what exactly you do that this becomes noticeable?

timflannagan:
Awesome - I'll add a TODO to test this in our environment tomorrow/Tuesday. At a glance, I think this should help alleviate some of the hot paths we were seeing in our pprof profiles for our controller. I'd love to share them to help paint a better picture, but they're from a closed source product, and I don't want legal knocking down my door 😅

alvaroaleman:
@timflannagan right I figured that, but could you share some more details, so you seem to be doing a high number of List requests with client.MatchingLabels - What is the aproximate number of labels, do you list against the cache or the api, do you set any other listOptions?

vincepri (approved):

timflannagan:
@alvaroaleman Just ran this commit through a couple of our heavier environments and these changes look like it fixes the performance issues we were previously seeing with 0.15.1 tag. Thanks for jumping on this (and the quick turn around). Any chance this could get backported to the release-0.15 release branch?

PR #2549, opened 2023-10-17
🌱 Allow webhook server to receive environment overrides

vincepri:
We're currently not exposing any flags generically for the webhook server across all controller runtime managers; this addition allows to customize ports, host, and certificate directory on a per-process basis through environment flags. Useful for testing, and for manager of managers.


vincepri:
/hold

Actually, should we have the manager's option struct (some, not all) to be configurable optionally via a env vars?

PR #2557, opened 2023-10-23
:bug: Always discover all versions for an apiGroup in RestMapper

### LINKED ISSUE #2556: RestMapper behavior is context specific
I am using the RestMapper implemented in this library and found that the behavior of a function call for the same arguments changed based what api calls are done before.

Here is an example:

1. First create a KIND cluster of 1.21.x

```bash
kind create cluster --image=kindest/node:v1.21.14
```

2. This K8s version had both v1 and v1beta1 EndpointSlice api. Now, I want to show how the behavior of the `RESTMapper().RESTMappings()` api call changes based on the context.

**Case 1:**

```go
mappings, err := kc.RESTMapper().RESTMappings(schema.GroupKind{
	Group: discoveryv1.GroupName,
	Kind:  "EndpointSlice",
})
if err != nil {
	return err
}
fmt.Println(len(mappings))

var list discoveryv1beta1.EndpointSliceList
err = kc.List(context.TODO(), &list)
if err != nil {
	return err
}
scheme := kc.Scheme()

mappings2, err := kc.RESTMapper().RESTMappings(schema.GroupKind{
	Group: discoveryv1.GroupName,
	Kind:  "EndpointSlice",
})
if err != nil {
	return err
}
fmt.Println(len(mappings2))
```

Here, we always get back both v1 and v1beta1 EndpointSlice resources.

**Case 2:**

```go
// mappings, err := kc.RESTMapper().RESTMappings(schema.GroupKind{
// 	Group: discoveryv1.GroupName,
// 	Kind:  "EndpointSlice",
// })
// if err != nil {
// 	return err
// }
// fmt.Println(len(mappings))

var list discoveryv1beta1.EndpointSliceList
err = kc.List(context.TODO(), &list)
if err != nil {
	return err
}
scheme := kc.Scheme()

mappings2, err := kc.RESTMapper().RESTMappings(schema.GroupKind{
	Group: discoveryv1.GroupName,
	Kind:  "EndpointSlice",
})
if err != nil {
	return err
}
fmt.Println(len(mappings2))
```

Here <ins>_we only get back the v1beta1 EndpointSlice resource even though the cluster has both v1 and v1beta1 EndpointSlice api._</ins>

**Why?**
The source of the issue seems to be this block of code:
https://github.com/kubernetes-sigs/controller-runtime/blob/15d79283523517f52ee30b28d2ae4c1ea3f113da/pkg/client/apiutil/restmapper.go#L161-L172

In case 2, since `v1beta1` is passed as version during the first call to the RestMapper, it only discovers that version and caches it. So, later calls only returns only v1beta1 api version.

One way to make this consistent will be to always discover all api versions for a GK the first time.

This was a surprising behavior to me and I wonder if this is intentional or this is a bug.

tamalsaha:
Fixes #2556

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -64,7 +64,7 @@ type mapper struct {
 func (m *mapper) KindFor(resource schema.GroupVersionResource) (schema.GroupVersionKind, error) {
 	res, err := m.getMapper().KindFor(resource)
 	if meta.IsNoMatchError(err) {
-		if err := m.addKnownGroupAndReload(resource.Group, resource.Version); err != nil {
+		if err := m.addKnownGroupAndReload(resource.Group); err != nil {
 			return schema.GroupVersionKind{}, err
 		}
 		res, err = m.getMapper().KindFor(resource)
@@ -77,7 +77,7 @@ func (m *mapper) KindFor(resource schema.GroupVersionResource) (schema.GroupVers
 func (m *mapper) KindsFor(resource schema.GroupVersionResource) ([]schema.GroupVersionKind, error) {
 	res, err := m.getMapper().KindsFor(resource)
 	if meta.IsNoMatchError(err) {
-		if err := m.addKnownGroupAndReload(resource.Group, resource.Version); err != nil {
+		if err := m.addKnownGroupAndReload(resource.Group); err != nil {
 			return nil, err
 		}
 		res, err = m.getMapper().KindsFor(resource)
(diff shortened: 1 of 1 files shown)


vincepri:
/hold

You can provide your own RESTMapper https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L103-L107 — the default one is dynamic and lazy

tamalsaha:
I think the default one should still correctly reflect the cluster state (say both api versions are available).

vincepri:
@alvaroaleman has a different idea that might work, in general, the built-in one we've opted to be more efficient for the client's main use

alvaroaleman:
> I think the default one should still correctly reflect the cluster state (say both api versions are available).

The default is meant to be lazy for how c-r uses it. That means one apicall per type. What you are building here means an api call per version of type.

What we can do is change `RestMapper()` on the client to return an upstream restmapper rather than the built-in lazy one, as the latter is only guaranteed to work correctly in the context of how CR uses it (always call with exactly one of group, kind and version).

PR #2560, opened 2023-10-23
🌱 envtest: improve process cleanup

vincepri:
When envtest creates new processes, it wasn't setting the process group to the parent, when the main process is killed or it panics, all the other processes are orphaned, and need to be killed separately.

In addition, if for some reason we can't shutdown a process cleanly, we should try to SIGKILL before returning from Stop.

/assign @alvaroaleman

### DIFF
--- a/pkg/internal/testing/process/process.go
+++ b/pkg/internal/testing/process/process.go
@@ -155,6 +155,7 @@ func (ps *State) Start(stdout, stderr io.Writer) (err error) {
 	ps.Cmd = exec.Command(ps.Path, ps.Args...)
 	ps.Cmd.Stdout = stdout
 	ps.Cmd.Stderr = stderr
+	ps.Cmd.SysProcAttr = GetSysProcAttr()
 
 	ready := make(chan bool)
 	timedOut := time.After(ps.StartTimeout)
@@ -265,6 +266,9 @@ func (ps *State) Stop() error {
 	case <-ps.waitDone:
 		break
 	case <-timedOut:
+		if err := ps.Cmd.Process.Signal(syscall.SIGKILL); err != nil {
+			return fmt.Errorf("unable to kill process %s: %w", ps.Path, err)
+		}
 		return fmt.Errorf("timeout waiting for process %s to stop", path.Base(ps.Path))
 	}
 	ps.ready = false
--- a/pkg/internal/testing/process/process_test.go
+++ b/pkg/internal/testing/process/process_test.go
@@ -352,16 +352,7 @@ var _ = Describe("Init", func() {
 })
 
 var simpleBashScript = []string{
-	"-c",
-	`
-		i=0
-		while true
-		do
-			echo "loop $i" >&2
-			let 'i += 1'
-			sleep 0.2
-		done
-	`,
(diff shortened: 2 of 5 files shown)


pkg/internal/testing/process/process.go
```diff
@@ -155,6 +155,7 @@ func (ps *State) Start(stdout, stderr io.Writer) (err error) {
 	ps.Cmd = exec.Command(ps.Path, ps.Args...)
 	ps.Cmd.Stdout = stdout
 	ps.Cmd.Stderr = stderr
+	ps.Cmd.SysProcAttr = getSysProcAttr()
```
alvaroaleman, line 158: Lol interesting. I ran into this, too in a completely different context

alvaroaleman (approved):

inteon:
/lgtm
Interesting, I didn't know this fix.

pkg/internal/testing/process/process_test.go
```diff
@@ -352,16 +352,7 @@ var _ = Describe("Init", func() {
...
-	`
-		i=0
-		while true
-		do
-			echo "loop $i" >&2
-			let 'i += 1'
-			sleep 0.2
-		done
-	`,
+	"-c", "tail -f /dev/null",
```
vincepri, line 355: @alvaroaleman @inteon I noticed the bash shells were never getting cleaned up, it seems this hot loop was the cause, because it never catches that it should go away.

alvaroaleman: ...what is this used for?

pkg/internal/testing/process/process.go
```diff
@@ -265,6 +266,9 @@ func (ps *State) Stop() error {
 	case <-ps.waitDone:
 		break
 	case <-timedOut:
+		if err := ps.Cmd.Process.Signal(syscall.SIGKILL); err != nil {
```
ncdc, line 269: You sure you don't want this? https://github.com/kcp-dev/kcp/pull/2707/commits/1cfbe44a0e9d40c848e15d3ec504d4284262d147

ncdc: Or at least put the kill-negativepid-sigterm up above where we're sending SIGTERM?

pkg/internal/testing/process/process_test.go
vincepri: Block forever basically

vincepri: It's just for testing

pkg/internal/testing/process/process.go
vincepri: Hm, wouldn't that kill the entire process group? What if we're only stopping a single piece?

alvaroaleman: Lol all the funky stuff you need to actually operationalize things seems to go back to work by @felixge

ncdc: Um 😄 was writing quickly as I remembered doing this for kcp. I'll defer to you for what the needs are here, in terms of 1 pid vs. multiple.

vincepri:
@alvaroaleman @ncdc ptal

Issue #2424, opened 2023-07-26
`DynamicRESTMapper` does not return `No{Kind,Resource}MatchError` for non-present API groups


timebertt:
When calling `RESTMapping()` and friends for an API Group/GroupVersion that is not present in the system (e.g., CRDs not yet installed), a `RESTMapper` should return `No{Kind,Resource}MatchError`.

However, with v0.15.0, the new `DynamicRESTMapper` returns `ErrGroupDiscoveryFailed` instead, e.g.
```
failed to get API group resources: unable to retrieve the complete list of server APIs: some.group/v1: the server could not find the requested resource
```

A consequence of this behavior is that managers will crash if they register controllers for resources that will only be registered during runtime. This was supported previously with the old `DynamicRESTMapper` implementation before v0.15.0.

/kind bug

troy0820:
Is this issue still necessary?  I see that this PR https://github.com/kubernetes-sigs/controller-runtime/pull/2425 is closed but the above PR that was merged took changes from that PR and applied it.  
@timebertt

timebertt:
Yes!
/close

PR #2571, opened 2023-11-02
⚠️  RESTMapper: don't treat non-existing GroupVersions as errors

ary1992:
/kind bug

This PR fixes the new RESTMapper implementation to properly handle cases where an API Group / GroupVersion is not present in the system. Clients expect No{Kind,Resource}MatchErrors in this case.

The first commit increases the accuracy of the RESTMapper tests to verify the described expectations.
The second commit implements the actual fix.

The changes are from @timebertt PR https://github.com/kubernetes-sigs/controller-runtime/pull/2425. 
The PR #2425 is still needed as with non existing CRDs  we are getting `failed to find API group` error.

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"sync"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


ary1992:
cc @alvaroaleman

vincepri:
/hold

This is in direct contrast with hold

ary1992:
> This is in direct contrast #2424

WDYM ? We still need this fix right? refer https://github.com/kubernetes-sigs/controller-runtime/pull/2425#issuecomment-1725606042 . PR #2425 was closed saying it was fixed with #2472 but that is not the case. The issue is still there.

alvaroaleman:
Wasn't this issue already fixed by https://github.com/kubernetes-sigs/controller-runtime/pull/2472 ?

ary1992:
> Wasn't this issue already fixed by #2472 ?

I don't think so
https://github.com/kubernetes-sigs/controller-runtime/pull/2425#issuecomment-1725606042

akalenyu:
gentle ping :pray: @alvaroaleman

sbueringer:
Hm are we okay with the cache filling up with invalid groups after this change?

image

alvaroaleman:
> Hm are we okay with the cache filling up with invalid groups (that we didn't find) after this change?

I guess the alternate is to handle absence of an entry in the map? Not sure if it will matter too much in practise as there isn't too much data in a group and this will only happen when requesting an absent group.

Issue #2582, opened 2023-11-17
Add a higher-level "canonical" reconciler API


JamesOwenHall:
Controller-runtime strikes a good balance between simplifying the development of Kubernetes controllers and still being very flexible. The flexibility of the `reconcile.Reconciler` interface should be maintained but I believe there would be value in also providing a slightly higher-level abstraction to reduce the duplication of common controller patterns.

In my experience, controllers can end up looking like the following before even implementing any business logic.

```go
const ExampleFinalizer = "my-finalizer"

func (r *ExampleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	example := &examplev1.Example{}
	if err := r.Get(ctx, req.NamespacedName, example); err != nil {
		return ctrl.Result{}, fmt.Errorf("failed to get example: %w", err)
	}

	if !example.ObjectMeta.DeletionTimestamp.IsZero() {
		if controllerutil.ContainsFinalizer(example, ExampleFinalizer) {
			if err := r.finalize(); err != nil {
				return ctrl.Result{}, fmt.Errorf("failed to finalize example: %w", err)
			}

			controllerutil.RemoveFinalizer(example, ExampleFinalizer)
			if err := r.Update(ctx, example); err != nil {
				return ctrl.Result{}, fmt.Errorf("failed to update example: %w", err)
			}
		}
	
		return ctrl.Result{}, nil
	}

	if !controllerutil.ContainsFinalizer(example, ExampleFinalizer) {
		controllerutil.AddFinalizer(example, ExampleFinalizer)
		if err := r.Update(ctx, example); err != nil {
			return ctrl.Result{}, fmt.Errorf("failed to update example: %w", err)
		}

		return ctrl.Result{}, nil
	}

	original := example.DeepCopy()
	businessLogicErr := r.doBusinessLogic(example)

	if equality.Semantic.DeepEqual(original, example) {
		return ctrl.Result{}, nil
	}

	if err := r.Update(ctx, example); err != nil {
		return ctrl.Result{}, fmt.Errorf("failed to update example: %w", err)
	}

	if businessLogicErr != nil {
		return ctrl.Result{}, fmt.Errorf("failed to do business logic: %w", businessLogicErr)
	}

	return ctrl.Result{}, nil
}
```

I propose adding a higher-level interface like the following to complement the existing `reconcile.Reconciler` interface. This builds upon #2221 but extends beyond just the use of generics.

```go
package canonical

[...]

PR #2592, opened 2023-11-23
✨ Add reconcile.ObjectReconciler

### LINKED ISSUE #2221: Consider using generics for the reconcile.Reconciler interface
Most reconcilers reconcile some kubernetes object. Any such reconciler needs to start off with a check if the object still exists and do nothing if not. We could avoid this check by doing it in controller-runtime and adding a new interface`Reconcile[o ctrlclient.Object](context.Context, o)(reconcile.Result, error)` that people can use instead of the current one. We can not remove the current one, because not all reconcilers act on k8s objects.

The main challenge here is that there are also reconcilers that do not reconcile a kubernetes object and that need to be able to use some kind of arbitrary key.

/cc @sbueringer @vincepri

JamesOwenHall:
Part of #2582.
Fixes #2221.

This PR adds a new `reconcile.ObjectReconciler` interface to eliminate the following common boilerplate.

```go
func (r *MyReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
	example := &examplev1.Example{}
	if err := r.Get(ctx, req.NamespacedName, example); err != nil {
		return ctrl.Result{}, client.IgnoreNotFound(err)
	}

	// Do business logic…
}
```

`ObjectReconciler` leverages generics to write reconcilers in terms of deserialized instances of `client.Object`. It does not replace the existing `reconcile.Reconciler` interface.

Developers can create a controller with an `ObjectReconciler` by calling the `AsReconciler` function.

```go
builder.ControllerManagedBy(manager).
	For(&corev1.ConfigMap{}).
	Complete(reconcile.AsReconciler(client, &MyObjectReconciler{}))
```

### DIFF
--- a/pkg/reconcile/reconcile.go
+++ b/pkg/reconcile/reconcile.go
@@ -19,9 +19,11 @@ package reconcile
 import (
 	"context"
 	"errors"
+	"reflect"
 	"time"
 
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Result contains the result of a Reconciler invocation.
@@ -109,6 +111,36 @@ var _ Reconciler = Func(nil)
 
 // Reconcile implements Reconciler.
 func (r Func) Reconcile(ctx context.Context, o Request) (Result, error) { return r(ctx, o) }
+
+// ObjectReconciler is a specialized version of Reconciler that acts on instances of client.Object. Each reconciliation
+// event gets the associated object from Kubernetes before passing it to Reconcile. An ObjectReconciler can be used in
+// Builder.Complete by calling AsReconciler. See Reconciler for more details.
+type ObjectReconciler[T client.Object] interface {
+	Reconcile(context.Context, T) (Result, error)
+}
+
+// AsReconciler creates a Reconciler based on the given ObjectReconciler.
+func AsReconciler[T client.Object](client client.Client, rec ObjectReconciler[T]) Reconciler {
+	return &objectReconcilerAdapter[T]{
+		objReconciler: rec,
+		client:        client,
+	}
+}
+
+type objectReconcilerAdapter[T client.Object] struct {
+	objReconciler ObjectReconciler[T]
+	client        client.Client
+}
+
+// Reconcile implements Reconciler.
+func (a *objectReconcilerAdapter[T]) Reconcile(ctx context.Context, req Request) (Result, error) {
+	o := reflect.New(reflect.TypeOf(*new(T)).Elem()).Interface().(T)
+	if err := a.client.Get(ctx, req.NamespacedName, o); err != nil {
+		return Result{}, client.IgnoreNotFound(err)
+	}
+
+	return a.objReconciler.Reconcile(ctx, o)
+}
 
 // TerminalError is an error that will not be retried but still be logged
 // and recorded in metrics.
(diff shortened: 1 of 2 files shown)


pkg/reconcile/reconcile.go
```diff
@@ -110,6 +112,36 @@ var _ Reconciler = Func(nil)
...
 
+// ObjectReconciler is a specialized version of Reconciler that acts on instances of client.Object. Each reconciliation
+// event gets the associated object from Kubernetes before passing it to Reconcile. An ObjectReconciler can be used in
+// Builder.Complete by calling AsReconciler. See Reconciler for more details.
+type ObjectReconciler[T client.Object] interface {
+	Reconcile(context.Context, T) (Result, error)
+}
+
+// AsReconciler creates a Reconciler based on the given ObjectReconciler.
+func AsReconciler[T client.Object](client client.Client, rec ObjectReconciler[T]) Reconciler {
```
JamesOwenHall, line 123: In https://github.com/kubernetes-sigs/controller-runtime/issues/2582#issuecomment-1819720649, @alvaroaleman proposed adding a new method to `builder.Builder` so that an adapter function like `AsReconciler` here wouldn't be necessary. However, Go does not allow a method to add new type parameters, only a method receiver can define type parameters. Therefore, I believe that adding a `Builder.CompleteObjectReconciler` method would only be possible if `Builder` defined a type parameter, i.e. `Builder[T client.Object]`. I don't think this type parameter makes much sense for instances of `reconcile.Reconciler` so I opted to use an adapter function instead.

alvaroaleman: I actually realized the same thing later on while being on a walk, so agree 👍

pkg/reconcile/reconcile_test.go
```diff
@@ -102,4 +114,75 @@ var _ = Describe("reconcile", func() {
...
+				}
+				key = client.ObjectKeyFromObject(cm)
+
+				err := testClient.Create(context.Background(), cm)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should Get the object and call the ObjectReconciler", func() {
+				var actual *corev1.ConfigMap
+				reconciler := reconcile.AsReconciler[*corev1.ConfigMap](testClient, &mockObjectReconciler{
```
alvaroaleman, line 154: I am very surprised this is needed - Could you try if upgrading to go 1.21 fixes that?

JamesOwenHall: This is indeed not needed in Go 1.21. I originally tested my PR locally against the Go 1.21 toolchain, but it appears that CI runs on Go 1.20. Are you suggesting upgrading the Go version in CI? If so, how do we do that?

sbueringer: Looks like the Go 1.21 bump was done in the meantime. So I think you can use 1.21 now

JamesOwenHall: Rebase fixed it, thanks for the heads up.

troy0820:
I'm not sure what else this needs but after reviewing it, it looks good to me. 

/lgtm

alvaroaleman:
So @vincepri @sbueringer and myself had a bit of discussion around this and weren't really sure if this is what we want. Given that this is a rather trivial change that can easily be managed downstream, we'd like to defer merging this for now.

We can probably use one of the issues to discuss this and see if we can up with something everyone is comfortable with.

vincepri (approved):

alvaroaleman:
/hold cancel
nvm, looks like ppl are actually fine with it. Thanks for your work :)

PR #2595, opened 2023-11-24
 🐛Remove owner passed in to RemoveControllerReference 

### LINKED ISSUE #2594: RemoveControllerReference in controllerutil bug.
In pull request #2509 when using the `RemoveControllerReference` the function checks if ann owner exists but only removes the owner passed in which may not be the owner reference with controlled = true. 

Update the function to ensure that the owner is in fact the controller reference to be removed. 

/kind bug 
/assign @troy0820

troy0820:
Fixes initial PR of #2509 

Resolves #2594 

- [x] Add a check to find owner reference that equals the passed in owner and remove if controller = true

/assign @sbueringer

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -171,7 +171,32 @@ func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Sche
 	if ok := HasControllerReference(object); !ok {
 		return fmt.Errorf("%T does not have a owner reference with controller equals true", object)
 	}
-	return RemoveOwnerReference(owner, object, scheme)
+	ro, ok := owner.(runtime.Object)
+	if !ok {
+		return fmt.Errorf("%T is not a runtime.Object, cannot call RemoveControllerReference", owner)
+	}
+	gvk, err := apiutil.GVKForObject(ro, scheme)
+	if err != nil {
+		return err
+	}
+	ownerRefs := object.GetOwnerReferences()
+	index := indexOwnerRef(ownerRefs, metav1.OwnerReference{
+		APIVersion: gvk.GroupVersion().String(),
+		Name:       owner.GetName(),
+		Kind:       gvk.Kind,
+	})
+
+	if index == -1 {
+		return fmt.Errorf("%T does not have an controller reference for %T", object, owner)
+	}
+
+	if ownerRefs[index].Controller == nil || !*ownerRefs[index].Controller {
+		return fmt.Errorf("%T owner is not the controller reference for %T", owner, object)
+	}
+
+	ownerRefs = append(ownerRefs[:index], ownerRefs[index+1:]...)
+	object.SetOwnerReferences(ownerRefs)
+	return nil
 }
 
 func upsertOwnerRef(ref metav1.OwnerReference, object metav1.Object) {
@@ -219,7 +244,6 @@ func referSameObject(a, b metav1.OwnerReference) bool {
 	if err != nil {
 		return false
 	}
-
 	return aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/controllerutil/controllerutil.go
```diff
@@ -171,7 +171,27 @@ func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Sche
...
+		Name:       owner.GetName(),
+		Kind:       gvk.Kind,
+		Controller: ptr.To(true),
+	})
+	if index == -1 {
+		return fmt.Errorf("%T does not have an controller reference for %T", object, owner)
+	}
+	ownerRefs = append(ownerRefs[:index], ownerRefs[index+1:]...)
+	object.SetOwnerReferences(ownerRefs)
+	return nil
```
troy0820, line 194: Not a fan of this replicated logic but didn't want to assert it to a `runtime.Object` back to a `metav1.Object` to utilize the reuse.  

This seems to make sure the only controllerReference is the owner that is passed in because 
1. Only one ControllerReference (1:1) be present on an object but many OwnerReferences (1:N)
2. The index returned verifies that the above owner is in fact the controller reference on the object if the index is greater than -1.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -171,7 +171,27 @@ func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Sche
 	if ok := HasControllerReference(object); !ok {
```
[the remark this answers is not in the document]
troy0820: I can do that.  My work here still doesn't show that the owner reference that exists is in fact the actual controller Reference that is being removed.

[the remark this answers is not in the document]
troy0820: The test I added was to show that this errors when passing in the wrong owner.  @sbueringer

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -171,7 +171,27 @@ func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Sche
...
+	gvk, err := apiutil.GVKForObject(ro, scheme)
+	if err != nil {
+		return err
+	}
+	ownerRefs := object.GetOwnerReferences()
+	index := indexOwnerRef(ownerRefs, metav1.OwnerReference{
+		APIVersion: gvk.GroupVersion().String(),
+		Name:       owner.GetName(),
+		Kind:       gvk.Kind,
+		Controller: ptr.To(true),
```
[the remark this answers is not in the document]
troy0820: I've updated the `referSameObject` func to check the `controller == *ptr.To(true)`.  

So now the indexOwnerRef will check to see if the controller is set/not set on the owner ref

pkg/controller/controllerutil/controllerutil_test.go
```diff
@@ -229,6 +230,20 @@ var _ = Describe("Controllerutil", func() {
...
+			rs := &appsv1.ReplicaSet{
+				ObjectMeta: metav1.ObjectMeta{},
+			}
+			dep := &extensionsv1beta1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "foo", UID: "foo-uid-2"},
+			}
+			dep2 := &extensionsv1beta1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "foo-2", UID: "foo-uid-42"},
+			}
+			Expect(controllerutil.SetControllerReference(dep, rs, scheme.Scheme)).ToNot(HaveOccurred())
```
sbueringer, line 243:
```suggestion
			Expect(controllerutil.SetControllerReference(dep, rs, scheme.Scheme)).ToNot(HaveOccurred())
			Expect(controllerutil.SetOwnerReference(dep2, rs, scheme.Scheme)).ToNot(HaveOccurred())
```

Wondering if that would add useful coverage. Basically we error out because the owner we want to remove is only owner but not the controller

troy0820: So it exists, but the owner we passed in would still error because it doesn't have the controller == true.  Makes sense.

inteon:
@troy0820 could you fix the typo in the PR title?

alvaroaleman (approved):

pkg/controller/controllerutil/controllerutil_test.go
sbueringer: Thx!

PR #2598, opened 2023-11-28
⚠️ Use limited reader in webhooks

inteon:
This PR adds a LimitedReader which prevents DOS attacks due to OOM caused by the `io.ReadAll` function.

Depends on https://github.com/kubernetes-sigs/controller-runtime/pull/2604.

Code is based on https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -34,6 +34,9 @@ import (
 var admissionScheme = runtime.NewScheme()
 var admissionCodecs = serializer.NewCodecFactory(admissionScheme)
 
+// based on https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
+var maxRequestSize = int64(3 * 1024 * 1024)
+
 func init() {
 	utilruntime.Must(v1.AddToScheme(admissionScheme))
 	utilruntime.Must(v1beta1.AddToScheme(admissionScheme))
@@ -55,9 +58,16 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	defer r.Body.Close()
-	body, err := io.ReadAll(r.Body)
+	limitedReader := &io.LimitedReader{R: r.Body, N: maxRequestSize}
+	body, err := io.ReadAll(limitedReader)
 	if err != nil {
 		wh.getLogger(nil).Error(err, "unable to read the body from the incoming request")
+		wh.writeResponse(w, Errored(http.StatusBadRequest, err))
+		return
+	}
+	if limitedReader.N <= 0 {
+		err := fmt.Errorf("request entity is too large; limit is %d bytes", maxRequestSize)
+		wh.getLogger(nil).Error(err, "unable to read the body from the incoming request; limit reached")
 		wh.writeResponse(w, Errored(http.StatusBadRequest, err))
 		return
 	}
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -34,8 +34,23 @@ import (
 var admissionScheme = runtime.NewScheme()
 var admissionCodecs = serializer.NewCodecFactory(admissionScheme)
 
-// based on https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
-var maxRequestSize = int64(3 * 1024 * 1024)
+// adapted from https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
+//
+// From https://github.com/kubernetes/apiserver/blob/d6876a0600de06fef75968c4641c64d7da499f25/pkg/server/config.go#L433-L442C5:
+//
+//	     1.5MB is the recommended client request size in byte
+//		 the etcd server should accept. See
+//		 https://github.com/etcd-io/etcd/blob/release-3.4/embed/config.go#L56.
+//		 A request body might be encoded in json, and is converted to
+//		 proto when persisted in etcd, so we allow 2x as the largest request
+//		 body size to be accepted and decoded in a write request.
+//
+// For the admission request, we can infer that it contains at most two objects
+// (the old and new versions of the object being admitted), each of which can
+// be at most 3MB in size. For the rest of the request, we can assume that
+// it will be less than 1MB in size. Therefore, we can set the max request
+// size to 7MB.
+var maxRequestSize = int64(7 * 1024 * 1024)
 
 func init() {
 	utilruntime.Must(v1.AddToScheme(admissionScheme))
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -50,6 +50,8 @@ var admissionCodecs = serializer.NewCodecFactory(admissionScheme)
 // be at most 3MB in size. For the rest of the request, we can assume that
 // it will be less than 1MB in size. Therefore, we can set the max request
 // size to 7MB.
+// If your use case requires larger max request sizes, please
+// open an issue (https://github.com/kubernetes-sigs/controller-runtime/issues/new).
 var maxRequestSize = int64(7 * 1024 * 1024)
 
 func init() {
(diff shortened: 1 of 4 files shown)


alvaroaleman:
@inteon would you mind splitting this up in one PR per change, so discussions can stay targeted and the changelog will have a more meaningful entry than "Improve Webhook"?

inteon:
> @inteon would you mind splitting this up in one PR per change, so discussions can stay targeted and the changelog will have a more meaningful entry than "Improve Webhook"?

Ok, this PR now focuses on adding `limited reader`.

pkg/webhook/admission/http.go
```diff
@@ -34,6 +34,9 @@ import (
 var admissionScheme = runtime.NewScheme()
 var admissionCodecs = serializer.NewCodecFactory(admissionScheme)
 
+// based on https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
+var maxRequestSize = int64(3 * 1024 * 1024)
```
alvaroaleman, line 38: This is technically a breaking change. More importantly, how do we know this size is appropriate? I guess for admission we can inferr that a request contains at most two objects which each have a maximum size of one meg plus the request itself which is also not likely to exceed one meg. But how about requests in the authentication webhook?

inteon: Upstream, this value seems to come from here: https://github.com/kubernetes/apiserver/blob/master/pkg/server/config.go#L433-L442C5

inteon: So, k/k has made a mistake using that value for this admission webhook?: https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go

inteon: @alvaroaleman I updated the maxRequestSize based on the reasoning you explained and added comments: https://github.com/kubernetes-sigs/controller-runtime/pull/2598/commits/9dcfc53b6dd7852c484034b763b015258dd382a6

pkg/webhook/admission/http.go
```diff
@@ -55,12 +73,19 @@ func (wh *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	body, err := io.ReadAll(limitedReader)
 	if err != nil {
 		wh.getLogger(nil).Error(err, "unable to read the body from the incoming request")
 		wh.writeResponse(w, Errored(http.StatusBadRequest, err))
 		return
 	}
+	if limitedReader.N <= 0 {
+		err := fmt.Errorf("request entity is too large; limit is %d bytes", maxRequestSize)
+		wh.getLogger(nil).Error(err, "unable to read the body from the incoming request; limit reached")
+		wh.writeResponse(w, Errored(http.StatusBadRequest, err))
```
sbueringer, line 86:
```suggestion
		wh.writeResponse(w, Errored(http.StatusRequestEntityTooLarge, err))
```

Seems more precise

pkg/webhook/admission/http.go
```diff
@@ -34,6 +34,24 @@ import (
 var admissionScheme = runtime.NewScheme()
 var admissionCodecs = serializer.NewCodecFactory(admissionScheme)
 
+// adapted from https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
+//
+// From https://github.com/kubernetes/apiserver/blob/d6876a0600de06fef75968c4641c64d7da499f25/pkg/server/config.go#L433-L442C5:
+//
+//	     1.5MB is the recommended client request size in byte
+//		 the etcd server should accept. See
```
sbueringer, line 42: Given all the stuff people are doing today. Can we really still assume that controller-runtime is only used in environments where etcd is used?

pkg/webhook/authentication/http.go
```diff
@@ -34,6 +34,20 @@ import (
...
+// adapted from https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go
+//
+// From https://github.com/kubernetes/apiserver/blob/d6876a0600de06fef75968c4641c64d7da499f25/pkg/server/config.go#L433-L442C5:
+//
+//	     1.5MB is the recommended client request size in byte
+//		 the etcd server should accept. See
+//		 https://github.com/etcd-io/etcd/blob/release-3.4/embed/config.go#L56.
+//		 A request body might be encoded in json, and is converted to
+//		 proto when persisted in etcd, so we allow 2x as the largest request
+//		 body size to be accepted and decoded in a write request.
```
sbueringer, line 46: Q: What does this have to do with the TokenReview request?

pkg/webhook/admission/http.go
vincepri: Probably not, but this limiter should only apply to admission webhooks; which seems fair?

pkg/webhook/authentication/http.go
inteon: You are right, the TokenReview request does not come from etcd.
Not sure what a better upper-limit would be.
I did read somewhere that JWTs have a upper size limit because they have to be includable in the header of a HTTP request, maybe that is a starting point?

pkg/webhook/admission/http.go
sbueringer: I assume the apiserver calls admission webhooks independent of what datastore is used behind it.

I wasn't thinking about non-apiserver scenarios. More about the cases where folks are just not using etcd as storage

sbueringer: Nevermind. I missed that every object that is stored goes through the apiserver which enforces the limit

pkg/webhook/authentication/http.go
sbueringer: After a quick look at google/stackoverflow it seems that neither JWT nor HTTP defines a max size for a JWT / a HTTP header. Just that there are limits that webservers have in place: https://stackoverflow.com/questions/686217/maximum-on-http-header-values

It seems usually HTTP header limits are between 4 and 48k. Maybe let's use 1 MB as a limit which should be low enough given what we're protecting against.

Then I would add a comment that the TokenReview mostly contains a bearer token which at most should have a few kb's of size, so we picked 1 MB to have plenty of buffer.

sbueringer: @alvaroaleman @vincepri ^^ any other ideas on how to pick a limit?

vincepri: The logic above makes sense to me, in both cases I'd add a comment that if this happens, folks should open an issue, and link to https://github.com/kubernetes-sigs/controller-runtime/issues/new

sbueringer:
@inteon PTAL when you have some time :)

inteon:
@sbueringer I updated the limits and comments based on the review feedback.
Please let me know in case I missed something.

sbueringer:
Thx!

/lgtm

pkg/webhook/admission/http.go
```diff
@@ -34,6 +34,26 @@ import (
...
+//		 body size to be accepted and decoded in a write request.
+//
+// For the admission request, we can infer that it contains at most two objects
+// (the old and new versions of the object being admitted), each of which can
+// be at most 3MB in size. For the rest of the request, we can assume that
+// it will be less than 1MB in size. Therefore, we can set the max request
+// size to 7MB.
+// If your use case requires larger max request sizes, please
+// open an issue (https://github.com/kubernetes-sigs/controller-runtime/issues/new).
+var maxRequestSize = int64(7 * 1024 * 1024)
```
alvaroaleman, line 55: `const`

pkg/webhook/authentication/http.go
```diff
@@ -34,6 +34,13 @@ import (
 var authenticationScheme = runtime.NewScheme()
 var authenticationCodecs = serializer.NewCodecFactory(authenticationScheme)
 
+// The TokenReview resource mostly contains a bearer token which
+// at most should have a few KB's of size, so we picked 1 MB to
+// have plenty of buffer.
+// If your use case requires larger max request sizes, please
+// open an issue (https://github.com/kubernetes-sigs/controller-runtime/issues/new).
+var maxRequestSize = int64(1 * 1024 * 1024)
```
alvaroaleman, line 42: const

alvaroaleman (approved):

PR #2601, opened 2023-12-02
✨ Add support for controller-manager webhook shutdown delay

dippynark:
Currently the controller-runtime signal handler cancels the context as soon as a termination signal is received: https://github.com/kubernetes-sigs/controller-runtime/blob/1657cf64eb90e0d9143873122502c872e60e825c/pkg/manager/signals/signal.go#L27-L30

This is problematic for controller-managers managing a webhook server since graceful webhook server shutdown may begin before kube-proxy has observed the termination (and reconciled local iptables rules) leading to connection attempts after webhook server listeners have already been closed.

This PR adds support for a `ShutdownDelay` webhook server option allowing users to delay webhook server shutdown and give time for routes to be updated so that clients connect to other server replicas before closing server listeners. In particular, on GKE the kube-proxy `--iptables-min-sync-period` flag is currently set to 10 seconds, so a delay duration of 15 seconds might make sense on such clusters.

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -101,6 +101,12 @@ type Options struct {
 
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
+
+	// ShutdownDelay delays server shutdown to wait for clients to stop opening
+	// new connections before closing server listeners. HTTP keep-alives are
+	// disabled during this time to allow persistent connections to be closed
+	// gracefully. Defaults to 0.
+	ShutdownDelay time.Duration
 }
 
 // NewServer constructs a new webhook.Server from the provided options.
@@ -247,9 +253,16 @@ func (s *DefaultServer) Start(ctx context.Context) error {
 	go func() {
 		<-ctx.Done()
 		log.Info("Shutting down webhook server with timeout of 1 minute")
-
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 		defer cancel()
+		// Disable HTTP keep-alives to close persistent connections gracefully
+		srv.SetKeepAlivesEnabled(false)
+		// Wait for the specified shutdown delay or until the shutdown context
+		// expires, whichever happens first
+		select {
+		case <-time.After(s.Options.ShutdownDelay):
+		case <-ctx.Done():
+		}
 		if err := srv.Shutdown(ctx); err != nil {
 			// Error from closing listeners, or context timeout
 			log.Error(err, "error shutting down the HTTP server")
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -103,9 +103,7 @@ type Options struct {
 	WebhookMux *http.ServeMux
 
 	// ShutdownDelay delays server shutdown to wait for clients to stop opening
-	// new connections before closing server listeners. HTTP keep-alives are
-	// disabled during this time to allow persistent connections to be closed
-	// gracefully. Defaults to 0.
+	// new connections before closing server listeners. Defaults to 0.
 	ShutdownDelay time.Duration
 }
 
@@ -256,8 +254,6 @@ func (s *DefaultServer) Start(ctx context.Context) error {
 
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 		defer cancel()
-		// Disable HTTP keep-alives to close persistent connections gracefully
-		srv.SetKeepAlivesEnabled(false)
 		// Wait for the specified shutdown delay or until the shutdown context
 		// expires, whichever happens first
 		select {
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -257,6 +257,7 @@ func (s *DefaultServer) Start(ctx context.Context) error {
 		srv.SetKeepAlivesEnabled(false)
 		<-time.After(s.Options.ShutdownDelay)
 		log.Info("Shutting down webhook server with timeout of 1 minute")
+
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 		defer cancel()
 		if err := srv.Shutdown(ctx); err != nil {
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -250,14 +250,16 @@ func (s *DefaultServer) Start(ctx context.Context) error {
 	idleConnsClosed := make(chan struct{})
 	go func() {
 		<-ctx.Done()
+
 		// Disable HTTP keep-alives to close persistent connections after next
 		// HTTP request. Clients may reconnect until routes are updated and
 		// server listeners are closed however this should start gradually
 		// migrating clients to server instances that are not about to shutdown
 		srv.SetKeepAlivesEnabled(false)
+		// Wait before shutting down webhook server
 		<-time.After(s.Options.ShutdownDelay)
-		log.Info("Shutting down webhook server with timeout of 1 minute")
 
+		log.Info("Shutting down webhook server with timeout of 1 minute")
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 		defer cancel()
 		if err := srv.Shutdown(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


inteon:
@dippynark Should we not create a webhook-specific solution?

I do remember reading a great article on how to properly shutdown a webserver on Kubernetes; unfortunately, I cannot find it anymore.
I do remember reading that new connections should still be accepted, but old connections should start being drained when you receive a shutdown signal. After a certain amount of time, the server can stop accepting new connections and fully shutdown.
It would be great if controller-runtime could apply some of these best-practices by default (only to its webhooks). I'm not sure just delaying the context cancel is the most elegant solution.

inteon:
> @inteon Nice! When I looked at this I couldn't see how to add some time between triggering shutdown and closing listeners: https://github.com/golang/go/blob/de5b418bea70aaf27de1f47e9b5813940d1e15a4/src/net/http/server.go#L2986-L2989
> 
> But didn't know about `SetKeepAlivesEnabled`! Will have a go at doing shutdown this way

I already started creating something. Will share with you shortly.

alvaroaleman:
Another thing that I do not understand about this: The Kube-Apiserver doesn't actually use kube-proxy to reach webhooks and extended apiserves, it directly resolves the endpoints internally which means the described problem shouldn't apply...?

dippynark:
@alvaroaleman I thought this was the case too before working on this, but as far as I can tell the standard API Server -> webhook setup does rely on kube-proxy.

Definitely not 100% on this, but my understanding is that when the API server wants to connect to a webhook behind a Kubernetes Service, it typically proxies the request via the [Konnectivity Agent](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1281-network-proxy) running on the cluster which performs Service DNS resolution and connects to the Service VIP, making use of kube-proxy routing; this allows connectivity even when the control plane and webhook Pods are in disjoint networks.

I tried to follow through the API server code which seems to confirm this. For example, [here](https://github.com/kubernetes/apiserver/blob/d6876a0600de06fef75968c4641c64d7da499f25/pkg/util/webhook/client.go#L176-L179) the API server is setting up the webhook client using the Service hostname before delegating to the cluster dialer provided by the EgressSelector instead of picking one of the endpoints itself.

pkg/webhook/server.go
```diff
@@ -247,9 +251,19 @@ func (s *DefaultServer) Start(ctx context.Context) error {
...
+		// Disable HTTP keep-alives to close persistent connections after next
+		// HTTP request. Clients may reconnect until routes are updated and
+		// server listeners are closed however this should start gradually
+		// migrating clients to server instances that are not about to shutdown
+		srv.SetKeepAlivesEnabled(false)
+		// Wait for the specified shutdown delay or until the shutdown context
+		// expires, whichever happens first
+		select {
+		case <-time.After(s.Options.ShutdownDelay):
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
dippynark: hmm yeah I don't think it's great being capped at 1 minute. Does it look better now? I've changed it to always wait for the full shutdown delay

pkg/webhook/server.go
```diff
@@ -246,6 +250,15 @@ func (s *DefaultServer) Start(ctx context.Context) error {
...
 	go func() {
 		<-ctx.Done()
+
+		// Disable HTTP keep-alives to close persistent connections after next
+		// HTTP request. Clients may reconnect until routes are updated and
+		// server listeners are closed however this should start gradually
+		// migrating clients to server instances that are not about to shutdown
+		srv.SetKeepAlivesEnabled(false)
+		// Wait before shutting down webhook server
+		<-time.After(s.Options.ShutdownDelay)
```
inteon, line 260:
```suggestion
		time.Sleep(s.Options.ShutdownDelay)
```
We don't need to create a channel, I think.

pkg/webhook/server.go
```diff
@@ -101,6 +101,10 @@ type Options struct {
 
 	// WebhookMux is the multiplexer that handles different webhooks.
 	WebhookMux *http.ServeMux
+
+	// ShutdownDelay delays server shutdown to wait for clients to stop opening
+	// new connections before closing server listeners. Defaults to 0.
```
inteon, line 106: Open question:
Would it make sense to set a > 0 default, eg. 15s which is less than the 30 seconds default graceful shutdown delay?

inteon:
> Another thing that I do not understand about this: The Kube-Apiserver doesn't actually use kube-proxy to reach webhooks and extended apiserves, it directly resolves the endpoints internally which means the described problem shouldn't apply...?

There seems to be a `--enable-aggregator-routing` flag, maybe that is related?
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
```Turns on aggregator routing requests to endpoints IP rather than cluster IP.```

File: pkg/webhook/server.go (hunk 2 above)
Comment (dippynark) [reply pkg/webhook/server.go]: Haha oops, fixed

File: pkg/webhook/server.go (hunk 3 above)
Comment (dippynark) [reply pkg/webhook/server.go:106]: It's a tricky one for sure, but because it's so specific to the Kubernetes cluster you're running on it doesn't feel right to align to any particular cloud provider (e.g. I'm not sure what value EKS and AKS use for example).

IMO we should keep it at 0 for backwards compatibility and to ensure that connection draining 'doesn't work' consistently across clusters by default. I think this is also more likely to encourage applications that care a lot about this (e.g. Gatekeeper) to expose a flag to change this to make it easier to reconfigure for the cluster you're running on.

No problems with changing to 15 seconds by default though if you were leaning towards that (that would be better for us anyway because we're using GKE 😄). We could alternatively align to the kube-proxy default of 1 second? But on GKE at least this won't help much.

**Comment by sbueringer** [comment]
Q: Did I get this correctly that kube-proxy reacts to the deletionTimestamp on the Pod and not only on readiness when dropping a Pod from a Service?

**Comment by dippynark** [comment]
@sbueringer Yeah, the following actions are triggered simultaneously when a Pod's deletion timestamp is set:

- The kubelet runs any preStop hooks and then sends a TERM signal to the container processes
- EndpointSlices are updated to indicate that the Pod is terminating which triggers kube-proxy to remove corresponding iptables rules

There are more details here: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

Note that there is currently no way to guarentee that all kube-proxy instances have reconciled EndpointSlice updates before shutting down (leading to issues like I described in this PR). This best mechanism I am aware of is to delay shutdown for slightly longer than the `--iptables-min-sync-period` kube-proxy flag: https://github.com/kubernetes/kubernetes/issues/106476

File: pkg/webhook/server.go [hunk 4]
```diff
@@ -246,6 +250,15 @@ func (s *DefaultServer) Start(ctx context.Context) error {
 	idleConnsClosed := make(chan struct{})
 	go func() {
 		<-ctx.Done()
+
+		// Disable HTTP keep-alives to close persistent connections after next
+		// HTTP request. Clients may reconnect until routes are updated and
+		// server listeners are closed however this should start gradually
+		// migrating clients to server instances that are not about to shutdown
+		srv.SetKeepAlivesEnabled(false)
```
vincepri, line 258: This seems a good addition

pkg/webhook/server.go
```diff
@@ -246,6 +250,15 @@ func (s *DefaultServer) Start(ctx context.Context) error {
...
+		// Wait before shutting down webhook server
+		time.Sleep(s.Options.ShutdownDelay)
```
vincepri, line 260: Why is this needed? Reading the intent of the PR, there is a context with a timeout that will be given to `shutdown` which allows those connections to be closed, was that not enough?

dippynark: Although `srv.Shutdown(ctx)` allows active connections to finish, it will close listeners straight away before kube-proxy has had a chance to remove the iptables rules from all cluster nodes that are routing new connections. Adding a sleep here gives time for kube-proxy to sync ready endpoints to avoid it routing new connections to closed listeners

This comment speaks more about how kube-proxy iptables are involved in webhook connectivity: https://github.com/kubernetes-sigs/controller-runtime/pull/2601#issuecomment-1837570916

dippynark:
I've reopened this PR as there seems to be renewed interest: https://github.com/kubernetes-sigs/controller-runtime/issues/3113

PR #2605, opened 2023-12-03
🐛 Webhook: Handle http.NoBody

inteon:
Add handling of `http.NoBody`.

Code is based on https://github.com/kubernetes/kubernetes/blob/c28c2009181fcc44c5f6b47e10e62dacf53e4da0/staging/src/k8s.io/pod-security-admission/cmd/webhook/server/server.go


inteon:
> `http.NoBody` is go-doced as:
> 
> > It can be used in an outgoing client request to explicitly signal that a request has zero bytes.
> 
> How would this ever be set in an incoming request?

Upstream the change is explained as follows: https://github.com/kubernetes/kubernetes/pull/115908
That PR refers to the docs here: https://pkg.go.dev/net/http#Request

PR #2611, opened 2023-12-06
⚠️ Remove apiutil.NewDiscoveryRESTMapper, use DynamicRESTMapper by default for cache

alvaroaleman:
The discovery restmapper was unused except for two places that we missed (when creating a cache directly as opposed through the manager and in a test). The discovery restmapper only ever loads mappings once during startup and is unable to reload them if they change. It also isn't lazy and we don't want anyone using it.

Instead, the `DynamicRESTMapper` should be used.


vincepri (approved):

vincepri:
/retitle ⚠️ Remove apiutil.NewDiscoveryRESTMapper, use DynamicRESTMapper by default for cache

PR #2623, opened 2023-12-14
✨Add flowcontrol api v1 for Kubernetes v1.29.0

troy0820:
Kubernetes released [v1.29.0](https://kubernetes.io/blog/2023/12/13/kubernetes-v1-29-release/#removal-of-the-v1beta2-flow-control-api-group) and with that release was the deprecation and removal of flowcontrol v1beta2.  

This PR changes the fake client to not serve that for v1.29.0 when controller-runtime v0.17 gets released 

- [x] Change v1beta2 => v1 in flowcontrol in `inTreeResourcesWithStatus()` function

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1254,6 +1254,8 @@ func inTreeResourcesWithStatus() []schema.GroupVersionKind {
 
 		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1beta2", Kind: "FlowSchema"},
 		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1beta2", Kind: "PriorityLevelConfiguration"},
+		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1", Kind: "FlowSchema"},
+		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1", Kind: "PriorityLevelConfiguration"},
 	}
 }
 


pkg/client/fake/client.go
```diff
@@ -1252,8 +1252,8 @@ func inTreeResourcesWithStatus() []schema.GroupVersionKind {
 
 		{Group: "apiextensions.k8s.io", Version: "v1", Kind: "CustomResourceDefinition"},
 
-		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1beta2", Kind: "FlowSchema"},
-		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1beta2", Kind: "PriorityLevelConfiguration"},
+		{Group: "flowcontrol.apiserver.k8s.io", Version: "v1", Kind: "FlowSchema"},
```
alvaroaleman, line 1255: I think we should have both. This config is just for the status subresource. Most of our users will not be using kube 1.29 right away and so they might be forced to use the older version. I don't see a reason to break that.

troy0820: That's fair, I'll update it with the v1beta2 as well.

alvaroaleman (approved):
/retest-required

sbueringer:
Thx! @troy0820 can you please update the PR description. It's just a bit misleading now :)

PR #2631, opened 2023-12-26
:bug: CreateOrPatch fails to update status when creating object

### LINKED ISSUE #2627: CreateOrPatch / CreateOrUpdate do not set status properly when the object is created
When using the `CreateOrPatch` or `CreateOrUpdate` methods, the status of the object won't be persisted if the object does not exist.

In both functions:
- https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/controller/controllerutil/controllerutil.go#L247
- https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/controller/controllerutil/controllerutil.go#L208

The method returns after the Create call.

However, this does not set the status of the object (see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource for instance `PUT/POST/PATCH requests to the custom resource ignore changes to the status stanza.`)

The code needs to be updated to actually do the status update part even on creation. I'll have a look at it and make a proposal for the change

jfremy:
The `CreateOrPatch` code directly exits after creating an object, it does not consider whether the status was modified or not.
This means that any change made to the status will be discarded.

Since the code correctly takes care of the status in the resource update case, I’ve updated the code to also do it in the create code path. I’ve tried to re-use as much as possible the existing code.

I’ve introduced a new operation result value (`OperationResultCreatedStatus` to indicate an object was created and its status updated as well) but we might want to hold on that since that could create issues with existing users of the method and keep returning the `OperationResultCreated` code to maintain backward compatibility.

I also included a non regression test to validate the change is working as expected.

Fixes #2627

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -261,6 +261,8 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 	OperationResultUpdatedStatus OperationResult = "updatedStatus"
 	// OperationResultUpdatedStatusOnly means that only an existing status is updated.
 	OperationResultUpdatedStatusOnly OperationResult = "updatedStatusOnly"
+	// OperationResultCreatedStatus means that an object was created and its status updated
+	OperationResultCreatedStatus OperationResult = "createdStatus"
 )
 
 // CreateOrUpdate creates or updates the given object in the Kubernetes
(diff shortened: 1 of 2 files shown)


troy0820:
/hold 
Discussion in issue submitted.

PR #2633, opened 2024-01-03
:warning: Fakeclient: Only set TypeMeta for unstructured

alvaroaleman:
Currently, the fakeclient unconditionally sets metadata for all objects retrieved through it. This causes two problems:
* It differs from the behavior of the real client, except for the cached one if `DeepCopy` is enabled and might lead ppl to think that they can rely on `TypeMeta` being populated, which is absolutely not the case
* It causes panics for types that have `TypeMeta` defined as pointer, making it impossible to use the client with them

This PR changes the behavior to only populate TypeMeta for `unstructured`.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -334,10 +334,12 @@ func (t versionedTracker) Create(gvr schema.GroupVersionResource, obj runtime.Ob
 // tries to assign whatever it finds into a ListType it gets from schema.New() - Thus we have to ensure
 // we save as the very same type, otherwise subsequent List requests will fail.
 func convertFromUnstructuredIfNecessary(s *runtime.Scheme, o runtime.Object) (runtime.Object, error) {
-	gvk := o.GetObjectKind().GroupVersionKind()
-
 	u, isUnstructured := o.(runtime.Unstructured)
-	if !isUnstructured || !s.Recognizes(gvk) {
+	if !isUnstructured {
+		return o, nil
+	}
+	gvk := o.GetObjectKind().GroupVersionKind()
+	if !s.Recognizes(gvk) {
 		return o, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -527,21 +529,21 @@ func (c *fakeClient) List(ctx context.Context, obj client.ObjectList, opts ...cl
 		return err
 	}
 
-	ta, err := meta.TypeAccessor(o)
-	if err != nil {
-		return err
+	if _, isUnstructured := obj.(runtime.Unstructured); isUnstructured {
```
sbueringer, line 532: Do we need the same for patch in l.874++?

(+ corresponding test cases)

pkg/client/fake/client_test.go
```diff
@@ -1877,13 +1873,136 @@ var _ = Describe("Fake client", func() {
...
 
+type WithPointerMetaList struct {
+	*metav1.ListMeta
+	*metav1.TypeMeta
+	Items []*WithPointerMeta
+}
+
+func (t *WithPointerMetaList) DeepCopy() *WithPointerMetaList {
+	l := &WithPointerMetaList{
+		ListMeta: t.ListMeta.DeepCopy(),
```
sbueringer, line 1964: nit: does it make sense to also cover the nil cases for ListMeta (and ObjectMeta below)

pkg/client/fake/client.go
alvaroaleman: Great catch, updated it and added testcases for patch, update and delete

pkg/client/fake/client_test.go
alvaroaleman: The `DeepCopy` implementations do nothing when they copy from a nil object which means that we implicitly have that for the `ListMeta` case. For the `ObjectMeta` case that isn't possible, because you can not access an object without specyfing at least a name anyways.

sbueringer: Makes sense, thx! Didn't really think about that the Meta objects could be just nil safe :)

sbueringer:
/lgtm

/hold
Feel free to merge, not sure if you want additional reviews

PR #2637, opened 2024-01-05
:warning: Rename leader election config fields

### LINKED ISSUE #2636: Proposal: Naming consistency for leader election lock options
This is probably a very minor naming suggestion but when I first started out using controller-runtime, the `LeaderElectionNamespace` option's identifier confused me a bit as to what it does and made me look under the covers to see how it is used.

I think clarifying this as follows would address the problem:

```diff
@@ -176,13 +176,13 @@ type Options struct {
 	// act on the same resources concurrently.
 	LeaderElectionResourceLock string
 
-	// LeaderElectionNamespace determines the namespace in which the leader
+	// LeaderElectionLockNamespace determines the namespace in which the leader
-	// election resource will be created.
+	// election lock resource will be created.
-	LeaderElectionNamespace string
+	LeaderElectionLockNamespace string
 
-	// LeaderElectionID determines the name of the resource that leader election
+	// LeaderElectionLockID determines the name of the resource that leader election
 	// will use for holding the leader lock.
-	LeaderElectionID string
+	LeaderElectionLockID string
 
 	// LeaderElectionConfig can be specified to override the default configuration
 	// that is used to build the leader election client.
```

If we go through with this, I think changing `LeaderElectionLockID` → `LeaderElectionLockName` would  be quite useful as well (since ID is not the Kubernetes term for unique resource names).

I understand these names have been there since the inception, but since I observed this repo regularly changes Options structs (and this is particularly not a troublesome migration), I'm proposing this. Thoughts?

/kind feature

ahmetb:
Changing the `manager.Options` and `leaderelection.Options` field names around leader election configuration to be more explanatory and reflective of what they actually do.

- `LeaderElectionNamespace` -> `LeaderElectionLockNamespace` since it actually is the namespace for the lock.

- `LeaderElectionID` -> `LeaderElectionLockName` since it actually is the resource name of the lock.

- Also updated `pkg/leaderelection` to remove stuttering in the field names since the type is already `leaderelection.Options` and there is no need to repeat the `LeaderElection` prefix.

Closes #2636.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -176,13 +176,13 @@ type Options struct {
 	// act on the same resources concurrently.
 	LeaderElectionResourceLock string
 
-	// LeaderElectionNamespace determines the namespace in which the leader
-	// election resource will be created.
-	LeaderElectionNamespace string
+	// LeaderElectionLockNamespace determines the namespace of the leader election lock
+	// resource will be created in.
+	LeaderElectionLockNamespace string
 
-	// LeaderElectionID determines the name of the resource that leader election
+	// LeaderElectionLockName determines the name of the resource that leader election
 	// will use for holding the leader lock.
-	LeaderElectionID string
+	LeaderElectionLockName string
 
 	// LeaderElectionConfig can be specified to override the default configuration
 	// that is used to build the leader election client.
@@ -196,7 +196,7 @@ type Options struct {
 	LeaderElectionReleaseOnCancel bool
 
 	// LeaderElectionResourceLockInterface allows to provide a custom resourcelock.Interface that was created outside
-	// of the controller-runtime. If this value is set the options LeaderElectionID, LeaderElectionNamespace,
+	// of the controller-runtime. If this value is set the options LeaderElectionLockName, LeaderElectionLockNamespace,
 	// LeaderElectionResourceLock, LeaseDuration, RenewDeadline and RetryPeriod will be ignored. This can be useful if you
 	// want to use a locking mechanism that is currently not supported, like a MultiLock across two Kubernetes clusters.
 	LeaderElectionResourceLockInterface resourcelock.Interface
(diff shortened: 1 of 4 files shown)


troy0820:
Failing api-diff, unsure how to deal with removals and utilize this tool to not flag what is being renamed.  @joelanford  👀

ahmetb:
Peeked at old PRs that also did similar changes, they also had the failure at the time of merging. I think that failure is largely informational.

pkg/manager/manager.go
```diff
@@ -176,13 +176,13 @@ type Options struct {
 	// act on the same resources concurrently.
 	LeaderElectionResourceLock string
 
-	// LeaderElectionNamespace determines the namespace in which the leader
-	// election resource will be created.
-	LeaderElectionNamespace string
+	// LeaderElectionLockNamespace determines the namespace of the leader election lock
```
alvaroaleman, line 179: While I do agree that the new names are better, renaming these will result in a lot of churn for consumers, I am not sure if that is worth it

ahmetb: Yeah totally valid. It doesn't offer any net new improvement.

joelanford:
> Failing api-diff, unsure how to deal with removals and utilize this tool to not flag what is being renamed. @joelanford 👀

Yes, this is just an informational test that helps us ensure we avoid accidental API breakage. A failure for intentional API breakage does not block merge.

pkg/manager/manager.go
troy0820: Would leaving a comment in the docs for this be better?  I don't see this getting merged being that the churn associated with this and the complexity @joelanford  mentioned with trying to promote these fields and deprecating others would help either.

vincepri: To echo other's comments, given that the field are self explanatory, maybe not 100% accurate I guess (for historical reasons these have changed the internal meaning); could we add more information on the comment, rather than renaming them?

ahmetb:
Closing based on the decision on the proposal.
/close

PR #2638, opened 2024-01-05
✨ Add 100/1000s buckets for prometheus workqueue histograms

### LINKED ISSUE #2625: higher buckets for workqueue histograms
The workqueue subsystem exposes several metrics, including the `workqueue_queue_duration_seconds` and `workqueue_work_duration_seconds` buckets.
These currently have exponential buckets going from 1ns -> 10s.

https://github.com/kubernetes-sigs/controller-runtime/blob/1b80b9629af8ec55ae200e2c649a9603533eed20/pkg/metrics/workqueue.go#L53-L65

I think these should go up a further 2 steps to include 100s and 1000s as buckets (and maybe drop the lowest 2 buckets if we're concerned about the number), especially for `workqueue_queue_duration_seconds` as a stuck / poorly implemented controller can have quite high latency in processing reconciliation requests

seankhliao:
Controllers making many external requests in large clusters may have normal operating latency on the order of ~100s. Add buckets that cover the range, allowing metric backends to interpolate percentile estimates properly.

For #2625

### DIFF
--- a/pkg/metrics/workqueue.go
+++ b/pkg/metrics/workqueue.go
@@ -54,14 +54,14 @@ var (
 		Subsystem: WorkQueueSubsystem,
 		Name:      QueueLatencyKey,
 		Help:      "How long in seconds an item stays in workqueue before being requested",
-		Buckets:   prometheus.ExponentialBuckets(10e-9, 10, 10),
+		Buckets:   prometheus.ExponentialBuckets(10e-9, 10, 12),
 	}, []string{"name"})
 
 	workDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Subsystem: WorkQueueSubsystem,
 		Name:      WorkDurationKey,
 		Help:      "How long in seconds processing an item from workqueue takes.",
-		Buckets:   prometheus.ExponentialBuckets(10e-9, 10, 10),
+		Buckets:   prometheus.ExponentialBuckets(10e-9, 10, 12),
 	}, []string{"name"})
 
 	unfinished = prometheus.NewGaugeVec(prometheus.GaugeOpts{


troy0820:
/retitle ✨ Add 100/1000s buckets for prometheus workqueue histograms

troy0820:
It can't merge without approval but I want the maintainers to be able to take a look at this to get consensus around this solution. 

/hold

alvaroaleman:
> Controllers making many external requests in large clusters may have normal operating latency on the order of ~100s

Not really, no. This would be indicative of doing long-running and/or blocking operations in your Reconcile which you shouldn't do, use `RequeueAfter` instead.

seankhliao:
Or you have a poorly written vendor controller, or we just have a lot of work and we're starved of parallelism. Either way, I'd like to be able to observe how poorly it's performing beyond just 10s.

vincepri:
Not opposed to this, @alvaroaleman ? Seems a small improvement?

sbueringer:
In the past I used `controller_runtime_reconcile_time_seconds` for this purpose. Exactly because the workqueue metrics are capped at 10 seconds.

https://github.com/kubernetes-sigs/controller-runtime/blob/d10ae95a3583b4b75a682ec3288cf970f9449a5a/pkg/internal/controller/metrics/metrics.go#L51-L56

(Might be helpful retroactively in any case)

In general no objection to adding additional buckets. I have no idea what the ideal buckets are to be honest.

sbueringer:
For folks not super familiar with metrics it can look pretty confusing that the metrics are capped at 10s (as it looks like the durations are never slower than that)

troy0820:
> In general no objection to adding additional buckets. I have no idea what the ideal buckets are to be honest.

I have no idea what is ideal as well. Initially in the issue, I wasn't trying to encourage more buckets as that may or may not impact performance, however the need for an upper bound outside of 10s seemed to be necessary.  

> For what it's worth upstream still only has buckets until 10s: https://github.com/kubernetes/component-base/blob/master/metrics/prometheus/workqueue/metrics.go#L55-L69

However, however, to stay consistent with upstream, would this be beneficial to add and not use the other metrics as you proposed?

sbueringer:
> However, however, to stay consistent with upstream, would this be beneficial to add and not use the other metrics as you proposed?

The controller_runtime_reconcile_time_seconds metric only ~ matches one of the workqueue metrics. So it definitely makes sense for folks to also use the work queue metrics. They are also not 100% the same data, just very close.

I think performance-wise it's probably fine. I guess it's just 20% more data for the two histogram metrics. As the number of workqueues in a controller is usually not that high that it should become a problem I think it's okay.

sbueringer:
I discussed it with Alvaro. We're fine with the change for now.

A larger issue we have is that all the metric configuration is done in `init` functions. Ideally we would instead configure metrics during startup (~ manager.Start()). This would make it possible for folks to configure metrics depending on their needs instead of having to live with the hard-coded configuration we have. We'll try to get this improved in the upcoming release cycle. At this point we also might re-evaluate the "default configuration" of the metrics.

/lgtm
/approve
/hold cancel

Issue #2596, opened 2023-11-27
Changes to webhook.Validator pattern cause version update problems


nathanperkins:
The webhook.Validator interface is implemented on API types to provide easy webhook creation. However, tying this interface to API types means that API code is coupled with implementation code.

When the webhook.Validator interface is changed, modules cannot import each other across the breaking change boundary. It is difficult to achieve synchronous updates across a single organization maintaining many controllers, let alone the entire community.

For example, in 0.15, the webhookValidate interface was changed to `return (admission.Warnings, error)` instead of `return error`. We have updated our controllers to 0.15, but can't import [metallb's APIs](https://github.com/metallb/metallb/blob/main/api/v1beta1/addresspool_webhook.go) until they are also at 0.15, because their use of webhook.Validator is incompatible with our module's choice to use 0.15.

Breaking changes in other areas of the code are less problematic because it is common for repos to only import each other for APIs. If the code dependent on the breaking change is not actually used, it will not be included in compilation and it will compile fine, even though a complete compilation would fail.

We were not aware of the danger of this pattern, and only recently started moving away from it for our controllers to reduce the scope of this problem. But we cannot fix it in third party repos, without forking it and fixing it locally ourselves. We could do pull requests, but they will run into their own problems where updating their repo requires them to update all the downstream repos.

Is there a solution for controller-runtime users? If not, is there something that can be done in the project to reduce this issue?

Related: #2327

alvaroaleman:
API packages should not import anything other than other api packages to avoid precisely this problem. Controller-Runtime does not guarantee compatibility across versions.

troy0820:
@nathanperkins with your resolution from the "edit" you had in your comment, does @alvaroaleman's explanation along with the fix you provided settle this issue?

nathanperkins:
It doesn't settle the issue. The problem is that implementing the webhook.Validator pattern requires importing controller-runtime in the API packages because the return signatures include admission.Warnings and these are supposed to be implemented on your CRD types.

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission#Validator

>Validator defines functions for validating an operation. The custom resource kind which implements this interface can validate itself. To validate the custom resource with another specific struct, use CustomValidator instead.

There are other ways to implement webhook validation but this pattern causes major issues with compatibility in the K8s ecosystem.

sbueringer:
[...]

PR #2639, opened 2024-01-06
:warning: Deprecate admission.Validator and admission.Defaulter

alvaroaleman:
`admission.Validator/Defaulter` require to import controller-runtime into api packages. This is not recommended, as api packages should themselves not have any dependencies except to other api packages in order to make importing them easy without dependency issues and possibly incompatibilities.

This change marks the two as deprecated. We are not going to remove them yet to give folks some time to migrate off them.

Closes https://github.com/kubernetes-sigs/controller-runtime/issues/2596

### DIFF
--- a/examples/crd/pkg/resource.go
+++ b/examples/crd/pkg/resource.go
@@ -17,14 +17,8 @@ limitations under the License.
 package pkg
 
 import (
-	"fmt"
-	"time"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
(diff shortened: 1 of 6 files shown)


troy0820:
/lgtm 

Will kubebuilder need to change their scaffolding to improve this going forward?

examples/crd/pkg/resource.go
```diff
@@ -62,61 +56,6 @@ type ChaosPodList struct {
...
-
-// ValidateCreate implements webhookutil.validator so a webhook will be registered for the type
-func (c *ChaosPod) ValidateCreate() (admission.Warnings, error) {
-	log.Info("validate create", "name", c.Name)
-
-	if c.Spec.NextStop.Before(&metav1.Time{Time: time.Now()}) {
-		return nil, fmt.Errorf(".spec.nextStop must be later than current time")
-	}
-	return nil, nil
-}
```
vincepri, line 77: Can we add an example with the CustomValidator interface?

alvaroaleman: There are existing examples in https://github.com/kubernetes-sigs/controller-runtime/tree/main/examples/builtins

PR #2643, opened 2024-01-08
🌱 Downgrade github.com/evanphx/json-patch library to v4 to match upstream

aerfio:
Changes an import of `github.com/evanphx/json-patch` to `github.com/evanphx/json-patch/v5`. Other codepaths use only json-patch/v5. It removes 1 direct dependency

### DIFF
--- a/examples/scratch-env/go.sum
+++ b/examples/scratch-env/go.sum
@@ -8,8 +8,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/emicklei/go-restful/v3 v3.11.0 h1:rAQeMHw1c7zTmncogyy8VvRZwtkmkZ4FxERmMY4rD+g=
 github.com/emicklei/go-restful/v3 v3.11.0/go.mod h1:6n3XBCmQQb25CM2LCACGz8ukIrRry+4bhvbpWn3mrbc=
-github.com/evanphx/json-patch v5.7.0+incompatible h1:vgGkfT/9f8zE6tvSCe74nfpAVDQ2tG6yudJd8LBksgI=
-github.com/evanphx/json-patch v5.7.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/evanphx/json-patch v4.12.0+incompatible h1:4onqiflcdA9EOZ4RxV643DvftH5pOlLGNtQ5lPWQu84=
+github.com/evanphx/json-patch v4.12.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
 github.com/evanphx/json-patch/v5 v5.7.0 h1:nJqP7uwL84RJInrohHfW0Fx3awjbm8qZeFv0nW9SYGc=
 github.com/evanphx/json-patch/v5 v5.7.0/go.mod h1:VNkHZ/282BpEyt/tObQO8s5CMPmYYq14uClGH4abBuQ=
 github.com/fsnotify/fsnotify v1.7.0 h1:8JEhPFa5W2WU7YfeZzPNqzMP6Lwt7L2715Ggo0nosvA=
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.21
 toolchain go1.21.0
 
 require (
-	github.com/evanphx/json-patch v5.7.0+incompatible
+	github.com/evanphx/json-patch v4.12.0+incompatible // Using v4 to match upstream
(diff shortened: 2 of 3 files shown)


pkg/client/fake/client.go
```diff
@@ -29,8 +29,7 @@ import (
 	"sync"
 	"time"
 
-	// Using v4 to match upstream
```
sbueringer, line 32: /hold

This was intentional, see the godoc comment. Did upstream move to v5?

inteon:
FYI: https://github.com/kubernetes/kubernetes/pull/120327

pkg/client/fake/client.go
aerfio: I saw it, but I concluded that it was stale comment because in go.mod there was only v5/v5+incompatible. Now that I know there was some issue with newer versions of this lib (thanks @inteon) I'll research what to do next.

sbueringer: The reason we use the older version here was that we copied code from upstream and wanted it to use the same library as upstream. 

I wasn't aware of general issues in the library. I wonder if we should re-visit our usage of v5 with this in mind.

sbueringer: cc @vincepri @alvaroaleman We should probably figure out if we have to do something before we cut v0.17

aerfio: https://github.com/kubernetes/kubernetes/issues/120326 mentions that v5.6.0 does not include https://github.com/evanphx/json-patch/commit/50fdc0b4c9ab36933e6a0c9288fef3ac5df2b907, but it's part of v5.7.0 btw
image

sbueringer: Maybe it's only if you use 

`github.com/evanphx/json-patch v5.6.0+incompatible`

And not for

`github.com/evanphx/json-patch/v5 v5.6.0`

sbueringer: Okay so my interpretation:
* Kubernetes intentionally stayed on v4 because of: https://github.com/kubernetes/kubernetes/pull/91622#issuecomment-637087847
* For the `github.com/evanphx/json-patch` module  "v5.6.0+incompatible is older than v4.12.0+incompatible" (https://github.com/kubernetes/kubernetes/issues/120326#issue-1876406255) 
* client-go v0.29 is using `v4.12.0+incompatible` https://github.com/kubernetes/client-go/blob/v0.29.0/go.mod#L8
* we intentionally use `github.com/evanphx/json-patch` because we copied code from https://github.com/kubernetes/client-go/blob/783d0d33626e59d55d52bfd7696b775851f92107/testing/fixture.go#L146-L194

Based on all this:
* I would "upgrade" `github.com/evanphx/json-patch` to be in sync with client-go v0.29
* I would keep our regular `github.com/evanphx/json-patch/v5` dependency. I assume downgrading to v4 is a breaking change just like upgrading to v5 for Kubernetes
  * I didn't dig deeper into https://github.com/kubernetes/kubernetes/pull/91622#issuecomment-637087847 (which was the reason why Kubernetes stayed on v4)

aerfio: Ok, I've downgraded this dep to v4. I also copied the comment that you originally asked about to the go.mod, so that any changes to this line will be visible in the future. I hope this will prevent accidental changes

aerfio:
Hmm the `pull-controller-runtime-apidiff` job fails with:
```
verifying api diff
invoking: 'hack/tools/bin/go-apidiff 984a121dbb8f6273eef55ebb4016e3b3deef1919 --print-compatible'
sigs.k8s.io/controller-runtime/pkg/reconcile
  Incompatible changes:
  - AsReconciler: changed from func(sigs.k8s.io/controller-runtime/pkg/client.Client, ObjectReconciler[T]) Reconciler to func(sigs.k8s.io/controller-runtime/pkg/client.Client, ObjectReconciler[T]) Reconciler
  - ObjectReconciler: changed from ObjectReconciler[T sigs.k8s.io/controller-runtime/pkg/client.Object] to ObjectReconciler[T sigs.k8s.io/controller-runtime/pkg/client.Object]
```
But there's no difference between changed to vs from. Is go-apidiff generics compatible? This PR is running a couple hours after https://github.com/kubernetes-sigs/controller-runtime/pull/2592 was merged, which is the first (IIRC) public usage of generics in c-r, it might be related

vincepri:
Try rebasing

aerfio:
@vincepri already did that with the earlier force-push

vincepri (approved):

vincepri:
@sbueringer hold is yours to clear

pkg/client/fake/client.go
sbueringer: Thank you!

sbueringer:
Thx!

/lgtm

/hold cancel

sbueringer:
No idea why but tide was blocked because it though the hold label was there even though it wasn't. Force pushed the PR to force a re-run of everything

aerfio:
Just an additional thought, if I understand this situation correctly: it might be nice to mention in the release notes that c-r users should downgrade github.com/evanphx/json-patch to v4.12.0 to match client-go (or upgrade all the way up to v5.7.0) so that the fix https://github.com/evanphx/json-patch/commit/50fdc0b4c9ab36933e6a0c9288fef3ac5df2b907 is included

sbueringer:
@aerfio Good idea, will do

PR #2645, opened 2024-01-10
🌱 Avoid dependency on apiserver/options if metrics/filters are used

sbueringer:

This also avoids a dependency on component-base/metrics/prometheus/workqueue, which we really want to avoid as it initialized the workqueue metrics before our init.

```go
 go mod why k8s.io/component-base/metrics/prometheus/workqueue
# k8s.io/component-base/metrics/prometheus/workqueue
sigs.k8s.io/cluster-api/util/flags
sigs.k8s.io/controller-runtime/pkg/metrics/filters
k8s.io/apiserver/pkg/server/options
k8s.io/apiserver/pkg/server
k8s.io/apiserver/pkg/endpoints/filters
k8s.io/component-base/metrics/prometheus/workqueue
```


sbueringer:
Re: apidiff. My guess is apidiff might has problems with generics

sbueringer:
> Re: apidiff. My guess is apidiff might has problems with generics

^^ @joelanford just fyi

pkg/metrics/filters/filters.go
```diff
@@ -47,7 +47,12 @@ func WithAuthenticationAndAuthorization(config *rest.Config, httpClient *http.Cl
 		CacheTTL:                 1 * time.Minute,
 		TokenAccessReviewClient:  authenticationV1Client,
 		TokenAccessReviewTimeout: 10 * time.Second,
-		WebhookRetryBackoff:      options.DefaultAuthWebhookRetryBackoff(),
+		WebhookRetryBackoff: &wait.Backoff{
```
inteon, line 50: Could you add a comment that explains where these values were copied from?

PR #2647, opened 2024-01-10
:book: Fix Client Read/Write Functionality Comments

fxierh:
Fix comments related to the read/write functionality of the client to reflect that the client reads from a local cache or API server, and writes directly to the API server.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -90,10 +90,11 @@ type CacheOptions struct {
 type NewClientFunc func(config *rest.Config, options Options) (Client, error)
 
 // New returns a new Client using the provided config and Options.
-// The returned client reads *and* writes directly from the server
-// (it doesn't use object caches).  It understands how to work with
-// normal types (both custom resources and aggregated/built-in resources),
-// as well as unstructured types.
+// The returned client reads from a local cache or directly from the API server,
+// and writes are always performed directly on the API server.
+// (read operations may use object caches, but write operations do not).
+// It understands how to work with normal types (both custom resources
+// and aggregated/built-in resources), as well as unstructured types.
 //
 // In the case of normal types, the scheme will be used to look up the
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -90,10 +90,11 @@ type CacheOptions struct {
 type NewClientFunc func(config *rest.Config, options Options) (Client, error)
 
 // New returns a new Client using the provided config and Options.
-// The returned client reads *and* writes directly from the server
-// (it doesn't use object caches).  It understands how to work with
-// normal types (both custom resources and aggregated/built-in resources),
-// as well as unstructured types.
+// The returned client reads from a local cache or directly from the API server,
```
troy0820, line 93: Is this `or` intentional?  Did the previous client not only read directly from the apiserver?

troy0820: Never mind, I see that if you set `options.Cache` it will read from the cache and not directly from the api server.  However the default case would return a client that reads directly from the api-server.

fxierh: Hi @troy0820, thanks for the comments. The objective is to disambiguate the comments and to reflect that the read could be performed on a local cache (and not necessarily on the API server).

fxierh:
/assign @inteon @joelanford 

Would you please take a look when you get a chance? Thanks.

pkg/client/client.go
inteon: Could you explain in the comment in which cases a cached client is returned? (By referring to the options passed to this function)

fxierh: @inteon Thanks for pointing this out. The comment is updated accordingly. Could you please take another look ?

sbueringer:
Perfect. Thank you very much!

/lgtm
/approve

honnix:
To be honest, I actually prefer the previous delegation model that was removed by #2150, which had better separation of concern and well-defined behaviour.

Issue #895, opened 2020-04-11
RFC: Deprecation and removal of ComponentConfig


christopherhein:
This issue tracks the deprecation and removal of ComponentConfig from Controller Runtime.

vincepri:
/lifecycle frozen
/milestone v0.7.x

@christopherhein Let me know if this should fall into 0.7, or we can wait for later as well

christopherhein:
What is the v0.7 timeframe? 

I'm working on updating the PR this week and will likely have something Friday or early next week…

vincepri:
1-2 months, probably, it all depends when 1.19 is release and when the current planned breaking changes will merge

christopherhein:
Perfect, let's aim for that milestone since as we've discussed this will also have breaking changes. Thanks!

christopherhein:
@vincepri I think we need to remove this tracking issue from the 1.19 release. For now, we should just have #891 give this is tracking the entire feature through alpha -> beta -> ga we'll want the configfile to bake a bit.

christopherhein:
@DirectXMan12 @vincepri looking into at the next phases of this… 

I'd originally documented moving to use `configv1alpha1.ClientConnectionConfiguration` - https://github.com/kubernetes/component-base/blob/master/config/v1alpha1/types.go#L69 as a standard type that we could configure `rest.Config` with. Looking into how this is implemented in `manager.New()` it takes in the `*rest.Config, Options{}` as parameters meaning if we included the fields from `configv1alpha1.ClientConnectionConfiguration` and supported `nil` values for `rest.Config` without erroring (https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L306-L308) we could configure the `*rest.Config` with the values from the file as long as those same fields were defined on the Options struct…

For reference:
```
type ClientConnectionConfiguration struct {
	// kubeconfig is the path to a KubeConfig file.
	Kubeconfig string `json:"kubeconfig"`
	// acceptContentTypes defines the Accept header sent by clients when connecting to a server, overriding the
	// default value of 'application/json'. This field will control all connections to the server used by a particular
	// client.
	AcceptContentTypes string `json:"acceptContentTypes"`
	// contentType is the content type used when sending data to the server from this client.
	ContentType string `json:"contentType"`
	// qps controls the number of queries per second allowed for this connection.
	QPS float32 `json:"qps"`
	// burst allows extra queries to accumulate when a client is exceeding its rate.
[...]

PR #2648, opened 2024-01-10
⚠️ Remove deprecated manager options

troy0820:
`options.AndFrom` and `options.AndFromOrDie` have been deprecated over several releases. This PR will remove them where consumers will have to migrate to their own config solution.   

- [x] Remove function `options.AndFrom`
- [x] Remove function `options.AndFromOrDie`

/kind cleanup


camilamacedo86 (commented):
This code ensures that the options work
So, IMO it should only be removed when the options be removed and are no longer supported.

inteon:
I think we can only remove these functions when we remove `sigs.k8s.io/controller-runtime/pkg/config` too.

troy0820:
> This code ensures that the options work
So, IMO it should only be removed when the options be removed and are no longer supported.

The `options.AndFrom` was loading a config that @inteon agreed that we can only deprecate this if we deprecate the config package.  

Do you mean that the options that get merged with the config options shouldn't be removed? This doesn't remove any options.

https://github.com/kubernetes-sigs/controller-runtime/issues/895

> 👍 I think we can only remove these functions when we remove sigs.k8s.io/controller-runtime/pkg/config too.

Very true, I can delete those as well.

Makefile
```diff
@@ -102,10 +102,6 @@ modules: ## Runs go mod to ensure modules are up to date.
 	cd $(ENVTEST_DIR); go mod tidy
 	cd $(SCRATCH_ENV_DIR); go mod tidy
 
-.PHONY: generate
-generate: $(CONTROLLER_GEN) ## Runs controller-gen for internal types for config file
-	$(CONTROLLER_GEN) object paths="./pkg/config/v1alpha1/...;./examples/configfile/custom/v1alpha1/..."
```
camilamacedo86, line 107: Why it has been removed?

Makefile
```diff
@@ -126,9 +122,4 @@ verify-modules: modules ## Verify go modules are up to date
 		echo "go module files are out of date, please run 'make modules'"; exit 1; \
 	fi
 
-.PHONY: verify-generate
-verify-generate: generate ## Verify generated files are up to date
-	@if !(git diff --quiet HEAD); then \
-		git diff; \
-		echo "generated files are out of date, run make generate"; exit 1; \
-	fi
```
camilamacedo86, line 134: Why it has been removed?

troy0820:
> /lgtm PR LGTM, we now just need an approver to agree that we want to remove this deprecated feature already.

Thank you for the review, I know that this has to be coordinated and maybe targeted for v0.18.0.

Makefile
```diff
@@ -116,10 +116,6 @@ modules: ## Runs go mod to ensure modules are up to date.
 	cd $(ENVTEST_DIR); go mod tidy
 	cd $(SCRATCH_ENV_DIR); go mod tidy
```
sbueringer, line 118: verify-generate can be deleted as well

troy0820: I will make this change.

sbueringer:
Thank you very much!

/lgtm

alvaroaleman:
/retitle ⚠️ Remove deprecated manager options

The current title sounds to me like we were un-deprecating something, WDYT?
/lgtm
/approve
/hold

In case @vincepri wants to have a look

sbueringer:
@troy0820 Thank you very much!

PR #2663, opened 2024-01-24
🐛 Clean restmapper cache if a version is notFound

g-gaston:
This avoids:
- Extra calls to `https://host/apis/<group>/<version>` when a version seen before and cached in apiGroups is deleted or marked as not served.
- Returnning a valid mapping for a cached version that is deleted or not served anymore.

I would like backport this fix to v0.16 (or at version of it since it won't apply cleanly). It would help us solve this issue in cluster-api https://github.com/kubernetes-sigs/cluster-api/issues/10032

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -185,7 +185,11 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 
 	// Update information for group resources about versioned resources.
 	// The number of API calls is equal to the number of versions: /apis/<group>/<version>.
-	groupVersionResources, err := m.fetchGroupVersionResources(groupName, versions...)
+	// If we encounter a missing API version (NotFound error), we will remove the group from
+	// the m.apiGroups and m.knownGroups caches.
+	// If this happens, in the next call the group will be added back to apiGroups
+	// and only the existing versions will be loaded in knownGroups.
+	groupVersionResources, err := m.fetchGroupVersionResourcesLocked(groupName, versions...)
 	if err != nil {
 		return fmt.Errorf("failed to get API group resources: %w", err)
 	}
@@ -194,14 +198,12 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 		groupResources = m.knownGroups[groupName]
 	}
 
-	for version, resources := range groupVersionResources {
-		groupResources.VersionedResources[version.Version] = resources.APIResources
-	}
-
 	// Update information for group resources about the API group by adding new versions.
 	// Ignore the versions that are already registered.
-	for groupVersion := range groupVersionResources {
+	for groupVersion, resources := range groupVersionResources {
 		version := groupVersion.Version
+
+		groupResources.VersionedResources[version] = resources.APIResources
 		found := false
 		for _, v := range groupResources.Group.Versions {
 			if v.Version == version {
@@ -268,9 +270,9 @@ func (m *mapper) findAPIGroupByName(groupName string) (*metav1.APIGroup, error) 
 	return m.apiGroups[groupName], nil
 }
 
-// fetchGroupVersionResources fetches the resources for the specified group and its versions.
+// fetchGroupVersionResourcesLocked fetches the resources for the specified group and its versions.
 // This method might modify the cache so it needs to be called under the lock.
-func (m *mapper) fetchGroupVersionResources(groupName string, versions ...string) (map[schema.GroupVersion]*metav1.APIResourceList, error) {
+func (m *mapper) fetchGroupVersionResourcesLocked(groupName string, versions ...string) (map[schema.GroupVersion]*metav1.APIResourceList, error) {
 	groupVersionResources := make(map[schema.GroupVersion]*metav1.APIResourceList)
 	failedGroups := make(map[schema.GroupVersion]error)
 
@@ -283,6 +285,7 @@ func (m *mapper) fetchGroupVersionResources(groupName string, versions ...string
 			// so it gets refreshed on the next call.
 			delete(m.apiGroups, groupName)
 			delete(m.knownGroups, groupName)
+			continue
 		} else if err != nil {
 			failedGroups[groupVersion] = err
 		}
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -182,9 +182,6 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 		Group:              metav1.APIGroup{Name: groupName},
 		VersionedResources: make(map[string][]metav1.APIResource),
 	}
-	if _, ok := m.knownGroups[groupName]; ok {
-		groupResources = m.knownGroups[groupName]
-	}
 
 	// Update information for group resources about versioned resources.
 	// The number of API calls is equal to the number of versions: /apis/<group>/<version>.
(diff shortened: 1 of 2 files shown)


g-gaston:
cc @fabriziopandini @sbueringer

pkg/client/apiutil/restmapper.go
```diff
@@ -182,23 +182,26 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 		Group:              metav1.APIGroup{Name: groupName},
 		VersionedResources: make(map[string][]metav1.APIResource),
 	}
-	if _, ok := m.knownGroups[groupName]; ok {
-		groupResources = m.knownGroups[groupName]
-	}
 
 	// Update information for group resources about versioned resources.
 	// The number of API calls is equal to the number of versions: /apis/<group>/<version>.
```
sbueringer, line 187: Maybe let's add a note here as well that this func cleans up / resets m.apiGroups & m.knownGroups if we encounter a missing apiVersion

And that this will lead to:
* apiGroups => we'll load the APIGroup again on the next call of findAPIGroupByName
* knownGroups => we'll only add the now existing apiVersions further down (and the removed ones are dropped)

pkg/client/apiutil/restmapper_test.go
```diff
@@ -564,6 +550,131 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+		g.Expect(err).NotTo(gmg.HaveOccurred())
+
+		s := scheme.Scheme
+		err = apiextensionsv1.AddToScheme(s)
+		g.Expect(err).NotTo(gmg.HaveOccurred())
+
+		c, err := client.New(restCfg, client.Options{Scheme: s})
+		g.Expect(err).NotTo(gmg.HaveOccurred())
+
+		// Register a new CRD ina  new group to avoid collisions when deleting versions - "taxi.inventory.example.com".
```
sbueringer, line 574:
```suggestion
		// Register a new CRD in a new group to avoid collisions when deleting versions - "taxis.inventory.example.com".
```

pkg/client/apiutil/restmapper_test.go
```diff
@@ -564,6 +550,131 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+		g.Expect(mapping.GroupVersionKind.Kind).To(gmg.Equal(kind))
+		g.Expect(crt.GetRequestCount()).To(gmg.Equal(4))
+		crt.Reset() // We reset the counter to check how many additional requests are made later.
+
+		// At this point v2 should be cached
+		_, err = lazyRestMapper.RESTMapping(schema.GroupKind{Group: group, Kind: kind}, "v2")
+		g.Expect(err).NotTo(gmg.HaveOccurred())
+		g.Expect(crt.GetRequestCount()).To(gmg.Equal(0))
+
+		// We update the CRD to only have v1 version.
```
sbueringer, line 616: nit. Maybe let's get rid of v1. Just because it's closer to what will be done in reality

pkg/client/apiutil/restmapper_test.go
```diff
@@ -564,6 +550,131 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+		g.Expect(c.Get(ctx, types.NamespacedName{Name: crdName}, crd)).To(gmg.Succeed())
+		var v1 apiextensionsv1.CustomResourceDefinitionVersion
+		for i, version := range crd.Spec.Versions {
+			if version.Name == "v1" {
+				crd.Spec.Versions[i].Storage = true
```
sbueringer, line 621: nit: Do we have to set Storage twice?

pkg/client/apiutil/restmapper_test.go
```diff
@@ -564,6 +550,131 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+		crd.Spec.Versions = []apiextensionsv1.CustomResourceDefinitionVersion{v1}
+		g.Expect(c.Update(ctx, crd)).To(gmg.Succeed())
+
+		// We wait until v2 is not available anymore.
+		g.Eventually(func(g gmg.Gomega) {
+			_, err = discClient.ServerResourcesForGroupVersion(group + "/v2")
+			g.Expect(apierrors.IsNotFound(err)).To(gmg.BeTrue(), "v2 should not be available anymore")
+		}).Should(gmg.Succeed())
+
+		// Although v2 is not available anymore, the cache is not invalidated yet so it should return a mapping.
```
sbueringer, line 635: Just a question. I didn't realize this before, but once the CAPI controllers are all running for a bit, they wouldn't hit the meta.IsNoMatchError case, right?

So this only happens if the apiVersion goes away at the wrong time and then (in CAPI) we do an external.Get on a resource that isn't already cached?

pkg/client/apiutil/restmapper_test.go
```diff
@@ -564,6 +550,131 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+		_, err = lazyRestMapper.RESTMapping(schema.GroupKind{Group: group, Kind: "Limo"})
+		g.Expect(err).To(beNoMatchError())
+		g.Expect(crt.GetRequestCount()).To(gmg.Equal(2))
+		crt.Reset()
+
+		// Now we request v2 again and it should return an error since the cache was invalidated.
+		// 	#1: GET https://host/apis/inventory.example.com/v2
+		_, err = lazyRestMapper.RESTMapping(schema.GroupKind{Group: group, Kind: kind}, "v2")
+		g.Expect(err).To(beNoMatchError())
+		g.Expect(crt.GetRequestCount()).To(gmg.Equal(1))
```
sbueringer, line 654: Should we do one more call to verify calling RESTMappings without versions set would work and only return v1?

(I think this is the case that we hit through apiutil.IsGVKNamespaced)

sbueringer (commented):
Nice work, thx!

pkg/client/apiutil/restmapper_test.go
g-gaston: Yeah. From capi's perspective, the error happens (only in v0.16.x) when a group is cached in m.apiGroups with a particular version that will be deleted. If after this is cached, we request a resource that for some reason is not yet in the underlying RESTMapper, that triggers a "re-read" of all the cached versions for the resources's group. If one of these cached versions have been deleted, it will return an error. This will keep happening forever since m.apigroups is not cleaned.

In v0.17.x this "re-read" for a non-existing version still happens every time there is a miss in the underlying RESTMapper, but the error is ignored, so the caller continues as normal.

pkg/client/apiutil/restmapper_test.go
g-gaston: That wouldn't happen with this version of CR since not found errors were ignored already. Added an extra call to make it explicit jic.

pkg/client/apiutil/restmapper_test.go
g-gaston: good point, I was just being lazy

I changed it to have v1 and v1alpha1, with v1 being the storage version.
Then I delete v1alpha1 and I expect v1 to be the only available version.

sbueringer: Thx!

pkg/client/apiutil/restmapper_test.go
sbueringer: Good point

pkg/client/apiutil/restmapper_test.go
sbueringer: Perfect, thx!

sbueringer:
Thank you very much!

/lgtm


PTAL :)

pkg/client/apiutil/restmapper.go
```diff
@@ -265,18 +270,26 @@ func (m *mapper) findAPIGroupByName(groupName string) (*metav1.APIGroup, error)
...
+// fetchGroupVersionResourcesLocked fetches the resources for the specified group and its versions.
+// This method might modify the cache so it needs to be called under the lock.
+func (m *mapper) fetchGroupVersionResourcesLocked(groupName string, versions ...string) (map[schema.GroupVersion]*metav1.APIResourceList, error) {
 	groupVersionResources := make(map[schema.GroupVersion]*metav1.APIResourceList)
 	failedGroups := make(map[schema.GroupVersion]error)
 
 	for _, version := range versions {
 		groupVersion := schema.GroupVersion{Group: groupName, Version: version}
 
 		apiResourceList, err := m.client.ServerResourcesForGroupVersion(groupVersion.String())
```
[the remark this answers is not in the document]
sbueringer: l.283 until 291 are handling the returned err of this call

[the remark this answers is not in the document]
vincepri: ahh, ok nevermind then!

vincepri (approved):
/approve

sbueringer:
As mentioned in the issue. We would like to backport this into v0.17 and v0.16. v0.16 requires a manual backport. Let's see if we get lucky with 0.17

/cherry-pick release-0.17

sbueringer:
Pushed the release-0.17 branch based on v0.17.0

PR #2664, opened 2024-01-24
🌱 adding SubjectAccessReview.authorization.k8s.io/v1 webhook support

everpeace:
This PR is the second try which adds SubjectAccessReview.authorization.k8s.io/v1 webhook support. The original PR is #1535 posted by @rfranzke, which is closed by stale long time ago.

After the long time, I'm now interested in this feature(https://github.com/kubernetes-sigs/controller-runtime/pull/1535#issuecomment-1908310685). Thus, this PR is based on #1535 and added some cleanups and migrations on the top of it.

cc/ @rfranzke @alvaroaleman @christopherhein @vincepri @estroz

Note: I would like to use this feature in implementing https://github.com/karmada-io/karmada/pull/4593

### DIFF
--- a/examples/subjectaccessreview/main.go
+++ b/examples/subjectaccessreview/main.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


alvaroaleman:
Hi, thanks for this PR. Controller-Runtime is a library for building controllers, so an authorization webhook does IMHO not really fit into it.

rfranzke:
> Hi, thanks for this PR. Controller-Runtime is a library for building controllers, so an authorization webhook does IMHO not really fit into it.

I somehow missed this comment, but thanks to @everpeace's activity, I now saw it. With this line of thought, why there is support for [authn webhooks](https://github.com/kubernetes-sigs/controller-runtime/tree/main/examples/tokenreview) then? Shouldn't this be deprecated and removed accordingly?

sbueringer:
It's always a lot harder to drop something and break existing users compared to not adding something

PR #2668, opened 2024-01-31
:seedling: WaitForDefaultNamespace while starting up envtest

### LINKED ISSUE #2626: Seeing flakey testing where `default` namespace doesn't exist
Currently, we are running `envtest` in Karpenter where we are starting up all of the binaries and then applying pods to the `default` namespace. Every now and then, we are seeing failures in our pod metric testing (seen here: https://github.com/kubernetes-sigs/karpenter/actions/runs/7251475446/job/19753909305?pr=885). When this happens, it always contains the same error

```bash
Pod Metrics [It] should update the pod state metrics
/home/runner/work/karpenter/karpenter/pkg/controllers/metrics/pod/suite_test.go:56

  [FAILED] Expected success, but got an error:
      <*errors.StatusError | 0xc000134aa0>: 
      namespaces "default" not found
      {
          ErrStatus: {
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "namespaces \"default\" not found",
              Reason: "NotFound",
              Details: {Name: "default", Group: "", Kind: "namespaces", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 404,
          },
      }
  In [It] at: /home/runner/work/karpenter/karpenter/pkg/controllers/metrics/pod/suite_test.go:58 @ 12/18/23 17:16:52.[71](https://github.com/kubernetes-sigs/karpenter/actions/runs/7251475446/job/19753909305?pr=885#step:5:72)9
```

This error indicates that the `default` namespace doesn't exist at the apiserver when we are applying the object; however, I would generally expect this namespace to exist once the apiserver starts up.

It's worth noting that we don't see this error in other testing where we are running `envtest.Environment` with the `CRDs` option set. I suspect that this is due to the fact that this gives the setup a little more time to populate this namespace at the apiserver before it returns. 

Is there a general recommendation here? Obviously we could wait for the `default` namespace to exist in our own code, but it would be nice if we could be assured that it would always exist when we start testing from upstream.

jonathan-innis:
Fixes #2626

This change adds a `waitForDefaultNamespace` call which polls the apiserver for the existence of the default namespace immediately after control plane startup. This is needed because we've seen that there can be flakes when leveraging `envtest` if you expect the `default` namespace to be there immediately after startup but the namespace hasn't actually propagated so that the apiserver can see it.

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -17,15 +17,20 @@ limitations under the License.
 package envtest
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
@@ -265,6 +270,12 @@ func (te *Environment) Start() (*rest.Config, error) {
 		te.Scheme = scheme.Scheme
 	}
 
+	// If we are bringing etcd up for the first time, it can take some time for the
+	// default namespace to actually be created and seen as available to the apiserver
+	if err := te.waitForDefaultNamespace(te.Config); err != nil {
+		return nil, fmt.Errorf("default namespace didn't register within deadline: %w", err)
+	}
+
 	// Call PrepWithoutInstalling to setup certificates first
 	// and have them available to patch CRD conversion webhook as well.
 	if err := te.WebhookInstallOptions.PrepWithoutInstalling(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/envtest/server.go
```diff
@@ -322,6 +333,20 @@ func (te *Environment) startControlPlane() error {
 	return nil
 }
 
+func (te *Environment) waitForDefaultNamespace(config *rest.Config) error {
+	cs, err := client.New(config, client.Options{})
+	if err != nil {
+		return fmt.Errorf("unable to create client: %w", err)
+	}
+	// It shouldn't take longer than 5s for the default namespace to be brought up in etcd
+	return wait.PollUntilContextTimeout(context.TODO(), time.Millisecond*500, time.Second*5, true, func(ctx context.Context) (bool, error) {
```
sbueringer, line 342:
```suggestion
	return wait.PollUntilContextTimeout(context.TODO(), time.Millisecond*50, time.Second*5, true, func(ctx context.Context) (bool, error) {
```

Maybe let's do this a bit quicker so we don't lose a full 500ms if the namespace doesn't show up immediately

jonathan-innis: Yep, makes sense. Updated.

pkg/envtest/server.go
```diff
@@ -334,12 +334,13 @@ func (te *Environment) startControlPlane() error {
 }
 
 func (te *Environment) waitForDefaultNamespace(config *rest.Config) error {
+	fmt.Println("waiting...")
```
sbueringer, line 337: If we want to add a log here. Let's please also write what we are waiting for and log the same way as in other places (e.g. l.299: `log.V(1).Info("installing webhooks")`)

jonathan-innis: Shoot. Yep, you're totally right. I don't even think that we want to log here. This was just an ad-how change that I forgot to remove before pushing

jonathan-innis: Updated.

sbueringer:
Thx!

/lgtm
/approve

Issue #2644, opened 2024-01-09
Support startup probes


ramperher:
Hi all! I was working in the update of some operators I'm building with [operator-sdk](https://github.com/operator-framework/operator-sdk), which is consuming from this project, and I was able to define liveness and readiness probes for the pods, feature already supported based on https://github.com/kubernetes-sigs/controller-runtime/issues/297.

However, I also wanted to introduce startup probes, which are also documented in [k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), but there's no reference in operator-sdk project for that. I evaluated the code to try to implement that, but then I discovered controller-runtime seems not to have support for that probe yet.

I would like to ask if it's feasible to introduce this feature in the controller-runtime, probably following the same work it was done [here](https://github.com/kubernetes-sigs/controller-runtime/pull/419). I can work on that if you agree on that. Thanks!

ramperher:
> What's the underlying use case we're trying to solve?

Hi @vincepri, currently, the use case where we want to introduce this feature is a dummy workload, called [example-cnf](https://github.com/openshift-kni/example-cnf), that our team is maintaining to define a well-developed CNF, according to [CNF best practices](https://github.com/test-network-function/cnf-best-practices-guide). In that guide, which is eventually taken as reference for [Red Hat CNF Certification](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.72/html/red_hat_software_certification_workflow_guide/con_cnf-certification_openshift-sw-cert-workflow-submitting-your-helm-chart-for-certification), there's a specific section for _Liveness, readiness and startup probes_, referencing to this [unit test](https://github.com/test-network-function/cnf-certification-test/blob/main/CATALOG.md#lifecycle-startup-probe) that is the one used to evaluate if a CNF has a startup probe defined.

So, as a summary, startup probes are needed then for passing one of the tests of this certification workflow without any issue, and currently, the operators built with operator-sdk (which is consuming from this repo) don't have this feature.

troy0820:
This seems to be on the pod of the workload and not the underlying controller runtime components. 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ shows how to do this on workloads along with liveness/readiness probes.

ramperher:
> This seems to be on the pod of the workload and not the underlying controller runtime components.
> 
[...]

PR #2669, opened 2024-01-31
✨ Add support for startup probes

ramperher:
This PR address this issue: https://github.com/kubernetes-sigs/controller-runtime/issues/2644, including the support for a separate endpoint for startup probe, defined in [k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/). Current code only support readiness and liveness probes, but doesn't include support for startup probe.

### DIFF
--- a/pkg/config/v1alpha1/types.go
+++ b/pkg/config/v1alpha1/types.go
@@ -135,6 +135,10 @@ type ControllerHealth struct {
 	// LivenessEndpointName, defaults to "healthz"
 	// +optional
 	LivenessEndpointName string `json:"livenessEndpointName,omitempty"`
+
+	// StartupEndpointName, defaults to "startz"
+	// +optional
+	StartupEndpointName string `json:"startupEndpointName,omitempty"`
 }
 
 // ControllerWebhook defines the webhook server for the controller.
--- a/designs/component-config.md
+++ b/designs/component-config.md
@@ -1,7 +1,7 @@
 # ComponentConfig Controller Runtime Support
 Author: @christopherhein
 
-Last Updated on: 03/02/2020
+Last Updated on: 31/01/2024
 
 ## Table of Contents
 
@@ -91,6 +91,7 @@ type ManagerConfiguration interface {
 
 	GetReadinessEndpointName() string
 	GetLivenessEndpointName() string
+	GetStartupEndpointName() string
 
 	GetPort() int
 	GetHost() string
@@ -161,6 +162,7 @@ type ControllerManagerConfigurationHealth struct {
 
 	ReadinessEndpointName string `json:"readinessEndpointName,omitempty"`
 	LivenessEndpointName  string `json:"livenessEndpointName,omitempty"`
+	StartupEndpointName  string `json:"startupEndpointName,omitempty"`
 }
 ```
 
--- a/designs/move-cluster-specific-code-out-of-manager.md
+++ b/designs/move-cluster-specific-code-out-of-manager.md
@@ -116,6 +116,9 @@ type Manager interface {
 	// AddReadyzCheck allows you to add Readyz checker
 	AddReadyzCheck(name string, check healthz.Checker) error
 
+	// AddStartzCheck allows you to add Startz checker
+	AddStartzCheck(name string, check healthz.Checker) error
+
 	// Start starts all registered Controllers and blocks until the Stop channel is closed.
 	// Returns an error if there is an error starting any controller.
 	// If LeaderElection is used, the binary must be exited immediately after this returns,
--- a/pkg/healthz/doc.go
+++ b/pkg/healthz/doc.go
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package healthz contains helpers from supporting liveness and readiness endpoints.
-// (often referred to as healthz and readyz, respectively).
+// Package healthz contains helpers from supporting liveness, readiness and startup endpoints.
+// (often referred to as healthz, readyz and startz, respectively).
 //
 // This package draws heavily from the apiserver's healthz package
 // ( https://github.com/kubernetes/apiserver/tree/master/pkg/server/healthz )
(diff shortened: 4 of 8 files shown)


pkg/config/v1alpha1/types.go
```diff
@@ -135,6 +135,10 @@ type ControllerHealth struct {
 	// LivenessEndpointName, defaults to "healthz"
 	// +optional
 	LivenessEndpointName string `json:"livenessEndpointName,omitempty"`
+
+	// StartupEndpointName, defaults to "startz"
+	// +optional
+	StartupEndpointName string `json:"startupEndpointName,omitempty"`
```
troy0820, line 141: I know this is where the other endpoint names live but this is deprecated.  Do we want to keep it here with the risk of it being moved and/or removed for configuration?

ramperher: I was not aware of this file being deprecated. Where should it be instead? I didn't find another place where the endpoint name is defined. I suppose that the risk for StartupEndpointName is the same than for the others, is it correct?

troy0820: >// ControllerHealth defines the health configs.
//
// Deprecated: The component config package has been deprecated and will be removed in a future release. Users should migrate to their own config implementation, please share feedback in https://github.com/kubernetes-sigs/controller-runtime/issues/895.

There is this PR that removes this file https://github.com/kubernetes-sigs/controller-runtime/pull/2648 because most of the config is deprecated.  Adding this functionality there doesn't seem like it will stay long as the plan would be to get rid of this configuration. 

Now that doesn't mean it doesn't have a place defined in this repo, but I would probably place it as an option and not as deprecated option within the

ramperher: Understood, thank you. However, from my point of view, isn't it better to remove this config from my change when https://github.com/kubernetes-sigs/controller-runtime/pull/2648 is merged? Because right now it's still in progress

After merging your change, I can rebase my change to bring your changes and then I can remove my code from the deprecated files.

troy0820: Yours may get merged first, so with that if the maintainers lgtm/approve, I can work on finding a spot for this to move it out of the deprecated struct.

alvaroaleman:
I still do not see what problem this solves other than some certification of yours. You are talking about use-cases without actually giving any concrete example of a use case that would require this. You can run a custom http server to server any endpoint of your choosing or even add custom endpoints to the webhook server, none of that requires changes in controller-runtime.

Before we merge anything like this, I want to see a specific example of a controller-manager that requires this.
/hold

ramperher:
> I still do not see what problem this solves other than some certification of yours. You are talking about use-cases without actually giving any concrete example of a use case that would require this. You can run a custom http server to server any endpoint of your choosing or even add custom endpoints to the webhook server, none of that requires changes in controller-runtime.
> 
> Before we merge anything like this, I want to see a specific example of a controller-manager that requires this. /hold

@alvaroaleman , just to clarify the rationale behind the work: this appeared as a consequence of working with some cases submitted for Red Hat CNF Certification, but this is not a fix for that. This work is to provide the capability of defining a custom startup probe, decoupling it from liveness/readiness probe, for [operator-sdk](https://github.com/operator-framework/operator-sdk/issues/6659)'s controller-manager pods. These pods, which manage the lifecycle of the resources deployed by the operator, rely on the controller-runtime code to be built, and currently, it's not possible to define a custom startup probe if needed, as documented in the issue.

In our case, right now, we are working in a Telco CNF called [example-cnf](https://github.com/openshift-kni/example-cnf), where the operators are made with operator-sdk, and in the controller-manager (example [here](https://github.com/openshift-kni/example-cnf/blob/main/testpmd-operator/config/manager/manager.yaml#L41-L58)), we have to point to `/healthz` endpoint for startup probe because there's no endpoint defined for startup probe, currently.

In the issue opened in  [operator-sdk](https://github.com/operator-framework/operator-sdk/issues/6659), the current status is just waiting for the resolution of this work in controller-runtime, then the endpoint would be added there.

After saying this, I'd also like to quote [k8s official docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#:~:text=The%20kubelet%20uses%20startup%20probes,interfere%20with%20the%20application%20startup.) regarding this feature:

_The kubelet uses startup probes to know when a container application has started. If such a probe is configured, liveness and readiness probes do not start until it succeeds, making sure those probes don't interfere with the application startup._

This means that the startup probe is the first one being executed. Liveness and readiness probes are not started till startup probe execution succeed. So, for me, it makes sense that there may be cases where a controller-manager pod from operator-sdk wants to control what happens just after starting the containers, using this startup probe feature, and managing it in a different way than liveness and readiness cases.

Hope this clarifies the purpose of this change.

ramperher:
> > we have to point to /healthz endpoint for startup probe because there's no endpoint defined for startup probe, currently.
> 
> And what is the problem with that? Please, start with the problem description. "It doesn't have a startupProbe enndpoint" is not a problem description, its jumping to a problem in an assumed solution of a problem that wasn't described.
> 
> You need to describe why you need the `startupProbe` in the first place. What is not working correctly when not using a startupProbe or using the healthz endpoint for it?



> > we have to point to /healthz endpoint for startup probe because there's no endpoint defined for startup probe, currently.
> 
> And what is the problem with that? Please, start with the problem description. "It doesn't have a startupProbe enndpoint" is not a problem description, its jumping to a problem in an assumed solution of a problem that wasn't described.
> 
> You need to describe why you need the `startupProbe` in the first place. What is not working correctly when not using a startupProbe or using the healthz endpoint for it?

As I can extract from k8s documentation, startup probe targets a different container lifecycle stage compared to readiness and liveness probes. My rationale for this change is, if the feature is different, shouldn't it have a different endpoint too?

ramperher:
I've been analyzing this KEP with @vincepri: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/950-liveness-probe-holdoff/README.md, and there, same endpoint is used for startup probe and liveness probe.

As I said, for me, it made sense to separate features in different endpoints, but, if by design, in this KEP, it was decided to do the opposite, I don't have much more arguments to justify my change.

So, having said that, I'll close the PR. Thanks for your reviews here.

Issue #2670, opened 2024-02-01
Create a separate registry for work-queue metrics


sonamkshenoy:
Hi all! 

We are using custom controllers to manage our custom resources, and wish to export the workqueue metrics to Prometheus.
This is done by using the [`workqueue.SetProvider()`](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go#L262) provided by `kubernetes.io/client-go`. 
When we try to register our metrics using this function in our `init()`, this doesn't take effect, as the `init()` function [being used](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/workqueue.go#L90) in the `metrics` package of this repo seems to making the call to `SetProvider()` first.

We can make use of this registry that you provide. It however comes with a [whole lot](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/client_go_adapter.go) of other metrics which we do not wish to export, such as - leader election, latency, request size, response size and other client metrics. And since they're `internal` we aren't able to unregister them.

Wonder if we can either:
1. [The best option in my opinion] Create a separate registry for just workqueue metrics in this repo and not add the other metrics to it, so that we can use it in our project. We can then combine all those metrics along with the non-workqueue related ones in a separate registry, thus unaffecting the previous behaviour.
2. Not have the `init()` function altogether in the [`metrics.go`](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go) file. Instead provide those functionalities as a separate function, so that we users can call it in our `init()`. This way, if we choose to, we can create our own metrics and call `SetProvider()` ourselves.
3. Any other option which might not have been covered above...
4. Or is there a way we can tackle this on our side without having to make any changes here?

I'm up for working on this issue.

troy0820:
/kind support feature 

> [The best option in my opinion] Create a separate registry for just workqueue metrics in this repo and not add the other metrics to it, so that we can use it in our project. We can then combine all those metrics along with the non-workqueue related ones in a separate registry, thus unaffecting the previous behaviour.

I think this is a reasonable request.  Do you mean that the defined metrics can then be registered to this "new" registry and pick the metrics you want to export rather than having all in one?  Such as workqueue, etc?  

And allow the user to register the metrics they want to expose?

sonamkshenoy:
Yep! 
Something like: https://github.com/kubernetes-sigs/controller-runtime/pull/2671

[...]

PR #2671, opened 2024-02-01
✨ Add a separate registry for workqueue metrics

sonamkshenoy:
This PR addresses #2670 

A separate registry has been created for work queue metrics, so that we can import just this registry directly as `metrics.WorkqueueRegistry` without getting loads of other unrelated metrics.
This retains the previous behaviour as all the metrics are still available in the main `Registry`.

### DIFF
--- a/pkg/metrics/workqueue.go
+++ b/pkg/metrics/workqueue.go
@@ -87,14 +87,18 @@ var (
 	}, []string{"name"})
 )
 
+var WorkqueueRegistry = prometheus.NewRegistry()
+
 func init() {
-	Registry.MustRegister(depth)
-	Registry.MustRegister(adds)
-	Registry.MustRegister(latency)
-	Registry.MustRegister(workDuration)
-	Registry.MustRegister(unfinished)
-	Registry.MustRegister(longestRunningProcessor)
-	Registry.MustRegister(retries)
+	WorkqueueRegistry.MustRegister(depth)
+	WorkqueueRegistry.MustRegister(adds)
+	WorkqueueRegistry.MustRegister(latency)
+	WorkqueueRegistry.MustRegister(workDuration)
+	WorkqueueRegistry.MustRegister(unfinished)
+	WorkqueueRegistry.MustRegister(longestRunningProcessor)
+	WorkqueueRegistry.MustRegister(retries)
+
+	Registry.Register(WorkqueueRegistry)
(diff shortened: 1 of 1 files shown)


sonamkshenoy:
Hey @troy0820 

Yep, I need just the work queue related metrics in my application. So, in this case (which implements the approach 1), I could have these metrics directly as `metrics.WorkqueueRegistry`. And I don't need the `metrics.Registry` anymore.

> If what you need are the metrics defined by this package, to have it be provided to the new registry, is there a way we can do this on the consumer side where you can push them to where ever you want?

I'm not sure there is a way I can do this? The init() function in this package gets called before the init() function in our application as it is being imported (and probably called first). Thus, even if we create our metrics, we can't register them with the work queue, unless remove the `SetProvider()` function in the init() of this file.

pkg/metrics/workqueue.go
```diff
@@ -87,14 +87,18 @@ var (
...
-	Registry.MustRegister(retries)
+	WorkqueueRegistry.MustRegister(depth)
+	WorkqueueRegistry.MustRegister(adds)
+	WorkqueueRegistry.MustRegister(latency)
+	WorkqueueRegistry.MustRegister(workDuration)
+	WorkqueueRegistry.MustRegister(unfinished)
+	WorkqueueRegistry.MustRegister(longestRunningProcessor)
+	WorkqueueRegistry.MustRegister(retries)
+
+	Registry.Register(WorkqueueRegistry)
```
alvaroaleman, line 101: I understand the problem, but this means that the workqueue metrics will disappear for all existing users. That is not an option. We need to find a solution that does not break existing users.

/hold

sonamkshenoy: Hey @alvaroaleman , these metrics won't disappear actually. We are adding those metrics back to the main `Registry` with `Registry.Register(WorkqueueRegistry)`. So this won't affect anything for existing users.

sonamkshenoy:
Hey @troy0820 @alvaroaleman 

There are 2 more options I could think of:
1. Just capitalise the metrics in this file `workqueue.go`, so that they can be exposed for use to users.
2. Functionality for users to register metrics of their choice as in https://github.com/kubernetes-sigs/controller-runtime/pull/2672. I guess this is what @troy0820  was referring to?

PR #2673, opened 2024-02-02
✨ custom equalities in CreateOrUpdate

tareksha:
Currently `CreateOrUpdate` always uses `equalities.Semantic` to decide whether an update is needed after mutating the existing object. Sometimes custom behavior is needed, for example when some fields are protobuf types (proto.Message).

Add `CreateOrUpdateWithEqualities` to allow passing a custom `Equalities` for performing the deepEqual comparison. The existing `CreateOrUpdate` continues to use `equalities.Semantic`.

Without this fix CreateOrUpdate always updates resources that contain fields with types that Semantic does not support as-is, even when not needed

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -25,6 +25,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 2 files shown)


alvaroaleman:
Do you have a concrete example where the `equalities.Semantic` doesn't work?

tareksha:
> Do you have a concrete example where the `equalities.Semantic` doesn't work?

sure. take istio's `istio.io/client-go` for example. it wraps protobuf types in each CRD's spec, all taken from the general `istio.io/api`.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -274,6 +275,11 @@ const ( // They should complete the sentence "Deployment default/foo has been ..
 // Note: changes made by MutateFn to any sub-resource (status...), will be
 // discarded.
 func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f MutateFn) (OperationResult, error) {
+	return CreateOrUpdateWithEqualities(ctx, c, obj, equality.Semantic, f)
+}
+
+// CreateOrUpdateWithEqualities is like CreateOrUpdate but allows customizing the equality check.
+func CreateOrUpdateWithEqualities(ctx context.Context, c client.Client, obj client.Object, eq conversion.Equalities, f MutateFn) (OperationResult, error) {
```
alvaroaleman, line 282: Alternatively maybe add a variadic `opts` arg to `CreateOrUpdate` that allows setting a custom comparison?

Also, wouldn't it be better to use a function type for this rather than a single-method interface, given that functions can be implemented inline and interfaces can not?

PR #2685, opened 2024-02-13
✨Move Source implementations to a single folder

inteon:
Moves all the source logic to the same directory, making it easier to understand and refactor in the future.
Currently `pkg/source/source.go` contains the Channel, Informer and Func source, while the Kind source lives in the `pkg/source/internal/kind.go` file.


inteon:
@alvaroaleman could you take a look at this PR too?
It should be purely code cleanup/ re-organisation.

alvaroaleman (changes requested):
> Moves all the source logic to the same directory, making it easier to understand and refactor in the future.

What is or isn't easier to understand in terms of file layout extremely subjective while not mattering too much in practise due to IDE tooling. At the same time a change like this has a cost in review capacity and in that it breaks git blame. Given that there is no clear benefit but a clear cost I'm -1 to this.

inteon:
> > Moves all the source logic to the same directory, making it easier to understand and refactor in the future.
> 
> What is or isn't easier to understand in terms of file layout extremely subjective while not mattering too much in practise due to IDE tooling. At the same time a change like this has a cost in review capacity and in that it breaks git blame. Given that there is no clear benefit but a clear cost I'm -1 to this.

Thanks for the feedback. I updated https://github.com/kubernetes-sigs/controller-runtime/pull/2686 so it does no longer depend on this PR.

PR #2686, opened 2024-02-13
⚠Only allow sources to start once

inteon:
Currently, only the Informer and Channel source can be started more than once (Start function can be called twice on same object).
I do believe however that this pattern is flawed, each source should only be started once, so that the context passed to that Start function is only used to stop that one instance from running and all the properties in the source belong to that single instance. The Channel source, for example, currently stops all sources that were started from the same instance when the first context passed to Start gets canceled (see https://github.com/kubernetes-sigs/controller-runtime/pull/2686#discussion_r1575306403).

For the Channel source, allowing the source to start only once, requires the API to be altered, since we want to make it possible to listen to the same channel using multiple Channel sources (matching the current behavior). This requires a new Channel Broadcaster to be introduced.

### DIFF
--- a/pkg/source/source.go
+++ b/pkg/source/source.go
@@ -23,14 +23,13 @@ import (
 	"sync"
 
 	"k8s.io/client-go/util/workqueue"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	internal "sigs.k8s.io/controller-runtime/pkg/internal/source"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // Source is a source of events (e.g. Create, Update, Delete operations on Kubernetes Objects, Webhook callbacks, etc)
@@ -53,6 +52,9 @@ type SyncingSource interface {
 	Source
 	WaitForSync(ctx context.Context) error
 }
+
+var _ Source = &internal.Kind[client.Object]{}
+var _ SyncingSource = &internal.Kind[client.Object]{}
 
 // Kind creates a KindSource with the given cache provider.
 func Kind[T client.Object](cache cache.Cache, object T, handler handler.TypedEventHandler[T], predicates ...predicate.TypedPredicate[T]) SyncingSource {
@@ -80,17 +82,18 @@ func WithPredicates[T any](p ...predicate.TypedPredicate[T]) ChannelOpt[T] {
 // default, the buffer size is 1024.
 func WithBufferSize[T any](bufferSize int) ChannelOpt[T] {
 	return func(c *channel[T]) {
-		c.bufferSize = &bufferSize
+		c.bufferSize = bufferSize
 	}
 }
 
 // Channel is used to provide a source of events originating outside the cluster
 // (e.g. GitHub Webhook callback).  Channel requires the user to wire the external
 // source (e.g. http handler) to write GenericEvents to the underlying channel.
-func Channel[T any](source <-chan event.TypedGenericEvent[T], handler handler.TypedEventHandler[T], opts ...ChannelOpt[T]) Source {
+func Channel[T any](broadcaster *channelBroadcaster[T], handler handler.TypedEventHandler[T], opts ...ChannelOpt[T]) Source {
 	c := &channel[T]{
-		source:  source,
-		handler: handler,
+		broadcaster: broadcaster,
+		handler:     handler,
+		bufferSize:  1024,
 	}
 	for _, opt := range opts {
 		opt(c)
@@ -99,24 +102,21 @@ func Channel[T any](source <-chan event.TypedGenericEvent[T], handler handler.Ty
 	return c
 }
 
-type channel[T any] struct {
-	// once ensures the event distribution goroutine will be performed only once
-	once sync.Once
+var _ Source = &channel[string]{}
 
-	// source is the source channel to fetch GenericEvents
-	source <-chan event.TypedGenericEvent[T]
+type channel[T any] struct {
+	// broadcaster contains the source channel for events.
+	broadcaster *channelBroadcaster[T]
 
 	handler handler.TypedEventHandler[T]
 
 	predicates []predicate.TypedPredicate[T]
 
-	bufferSize *int
+	bufferSize int
 
-	// dest is the destination channels of the added event handlers
-	dest []chan event.TypedGenericEvent[T]
-
-	// destLock is to ensure the destination channels are safely added/removed
-	destLock sync.Mutex
+	mu sync.Mutex
+	// isStarted is true if the source has been started. A source can only be started once.
+	isStarted bool
 }
 
 func (cs *channel[T]) String() string {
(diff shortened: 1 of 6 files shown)


pkg/source/source.go
```diff
@@ -64,200 +60,68 @@ func Kind[T client.Object](cache cache.Cache, object T, handler handler.TypedEve
...
-
-	dst := make(chan event.TypedGenericEvent[T], *cs.bufferSize)
-
-	cs.destLock.Lock()
-	cs.dest = append(cs.dest, dst)
-	cs.destLock.Unlock()
-
-	cs.once.Do(func() {
-		// Distribute GenericEvents to all EventHandler / Queue pairs Watching this source
-		go cs.syncLoop(ctx)
```
inteon, line 151: This is the main issue that this PR solves: here, the context passed to the first call of Start is used to manage the goroutine that is used for all future calls to Start.

sbueringer:
I'm really having some trouble understanding what the goal for users of controller-runtime is with this PR.

inteon:
@sbueringer Below is a small example test to illustrate the issue at hand.
I start two controllers which both listen to the same channel.
If the first controller is stopped (context passed to Start is canceled), the second controller also stops receiving events.

On master:
```go
func TestMultiStart(t *testing.T) {
	rootCtx := context.Background()

	watchChan := make(chan event.GenericEvent, 10)
	watch := source.Channel(watchChan, &handler.EnqueueRequestForObject{})

	m, err := manager.New(&rest.Config{}, manager.Options{})
	require.NoError(t, err)

	// Controller 1 is started and stopped directly
	{
		c1, err := controller.New("controller-1", m, controller.Options{Reconciler: reconcile.Func(func(ctx context.Context, r reconcile.Request) (reconcile.Result, error) {
			t.Fail() // Reconcilder should not get called

		})})
		require.NoError(t, err)
		require.NoError(t, c1.Watch(watch))
		ctx1, cancel1 := context.WithCancel(rootCtx)
		cancel1()
		require.NoError(t, c1.Start(ctx1))
	}

	// Controller 2 is listening on the same channel
	callCount := int32(0)
	c2, err := controller.New("controller-2", m, controller.Options{Reconciler: reconcile.Func(func(ctx context.Context, o reconcile.Request) (reconcile.Result, error) {
		t.Log("got event!", atomic.AddInt32(&callCount, 1))

	})})
	require.NoError(t, err)
	require.NoError(t, c2.Watch(watch))

	ctx2, stop := context.WithTimeout(rootCtx, time.Second*5)
	defer stop()
	done := make(chan struct{})
	go func() {
		defer close(done)
		require.NoError(t, c2.Start(ctx2))
	}()

	// Start sending messages on the channel
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test1"}}}
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test2"}}}
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test3"}}}

	<-done

	// We waited 5 seconds, which should be enough to have at least received
	// one of the events on the channel
	require.GreaterOrEqual(t, callCount, int32(1))
}
```
```bash
$ go test -count=5 -timeout 30s -run ^TestMultiStart$ sigs.k8s.io/controller-runtime
--- FAIL: TestMultiStart (5.00s)
    aa_test.go:71: 
                Error Trace:    /home/tramlot/projects/controller-runtime/aa_test.go:71
                Error:          "0" is not greater than or equal to "1"
                Test:           TestMultiStart
--- FAIL: TestMultiStart (5.00s)
    aa_test.go:71: 
                Error Trace:    /home/tramlot/projects/controller-runtime/aa_test.go:71
                Error:          "0" is not greater than or equal to "1"
                Test:           TestMultiStart
--- FAIL: TestMultiStart (5.00s)
    aa_test.go:71: 
                Error Trace:    /home/tramlot/projects/controller-runtime/aa_test.go:71
                Error:          "0" is not greater than or equal to "1"
                Test:           TestMultiStart
--- FAIL: TestMultiStart (5.00s)
    aa_test.go:71: 
                Error Trace:    /home/tramlot/projects/controller-runtime/aa_test.go:71
                Error:          "0" is not greater than or equal to "1"
                Test:           TestMultiStart
--- FAIL: TestMultiStart (5.00s)
    aa_test.go:71: 
                Error Trace:    /home/tramlot/projects/controller-runtime/aa_test.go:71
                Error:          "0" is not greater than or equal to "1"
                Test:           TestMultiStart
FAIL
FAIL    sigs.k8s.io/controller-runtime  25.019s
FAIL
```

With this PR:
```go
func TestMultiStart(t *testing.T) {
	rootCtx := context.Background()

	watchChan := make(chan event.GenericEvent, 10)
	broadcaster := source.NewChannelBroadcaster(watchChan)
	watch1 := source.Channel(broadcaster, &handler.EnqueueRequestForObject{})
	watch2 := source.Channel(broadcaster, &handler.EnqueueRequestForObject{})

	m, err := manager.New(&rest.Config{}, manager.Options{})
	require.NoError(t, err)

	// Controller 1 is started and stopped directly
	{
		c1, err := controller.New("controller-1", m, controller.Options{Reconciler: reconcile.Func(func(ctx context.Context, r reconcile.Request) (reconcile.Result, error) {
			t.Fail() // Reconcilder should not get called

		})})
		require.NoError(t, err)
		require.NoError(t, c1.Watch(watch1))
		ctx1, cancel1 := context.WithCancel(rootCtx)
		cancel1()
		require.NoError(t, c1.Start(ctx1))
	}

	// Controller 2 is listening on the same channel
	callCount := int32(0)
	c2, err := controller.New("controller-2", m, controller.Options{Reconciler: reconcile.Func(func(ctx context.Context, o reconcile.Request) (reconcile.Result, error) {
		t.Log("got event!", atomic.AddInt32(&callCount, 1))

	})})
	require.NoError(t, err)
	require.NoError(t, c2.Watch(watch2))

	ctx2, stop := context.WithTimeout(rootCtx, time.Second*5)
	defer stop()
	done := make(chan struct{})
	go func() {
		defer close(done)
		require.NoError(t, c2.Start(ctx2))
	}()

	// Start sending messages on the channel
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test1"}}}
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test2"}}}
	watchChan <- event.GenericEvent{Object: &corev1.Pod{ObjectMeta: v1.ObjectMeta{Name: "Test3"}}}

	<-done

	// We waited 5 seconds, which should be enough to have at least received
	// one of the events on the channel
	require.GreaterOrEqual(t, callCount, int32(1))
}
```
```bash
$ go test -count=5 -timeout 30s -run ^TestMultiStart$ sigs.k8s.io/controller-runtime
ok      sigs.k8s.io/controller-runtime  25.017s
```

sbueringer:
I get the point that calling Start twice on the same Channel is problematic because of the strange behavior on cancel.

I'm not sure if that necessarily means that we should have the channelBroadcaster implementation within controller-runtime.

inteon:
> I get the point that calling Start twice on the same Channel is problematic because of the strange behavior on cancel.
> 
> I'm not sure if that necessarily means that we should have the channelBroadcaster implementation within controller-runtime to support having multiple Sources on the same chan.

We can also say that 1 channel source = 1 channel.
However, that means we drop the broadcasting feature that we have today.

sbueringer:
> We can also say that 1 channel source = 1 channel.
> However, that means we drop the broadcasting feature that we have today.

I'm not sure if that is/was an intended feature or it just mostly worked (- the cancel behavior of the context).
(I just don't have the history on that to be honest).

Let's see what @alvaroaleman thinks.

alvaroaleman:
> I'm not sure if that is/was an intended feature or it just mostly worked (minus the cancel behavior of the context).
(I just don't have the history on that to be honest).

I don't know either if this was intentional, but the cancel behavior means that this never worked correctly. This change means that the UX of `source.Channel` gets worse for everyone to support something that never worked properly to begin with. If someone wants this (and we don't know if that is the case), they should multiplex a single source channel onto multiple target channels themselves.

PR #2687, opened 2024-02-14
🐛 Fix lazy rest mapper cache invalidation

g-gaston:
This fixes a bug introduced by my previous PR https://github.com/kubernetes-sigs/controller-runtime/pull/2663

When a group version is not found, if the group version is cached in apiGroups but not cached in knownGroups, the cache is not invalidated. Moreover and even worse, in that scenario an error is returned.

This is only an issue under the very rare case where a version has been removed during a call just after caching in apiGroups but before it's cached in knownGroups.

After the change now an error is never returned for not found versions no matter if the version was cached or not. And now the two different caches are checked and invalidated independently.


sbueringer:
Perfect. Thank you very much!

/lgtm
/approve

sbueringer:
@g-gaston Do you have some time to open cherry-pick PRs for release-0.17 and release-0.16?

PR #2693, opened 2024-02-22
⚠ Bump k8s.io/ to v1.30.0-beta.0

twz123:
Kubernetes 1.30 has a breaking API change around leader election metrics, introduced in kubernetes/kubernetes#122069. Not sure if there's a trick to fixing this in a way that's compatible with both 1.29 and 1.30 at the same time. Not sure when controller-runtime wants to jump to k8s 1.30, so leaving this as a draft.


sbueringer:
@twz123 Do you have time to rebase to v1.30.0-beta.0? In general I would bump to v1.30 now for early feedback.

I'm not sure I understand what the breaking change around leader election metrics is

twz123:
Is there some way to let Prow use Go 1.22? Kubernetes 1.30 [requires](https://github.com/kubernetes/api/blob/v0.30.0-beta.0/go.mod#L5) it.

sbueringer:
> Is there some way to let Prow use Go 1.22? Kubernetes 1.30 [requires](https://github.com/kubernetes/api/blob/v0.30.0-beta.0/go.mod#L5) it.

Yes, you have to bump the two images here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-main.yaml

Can you open a PR for that? (then I can approve)

go.mod
```diff
@@ -1,8 +1,8 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.21
+go 1.22.0
 
-toolchain go1.21.0
+toolchain go1.22.1
```
sbueringer, line 5: Do you know what the exact implications are of using go 1.22.0 and toolchain go1.22.1?

(specifically the exact patch version in each)

twz123: According to https://go.dev/doc/toolchain#config, IIUC, the `go' directive specifies the minimum required version, while the `toolchain' directive specifies the suggested version. Perhaps the toolchain directive could be removed altogether here?

twz123:
There you go:
* kubernetes/test-infra#32265

go.mod
sbueringer: Maybe. Is it automatically added by `go mod tidy`?

twz123: I've seen this added automatically sometimes, but I haven't caught the pattern yet. It definitely won't be added by `go mod tidy` if everything is in sync, at least not for me.

sbueringer: Did a bit of reserach:

* if `go` and `toolchain` use the same version, `go mod tidy` will remove `toolchain` (i.e. it's only preserved if the versions are different)
* we introduced `toolchain` when bumping to Go 1.21. I believe either this was added because we used `1.21` (instead of `1.21.0`) for the go directive or because the cleanup in `go mod tidy` wasn't implemented yet.

Kubernetes today only sets the Go directive to `1.22.0` and omits the `toolchain` directive: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L9

I would like to follow this pattern for now, if we don't have a good reason to set the `toolchain` directive as well.

sbueringer: This one is interesting

> Commands that incorporate new module versions that require new Go versions write the new minimum go version requirement to the current workspace’s go.work file or the main module’s go.mod file, updating the go line. For [repeatability](https://research.swtch.com/vgo-principles#repeatability), any command that updates the go line also updates the toolchain line to record its own toolchain name. The next time the go command runs in that workspace or module, it will use that updated toolchain line during [toolchain selection](https://go.dev/doc/toolchain#select).

sbueringer:
Thank you!

/lgtm
/approve

pmalek:
Can we have this in `release-0.17` to prevent

> Error: ../../../go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/metrics/leaderelection.go:26:71: undefined: leaderelection.SwitchMetric

with recent k8s 1.30 packages?

sbueringer:
No because this would make CR v0.17.x incompatible with client-go v0.29

pmalek:
ACK. When do we expect to ship 0.18?

sbueringer:
Current plan is ~ this week. Can't guarantee it though.

Mostly depends on #2783 as far as I can tell

PR #2698, opened 2024-02-29
🌱 Make event handler type aware

Danil-Grigorev:
This ensures that type casting check executed inside `source.Kind` logic guaranties type safety for propagated objects.

Currently the change will only have effect in `builder.WatchesRawSource`, when the source.Kind is constructed externally.

For such (raw) typed source, `OnlyMetadata` option is no longer doing anything, so this forces user to pass a fully qualified partial object metadata type during construction. This also supposed to be correct way to do it:
```go
pod := &metav1.PartialObjectMetadata{}
pod.SetGroupVersionKind(schema.GroupVersionKind{
	Group:   "",
	Version: "v1",
	Kind:    "Pod",
})
builder.WatchesRawSource(source.Kind(cache, pod), eventHandler, opts...)
```

### DIFF
--- a/pkg/builder/controller.go
+++ b/pkg/builder/controller.go
@@ -135,7 +135,7 @@ type WatchesInput struct {
 // This is the equivalent of calling
 // WatchesRawSource(source.Kind(cache, object), eventHandler, opts...).
 func (blder *Builder) Watches(object client.Object, eventHandler handler.EventHandler, opts ...WatchesOption) *Builder {
-	src := source.Kind(blder.mgr.GetCache(), object)
+	src := source.ObjectKind(blder.mgr.GetCache(), object)
 	return blder.WatchesRawSource(src, eventHandler, opts...)
 }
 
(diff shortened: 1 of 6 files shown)


alvaroaleman (commented):
/hold

This topic needs a wider discussion. If this doesn't provide proper typesafety for the whole source/predicate/handler chain, there is no value in doing this in the first place.

The only externally-visible change is this:
```
func ObjectKind[T client.Object](cache cache.Cache, object T) SyncingSource {
	return &internal.Kind[T]{Type: object, Cache: cache}
}
```

I don't see how this is any better than what we already have.

Danil-Grigorev:
@alvaroaleman When you are adding predicate on the type, you need to typecast from `client.Object`. Anything can be inside this interface, so you need to error check. Even if I expect a v1.Pod there, still, I may receive some other version of v1.Pod, which would cause panic if I don’t check safely for a type. Whatever you wrap in source.ObjectKind is then strictly checked on event handler level so possibility of that is eliminated. 

This change is the first step which allows than to include type info into predicates and make it reliable later. Also it is small and does not introduce API conflicts, which I believe is a reason to merge it and get into release earlier.

I can extend this PR and introduce types in other parts, but it would be much harder to review, and likely will not go without conflicts. That depends on how conflicts will be accepted in such refactoring context. I’m not against doing it all at once, anyway.

alvaroaleman:
> This change is the first step which allows than to include type info into predicates and make it reliable later. 

There is no point in a first step if we don't know if the direction is right. We have to have a plan how this will look in the end, we can't just do random changes that combined might or might not lead us somewhere.

PR #2702, opened 2024-03-07
⚠ Client Server side apply 

troy0820:
This is an attempt at server side apply being applied to the `client.Writer` interface.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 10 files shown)


troy0820:
This feels wrong, but may be right? Any feedback is definitely appreciated. I do know that the test failed on core object without the type meta because I didn't get the gvk of the client.Object before converting, however this isn't tested against a CRD.  

Also, not sure if we want to allow applyconfigurations or just deal with the unstructured parts we are tryin to "apply".  

Didn't want to try to resolve this myself as this may need discussion.

pkg/client/client.go
```diff
@@ -347,6 +348,21 @@ func (c *client) Patch(ctx context.Context, obj Object, patch Patch, opts ...Pat
 	}
 }
 
+func (c *client) Apply(ctx context.Context, obj Object, fieldOwner string) error {
```
alvaroaleman, line 351: Pretty heavily opposed to do anything in here before https://github.com/kubernetes/kubernetes/issues/118138 is done because:
* The only type that can be used in here is `unstructured.Unstructured`
* The converting as you do it down below doesn't fix the issue that the normal types end up putting zero values in fields
* Which is the reason why applytypes exist, but you can not actually use applytypes here, because they do not fulfill `client.Object`
/hold

troy0820: Definitely hold, and no strong opinions on the feedback beyond understanding the gaps that exists. Waiting on this make sense and provided the support going forward, strategy around this seems crucial.  Here for that conversation.

alvaroaleman: The whole topic boils down to "someone needs to do the work":
1. The upstream issue I mentioned needs fixing, upstream signaled that they are ok with the idea but no one has shown up to actually put the work in
2. This PR needs to be finished: https://github.com/kubernetes-sigs/controller-tools/pull/818
3. A change like this in controller-runtime that adds an `Apply` method that only accepts apply types

troy0820:
I can close this PR but I wanted to track the work that's necessary and the conversation from this review.  The apply types that are needed from controller-tools still has a  WIP PR out, and the method must accept apply types.  

/remove-lifecycle rotten

troy0820:
/remove-lifecycle stale

This is still in draft but being used for discussion on implementation.

alvaroaleman:
Superseded by https://github.com/kubernetes-sigs/controller-runtime/pull/3253

PR #2705, opened 2024-03-13
:seedling: Compare labels and annotations using maps.Equal

avorima:
I didn't benchmark this or anything, but thought that using this was better than reflect.

The maps package was introduced in go 1.21.


alvaroaleman:
thanks!

PR #2709, opened 2024-03-15
🌱 Remove outdated testing framework approvers

vincepri:

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -25,12 +25,6 @@ aliases:
   - varshaprasad96
   - inteon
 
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -25,12 +25,6 @@ aliases:
   - varshaprasad96
   - inteon
 
-  # folks to can approve things in the directly-ported
```
sbueringer, line 28: @vincepri Should we get rid of the OWNERS file as well?

Doesn't make much sense if we don't use it

PR #2710, opened 2024-03-15
🌱 Use go-install for versioned dependencies

vincepri:


hack/tools/go.mod
```diff
@@ -3,59 +3,3 @@ module sigs.k8s.io/controller-runtime/hack/tools
 go 1.21
 
 toolchain go1.21.5
-
```
troy0820, line 6: Do we even need this anymore?  I feel like if there is nothing here we could delete the go.mod and go.sum

vincepri: IIRC we do need it for go-install to work in this direcotry

PR #2715, opened 2024-03-18
🌱 Improve github actions dependencies versions and permissions

vincepri:

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -2,6 +2,9 @@ on:
   pull_request_target:
     types: [opened, edited, reopened, synchronize]
 
+permissions:
+  checks: write # Allow access to checks to write check runs.
+
 jobs:
   verify:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/verify.yml
```diff
@@ -2,13 +2,16 @@ on:
...
+
 jobs:
   verify:
     runs-on: ubuntu-latest
     name: verify PR contents
     steps:
     - name: Verifier action
       id: verifier
-      uses: kubernetes-sigs/kubebuilder-release-tools@v0.4.3
+      uses: kubernetes-sigs/kubebuilder-release-tools@012269a88fa4c034a0acf1ba84c26b195c0dbab4 # tag=v0.4.3
```
alvaroaleman, line 15: Why not use the tags?

vincepri: These are pinned dependencies, to a hash. Versions might change the hash and that on its own has general security implications.

When updates are available, dependabot updates both the hash and the tag comment

alvaroaleman (approved):

.github/workflows/verify.yml
sbueringer: A while back dependabot didn't support updating tag comments, but since that's supported we've been using this in CAPI and it works perfectly

Issue #2720, opened 2024-03-22
`go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest` fails with ` go.mod:3: invalid go version '1.22.0': must match format 1.23`


kgoutham93:
Installing latest version of setup-envtest fails with the following error log.

```
go: downloading sigs.k8s.io/controller-runtime v0.17.2
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240322105421-affb96708000
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest (in sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240322105421-affb96708000): go.mod:3: invalid go version '1.22.0': must match format 1.23
```
Go version
```
$ go version
go version go1.20.7 linux/amd64
```

Could be related to https://github.com/golang/go/issues/61888 ?

fspaniol:
I guess it might be related to: https://github.com/kubernetes-sigs/controller-runtime/commit/4c2442e4d743d172a2e0126a841ecd274fffc228#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R3

Previously: 1.21, now 1.22.0

fspaniol:
> > I tried to open a quick fix for it, but unfortunately I don't have time to do to it properly and thus I closed it. 😓
> 
> @fspaniol So I assume just dropping the patch version wasn't enough? Did you find a problem with that?

Hey @sbueringer , I managed to find the time to go a bit deeper on this and opened a new PR with some insights: https://github.com/kubernetes-sigs/controller-runtime/pull/2722

karlkfi:
If `setup-envtest` had git tags, that might provide a workaround. Currently, if you try to `go get sigs.k8s.io/controller-runtime/tools/setup-envtest` you get an auto-generated version number, because it's not taking the parent directory's version.

Alternatively, using go workspaces might allow the parent tag to work for child modules.

I ran into a similar problem trying to use setup-envtest@latest, which now requires Go 1.22. Not having tags makes it hard to pin to an old version. I ended up using `go get sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.16` to puin to an old version.

sbueringer:
Thx for taking a closer look!

I think it's important that we have a go.mod file which won't be updated after running go mod tidy (or other go commands). 

[...]

PR #2722, opened 2024-03-22
🐛 fix go versions in go modules

fspaniol:
Related to https://github.com/kubernetes-sigs/controller-runtime/issues/2720.

I have tried `go get` locally and I was getting the following errors before:

```
invalid go version '1.22.0': must match format 1.23
```

With this change, this is solved.

To provide more context on to why this happens:

I have a local setup where I use go 1.20 but however was installing `envtest` using the latest reference.

After opening this PR, as a mean of trying out, I upgraded my local Go version to 1.22, and `go get` worked with the current HEAD version. And if I execute it with this PR's state, it even changed it to `1.22.0` again:

```
go get .
go: upgraded go 1.22 => 1.22.0
go: added toolchain go1.22.1
```

Now, for whomever has older Go versions, using the current main will not work, but on the other hand, if we merge this, it will always modify the go.mod for people with newer go versions. I'll defer to the maintainers on what's the wanted approach here.

As I said, in my project I switched from `@latest` to `@release-0.16` and I can move on.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.22.0
+go 1.22
 
 require (
 	github.com/evanphx/json-patch v4.12.0+incompatible // Using v4 to match upstream
--- a/tools/setup-envtest/go.mod
+++ b/tools/setup-envtest/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime/tools/setup-envtest
 
-go 1.22.0
+go 1.22
 
 require (
 	github.com/go-logr/logr v1.2.4


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.22.0
+go 1.22
```
sbueringer, line 3: @fspaniol Can you try to add additional `toolchain 1.22.0` statements in both go files? I think that might solve the issue for both cases

sbueringer, line 3: To be honest. Not sure what we should do here. We won't be able to get CI green if go mod tidy changes the go mod file. I also don't want to check in a go.mod file that is updated as soon as a go command is run

sbueringer: Responded on the issue: https://github.com/kubernetes-sigs/controller-runtime/issues/2720#issuecomment-2016461180 (just saw the comments there now)

sbueringer:
/close

based on https://github.com/kubernetes-sigs/controller-runtime/issues/2720#issuecomment-2016461180

Feel free to reopen if you have an idea how else we could resolve this :)

Issue #2719, opened 2024-03-20
When a non-leader shuts down, leader election runnables are started


pleshakov:
Hi folks,

Noticed the following behavior.

When a non-leader manager shuts down, it starts leader election runnables.

In our project, we see the following in the logs

```
(this manager is a non-leader)
I0320 19:27:59.274077       7 leaderelection.go:250] attempting to acquire leader lease nginx-gateway/ngf-nginx-gateway-fabric-leader-election...


{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"Stopping and waiting for leader election runnables"}

(Next 5 lines are from the jobs started by leader election runnables)

{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"statusUpdater","msg":"Writing last statuses"}
{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"statusUpdater","msg":"Updating Gateway API statuses"}
{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"statusUpdater","msg":"Updating Nginx Gateway status"}
{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"telemetryJob","msg":"Starting cronjob"}
{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"telemetryJob","msg":"Stopping cronjob"}

{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"Stopping and waiting for caches"}
{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"Stopping and waiting for webhooks"}
{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"Stopping and waiting for HTTP servers"}
{"level":"info","ts":"2024-03-20T19:28:58Z","logger":"controller-runtime.metrics","msg":"Shutting down metrics server with timeout of 1 minute"}
{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"shutting down server","kind":"health probe","addr":"[::]:8081"}
{"level":"info","ts":"2024-03-20T19:28:58Z","msg":"Wait completed, proceeding to shutdown the manager"}
```

Note that after the start, the contexts passed to leader election runnables is closed shortly after, but there is enough time for those jobs to run. For example, to send an HTTP request.

Looking at the controller runtime code, this behavior occurs because when a runnable group is being StopAndWait, it always calls r.Start.

My understanding of the code is like this:
[...]

PR #2724, opened 2024-03-23
🐛 prevent leader election when shutting down a non-elected manager

alexandremahdhaoui:
## Context

This PR should address a bug where the runnable group for Leader Election would start while shutting down the manager: https://github.com/kubernetes-sigs/controller-runtime/issues/2719

What's happening?
- https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L344
- https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L521
- https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L521
- Causes the side-effect of starting the runnableGroup: https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/runnable_group.go#L277

However when leader election is enabled, a non-leader manager will never start the LeaderElection runnable group. Thus, the sync.Once allow starting a new election during shutdown. The change in this PR ensures `Start` is ineffective during shutdown.
- This func calls the `manager.LeaderElection`'s `runnableGroup.Start()` https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L557
- Which is obviously only called when a member becomes leader:  https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L568
- In the case leader election is enabled: https://github.com/kubernetes-sigs/controller-runtime/blob/395cfc7486e652d19fe1b544a436f9852ba26e4f/pkg/manager/internal.go#L429

## Changes

- On shutdown, perform a noop call to the `sync.Once` func associated with the LeaderElection `runnableGroup` to prevent it from starting a new leader election.

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -518,6 +518,8 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 
 		// Stop all the leader election runnables, which includes reconcilers.
 		cm.logger.Info("Stopping and waiting for leader election runnables")
+		// Prevent leader election when shutting down a non-elected manager
+		cm.runnables.LeaderElection.startOnce.Do(func() {})
 		cm.runnables.LeaderElection.StopAndWait(cm.shutdownCtx)
 
 		// Stop the caches before the leader election runnables, this is an important
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -378,6 +378,85 @@ var _ = Describe("manger.Manager", func() {
 
 				Expect(cm.gracefulShutdownTimeout.Nanoseconds()).To(Equal(int64(0)))
 			})
+
+			It("should prevent leader election when shutting down a non-elected manager", func() {
+				var rl resourcelock.Interface
+				m1, err := New(cfg, Options{
+					LeaderElection:          true,
+					LeaderElectionNamespace: "default",
+					LeaderElectionID:        "test-leader-election-id",
+					newResourceLock: func(config *rest.Config, recorderProvider recorder.Provider, options leaderelection.Options) (resourcelock.Interface, error) {
+						var err error
+						rl, err = leaderelection.NewResourceLock(config, recorderProvider, options)
+						return rl, err
+					},
+					HealthProbeBindAddress: "0",
+					Metrics:                metricsserver.Options{BindAddress: "0"},
+					PprofBindAddress:       "0",
+				})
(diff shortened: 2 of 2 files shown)


pkg/manager/internal.go
```diff
@@ -518,6 +518,8 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 
 		// Stop all the leader election runnables, which includes reconcilers.
 		cm.logger.Info("Stopping and waiting for leader election runnables")
+		// Prevent leader election when shutting down a non-elected manager
+		cm.runnables.LeaderElection.startOnce.Do(func() {})
```
alvaroaleman, line 522: Could you add a test, please?

alvaroaleman (commented):
/ok-to-test

pkg/manager/internal.go
alexandremahdhaoui: Added the test in 06d575d6660414965c861063398b4019d7fd2659.

pkg/manager/manager_test.go
```diff
@@ -378,6 +378,82 @@ var _ = Describe("manger.Manager", func() {
...
+					Metrics:                metricsserver.Options{BindAddress: "0"},
+					PprofBindAddress:       "0",
+				})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(m2).ToNot(BeNil())
+				Expect(rl.Describe()).To(Equal("default/test-leader-election-id"))
+
+				Expect(m1.Add(RunnableFunc(func(ctx context.Context) error {
+					defer GinkgoRecover()
+					return nil
```
inteon, line 424: Can you simulate a Runnable that blocks until the context is canceled?

inteon: & make sure it is stopped before manager.Start returns?

alexandremahdhaoui: Yes I will

alvaroaleman (approved):
/cherrypick release-0.17

Issue #2678, opened 2024-02-09
Fakeclient: Patching an object that has allowUnconditionalUpdate=false where the patched version has no RV should succeed


alvaroaleman:
Easiest to show with a testcase, the following test should pass but does not:
```diff
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -359,6 +359,33 @@ var _ = Describe("Fake client", func() {
                        Expect(list.Items).To(ConsistOf(*dep2))
                })

+               FIt("should not return a conflict error from a patch operation with an empty RV and allowsUnconditionalUpdate=false", func() {
+                       schemeBuilder := &scheme.Builder{GroupVersion: schema.GroupVersion{Group: "test", Version: "v1"}}
+                       schemeBuilder.Register(&WithPointerMeta{}, &WithPointerMetaList{})
+                       scheme := runtime.NewScheme()
+                       Expect(schemeBuilder.AddToScheme(scheme)).NotTo(HaveOccurred())
+                       cl := NewClientBuilder().WithScheme(scheme).Build()
+                       original := &WithPointerMeta{ObjectMeta: &metav1.ObjectMeta{
+                               Name:      "obj",
+                               Namespace: "ns2",
+                       }}
+                       err := cl.Create(context.Background(), original)
+                       Expect(err).ToNot(HaveOccurred())
+
+                       newObj := &WithPointerMeta{ObjectMeta: &metav1.ObjectMeta{
+                               Name:      original.Name,
+                               Namespace: original.Namespace,
+                               Annotations: map[string]string{
+                                       "foo": "bar",
+                               },
+                       }}
+                       Expect(cl.Patch(context.Background(), newObj, client.MergeFrom(original))).To(Succeed())
+
+                       patched := &WithPointerMeta{}
+                       Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(original), patched)).To(Succeed())
+                       Expect(patched.Annotations).To(Equal(map[string]string{"foo": "bar"}))
+               })
+
                It("should be able to Create", func() {
                        By("Creating a new configmap")
                        newcm := &corev1.ConfigMap{
```

We never noticed this, because all the tests use built-in resoruces and they all allowUnconditionalUpdate

alvaroaleman:
Actually, at least that precise testcase is wrong as it will end up generating a patch that explicitly sets ResourceVersion to `null` which is indeed getting rejected by the server.

[...]

PR #2725, opened 2024-03-24
🐛 allow fakeclient to patch CR with no RV

alexandremahdhaoui:
# Context

This PR should fix a bug https://github.com/kubernetes-sigs/controller-runtime/issues/2678 where patching a resource would throw a conflict error when the new object doesn't specify any RV.

# Changes

- Fix patch by setting the new accessors' RV to the old accessors' RV.
- Add test to ensure the correct behavior

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -423,9 +423,18 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
 
 	// If the new object does not have the resource version set and it allows unconditional update,
 	// default it to the resource version of the existing resource
-	if accessor.GetResourceVersion() == "" && allowsUnconditionalUpdate(gvk) {
-		accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+	if accessor.GetResourceVersion() == "" {
+		switch {
+		case allowsUnconditionalUpdate(gvk):
+			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+		case bytes.
+			Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeClient).Patch")):
+			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
+			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Patch" func.
+			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+		}
 	}
+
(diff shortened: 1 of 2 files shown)


pkg/client/patch.go
```diff
@@ -112,6 +112,8 @@ func (s *mergeFromPatch) Data(obj Object) ([]byte, error) {
 
 		modified = modified.DeepCopyObject().(Object)
 		modified.SetResourceVersion(version)
+	} else if modified.GetResourceVersion() == "" {
```
troy0820, line 115: This isn't the fakeClient. This is the real client.  Are we supposed to be modifying this rather than `pkg/client/fake/client.go`?

troy0820: I believe you want to look around here 
https://github.com/kubernetes-sigs/controller-runtime/blob/d0396a3d6f9fb554ef2da382a3d0bf05f7565e65/pkg/client/fake/client.go#L426

alexandremahdhaoui: Thanks for the review. I'll look into it

alexandremahdhaoui: Hi Troy! I've fixed it using the "debug stack hack".

pkg/client/fake/client.go
```diff
@@ -423,9 +423,18 @@ func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Ob
...
+	if accessor.GetResourceVersion() == "" {
+		switch {
+		case allowsUnconditionalUpdate(gvk):
+			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+		case bytes.
+			Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeClient).Patch")):
+			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
+			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Patch" func.
+			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+		}
```
alexandremahdhaoui, line 435: Even though both those switch branches leads to the same executed code, I found it to be a bit easier to read than an `if ( xx && yy ) || ( xx && zz )` block.

pkg/client/patch.go
sbueringer: Thx!

sbueringer:
Thx!

/lgtm

alvaroaleman (approved):
thanks!

PR #2207, opened 2023-02-21
WIP: ✨ Cluster Provider and cluster-aware controllers

vincepri:

PR #2726, opened 2024-03-25
✨ WIP: Cluster Provider and cluster-aware controllers

sttts:
Replacing https://github.com/kubernetes-sigs/controller-runtime/pull/2207.


pkg/cluster/cluster.go
```diff
@@ -28,16 +28,45 @@ import (
...
+	// Implementers should return an error if they cannot start operations for the given Cluster,
+	// and should ensure this operation is re-entrant and non-blocking.
+	//
+	//	\_________________|)____.---'--`---.____
+	//              ||    \----.________.----/
+	//              ||     / /    `--'
+	//            __||____/ /_
+	//           |___         \
+	//               `--------'
+	Engage(context.Context, Cluster) error
```
mjudeikis, line 54: Love Start trek analogy :D

pkg/builder/controller.go
```diff
@@ -66,15 +69,16 @@ type Builder struct {
...
 }
 
 // ForInput represents the information set by the For method.
 type ForInput struct {
-	object           client.Object
-	predicates       []predicate.Predicate
-	objectProjection objectProjection
-	err              error
+	object              client.Object
+	forceDefaultCluster bool // in cluster-aware mode, force the object to be watched in the default cluster
```
mjudeikis, line 78: So its set true in cluster-aware mode or other way around?

pkg/builder/options.go
```diff
@@ -154,3 +154,21 @@ type matchEveryOwner struct{}
...
 }
+
+// InDefaultCluster configures the input to only watch objects on the default
+// cluster, even if a cluster provider is set.
+var InDefaultCluster = inDefaultCluster{}
+
+type inDefaultCluster struct{}
+
+func (f inDefaultCluster) ApplyToFor(opts *ForInput) {
+	opts.forceDefaultCluster = true
```
mjudeikis, line 165: Should we invest this where "false" would be default? for "single cluster"?

mjudeikis (commented):
This is way cleaner than previous one. I like it.

pkg/builder/controller.go
sttts: actually I reverted that, but left some bread crumps here. Not sure we really want it. To be discussed.

pkg/builder/options.go
sttts: In single cluster it is always the default cluster that gets the watches. In multi-cluster mode the provider clusters get the watches. This variable is to override the later. It's a noop in the former.

pkg/builder/controller.go
```diff
@@ -58,6 +59,8 @@ type Builder struct {
 	ownsInput        []OwnsInput
 	watchesInput     []WatchesInput
 	mgr              manager.Manager
+	cluster          cluster.Cluster
+	clusterName      string
```
sbueringer, line 63: I know the PR is WIP and I'm just taking a look to get an overview.

The ClusterName field is never written atm.

pkg/config/controller.go
```diff
@@ -46,4 +46,9 @@ type Controller struct {
 	// NeedLeaderElection indicates whether the controller needs to use leader election.
 	// Defaults to true, which means the controller will use leader election.
 	NeedLeaderElection *bool
+
+	// WatchProviderClusters indicates whether the controller should
+	// only watch clusters that are engaged by the cluster provider. Defaults to false
+	// if no provider is set, and to true if a provider is set.
+	WatchProviderClusters *bool
```
sbueringer, line 53: Should this be named WatchOnlyProviderClusters? (currently it sounds like we watch provider clusters in addition to the hub cluster)

sbueringer: I guess that also wouldn't be better because false would signal that we're watching both the provider and hub clusters

pkg/builder/controller.go
sttts: Updated the PR. This is not relevant anymore.

pkg/config/controller.go
sttts: Updated the PR. This is not relevant anymore.

Compare in controller options:
```golang
	// EngageWithDefaultCluster indicates whether the controller should engage
	// with the default cluster of a manager. This defaults to false through the
	// global controller options of the manager if a cluster provider is set,
	// and to true otherwise. Here it can be overridden.
	EngageWithDefaultCluster *bool
	// EngageWithProvidedClusters indicates whether the controller should engage
	// with the provided clusters of a manager. This defaults to true through the
	// global controller options of the manager if a cluster provider is set,
	// and to false otherwise. Here it can be overridden.
	EngageWithProviderClusters *bool
```

sbueringer: That's way clearer, thx!

sttts:
Replaced by https://github.com/kubernetes-sigs/controller-runtime/pull/3019.

PR #2738, opened 2024-03-27
✨ allow setting watchTimeoutPeriod when creating informers

ccding:
We use kubebuilder to create operators. These operators watch a lot of resources. The default watch timeout is a random number between 5 and 10 minutes, which triggers a re-watch every 7.5 minutes on average. However, this behavior causes a log of QPS to our k8s API server. We want to increase the watch timeout to reduce the overhead of the API server, thus this PR makes it configurable.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -392,7 +392,7 @@ func newCache(restConfig *rest.Config, opts Options) newCacheFunc {
 				WatchErrorHandler:     opts.DefaultWatchErrorHandler,
 				UnsafeDisableDeepCopy: ptr.Deref(config.UnsafeDisableDeepCopy, false),
 				NewInformer:           opts.newInformer,
-				MinWatchTimeoutPeriod: opts.MinRewatchPeriod,
+				MinWatchTimeout:       opts.MinRewatchPeriod,
 			}),
 			readerFailOnMissingInformer: opts.ReaderFailOnMissingInformer,
 		}
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -170,6 +170,12 @@ type Options struct {
 	// instead of `reconcile.Result{}`.
 	SyncPeriod *time.Duration
 
+	// WatchTimeoutPeriod is the timeout for the watch request. If the watch request
+	// times out, the cache will close the watch and reconnect.
+	//
+	// Defaults to a random duration between 5 and 10 minutes if unset.
+	WatchTimeoutPeriod *time.Duration
+
 	// ReaderFailOnMissingInformer configures the cache to return a ErrResourceNotCached error when a user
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -170,6 +170,12 @@ type Options struct {
 	// instead of `reconcile.Result{}`.
 	SyncPeriod *time.Duration
 
+	// WatchTimeoutPeriod is the timeout for the watch request. If the watch request
+	// times out, the cache will close the watch and reconnect.
+	//
+	// Defaults to a random duration between 5 and 10 minutes if unset.
+	WatchTimeoutPeriod *time.Duration
```
FillZpp, line 177: LG. Will it be easier to understand if name it like `RewatchPeriod`?

A `WatchTimeoutPeriod` field in `cache.Options` seems to be the timeout that waits for the cache starting to watch?

WDYT @xiang90 @sbueringer @vincepri

pkg/cache/internal/informers.go
```diff
@@ -354,6 +360,9 @@ func (ip *Informers) addInformerToMap(gvk schema.GroupVersionKind, obj runtime.O
 		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
 			ip.selector.ApplyToList(&opts)
 			opts.Watch = true // Watch needs to be set to true separately
+			if ip.watchTimeoutPeriod != nil {
+				opts.TimeoutSeconds = ptr.To(int64(ip.watchTimeoutPeriod.Seconds()))
```
FillZpp, line 364: Since this is a certain time without any random, it might bring a large number of watch requests to apiserver when all the old watches closed and try to reconnect, if your operator watch a lot of resources as you described in this pr.

This is why the default timeout is a random number between 5 and 10 minutes. So how about add a little random time base on the given period?

ccding: Thanks. That makes sense. Fixed following the style of https://github.com/kubernetes/client-go/blob/8c4efe8d079e405329f314fb789a41ac6af101dc/tools/cache/reflector.go#L418

FillZpp:
/lgtm

/assign @alvaroaleman @vincepri @sbueringer 
PTAL :)

alvaroaleman:
>  However, this behavior causes a log of QPS to our k8s API server

The majority of the server-side cost to a watch is in sending the data, not in a client establishing a connection. Clients resume watches from where they left off using the ResourceVersion, so the only thing this could possibly save is some TLS handshakes, which seems neglible to me. Have you done any benchmarking to see the impact of this?

This is an extremely low-level setting and people might not understand the implications of changing it. If we are going to expose something like this, I want to see some data that this actually makes a meaningful difference.

sbueringer:
/close

as there was no further feedback

Issue #2739, opened 2024-03-27
🐛 Webhook builder allows passing in more than one `For` type which gets overridden by the last type.


troy0820:
This issue is in reference to this issue https://github.com/kubernetes-sigs/controller-runtime/issues/1173. 

I realized that this was being done within the reconciler builder but not the webhook.  To enforce a sense of consistency, and error to the caller when this happens within the webhook builder as well.

/kind bug

PR #2740, opened 2024-03-27
🐛 Return error if more than one For() is used on webhook builder

troy0820:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2739

Adds an error to be returned if the `For()` was done multiple times on a webhook builder instead of accepting the last apiType.  This mimics the behavior expected when dealing with the reconciler builder.  Reference issue above and issue https://github.com/kubernetes-sigs/controller-runtime/issues/1173


alvaroaleman (approved):
thanks!

PR #2745, opened 2024-03-30
🐛 Quiet context.Canceled errors during shutdown

### LINKED ISSUE #1927: bug: "error received after stop ..." is logged for context cancellation errors
https://github.com/kubernetes-sigs/controller-runtime/blob/c7a98aa706379c4e5c79ea675c7f333192677971/pkg/manager/internal.go#L541

Context cancellation errors should not be logged as *errors* in this case: presumably the context is cancelled *because* the stop sequence is engaged, so logging an error here is noisy at best, misleading at worst.

cbandy:
Runnable implementations that return ctx.Err() cause a spurious "error received after stop" log message.

Fixes #1927

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -507,8 +507,8 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 				cm.internalCancel()
 			})
 			select {
-			case err, ok := <-cm.errChan:
-				if ok {
+			case err := <-cm.errChan:
+				if !errors.Is(err, context.Canceled) {
 					cm.logger.Error(err, "error received after stop sequence was engaged")
 				}
 			case <-stopComplete:
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/go-logr/logr/funcr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/prometheus/client_golang/prometheus"
@@ -980,6 +981,57 @@ var _ = Describe("manger.Manager", func() {
 					defer GinkgoRecover()
 					return nil
 				}))).NotTo(Succeed())
+			})
+
+			It("should not return runnables context.Canceled errors", func() {
+				Expect(options.Logger).To(BeZero(), "this test overrides Logger")
+
+				var log struct {
+					sync.Mutex
+					messages []string
+				}
+				options.Logger = funcr.NewJSON(func(object string) {
+					log.Lock()
+					log.messages = append(log.messages, object)
+					log.Unlock()
+				}, funcr.Options{})
+
+				m, err := New(cfg, options)
+				Expect(err).NotTo(HaveOccurred())
+				for _, cb := range callbacks {
+					cb(m)
+				}
+
+				// Runnables may return ctx.Err() as shown in some [context.Context] examples.
+				started := make(chan struct{})
+				Expect(m.Add(RunnableFunc(func(ctx context.Context) error {
+					close(started)
+					<-ctx.Done()
+					return ctx.Err()
+				}))).To(Succeed())
+
+				stopped := make(chan error)
+				ctx, cancel := context.WithCancel(context.Background())
+				go func() {
+					stopped <- m.Start(ctx)
+				}()
+
+				// Wait for runnables to start, signal the manager, and wait for it to return.
+				<-started
+				cancel()
+				Expect(<-stopped).To(Succeed())
+
+				// The leader election goroutine emits one more log message after Start() returns.
+				// Take the lock here to avoid a race between it writing to log.messages and the
+				// following read from log.messages.
+				if options.LeaderElection {
+					log.Lock()
+					defer log.Unlock()
+				}
+
+				Expect(log.messages).To(Not(ContainElement(
+					ContainSubstring(context.Canceled.Error()),
+				)))
 			})
 
 			It("should return both runnables and stop errors when both error", func() {


alvaroaleman (commented):
Could you add a test, please?
/ok-to-test

cbandy:
@alvaroaleman Thanks for taking a look. Test added in an amended commit.

pkg/manager/manager_test.go
```diff
@@ -1044,6 +1045,44 @@ var _ = Describe("manger.Manager", func() {
 				}))).NotTo(Succeed())
 			})
 
+			It("should not return runnables context.Canceled errors", func() {
+				Expect(options.Logger).To(BeZero(), "this test overrides Logger")
+
+				logs := []string{}
+				options.Logger = funcr.NewJSON(func(object string) {
+					logs = append(logs, object)
```
alvaroaleman, line 1053: This is a datarace which is why the tests are failing. You need to protect `logs` with a mutex or some such

cbandy: A mutex here solved a race triggered by `append`, but that revealed a race between this line and the assertion on L1091 during the LeaderElection suite. My guess is that this goroutine is to blame: https://github.com/kubernetes-sigs/controller-runtime/blob/21368602d84b669eacfb36ec49b10a665b2d4d49/pkg/manager/internal.go#L423-L427


😞 I added a `time.Sleep` for that suite for now.

pkg/manager/internal.go
```diff
@@ -485,8 +485,8 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 				cm.internalCancel()
 			})
 			select {
-			case err, ok := <-cm.errChan:
-				if ok {
+			case err := <-cm.errChan:
+				if !errors.Is(err, context.Canceled) {
```
vincepri, line 489: Why not keep checking the `ok` value here?

cbandy: I don't see any calls to close `errChan`, so `ok` is always true (today).

Also, in my experience, reading from a closed channel in `select` is for some kind of control flow (break or return or so), or the channel is assigned `nil` so that it is not selected again.

pkg/manager/internal.go
```diff
@@ -485,8 +485,8 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 				cm.internalCancel()
 			})
 			select {
-			case err, ok := <-cm.errChan:
-				if ok {
+			case err := <-cm.errChan:
```
sbueringer, line 488: Wondering if this actually fixes the issues.

Isn't it a bit random if the err is read here or in l.448?

sbueringer: But maybe good enough, whoever is calling manager.Start can also decide ignoring Context.Canceled

pkg/manager/internal.go
sbueringer: Sounds okay to me. I guess if we ever start closing the channel we have to check all usages of the channel anyway.

pkg/manager/manager_test.go
```diff
@@ -1044,6 +1045,54 @@ var _ = Describe("manger.Manager", func() {
...
+				stopped := make(chan error)
+				ctx, cancel := context.WithCancel(context.Background())
+				go func() {
+					stopped <- m.Start(ctx)
+				}()
+
+				// Wait for runnables to start, signal the manager, and wait for it.
+				<-started
+				cancel()
+				Expect(<-stopped).To(Succeed())
```
sbueringer, line 1084: Q: Couldn't it be that l.448 reads and returns the error from errChan?

cbandy: Here, in this test? I don't think so.

I don't fully understand all the startup logic, but the runnable groups don't appear to send to that channel during Start (...exactly). The leader election goroutine does.

sbueringer: Let me rephrase.

As far as I can tell there are two places reading from errChan:
```go
	select {
	case <-ctx.Done():
		// We are done
		return nil
	case err := <-cm.errChan:
		// Error starting or running a runnable
		return err
	}
```

```
			select {
			case err := <-cm.errChan:
				if !errors.Is(err, context.Canceled) {
					cm.logger.Error(err, "error received after stop sequence was engaged")
				}
			case <-stopComplete:
				return
			}
```

This test case verifies that the cancel error ends up in the logger and not as a return parameter of `stopped <- m.Start(ctx)`. I'm wondering if the cancel error always ends up in the logger, and if yes, why.

sbueringer: Nevermind. engageStopProcedure is only called after start returns

pkg/manager/internal.go
sbueringer: Just realized that engageStopProcedure is only called after start returns, so all good

pkg/manager/manager_test.go
```diff
@@ -1044,6 +1045,54 @@ var _ = Describe("manger.Manager", func() {
...
+
+				// TODO: Diagnose why LeaderElection races after Start() returns.
+				if options.LeaderElection {
```
sbueringer, line 1087: What happened without the sleep?

sbueringer: I'll answer above

pkg/manager/manager_test.go
sbueringer: So the idea is that currently with leaderelection engageStopProcedure is still writing the the logger while the test is asserting on the logs?

sbueringer: My guess would be because the RunnableFunc is not a LeaderElectionRunnable. So it won't wait for leaderelection to close start

sbueringer: Okay reproduced locally. ^^ doesn't make sense as everything that doesn't implement LeaderElectionRunnable is treated like it needs leader election

sbueringer: I think the problem is that nothing is actually waiting for this go routine to return:
```go
	go func() {
		for {
			// Closing in the for loop is required to avoid race conditions between
			// the closure of all internal procedures and making sure to have a reader off the error channel.
			closeOnce.Do(func() {
				// Cancel the internal stop channel and wait for the procedures to stop and complete.
				close(cm.internalProceduresStop)
				cm.internalCancel()
			})
			select {
			case err := <-cm.errChan:
				if !errors.Is(err, context.Canceled) {
					cm.logger.Error(err, "error received after stop sequence was engaged")
				}
			case <-stopComplete:
				return
			}
		}
	}()
```

While we call `defer close(stopComplete)` at the end of Start, this only means that the go routine will return pretty soon after, but we don't wait for it before we return. So it makes sense that the goroutine still writes to the logger even after Start returned

sbueringer: I think the reason why it only occurs with leader election is because in that case we get one more log message:
image
(the other log messages are all produced "synchronously" within engageStopProcedure)

sbueringer: @cbandy @alvaroaleman If that makes sense to you both, let's add a godoc comment before the sleep and I think we're good

cbandy: 🤔 The fact that `Start` aggregates its `err` with `stopErr` makes me think it expects `engageStopProcedure` to return the outcome of stopping. Why does it inject a channel at all?

Should `engageStopProcedure` wait for the `errChan`-draining goroutine to finish?

sbueringer: > 🤔 The fact that Start aggregates its err with stopErr makes me think it expects engageStopProcedure to return the outcome of stopping. Why does it inject a channel at all?

Not sure. I think the goal of this case `case err := <-cm.errChan:` in `engageStopProcedure` is mostlyl to avoid running into deadlocks because the errChan blocks.

> Should engageStopProcedure wait for the errChan-draining goroutine to finish?

With the way it currently works it can't because this defer (`defer close(stopComplete)`) in Start is called too late.

sbueringer: I guess we could wait in the Start func that the `errChan`-draining goroutine also finishes.

It would be a bit nicer, but I'm not sure if I want to introduce even more complexity into all of this. It's already pretty had to reason about it.

pkg/manager/manager_test.go
```diff
@@ -1044,6 +1045,54 @@ var _ = Describe("manger.Manager", func() {
...
+					time.Sleep(time.Second)
+				}
+
```
alvaroaleman, line 1090: The reason this races is that you access the log output without acquiring the lock here. I am actually surprised the sleep fixed that. Proper fix would be this:
```suggestion
				log.Lock()
				defer log.Unlock()


```

cbandy: This is called after `m.Start` has returned. Is `m` expected to log in that case?

cbandy: I like `Lock` more than `Sleep`. I'll rebase and try that out.

cbandy:
Rebased with recommended changes. PTAL.

alvaroaleman (approved):
Thanks! And sorry this ended up taking so long

PR #2746, opened 2024-03-31
:book: Add designs/multi-cluster.md

sttts:
Controller-runtime today allows to write controllers against one cluster only.
Multi-cluster use-cases require the creation of multiple managers and/or cluster
objects. This proposal is about adding native support for multi-cluster use-cases
to controller-runtime.

The proposed changes are prototyped in https://github.com/kubernetes-sigs/controller-runtime/pull/2726.

### DIFF
--- a/designs/multi-cluster.md
+++ b/designs/multi-cluster.md
@@ -1,0 +1,281 @@
+# Multi-Cluster Support
+Author: @sttts
+Initial implementation: @vincepri
+
+Last Updated on: 03/26/2024
+
+## Table of Contents
+
+<!--ts-->
+- [Multi-Cluster Support](#multi-cluster-support)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Examples](#examples)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [Multi-Cluster-Compatible Reconcilers](#multi-cluster-compatible-reconcilers)
+  - [User Stories](#user-stories)
+    - [Controller Author with no interest in multi-cluster wanting to old behaviour.](#controller-author-with-no-interest-in-multi-cluster-wanting-to-old-behaviour)
+    - [Multi-Cluster Integrator wanting to support cluster managers like Cluster-API or kind](#multi-cluster-integrator-wanting-to-support-cluster-managers-like-cluster-api-or-kind)
+    - [Multi-Cluster Integrator wanting to support apiservers with logical cluster (like kcp)](#multi-cluster-integrator-wanting-to-support-apiservers-with-logical-cluster-like-kcp)
+    - [Controller Author without self-interest in multi-cluster, but open for adoption in multi-cluster setups](#controller-author-without-self-interest-in-multi-cluster-but-open-for-adoption-in-multi-cluster-setups)
+    - [Controller Author who wants to support certain multi-cluster setups](#controller-author-who-wants-to-support-certain-multi-cluster-setups)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Alternatives](#alternatives)
+  - [Implementation History](#implementation-history)
+
+<!--te-->
+
+## Summary
+
+Controller-runtime today allows to write controllers against one cluster only.
+Multi-cluster use-cases require the creation of multiple managers and/or cluster
+objects. This proposal is about adding native support for multi-cluster use-cases
+to controller-runtime.
+
+## Motivation
+
+This change is important because:
+- multi-cluster use-cases are becoming more and more common, compare projects
+  like Kamarda, Crossplane or kcp. They all need to write (controller-runtime)
+  controllers that operate on multiple clusters.
+- writing controllers for upper systems in a **portable** way is hard today.
+  Consequently, there is no multi-cluster controller ecosystem, but could and
+  should be.
+- kcp maintains a [controller-runtime fork with multi-cluster support](https://github.com/kcp-dev/controller-runtime)
+  because adding support on-top leads to an inefficient controller design, and
+  even more important leads of divergence in the ecosystem.
+
+### Goals
+
+- Provide a way to natively write controllers that
+  1. (UNIFORM MULTI-CLUSTER CONTROLLER) operate on multiple clusters in a uniform way,
+     i.e. reconciling the same resources on multiple clusters, **optionally**
+     - sourcing information from one central hub cluster
+     - sourcing information cross-cluster.
+
+     Example: distributed `ReplicaSet` controller, reconciling `ReplicaSets` on multiple clusters.
+  2. (AGGREGATING MULTI-CLUSTER CONTROLLER) operate on one central hub cluster aggregating information from multiple clusters.
+     
+     Example: distributed `Deployment` controller, aggregating `ReplicaSets` back into the `Deployment` object.
+- Allow clusters to dynamically join and leave the set of clusters a controller operates on.
+- Allow event sources to be cross-cluster:
+  1. Multi-cluster events that trigger reconciliation in the one central hub cluster.
+  2. Central hub cluster events to trigger reconciliation on multiple clusters.
+- Allow (informer) indexes that span multiple clusters.
+- Allow logical clusters where a set of clusters is actually backed by one physical informer store.
+- Allow 3rd-parties to plug in their multi-cluster adapter (in source code) into
+  an existing multi-cluster-compatible code-base.
+- Minimize the amount of changes to make a controller-runtime controller 
+  multi-cluster-compatible, in a way that 3rd-party projects have no reason to
+  object these kind of changes.
+
+Here we call a controller to be multi-cluster-compatible if the reconcilers get
+reconcile requests in cluster `X` and do all reconciliation in cluster `X`. This 
+is less than being multi-cluster-aware, where reconcilers implement cross-cluster
+logic.
+
+### Examples
+
+- Run a controller-runtime controller against a kubeconfig with arbitrary many contexts, all being reconciled.
+- Run a controller-runtime controller against cluster-managers like kind, Cluster-API, Open-Cluster-Manager or Hypershift.
+- Run a controller-runtime controller against a kcp shard with a wildcard watch.
+
+### Non-Goals/Future Work
+
+- Ship integration for different multi-cluster setups. This should become 
+  out-of-tree subprojects that can individually evolve and vendor'ed by controller authors.
+- Make controller-runtime controllers "binary pluggable".
+- Manage one manager per cluster.
+- Manage one controller per cluster with dedicated workqueues.
+
+## Proposal
+
+The `ctrl.Manager` _SHOULD_ be extended to get an optional `cluster.Provider` via
+`ctrl.Options` implementing
+
+```golang
+// pkg/cluster
+type Provider interface {
+   Get(ctx context.Context, clusterName string, opts ...Option) (Cluster, error)
+   List(ctx context.Context) ([]string, error)
+   Watch(ctx context.Context) (Watcher, error)
+}
+```
+The `cluster.Cluster` _SHOULD_ be extended with a unique name identifier:
+```golang
+// pkg/cluster:
+type Cluster interface {
+    Name() string
+	...
+}
+```
+
+The `ctrl.Manager` will use the provider to watch clusters coming and going, and
+will inform runnables implementing the `cluster.AwareRunnable` interface:
+
+```golang
+// pkg/cluster
+type AwareRunnable interface {
+	Engage(context.Context, Cluster) error
+	Disengage(context.Context, Cluster) error
+}
+```
+In particular, controllers implement the `AwareRunnable` interface. They react
+to engaged clusters by duplicating and starting their registered `source.Source`s
+and `handler.EventHandler`s for each cluster through implementation of
+```golang
+// pkg/source
+type DeepCopyableSyncingSource interface {
+	SyncingSource
+	DeepCopyFor(cluster cluster.Cluster) DeepCopyableSyncingSource
+}
+
+// pkg/handler
+type DeepCopyableEventHandler interface {
+    EventHandler
+    DeepCopyFor(c cluster.Cluster) DeepCopyableEventHandler
+}
+```
+The standard implementing types, in particular `internal.Kind` will adhere to
+these interfaces.
+
+The `ctrl.Manager` _SHOULD_ be extended by a `cluster.Cluster` getter:
+```golang
+// pkg/manager
+type Manager interface {
+    // ...
+    GetCluster(ctx context.Context, clusterName string) (cluster.Cluster, error)
+}
+```
+The embedded `cluster.Cluster` corresponds to `GetCluster(ctx, "")`. We call the
+clusters with non-empty name "provider clusters" or "enganged clusters", while
+the embedded cluster of the manager is called the "default cluster" or "hub 
+cluster".
+
+The `reconcile.Request` _SHOULD_ be extended by an optional `ClusterName` field:
+```golang
+// pkg/reconile
+type Request struct {
+    ClusterName string
+    types.NamespacedName
+}
+```
+
+With these changes, the behaviour of controller-runtime without a set cluster
+provider will be unchanged.
+
+### Multi-Cluster-Compatible Reconcilers
+
+Reconcilers can be made multi-cluster-compatible by changing client and cache
+accessing code from directly accessing `mgr.GetClient()` and `mgr.GetCache()` to
+going through `mgr.GetCluster(req.ClusterName).GetClient()` and
+`mgr.GetCluster(req.ClusterName).GetCache()`.
+
+When building a controller like
+```golang
+builder.NewControllerManagedBy(mgr).
+    For(&appsv1.ReplicaSet{}).
+	Owns(&v1.Pod{}).
+    Complete(reconciler)
+```
+with the described change to use `GetCluster(ctx, req.ClusterName)` will automatically 
+act as *uniform multi-cluster controller*. It will reconcile resources from cluster `X`
+in cluster `X`.
+
+For a manager with `cluster.Provider`, the builder _SHOULD_ create a controller
+that sources events **ONLY** from the provider clusters that got engaged with
+the controller.
+
+Controllers that should be triggered by events on the hub cluster will have to
+opt-in like in this example:
+
+```golang
+builder.NewControllerManagedBy(mgr).
+    For(&appsv1.Deployment{}, builder.InDefaultCluster).
+	Owns(&v1.ReplicaSet{}).
+    Complete(reconciler)
+```
+A mixed set of sources is possible as shown here in the example.
+
+## User Stories
+
+### Controller Author with no interest in multi-cluster wanting to old behaviour.
+
+- Do nothing. Controller-runtime behaviour is unchanged.
+
+### Multi-Cluster Integrator wanting to support cluster managers like Cluster-API or kind
+
+- Implement the `cluster.Provider` interface, either via polling of the cluster registry
+  or by watching objects in the hub cluster.
+- For every new cluster create an instance of `cluster.Cluster`.
+
+### Multi-Cluster Integrator wanting to support apiservers with logical cluster (like kcp)
+
+- Implement the `cluster.Provider` interface by watching the apiserver for logical cluster objects
+  (`LogicalCluster` CRD in kcp).
+- Return a facade `cluster.Cluster` that scopes all operations (client, cache, indexers) 
+  to the logical cluster, but backed by one physical `cluster.Cluster` resource.
+- Add cross-cluster indexers to the physical `cluster.Cluster` object.
+
+### Controller Author without self-interest in multi-cluster, but open for adoption in multi-cluster setups
+
+- Replace `mgr.GetClient()` and `mgr.GetCache` with `mgr.GetCluster(req.ClusterName).GetClient()` and `mgr.GetCluster(req.ClusterName).GetCache()`.
+- Make manager and controller plumbing vendor'able to allow plugging in multi-cluster provider.
+
+### Controller Author who wants to support certain multi-cluster setups
+
+- Do the `GetCluster` plumbing as described above.
+- Vendor 3rd-party multi-cluster providers and wire them up in `main.go`
+
+## Risks and Mitigations
+
+- The standard behaviour of controller-runtime is unchanged for single-cluster controllers.
+- The activation of the multi-cluster mode is through attaching the `cluster.Provider` to the manager. 
+  To make it clear that the semantics are experimental, we make the `Options.provider` field private
+  and adds `Options.WithExperimentalClusterProvider` method.
+- We only extend these interfaces and structs:
+  - `ctrl.Manager` with `GetCluster(ctx, clusterName string) (cluster.Cluster, error)`
+  - `cluster.Cluster` with `Name() string`
+  - `reconcile.Request` with `ClusterName string`
+  We think that the behaviour of these extensions is well understood and hence low risk.
+  Everything else behind the scenes is an implementation detail that can be changed
+  at any time.
+
+## Alternatives
+
+- Multi-cluster support could be built outside of core controller-runtime. This would
+  lead likely to a design with one manager per cluster. This has a number of problems:
+  - only one manager can serve webhooks or metrics
+  - cluster management must be custom built
+  - logical cluster support would still require a fork of controller-runtime and
+    with that a divergence in the ecosystem. The reason is that logical clusters
+    require a shared workqueue because they share the same apiserver. So for
+    fair queueing, this needs deep integration into one manager.
+  - informers facades are not supported in today's cluster/cache implementation.
+- We could deepcopy the builder instead of the sources and handlers. This would
+  lead to one controller and one workqueue per cluster. For the reason outlined
+  in the previous alternative, this is not desireable.
+- We could skip adding `ClusterName` to `reconcile.Request` and instead pass the
+  cluster through in the context. On the one hand, this looks attractive as it
+  would avoid having to touch reconcilers at all to make them multi-cluster-compatible.
+  On the other hand, with `cluster.Cluster` embedded into `manager.Manager`, not
+  every method of `cluster.Cluster` carries a context. So virtualizing the cluster
+  in the manager leads to contradictions in the semantics.
+
+  For example, it can well be that every cluster has different REST mapping because
+  installed CRDs are different. Without a context, we cannot return the right
+  REST mapper.
+
+  An alternative would be to add a context to every method of `cluster.Cluster`,
+  which is a much bigger and uglier change than what is proposed here.
+
+
+## Implementation History
+
+- [PR #2207 by @vincepri : WIP: ✨ Cluster Provider and cluster-aware controllers](https://github.com/kubernetes-sigs/controller-runtime/pull/2207) – with extensive review
+- [PR #2208 by @sttts replace #2207: WIP: ✨ Cluster Provider and cluster-aware controllers](https://github.com/kubernetes-sigs/controller-runtime/pull/2726) – 
+  picking up #2207, addressing lots of comments and extending the approach to what kcp needs, with a `fleet-namespace` example that demonstrates a similar setup as kcp with real logical clusters.
+- [github.com/kcp-dev/controller-runtime](https://github.com/kcp-dev/controller-runtime) – the kcp controller-runtime fork


designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+and `handler.EventHandler`s for each cluster through implementation of
+```golang
+// pkg/source
+type DeepCopyableSyncingSource interface {
+	SyncingSource
+	DeepCopyFor(cluster cluster.Cluster) DeepCopyableSyncingSource
+}
+
+// pkg/handler
+type DeepCopyableEventHandler interface {
```
alvaroaleman, line 137: The eventhandlers are stateless, why do we need the deepcopy for them?

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+will inform runnables implementing the `cluster.AwareRunnable` interface:
+
+```golang
+// pkg/cluster
+type AwareRunnable interface {
+	Engage(context.Context, Cluster) error
+	Disengage(context.Context, Cluster) error
+}
+```
+In particular, controllers implement the `AwareRunnable` interface. They react
```
alvaroaleman, line 126: Rather than changing the controller type directly and requiring all its dependencies to known how to deepcopy themselves, how about having something like a `controllerconstructor` (name tbd) in between that is filled with a `[]watchConstructor{source func(Cluster) source.Source, handler func(Cluster) handler.Handler, predicate func(cluster) []predicate.Predicate}`?

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+## Proposal
+
+The `ctrl.Manager` _SHOULD_ be extended to get an optional `cluster.Provider` via
+`ctrl.Options` implementing
+
+```golang
+// pkg/cluster
+type Provider interface {
+   Get(ctx context.Context, clusterName string, opts ...Option) (Cluster, error)
+   List(ctx context.Context) ([]string, error)
```
alvaroaleman, line 103: Why return `[]string` here rather than `[]Cluster`?

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+The `cluster.Cluster` _SHOULD_ be extended with a unique name identifier:
+```golang
+// pkg/cluster:
+type Cluster interface {
+    Name() string
+	...
+}
+```
+
+The `ctrl.Manager` will use the provider to watch clusters coming and going, and
```
alvaroaleman, line 116: I'll have to think about if and how this is doable, but ideally the "thing that comes and goes" wouldn't be typed to `cluster.Cluster` but can be anything, so this mechanism can also be used if folks have sources that are not kube watches

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+Controller-runtime today allows to write controllers against one cluster only.
+Multi-cluster use-cases require the creation of multiple managers and/or cluster
+objects. This proposal is about adding native support for multi-cluster use-cases
+to controller-runtime.
+
+## Motivation
+
+This change is important because:
+- multi-cluster use-cases are becoming more and more common, compare projects
+  like Kamarda, Crossplane or kcp. They all need to write (controller-runtime)
```
embik, line 42: Uber nitpick: The project in question is called Karmada.

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+
+Here we call a controller to be multi-cluster-compatible if the reconcilers get
+reconcile requests in cluster `X` and do all reconciliation in cluster `X`. This 
+is less than being multi-cluster-aware, where reconcilers implement cross-cluster
+logic.
+
+### Examples
+
+- Run a controller-runtime controller against a kubeconfig with arbitrary many contexts, all being reconciled.
+- Run a controller-runtime controller against cluster-managers like kind, Cluster-API, Open-Cluster-Manager or Hypershift.
```
elmiko, line 83: given the cluster-api example here, is the intention that controllers will be able to reconcile CRDs in clusters that they know about that may only exist in a subset of clusters (e.g. Machine objects in the management cluster but not in the workload cluster) ?

elmiko (commented):
i think this is an interesting idea and i could see using it, i just have a question about some of the mechanics.

for context, i am investigating a cluster-api provider for karpenter and it would be nice to have the controllers discriminate between objects in the management cluster and objects in the workload clusters.

designs/multi-cluster.md
sbueringer: Good point. I think that has to be possible. Otherwise we need all resources that we watch in all clusters

designs/multi-cluster.md
sbueringer: +1 Would be good for consistency with the Get func

designs/multi-cluster.md
```diff
@@ -0,0 +1,281 @@
...
+The embedded `cluster.Cluster` corresponds to `GetCluster(ctx, "")`. We call the
+clusters with non-empty name "provider clusters" or "enganged clusters", while
+the embedded cluster of the manager is called the "default cluster" or "hub 
+cluster".
+
+The `reconcile.Request` _SHOULD_ be extended by an optional `ClusterName` field:
+```golang
+// pkg/reconile
+type Request struct {
+    ClusterName string
```
sbueringer, line 162: Would it make sense to have a struct here instead? So we end up with req.Cluster.Name?

(Probably nicer if we anticipate that at some point we might want to add more information about the cluster)

designs/multi-cluster.md
sbueringer: Would this be mostly about a more generic name? (can't think of much that would work, maybe something like scope)

designs/multi-cluster.md
sbueringer: Looking at the propotype. I think this is because EventHandler then would store the Cluster (it is using that to add the Cluster to the request)

designs/multi-cluster.md
sbueringer: I think this would require more invasive changes to our public API (the Controller interface)

alvaroaleman: No, you can call `Watch` on an existing controller. The idea is to not let the `Controller` or its dependencies have any knowledge about this but instead have a thing on top of the `Controller` that is configured with constructors that take a `cluster.Cluster` and return a source/predicate/handler and then uses those to call `Watch` when a new cluster appears.

When one disappears, it would cancel the `context` on the Source.

The idea really is the opposite, I do not want the `Controller` to know how to extend itself like this, IMHO this is a higher-level abstraction.

designs/multi-cluster.md
sttts: fixed

designs/multi-cluster.md
sttts: Good point. Question is whether one would rather group them in managers such that every manager has a uniform set of clusters.

designs/multi-cluster.md
sttts: There is a misunderstanding of the interface. The getter is actually the constructor. The life-cycle of the returned clusters is owned by the manager (they are added as runnables). Hence, the `List` returns names, not clusters. We should rather rename `Get` to `Create` or `Connect`.

designs/multi-cluster.md
sttts: Compare https://github.com/kubernetes-sigs/controller-runtime/pull/2726 after latest push. I have implemented @alvaroaleman's idea via a `MultiClusterController` wrapper implementing `cluster.AwareRunnable` and just calling `Watch` on the actual controller. All the deepcopy'ing is gone 🎉 Much nicer IMO. @alvaroaleman great intuition!

designs/multi-cluster.md
sttts: See my updated PR https://github.com/kubernetes-sigs/controller-runtime/pull/2726. You can now opt into provider and/or the default cluster per controller via options:
```
	// EngageWithDefaultCluster indicates whether the controller should engage
	// with the default cluster of a manager. This defaults to false through the
	// global controller options of the manager if a cluster provider is set,
	// and to true otherwise. Here it can be overridden.
	EngageWithDefaultCluster *bool
	// EngageWithProvidedClusters indicates whether the controller should engage
	// with the provided clusters of a manager. This defaults to true through the
	// global controller options of the manager if a cluster provider is set,
	// and to false otherwise. Here it can be overridden.
	EngageWithProviderClusters *bool
```

There is no logic yet for a controller to decide whether to engage with a provider cluster or not. Now it's with all of them. If the setup is more diverse, we might want such a functionality, e.g. some kind of pre-check: ctrl.WantsToEngage(ctx, cluster) bool`.

designs/multi-cluster.md
sttts: Was thinking that you can always get the cluster from the manager. This is really just the id aka name. We could drop the `Name` suffix. I tried to use `ClusterName` instead of `Cluster` everywhere to make clear what is meant. But no strong opinion here.

designs/multi-cluster.md
sttts: This is gone now in https://github.com/kubernetes-sigs/controller-runtime/pull/2726.

Will update the design here.

designs/multi-cluster.md
elmiko: i'm still understanding the changes in #2726, but i think what you are saying here makes sense to me and would solve the issue.

> some kind of pre-check: ctrl.WantsToEngage(ctx, cluster) bool`.

+1, i think we definitely need some way for the client user to specify when it should check a specific cluster for a resource.

designs/multi-cluster.md
sbueringer: Definitely sounds helpful to clearly differentiate between the Cluster struct and the ClusterName

designs/multi-cluster.md
mjudeikis: I somehow think it should be the author's and managers responsibility (for now) to group them into groups which are working with the pattern. At this point, we don't know what we don't know. Once this is released, we can gather some feedback on edge cases and take it from there. I suspect the majority of use cases will be still single cluster reconcile loops.

Maybe document this edge case and mark this feature overall as experimental? This way we not committing to full production level stability, and allow to gather more feedback?

sttts:
For those reading, this is currently a little outdated. https://github.com/kubernetes-sigs/controller-runtime/pull/2726 has a changed design proposed by @alvaroaleman. Will come back soon to both PRs.

embik:
/reopen

We'd like to continue working on this, time is simply a bit scarce at the moment.

embik:
@Gomaya I'm working on a prototype that attempts to address the review comments in #2726. Once everyone is back from KubeCon, I plan to run this by everyone involved and try to move the feature forward.

designs/multi-cluster.md
embik: With the BYO request/eventhandler changes in #3019, I brought this back after remembering it was mentioned in the proposal. The previous version of the prototype had a weird second layer of event handler that was wrapping the actual event wrapper and was using the event object to communicate the cluster name in. That felt all kinds of weird.

Because we now have BYO EventHandlers, it's possible that they are not entirely stateless (as @sbueringer pointed out, some event handlers might have to store the cluster name in absence of any information on the event object itself). So I think this approach is the most clean, to be honest. It's entirely optional in #3019, existing EventHandlers don't need to be changed.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+}
+```
+
+A new interface for cluster-aware runnables will be provided:
+
+```golang
+// pkg/cluster
+type Aware interface {
+	// Engage gets called when the component should start operations for the given Cluster.
+	// The given context is tied to the Cluster's lifecycle and will be cancelled when the
```
yastij, line 144: how would a context be selected from the kubeconfig passed to the controller?

embik: Just to avoid any confusion, this description is talking about a `context.Context`, not a kubeconfig context.

But in general: This would be something to implement in a `Provider`, a kubeconfig provider could be a very simple implementation (although the focus is on _dynamic_ providers, and a kubeconfig would probably be fairly static). How that provider translates the `clusterName` parameter in the `Get` method (see the `Provider` interface above) to a kubeconfig context would be up to the implementation, but I could see the context name being the identifier here (since `Get` returns a `cluster.Cluster`, we need credentials for the embedded client, so a context makes a lot of sense here).

Does that make sense? 🤔

sttts:
Linked the updated implementation https://github.com/kubernetes-sigs/controller-runtime/pull/3019.

maximilianbraun:
+1 highly appreciated from SAP & in needed in context of our open source efforts with our [european partners](https://www.bmwk.de/Redaktion/EN/Artikel/Industry/ipcei-cis.html).

mirzakopic:
+1 We are very excited about this and are looking forward us it in our open source efforts with https://apeirora.eu/ @SAP

alvaroaleman (commented):
There are actually two subcases in here that we don't seem to be differentiating well:
a) Controller whose `request` will always contain the cluster the event originated from
b) Controller whose `request` may contain a cluster that is not the one the event originated from

I _think_ you are mostly thinking about a) with this proposal. You write:
> We could deepcopy the builder instead of the sources and handlers. This would
  lead to one controller and one workqueue per cluster. For the reason outlined
  in the previous alternative, this is not desireable.

I am having trouble finding the "reasons outlined" and I do think instantiating a controller per cluster rather than adding/removing eventsources is preferrable whenever possible, because:
1) Otherwise workqueue metrics get dilluted, this is a big issue for ops, workqueue_depth is usually one of the main metrics people alert on
2) It requires zero changes in existing reconcilers

I'd love to hear your thoughts on this and I do think we should clearly differentiate the two cases in this doc. What also seems to get lost a bit is that the main issue today is with doing all of this dynamically - pkg/cluster already exists, if the cluster set is static, this design adds very little.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+This change is important because:
+- multi-cluster use-cases are becoming more and more common, compare projects
+  like Kamarda, Crossplane or kcp. They all need to write (controller-runtime)
+  controllers that operate on multiple clusters.
+- writing controllers for upper systems in a **portable** way is hard today.
+  Consequently, there is no multi-cluster controller ecosystem, but could and
+  should be.
+- kcp maintains a [controller-runtime fork with multi-cluster support](https://github.com/kcp-dev/controller-runtime)
+  because adding support on-top leads to an inefficient controller design, and
```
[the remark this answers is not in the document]
sttts: We explicitly don't want one controller (with its own workqueue) per cluster.

Example: forget about workspaces. Imagine controller-runtime only support controller per one (!) namespace, each with its own namespace. Same argument here, just level higher.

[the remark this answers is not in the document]
sttts: And independently you could imagine cases where the same is true e.g. for cluster-api cases where the workqueue should be shared. That's what this enhancement is enabling.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+- Allow event sources to be cross-cluster such that:
+  1. Multi-cluster events can trigger reconciliation in the one central hub cluster.
+  2. Central hub cluster events can trigger reconciliation on multiple clusters.
+- Allow reconcilers to look up objects through (informer) indexes from specific other clusters.
+- Minimize the amount of changes to make a controller-runtime controller 
+  multi-cluster-compatible, in a way that 3rd-party projects have no reason to
+  object to these kind of changes.
+
+Here we call a controller to be multi-cluster-compatible if the reconcilers get
+reconcile requests in cluster `X` and do all reconciliation in cluster `X`. This
```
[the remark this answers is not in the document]
sttts: Related to "one workqueue" I guess. Start/Stop means another workqueue, which we don't want.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+	//               `--------'
+	Engage(context.Context, Cluster) error
+
+	// Disengage gets called when the component should stop operations for the given Cluster.
+	Disengage(context.Context, Cluster) error
+}
+```
+
+`ctrl.Manager` will implement `cluster.Aware`. As specified in the `Provider` interface,
+it is the cluster provider's responsibility to call `Engage` and `Disengage` on a `ctrl.Manager`
```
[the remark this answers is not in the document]
sttts: That's out of scope of the goal of the interface. Wiring the manager in will happen when starting the provider:
```golang
prov := NewProvider()
mgr := NewManager(..., Options: {provider: prov})
go mgr.Start(ctx)
go prov.Start(ctx, mgr)
```

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+cluster providers. Think of generic CNCF projects like cert-manager wanting to
+support multi-cluster setups generically without forking the codebase.
+
+### BYO Request Type
+
+Instead of using the new `reconcile.ClusterAwareRequest`, implementations _CAN_ also bring their
+own request type through the generics support in `Typed*` types (`request comparable`).
+
+**Note:** these kind of controllers won't be uniform in the sense of compatibility
+with arbitrary cluster providers, but for use-cases that are tiedly integrated
```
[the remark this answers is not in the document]
sttts: > The cluster provider doesn't know anything about what request type controllers use?

This is not about the provider, but about controllers. We want to allow allow "uniform cluster support" without minimal changes to existing controllers. We don't control the controllers. These are 3rdparty controllers. If we demand lots of Golang generics type magic, it's a no go.

Hence:

The controllers cannot be generic as outlined above. Imagine they were:
```
Reconcile(ctx context.Context, req T) (reconcile.Result, error) {
  existing code ... req.Namespace ... existing code
}
```
This does not compile because the Go compiler resolves `.Namespace` against an arbitrary `T any`. Hence, have two ways out:
1. use some interface for T that provides `Name+Namespace() string` as a method. This is diverging from 99.9% of controllers today.
2. have a struct that is common aka upstream, and with that has least resistence to be used in practice.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+builder.NewControllerManagedBy(mgr).
+    WithOptions(controller.TypedOptions{
+        EngageWithDefaultCluster: ptr.To(true),
+        EngageWithProviderClusters: ptr.To(false),
+    }).
+    For(&appsv1.Deployment{}).
+	Owns(&v1.ReplicaSet{}).
+    Complete(reconciler)
+```
+
```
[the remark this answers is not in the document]
sttts: See implementation proposal https://github.com/kubernetes-sigs/controller-runtime/pull/3019.

designs/multi-cluster.md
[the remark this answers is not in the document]
alvaroaleman: > We explicitly don't want one controller (with its own workqueue) per cluster.

Others might disagree. This question needs a proper evaluation with pro/cons of both approaches rather than jumping to conclusions.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+Here we call a controller to be multi-cluster-compatible if the reconcilers get
+reconcile requests in cluster `X` and do all reconciliation in cluster `X`. This 
+is less than being multi-cluster-aware, where reconcilers implement cross-cluster
+logic.
+
+### Examples
+
+- Run a controller-runtime controller against a kubeconfig with arbitrary many contexts, all being reconciled.
+- Run a controller-runtime controller against cluster managers like kind, Cluster API, Open-Cluster-Manager or Hypershift.
+- Run a controller-runtime controller against a kcp shard with a wildcard watch.
```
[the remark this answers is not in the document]
embik: We agree things should be highly pluggable, that's why this is just an (incomplete) list of things that you could eventually plug in. Agreed that kcp is an uncommon use case, but so far we've made (recent) design decisions with Kubernetes clusters in general in mind. The "kcp" provider we'd like to built is itself far from ready yet.

designs/multi-cluster.md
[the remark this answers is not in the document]
sttts: I agree that both topologies can have their place. Am not even sure pro/con is helpful. We shouldn't be opinionated, but give the primitives for the developer to decide.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+// and to engage or disengage clusters with the manager the provider is
+// run against.
+//
+// Example: A Cluster API provider would be responsible for discovering and
+// managing clusters that are backed by Cluster API resources, which can live
+// in multiple namespaces in a single management cluster.
+type Provider interface {
+	// Get returns a cluster for the given identifying cluster name. Get
+	// returns an existing cluster if it has been created before.
+	Get(ctx context.Context, clusterName string) (Cluster, error)
```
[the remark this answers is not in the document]
embik: Are you thinking of [`logical.Name`](https://github.com/kubernetes-sigs/logical-cluster/blob/ecfa507978904b20314431ed7ad194a647c25bc9/types.go#L21) here or more a struct?

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+A new interface for cluster-aware runnables will be provided:
+
+```golang
+// pkg/cluster
+type Aware interface {
+	// Engage gets called when the component should start operations for the given Cluster.
+	// The given context is tied to the Cluster's lifecycle and will be cancelled when the
+	// Cluster is removed or an error occurs.
+	//
+	// Implementers should return an error if they cannot start operations for the given Cluster,
```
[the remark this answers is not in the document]
embik: The idea is that anything that needs to be done for starting operations on a cluster is "blocking" (and thus would return an error), but the operations on the engaged cluster themselves are not blocking.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+depending on the `request` type used.
+
+With the described changes (use `GetCluster(ctx, req.ClusterName)`, making `reconciler`
+a `TypedFunc[reconcile.ClusterAwareRequest]`) an existing controller will automatically act as
+*uniform multi-cluster controller* if a cluster provider is configured.
+It will reconcile resources from cluster `X` in cluster `X`.
+
+For a manager with `cluster.Provider`, the builder _SHOULD_ create a controller
+that sources events **ONLY** from the provider clusters that got engaged with
+the controller.
```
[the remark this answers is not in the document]
sttts: It can be configured by controller. We default to writing a uniform multi-cluster controller, i.e. one that only reacts to provider clusters. It is not common afaik to have the same semantics for both a local cluster (the hub usually) and provider clusters.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+with specific cluster providers, this might be useful.
+
+Optionally, a passed `TypedEventHandler` will be duplicated per engaged cluster if they
+fullfil the following interface:
+
+```golang
+// pkg/handler
+type TypedDeepCopyableEventHandler[object any, request comparable] interface {
+	TypedEventHandler[object, request]
+	DeepCopyFor(c cluster.Cluster) TypedDeepCopyableEventHandler[object, request]
```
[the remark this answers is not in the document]
embik: The reason for this was the attempt to keep existing function signatures stable while enabling them to be multi-cluster aware. A `HandlerConstructor` would probably need a new separate builder function to be passed as argument, so e.g. something `Watches` vs `ClusterAwareWatches` (or whatever).

I'm totally open to changing this if you prefer it.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+	Complete(reconciler)
+```
+
+The builder will chose the correct `EventHandler` implementation for both `For` and `Owns`
+depending on the `request` type used.
+
+With the described changes (use `GetCluster(ctx, req.ClusterName)`, making `reconciler`
+a `TypedFunc[reconcile.ClusterAwareRequest]`) an existing controller will automatically act as
+*uniform multi-cluster controller* if a cluster provider is configured.
+It will reconcile resources from cluster `X` in cluster `X`.
```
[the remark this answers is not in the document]
embik: Would you be comfortable with a way to tell this design to either use a shared queue (e.g. start sources) or to start controllers with a config switch or similar?

designs/multi-cluster.md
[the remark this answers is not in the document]
embik: Copying my comment from [here](https://github.com/kubernetes-sigs/controller-runtime/pull/2746/files/34a44f9356707ac415ca4230796ca074295ec293#r1951291393) since this is the bigger thread: 

> Would you be comfortable with a way to tell this design to either use a shared queue (e.g. start sources) or to start controllers with a config switch [in the manager] or similar?

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+## Motivation
+
+This change is important because:
+- multi-cluster use-cases are becoming more and more common, compare projects
+  like Kamarda, Crossplane or kcp. They all need to write (controller-runtime)
+  controllers that operate on multiple clusters.
+- writing controllers for upper systems in a **portable** way is hard today.
+  Consequently, there is no multi-cluster controller ecosystem, but could and
+  should be.
+- kcp maintains a [controller-runtime fork with multi-cluster support](https://github.com/kcp-dev/controller-runtime)
```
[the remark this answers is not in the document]
embik: I think we tried to cover the high-level part with the first bullet point. The kcp controller-runtime fork is just mentioned to give personal motivation, but I don't think we have to mention it here if that is preferred.

designs/multi-cluster.md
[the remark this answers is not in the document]
alvaroaleman: > I agree that both topologies can have their place. Am not even sure pro/con is helpful. We shouldn't be opinionated, but give the primitives for the developer to decide.

A stated goal of this doc is to avoid `divergence in the ecosystem`, writing that down while at the same time handwaving away comments about this not actually being a good approach and saying we shouldn't be opinionated is not particularly convincing.

Our goal is be make the majority use case simple and other use-cases possible. This is not possible if we refuse to even look into the question what the majority use-case is and default to assuming that the use-case of the author of a design must be the majority use-case.

designs/multi-cluster.md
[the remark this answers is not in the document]
sbueringer: Can you give an example?

I have some trouble understanding the difference between "anything that needs to be done for starting operations on a cluster" vs "operations on the engaged cluster" that both seems to be done or aysnchronously triggered (maybe?) in the Engage func

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+// pkg/controller
+type TypedMultiClusterController[request comparable] interface {
+	cluster.Aware
+	TypedController[request]
+}
+```
+
+The multi-cluster controller implementation reacts to engaged clusters by starting
+a new `TypedSyncingSource` that also wraps the context passed down from the call to `Engage`,
+which _MUST_ be canceled by the cluster provider at the end of a cluster's lifecycle.
```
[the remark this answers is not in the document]
sbueringer: This seems to imply that the context used to call Start() on a Source will be used to stop the Source?

I think stopping the Source is not possible by just canceling this context for our current source implementations (or at least some of them)

IIRC the only similar functionality that we have today is that cancelling the context passed into Cache.Start() will stop all informers.

I also wonder how this works when multiple controllers are sharing the same underlying informer. I think if a controller doesn't exclusively own an informer it also shouldn't just shut it down. Or is my assumption wrong that they usually would share informers? (like it works today when multiple controllers are sharing the same cache & underlying informers).

For additional context. If I apply what we currently do in Cluster API to this proposal, it would be the cluster provider that shuts down the cache and all underlying informers.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+Controllers that should be triggered by events on the hub cluster can continue
+to use `For` and `Owns` and explicitly pass the intention to engage only with the
+"default" cluster (this is only necessary if a cluster provider is plugged in):
+
+```golang
+builder.NewControllerManagedBy(mgr).
+    WithOptions(controller.TypedOptions{
+        EngageWithDefaultCluster: ptr.To(true),
+        EngageWithProviderClusters: ptr.To(false),
```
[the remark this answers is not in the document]
sbueringer: Wondering if it would make sense to just always call Engage and then the Engage function always can just do nothing if it doesn't want to engage a Cluster. This seems the most flexible option.

(If necessary Engage could have a bool return parameter signalling if the controller actually engaged a cluster or not)

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+  - `ctrl.Manager` with `GetCluster(ctx, clusterName string) (cluster.Cluster, error)` and `cluster.Aware`.
+  - `cluster.Cluster` with `Name() string`.
+  We think that the behaviour of these extensions is well understood and hence low risk.
+  Everything else behind the scenes is an implementation detail that can be changed
+  at any time.
+
+## Alternatives
+
+- Multi-cluster support could be built outside of core controller-runtime. This would
+  lead likely to a design with one manager per cluster. This has a number of problems:
```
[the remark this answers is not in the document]
sbueringer: I would be curious about details why it is not possible to implement this outside of CR.

We have implemented a component (called ClusterCache) in Cluster API that seems to come very close to what this design is trying to achieve (apart from that it is Cluster API specific of course). Especially since the generic support was added to CR.

Basically ClusterCache in CAPI:
* discovers Clusters
* maintains a Cache per Cluster
* allows retrieving Clients for a Cluster
* allows adding Watches (kind Sources) for a Cluster

xref: https://github.com/kubernetes-sigs/cluster-api/tree/main/controllers/clustercache

designs/multi-cluster.md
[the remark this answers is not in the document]
sbueringer: I think the confusion comes from that this sentence makes it sound like using another request type will make a controller only compatible with a subset of cluster.Provider implementations. I have some trouble seeing the connection (I got your point though)

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+- Implement the `cluster.Provider` interface by watching the apiserver for logical cluster objects
+  (`LogicalCluster` CRD in kcp).
+- Return a facade `cluster.Cluster` that scopes all operations (client, cache, indexers) 
+  to the logical cluster, but backed by one physical `cluster.Cluster` resource.
+- Add cross-cluster indexers to the physical `cluster.Cluster` object.
+
+### Controller Author without self-interest in multi-cluster, but open for adoption in multi-cluster setups
+
+- Replace `mgr.GetClient()` and `mgr.GetCache` with `mgr.GetCluster(req.ClusterName).GetClient()` and `mgr.GetCluster(req.ClusterName).GetCache()`.
+- Make manager and controller plumbing vendor'able to allow plugging in multi-cluster provider and BYO request type.
```
sbueringer, line 334: Is the idea that something like cert-manager would decide on startup which cluster provider is used and can then only work with one cluster provider at a time?

Phrased differently. Do we also have to support multiple cluster providers at the same time?

sttts: - One provider per manager at a time.
- Low friction to make reconcilers uniform multi-cluster capable (this basically means using the cluster-enabled request and to call `mgr.GetCluster(name)` instead of accessing cluster methods directly).
- If a controller project wants to add support for a number of provider in their repository, this is fine, but not necessarily the goal.
- Instead it should be easy to instantiate the controllers from an alternative main.go with a provider of your choice.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+- Run a controller-runtime controller against a kubeconfig with arbitrary many contexts, all being reconciled.
+- Run a controller-runtime controller against cluster managers like kind, Cluster API, Open-Cluster-Manager or Hypershift.
+- Run a controller-runtime controller against a kcp shard with a wildcard watch.
+
+### Non-Goals/Future Work
+
+- Ship integration for different multi-cluster setups. This should become 
+  out-of-tree subprojects that can individually evolve and vendor'ed by controller authors.
+- Make controller-runtime controllers "binary pluggable".
```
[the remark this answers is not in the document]
sttts: Something like https://pkg.go.dev/plugin to dynamically load providers.

designs/multi-cluster.md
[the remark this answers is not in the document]
sttts: The main "blocker" for this is that the builder is pretty common in 3rdparty controller code. If we do all of this outside of CR, this will very likely mean a fork of `pkg/builder`. I think everything else in the implementation PR could be done outside of CR.

designs/multi-cluster.md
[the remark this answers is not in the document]
sttts: > I think stopping the Source is not possible by just canceling this context for our current source implementations (or at least some of them)

Am curious which you have in mind.

designs/multi-cluster.md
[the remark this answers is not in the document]
sttts: Fair enough. I didn't mean we shouldn't think about which workqueue topology is useful when. I meant that there are good reasons for a joint workqueue in some situations (like when I want to throttle all reconciles in a process because that throughput it limited), and independent ones in other situations (like when e.g. writes to a cluster are the limited factor).

I played with a `TypedFair` queue:
```golang
// Fair is a queue that ensures items are dequeued fairly across different
// fairness keys while maintaining FIFO order within each key.
type Fair TypedFair[any]

// FairnessKeyFunc is a function that returns a string key for a given item.
// Items with different keys are dequeued fairly.
type FairnessKeyFunc[T comparable] func(T) string

// NewFair creates a new Fair instance.
func NewFair(keyFunc FairnessKeyFunc[any]) *Fair {
	return (*Fair)(NewTypedFair[any](keyFunc))
}
```
that could be plugged in here, wrapped by throttling and delays.

sttts:
I already discussed this with @sbueringer: 

I went ahead earlier this week to implement the design outside of controller-runtime: https://github.com/multicluster-runtime/multicluster-runtime. 

As written in the [upper comment](https://github.com/kubernetes-sigs/controller-runtime/pull/2746#discussion_r1953484877), everything but the builder is easy to implement by wrapping controller-runtime concepts. Thanks to the extensive work to get Go generics into CR, those wrappers are easy and natural. 

The builder consists of roughly 500 lines of code with mostly mechanical changes, which should be feasible to regularly rebase onto the latest state in controller-runtime.

Our plan for the moment is to use https://github.com/multicluster-runtime/multicluster-runtime to prove the design with non-trivial projects. Last but not least, it will help to iterate fast and learn on the way. **It is not ruled out that parts of it should flow back into controller-runtime at some point. Definitely it can and should influence the design**.

designs/multi-cluster.md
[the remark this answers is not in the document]
sbueringer: At least source.Kind. The ctx passed into Start can be used to cancel the start process (e.g. WaitForCacheSync) but not the informer

sbueringer:
@sttts What do we want to do with this PR? :)

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+	// Implementers should return an error if they cannot start operations for the given Cluster,
+	// and should ensure this operation is re-entrant and non-blocking.
+	//
+	//	\_________________|)____.---'--`---.____
+	//              ||    \----.________.----/
+	//              ||     / /    `--'
+	//            __||____/ /_
+	//           |___         \
+	//               `--------'
+	Engage(context.Context, Cluster) error
```
n3wscott, line 156: I think `Engage` is fun, but I wonder if `Connect` or `Authenticate` might be more clear what they heck it does.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+**Note:** controller-runtime must provide this cluster-aware request type to
+allow writing *uniform* multi-cluster controllers (see goals), i.e. controllers 
+that both work as single-cluster and multi-cluster controllers against arbitrary
+cluster providers. Think of generic CNCF projects like cert-manager wanting to
+support multi-cluster setups generically without forking the codebase.
+
+### BYO Request Type
+
+Instead of using the new `reconcile.ClusterAwareRequest`, implementations _CAN_ also bring their
```
n3wscott, line 229: love it. I have wanted this for a long time. The current assumption that the key is always `namespace/name` with no type information limits ducktype controllers.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+```
+
+This might be necessary if a BYO `TypedEventHandler` needs to store information about
+the engaged cluster (e.g. because the events do not supply information about the cluster in
+object annotations) that it has been started for.
+
+### Multi-Cluster-Compatible Reconcilers
+
+Reconcilers can be made multi-cluster-compatible by changing client and cache
+accessing code from directly accessing `mgr.GetClient()` and `mgr.GetCache()` to
```
n3wscott, line 254: it would be interesting to move that client access code to the request key?

```
Reconcile(ctx context.Context, req T) (reconcile.Result, error) {
  ...
  req.Client().Status().Update(ctx, update) // defaults to the target cluster
  
  // and if needing a specific cross-cluster client:
  req.Cluster(ref).Client().Update(...)
}
```

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+going through `mgr.GetCluster(req.ClusterName).GetClient()` and
+`mgr.GetCluster(req.ClusterName).GetCache()`.
+
+A typical snippet at the beginning of a reconciler to fetch the client could look like this:
+
+```golang
+cl, err := mgr.GetCluster(ctx, req.ClusterName)
```
n3wscott, line 261: It is not normal to store the manager in the controller impl. So perhaps the pattern would be to store the cluster provider?

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+```golang
+// previous
+builder.TypedControllerManagedBy[reconcile.Request](mgr).
+	Named("single-cluster-controller").
+	For(&corev1.Pod{}).
+	Complete(reconciler)
+
+// new
+builder.TypedControllerManagedBy[reconcile.ClusterAwareRequest](mgr).
```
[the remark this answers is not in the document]
n3wscott: I think we should push for the default to be a ClusterAware everything. The default is cluster aware and there is only one cluster.

designs/multi-cluster.md
sttts: It got messy. I have a branch of multicluster-runtime somewhere implementing that. At the end we decided for strings. One can map strings to anything. The code is much simpler.

designs/multi-cluster.md
sttts: possible, but it's a little strange.

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+*uniform multi-cluster controller* if a cluster provider is configured.
+It will reconcile resources from cluster `X` in cluster `X`.
+
+For a manager with `cluster.Provider`, the builder _SHOULD_ create a controller
+that sources events **ONLY** from the provider clusters that got engaged with
+the controller.
+
+Controllers that should be triggered by events on the hub cluster can continue
+to use `For` and `Owns` and explicitly pass the intention to engage only with the
+"default" cluster (this is only necessary if a cluster provider is plugged in):
```
n3wscott, line 298: The example does not make it clear how I could control which cluster I am reconciling for and which I am watching resources in. I would expect to be able to mix clusters where I am watching resource Foo in cluster A, but I reconcile Bar in cluster B. 

```
builder.NewControllerManagedBy(mgr).
  for(Bar{}).      // Cluster B
  watch(Foo{}) // Cluster A
```

Or is it out of scope to mix clusters?

designs/multi-cluster.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Controller Author with no interest in multi-cluster wanting to old behaviour.
+
+- Do nothing. Controller-runtime behaviour is unchanged.
+
+### Multi-Cluster Integrator wanting to support cluster managers like Cluster API or kind
+
+- Implement the `cluster.Provider` interface, either via polling of the cluster registry
+  or by watching objects in the hub cluster.
+- For every new cluster create an instance of `cluster.Cluster` and call `mgr.Engage`.
```
n3wscott, line 321: Seems like we might need a new kind of ref to link resources between clusters?

designs/multi-cluster.md
n3wscott: json can be a string 😬

designs/multi-cluster.md
embik: Naming is hard, but at least to me "engaging" a cluster feels very natural now that we have had it for a while. Connecting has the implication we are connecting to the cluster, when in reality we're more passing a cluster to the multicluster controllers to also reconcile. Establishing the connection happens earlier, in the provider.

Not sure here, there's probably better language, but overall I think the choice was okay.

embik:
@n3wscott (and others, potentially): Heads up, the multicluster-runtime experiment that @sttts linked above has been transferred under SIG Multicluster and is available here now: https://github.com/kubernetes-sigs/multicluster-runtime. So the design in this document has been implemented out-of-tree and is available to try out.

We should probably transfer this design document to the multicluster-runtime project so it doesn't get lost.

designs/multi-cluster.md
embik: That's not really how it works, you can either trigger events from the "default" hub cluster or from the cluster fleet, but not chose specific clusters because you don't really know if cluster A and cluster B will exist.

With that being said, I believe you should still be able to pass raw sources and similar, so there's maybe a way to still do this.

PR #2750, opened 2024-04-02
🐛 Revert "🌱 Bump github.com/prometheus/client_golang"

### LINKED ISSUE #2742: github.com/prometheus/common v0.48.0 is incompatible with Kubernetes
That dependency has been bumped in #2699. Upstream Kubernetes [uses v0.44](https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/go.mod#L58). In prometheus/common#576, which has been released as v0.48.0, a public symbol has been made private. That public symbol is being used in Kubernetes here: https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/staging/src/k8s.io/component-base/metrics/testutil/metrics.go#L73.

I can workaround that by adding to `replace github.com/prometheus/common => github.com/prometheus/common v0.47.0` in `go.mod`. Is that something to be fixed here by downgrading the dependency or should I live with the replace directive?

twz123:
This reverts PR #2699.

The bump of github.com/prometheus/common to v0.48.0 is incompatible with upstream Kubernetes, which uses v0.44.0. Kubernetes uses a public symbol which has been made private in v0.48.0.

Fixes #2742.


sbueringer:
/ok-to-test

I'll try to look into it tomorrow

alvaroaleman (approved):

PR #2758, opened 2024-04-04
:bug: Prevent race when informers are started more than once

alvaroaleman:
If `Informers` are started a second time, there is a possibility for a data race because it sets a `ctx` field on itself. This write is protected by a mutex, but reads from that field are not.

As it is generally unclear and untested what happens when a cache is started a second time, simply error out to make the user aware that they were starting the cache a second time.

### DIFF
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -18,6 +18,7 @@ package internal
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net/http"
@@ -186,9 +187,13 @@ type Informers struct {
 // Start calls Run on each of the informers and sets started to true. Blocks on the context.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
 func (ip *Informers) Start(ctx context.Context) error {
-	func() {
+	if err := func() error {
 		ip.mu.Lock()
 		defer ip.mu.Unlock()
+
+		if ip.started {
+			return errors.New("Informer already started") //nolint:stylecheck
+		}
 
 		// Set the context so it can be passed to informers that are added later
 		ip.ctx = ctx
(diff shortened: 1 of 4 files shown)


alvaroaleman:
Needs a test

pkg/cache/internal/informers.go
```diff
@@ -186,6 +187,12 @@ type Informers struct {
 // Start calls Run on each of the informers and sets started to true. Blocks on the context.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
 func (ip *Informers) Start(ctx context.Context) error {
+	select {
+	case <-ip.startWait:
+		return errors.New("Informrmer already started")
```
troy0820, line 192: nit: typo
```suggestion
		return errors.New("Informer already started")
```

pkg/cache/internal/informers.go
```diff
@@ -186,6 +187,12 @@ type Informers struct {
 // Start calls Run on each of the informers and sets started to true. Blocks on the context.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
 func (ip *Informers) Start(ctx context.Context) error {
+	select {
+	case <-ip.startWait:
```
sbueringer, line 191: Do we need a lock around this?

alvaroaleman: No, this is a channel

pkg/cache/internal/informers.go
alvaroaleman: fixed

pkg/cache/internal/informers.go
sbueringer: That's not how I meant it :)

What if you have one goroutine in l.198 and another one then just going through l.191?
I would assume the same data race.

Don't we have to make sure an additional goroutine can't go through l.190-l.195 while another one is inside the critical section in l.197-216?

pkg/cache/internal/informers.go
```diff
@@ -186,10 +187,17 @@ type Informers struct {
 // Start calls Run on each of the informers and sets started to true. Blocks on the context.
 // It doesn't return start because it can't return an error, and it's not a runnable directly.
 func (ip *Informers) Start(ctx context.Context) error {
-	func() {
+	if err := func() error {
 		ip.mu.Lock()
 		defer ip.mu.Unlock()
 
+		select {
```
sbueringer, line 194: Not much different but what about simply checking `if ip.started`?

pkg/cache/internal/informers.go
alvaroaleman: Put all of it under the lock

sbueringer:
/lgtm

/hold
If you want Vince to also take a look

PR #2765, opened 2024-04-09
🌱 Downgrade prometheus client libs to versions used in k/k 1.30

sbueringer:


sbueringer:

We recently had some problems when upgrading even further. But I think we should play it safe and use the exact same versions that k/k is using.

alvaroaleman (approved):

aerfio:
> We recently had some problems when upgrading even further. But I think we should play it safe and use the exact same versions that k/k is using.

@sbueringer could you please provide a link to an issue/PR which explains those problems? Thanks

vincepri:
As a general rule we match controller-runtime dependencies with upstream ones, if possible. @sbueringer we might want to add a validation script to keep them in sync

sbueringer:
@aerfio The initial problem we found in this context was this: https://github.com/kubernetes-sigs/controller-runtime/pull/2750

Nobody complained *yet* about the diff we had after this PR, but it might just be a matter of time until someone uses another package where additional errors pop up.  So to play it safe we try to keep the dependencies in sync to avoid compile errors when folks start adopting a new CR version

PR #2766, opened 2024-04-09
 🐛 Fix SubResourceCreateOptions signature in subresource client

Fricounet:
The current signature of SubResourceCreateOptions does not match the interface definition of SubResourceCreateOption because of a typo in the method `ApplyToSubResourceCreate` name. This PR fixes the name.
It also modifies the existing test to actually use the options provided so that if an error like this were to occur again, it would fail to compile.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -523,8 +523,8 @@ func (co *SubResourceCreateOptions) ApplyOptions(opts []SubResourceCreateOption)
 	return co
 }
 
-// ApplyToSubresourceCreate applies the the configuration on the given create options.
-func (co *SubResourceCreateOptions) ApplyToSubresourceCreate(o *SubResourceCreateOptions) {
+// ApplyToSubResourceCreate applies the the configuration on the given create options.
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
thanks

Fricounet:
@alvaroaleman it seems like the api-diff test failed. Does it mean that I should modify the PR title to be a breaking change? Or is it fine to have it failed for a bugfix? Because at the moment, the current API is simply broken for users :thinking:

alvaroaleman:
@Fricounet its fine given this is a bugfix and I can not imagine that someone manually called this method before, which would be the only situation where this would break someone

sbueringer:
No strong opinion, but should this be cherry-picked? Looks like before this PR the option was basically useless

alvaroaleman:
> No strong opinion, but should this be cherry-picked? Looks like before this PR the option was basically useless

I'd leave it unpicked for the moment as in theory it could be breaking and given that no one even noticed this, it doesn't appear to be widely used.

PR #2770, opened 2024-04-11
:book: Use `Should` instead of `To` in `komega` samples

afritzler:
As the `AsyncAssertion` of `gomega` does not support `To` the samples in the documentation have been updated to use `Should`.


sbueringer:
Nice catch, thx!

/lgtm
/approve

/ok-to-test

PR #2771, opened 2024-04-11
✨ client: Add client-wide fieldManager

ahmetb:
## Summary

This PR adds a new `client` function `WithFieldOwner` that wraps a `client.Client` and adds a `client.FieldOwner` option to all writes from this client. If additional FieldOwner options are specified on methods of this client, the value specified here will be overridden.

## Why do we need it?

I have found myself trying to distinguish between different reconcilers in the same controller-manager process for both troubleshooting purposes (i.e having `managedFields` show the specific controller) and prevent colocated reconcilers from managing the same field.

The alternative was to add `client.FieldManager` to each individual call, which is tedious and error-prone. Instead, I would prefer to have the client use the same fieldManager across all its write calls.

### DIFF
--- a/pkg/client/fieldmanager_test.go
+++ b/pkg/client/fieldmanager_test.go
@@ -1,0 +1,148 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client_test
+
+import (
+	"context"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


vincepri (approved):

ahmetb:
/cherrypick release-0.17
/cherrypick release-0.16
per @alvaroaleman since it's a net-new change.

pkg/client/fieldmanager_test.go
```diff
@@ -0,0 +1,148 @@
+/*
```
sbueringer, line 1: @ahmetb Can you please adjust the filename to fieldowner_test.go? (just for main is okay)

pkg/client/fieldowner.go
```diff
@@ -0,0 +1,106 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// WithFieldOwner wraps a Client and adds the fieldOwner as the field
+// manager to all write requests from this client. If additional [FieldOwner]
+// options are specified on methods of this client, the value specified here
+// will be overridden.
+func WithFieldOwner(c Client, fieldOwner string) Client {
+	return &clientWithFieldManager{
+		manager: fieldOwner,
```
sbueringer, line 33: nit: Maybe we should call this owner or fieldOwner consistently (compared to manager in one client and owner in the other)

pkg/client/fieldmanager_test.go
ahmetb: sent #2777

PR #2783, opened 2024-04-16
⚠️ Source, Event, Predicate, Handler: Add generics support

alvaroaleman:
This change adds support for generics into the `Event`, `Predicate` and `Handler` chain. The motivation for this is explained in https://github.com/kubernetes-sigs/controller-runtime/issues/2214, but basically the current `client.Object` typing frequently requires type assertions when using custom handlers or predicates and is completely inappropriate when building a controller that is not based on Kubernetes objects.

It aims to minimize breaking changes as much as possible. It does so in two parts:

* Source: Move Handler and Predicate out from Start and into constructor: This is needed because of https://github.com/golang/go/issues/49085, otherwise we either
    * Can only support one type per `Controller` which makes no sense
    * Would need to change `controller.Watch` to take a wrapper type that encapsulates the generic part and hides it from the controller. This is pretty much the same, but more complicated
    * This change means that the concept of a `Handler` and `Predicate` effectively becomes optional for a `Source` which I think makes sense, as a custom source may decide it is easier not to implement these. The existing concept is also visibly shoehorned insofar as that Event/PredicateHandler have Create/Update/Delete and Generic variants, whereas the first three are only used in the context of a kube watch and the latter only outside of it
 * Source, Event, Predicate, Handler: Add generics support. This is achieved by adding a `Typed` variant to all constructors/types in the latter three packages and making the `Kind` and `Channel` source constructors generic. This means that this change is 100% backwards-compatible in the `builder`

This change will require users to adjust their code only if:
* They directly call `controller.Watch` rather than the builder
* They use a `Channel` source
* They call `builder.WatchRawSource` with `...builder.WatchesOption`
* They built their own `source.Source`

The above is IMHO likely a relatively small fraction of our users and can be assumed to have advanced knowledge of controller-runtime.

As a side-effect of keeping the `builder` compatible, users of it will not see this feature. Finding a way to support this in the builder outside of `WatchesRawSource` is left as a future task and will likely not have a solution with a nice UX due to https://github.com/golang/go/issues/49085

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2214


sbueringer:
Just did a first high-level review. Overall pretty nice!

pkg/internal/controller/controller.go
```diff
@@ -98,9 +96,7 @@ type Controller struct {
 
 // watchDescription contains all the information necessary to start a watch.
 type watchDescription struct {
```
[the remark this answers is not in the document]
alvaroaleman: So I actually realized I made a mistake with how I designed it with having methods in the internalSource to enable configuring a metadata watch after the fact: We assert the incoming type to the sources generic type and that will completely fail if object is a metadata object. So In re-designed this to not store a source at all, but all parts to construct a source in the `watchDescription`, that way we do not have to manipulate it after constructing it and IMHO its much cleaner.

pkg/internal/source/kind.go
```diff
@@ -4,36 +4,49 @@ import (
...
 
 // Start is internal and should be called only by the Controller to register an EventHandler with the Informer
 // to enqueue reconcile.Requests.
-func (ks *Kind) Start(ctx context.Context, handler handler.EventHandler, queue workqueue.RateLimitingInterface,
-	prct ...predicate.Predicate) error {
+func (ks *Kind[T]) Start(ctx context.Context, queue workqueue.RateLimitingInterface) error {
+	ks.startedLock.Lock()
+	defer ks.startedLock.Unlock()
+
 	if ks.Type == nil {
```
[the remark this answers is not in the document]
alvaroaleman: Added that

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
Danil-Grigorev: There is a big disclaimer on watchDescription that it is not supposed to be used alone, and that the `PartialObjectMetadata` should be used. I believe that this field even does not make any effect, so I don’t think it should be carried over into new implementations. Generics will work even better with `PartialObjectMetadata` passed. I mentioned this on the event handler PR https://github.com/kubernetes-sigs/controller-runtime/pull/2698#issue-2160848733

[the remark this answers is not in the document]
alvaroaleman: I am realizing in the meantime it coincidentally worked, because client.Object is implemented by both the metaobject and any type, so the assertion actually succeeded.

pkg/handler/enqueue.go
```diff
@@ -33,13 +35,18 @@ type empty struct{}
...
+// TypedEnqueueRequestForObject enqueues a Request containing the Name and Namespace of the object that is the source of the Event.
+// (e.g. the created / deleted / updated objects Name and Namespace).  handler.TypedEnqueueRequestForObject is used by almost all
+// Controllers that have associated Resources (e.g. CRDs) to reconcile the associated Resource.
+type TypedEnqueueRequestForObject[T client.Object] struct{}
 
 // Create implements EventHandler.
-func (e *EnqueueRequestForObject) Create(ctx context.Context, evt event.CreateEvent, q workqueue.RateLimitingInterface) {
-	if evt.Object == nil {
+func (e *TypedEnqueueRequestForObject[T]) Create(ctx context.Context, evt event.TypedCreateEvent[T], q workqueue.RateLimitingInterface) {
+	if isNil(evt.Object) {
```
Danil-Grigorev, line 49: Is there a historical reason for this nil check? Can API server return an empty Object still? Maybe this can just be avoided, as having T to implement comparable will make it incompatible with kube resources.

alvaroaleman: Yeah, its historic, I do not want to break anyone who is relying on this

pkg/event/event.go
```diff
@@ -18,38 +18,55 @@ package event
...
-	Object client.Object
+	Object T
 }
 
-// UpdateEvent is an event where a Kubernetes object was updated.  UpdateEvent should be generated
-// by a source.Source and transformed into a reconcile.Request by an handler.EventHandler.
-type UpdateEvent struct {
+// TypedUpdateEvent is an event where a Kubernetes object was updated. TypedUpdateEvent should be generated
+// by a source.Source and transformed into a reconcile.Request by an handler.TypedEventHandler.
+type TypedUpdateEvent[T any] struct {
```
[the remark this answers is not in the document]
Danil-Grigorev: Rather leave Event structures for non generic types without any change, and introduce a new handler for generic types. I did it this way: https://github.com/Danil-Grigorev/controller-runtime/blob/4d770047de3a10110f5ded3bce46c178d68d0500/pkg/handler/eventhandler.go#L61-L73

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
Danil-Grigorev: I actually got rid of the `watchDescription` in favor of building `source.Source` via changed interfaces before starting it. It can be used like:
```go
var src source.PrepareSyncing
var prepared source.SyncingSource
prepared = src.Prepare(handler, predicates…)
```
Then you can do something like
```go
c.Watch(source.Kind(mgr.GetCache(), &appsv1.ReplicaSet{}).Prepare(&handler.EnqueueRequestForObject{}))
```
I opened a draft #2784 to compare implementations, also did a test rewrite [here](https://github.com/Danil-Grigorev/cluster-api/pull/1) (to see how it feels) for cluster-api, as it has lots of predicates and map functions where generics will be the most useful.

pkg/event/event.go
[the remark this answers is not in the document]
alvaroaleman: What is the advantage of that? That means sources have to be able to deal with either or or there needs to be some sort of adapter to convert, both of which is very inconvenient. How does this improve the overall UX?

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
alvaroaleman: This PR avoids that entire issue by moving all args into source construction. IMHO that is a lot simpler than requiring sources to implement a `Prepare`.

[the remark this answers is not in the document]
alvaroaleman: @Danil-Grigorev your approach requires something like this:
```
builder.Add(builder.For(b,
		&bootstrapv1.KubeadmConfig{},
		predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue, &bootstrapv1.KubeadmConfig{}),
	)).
```

IMHO something like that is a huge no-no:
* We break every single user of controller-runtime bigtimes
* It is super annoying to have to have the custom wrappers for methods like `For` etc, especially in contexts where none of this matters like you just want to enqueue the object, its owner etc - Which is the vast majority use case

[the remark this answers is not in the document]
Danil-Grigorev: But you can still use all of the existing methods of the builder in a non-generic context, I don’t see how is this breaking. And additional wrappers allow you to typecheck all your event handlers and predicates on the compiler level within the method invocation. Which is extremely useful if you go beyond 1-2 predicate methods and do complex logic with And/Or. I noticed this while doing CAPI code rewrite. Adapters are mainly made for scenarios when you explicitly don’t want to rewrite your code to generics (like readability concerns). I would love to get rid of the builder requirement in these wrappers, it is only needed for passing validation (have For/Owns used at least once), otherwise passing a Manager would suffice. It is after all always available in `SetupWithManager` context.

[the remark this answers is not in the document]
alvaroaleman: @Danil-Grigorev okay I missed that this is optional. But even then, the approach seems more complicated, you need a wraper to build a source. Why not just change things so that source construction itself takes all required args rather than needing an additional constructor func on top? I can see that it might be nice to optionally have the builder `For` and `Own` but that is it. Also, there is btw already `buillder.WatchesRawSource` of which your `Add` is a duplicate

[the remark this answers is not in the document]
Danil-Grigorev: `Source` interface is being weakened by the rewrite of the start method. I’d like to preserve source requirements to pass predicates and a handler, maybe even just by a composite interface embedding updated `Source`. You are right, `WatchesRawSource` is duplicate of `Add`, but this allows for both to be stored in a single array without type erasure. Here is also an example for `builder.Watches` where every predicate or map method should always route to clusterv1.Cluster. This removes margin for error:
```go
Add(builder.Watches(mgr,
	&clusterv1.Cluster{},
	handler.EnqueueRequestsFromObjectMap(r.ClusterToKubeadmControlPlane),
	predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue, &clusterv1.Cluster{}),
	predicates.ClusterUnpausedAndInfrastructureReady(ctrl.LoggerFrom(ctx)),
)
```

My original motivation to go for a more complicated way was the hope to have as little breaking changes as possible, and yet allow for optional use of generics as a sort of first-class feature.

pkg/event/event.go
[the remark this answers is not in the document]
Danil-Grigorev: Mostly for the ability to infer type, I’d say. Anytime that you specify `[anything]` before the method, the syntax highlighting breaks. Might be only my personal observation, env, etc. The goal is that as controller-runtime is a library, exposing a list of interfaces to extend (in case you need this), having one which unconditionally requires to use generics might make it hard for newcomers to implement custom logic. With a method one can always use just `*corev1.Pod` as an arg to custom EventHandler implementation and won’t have to use advanced features immediately. And wrapping EventHandle into constructors to eliminate the issue didn’t make sense to me. I can’t imagine how a structure for events may grow beyond just hosting old/new objects, which is low enough to consider a method.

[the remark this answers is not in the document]
Danil-Grigorev: I’m trying to find a reason why I abandoned aliasing in the first place. Yes, it seems to be working, and having to specify a typedEvent[*Pod] is a minor inconvenience. My guess would be that the issue discussion lead to overcomplicating the solution. But I see a difference with using `Source().Prepare()` that the current builder logic was not refactored in my implementation. This means there will not be issues with existing functionality

pkg/handler/enqueue_mapped.go
```diff
@@ -40,44 +44,58 @@ type MapFunc func(context.Context, client.Object) []reconcile.Request
...
+// of reconcile.Requests on each Event.  The reconcile.Requests may be for an arbitrary set of objects
+// defined by some user specified transformation of the source Event.  (e.g. trigger Reconciler for a set of objects
+// in response to a cluster resize event caused by adding or deleting a Node)
+//
+// EnqueueRequestsFromTypedMapFunc is frequently used to fan-out updates from one object to one or more other
+// objects of a differing type.
+//
+// For TypedUpdateEvents which contain both a new and old object, the transformation function is run on both
+// objects and both sets of Requests are enqueue.
+func EnqueueRequestsFromTypedMapFunc[T any](fn TypedMapFunc[T]) TypedEventHandler[T] {
```
Danil-Grigorev, line 60: Here is one problem as I can see. There should be an adapter method for `EventHandler` for a non-generic controller using type specific generic map method. Something like:
```go
func EnqueueRequestsFromTypedMapFunc[T client.Object](fn TypedMapFunc[T]) EventHandler
```
The problem now is that `EventHandler = TypedEventHandler[client.Object]` and you can’t type cast from one generic type to another, regardless of the fact that both implement `client.Object`. That’s where the simpler form of type aliasing broke for me, and where this (obviously simpler) design is a breaking change. Sometimes you can’t pair generics with a project due to reasons like the one we are facing - using builder or any sort of nested structures. Go generics just don’t work where there is a type ambiguity. But using generics should not be either/or decision either, it just needs a way to make a split at some point where it no longer works or does not make sense, I think.

As for real world example where this is important - I rebased CAPI testing PR on your fork + new builder methods, all works well except one common [method](https://github.com/kubernetes-sigs/cluster-api/pull/10464/files#diff-ef84e95c01630a75038b4d2cda0709d9d1d3d556459304fb39789913fb98eeaeR332). It is the only blocker for it to compile and work, but it is a deal breaker if this has to be continued (cc @sbueringer):
```go
return r.Tracker.Watch(ctx, remote.WatchInput{
    Name:         "machinepool-watchNodes",
 	Cluster:      util.ObjectKey(cluster),
 	Watcher:      &controller.ControllerAdapter{Controller: r.controller},
 	Kind:         &corev1.Node{},
 	EventHandler: handler.EnqueueRequestsFromMapFunc(r.nodeToMachinePool), // Does not compile because expects TypedMapFunc[client.Object] and gets TypedMapFunc[*corev1.Node]
 })

func (r *MachinePoolReconciler) nodeToMachinePool(ctx context.Context, node *corev1.Node) []reconcile.Request {
```
And there is no workaround except for having `TypedEventHandler` and `EventHandler` being separate interfaces with different methods working on `client.Object` and `T` simultaneously. Here is [how](https://github.com/kubernetes-sigs/controller-runtime/pull/2784/files#diff-37a34307ac3b7d0c687b0b2fd11f1a595a7987c90fc3f3b85881c3fdddfae981) I did it originally. I didn’t see a point in having `TypedEvent` structure either, as a simplification. Another user story for this - one less `controller-runtime` dependency for reusable business logic around event mapping exported in a package at the API level.

alvaroaleman: Cluster-API was also bumped here and it doesn't look like the problem you are describing happened there: https://github.com/kubernetes-sigs/cluster-api/pull/10460

> The problem now is that EventHandler = TypedEventHandler[client.Object] and you can’t type cast from one generic type to another,

You can not type cast but you can use them interchangably. I still do not quite understand what this breaks that your approach doesn't. I am not very familiar with cluster-api, can you give a simpler example?

sbueringer: I'll take a look

sbueringer: > Cluster-API was also bumped here and it doesn't look like the problem you are describing happened there: https://github.com/kubernetes-sigs/cluster-api/pull/10460

I think this didn't happen here because Christian rebased on this PR, and not on "your fork + new builder methods"

sbueringer: @Danil-Grigorev Can you provide more details how this func is broken? (and push a branch somewhere where it is broken)

alvaroaleman: One detail that maybe is what you are referring to but I am not sure: It makes a difference if you do
```
type foo = bar[something]
```

or

```
type foo bar[something]
```

The latter means that you do not get the methods of `bar` and as a result, `foo` can not be used in place of `bar[something]`

But yeah, I also don't really understand the example given so I don't know if this might be what you mean.

Danil-Grigorev: Yes, as I mentioned, I just had to stick with non generic `nodeToMachinePool` method which kept using `client.Object` and that probably it wasn’t the right example, because `nodeToMachinePool` is only used once - in the tracker setup.

My concern is about more complex scenarios, when the same method has to be used with generic implementation elsewhere. It can’t be done without an adapter or code duplication purely for type casting. Which might lead to decision against using generics due to added complexity in having use methods interchangeably.

The example I provided is good in outlining the second problem related to golang/go#49085 as in builder pattern UX changes. We can’t make `Tracker` a generic method with specialization on `*corev1.Node` for the CAPI use case to make it simple, because the `Tracker` is stored in `MachinePoolReconciler` which works with `MachinePools` - it would look strange and may block `MachinePoolReconciler` from being a generic method on `MachinePoolReconciler[*MachinePool]` for example if needed.

This is theoretical, I’m struggling to find a real world example which would explain the point better, still looking though. I think `controller-runtime` should take a couple more steps further to make the library UX better aligned with generics, despite the golang/go#49085 limitation, and get the design of how it looks in the end with possible edge scenarios covered.

pkg/internal/controller/controller_test.go
```diff
@@ -164,7 +163,7 @@ var _ = Describe("controller", func() {
 			c = &cacheWithIndefinitelyBlockingGetInformer{c}
 			ctrl.startWatches = []watchDescription{{
 				src: &singnallingSourceWrapper{
-					SyncingSource: source.Kind(c, &appsv1.Deployment{}),
+					SyncingSource: source.Kind[client.Object](c, &appsv1.Deployment{}, &handler.EnqueueRequestForObject{}),
```
sbueringer, line 166: I assume we don't use a TypedEnqueueRequestForObject to show that we don't have to?

pkg/internal/controller/controller_test.go
```diff
@@ -227,20 +226,21 @@ var _ = Describe("controller", func() {
 				Object: p,
 			}
 
-			ins := &source.Channel{Source: ch}
-			ins.DestBufferSize = 1
+			ins := source.Channel(
```
sbueringer, line 229: Q: Did you intentionally drop `ins.DestBufferSize = 1`?

pkg/source/source.go
```diff
@@ -58,54 +54,64 @@ type SyncingSource interface {
...
-	// Source is the source channel to fetch GenericEvents
-	Source <-chan event.GenericEvent
+	// source is the source channel to fetch GenericEvents
+	source <-chan event.TypedGenericEvent[T]
 
-	// dest is the destination channels of the added event handlers
-	dest []chan event.GenericEvent
+	handler handler.TypedEventHandler[T]
+
+	predicates []predicate.TypedPredicate[T]
```
sbueringer, line 89: Would it make sense to preserve this option?

sbueringer (commented):
Overall looks good

pkg/builder/controller.go
```diff
@@ -135,8 +135,17 @@ type WatchesInput struct {
 // This is the equivalent of calling
 // WatchesRawSource(source.Kind(cache, object), eventHandler, opts...).
```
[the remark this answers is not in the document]
alvaroaleman: Updated

pkg/builder/controller.go
```diff
@@ -302,27 +306,29 @@ func (blder *Builder) doWatch() error {
...
+		src := source.Kind(blder.mgr.GetCache(), obj, hdler, allPredicates...)
+		if err := blder.ctrl.Watch(src); err != nil {
 			return err
 		}
 	}
 
 	// Do the watch requests
-	if len(blder.watchesInput) == 0 && blder.forInput.object == nil {
+	if len(blder.watchesInput) == 0 && blder.forInput.object == nil && len(blder.rawSources) == 0 {
 		return errors.New("there are no watches configured, controller will never get triggered. Use For(), Owns() or Watches() to set them up")
```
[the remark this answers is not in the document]
alvaroaleman: Added them

pkg/builder/controller.go
```diff
@@ -302,27 +306,29 @@ func (blder *Builder) doWatch() error {
...
 	}
+	allPredicates := append([]predicate.Predicate(nil), blder.globalPredicates...)
```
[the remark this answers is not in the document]
alvaroaleman: That is just how I found it so I left it like that. Changed it to also be inside the loop.

pkg/builder/controller.go
```diff
@@ -302,27 +306,29 @@ func (blder *Builder) doWatch() error {
...
-		allPredicates := append([]predicate.Predicate(nil), blder.globalPredicates...)
-		allPredicates = append(allPredicates, w.predicates...)
-		if err := blder.ctrl.Watch(w.src, w.eventHandler, allPredicates...); err != nil {
+		allPredicates := append(allPredicates, w.predicates...)
+		if err := blder.ctrl.Watch(source.Kind(blder.mgr.GetCache(), projected, w.handler, allPredicates...)); err != nil {
+			return err
+		}
+	}
+	for _, src := range blder.rawSources {
+		if err := blder.ctrl.Watch(src); err != nil {
```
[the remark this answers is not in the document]
alvaroaleman: Added it to the godocs

pkg/handler/enqueue_owner.go
```diff
@@ -32,12 +32,12 @@ import (
...
 )
 
-var _ EventHandler = &enqueueRequestForOwner{}
+var _ EventHandler = &enqueueRequestForOwner[client.Object]{}
 
 var log = logf.RuntimeLog.WithName("eventhandler").WithName("enqueueRequestForOwner")
 
 // OwnerOption modifies an EnqueueRequestForOwner EventHandler.
-type OwnerOption func(e *enqueueRequestForOwner)
+type OwnerOption func(e enqueueRequestForOwnerIface)
```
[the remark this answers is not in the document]
alvaroaleman: Updated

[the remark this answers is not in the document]
alvaroaleman: Updated

pkg/handler/eventhandler.go
```diff
@@ -41,65 +42,88 @@ import (
...
+//
+// Identical reconcile.Requests will be batched together through the queuing mechanism before reconcile is called.
+//
+// * Use TypedEnqueueRequestForObject to reconcile the object the event is for
+// - do this for events for the type the Controller Reconciles. (e.g. Deployment for a Deployment Controller)
+//
+// * Use TypedEnqueueRequestForOwner to reconcile the owner of the object the event is for
+// - do this for events for the types the Controller creates.  (e.g. ReplicaSets created by a Deployment Controller)
+//
+// * Use TypedEnqueueRequestsFromMapFunc to transform an event for an object to a reconcile of an object
```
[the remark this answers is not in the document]
alvaroaleman: I ended up renaming it to `TypedEnqueueRequestFromMapFunc` just because I recall that when I migrated our internal codebase I stumbled upon this inconsistency in naming a couple of times. I think the previous name sounds nicer, but consistency is IMHO more important to make this easy to find

pkg/internal/controller/controller.go
```diff
@@ -98,9 +96,7 @@ type Controller struct {
...
-	src        source.Source
-	handler    handler.EventHandler
-	predicates []predicate.Predicate
+	src source.Source
```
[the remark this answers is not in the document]
alvaroaleman: You are right, it is not needed anymore, removed it

pkg/internal/controller/controller_test.go
alvaroaleman: No, just because this was the smaller change :P

pkg/source/source.go
alvaroaleman: The reason I ended up not preserving it is that I unexported source.Channel and exported a constructor for it instead. That was done to avoid silently (as in: Without compilation failures) breaking ppl who do not pass in a `Handler` which is now mandatory (it was also mandatory previously, but passed in through `Start()`).

The constructor func already uses its variadic arg for `Predicates` as they are optional. This setting would also be optional, so in order to make that happen, we would need to make the variadic arg some form of `ChannelOption` and that would make it more annoying to pass in `Predicates`. I assumed that this is rarely if ever configured and thus that its best to use the variadic arg for the `Predicates`. I don't have any evidence one way or the other though and not strong take, happy to change this if you think we should keep this configurable.

pkg/handler/eventhandler.go
```diff
@@ -41,65 +42,88 @@ import (
 //
 // Unless you are implementing your own EventHandler, you can ignore the functions on the EventHandler interface.
 // Most users shouldn't need to implement their own EventHandler.
-type EventHandler interface {
+type EventHandler TypedEventHandler[client.Object]
+
+// TypedEventHandler enqueues reconcile.Requests in response to events (e.g. Pod Create). TypedEventHandlers map an Event
+// for one object to trigger Reconciles for either the same object or different objects - e.g. if there is an
+// Event for object with type Foo (using source.KindSource) then reconcile one or more object(s) with type Bar.
```
Danil-Grigorev, line 49:
```suggestion
// Event for object with type Foo (using source.Kind) then reconcile one or more object(s) with type Bar.
```

pkg/internal/source/event_handler.go
```diff
@@ -43,19 +43,19 @@ func NewEventHandler(ctx context.Context, queue workqueue.RateLimitingInterface,
 }
 
 // EventHandler adapts a handler.EventHandler interface to a cache.ResourceEventHandler interface.
-type EventHandler struct {
+type EventHandler[T client.Object] struct {
```
Danil-Grigorev, line 46: This can also use `EventHandler[T any]` without changes to allow watching non-kubernetes resources.

pkg/predicate/predicate.go
```diff
@@ -153,17 +180,35 @@ func (ResourceVersionChangedPredicate) Update(e event.UpdateEvent) bool {
...
+//
+// Caveats:
+//
+// * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
+// E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
+// For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
+//
+// * With this predicate, any update events with writes only to the status field will not be reconciled.
+// So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.
+type TypedGenerationChangedPredicate[T client.Object] struct {
```
Danil-Grigorev, line 201: For this predicate and for annotation + labels change it might make sense to have a smaller scoped interfaces? This only needs .GetGeneration() and the other GetAnnotations() or GetLabels().

pkg/source/source.go
```diff
@@ -31,11 +32,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-const (
-	// defaultBufferSize is the default number of event notifications that can be buffered.
-	defaultBufferSize = 1024
```
Danil-Grigorev, line 36: Doesn’t this upset linter now? Maybe it shouldn’t just checking.

pkg/source/source.go
```diff
@@ -212,12 +222,11 @@ func (is *Informer) String() string {
 var _ Source = Func(nil)
 
 // Func is a function that implements Source.
-type Func func(context.Context, handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
+type Func func(context.Context, workqueue.RateLimitingInterface) error
```
Danil-Grigorev, line 225: It does implement source, but does it make sense? There will be no handler available to funnel the events, no place to store it, maybe an example of how this will work now be helpful? Or just remove it

pkg/internal/source/kind.go
```diff
@@ -4,47 +4,55 @@ import (
...
+
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // Kind is used to provide a source of events originating inside the cluster from Watches (e.g. Pod Create).
-type Kind struct {
+type Kind[T client.Object] struct {
```
Danil-Grigorev, line 22: Technically this can also be any, if the `Kind` will be able to accept GVK and use `GetInfromerForKind`. I suppose users will have to write their own `[any]` source for now.

pkg/builder/controller.go
```diff
@@ -302,27 +308,29 @@ func (blder *Builder) doWatch() error {
...
 	}
 	for _, w := range blder.watchesInput {
-		// If the source of this watch is of type Kind, project it.
-		if srcKind, ok := w.src.(*internalsource.Kind); ok {
-			typeForSrc, err := blder.project(srcKind.Type, w.objectProjection)
-			if err != nil {
-				return err
-			}
-			srcKind.Type = typeForSrc
+		projected, err := blder.project(w.obj, w.objectProjection)
```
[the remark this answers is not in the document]
alvaroaleman: > Projection might not work for different sources.

This is only used to construct a soruce.Kind.

> I think here a [change](https://github.com/kubernetes-sigs/controller-runtime/pull/2685/files#diff-e77b9468ab935d5ea4d5cdae3b994114bada17df0001570e4b1436419afd50ccR315) suggested in https://github.com/kubernetes-sigs/controller-runtime/pull/2685 will fit better.

I disagree. Both what we did previously and what is done over there is IMHO hacky and will not work after the Source was started. Deferring the creation and creating it with the right args rather than manipulating it later on is IMHO much cleaner.

pkg/internal/source/event_handler.go
alvaroaleman: This is internal and only used by `source.Kind` which only works with `client.Object`, so there is no reason to have a different type constraint here

pkg/predicate/predicate.go
alvaroaleman: Changed them to use metav1.Object

pkg/source/source.go
alvaroaleman: > It does implement source, but does it make sense? There will be no handler available to funnel the events, no place to store it, maybe an example of how this will work now be helpful?

As mentioned earlier, there is no need for the concept of a handler when not using a kube watch. A simple example of how to use this would be something like the below, that being said the point of this is to simplify implementing the interface for whatever people want, so if we added the example in code, people might think that is how its supposed to be used but its only one of potentially many ways:
```
queueInitialized := make(chan struct{})
var workqueue workqueue.RateLimitingInterface

ctrl.Watch(source.Func(func(_ context.Context, w workqueue.RateLimitingInterface) error {
     workqueue = w
     close(queueInitialized)
}))

add := func(in types.NamespacedName){
    <- queueInitialized
    workqueue.Add(in)
}
```

pkg/internal/source/kind.go
alvaroaleman: > Technically this can also be any, if the Kind will be able to accept GVK and use GetInfromerForKind. I suppose users will have to write their own [any] source for now.

Not sure I understand. The `Source` has to have a type in the golang sense. There is no way to express a GVK as a type in the golang sense other than using a `client.Object` from which we then derive the GVK.

Danil-Grigorev: This is not what I mean, GVK can’t be used as a type, true. But if you are able to supply a GVK externally (a field in the struct) then the generic type supplied may be anything. GVK can be constructed manually, and the object Kind will be able to work with something which does not need to be anywhere close to `client.Object`. It’s for the case of “build arbitrary controllers that might not even act on kube resources” as I saw it from the issue discussion.

alvaroaleman: > It’s for the case of “build arbitrary controllers that might not even act on kube resources” as I saw it from the issue discussion.

Right, but `source.Kind` isn't meant for that but specifically for kube resources.

pkg/handler/enqueue_mapped.go
Danil-Grigorev: Sounds good to me. Tried this once again and this approach seems to work:
```go
func convert[T any](fn TypedMapFunc[T]) MapFunc {
	return func(ctx context.Context, obj client.Object) (req []reconcile.Request) {
		if typed, ok := obj.(T); !ok {
			return req
		} else {
			return fn(ctx, typed)
		}
	}
}

func EnqueueRequestsFromTypedMapFunc[T any](fn TypedMapFunc[T]) EventHandler {
	return &enqueueRequestsFromMapFunc[client.Object]{
		toRequests: convert(fn),
	}
}
```

pkg/predicate/predicate.go
```diff
@@ -358,3 +409,41 @@ func LabelSelectorPredicate(s metav1.LabelSelector) (Predicate, error) {
...
+		v.Kind() == reflect.Map ||
+		v.Kind() == reflect.Chan ||
+		v.Kind() == reflect.Func) && v.IsNil()) {
+		return true
+	}
+	return false
+}
+
+// ToTyped converts a Predicate to a TypedPredicate.
+func ToTyped[T client.Object](p Predicate) TypedPredicate[T] {
```
Danil-Grigorev, line 426: I think this is method is not needed. As generic sources are not released yet, anyone considering to use them will have to refactor their code after it is supported. And using this method negates benefits the more robust compile time type checking system brings, I don’t think anyone will use it. On the other hand something like the `ToUntyped` may be useful, if you don’t want to immediately refactor. Use case is similar to `AsReconciler` - just to spare a couple lines of code which will compound as time goes on and project grows.

This can be done as a separate PR as well if you think it’s better.

```go
// ToUntyped converts a TypedPredicate to a regular Predicate.
func ToUntyped[T client.Object](p TypedPredicate[T]) Predicate {
	return TypedFuncs[client.Object]{
		CreateFunc: func(e event.TypedCreateEvent[client.Object]) bool {
			obj, ok := e.Object.(T)
			return ok && p.Create(event.TypedCreateEvent[T]{Object: obj})
		},
		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
			objOld, okOld := e.ObjectOld.(T)
			objNew, okNew := e.ObjectNew.(T)
			return okOld && okNew && p.Update(event.TypedUpdateEvent[T]{
				ObjectOld: objOld,
				ObjectNew: objNew,
			})
		},
		DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
			obj, ok := e.Object.(T)
			return ok && p.Delete(event.TypedDeleteEvent[T]{
				Object:             obj,
				DeleteStateUnknown: e.DeleteStateUnknown,
			})
		},
		GenericFunc: func(e event.TypedGenericEvent[client.Object]) bool {
			obj, ok := e.Object.(T)
			return ok && p.Generic(event.TypedGenericEvent[T]{
				Object: obj,
			})
		},
	}
}
```

pkg/handler/example_test.go
```diff
@@ -68,19 +68,20 @@ func ExampleEnqueueRequestForOwner() {
...
-					Namespace: a.GetNamespace(),
-				}},
-				{NamespacedName: types.NamespacedName{
-					Name:      a.GetName() + "-2",
-					Namespace: a.GetNamespace(),
-				}},
-			}
-		}),
+		source.Kind(mgr.GetCache(), &appsv1.Deployment{},
+			handler.TypedEnqueueRequestsFromMapFunc[*appsv1.Deployment](func(ctx context.Context, a *appsv1.Deployment) []reconcile.Request {
```
Danil-Grigorev, line 72:
```suggestion
			handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, a *appsv1.Deployment) []reconcile.Request {
```

Danil-Grigorev (commented):
Overall looks good to me, after some more time spent reviewing, only a couple more minor things I still think are nice to haves.

pkg/handler/example_test.go
alvaroaleman: Updated

pkg/predicate/predicate.go
alvaroaleman: Hm you are right, I think I used this at some point in the builder but apparently not anymore. Removed it, we can still add adapters like that later if we find that there is demand.

alvaroaleman:
> Instead of passing both a handler and a set of predicates to each source, we could also use the predicates to filter what is > passed to the handler and return a new handler. I created a POC for this here: https://github.com/alvaroaleman/controller-runtime/pull/1.
>PTAL and let me know what you think.

I ended up adding an option type for `source.Channel` after finding that there are some examples where ppl set a custom buffer size, e.g. here: https://github.com/alibaba/hybridnet/blob/b847fd2505c05ec4f48c4d387f511401fb6b4f0b/pkg/controllers/networking/subnetstatus_controller.go#L145

It doesn't seem very common but it would be bad to have to break ppl a second time to add this.

WRT the idea of modeling `Predicates` as a `handler` wrapper you are right that that is possible but I think it might cause a lot of confusion if we do and enforce that.

pkg/builder/controller.go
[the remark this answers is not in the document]
sbueringer: It actually was inside the for loop. But I think that just happened during some refactoring :D

pkg/source/source.go
```diff
@@ -58,54 +55,92 @@ type SyncingSource interface {
...
 }
 
-var _ Source = &Channel{}
+var _ Source = &channel[string]{}
+
+// ChannelOpt allows to configure a source.Channel.
+type ChannelOpt[T any] func(*channel[T])
+
+// WithPrededicates adds the configured predicates to a source.Channel.
+func WithPrededicates[T any](p ...predicate.TypedPredicate[T]) ChannelOpt[T] {
```
sbueringer, line 73: Typo

sbueringer: PR: https://github.com/kubernetes-sigs/controller-runtime/pull/2792

PR #2791, opened 2024-04-21
:sparkling: Add TransformStripManagedFields

alvaroaleman:
This change adds `TransformStripManagedFields` to the Cache, a transform func that strips off all managed fields for objects. There are no known issues when using this as a `DefaultTransform` unless there is explicit code to access objects `ManagedFields` and can lead to a significant reduction in memory usage.


Danil-Grigorev (approved):
This is very useful when using client.Patch with Apply logic, thanks
/lgtm
/hold
if anyone wants to take a look

sbueringer:
Thx!!

/approve
/lgtm

/hold cancel

akutz:
Ohhh, it's because an update from controller-runtime will have a manager set to `manager` (by default) and thus not impact the managed field from another manager like `kubectl`, right?

sbueringer:
Let's see if someone else knows more. I simply never tried what happens with Update

alvaroaleman:
So a quick test shows that update with nil managed fields does not clear them:
```
k get deployment coredns -ojson --show-managed-fields=true|jq .metadata.managedFields >/tmp/original
cat main.go
package main

import (
	"context"

	appsv1 "k8s.io/api/apps/v1"
)

func main() {
		Cache: cache.Options{DefaultTransform: cache.TransformStripManagedFields()},
	})
	if err != nil {
		panic(err)
	}

	go func() {
		if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
			panic(err)
		}
	}()

	ctx := context.Background()
	if _, err := mgr.GetCache().GetInformer(ctx, &appsv1.Deployment{}); err != nil {
		panic(err)
	}
	dep := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{
		Name:      "coredns",
		Namespace: "kube-system",
	}}
	if err := mgr.GetClient().Get(ctx, client.ObjectKeyFromObject(dep), dep); err != nil {
		panic(err)
	}

	dep.Annotations["foo"] = "bar"
	if err := mgr.GetClient().Update(ctx, dep); err != nil {
		panic(err)
	}
}

go run ./
k get deployment coredns -ojson --show-managed-fields=true|jq .metadata.managedFields >/tmp/updated
$ diff -u /tmp/original /tmp/updated
--- /tmp/original	2024-09-11 21:42:10
+++ /tmp/updated	2024-09-11 21:44:14
@@ -164,6 +164,20 @@
     "fieldsV1": {
       "f:metadata": {
         "f:annotations": {
+          "f:foo": {}
+        }
+      }
+    },
+    "manager": "foo.bar",
+    "operation": "Update",
+    "time": "2024-09-12T01:44:07Z"
+  },
+  {
+    "apiVersion": "apps/v1",
+    "fieldsType": "FieldsV1",
+    "fieldsV1": {
+      "f:metadata": {
+        "f:annotations": {
           ".": {},
           "f:deployment.kubernetes.io/revision": {}
         }
@@ -200,6 +214,6 @@
     "manager": "kube-controller-manager",
     "operation": "Update",
     "subresource": "status",
-    "time": "2024-08-31T22:23:53Z"
+    "time": "2024-09-12T01:44:07Z"
   }
 ]
```

It should also be mentioned that upstream did this change [for the controller-manager](https://github.com/kubernetes/kubernetes/pull/118455) and [scheduler](https://github.com/kubernetes/kubernetes/pull/119556)), those two are what prompted me to do this change. 

According to that scheduler PR `even if scheduler use Update method on Pod in the future, updating Pod with managedFields being nil has no impact as kube-apiserver calculates the managed fields from the live object` so I do think this is safe.

sbueringer:
Nice! Thx for testing! So there is special handling for nil. I assume if some managedfields are set/modified it would update accordingly

PR #2796, opened 2024-04-23
📖  Mark Typed Handlers as experimental

alvaroaleman:
We will need to break them if we want to introduce a generic reconciler with a generic workqueue so mark them as experimental.


sbueringer:
Thx!

/lgtm
/approve

PR #2799, opened 2024-04-26
:warning: Add TypedReconciler

alvaroaleman:
This change adds a TypedReconciler which allows to customize the type being used in the workqueue. There is a number of situations where a custom type might be better than the default `reconcile.Request`:
* Multi-Cluster controllers might want to put the clusters in there
* Some controllers do not reconcile individual resources of a given type but all of them at once, for example IngressControllers might do this
* Some controllers do not operate on Kubernetes resources at all


alvaroaleman:
/hold
This needs some more testing

pkg/handler/eventhandler.go
```diff
@@ -64,70 +65,70 @@ type EventHandler TypedEventHandler[client.Object]
 // Most users shouldn't need to implement their own TypedEventHandler.
 //
 // TypedEventHandler is experimental and subject to future change.
-type TypedEventHandler[T any] interface {
+type TypedEventHandler[object any, request comparable] interface {
```
alvaroaleman, line 68: The main annoyance I found with this change for standard k8s controllers is this, when using a typed eventhandler to have the added typesafety in there it is now also needed to pass the request type. Removing it would mean it is impossible to use a custom request type with `source.Kind` though, which is for example useful when having watches in different clusters

pkg/source/source.go
```diff
@@ -248,16 +293,16 @@ func (is *Informer) String() string {
 	return fmt.Sprintf("informer source: %p", is.Informer)
 }
 
-var _ Source = Func(nil)
+var _ Source = Func[reconcile.Request](nil)
 
 // Func is a function that implements Source.
-type Func func(context.Context, workqueue.RateLimitingInterface) error
+type Func[request comparable] func(context.Context, workqueue.TypedRateLimitingInterface[request]) error
```
alvaroaleman, line 299: This is a breaking change but I think `source.Func` is barely used. Prior to the first generics PR, I didn't know it exists.

alvaroaleman:
I've tried this change with some internal projects and the changes required there were minimal and mostly around the interface change in `TypedEventHandler`. It is really cool to be able to use a custom request type and improves readability and type safety a lot in places where something other than the default is needed.

examples/typed/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func run() error {
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
+	if err != nil {
+		return fmt.Errorf("failed to construct manager: %w", err)
+	}
+
+	type request struct{}
+
+	r := reconcile.TypedFunc[request](func(ctx context.Context, _ request) (reconcile.Result, error) {
```
sbueringer, line 31: Might be good to just add a short comment that the goal of this pattern is to de-duplicate events per Ingress object

pkg/builder/controller.go
```diff
@@ -272,17 +291,23 @@ func (blder *Builder) project(obj client.Object, proj objectProjection) (client.
...
 	if blder.forInput.object != nil {
 		obj, err := blder.project(blder.forInput.object, blder.forInput.objectProjection)
 		if err != nil {
 			return err
 		}
-		hdler := &handler.EnqueueRequestForObject{}
+
+		var hdler handler.TypedEventHandler[client.Object, request]
+		if reflect.TypeFor[request]() != reflect.TypeOf(reconcile.Request{}) {
+			return errors.New("For() is not supported for TypedBuilder")
```
sbueringer, line 304: I think that's technically not correct. Only TypedBuilder which are for a type which is not reconcile.Request are not supported (the Builder itself is a supported TypedBuilder)

alvaroaleman: We only run into this if if the request type is not `reconcile.Request` and I guess the underlying assumption was that people was use the `Builder` type when they want a `TypedBuilder[reconcile.Request]`

examples/typed/main.go
alvaroaleman: Added a comment, WDYT?

pkg/source/source.go
```diff
@@ -41,54 +42,98 @@ import (
 // * Use Channel for events originating outside the cluster (e.g. GitHub Webhook callback, Polling external urls).
 //
 // Users may build their own Source implementations.
-type Source interface {
+type Source TypedSource[reconcile.Request]
```
sbueringer, line 45:
```suggestion
type Source = TypedSource[reconcile.Request]
```
Maybe?

See: https://github.com/kubernetes-sigs/cluster-api/pull/10803/files#r1656585773

sbueringer: Did a quick test. This change fixes the issue in CAPI

pkg/controller/controller.go
```diff
@@ -27,13 +27,15 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // Options are the arguments for creating a new Controller.
-type Options struct {
+type Options = TypedOptions[reconcile.Request]
+
+// TypedOptions are the arguments for creating a new Controller.
+type TypedOptions[T comparable] struct {
```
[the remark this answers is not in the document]
sbueringer: Same for other cases of [T] in this PR

pkg/internal/controller/controller.go
```diff
@@ -31,13 +31,12 @@ import (
...
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/internal/controller/metrics"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/ratelimiter"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // Controller implements controller.Controller.
-type Controller struct {
+type Controller[T comparable] struct {
```
sbueringer, line 39:
```suggestion
type Controller[request comparable] struct {
```

pkg/internal/controller/controller_test.go
```diff
@@ -400,8 +379,10 @@ var _ = Describe("controller", func() {
 		// TODO(directxman12): we should ensure that backoff occurrs with error requeue
 
 		It("should not reset backoff until there's a non-error result", func() {
-			dq := &DelegatingQueue{RateLimitingInterface: ctrl.NewQueue("controller1", nil)}
-			ctrl.NewQueue = func(string, ratelimiter.RateLimiter) workqueue.RateLimitingInterface { return dq }
+			dq := &DelegatingQueue{TypedRateLimitingInterface: ctrl.NewQueue("controller1", nil)}
```
sbueringer, line 382: Also wondering here if this should stay "RateLimitingInterface" as it's pinned to reconcile.Request

pkg/reconcile/reconcile.go
```diff
@@ -101,16 +108,19 @@ type Reconciler interface {
...
-	Reconcile(context.Context, Request) (Result, error)
+	Reconcile(context.Context, request) (Result, error)
 }
 
 // Func is a function that implements the reconcile interface.
-type Func func(context.Context, Request) (Result, error)
+type Func = TypedFunc[Request]
+
+// TypedFunc is a function that implements the reconcile interface.
+type TypedFunc[T comparable] func(context.Context, T) (Result, error)
```
sbueringer, line 118:
```suggestion
type TypedFunc[request comparable] func(context.Context, request) (Result, error)
```
(same in l.123)

pkg/source/source.go
```diff
@@ -41,54 +42,98 @@ import (
...
+
+// TypedSource is a generic source of events (e.g. Create, Update, Delete operations on Kubernetes Objects, Webhook callbacks, etc)
+// which should be processed by event.EventHandlers to enqueue reconcile.Requests.
```
sbueringer, line 48: Not necessarily reconcile.Requests, right?

pkg/source/source.go
```diff
@@ -41,54 +42,98 @@ import (
...
+	cache cache.Cache,
+	object T,
+	handler handler.TypedEventHandler[T, reconcile.Request],
+	predicates ...predicate.TypedPredicate[T],
+) SyncingSource {
+	return TypedKind(cache, object, handler, predicates...)
+}
+
+// TypedKind creates a KindSource with the given cache provider.
+func TypedKind[objectType client.Object, request comparable](
```
sbueringer, line 83: Let's make the type parameters a bit more consistent. I think in most parts of the PR it would be object and the argument name below would then be o?

(Same in the Kind func above)

pkg/source/source.go
```diff
@@ -228,7 +273,7 @@ var _ Source = &Informer{}
 
 // Start is internal and should be called only by the Controller to register an EventHandler with the Informer
```
sbueringer, line 274: Should `Handler    handler.EventHandler` be handler.TypedEventHandler? (same for predicates)

(disclaimer: might not make any sense at all :))

sbueringer: Additionally, I see there is a mismatch in the sense that NewEventHandler only works with client.Object while TypedEventHandler could handle any. Wondering if NewEventHandler should be able to handle any

sbueringer (commented):
Really really nice work! :)

Two high-level points:
* would be good to have consistent type parameter & corresponding argument names
* might be good to take another look if we have enough test coverage for the new "Typed" types

pkg/handler/eventhandler_test.go
```diff
@@ -41,12 +41,12 @@ import (
...
 var _ = Describe("Eventhandler", func() {
 	var ctx = context.Background()
-	var q workqueue.RateLimitingInterface
+	var q workqueue.TypedRateLimitingInterface[reconcile.Request]
 	var instance handler.EnqueueRequestForObject
 	var pod *corev1.Pod
 	var mapper meta.RESTMapper
 	BeforeEach(func() {
-		q = &controllertest.Queue{Interface: workqueue.New()}
+		q = &controllertest.Queue{TypedInterface: workqueue.NewTyped[reconcile.Request]()}
```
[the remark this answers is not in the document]
alvaroaleman: `TypedInterface` is embedded which is why we can't change the field name. We could not embedd it anymore, but then we'd have to add a bunch of methods to controllertest.Queue to delegate to the underlying queue

examples/typed/main.go
```diff
@@ -0,0 +1,60 @@
...
+		ingressList := &networkingv1.IngressList{}
+		if err := mgr.GetClient().List(ctx, ingressList); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to list ingresses: %w", err)
+		}
+
+		buildIngressConfig(ingressList)
+		return reconcile.Result{}, nil
+	})
+	if err := builder.TypedControllerManagedBy[request](mgr).
```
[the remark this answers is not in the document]
alvaroaleman: You are right, it is not needed here. But I think these examples are often used as basis that is then extended, so I'd prefer to keep it that way so folks have the type available if needed in their handler or predicate.

pkg/handler/eventhandler.go
```diff
@@ -42,7 +43,7 @@ import (
 //
 // Unless you are implementing your own EventHandler, you can ignore the functions on the EventHandler interface.
 // Most users shouldn't need to implement their own EventHandler.
-type EventHandler TypedEventHandler[client.Object]
+type EventHandler TypedEventHandler[client.Object, reconcile.Request]
```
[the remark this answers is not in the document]
alvaroaleman: Great catch, very surprised I didn't notice this during testing

pkg/builder/controller.go
```diff
@@ -52,21 +52,32 @@ const (
...
+	ctrl             controller.TypedController[request]
+	ctrlOptions      controller.TypedOptions[request]
 	name             string
+	newController    func(name string, mgr manager.Manager, options controller.TypedOptions[request]) (controller.TypedController[request], error)
 }
 
 // ControllerManagedBy returns a new controller builder that will be started by the provided Manager.
-func ControllerManagedBy(m manager.Manager) *Builder {
-	return &Builder{mgr: m}
+func ControllerManagedBy(m manager.Manager) *TypedBuilder[reconcile.Request] {
```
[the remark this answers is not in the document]
alvaroaleman: Yep

pkg/builder/controller.go
```diff
@@ -272,17 +291,23 @@ func (blder *Builder) project(obj client.Object, proj objectProjection) (client.
...
+func (blder *TypedBuilder[request]) doWatch() error {
 	// Reconcile type
 	if blder.forInput.object != nil {
 		obj, err := blder.project(blder.forInput.object, blder.forInput.objectProjection)
 		if err != nil {
 			return err
 		}
-		hdler := &handler.EnqueueRequestForObject{}
+
+		var hdler handler.TypedEventHandler[client.Object, request]
```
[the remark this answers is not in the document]
alvaroaleman: Done

pkg/builder/controller.go
```diff
@@ -52,21 +52,32 @@ const (
...
+// The workqueue de-duplicates identical requests.
+type TypedBuilder[request comparable] struct {
 	forInput         ForInput
 	ownsInput        []OwnsInput
-	rawSources       []source.Source
-	watchesInput     []WatchesInput
+	rawSources       []source.TypedSource[request]
+	watchesInput     []WatchesInput[request]
 	mgr              manager.Manager
 	globalPredicates []predicate.Predicate
```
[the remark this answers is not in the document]
alvaroaleman: The predicate type is the type of the object. In order to be able to watch multiple sources with different types, the concrete type of any source must not leak into the builder and hence everything in the builder related to the object remains typed as client.Object. That was basically the motivation for the interface changes in https://github.com/kubernetes-sigs/controller-runtime/pull/2783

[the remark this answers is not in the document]
alvaroaleman: You probably ran into this when looking at `TypedPredicates`. The only reason they have a type is to be able to implement `ApplyToWatches(opts *WatchesInput[request])` because when referencing another generic type, you have to specify the generic, you can not say "applies regardless of generic type". The predicate itself doesn't care at all and the predicate it sets on `WatchesInput` is typed to `client.Object`. I'll have to think about this a bit if I can come up with something better, but the only alternative I see is to change the `ApplyToWatches` interface to apply to an interface implemented by `WatchesInput` rather than the concrete `WatchesInput` as that is the only way to reference it regardless of generic.

pkg/controller/controller.go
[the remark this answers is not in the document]
alvaroaleman: Updated them all, you are right, its definitely a readability improvement

pkg/controller/controller.go
```diff
@@ -108,6 +118,11 @@ func New(name string, mgr manager.Manager, options Options) (Controller, error)
 // NewUnmanaged returns a new controller without adding it to the manager. The
 // caller is responsible for starting the returned controller.
 func NewUnmanaged(name string, mgr manager.Manager, options Options) (Controller, error) {
+	return NewTyped(name, mgr, options)
```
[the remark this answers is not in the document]
alvaroaleman: Fixed, great catch

pkg/handler/enqueue.go
```diff
@@ -44,10 +44,10 @@ type EnqueueRequestForObject = TypedEnqueueRequestForObject[client.Object]
 // Controllers that have associated Resources (e.g. CRDs) to reconcile the associated Resource.
 //
 // TypedEnqueueRequestForObject is experimental and subject to future change.
-type TypedEnqueueRequestForObject[T client.Object] struct{}
+type TypedEnqueueRequestForObject[object client.Object] struct{}
```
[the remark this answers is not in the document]
alvaroaleman: Done

pkg/source/source.go
alvaroaleman: Done

pkg/source/source.go
alvaroaleman: I've left it without generics under the assumption that this is specific to kube objects and thus will always use reconcile.Request. I guess that might not be true in same cases, like @sttts wanting to have the cluster in the request type

pkg/source/source.go
alvaroaleman: Great catch, not sure how I missed this

pkg/builder/controller.go
alvaroaleman: That is actually a good idea, updated it

pkg/builder/controller.go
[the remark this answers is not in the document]
alvaroaleman: I got rid of the generics in `Predicate` and `Projection` in 10a38deccfbdd15d14916ae89222a023b5d09b40, let em know what you think. Will squash at the end.

pkg/builder/options.go
```diff
@@ -52,28 +53,31 @@ func WithPredicates(predicates ...predicate.Predicate) Predicates {
...
 // ApplyToWatches applies this configuration to the given WatchesInput options.
-func (w Predicates) ApplyToWatches(opts *WatchesInput) {
+func (w TypedPredicates[request]) ApplyToWatches(opts *WatchesInput[request]) {
 	opts.predicates = w.predicates
 }
 
 var _ ForOption = &Predicates{}
 var _ OwnsOption = &Predicates{}
-var _ WatchesOption = &Predicates{}
+var _ WatchesOption[reconcile.Request] = &Predicates{}
```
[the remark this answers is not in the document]
alvaroaleman: Generic is gone there in 10a38deccfbdd15d14916ae89222a023b5d09b40

pkg/builder/options.go
```diff
@@ -82,20 +86,20 @@ var _ WatchesOption = &Predicates{}
...
-func (p projectAs) ApplyToOwns(opts *OwnsInput) {
+func (p projectAs[request]) ApplyToOwns(opts *OwnsInput) {
 	opts.objectProjection = objectProjection(p)
 }
 
 // ApplyToWatches applies this configuration to the given WatchesInput options.
-func (p projectAs) ApplyToWatches(opts *WatchesInput) {
+func (p projectAs[request]) ApplyToWatches(opts *WatchesInput[request]) {
 	opts.objectProjection = objectProjection(p)
 }
```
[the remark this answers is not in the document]
alvaroaleman: 10a38deccfbdd15d14916ae89222a023b5d09b40

pkg/source/source.go
alvaroaleman: Yeah, updated it to do that. I initially thought its pretty much completely unused, but turns out there are a few users: https://github.com/search?type=code&q=source.Func%28+language%3AGo

pkg/builder/controller_test.go
```diff
@@ -76,10 +76,6 @@ func (l *testLogger) WithName(name string) logr.LogSink {
 }
 
 var _ = Describe("application", func() {
```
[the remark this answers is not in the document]
alvaroaleman: Added negative tests for For() and Owned() with a custom type and a positive test for the happy path of `TypedControllerManagedBy`

pkg/source/source.go
alvaroaleman: I played around with this. This would be a very breaking change the the `Informer` source and there is no good way to infer the `object` type even when changing it to be initialized through a constructor func, because there is nothing to infer it from - The informer doesn't have it in its signature.

My understanding is this source only exists to be able to re-use an existing client-go informer. In 99% of cases, ppl should be using `source.Kind`.

I think it is ok not to support the typing with this source and if we change our mind on that/ppl ask for it, we can still do it later on.

examples/typed/main.go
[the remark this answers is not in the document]
sbueringer: Sounds good!

pkg/handler/eventhandler.go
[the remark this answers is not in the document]
sbueringer: Btw, I grepped over all occurences of something like this, I think now we got all of them :)

pkg/handler/eventhandler_test.go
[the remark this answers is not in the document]
sbueringer: Ah yeah, good point. That would not be not worth it :)

pkg/source/source.go
sbueringer: Sounds good, thank you very much for looking into it!

pkg/builder/controller.go
sbueringer: Thx!

pkg/builder/controller.go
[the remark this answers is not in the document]
sbueringer: Looks good to me!

pkg/builder/controller_test.go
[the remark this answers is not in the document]
sbueringer: Thx!

sbueringer:
Last nit from my side, feel free to squash directly

pkg/builder/controller.go
```diff
@@ -175,8 +203,12 @@ func (blder *Builder) Watches(object client.Object, eventHandler handler.EventHa
...
 // concrete type on top of the metadata cache; this increases memory
 // consumption and leads to race conditions as caches are not in sync.
-func (blder *Builder) WatchesMetadata(object client.Object, eventHandler handler.EventHandler, opts ...WatchesOption) *Builder {
-	opts = append(opts, OnlyMetadata)
+func (blder *TypedBuilder[request]) WatchesMetadata(
+	object client.Object,
+	eventHandler handler.TypedEventHandler[client.Object, request],
+	opts ...WatchesOption,
+) *TypedBuilder[request] {
+	opts = append(opts, projectAs(projectAsMetadata))
```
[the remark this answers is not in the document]
alvaroaleman: Updated it

sbueringer:
Thx!

/lgtm

Keeping the hold in case you want additional reviews

examples/typed/main.go
```diff
@@ -0,0 +1,60 @@
...
+		return fmt.Errorf("failed to construct manager: %w", err)
+	}
+
+	// Use a request type that is always equal to itself so the workqueue
+	// de-duplicates all events.
+	// This can for example be useful for an ingress-controller that
+	// generates a config from all ingresses, rather than individual ones.
+	type request struct{}
+
+	r := reconcile.TypedFunc[request](func(ctx context.Context, _ request) (reconcile.Result, error) {
```
vincepri, line 35: To expand on this a bit more can we add some reasoning on why this is better than say have a normal reconciler that just ignores the incoming request?

examples/typed/main.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		buildIngressConfig(ingressList)
+		return reconcile.Result{}, nil
+	})
+	if err := builder.TypedControllerManagedBy[request](mgr).
+		WatchesRawSource(source.TypedKind(
+			mgr.GetCache(),
+			&networkingv1.Ingress{},
+			handler.TypedEnqueueRequestsFromMapFunc(func(context.Context, *networkingv1.Ingress) []request {
```
vincepri, line 48: I assume here is one of the most important parts? The custom requests, outside of `reconcile.Request` can only be used with custom enqueuer?

examples/typed/main.go
```diff
@@ -0,0 +1,60 @@
+package main
```
vincepri, line 1: The PR description mentions this can be a good use for multiple clusters. Could we add an example that expands on the current `reconcile.Request` to add Cluster information?

examples/typed/main.go
alvaroaleman: From the code comment a few lines above:
```
	// Use a request type that is always equal to itself so the workqueue
	// de-duplicates all events.
	// This can for example be useful for an ingress-controller that
	// generates a config from all ingresses, rather than individual ones.
```

examples/typed/main.go
alvaroaleman: Yes, you need a custom event handler to be able to use a custom request type

examples/typed/main.go
alvaroaleman: I've added one under examples/multiclustersync

vincepri (approved):

sbueringer:
Nice additional example, thx for adding it!

PR #2806, opened 2024-05-01
🐛 correct kind source type

terinjokes:
Reverse the nil check in the String function of the internal Kind source to print the type when known.

Fixes: 2add01e7 ("Event, source, handler, predicates: Use generics")


alvaroaleman (approved):

alvaroaleman:
/ok-to-test
/cherrpick release-0.18

Thank you!

Issue #2790, opened 2024-04-20
Use setup-envtest from (test) code


tomasaschan:
I like being able to run the tests of a project with simply `go test ./...`, without requiring any local setup before that (other than a Go toolchain of the appropriate version being available). If one uses [`envtest`][envtest] to mock a Kubernetes cluster, one is expected to run some version of [`setup-envtest`][setup-envtest], to ensure the appropriate binaries are available; typically either like this
```
source <(setup-envtest use <version> -p env)
go test ./...
```
or e.g.
```
KUBEBUILDER_ASSETS=$(setup-envtest use <version> -p path) go test./...
```

Other than `setup-envtest` being written with purely this use case in mind, there's no _technical_ reason we couldn't allow triggering this from within the test harness, rather than outside of it.

[Here's a proof-of-concept implementation](https://github.com/tomasaschan/envtest-at-runtime) that runs the relevant parts of `setup-envtest` triggered from within the test instead, allowing
```
go test ./...
```
even for tests that use [envtest].

The main reason it's not very straightforward to write this code, is that the error handling in `setup-envtest` is based on panic-and-recover, rather than passing error values. But it also feels a bit odd to go into the internals of the tool, rather than the tool being built to support the use case in the first place.

If I send a PR to make this a supported feature and use case for [setup-envtes], what are the chances of that being accepted? :)

[envtest]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest
[setup-envtest]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest

alvaroaleman:
Yeah, I agree adding code to the `envtest.Environment` to download the binaries if needed would be useful

sbueringer:
Also sounds useful to me, we just have to check what the impact on our dependencies is. setup-envtest today is a separate go module.

tomasaschan:
[...]

PR #2810, opened 2024-05-03
:sparkles: Enable using setup-envtest without a separate CLI

tomasaschan:
This implements the proposal in #2790.

The implementation is incomplete so far; opening this mostly to show my progress and solicit feedback, as well as to avoid too many conflicts with initiatives like [this one](https://github.com/kubernetes-sigs/controller-runtime/issues/2790#issuecomment-2085721841).

To do:

* [x] Copy existing implementation of supporting packages over into the main module
* [x] Implement `use` in the main module
* [x] Implement `list` in the main module
* [x] Implement `cleanup` in the main module
* [x] Implement `sideload` in the main module
* [x] Port tests for the workflows (at least the ones that make sense)
* [x] Clean up the `setup-envtest` module, removing duplicated code etc
* [ ] ~Remove dependency on `afero`~
* [x] Rebase onto whatever latest changes have happened in e.g. `go.mod` since I branched off
* [ ] ~(Optional, possibly in a later PR) Re-implement the arg parsing and help parts of the main module with something a little less home-grown, probably something like https://github.com/spf13/cobra/~


tomasaschan:
I will rebase this properly once #2811 merges, since I expect conflicts with that anyway.

tomasaschan:
@sbueringer I think this is ready for an initial review now!

I still want to get rid of the dependency on `afero` before we merge this, so I'm keeping the PR in a draft state, but those changes should be basically independent of the major ones here. Let me know if you have any questions or would like me to somehow document the structure of the new set of packages to make reviewing easier.

tomasaschan:
I've spent some time now trying to rebuild this without a dependency on [`github.com/spf13/afero`](https://pkg.go.dev/github.com/spf13/afero), and while I was initially encouraged by the existence of [`testing/fstest`](https://pkg.go.dev/testing/fstest) in the standard libarry, it turns out that while the `afero` implements a complete file system abstraction (and implementations for the two pieces we need; an OS wrapper and an in-memory one for testing), `fstest` only implements file actions for a read-only file system, expecting the tests to set up the appropriate input data _before_ executing the SUT.

While it's certainly possible to build a read/write file system abstraction (on top of `fstest` or standalone) that supports the operations we need for `setup-envtest`, I feel it's probably not worth it in maintenance cost, compared to just relying on the third-party library. `afero` [does not change rapidly](https://pkg.go.dev/github.com/spf13/afero?tab=versions) and so we're unlikely to cause dependency hell issues by taking a dependency on it.

What do you think?

If you agree with this assessment, I can rebase again to get the latest version bumps of everything, and then this should be ready to go!

sbueringer:
@tomasaschan Just a heads up. We'll merge a PR directly after 0.19 which drops the support for GCS. Basically because we wanted to do it anyway, but also to simplify maintenance & refactorings like this PR

tomasaschan:
Thanks for the heads up!

Have you had time to review these changes, and in particular my last comment about keeping the afero dependency?

sbueringer:
Didn't get around to it yet. Just recently found time to catch up on controller-runtime in general and this is a big PR. But I'll try to look at it soon. Definitely feel free to wait with further rebases until than

tomasaschan:
/lifecycle frozen

@sbueringer and potentially other maintainers: given the initial hesitation to introduce a dependency on `github.com/spf13/afero` and my analysis of the available options for properly testing this (see https://github.com/kubernetes-sigs/controller-runtime/pull/2810#issuecomment-2198260406), I'd like someone to let me know what will be required for this to be merged before I undertake the non-trivial effort to rebase this.

tomasaschan:
@sbueringer Any chance I can get a more explicit decision on afero than a 👍 on my question above? 😅

sbueringer:
@tomasaschan Sorry should have brought this up way earlier but somehow I didn't think about that.

What is the functionality we are actually looking for from an envtest point of view? Download & extract the binaries if they don't already exist locally?

Just seems that if that's mostly what we are looking for, setup-envtest has an insane level of complexity.

Also it's probably a lot easier to add a few functions that cover what we need to the envtest package (they can reuse e.g. the http_client underneath) than porting all the commands and workflows over.

And sorry I think I lost the overview a bit, the idea of the current PR was to basically move the entire setup-envtest implementation into the main module with some refactorings so that everything can be also called in test code and now only via the CLI, correct?

tomasaschan:
@sbueringer My motivation for this PR is outlined in the issue linked from the PR body. In short, I'd like to avoid the need for wrapping the call to `go test ./...` with calls to setup-envtest to ensure that binaries are both available to (the ie downloaded) and found by (typically through setting an env var) envtest. Concretely, I want to enable things like kubebuilder projects to not have to use `make test` for running tests.

Wrt afero, the main reason to use it in setup-envtest is to make the file-system dependent parts testable; afero provides an abstraction that has both file system backed and in-memory backed implementations, meaning it's relatively easy to write tests that verify the file system dependent operations (anything from downloading the right version to choosing the right one based on what's already locally available). I did survey for similar abstractions in the stdlib, but found none - the one that exists is basically read-only.

What if the change instead involved pulling envtest out of controller-runtime (rather than pulling setup-envtest into it)? That way, dependencies would not be such an important consideration (depending on envtest is much more optional that depending on all of controller-runtime), and users of envtest could still configure their test harness to run without ad-hoc wrapping scripts. Would that be a more acceptable approach?

sbueringer:
> What if the change instead involved pulling envtest out of controller-runtime (rather than pulling setup-envtest into it)? 

This would only help if controller-runtime itself stops using envtest for its own testing. Otherwise we end up with the same transitive dependencies.

If I look at this from an envtest perspective, we basically have to invoke HTTPClient.GetVersion and then add a bit of code around it to ensure the file is written to the right location, right?

I would hope that testing for this could be done without abstracting away the filesystem.

tomasaschan:
> If I look at this from an envtest perspective, we basically have to invoke HTTPClient.GetVersion and then add a bit of code around it to ensure the file is written to the right location (+ skip if it's already there), right?

Plus all of the parsing of desired version, and resolving that (which involves looking at what's available locally, potentially at a custom location, and online, potentially with a custom registry). Additionally, there's still a few places where error handling is basically `panic`, which I suppose would have to be refactored for it to really be usable in a test context. (Although this is much rarer now than it used to be, nice! 👍🏻) To "add a bit of code around it" would essentially re-implement parts of the `Store` if we don't just use that, so it feels much more reasonable to make sure that part of `setup-envtest` can also be made available from a package.

> I think testing for this can be done without abstracting away the filesystem.

The main benefit of in-memory file systems for tests over real ones, IMO, is that the in-memory ones share no state, so there is no coupling between the tests. If we write the tests to interact with a real file system, we'll have all kinds of issues around shared state, test ordering, etc. Do you have a proposed way to avoid that?

sbueringer:
> The main benefit of in-memory file systems for tests over real ones, IMO, is that the in-memory ones share no state, so there is no coupling between the tests. If we write the tests to interact with a real file system, we'll have all kinds of issues around shared state, test ordering, etc. Do you have a proposed way to avoid that?

Make paths configurable and use different tmp folder(s) for different tests

sbueringer:
I could miss something, but I think this PR would cover all the use cases: https://github.com/kubernetes-sigs/controller-runtime/pull/3135

tomasaschan:
@sbueringer Nice, thanks for making that change for me!

I tried it out, and it does indeed seem to work pretty well - except that it assumes that the binary assets path must already exist, which it might not (especially not in ephemeral test environments). It'd be nice if the downloader could first do something equivalent of `mkdir -p` for the target directory, so the user doesn't have to do that out-of-band.

sbueringer:
Perfect, great to hear it works for you as well :)

> I tried it out, and it does indeed seem to work pretty well - except that it assumes that the binary assets path must already exist, which it might not (especially not in ephemeral test environments). It'd be nice if the downloader could first do something equivalent of mkdir -p for the target directory, so the user doesn't have to do that out-of-band.

Damn, I tried to do that, but I failed :)

This PR should fix it up: https://github.com/kubernetes-sigs/controller-runtime/pull/3137
Please give it a try

sbueringer:
Ups sorry. I didn't know that I can close a PR by mentioning it with "fixes" in another PR :)

Independent of that, @tomasaschan are we fine with closing this PR?

In general if we need a bit more, I would probably prefer extending the funcionality added in #3135 over moving around setup-envtest code. There's just so much baggage / complexity in setup-envtest.

tomasaschan:
Yeah, the use case that motivated this work is solved by #3135 (and #3137) so I'm fine with not doing more here. (Thanks again for those!)

I still think it could be useful to have the code in `setup-envtest` available to use as imported libraries rather than a CLI, which would require - at least - re-working the error handling, but I don't have an immediate use case for it so I'm not going to spend any effort on it at the moment :)

sbueringer:
Sounds good! Sorry for the long delays.

If there are new use cases coming up that specifically require to call setup-envtest programatically we could consider improving the setup-envtest module in place. So folks that want to use that would import the existing setup-envtest module in addition to CR

(but I think it's nice that for the use cases that #3135 covers that won't be necessary)

tomasaschan:
What's the release schedule for controller-runtime? When can we expect a patch release out (so we don't have to pin to `@main`)?

sbueringer:
The next CR minor release will be out after Kubernetes v1.33. I would assume very late in April or early May.

I would be fine with backporting the pkg/envtest changes of #3135 if there are no objections from @alvaroaleman & @vincepri 

We can cut a patch release in the next 1-2 weeks or something (patch releases are mostly on-demand, +/- sometimes we collect a few changes)

PR #2813, opened 2024-05-06
🐛 Reintroduce AddMetricsServerExtraHandler on manager

vincepri:

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -179,6 +179,24 @@ func (cm *controllerManager) add(r Runnable) error {
 	return cm.runnables.Add(r)
 }
 
+// AddMetricsServerExtraHandler adds extra handler served on path to the http server that serves metrics.
+func (cm *controllerManager) AddMetricsServerExtraHandler(path string, handler http.Handler) error {
+	cm.Lock()
+	defer cm.Unlock()
+	if cm.started {
+		return fmt.Errorf("unable to add new metrics handler because metrics endpoint has already been created")
+	}
+	if cm.metricsServer == nil {
+		cm.GetLogger().Info("warn: metrics server is currently disabled, registering extra handler %q will be ignored", path)
+		return nil
+	}
+	if err := cm.metricsServer.AddExtraHandler(path, handler); err != nil {
+		return err
+	}
+	cm.logger.V(2).Info("Registering metrics http server extra handler", "path", path)
+	return nil
+}
+
 // AddHealthzCheck allows you to add Healthz checker.
 func (cm *controllerManager) AddHealthzCheck(name string, check healthz.Checker) error {
 	cm.Lock()
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -64,6 +64,15 @@ type Manager interface {
 	// election was configured.
 	Elected() <-chan struct{}
 
+	// AddMetricsServerExtraHandler adds an extra handler served on path to the http server that serves metrics.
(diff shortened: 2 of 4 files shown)


pkg/manager/internal.go
```diff
@@ -179,6 +179,20 @@ func (cm *controllerManager) add(r Runnable) error {
...
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
+func (cm *controllerManager) AddMetricsExtraHandler(path string, handler http.Handler) error {
+	cm.Lock()
+	defer cm.Unlock()
+	if cm.started {
+		return fmt.Errorf("unable to add new metrics handler because metrics endpoint has already been created")
+	}
+	if err := cm.metricsServer.AddExtraHandler(path, handler); err != nil {
```
sbueringer, line 189: I think metricsServer is nil if BindAddress = 0

I think we should avoid a panic here. Not sure if we want to just return nil or return an error. I think I'm leaning towards returning nil if cm.metricsServer is nil (because disabling the metrics by setting BindAddress to 0 could be an end-user choice, so the author of the controller might not have that under control)

pkg/manager/manager.go
```diff
@@ -64,6 +64,15 @@ type Manager interface {
...
 	Elected() <-chan struct{}
 
+	// AddMetricsExtraHandler adds an extra handler served on path to the http server that serves metrics.
+	// Might be useful to register some diagnostic endpoints e.g. pprof.
+	//
+	// Note that these endpoints meant to be sensitive and shouldn't be exposed publicly.
+	//
+	// If the simple path -> handler mapping offered here is not enough,
+	// a new http server/listener should be added as Runnable to the manager via Add method.
+	AddMetricsExtraHandler(path string, handler http.Handler) error
```
sbueringer, line 74: Fine for me, but wondering if AddMetricsServerExtraHandler would be better.

If you add a pprof handler it's not really a "metrics handler"

pkg/manager/manager.go
```diff
@@ -64,6 +64,15 @@ type Manager interface {
 	// election was configured.
 	Elected() <-chan struct{}
 
+	// AddMetricsExtraHandler adds an extra handler served on path to the http server that serves metrics.
+	// Might be useful to register some diagnostic endpoints e.g. pprof.
+	//
+	// Note that these endpoints meant to be sensitive and shouldn't be exposed publicly.
```
sbueringer, line 70:
```suggestion
	// Note that these endpoints are meant to be sensitive and shouldn't be exposed publicly.
```
nit

pkg/metrics/server/server.go
```diff
@@ -182,6 +185,23 @@ func (*defaultServer) NeedLeaderElection() bool {
 	return false
 }
 
+// AddMetricsExtraHandler adds extra handler served on path to the http server that serves metrics.
```
sbueringer, line 188:
```suggestion
// AddExtraHandler adds extra handler served on path to the http server that serves metrics.
```

pkg/manager/internal.go
vincepri: If the bind address is 0, the extra handlers should still be populated, but disabled, no?

pkg/manager/manager.go
vincepri: fixed

pkg/metrics/server/server.go
vincepri: fixed

pkg/manager/internal.go
vincepri: added check and log

PR #2828, opened 2024-05-14
:bug: fix(controllerutil): avoid panic when the MutateFn is nil

0xff-dev:
I checked the `CreateOrPatch` function and it checked if `MutateFn` is nil, but `CreateOrUpdate` did not check.

Or is it possible to modify the `mutate` function directly and execute `MutateFn` only when `MutateFn` is not nil?
https://github.com/kubernetes-sigs/controller-runtime/blob/5dcea7e94b7158e410d22f4a1eed16780497fe66/pkg/controller/controllerutil/controllerutil.go#L425

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -310,9 +310,12 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f M
 		if !apierrors.IsNotFound(err) {
 			return OperationResultNone, err
 		}
-		if err := mutate(f, key, obj); err != nil {
-			return OperationResultNone, err
+		if f != nil {
+			if err := mutate(f, key, obj); err != nil {
+				return OperationResultNone, err
+			}
 		}
+
 		if err := c.Create(ctx, obj); err != nil {
 			return OperationResultNone, err
 		}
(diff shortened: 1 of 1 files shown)


troy0820:
Don't we want this to panic because the caller must apply a mutatefn?  This function is supposed to be called regardless and adding the nil check suggests that we can perform this without this function being called.

0xff-dev:
> Don't we want this to panic because the caller must apply a mutatefn? This function is supposed to be called regardless and adding the nil check suggests that we can perform this without this function being called.

You're right, by the function comment it states that the function would have been called anyway, but `CreateOrPatch` states the same thing, in the code it does a check. I think that even though the comment states the behavior of the function, I think it's still important to prevent special input from the user.

troy0820:
> You're right, by the function comment it states that the function would have been called anyway, but CreateOrPatch states the same thing, in the code it does a check. I think that even though the comment states the behavior of the function, it's still important to prevent special input from the user.

Fair point, I know the functions exhibit different behavior (Patching subresource doesn't happen in CreateOrUpdate), and may be a reason to make sure that the object doesn't.  I can't speak to this difference in behavior but will agree with your point. 

/ok-to-test

sbueringer:
I think it's okay to ensure both funcs have the same behavior. Panic'ing is not acceptable.

There could be a general point around if passing in a nil function is rather an error case and *both* funcs should return an error. But for now fixing the panic and ensuring both funcs behave the same in this regard seems fine to me.

/lgtm
/approve

Issue #2846, opened 2024-06-05
controllerutil: SetControllerReference does not allow disabling BlockOwnerDeletion


muwaqar:
controllerutil `SetControllerReference()` implementation necessarily sets `BlockOwnerDeletion = ptr.To(true)`. This means that the user needs to explicitly delete the referent resource before the owner can be cleaned up (see [foreground deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) details).

users should be able to configure `BlockOwnerDeletion` while using these utility methods conveniently.

PR #2847, opened 2024-06-05
✨ controllerutil: configure BlockOwnerDeletion when setting OwnerReference

muwaqar:
Currently, `SetOwnerReference()` and `SetControllerReference()` in controllerutil do not allow configuring `BlockOwnerDeletion` field of the underlying `OwnerReference`. This is limiting for scenarios where one would like to use these helper methods but configure `BlockOwnerDeletion` field in a different way than the current defaults.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2846

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -52,12 +52,22 @@ func newAlreadyOwnedError(obj metav1.Object, owner metav1.OwnerReference) *Alrea
 	}
 }
 
+// OwnerReferenceOption is a function that can modify a `metav1.OwnerReference`.
+type OwnerReferenceOption func(*metav1.OwnerReference)
+
+// WithBlockOwnerDeletion allows configuring the BlockOwnerDeletion field on the `metav1.OwnerReference`.
+func WithBlockOwnerDeletion(blockOwnerDeletion bool) OwnerReferenceOption {
+	return func(ref *metav1.OwnerReference) {
+		ref.BlockOwnerDeletion = &blockOwnerDeletion
+	}
+}
+
 // SetControllerReference sets owner as a Controller OwnerReference on controlled.
 // This is used for garbage collection of the controlled object and for
 // reconciling the owner object on changes to controlled (with a Watch + EnqueueRequestForOwner).
 // Since only one OwnerReference can be a controller, it returns an error if
 // there is another OwnerReference with Controller flag set.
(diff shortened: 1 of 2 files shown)


troy0820:
/hold 

In your issue you state 

> controllerutil SetControllerReference() implementation necessarily sets BlockOwnerDeletion = ptr.To(true). This means that the user needs to explicitly delete the referent resource before the owner can be cleaned up (see [foreground deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) details).
 users should be able to configure BlockOwnerDeletion while using these utility methods conveniently

but doesn't that go against what we want to use `SetControllerReference` for?  

Allowing the BlockOwnerDeletion to be configurable changes the intention of why that is necessary and may conflate the workflow with the owner/controller references.  

The requisite helper functions alleviate the need to change the configuration of SetControllerReference by using `RemoveControllerReferfence`

muwaqar:
> but doesn't that go against what we want to use SetControllerReference for?

@troy0820: based on my understanding of the original [ControllerRef proposal](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md), the controller reference is used to identify the owning controller only so that multiple controllers with sharing labels do not clash over the resource, but does not dictate any behavior for garbage collection of owned resources, and says:

> When a controller is deleted, the objects it owns will either be orphaned or deleted according to the normal [Garbage Collection](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/garbage-collection.md) behavior, based on OwnerReferences.

i interpret this to be users should be able to configure the cascade deletion behavior.

alvaroaleman:
> but doesn't that go against what we want to use SetControllerReference for?

> Allowing the BlockOwnerDeletion to be configurable changes the intention of why that is necessary and may conflate the workflow with the owner/controller references.

The handler we have in `Owns` requires the ownerRef to have `controller: true`. Ending up with foreground deletion just because you use `SetControllerRef` is surprising IMHO, those two settings are orthogonal, if you do or do not want foreground deletion can not be derived from controller: true

/ok-to-test

sbueringer:
Agree with what @muwaqar & @alvaroaleman wrote

alvaroaleman (approved):

PR #2855, opened 2024-06-17
✨ Add scale subresource logic to fake client

TheSpiritXIII:
Adds basic GET/UPDATE support to the scale sub-resource according to the docs: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#SubResourceClientConstructor

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -31,6 +31,8 @@ import (
 
 	// Using v4 to match upstream
 	jsonpatch "gopkg.in/evanphx/json-patch.v4"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
@@ -50,6 +52,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/testing"
+	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -83,6 +86,8 @@ const (
 	maxNameLength          = 63
 	randomLength           = 5
 	maxGeneratedNameLength = maxNameLength - randomLength
+
+	subResourceScale = "scale"
 )
 
 // NewFakeClient creates a new fake client for testing.
@@ -1111,7 +1116,26 @@ type fakeSubResourceClient struct {
 }
 
 func (sw *fakeSubResourceClient) Get(ctx context.Context, obj, subResource client.Object, opts ...client.SubResourceGetOption) error {
-	panic("fakeSubResourceClient does not support get")
+	switch sw.subResource {
+	case subResourceScale:
+		// Actual client looks up resource, then extracts the scale sub-resource:
+		// https://github.com/kubernetes/kubernetes/blob/fb6bbc9781d11a87688c398778525c4e1dcb0f08/pkg/registry/apps/deployment/storage/storage.go#L307
+		if err := sw.client.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+			return err
+		}
+		scale, isScale := subResource.(*autoscalingv1.Scale)
+		if !isScale {
+			return apierrors.NewBadRequest(fmt.Sprintf("expected Scale, got %t", subResource))
+		}
+		scaleOut, err := extractScale(obj)
+		if err != nil {
+			return err
+		}
+		*scale = *scaleOut
+		return nil
+	default:
+		return fmt.Errorf("fakeSubResourceClient does not support get for %s", sw.subResource)
+	}
 }
 
 func (sw *fakeSubResourceClient) Create(ctx context.Context, obj client.Object, subResource client.Object, opts ...client.SubResourceCreateOption) error {
@@ -1138,11 +1162,30 @@ func (sw *fakeSubResourceClient) Update(ctx context.Context, obj client.Object, 
 	updateOptions := client.SubResourceUpdateOptions{}
 	updateOptions.ApplyOptions(opts)
 
-	body := obj
-	if updateOptions.SubResourceBody != nil {
-		body = updateOptions.SubResourceBody
+	switch sw.subResource {
+	case subResourceScale:
+		if err := sw.client.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+			return err
+		}
+		if updateOptions.SubResourceBody == nil {
+			return apierrors.NewBadRequest("missing SubResourceBody")
+		}
+
+		scale, isScale := updateOptions.SubResourceBody.(*autoscalingv1.Scale)
+		if !isScale {
+			return apierrors.NewBadRequest(fmt.Sprintf("expected Scale, got %t", updateOptions.SubResourceBody))
+		}
+		if err := applyScale(obj, scale); err != nil {
+			return err
+		}
+		return sw.client.update(obj, false, &updateOptions.UpdateOptions)
+	default:
+		body := obj
+		if updateOptions.SubResourceBody != nil {
+			body = updateOptions.SubResourceBody
+		}
+		return sw.client.update(body, true, &updateOptions.UpdateOptions)
 	}
-	return sw.client.update(body, true, &updateOptions.UpdateOptions)
 }
 
 func (sw *fakeSubResourceClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.SubResourcePatchOption) error {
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -1278,3 +1315,88 @@ func zero(x interface{}) {
...
+	case *corev1.ReplicationController:
+		obj.Spec.Replicas = &scale.Spec.Replicas
+	case *appsv1.StatefulSet:
+		obj.Spec.Replicas = &scale.Spec.Replicas
+	default:
+		// TODO: CRDs https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource
+		return fmt.Errorf("unable to extract scale from type %T", obj)
+	}
+	return nil
+}
```
troy0820, line 1402: Wouldn't this not allow people who make CRs who wish to use the sub resource scale be limited to only allow the types in the case statement available?

pkg/client/fake/client.go
```diff
@@ -1080,7 +1082,24 @@ type fakeSubResourceClient struct {
...
+		}
+		scaleOut, err := extractScale(obj)
+		if err != nil {
+			return err
+		}
+		*scale = scaleOut
+		return nil
+	default:
+		return fmt.Errorf("fakeSubResourceClient does not support get for %s", sw.subResource)
+	}
```
troy0820, line 1102: Having to use the subresource client to then use the actual client seems like we wouldn't need the sub resource client to do the GET.  The value of the scale is indicative on the spec and the subresource client has to get that value from the actual client where it's derived.

pkg/client/fake/client.go
```diff
@@ -1108,10 +1127,28 @@ func (sw *fakeSubResourceClient) Update(ctx context.Context, obj client.Object,
 	updateOptions.ApplyOptions(opts)
 
 	body := obj
-	if updateOptions.SubResourceBody != nil {
-		body = updateOptions.SubResourceBody
+	switch sw.subResource {
+	case "scale":
```
alvaroaleman, line 1131: Lets use a `const` for the subresource name

pkg/client/fake/client.go
```diff
@@ -1108,10 +1127,28 @@ func (sw *fakeSubResourceClient) Update(ctx context.Context, obj client.Object,
...
+		if updateOptions.SubResourceBody == nil {
+			return apierrors.NewBadRequest("expected SubResourceBody")
+		}
+		scale, isScale := updateOptions.SubResourceBody.(*autoscalingv1.Scale)
+		if !isScale {
+			return apierrors.NewBadRequest(fmt.Sprintf("got invalid type %t, expected Scale", updateOptions.SubResourceBody))
```
alvaroaleman, line 1137: Is this the behavior of the KAS, to first complain with a bad request about the request body before checking if the main resource even exists?

pkg/client/fake/client.go
```diff
@@ -1278,3 +1315,88 @@ func zero(x interface{}) {
 	res := reflect.ValueOf(x).Elem()
 	res.Set(reflect.Zero(res.Type()))
 }
+
+func extractScale(obj client.Object) (autoscalingv1.Scale, error) {
+	switch obj := obj.(type) {
+	case *appsv1.Deployment:
```
alvaroaleman, line 1321: Is go smart enough to allow expression this in a single switch case:
```suggestion
	case *appsv1.Deployment, *appsv1.ReplicaSet:
```

pkg/client/fake/client.go
```diff
@@ -1278,3 +1315,88 @@ func zero(x interface{}) {
...
+				Replicas: replicas,
+			},
+			Status: autoscalingv1.ScaleStatus{
+				Replicas: obj.Status.Replicas,
+				Selector: obj.Spec.Selector.String(),
+			},
+		}, nil
+	default:
+		// TODO: CRDs https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource
+		return autoscalingv1.Scale{}, fmt.Errorf("unable to extract scale from type %T", obj)
```
alvaroaleman, line 1383: This is not a good error, the error should clarify that scale is not implemented in the fake client for %T. Ideally we would also exclude all in-tree resources that do not have a scale subresource and return the apprpropriate error for that rather than falling through here

alvaroaleman (commented):
Thanks a lot for looking into this, really appreciated! Few comments, we need to make really sure we behave the same way as the KAS does everywhere, because as soon as we release this, some ppl will start depending on whatever behavior it exhibits which might cause tests to incorrectly pass if this doesn't exactly match the kube apiserver behavior.

pkg/client/fake/client.go
TheSpiritXIII: Thanks! Added a look-up and more details here. Please let me know if you'd like any further changes.

pkg/client/fake/client.go
TheSpiritXIII: No, sadly :/

pkg/client/fake/client.go
TheSpiritXIII: This is implemented on the server-side as far as I know. I cannot just lookup on the client because the fake client doesn't know how to extract this information. It's something of a meta-resource.

pkg/client/fake/client.go
TheSpiritXIII: I was unable to find it; am I looking at the right place? https://github.com/search?q=repo%3Akubernetes%2Fapiserver+scale+-path%3Atest&type=code

Nonetheless, I was able to reproduce it with `client.Client.Get` and your intuition is correct: it looks up the object first! Great catch.

pkg/client/fake/client.go
troy0820: Fair and not criticizing the implementation but asking because you need to deal with this the way you are dealing with it.

alvaroaleman:
Will try to give this a proper review this week, sorry for the delay.
/ok-to-test

pkg/client/fake/client.go
alvaroaleman: I think the upstream code is here: https://github.com/kubernetes/kubernetes/blob/fb6bbc9781d11a87688c398778525c4e1dcb0f08/pkg/registry/apps/deployment/storage/storage.go#L320

Not exactly a quick read though it seems.

pkg/client/fake/client.go
```diff
@@ -1278,3 +1322,126 @@ func zero(x interface{}) {
...
+			Spec: autoscalingv1.ScaleSpec{
+				Replicas: replicas,
+			},
+			Status: autoscalingv1.ScaleStatus{
+				Replicas: obj.Status.Replicas,
+				Selector: obj.Spec.Selector.String(),
+			},
+		}, nil
+	default:
+		crd, err := getCustomResourceDefinition(ctx, c, obj)
```
alvaroaleman, line 1389: Can we remove this code and directly error out? For status for example we have an explicit config option on the fake client. I don't know how that would best look like for scale given it requires more config than just a binary enable/disable, but I would prefer not adding any code related to that until we actually support that.

pkg/client/fake/client.go
```diff
@@ -1278,3 +1322,126 @@ func zero(x interface{}) {
...
+
+func extractScale(ctx context.Context, c client.Client, obj client.Object) (*autoscalingv1.Scale, error) {
+	switch obj := obj.(type) {
+	case *appsv1.Deployment:
+		var replicas int32 = 1
+		if obj.Spec.Replicas != nil {
+			replicas = *obj.Spec.Replicas
+		}
+		return &autoscalingv1.Scale{
+			ObjectMeta: obj.ObjectMeta,
```
alvaroaleman, line 1334: I think its only namespace, name, uid, resourceVersion and creationTimestamp we should copy over, not for example labels or annotations. Please update this and the corresponding tests.

pkg/client/fake/client_test.go
```diff
@@ -2068,6 +2070,70 @@ var _ = Describe("Fake client", func() {
...
+		obj := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "foo",
+			},
+		}
+		cl := NewClientBuilder().WithScheme(scheme).WithObjects(obj).Build()
+
+		scale := &autoscalingv1.Scale{Spec: autoscalingv1.ScaleSpec{Replicas: 2}}
+		Expect(cl.SubResource(subResourceScale).Get(context.Background(), obj, scale)).ToNot(Succeed())
+		Expect(cl.SubResource(subResourceScale).Update(context.Background(), obj, client.WithSubResourceBody(scale))).ToNot(Succeed())
```
alvaroaleman, line 2088: Could we match the error rather than just checking for an error? Right now, this test would for example incorrectly pass if you did a GET for an object that doesn't exist

pkg/client/fake/client_test.go
```diff
@@ -2068,6 +2070,70 @@ var _ = Describe("Fake client", func() {
...
+		objOriginal := obj.DeepCopy()
+
+		scale := &autoscalingv1.Scale{}
+		Expect(cl.SubResource(subResourceScale).Get(context.Background(), obj, scale)).NotTo(HaveOccurred())
+
+		actual := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: obj.Name}}
+		Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(actual), actual)).To(Succeed())
+
+		objOriginal.APIVersion = scale.APIVersion
+		objOriginal.Kind = scale.Kind
```
alvaroaleman, line 2110: These lines confuse me. Neither `objOriginal` nor `scale` should have either `APIVersion` or `Kind` set - Why is this needed? Same question for the update test.

pkg/client/fake/client_test.go
```diff
@@ -2068,6 +2070,70 @@ var _ = Describe("Fake client", func() {
...
+		}
+		cl := NewClientBuilder().WithScheme(scheme).WithObjects(obj).Build()
+
+		scale := &autoscalingv1.Scale{Spec: autoscalingv1.ScaleSpec{Replicas: 2}}
+		Expect(cl.SubResource(subResourceScale).Get(context.Background(), obj, scale)).ToNot(Succeed())
+		Expect(cl.SubResource(subResourceScale).Update(context.Background(), obj, client.WithSubResourceBody(scale))).ToNot(Succeed())
+	})
+
+	It("should be able to Get scale subresources", func() {
+		obj := &appsv1.Deployment{
```
alvaroaleman, line 2092: It would be good to test this for all object kinds where we support this, maybe you can loop over a list of deployment/replicaset/statefulset/daemonset for that?

pkg/client/fake/client_test.go
```diff
@@ -2068,6 +2070,70 @@ var _ = Describe("Fake client", func() {
...
+		scale := &autoscalingv1.Scale{}
+		Expect(cl.SubResource(subResourceScale).Get(context.Background(), obj, scale)).NotTo(HaveOccurred())
+
+		actual := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: obj.Name}}
+		Expect(cl.Get(context.Background(), client.ObjectKeyFromObject(actual), actual)).To(Succeed())
+
+		objOriginal.APIVersion = scale.APIVersion
+		objOriginal.Kind = scale.Kind
+		objOriginal.ResourceVersion = scale.ResourceVersion
+		objOriginal.Spec.Replicas = ptr.To(scale.Spec.Replicas)
```
alvaroaleman, line 2112: This test needs to verify that they are equal, right now it doesn't do that

alvaroaleman:
@TheSpiritXIII will you be able to rebase and resolve the remaining comments?

TheSpiritXIII:
@alvaroaleman I rebased and addressed all comments. Please let me know if there's anything else! Thanks.

alvaroaleman (approved):
Thank you!

PR #2860, opened 2024-06-20
:sparkles: Adds a client that enables strict field validation by default.

### LINKED ISSUE #2859: Server-side field validation should be strict by default
Creating/updating unstructured objects is a supported use case for controller-runtime.

When a controller works with an unstructured objects, there is no Go type to ensure the object conforms to a schema. Instead, the controller relies on [server-side validation](https://kubernetes.io/blog/2023/04/24/openapi-v3-field-validation-ga/#server-side-field-validation).

By default, server-side validation is not strict. For example, if a controller creates or updates an object and adds an unknown field, the request is not rejected; at most, a warning message is logged.

```
I0612 13:30:46.003406  164090 warning_handler.go:65] "msg"="unknown field \"spec.exampleField\"" "logger"="KubeAPIWarningLogger" 
```

Warning messages are difficult to test against, and can be missed in production. In turn, controller bugs are missed.

All controllers would benefit from enabling strict validation, but some may be unable to use it, e.g. because of backward compatibility (this is also why strict field validation was not enabled by default in the API server).

Individual client requests can be configured so that the API server returns an error, although it requires understanding the use of the "raw" options:

```go
client.Update(ctx, obj, &client.UpdateOptions{Raw: &metav1.UpdateOptions{ FieldValidation: metav1.FieldValidationStrict}})
```

But, if all controllers would benefit from enabling strict validation, then it's poor UX to require controller authors to enable it in every client request.

**Therefore, I propose that controller-runtime**

**(a) allow the user to configure field validation at client create time, and
(b) consider enabling strict field validation the default**

dlipovetsky:
This allows our users to enable strict field validation by default for all create, update, and patch requests.

The client can be used in a similar way to the specialized fieldManager-configuring client added in #2771.

Fixes #2859

### DIFF
--- a/pkg/client/fieldvalidation.go
+++ b/pkg/client/fieldvalidation.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// WithStrictFieldValidation wraps a Client and configures strict field
+// validation, by default, for all write requests from this client. Users
+// can override the field validation for individual requests.
+func WithStrictFieldValidation(c Client) Client {
+	return &clientWithFieldValidation{
+		validation: metav1.FieldValidationStrict,
+		c:          c,
+		Reader:     c,
+	}
+}
+
+type clientWithFieldValidation struct {
+	validation string
+	c          Client
+	Reader
+}
+
+func (f *clientWithFieldValidation) Create(ctx context.Context, obj Object, opts ...CreateOption) error {
+	return f.c.Create(ctx, obj, append([]CreateOption{FieldValidation(f.validation)}, opts...)...)
+}
+
+func (f *clientWithFieldValidation) Update(ctx context.Context, obj Object, opts ...UpdateOption) error {
+	return f.c.Update(ctx, obj, append([]UpdateOption{FieldValidation(f.validation)}, opts...)...)
+}
+
+func (f *clientWithFieldValidation) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error {
+	return f.c.Patch(ctx, obj, patch, append([]PatchOption{FieldValidation(f.validation)}, opts...)...)
+}
+
+func (f *clientWithFieldValidation) Delete(ctx context.Context, obj Object, opts ...DeleteOption) error {
+	return f.c.Delete(ctx, obj, opts...)
+}
+
+func (f *clientWithFieldValidation) DeleteAllOf(ctx context.Context, obj Object, opts ...DeleteAllOfOption) error {
+	return f.c.DeleteAllOf(ctx, obj, opts...)
+}
+
+func (f *clientWithFieldValidation) Scheme() *runtime.Scheme     { return f.c.Scheme() }
+func (f *clientWithFieldValidation) RESTMapper() meta.RESTMapper { return f.c.RESTMapper() }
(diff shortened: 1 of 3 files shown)


pkg/client/fieldvalidation_test.go
```diff
@@ -0,0 +1,149 @@
...
+	_ = wrappedClient.SubResource("some-subresource").Patch(ctx, dummyObj, nil, client.FieldValidation(metav1.FieldValidationWarn))
+
+	if expectedCalls := 9; calls != expectedCalls {
+		t.Fatalf("wrong number of calls to assertions: expected=%d; got=%d", expectedCalls, calls)
+	}
+}
+
+// testFieldValidationClient is a helper function that checks if calls have the expected field validation,
+// and calls the callback function on each intercepted call.
+func testFieldValidationClient(t *testing.T, expectedFieldValidation string, callback func()) client.Client {
```
dlipovetsky, line 79: I've shamelessly copied and adapted the test code from fieldowner_test.go.

This is almost identical to the `testClient` helper function in that file. I'd be happy to refactor the two helper functions into one that can be used by both tests.

alvaroaleman (approved):
thanks, lgtm
Putting a hold so others can have a look, too
/hold

pkg/client/fieldvalidation.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// WithStrictFieldValidation wraps a Client and configures strict field
+// validation, by default, for all write requests from this client. Users
+// can override the field validation for individual requests.
+func WithStrictFieldValidation(c Client) Client {
```
sbueringer, line 31: If we make validation a parameter then this same func can be used for the other validation values as well

pkg/client/options.go
```diff
@@ -187,6 +220,25 @@ type CreateOptions struct {
...
+	// - Warn: This will send a warning via the standard warning response
+	// header for each unknown field that is dropped from the object, and
+	// for each duplicate field that is encountered. The request will
+	// still succeed if there are no other errors, and will only persist
+	// the last of any duplicate fields. This is the default in v1.23+
+	// - Strict: This will fail the request with a BadRequest error if
+	// any unknown fields would be dropped from the object, or if any
+	// duplicate fields are present. The error returned from the server
+	// will contain all unknown and duplicate fields encountered.
+	// +optional
```
sbueringer, line 239: This seems like one of the very few times that we use the optional marker here. 

Not sure what we want to do, but consistency would be probably (if we have optional only on some fields this would imply the others are required?)

pkg/client/fieldvalidation.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
```
sbueringer, line 27: Doesn't have to block this PR, but I think we should find a more scalable pattern to wrap clients.

If we continue to introduce a new wrapper for every single option we will end up with a lot of wrappers.

I guess the same way we could either:
* only have one wrapper taking various options and acting accordingly
  * This also makes the options more discoverable for users
  * Pretty sure we end up with more consistency then (see the list of wrappers below)
* add options to the main client (that's probalby not good because then they cannot be used with the fake client)


As of today we have at least:
* intercept.NewClient
* NewDryRunClient
* WithFieldOwner
* WithStrictFieldValidation

sbueringer (commented):
Overall, absolutely in favor! Nice addition

pkg/client/options.go
dlipovetsky: Just a consequence of copy-paste. I'll remove the marker.

pkg/client/fieldvalidation.go
dlipovetsky: I completely agree; it's time to find a more scalable pattern to wrap clients.

It would be a breaking change, and will require some back-and-forth. I'll open an issue.

dlipovetsky: https://github.com/kubernetes-sigs/controller-runtime/issues/2888

pkg/client/options.go
dlipovetsky: Done in https://github.com/kubernetes-sigs/controller-runtime/pull/2860/commits/725259f79a67a0bcce6f694ee3e50a26a352884b

pkg/client/fieldvalidation.go
dlipovetsky: Done in https://github.com/kubernetes-sigs/controller-runtime/pull/2860/commits/21366f20738216c29278e6b08a17c1c13116a533

pkg/client/fieldvalidation.go
sbueringer: Thx!

PR #2862, opened 2024-06-25
✨ MOD: Add a new config in manager.New to set on stopped leading function

calmkart:
manager.New() method can't set onStoppedLeading function to options, so we can't set some special LeaderCallbacks in controller-runtime New Method to create a manager.

just add a option onStoppedLeading in manager.Options and set it to manager.controllerManager.onStoppedLeading

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
+
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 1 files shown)


troy0820:
/retitle ✨ MOD: Add a new config in manager.New to set on stopped leading function

pkg/manager/manager.go
```diff
@@ -201,6 +202,10 @@ type Options struct {
 	// LeaseDuration time first.
 	LeaderElectionReleaseOnCancel bool
 
+	// OnStoppedLeading is callled when the leader election lease is lost.
+	// It can be overridden for tests.
+	OnStoppedLeading func()
```
alvaroaleman, line 207: You can just run this after the manager ended to achieve the same result or not?

calmkart: lost leader is not equal manager ended. so we can't do that in operator's preStop or other hook.

alvaroaleman: What is the difference? The manager ends after it lost the leader?

PR #2864, opened 2024-06-27
✨ Only use json-patch v4

skitt:
controller-runtime currently depends on both v4 and v5 of evanphx/json-patch. For various reasons k/k is staying on v4, so it makes sense for controller-runtime to follow suit. This drops the v5 dependency in favour of v4 only.


alvaroaleman:
I think this change would re-introduce https://github.com/kubernetes-sigs/controller-runtime/issues/2603

skitt:
> I think this change would re-introduce #2603

It would indeed, and your test and the explanations in the commit introducing it make it clear enough.

PR #2870, opened 2024-07-02
:sparkles: feat: allow adding new prometheus.Gatherers on top of default registry

jakobmoellerdev:
In case of libraries using prebuilt prometheus registries (such as controller runtime itself), it should be possible to add the gatherer to the default metrics server so that no second handler or path has to be created for them. This allows to add metrics that are precollected/registered in something outside of the own controller runtime registry

If this is rejected the only way to merge registries is to write their own custom metrics server, which seems excessive.

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -1303,6 +1303,50 @@ var _ = Describe("manger.Manager", func() {
 				Expect(ok).To(BeTrue())
 			})
 
+			It("should serve metrics in an extra registry", func() {
+				extraRegistry := prometheus.NewRegistry()
+				one := prometheus.NewCounter(prometheus.CounterOpts{
+					Name: "test_one",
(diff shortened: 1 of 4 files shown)


pkg/manager/manager_test.go
```diff
@@ -1303,6 +1303,50 @@ var _ = Describe("manger.Manager", func() {
 				Expect(ok).To(BeTrue())
 			})
 
+			It("should serve metrics in an extra registry", func() {
+				extraRegistry := prometheus.NewRegistry()
```
alvaroaleman, line 1307: The prometheus registry itself implements `Collector` so could instead just do a `metrics.Registry.MustRegister(extraRegistry)` or not?

jakobmoellerdev:
Youre completely right, what an oversight on my part. Closing out! :)

PR #2874, opened 2024-07-05
:bug: AddMetricsServerExtraHandler: Info takes key/value pairs

thriqon:
The `Info` method takes key/value pairs as arguments, not format strings. Triggering this code paths panics instead of logging the intended message.


alvaroaleman (approved):
thanks!

PR #2881, opened 2024-07-16
:sparkles: Add dynamic watcher tests

clyang82:
Add a test for dynamically adding a watcher for the controller.


alvaroaleman (approved):
thanks!

PR #2895, opened 2024-07-31
:sparkles: Workqueue: Add `controller` label

alvaroaleman:
Currently, all controller-related metrics use the `controller` label to identity the controller. The workqueue metrics however use the `name` label instead. This is very confusing and unintuitive. Add the `controller` label to the workqueue as well with the same value as the existing `name` label. This keeps the cardinality the same, is backwards-compatible and allows using one label throughout for filtering.


sbueringer:
Thx!

/lgtm

(cc @fabriziopandini @chrischdi should allows us to simplify our dashboards a bit, I'll make a note in CAPI to adjust once we bumped to the next CR release)

PR #2896, opened 2024-07-31
:warning: Deprecate client.Options.WarningHandler

dlipovetsky:
As a follow-up to #2887, and the conversation in https://github.com/kubernetes-sigs/controller-runtime/pull/2887#discussion_r1692085803, this deprecates `client.Options.WarningHandler`.

Instead of using `client.Options.WarningHandler` users define `config.WarningHandler`. This change
1. Reduces the number of "sources" of warning handler configuration from two to one, and thereby reduces confusion.
1. Allows users to define custom warning handlers, and thereby increases flexibility. 

Deprecation and future removal does not change the default behavior, which is to de-duplicate and surface warnings.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -52,6 +52,11 @@ type Options struct {
 
 	// WarningHandler is used to configure the warning handler responsible for
 	// surfacing and handling warnings messages sent by the API server.
+	//
+	// Deprecated: This is deprecated and will be removed in a future release.
+	// Deprecation and future removal does not change the default behavior, which is to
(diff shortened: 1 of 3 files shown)


dlipovetsky:
> @dlipovetsky Talked to Vince & Alvaro. We would be also fine with directly removing the field. I think that's okay as folks can just easily pass in our implementation via rest.Config

Thanks! To keep things simple, I'll open a new PR and reference this one. I think that's ok, since this PR hasn't had implementation reviews.

Issue #2897, opened 2024-07-31
Restmapper: Use aggregated discovery if available


alvaroaleman:
We've built a lazy restmapper that only fetches mappings when needed to reduce the number of api calls made and thus speed things up. In the meantime, upstream built [aggregated discovery](https://github.com/kubernetes/enhancements/issues/3352) which allows to fetch all restmappings with a single api call.

We should update our restmapper to use `aggregated discovery` if available.

/kind feature

sbueringer:
Sounds good!

Issue #2496, opened 2023-09-17
`DynamicRESTMapper` behavior changes, can't do mapping when only `resource` is set in gvr.


xuezhaojun:
Recently, we have updated `controller-runtime` from `v0.14.6` to the `latest` version, and found a behavior change of `DynamicRESTMapper`.

This is the sample code, it uses `latest` version and `v0.14.6` to call `KindFor`:

```go
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/client-go/tools/clientcmd"
	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
)

func main() {
	kubeconfigPath := "./kubeconfig.yaml"

	kubeconfig, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
	if err != nil {
		fmt.Println(err)
		return
	}

	// latest version
	// httpClient, err := rest.HTTPClientFor(kubeconfig)
	// if err != nil {
	// 	fmt.Println(err, "Unable to create http client.")
	// 	return
	// }

	// restMapper, err := apiutil.NewDynamicRESTMapper(kubeconfig, httpClient)
	// if err != nil {
	// 	fmt.Println(err, "Unable to create restmapper.")
	// 	return
	// }

	// v0.14.6
	restMapper, err := apiutil.NewDynamicRESTMapper(kubeconfig, apiutil.WithLazyDiscovery)
	if err != nil {
		fmt.Println(err, "Unable to create restmapper.")
		return
	}

	// Call restMapper on resource "deployment"
	gvk, err := restMapper.KindFor(schema.GroupVersionResource{
		Resource: "deployments",
	})
	if err != nil {
		fmt.Println(err, "Unable to get gvk.")
		return
	}

	fmt.Println(gvk)
}
```

With `v0.14.6`, it returns no error.

But with the `latest` code, it will return the error:
```
no matches for /, Resource=deployments
```

Is there any workaround for this? Do we consider this as a compatibility issue?

troy0820:
These are from the release notes of v0.15.0 which is a breaking change:
```release-notes
- client/apiutil.NewDynamicRESTMapper signature has changed and now requires an *http.Client as parameter.

- Use and offer a single dynamic lazy RESTMapper (https://github.com/kubernetes-sigs/controller-runtime/pull/2116, https://github.com/kubernetes-sigs/controller-runtime/pull/2296)
[...]

PR #2901, opened 2024-08-03
:sparkles: Use aggregated discovery if available

alvaroaleman:
This change makes the `RESTMapper` use [AggredatedDiscovery][0] if available. `AggregatedDiscovery` is beta and thus enabled by default since Kube 1.27.

What this means in particular is that during startup, we will always do a request to `/api` and `/apis`. If `AggregatedDiscovery` is enabled, that is all we have to do. If not, we have to do a third request.

The behavior for reloading remains the same: Do one request unless the request didn't include a version. In that case we have to find the group. If the group is unknown, we will keep on doing a request to `/api` and one to `/apis` to find it. If it is found we are again done in the `AggregatedDiscovery` case and have to do a third request otherwise.

All in all this means that with `AggregatedDiscovery` enabled, the only case where this is worse is if the client interacts with only one `GroupVersion`, as we end up doing one additional api request. If it is disabled, we effectively waste two api requests.

[0]: https://github.com/kubernetes/enhancements/issues/3352

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2897

Note: Disable whitespace when reviewing, that significantly reduces the size of the diff.

### DIFF
--- a/pkg/client/apiutil/apimachinery_test.go
+++ b/pkg/client/apiutil/apimachinery_test.go
@@ -18,6 +18,7 @@ package apiutil_test
 
 import (
 	"context"
+	"strconv"
 	"testing"
 
 	gmg "github.com/onsi/gomega"
@@ -32,127 +33,130 @@ import (
 )
 
 func TestApiMachinery(t *testing.T) {
-	restCfg, tearDownFn := setupEnvtest(t)
-	defer tearDownFn(t)
+	for _, aggregatedDiscovery := range []bool{true, false} {
+		t.Run("aggregatedDiscovery="+strconv.FormatBool(aggregatedDiscovery), func(t *testing.T) {
+			restCfg := setupEnvtest(t, !aggregatedDiscovery)
 
-	// Details of the GVK registered at initialization.
-	initialGvk := metav1.GroupVersionKind{
-		Group:   "crew.example.com",
-		Version: "v1",
-		Kind:    "Driver",
-	}
-
-	// A set of GVKs to register at runtime with varying properties.
-	runtimeGvks := []struct {
-		name   string
-		gvk    metav1.GroupVersionKind
-		plural string
-	}{
-		{
-			name: "new Kind and Version added to existing Group",
-			gvk: metav1.GroupVersionKind{
-				Group:   "crew.example.com",
-				Version: "v1alpha1",
-				Kind:    "Passenger",
-			},
-			plural: "passengers",
-		},
-		{
-			name: "new Kind added to existing Group and Version",
-			gvk: metav1.GroupVersionKind{
+			// Details of the GVK registered at initialization.
+			initialGvk := metav1.GroupVersionKind{
 				Group:   "crew.example.com",
 				Version: "v1",
-				Kind:    "Garage",
-			},
-			plural: "garages",
-		},
-		{
-			name: "new GVK",
-			gvk: metav1.GroupVersionKind{
-				Group:   "inventory.example.com",
-				Version: "v1",
-				Kind:    "Taxi",
-			},
-			plural: "taxis",
-		},
-	}
+				Kind:    "Driver",
+			}
 
-	t.Run("IsGVKNamespaced should report scope for GVK registered at initialization", func(t *testing.T) {
-		g := gmg.NewWithT(t)
+			// A set of GVKs to register at runtime with varying properties.
+			runtimeGvks := []struct {
+				name   string
+				gvk    metav1.GroupVersionKind
+				plural string
+			}{
+				{
+					name: "new Kind and Version added to existing Group",
+					gvk: metav1.GroupVersionKind{
+						Group:   "crew.example.com",
+						Version: "v1alpha1",
+						Kind:    "Passenger",
+					},
+					plural: "passengers",
+				},
+				{
+					name: "new Kind added to existing Group and Version",
+					gvk: metav1.GroupVersionKind{
+						Group:   "crew.example.com",
+						Version: "v1",
+						Kind:    "Garage",
+					},
+					plural: "garages",
+				},
+				{
+					name: "new GVK",
+					gvk: metav1.GroupVersionKind{
+						Group:   "inventory.example.com",
+						Version: "v1",
+						Kind:    "Taxi",
+					},
+					plural: "taxis",
+				},
+			}
 
-		httpClient, err := rest.HTTPClientFor(restCfg)
-		g.Expect(err).NotTo(gmg.HaveOccurred())
+			t.Run("IsGVKNamespaced should report scope for GVK registered at initialization", func(t *testing.T) {
+				g := gmg.NewWithT(t)
 
-		lazyRestMapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
-		g.Expect(err).NotTo(gmg.HaveOccurred())
+				httpClient, err := rest.HTTPClientFor(restCfg)
+				g.Expect(err).NotTo(gmg.HaveOccurred())
 
-		s := scheme.Scheme
-		err = apiextensionsv1.AddToScheme(s)
-		g.Expect(err).NotTo(gmg.HaveOccurred())
+				lazyRestMapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
+				g.Expect(err).NotTo(gmg.HaveOccurred())
 
-		// Query the scope of a GVK that was registered at initialization.
-		scope, err := apiutil.IsGVKNamespaced(
-			schema.GroupVersionKind(initialGvk),
-			lazyRestMapper,
-		)
-		g.Expect(err).NotTo(gmg.HaveOccurred())
-		g.Expect(scope).To(gmg.BeTrue())
-	})
+				s := scheme.Scheme
+				err = apiextensionsv1.AddToScheme(s)
+				g.Expect(err).NotTo(gmg.HaveOccurred())
 
-	for _, runtimeGvk := range runtimeGvks {
-		t.Run("IsGVKNamespaced should report scope for "+runtimeGvk.name, func(t *testing.T) {
-			g := gmg.NewWithT(t)
-			ctx := context.Background()
-
-			httpClient, err := rest.HTTPClientFor(restCfg)
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-
-			lazyRestMapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-
-			s := scheme.Scheme
-			err = apiextensionsv1.AddToScheme(s)
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-
-			c, err := client.New(restCfg, client.Options{Scheme: s})
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-
-			// Run a valid query to initialize cache.
-			scope, err := apiutil.IsGVKNamespaced(
-				schema.GroupVersionKind(initialGvk),
-				lazyRestMapper,
-			)
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-			g.Expect(scope).To(gmg.BeTrue())
-
-			// Register a new CRD at runtime.
-			crd := newCRD(ctx, g, c, runtimeGvk.gvk.Group, runtimeGvk.gvk.Kind, runtimeGvk.plural)
-			version := crd.Spec.Versions[0]
-			version.Name = runtimeGvk.gvk.Version
-			version.Storage = true
-			version.Served = true
-			crd.Spec.Versions = []apiextensionsv1.CustomResourceDefinitionVersion{version}
-			crd.Spec.Scope = apiextensionsv1.NamespaceScoped
-
-			g.Expect(c.Create(ctx, crd)).To(gmg.Succeed())
-			t.Cleanup(func() {
-				g.Expect(c.Delete(ctx, crd)).To(gmg.Succeed())
+				// Query the scope of a GVK that was registered at initialization.
+				scope, err := apiutil.IsGVKNamespaced(
+					schema.GroupVersionKind(initialGvk),
+					lazyRestMapper,
+				)
+				g.Expect(err).NotTo(gmg.HaveOccurred())
+				g.Expect(scope).To(gmg.BeTrue())
 			})
 
-			// Wait until the CRD is registered.
-			g.Eventually(func(g gmg.Gomega) {
-				isRegistered, err := isCrdRegistered(restCfg, runtimeGvk.gvk)
-				g.Expect(err).NotTo(gmg.HaveOccurred())
-				g.Expect(isRegistered).To(gmg.BeTrue())
-			}).Should(gmg.Succeed(), "GVK should be available")
+			for _, runtimeGvk := range runtimeGvks {
+				t.Run("IsGVKNamespaced should report scope for "+runtimeGvk.name, func(t *testing.T) {
+					g := gmg.NewWithT(t)
+					ctx := context.Background()
 
-			// Query the scope of the GVK registered at runtime.
-			scope, err = apiutil.IsGVKNamespaced(
-				schema.GroupVersionKind(runtimeGvk.gvk),
-				lazyRestMapper,
-			)
-			g.Expect(err).NotTo(gmg.HaveOccurred())
-			g.Expect(scope).To(gmg.BeTrue())
+					httpClient, err := rest.HTTPClientFor(restCfg)
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+
+					lazyRestMapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+
+					s := scheme.Scheme
+					err = apiextensionsv1.AddToScheme(s)
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+
+					c, err := client.New(restCfg, client.Options{Scheme: s})
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+
+					// Run a valid query to initialize cache.
+					scope, err := apiutil.IsGVKNamespaced(
+						schema.GroupVersionKind(initialGvk),
+						lazyRestMapper,
+					)
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+					g.Expect(scope).To(gmg.BeTrue())
+
+					// Register a new CRD at runtime.
+					crd := newCRD(ctx, g, c, runtimeGvk.gvk.Group, runtimeGvk.gvk.Kind, runtimeGvk.plural)
+					version := crd.Spec.Versions[0]
+					version.Name = runtimeGvk.gvk.Version
+					version.Storage = true
+					version.Served = true
+					crd.Spec.Versions = []apiextensionsv1.CustomResourceDefinitionVersion{version}
+					crd.Spec.Scope = apiextensionsv1.NamespaceScoped
+
+					g.Expect(c.Create(ctx, crd)).To(gmg.Succeed())
+					t.Cleanup(func() {
+						g.Expect(c.Delete(ctx, crd)).To(gmg.Succeed())
+					})
+
+					// Wait until the CRD is registered.
+					g.Eventually(func(g gmg.Gomega) {
+						isRegistered, err := isCrdRegistered(restCfg, runtimeGvk.gvk)
+						g.Expect(err).NotTo(gmg.HaveOccurred())
+						g.Expect(isRegistered).To(gmg.BeTrue())
+					}).Should(gmg.Succeed(), "GVK should be available")
+
+					// Query the scope of the GVK registered at runtime.
+					scope, err = apiutil.IsGVKNamespaced(
+						schema.GroupVersionKind(runtimeGvk.gvk),
+						lazyRestMapper,
+					)
+					g.Expect(err).NotTo(gmg.HaveOccurred())
+					g.Expect(scope).To(gmg.BeTrue())
+				})
+			}
 		})
 	}
 }
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
+	"k8s.io/utils/ptr"
 )
 
 // NewDynamicRESTMapper returns a dynamic RESTMapper for cfg. The dynamic
@@ -41,6 +42,7 @@ func NewDynamicRESTMapper(cfg *rest.Config, httpClient *http.Client) (meta.RESTM
 	if err != nil {
 		return nil, err
 	}
+
 	return &mapper{
 		mapper:      restmapper.NewDiscoveryRESTMapper([]*restmapper.APIGroupResources{}),
 		client:      client,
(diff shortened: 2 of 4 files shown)


pkg/client/apiutil/restmapper.go
```diff
@@ -230,35 +260,43 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
 	for i := range apiGroups.Groups {
 		group := &apiGroups.Groups[i]
 		m.apiGroups[group.Name] = group
 	}
+	for failedGV := range failedGVs {
+		delete(maybeResources, failedGV)
+	}
+	m.checkedIfSupportsAggregatedDiscovery.Store(true)
+	if len(maybeResources) > 0 {
+		m.supportsAggregatedDiscovery.Store(true)
```
sbueringer, line 297: Why is it correct to check len(maybeResources) > 0 after failedGVs have been removed to infer that aggregated discovery is supported?

I would have assumed it is supported as soon as len(maybeResources) > 0 independent of the failedGVs

pkg/client/apiutil/restmapper.go
```diff
@@ -194,13 +209,28 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
+	return nil
+}
 
+// addGroupVersionResourcesToCache does what the name suggests. The mutex must be held when
+// calling it.
+func (m *mapper) addGroupVersionResourcesToCache(gvr map[schema.GroupVersion]*metav1.APIResourceList) {
 	// Update information for group resources about the API group by adding new versions.
-	// Ignore the versions that are already registered.
-	for groupVersion, resources := range groupVersionResources {
+	// Ignore the versions that are already registered
```
sbueringer, line 222: Should we return early if len(groupVersionResources) == 0? (I think in that case it would recreate the mapper for nothing, basically a no-op)

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +168,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 	// If no specific versions are set by user, we will scan all available ones for the API group.
 	// This operation requires 2 requests: /api and /apis, but only once. For all subsequent calls
 	// this data will be taken from cache.
-	if len(versions) == 0 {
-		apiGroup, err := m.findAPIGroupByName(groupName)
+	//
+	// Calling this also tells us if the server supports aggregated discovery, which allows
+	// loading everything with two api calls which we assume is overall cheaper, hence we always
+	// do this once optimistically to check and keep doing it if its supported.
```
sbueringer, line 174: How is the "keep doing it if its supported" part implemented?

As far as I can tell we just always try it by calling GroupsAndMaybeResources, independent of if it's supported or not (which I think is correct, just a bit confused by the comment)

pkg/client/apiutil/apimachinery_test.go
```diff
@@ -18,6 +18,7 @@ package apiutil_test
```
sbueringer, line 18: nit for the PR description. Based on Kubernetes docs this is enabled per default since 1.27: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +168,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
-	// Create or fetch group resources from cache.
-	groupResources := &restmapper.APIGroupResources{
-		Group:              metav1.APIGroup{Name: groupName},
-		VersionedResources: make(map[string][]metav1.APIResource),
+		// No need to do anything further if aggregatedDiscovery is supported and we did a lookup
+		if m.supportsAggregatedDiscovery.Load() && didLookup {
+			failedGroups := make(map[schema.GroupVersion]error)
+			for _, version := range versions {
+				if m.knownGroups[groupName] == nil || m.knownGroups[groupName].VersionedResources[version] == nil {
+					failedGroups[schema.GroupVersion{Group: groupName, Version: version}] = &meta.NoResourceMatchError{}
```
sbueringer, line 191: Maybe worth a comment that we'll hit this case for:
* groupVersions that don't exist
* groupVersions that are stale (via aggregated discovery)

If I understand this correctly

sbueringer: Although I think we hit the latter case only if the groupsVersions always were stale?

sbueringer: Makes me wonder if we should drop stale groups from the cache or keep serving stale data. Probably the latter is better though

pkg/client/apiutil/restmapper.go
alvaroaleman: That comment is outdated actually, I originally made it always call only this if supported but later realized that for the reloading case, we know the GV and directly fetching that is cheaper as its only one api call vs two here. Updated the go docs accordingly.

pkg/client/apiutil/restmapper.go
alvaroaleman: Yeah I wasn't entirely sure what the right thing to do there is either. If I understand the kep correctly, stale just means that the CRD is not yet initialized which should rarely happen

pkg/client/apiutil/restmapper.go
alvaroaleman: What codepath do you see where this could happen (len(gvk) == 0)?

pkg/client/apiutil/restmapper.go
alvaroaleman: Good point, updated

pkg/client/apiutil/restmapper.go
```diff
@@ -230,35 +260,43 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
 		group := &apiGroups.Groups[i]
 		m.apiGroups[group.Name] = group
 	}
+	for failedGV := range failedGVs {
+		delete(maybeResources, failedGV)
+	}
+	m.checkedIfSupportsAggregatedDiscovery.Store(true)
+	if len(maybeResources) > 0 {
+		m.supportsAggregatedDiscovery.Store(true)
+	}
```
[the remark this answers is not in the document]
alvaroaleman: Hm yeah that is true but we only use this info to short circuit after the `FindAPIGroupByNameAndMaybeAggregatedDiscover` call, it is unused after

alvaroaleman:
@sbueringer one thing probably worth pointing out: Originally I thought this is always better in terms of api calls as it is one api call for everything. I then learned when implementing this that its actually two api calls for everything: One to /api, one to /apis. That means its not as obviously better as I originally thought: If one GV is used by the binary its worse. If two are used, its the same. Three or more, its better (but I would assume the majority of controllers uses two or less).

The other reason why _I think_ it would be nice is that it brings us closer to the upstream restmapper if aggrgated discovery is supported, as it means we have all available info in the cache and issues like https://github.com/kubernetes-sigs/controller-runtime/issues/2496 wouldn't happen. But I don't know if that is a super strong argument.

pkg/client/apiutil/apimachinery_test.go
alvaroaleman: Updated

pkg/client/apiutil/restmapper.go
sbueringer: I was thinking about
```go
m.addGroupVersionResourcesToCache(maybeResources)
```
for the case where aggregated discovery is not enabled

pkg/client/apiutil/restmapper.go
sbueringer: Thx! That clears up a lot of confusion on my side :)

pkg/client/apiutil/restmapper.go
sbueringer: My take on the KEP was that stale (also?) happens if the kube-apiserver is not able to communicate with aggregated apiserves anymore

sbueringer: Definiteily for initial sync, but probably also later if that happens

```
No, but if aggregated apiservers are present, the feature will attempt to contact and aggregate the data published from the aggregated apiserver on a set interval. If there is high error rate, stale data may be returned because the latest data was not able to be fetched from the aggregated apiserver.
```

sbueringer:
Just to avoid misunderstandings

>  If one GV is used by the binary its worse. If two are used, its the same. Three or more, its better (but I would assume the majority of controllers uses two or less).

With "GV used" you basically mean that the controller will ask the restmapper for a specific GV. I.e. "two are used" means that the restmapper would be used for at least two different GVs.

I think it's safe to assume that most controllers will use at least 2 GVs. At least 1 for their CRD and then in a lot of cases either another one for other CRDs or builtin resources. 

If I got it correctly the restmapper should be used for every informer / get / list / write / ... calls. I think a lot of controllers are not restricted to just one GV

pkg/client/apiutil/restmapper.go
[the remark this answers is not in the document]
sbueringer: Yup. Just wondering if it would be semantically better to additionally return a `didAggregatedLookup` bool and then use that to short circuit.

At the moment we always short circuit as soon as GroupsAndMaybeResources returned resources once. If later calls to GroupsAndMaybeResources didn't return resources (because we hit an apiserver without the feature) we still short circuit even though we shouldn't

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +168,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 	// If no specific versions are set by user, we will scan all available ones for the API group.
 	// This operation requires 2 requests: /api and /apis, but only once. For all subsequent calls
 	// this data will be taken from cache.
-	if len(versions) == 0 {
-		apiGroup, err := m.findAPIGroupByName(groupName)
+	//
+	// Calling this also tells us if the server supports aggregated discovery, which allows
+	// loading everything with two api calls which we assume is overall cheaper, hence we always
+	// do this once optimistically.
+	if len(versions) == 0 || !m.checkedIfSupportsAggregatedDiscovery.Load() {
```
sbueringer, line 175: (I assume that aggregated discovery working once doesn't tell us anything about future discovery calls)

Would it then be better to call this something like `m.initialDiscoveryDone`?

The current name (+ the fact that we only ever set it to true) sort of implies that we check once if aggregated discovery is supported and then we'll do something with that information / it stays true for the lifetime of the binary.

pkg/client/apiutil/restmapper.go
sbueringer: I think the alternative would be to differentiate between NoResourceMatchError and StaleGroupVersionError.

But maybe it's fine to surface always NoResourceMatchError basically StaleGroupVersionError means that the apiserver didn't fine a "fresh" version of the resource. Maybe NoResourceMatchError is close in enough even for that case?

pkg/client/apiutil/restmapper.go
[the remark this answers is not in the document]
sbueringer: (with didAggregatedLookup we can also get rid of the existing didLookup I think)

pkg/client/apiutil/restmapper.go
alvaroaleman: But should we return this as an error? After re-reading, I actually removed the deletion of stale groups, because I don't think that restmappings change that often, so having potentially stale informations is likely better than having no information. At the end of the day, controllers will retry anyways should there be a missing mapping and that triggers a new non-aggregated discovery call.

pkg/client/apiutil/restmapper.go
alvaroaleman: Good idea, I've renamed it to `initialDiscoveryDone`. I've also removed the `supportsAggregatedDiscovery` and changed the code to instead rely on the bool returned by `findAPIGroupByNameAndMaybeAggregatedDiscovery` to find out if aggregated discovery was done.

pkg/client/apiutil/restmapper.go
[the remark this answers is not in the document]
alvaroaleman: > Yup. Just wondering if it would be semantically better to additionally return a didAggregatedLookup (or didAggregatedDiscovery) bool and then use that to short circuit.

Done that

pkg/client/apiutil/restmapper.go
sbueringer: Sounds reasonable. I'll take another look at the code

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +167,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 	// If no specific versions are set by user, we will scan all available ones for the API group.
 	// This operation requires 2 requests: /api and /apis, but only once. For all subsequent calls
 	// this data will be taken from cache.
-	if len(versions) == 0 {
-		apiGroup, err := m.findAPIGroupByName(groupName)
+	//
+	// Calling this also tells us if the server supports aggregated discovery, which allows
```
sbueringer, line 171: really small nit. "also tells us if the server supports aggregated discovery"

I guess calling "findAPIGroupByNameAndMaybeAggregatedDiscovery" only tells us that aggregated discovery worked this time (nothing about future calls)

WDYT about:
```go
	// We always run this once, because if the server supports aggregated discovery, this will
	// load everything with two api calls which we assume is overall cheaper.
```

(but feel free to ignore if you prefer the current version or an alternative)

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +167,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
-		Group:              metav1.APIGroup{Name: groupName},
-		VersionedResources: make(map[string][]metav1.APIResource),
+		// No need to do anything further if aggregatedDiscovery is supported and we did a lookup
+		if didAggregatedDiscovery {
+			failedGroups := make(map[schema.GroupVersion]error)
+			for _, version := range versions {
+				if m.knownGroups[groupName] == nil || m.knownGroups[groupName].VersionedResources[version] == nil {
+					failedGroups[schema.GroupVersion{Group: groupName, Version: version}] = &meta.NoResourceMatchError{}
+				}
+			}
```
sbueringer, line 192: (I intentionally put this comment on a different line to start a new conversation)

WDYT about setting `PartialResource` in `NoResourceMatchError`. I think if we set a GVR with group and version we can get slightly better error messages

pkg/client/apiutil/restmapper.go
```diff
@@ -194,13 +208,28 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 		return fmt.Errorf("failed to get API group resources: %w", err)
```
sbueringer, line 208: I think we now don't have the lock anymore when calling fetchGroupVersionResourcesLocked in l.206. Probably we just have to move it up from below

pkg/client/apiutil/restmapper.go
```diff
@@ -194,13 +208,28 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
-	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.addGroupVersionResourcesToCache(groupVersionResources)
+	return nil
+}
 
+// addGroupVersionResourcesToCache does what the name suggests. The mutex must be held when
+// calling it.
+func (m *mapper) addGroupVersionResourcesToCache(gvr map[schema.GroupVersion]*metav1.APIResourceList) {
```
sbueringer, line 219:
```suggestion
func (m *mapper) addGroupVersionResourcesToCacheLocked(gvr map[schema.GroupVersion]*metav1.APIResourceList) {
```
nit

Just to be a bit safer going forward. Can you please go through all funcs in this file and add a Locked suffix to all that assume that a lock is held? (a few more example are: isGroupVersionCached, isAPIGroupCached)

I think it should be at least all funcs that are accessing m.knownGroups or m.apiGroups without locking themselves

pkg/client/apiutil/restmapper.go
```diff
@@ -162,25 +167,34 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
 
-	// Create or fetch group resources from cache.
-	groupResources := &restmapper.APIGroupResources{
-		Group:              metav1.APIGroup{Name: groupName},
-		VersionedResources: make(map[string][]metav1.APIResource),
+		// No need to do anything further if aggregatedDiscovery is supported and we did a lookup
+		if didAggregatedDiscovery {
+			failedGroups := make(map[schema.GroupVersion]error)
+			for _, version := range versions {
+				if m.knownGroups[groupName] == nil || m.knownGroups[groupName].VersionedResources[version] == nil {
```
sbueringer, line 189: Do we need a read lock to access knownGroups?

pkg/client/apiutil/restmapper.go
```diff
@@ -230,35 +259,36 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
 
 	m.mu.Lock()
-	for i := range apiGroups.Groups {
-		group := &apiGroups.Groups[i]
-		m.apiGroups[group.Name] = group
+	m.initialDiscoveryDone.Store(true)
+	if len(maybeResources) > 0 {
+		didAggregatedDiscovery = true
 	}
+	m.addGroupVersionResourcesToCache(maybeResources)
```
sbueringer, line 291: Follow-up to: https://github.com/kubernetes-sigs/controller-runtime/pull/2901#discussion_r1706902772

We can also only call addGroupVersionResourcesToCache if it is actually necessary, e.g.:
```go
	m.initialDiscoveryDone.Store(true)
	if len(maybeResources) > 0 {
		didAggregatedDiscovery = true
		m.mu.Lock()
		m.addGroupVersionResourcesToCache(maybeResources)
		m.mu.Unlock()
	}
```

pkg/client/apiutil/restmapper.go
sbueringer: Added a new comment with a different idea

pkg/client/apiutil/restmapper.go
sbueringer: Looks good!

Also it looks like failedGVs aren't even added to resources: https://github.com/kubernetes/client-go/blob/master/discovery/aggregated_discovery.go#L82-L83

(so the previous code was a no-op with the current aggregated discovery implementation)

pkg/client/apiutil/restmapper.go
```diff
@@ -267,7 +297,7 @@ func (m *mapper) findAPIGroupByName(groupName string) (*metav1.APIGroup, error)
 
 	// Don't return an error here if the API group is not present.
 	// The reloaded RESTMapper will take care of returning a NoMatchError.
-	return m.apiGroups[groupName], nil
+	return m.apiGroups[groupName], didAggregatedDiscovery, nil
 }
 
 // fetchGroupVersionResourcesLocked fetches the resources for the specified group and its versions.
```
sbueringer, line 303: Looks like the last change broke a few unit tests

pkg/client/apiutil/restmapper.go
alvaroaleman: Done

pkg/client/apiutil/restmapper.go
alvaroaleman: This was a good call. Updated things as suggested but also changed `addKnownGroupAndReload` to hold the lock everywhere rather than just when its accessing data stuctures that need locking. The reason is that without that, its possible andlikely during startup,that we might have multiple calls to `findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked` going on in parallel, as the previously-unlocked api calls take much longer than updating the locked data structures. 

As `fetchGroupVersionResources` already required us to hold the lock there is no case where things might get worse.

pkg/client/apiutil/restmapper.go
alvaroaleman: Fixed, see other comment

pkg/client/apiutil/restmapper.go
alvaroaleman: Updated

pkg/client/apiutil/restmapper.go
alvaroaleman: Added

pkg/client/apiutil/restmapper.go
alvaroaleman: Its moved up

pkg/client/apiutil/restmapper.go
sbueringer: Sounds good, thx!

pkg/client/apiutil/restmapper.go
```diff
@@ -159,28 +165,42 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
 		versions = nil
 	}
 
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
sbueringer, line 169: So much simpler, should have done this earlier :)

pkg/client/apiutil/restmapper.go
```diff
@@ -214,14 +247,14 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
 		}
-	}
 
-	// Update data in the cache.
-	m.knownGroups[groupName] = groupResources
+		// Update data in the cache.
+		m.knownGroups[groupVersion.Group] = groupResources
+	}
 
 	// Finally, update the group with received information and regenerate the mapper.
```
sbueringer, line 259:
```suggestion
	// Finally, reload the mapper.
```
nit: I think

pkg/client/apiutil/restmapper.go
```diff
@@ -194,13 +214,26 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
-	if _, ok := m.knownGroups[groupName]; ok {
-		groupResources = m.knownGroups[groupName]
-	}
+	m.addGroupVersionResourcesToCacheLocked(groupVersionResources)
+	return nil
+}
 
+// addGroupVersionResourcesToCacheLocked does what the name suggests. The mutex must be held when
+// calling it.
+func (m *mapper) addGroupVersionResourcesToCacheLocked(gvr map[schema.GroupVersion]*metav1.APIResourceList) {
```
sbueringer, line 223: I know the name of this func is getting long, but might be still good to mention that it also reloads the mapper, maybe `addGroupVersionResourcesToCacheAndReloadLocked` (would align to `addKnownGroupAndReload`)

pkg/client/apiutil/restmapper.go
```diff
@@ -230,44 +263,40 @@ func (m *mapper) addKnownGroupAndReload(groupName string, versions ...string) er
...
-func (m *mapper) findAPIGroupByName(groupName string) (*metav1.APIGroup, error) {
-	// Looking in the cache first.
-	{
-		m.mu.RLock()
-		group, ok := m.apiGroups[groupName]
-		m.mu.RUnlock()
-		if ok {
-			return group, nil
-		}
+// findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked tries to finds the passed apiGroup.
```
sbueringer, line 268:
```suggestion
// findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked tries to find the passed apiGroup.
```
nit

sbueringer (commented):
Very nice! Last round of nits from my side

pkg/client/apiutil/restmapper.go
alvaroaleman: Done

pkg/client/apiutil/restmapper.go
alvaroaleman: done

pkg/client/apiutil/restmapper.go
alvaroaleman: done

sbueringer:
Thank you very much! :)

/lgtm
/approve

PR #2902, opened 2024-08-03
:warning: Validate controller names are unique

alvaroaleman:
When re-using the same name among multiple controllers, they will report to the same metrics and use the same logger. This can be very confusing, might not be obvious and can happen accidentally.

Validate controller names are unique at runtime to avoid all of this.

### DIFF
--- a/pkg/controller/name.go
+++ b/pkg/controller/name.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var nameLock sync.Mutex
+var usedNames sets.Set[string]
+
+func checkName(name string) error {
+	nameLock.Lock()
+	defer nameLock.Unlock()
+	if usedNames == nil {
+		usedNames = sets.Set[string]{}
+	}
+
+	if usedNames.Has(name) {
+		return fmt.Errorf("controller with name %s already exists. Controller names must be unique to avoid multiple controllers reporting to the same metric", name)
+	}
+
+	usedNames.Insert(name)
+
+	return nil
+}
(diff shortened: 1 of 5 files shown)


pkg/controller/name.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var nameLock sync.Mutex
+var usedNamed sets.Set[string]
```
sbueringer, line 27:
```suggestion
var usedNames sets.Set[string]
```
I think

sbueringer:
Otherwise lgtm from my side

sbueringer:
Thx!

/lgtm
/approve

mythi:
FWIW, this seems to trigger an error in our `envtest` based tests where a new manager is started `BeforeEach` and its context cancelled `AfterEach`. I might have a race or doing something wrong, though. :thinking: 

AFAUI, the manager gets closed
```
...
I0812 16:39:27.717706 1182770 internal.go:550] "Stopping and waiting for caches"
I0812 16:39:27.718212 1182770 internal.go:554] "Stopping and waiting for webhooks"
I0812 16:39:27.718251 1182770 internal.go:557] "Stopping and waiting for HTTP servers"
I0812 16:39:27.718270 1182770 internal.go:561] "Wait completed, proceeding to shutdown the manager"
```

but the next test returns an error added here.

```
  [FAILED] Expected
      <*errors.errorString | 0xc000804160>: 
      controller with name <myname> already exists. Controller names must be unique to avoid multiple controllers reporting to the same metric
      {
          s: "controller with name <myname> already exists. Controller names must be unique to avoid multiple controllers reporting to the same metric",
      }
  to be nil
```

sbueringer:
@mythi @alvaroaleman PR: https://github.com/kubernetes-sigs/controller-runtime/pull/2918

mythi:
> @mythi @alvaroaleman PR: #2918

thanks! I'll give it a try on Tue.

pkg/controller/name.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var nameLock sync.Mutex
+var usedNames sets.Set[string]
```
RainbowMango, line 27: Hi @alvaroaleman @sbueringer 
Holding controller name with a global variable seems not a good idea, because the variable will be shared between tests in a test suite or in a test package. 

To be more specific, see following the test sample below:
```golang
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			mgr, err := controllerruntime.NewManager(&rest.Config{}, controllerruntime.Options{})

			controller := &FooController{}
			err = controller.SetupWithManager(mgr)
			assert.NoError(t, err) // raise error: Controller names must be unique to avoid multiple controllers reporting to the same metric

			// ... others
		})
	}
```
Even though each test creates a brand new controller manager, but the `usedNamed` is shared unexpected.

RainbowMango: My two cents:
The used controller names should be stored in a specific manager instance.

mythi: @RainbowMango my tests have the following: `ctrl.NewManager(cfg, ctrl.Options{Scheme: scheme.Scheme, Metrics: metricsserver.Options{BindAddress: "0"}, Controller: config.Controller{SkipNameValidation: &yes}})`

RainbowMango: Yeah, thanks. @mythi 
It's not a blocker for me.
I noticed the option as well, but I think it's more like a workaround.

PR #2903, opened 2024-08-05
:warning: Remove options.WarningHandler

dlipovetsky:
As a follow-up to https://github.com/kubernetes-sigs/controller-runtime/pull/2887, and https://github.com/kubernetes-sigs/controller-runtime/pull/2896, this removes client.Options.WarningHandler.

Instead of using client.Options.WarningHandler users define config.WarningHandler. This change:

- Reduces the number of "sources" of warning handler configuration from two to one, and thereby reduces confusion.
- Allows users to define custom warning handlers, and thereby increases flexibility.
- Removal does not change the default behavior, which is to de-duplicate and surface warnings.

Note: Because the PR changes the go module public API, the "pull-controller-runtime-apidiff" test fails, as expected.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -50,26 +50,8 @@ type Options struct {
 	// Cache, if provided, is used to read objects from the cache.
 	Cache *CacheOptions
 
-	// WarningHandler is used to configure the warning handler responsible for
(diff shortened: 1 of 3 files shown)


sbueringer:
Thx!

/lgtm

PR #2905, opened 2024-08-06
⚠ Recover panics per default

### LINKED ISSUE #2167: RecoverPanic - true by default
Just curious why this isn't set to true by default. Would anyone object if I made a PR to enable by default? https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/config/controller.go#L44

sbueringer:

Recover panics in reconcilers & webhooks per default. Also adds corresponding metrics.

Fixes #2167

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -85,8 +85,8 @@ func (blder *WebhookBuilder) WithLogConstructor(logConstructor func(base logr.Lo
 
 // RecoverPanic indicates whether panics caused by the webhook should be recovered.
 // Defaults to true.
-func (blder *WebhookBuilder) RecoverPanic(recoverPanic *bool) *WebhookBuilder {
-	blder.recoverPanic = recoverPanic
+func (blder *WebhookBuilder) RecoverPanic(recoverPanic bool) *WebhookBuilder {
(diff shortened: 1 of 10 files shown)


sbueringer:
/hold

I'll do a bit of testing

pkg/builder/webhook.go
```diff
@@ -84,8 +84,9 @@ func (blder *WebhookBuilder) WithLogConstructor(logConstructor func(base logr.Lo
 }
 
 // RecoverPanic indicates whether panics caused by the webhook should be recovered.
-func (blder *WebhookBuilder) RecoverPanic() *WebhookBuilder {
-	blder.recoverPanic = true
+// Defaults to true.
+func (blder *WebhookBuilder) RecoverPanic(recoverPanic *bool) *WebhookBuilder {
```
alvaroaleman, line 88: Why take a bool pointer and not a value?

sbueringer: Also sort of tried to be consistent with the  RecoverPanic *bool option we have for controllres, but I'm aware that we don't need that for these funcs :)

sbueringer: Fixed

sbueringer:
@alvaroaleman fixed, PTAL :)

PR #2909, opened 2024-08-07
🌱 Enforce correct file headers

sbueringer:

The boilerplate script has been mostly copied from Cluster API which adapted it from k/k


sbueringer:
@alvaroaleman @vincepri WDYT is that something we want to enforce? (if yes, I'll fix the open findings)

Issue #2888, opened 2024-07-24
Find one scalable pattern to wrap clients, and deprecate others


dlipovetsky:
As @sbueringer observed in https://github.com/kubernetes-sigs/controller-runtime/pull/2860#discussion_r1673871790,
we have, as of now, at least four different ways to wrap the client:

> - intercept.NewClient
> - NewDryRunClient
> - WithFieldOwner
> - WithStrictFieldValidation
>
> ... I think we should find a more scalable pattern to wrap clients.
> If we continue to introduce a new wrapper for every single option we will end up with a lot of wrappers.
>
> If we continue to introduce a new wrapper for every single option we will end up with a lot of wrappers.

> 
> I guess the same way we could either:
> 
> - only have one wrapper taking various options and acting accordingly
>   - This also makes the options more discoverable for users
>   - Pretty sure we end up with more consistency then (see the list of wrappers below)
> - add options to the main client (that's probalby not good because then they cannot be used with the fake client)

/kind feature

alvaroaleman:
When building this we should also try to solve: https://github.com/kubernetes-sigs/controller-runtime/issues/2967
All the wrappers we have don't properly deal with `client.WithWatch` and end up removing the `Watch` functionality.

sbueringer:
> All the wrappers we have don't properly deal with client.WithWatch and end up removing the Watch functionality.

To expand on this. We also have (at least one) wrapper like `intercepter.NewClient` which take and return client.WithWatch which makes it very annoying to use them with regular clients that don't implement the client.WithWatch interface (most of our clients don't implement client.WithWatch)

Maybe it would help if more clients would implement client.WithWatch, but not sure if it e.g. makes sense for CacheReader

yindia:
@sbueringer @dlipovetsky Took a closer look and put together a concrete direction, sharing here so maintainers can weigh in before I open a PR.

Of the four wrappers, three are the same shape: `NewDryRunClient`, `WithFieldOwner`, and `WithFieldValidation` each wrap a `Client` and inject one write option into every mutating verb and its subresource equivalent. The fourth, `intercept.NewClient`, is a different concern. It replaces whole method behavior with user supplied funcs, wraps `WithWatch` rather than `Client`, and lives in its own package, so I would leave it out of this consolidation.

Proposed shared backend for the three option injecting wrappers:

```go
type WriteOptions struct {
    Create      []CreateOption
    Update      []UpdateOption
    Patch       []PatchOption
[...]

PR #2929, opened 2024-08-18
:sparkles: Add client.WithWatch wrapper

### LINKED ISSUE #2967: Can't wrap existing client with WithWatch()
In some cases, one can wrap already-extant `client.Client`s to confer extra functionality. For example:

```go
cWithFieldOwner := client.WithFieldOwner(cWithoutFieldOwner, "owner")
```

However, it is not possible to confer `WithWatch` on an existing client: one can only obtain a `client.WithWatch` via a constructor:

```go
cWithWatch, err := client.NewWithWatch(cfg, opts)
```

The resulting client is "frozen" -- i.e. can't be subjected to decorators like `WithFieldOwner` -- because those wrappers return a `client.Client`, which can't `Watch()`!

It would be great to have a decorator function to allow an existing client to grow `WithWatch` functionality:

```go
c, err := client.New(cfg, opts)
c = client.WithFieldOwner(c, "owner")
// ... chain other decorators ...
cWithAllTheThings, err := client.AsWithWatch(c)
```

Notes:
- It would be better to call the wrapper `WithWatch`, but that name is already taken by the interface. Sad face.
- I can't think of a way to get around this client being "the end of the line". It can't be further wrapped, unless the wrapper is `WithWatch`-specific. Sad face.
- It would be great if you could do this to any `client.Client`; but as written today that would be hard because the `Watch()` implementation relies on internals of the specific `client` (lowercase) implementation in the library. Sad face.

See #2929 where I prototyped an imperfect version of the solution that suffers from all of the above issues.

2uasimojo:
In some cases, one can wrap already-extant `client.Client`s to confer extra functionality. For example:

```go
cWithFieldOwner := client.WithFieldOwner(cWithoutFieldOwner, "owner")
```

However, it has not heretofore been possible to confer `WithWatch` on an existing client: one could only obtain a `client.WithWatch` via a constructor:

```go
cWithWatch, err := client.NewWithWatch(cfg, opts)
```

The resulting client is "frozen" -- i.e. can't be subjected to decorators like `WithFieldOwner` -- because those wrappers return a `client.Client`, which can't `Watch()`!

This commit adds a decorator function to allow an existing client to grow `WithWatch` functionality:

```go
c, err := client.New(cfg, opts)
c = client.WithFieldOwner(c, "owner")
// ... chain other decorators ...
cWithAllTheThings, err := client.AsWithWatch(c)
```

Notes:
- We would have preferred to call the wrapper `WithWatch`, but that name is already taken by the interface. Sad face.
- This client is "the end of the line". It can't be further wrapped, unless the wrapper is `WithWatch`-specific. Sad face.
- As currently conceived, you can't just do this to any `client.Client` because the `Watch()` implementation relies on internals of the specific `client` (lowercase) implementation in the library. Sad face.

Closes #2967

### DIFF
--- a/pkg/client/watch.go
+++ b/pkg/client/watch.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,6 +34,15 @@ func NewWithWatch(config *rest.Config, options Options) (WithWatch, error) {
 		return nil, err
 	}
 	return &watchingClient{client: client}, nil
+}
+
+// AsWithWatch wraps an existing client in a WithWatch
+func AsWithWatch(c Client) (WithWatch, error) {
+	cl, ok := c.(*client)
+	if !ok {
+		return nil, errors.New("incompatible client type")
+	}
+	return &watchingClient{client: cl}, nil
 }
 
 type watchingClient struct {


2uasimojo:
Reviewers: I don't think this PR is acceptable without addressing at least the second and third Notes in the description.

2uasimojo:
This prototype is pretty far from a good solution. I'm hoping whatever comes out of #2888 will be clean and pretty.

PR #2931, opened 2024-08-22
🐛 Fix custom defaulter: avoid deleting unknown fields

alculquicondor:
When creating a patch for the custom defaulter, pass the original object through decoding.

Otherwise, the patch will include a remove instruction for fields that the go type doesn't contain.

This could happen when building a webhook for a resource that belongs to another project (including k8s types).

Added a test for it. This is the patch if we don't use the coder:

```golang
     <[]jsonpatch.Operation | len:2, cap:2>: [
          {Operation: "add", Path: "/replica", Value: <float64>2},
          {Operation: "remove", Path: "/newField", Value: nil},
      ]
```

### DIFF
--- a/pkg/webhook/admission/defaulter_custom.go
+++ b/pkg/webhook/admission/defaulter_custom.go
@@ -71,13 +71,14 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
 	ctx = NewContextWithRequest(ctx, req)
 
 	// Get the object in the request
-	obj := h.object.DeepCopyObject()
-	if err := h.decoder.Decode(req, obj); err != nil {
+	original := h.object.DeepCopyObject()
+	if err := h.decoder.Decode(req, original); err != nil {
 		return Errored(http.StatusBadRequest, err)
 	}
 
 	// Default the object
-	if err := h.defaulter.Default(ctx, obj); err != nil {
+	updated := original.DeepCopyObject()
+	if err := h.defaulter.Default(ctx, updated); err != nil {
 		var apiStatus apierrors.APIStatus
 		if errors.As(err, &apiStatus) {
 			return validationResponseFromStatus(false, apiStatus.Status())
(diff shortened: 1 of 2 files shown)


alculquicondor:
This will allow to fix https://github.com/kubernetes-sigs/kueue/issues/2878 once we upgrade

pkg/webhook/admission/defaulter_custom_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"gomodules.xyz/jsonpatch/v2"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var _ = Describe("CustomDefaulter Handler", func() {
+
+	It("should should not lose unknown fields", func() {
```
tenzen-y, line 18:
```suggestion
	It("should not lose unknown fields", func() {
```

:)

tenzen-y (commented):
Greate improvements!
Just a nit.

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -71,24 +71,32 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
...
 		return Denied(err.Error())
 	}
 
-	// Create the patch
-	marshalled, err := json.Marshal(obj)
+	// Create the patch.
+	// We need to decode and marshall the original because the type registered in the
+	// decoder might not match the latest version of the API.
+	// Creating a diff from the raw object might cause new fields to be dropped.
+	marshalledOriginal, err := json.Marshal(original)
```
sbueringer, line 93: I think this can lead to invalid patches. Now we are calculating patches based on how original looks like after unmarshal + mashal. But the apiserver will have to apply the patches on the original JSON object

(We had problems like this when implementing our own patch calculation like it is suggested here on this PR)

sbueringer: Example: Imagine an API type that through (probably incorrect) omitempty / pointer/non-pointer fields will add additional fields to marshalledOriginal. This could be e.g.:
1. scalar field with an empty value or
2. object fields with `{}`.

Now the resulting patches will in case:
1. generate a replace instead of an add
2. if a field should be added to this object it will just add the field and miss a patch to add the parent object

alculquicondor: I see your point, but both the existing and the proposed approaches are buggy.

Generally, the apiserver would have to go through the same unmarshal + marshal process, then the "raw" contents should not differ. A problem like the one you mention could arise when the webhook uses newer APIs than the API server has registered.

The problem I present arises when the API server has newer APIs than the webhook.

Not sure which can be more common.

sbueringer: For CRDs the apiserver doesn't even have the Go API types :)

alculquicondor: Can you please look at #2932?

alculquicondor:
/close

In favor of an approach closer to #2932

PR #2877, opened 2024-07-08
⚠ Remove deprecated Defaulter and Validator interfaces

troy0820:
We deprecated the `admission.Validator` and the `admission.Defaulter` interfaces and these are said to be removed in ~~release-0.19.~~ release-0.20  

 Closes #2641

- [x] Remove deprecated interfaces
- [x] ~~Remove tests for those types~~ Change tests for the deprecated types to the CustomDefaulter/CustomValidator
- [x] ~~Change~~ Remove log message that point to the now removed interfaces for the builder.
- [x] Moved type `Warnings` to `validator_custom.go`

Issue #2800, opened 2024-04-29
Optimize patches returned by Defaulter


sbueringer:
Today our Defaulter and CustomDefaulter implementations always compare the entire `req.Object.Raw` with the object mutated by the `Default` func.

The result is that we always return patches for metadata, spec and status if there are differences including cases where e.g. the status marhals to an empty object. For eample in the latter case the webhook returns a patch to add an empty status object.

We have the following cases where the Defaulter webhook is called:
* CRs without status subresource
    * [x] for the entire object => we actually have to return patches for the entire object (correct as of today) 
2. CRDs with status subresource
    * [ ] for spec & metadata => we should only return patches for spec & metadata (today we also return patches for status)
        * I'm not aware of any way to detect this at runtime, we could maybe introduce an option on the `WithDefaulter` func.
    * [ ] for status => we should only return patches for status (today we also return patches for spec & metadata)
        * We should be able to optimize this by dropping all non-status patches if `req.SubResource == status`

The reason why I'm bringing this up is because I've noticed that in a lot of cases our CAPI webhooks are returning patches for status fields that are never used (and eventually discarded by the apiserver). Would be better if we don't return those patches in the first place.

alvaroaleman:
Would't it be possible to solve this by constructing the Patch not with `req.Obj.Raw` as base but with the result of `h.decoder.Decode(req, obj)` before mutating it?

That would accurately capture the differences that the validator actually made, rather than potentially also including a bunch of unrelated stuff due to go type not perfectly matching with the configured CRD.

sbueringer:
> Would't it be possible to solve this by constructing the Patch not with req.Obj.Raw as base but with the result of h.decoder.Decode(req, obj) before mutating it?

I think this will lead to other problems. See the first two comments in this thread: https://github.com/kubernetes-sigs/controller-runtime/pull/2931#discussion_r1728364542

Basically the patches then won't be generated based on the actual JSON anymore, so then there are cases where the apiserver won't be able to apply the JSON patches.

trasc:
@sbueringer @alvaroaleman I've opened #2982 , in it, the approach is similar to the one in #2931, but we only make sure that the removals that are generated by the object decoding are not part of the final patch.

Additionally, we can make this behavior optional and maybe add some warnings in result indicating which fields wold have bean removed by the decoding itself.

PR #2932, opened 2024-08-23
✨ Allow to wire an object mutation handler

alculquicondor:
## Proposal

Allow wiring an `admission.Handler` to process mutations in a webhook, as part of the `builder.WebhookBuilder`.

## Motivation

Currently, mutations can be done through a `CustomDefaulter` which accepts and modifies a `runtime.Object`. The creation of patches is hidden from the developer. The jsonpatch is created from the difference of the request's raw yaml and the marshalled object modified by the `CustomDefaulter`.

This works for webhooks that are always released along with CRD changes. However, it doesn't work for webhooks written for CRDs from other projects. An example of this is Kueue, which writes webhooks for k8s Jobs, kubeflow, kuberay and others. Whenever there is a new version of the third party CRD, the hidden controller-runtime handler will produce a patch that removes all unknown fields.

A mechanism to fully override the handler allows developers to write patches that are safer against these version changes.

### Goals

- A mechanism to override jsonpatch generation for mutation webhooks.

### Non Goals

- A breaking-change in the libraries to build webhooks.
- A breaking-change in the way patches are built when using a CustomDefaulter.
- A mechanism to wire an `admission.Handler` to process validations. This could easily be added in a follow up proposal, but it doesn't seem to provide as much value as processing mutations.

## Alternatives

My first proposal was #2931, which is to round-trip the raw object to get rid of unknown fields, and generate a jsondiff from this object.
However, @sbueringer pointed out that the round-trip might also add fields that didn't exist in the raw object is the go types are missing omitempty, which might lead to patches that can't be applied to the raw object in the apiserver. So the suggestion was to give full control to developers over the handler to build patches as they see fit.

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -37,6 +37,7 @@ import (
 // WebhookBuilder builds a Webhook.
 type WebhookBuilder struct {
 	apiType         runtime.Object
+	mutatorFactory  admission.HandlerFactory
 	customDefaulter admission.CustomDefaulter
 	customValidator admission.CustomValidator
 	gvk             schema.GroupVersionKind
@@ -62,6 +63,12 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 		blder.err = errors.New("For(...) should only be called once, could not assign multiple objects for webhook registration")
 	}
 	blder.apiType = apiType
+	return blder
+}
+
+// WithMutatorFactory takes an admission.HandlerFactory, a MutatingWebhook will be wired for the handler that this factory creates.
+func (blder *WebhookBuilder) WithMutatorFactory(factory admission.HandlerFactory) *WebhookBuilder {
+	blder.mutatorFactory = factory
 	return blder
 }
 
@@ -169,14 +176,18 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
 }
 
 func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
-	if defaulter := blder.customDefaulter; defaulter != nil {
-		w := admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter)
-		if blder.recoverPanic != nil {
-			w = w.WithRecoverPanic(*blder.recoverPanic)
-		}
-		return w
+	var w *admission.Webhook
+	if factory := blder.mutatorFactory; factory != nil {
+		w = admission.WithHandlerFactory(blder.mgr.GetScheme(), blder.apiType, factory)
+	} else if defaulter := blder.customDefaulter; defaulter != nil {
+		w = admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter)
+	} else {
+		return nil
 	}
-	return nil
+	if blder.recoverPanic != nil {
+		w = w.WithRecoverPanic(*blder.recoverPanic)
+	}
+	return w
 }
 
 // registerValidatingWebhook registers a validating webhook if necessary.
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -37,7 +37,7 @@ import (
 // WebhookBuilder builds a Webhook.
 type WebhookBuilder struct {
 	apiType         runtime.Object
-	mutatorFactory  admission.HandlerFactory
+	mutationHandler admission.Handler
 	customDefaulter admission.CustomDefaulter
 	customValidator admission.CustomValidator
 	gvk             schema.GroupVersionKind
@@ -66,9 +66,9 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 	return blder
 }
 
-// WithMutatorFactory takes an admission.HandlerFactory, a MutatingWebhook will be wired for the handler that this factory creates.
-func (blder *WebhookBuilder) WithMutatorFactory(factory admission.HandlerFactory) *WebhookBuilder {
-	blder.mutatorFactory = factory
+// WithMutationHandler takes an admission.Handler, a MutatingWebhook will be wired for it.
+func (blder *WebhookBuilder) WithMutationHandler(h admission.Handler) *WebhookBuilder {
+	blder.mutationHandler = h
 	return blder
 }
 
@@ -147,7 +147,9 @@ func (blder *WebhookBuilder) registerWebhooks() error {
 	}
 
 	// Register webhook(s) for type
-	blder.registerDefaultingWebhook()
+	if err := blder.registerDefaultingWebhook(); err != nil {
+		return err
+	}
 	blder.registerValidatingWebhook()
 
 	err = blder.registerConversionWebhook()
@@ -158,8 +160,11 @@ func (blder *WebhookBuilder) registerWebhooks() error {
 }
 
 // registerDefaultingWebhook registers a defaulting webhook if necessary.
-func (blder *WebhookBuilder) registerDefaultingWebhook() {
-	mwh := blder.getDefaultingWebhook()
+func (blder *WebhookBuilder) registerDefaultingWebhook() error {
+	mwh, err := blder.getDefaultingWebhook()
+	if err != nil {
+		return err
+	}
 	if mwh != nil {
 		mwh.LogConstructor = blder.logConstructor
 		path := generateMutatePath(blder.gvk)
@@ -173,21 +178,27 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
 			blder.mgr.GetWebhookServer().Register(path, mwh)
 		}
 	}
+	return nil
 }
 
-func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
+func (blder *WebhookBuilder) getDefaultingWebhook() (*admission.Webhook, error) {
 	var w *admission.Webhook
-	if factory := blder.mutatorFactory; factory != nil {
-		w = admission.WithHandlerFactory(blder.mgr.GetScheme(), blder.apiType, factory)
-	} else if defaulter := blder.customDefaulter; defaulter != nil {
+	if handler := blder.mutationHandler; handler != nil {
+		w = &admission.Webhook{Handler: handler}
+	}
+	if defaulter := blder.customDefaulter; defaulter != nil {
+		if w != nil {
+			return nil, errors.New("a WebhookBuilder can only define a MutationHandler or a Defaulter, but not both")
+		}
 		w = admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter)
-	} else {
-		return nil
+	}
+	if w == nil {
+		return nil, nil
 	}
 	if blder.recoverPanic != nil {
 		w = w.WithRecoverPanic(*blder.recoverPanic)
 	}
-	return w
+	return w, nil
 }
 
 // registerValidatingWebhook registers a validating webhook if necessary.
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -67,6 +67,9 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 }
 
 // WithMutationHandler takes an admission.Handler, a MutatingWebhook will be wired for it.
+// A mutation handler is a low-level plug point for advanced use cases that need control
+// to generate patches. If in doubt, prefer using WithDefaulter to register a defaulter.
+// You can only use one of WithMutationHandler or WithDefaulter.
 func (blder *WebhookBuilder) WithMutationHandler(h admission.Handler) *WebhookBuilder {
 	blder.mutationHandler = h
 	return blder
--- a/pkg/builder/webhook_test.go
+++ b/pkg/builder/webhook_test.go
@@ -30,9 +30,12 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
+	"gomodules.xyz/jsonpatch/v2"
+	admissionv1 "k8s.io/api/admission/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/ptr"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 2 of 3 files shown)


pkg/builder/webhook.go
```diff
@@ -169,24 +176,24 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
...
-			w = w.WithRecoverPanic(*blder.recoverPanic)
-		}
-		return w
+	var w *admission.Webhook
+	if handler := blder.mutationHandler; handler != nil {
+		w = admission.WithObjectHandler(blder.mgr.GetScheme(), blder.apiType, handler)
+	} else if defaulter := blder.customDefaulter; defaulter != nil {
+		w = admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter)
+	} else if defaulter, ok := blder.apiType.(admission.Defaulter); ok {
+		w = admission.DefaultingWebhookFor(blder.mgr.GetScheme(), defaulter)
```
troy0820, line 185: I know you are not waiting on this [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2877) to merge but we are removing the Defaulter interface.  

Not that this has anything to do with the logic you present, just calling it out in case this changes how you approach this.

alculquicondor: It sounds like a good thing, but yeah, it doesn't affect this PR (other than the necessary rebase).

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -76,7 +88,14 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
 		return Errored(http.StatusBadRequest, err)
 	}
 
-	// Default the object
+	return h.handler.Handle(ctx, req.Object.Raw, obj)
```
sbueringer, line 91: If we're going to introduce another way to implement mutation, I wonder if we should go all the way:
* Should we just hand over ctx and the entire request? (~ l.83)
* Should we call this interface Mutator & the builder func WithMutator

I think then it would also work well for resource updates

sbueringer: @vincepri @alvaroaleman Opinions?

pkg/builder/webhook.go
sbueringer: #2877 has been merged node

pkg/webhook/admission/defaulter_custom.go
alculquicondor: I wanted to avoid doing `h.decoder.Decode(req, obj)` every time, but sure, better leave everything up to the developer.

alculquicondor:
The panic seems to be on a different suite

/retest

alculquicondor:
@sbueringer I addressed all the comments, PTAL.

pkg/webhook/admission/defaulter_custom.go
alvaroaleman: Sorry, I don't quite follow how the new thing helps with the problem at hand. The problem at hand is you need the raw request or not? Wouldn't it be better to change the existing `CustomDefaulter` interface to add it as an arg in there?

alculquicondor: Yes, I do need the raw request. But this is not the biggest problem, because the request is embedded in the context.

But the reason why I need the raw request is to be able to return json patches. And the CustomDefaulter currently just expects changes to the object. The creation of patches is completely hidden from the defaulter.

We could change the interface to return patches, but that's a breaking change. Isn't it better to add a fully manual mechanism to building webhook handlers?

sbueringer: My thinking was that we can lose information during json.Unmarshal and json.Marshal (either if the CRDs are not prefectly designed with omitempty, or if we have to wrong version of the CRD because it's not ours)

So I thought if we provide a low-level way to handle this better, let's go all the way and give folks all the info (req) and allow them to return the patches directly

sbueringer: Somewhat related issue https://github.com/kubernetes-sigs/controller-runtime/issues/2800

alculquicondor:
Can we move this forward? In Kueue, we are risking dropping fields all the time because we don't control the versions of the CRDs that we have webhooks for.

alvaroaleman:
> Can we move this forward? In Kueue, we are risking dropping fields all the time because we don't control the versions of the CRDs that we have webhooks for.

I understand, but c-r is a widely used project and this is an external interface change. Please have some patience until we find time to properly review it.

sbueringer:
If you are under pressure to implement a fix now in Kueue, I would highly recommend to implement an alternative on Kueue side. Otherwise we need some time until folks have time to review this PR.

None of us is paid to maintain controller-runtime. So we are working on CR when we find the time

vincepri:
Having read through this discussion, the best path forward imo would be a small proposal that outlines the goals/non-goals and the problem we're trying to solve for all Controller Runtime users, not just this specific use case or what the current code does.

alculquicondor:
See updated description.

pkg/builder/webhook_test.go
```diff
@@ -735,6 +811,12 @@ func (*TestCustomDefaulter) Default(ctx context.Context, obj runtime.Object) err
 
 var _ admission.CustomDefaulter = &TestCustomDefaulter{}
 
+func mutatorFactoryForTestDefaulter(scheme *runtime.Scheme) admission.HandlerFactory {
+	return func(obj runtime.Object, _ admission.Decoder) admission.Handler {
+		return admission.WithCustomDefaulter(scheme, obj, &TestCustomDefaulter{}).Handler
+	}
+}
```
vincepri, line 818: This is very confusing. We're using a mutator, but returning a custom defaulter handler?

pkg/builder/webhook.go
```diff
@@ -65,6 +66,12 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 	return blder
 }
 
+// WithMutatorFactory takes an admission.HandlerFactory, a MutatingWebhook will be wired for the handler that this factory creates.
+func (blder *WebhookBuilder) WithMutatorFactory(factory admission.HandlerFactory) *WebhookBuilder {
+	blder.mutatorFactory = factory
+	return blder
+}
```
vincepri, line 73: Adding this makes the `WithDefaulter` a noop, which doesn't return an error and it's a misleading UX

pkg/webhook/admission/webhook.go
```diff
@@ -114,6 +118,13 @@ func (f HandlerFunc) Handle(ctx context.Context, req Request) Response {
 	return f(ctx, req)
 }
 
+// WithHandlerFactory creates a new Webhook for a handler factory.
+func WithHandlerFactory(scheme *runtime.Scheme, obj runtime.Object, factory HandlerFactory) *Webhook {
```
vincepri, line 122: Why is this called Factory?

pkg/builder/webhook.go
```diff
@@ -169,14 +176,18 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() {
...
-		}
-		return w
+	var w *admission.Webhook
+	if factory := blder.mutatorFactory; factory != nil {
+		w = admission.WithHandlerFactory(blder.mgr.GetScheme(), blder.apiType, factory)
+	} else if defaulter := blder.customDefaulter; defaulter != nil {
+		w = admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter)
+	} else {
+		return nil
 	}
```
vincepri, line 186: Same here, what happens if both are defined?

vincepri (commented):
@alculquicondor Thanks for the description and extra context. The problem statement makes sense to me, although the proposed solution bring in a lot of questions for most users of this library.

The provided goal `A mechanism to wire an admission.Handler to process mutations` is an implementation detail, rather an an actual goal. Also to note, our [design proposals](https://github.com/kubernetes-sigs/controller-runtime/tree/main/designs) are collected in the root repository and discussed separately from the actual implementation.

pkg/webhook/admission/webhook.go
alculquicondor: Because it knows how to create a handler. It allows you to receive a decoder from the WebhookBuilder.

Another option could be just to accept a Handler (renaming the method to `WithMutationHandler`) and pass the decoder through the context.

alculquicondor:
> Also to note, our [design proposals](https://github.com/kubernetes-sigs/controller-runtime/tree/main/designs) are collected in the root repository and discussed separately from the actual implementation.

Up to you if you prefer me to write it there. This change seems so minimal that it shouldn't need such a permanent location for the design.

pkg/webhook/admission/webhook.go
alculquicondor: I decided to remove the factory and leave it to the implementation to decide how to build the object.

pkg/builder/webhook.go
alculquicondor: Since this function doesn't return an error, I'll leave it to the `Complete` function to report the error. Unless you are ok with a panic?

pkg/builder/webhook.go
alculquicondor: Added error handling

pkg/builder/webhook_test.go
alculquicondor: I moved this somewhere else, but I'm still using a CustomDefaulter. Let me know if you prefer adding a different handler.

alculquicondor:
There seems to be a difference in how the logger is set up. I'm investigating.

pkg/builder/webhook_test.go
alculquicondor: I added a custom handler.

alculquicondor:
/hold cancel

This is ready for review now

pkg/builder/webhook_test.go
```diff
@@ -570,7 +649,8 @@ func (*TestDefaultValidatorList) DeepCopyObject() runtime.Object   { return nil
 type TestCustomDefaulter struct{}
 
 func (*TestCustomDefaulter) Default(ctx context.Context, obj runtime.Object) error {
-	logf.FromContext(ctx).Info("Defaulting object")
+	log := logf.FromContext(ctx)
```
mimowo, line 652: non-blocking nit: I would suggest to revert the formatting

mimowo (commented):
LGTM

pkg/builder/webhook_test.go
alculquicondor: Oops, that was me trying to debug some logging issues. Reverted.

pkg/builder/webhook.go
```diff
@@ -65,6 +66,12 @@ func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
 	return blder
 }
 
+// WithMutationHandler takes an admission.Handler, a MutatingWebhook will be wired for it.
+func (blder *WebhookBuilder) WithMutationHandler(h admission.Handler) *WebhookBuilder {
```
alvaroaleman, line 70: Its super not obvious when and why to use this versus the defaulter and validator and almost guaranteed to lead to questions

alvaroaleman (commented):
So if I understand it correctly, the root of the issue is a mismatch between the go types in the project and what is configured as the CRD, resulting in the patch removing fields it doesn't know about. 

Can you use `unstructured.Unstructured` in your CustomDefaulter to avoid this?

pkg/webhook/admission/defaulter_custom.go
alvaroaleman: Would this work: https://github.com/kubernetes-sigs/controller-runtime/issues/2800#issuecomment-2381413954 ?

alvaroaleman:
How about https://github.com/kubernetes-sigs/controller-runtime/issues/2800#issuecomment-2381413954 instead?

pkg/builder/webhook.go
alculquicondor: Can we leave it as "more advanced use cases where you want control over the construction of patches"?

alculquicondor: Or we could explicitly add the use case of go types that might be in a different version than the CRD.

pkg/webhook/admission/defaulter_custom.go
sbueringer: Answered on the issue, nope

sbueringer:
/lgtm
in general

Not entirely sure if we want a bit godoc around when to use which mechanism (WithMutationHandler vs WithDefaulter)
(cc @vincepri @alvaroaleman)

alvaroaleman:
> @sbueringer thank you for reviewing as we are going to use it for Kueue webhooks to register the LossLessDefaulter ( [docs](https://github.com/kubernetes-sigs/kueue/blob/04bfc06479c4b755f3419fa3568f135abc3664e5/site/content/en/docs/tasks/dev/integrate_a_custom_job.md?plain=1#L78C7-L92), and [code](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/webhook/defaulter.go)). We are also working on improving it to be https://github.com/kubernetes-sigs/kueue/pull/3186.

This seems extreme. The tl,dr of the problem still seems to be that we don't know of a good solution, so instead we add a low-level plug point so people can work around the problem. If we were to add something like what you've built in Kueue we effectively bless it and call it a good/appropriate solution and I don't think that is what it is.

pkg/builder/webhook.go
alvaroaleman: Please write something along the lines of that this is a low-level plug point for advanced users that need more control over how exactly the patch is being constructed and if in doubt, people should always use `WithDefaulter` instead.

alculquicondor:
@alvaroaleman, I think the proposal in https://github.com/kubernetes-sigs/kueue/pull/3194 is actually generic and it doesn't seem to have any of the problems in earlier proposals.

It just removes any `remove` operations that refer to fields not present in the go type.

sbueringer:
Maybe it would be better to first try to get the current PR merged and then we can discuss on a new issue if we want to add some built-in handlers?

Except if it's all or nothing from your side

mimowo:
Yeah. that's for sure, we should first move it to the point where we can merge it into Kueue. I'm also perfectly ok to incubate it for some time (like a release or more) in Kueue, so that we detect potential issues. 

Once it is merged into Kueue and we feel comfortable about it I would like to follow up with an issue in controller-runtime to discuss the transfer and potential gaps. WDYT?

alculquicondor:
Yes please, let's get this PR merged, as there will always be cases where people need to have fine grained control over patches.

The current way of doing that requires a lot of duplicated code from controller-runtime.

sbueringer:
> Once it is merged into Kueue and we feel comfortable about it I would like to follow up with an issue in controller-runtime to discuss the transfer and potential gaps. WDYT?

No guarantees about the outcome, but sure we can absolutely open an issue and discuss

pkg/builder/webhook.go
alculquicondor: Added

alculquicondor:
@alvaroaleman any other comment or ask before approval?

sbueringer:
/close

Merged #2982 instead

PR #2935, opened 2024-08-29
:seedling: Enable using errors.Is with ErrResourceDiscoveryFailed error

tamalsaha:

### DIFF
--- a/pkg/client/apiutil/errors.go
+++ b/pkg/client/apiutil/errors.go
@@ -52,3 +52,9 @@ func (e *ErrResourceDiscoveryFailed) Unwrap() []error {
 	}
 	return subErrors
 }
+
+// Is makes it possible for the callers to use `errors.Is(` helper on errors wrapped with ErrResourceDiscoveryFailed error.
+func (e *ErrResourceDiscoveryFailed) Is(target error) bool {
+	_, ok := target.(*ErrResourceDiscoveryFailed)
+	return ok
+}


pkg/client/apiutil/errors.go
```diff
@@ -52,3 +52,9 @@ func (e *ErrResourceDiscoveryFailed) Unwrap() []error {
 	}
 	return subErrors
 }
+
+// Is makes it possible for the callers to use `errors.Is(` helper on errors wrapped with ErrResourceDiscoveryFailed error.
+func (e *ErrResourceDiscoveryFailed) Is(target error) bool {
+	_, ok := target.(*ErrResourceDiscoveryFailed)
+	return ok
```
troy0820, line 59:
```suggestion
       ep := &ErrResourceDiscoveryFailed{}
       return errors.As(target,ep)
```

We have done this for Terminal error in `pkg/reconcile/reconcile.go:179`
https://github.com/kubernetes-sigs/controller-runtime/blob/e6c3d139d2b6c286b1dbba6b6a95919159cfe655/pkg/reconcile/reconcile.go#L179-L182

I am not sure if this is needed but I would add a test if you are adding this change.

tamalsaha: Screenshot 2024-08-29 at 1 06 35 PM

I don't think `Is` should be implemented using `As`, since `As` calls unwrap.

alvaroaleman: It doesn't matter, if the `Is` implementation doesn't call unwrap, `errors.Is` does, so the result is the same. Please add a test though.

sbueringer:
@tamalsaha Do you have time to address the finding above?

tamalsaha:
Sorry, I won't have time.

PR #2936, opened 2024-09-02
✨Expose RegisterClientMetrics to register other clien…

Yesphet:
…t-go metrics

Users of controller-runtime can not register custom metrics of `client-go` since `k8s.io/client-go/tools/metrics.Register` can only be called once and it's called in the `init()` function of  `pkg/metrics`. So this PR expose `RegisterClientMetrics` for registering custom `client-go` metrics. If it is not called within the first 30s of startup, only the client latency metrics will be registered.

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,6 +18,8 @@ package metrics
 
 import (
 	"context"
+	"sync"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	clientmetrics "k8s.io/client-go/tools/metrics"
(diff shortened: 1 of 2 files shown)


Yesphet:
@alvaroaleman @inteon  Hi, I hope you’re doing well! Just a quick reminder to review this pr (#2936) submitted two weeks ago. Your feedback would be greatly appreciated!

pkg/metrics/client_go_adapter.go
```diff
@@ -37,20 +39,30 @@ var (
...
-
-// registerClientMetrics sets up the client latency metrics from client-go.
-func registerClientMetrics() {
 	// register the metrics with our registry
 	Registry.MustRegister(requestResult)
 
-	// register the metrics with client-go
-	clientmetrics.Register(clientmetrics.RegisterOpts{
-		RequestResult: &resultAdapter{metric: requestResult},
+	time.AfterFunc(30*time.Second, func() {
```
alvaroaleman, line 50: I am not a fan of this because it means you might not get metrics for the first thirty seconds of the binary. Another way I can think of is to maybe call this before first use in the controller or so, that would still allow to register them yourselves by doing it before you build your controllers and avoid this issue?

pkg/metrics/client_go_adapter.go
```diff
@@ -63,9 +75,13 @@ func registerClientMetrics() {
...
-	metric *prometheus.CounterVec
+	metric       *prometheus.CounterVec
+	customMetric clientmetrics.ResultMetric
 }
 
-func (r *resultAdapter) Increment(_ context.Context, code, method, host string) {
+func (r *resultAdapter) Increment(ctx context.Context, code, method, host string) {
 	r.metric.WithLabelValues(code, method, host).Inc()
+	if r.customMetric != nil {
+		r.customMetric.Increment(ctx, code, method, host)
```
alvaroaleman, line 85: Sorry, I do not understand this. You both want to register them yourself and plug your own functionality into the existing adapter? Why isn't one of the two enough?

pkg/metrics/client_go_adapter.go
Yesphet: Agreed, I moved this call into the NewServer function of metric server in 639312c0e24c606365d41c98d95d163913a209fc

pkg/metrics/client_go_adapter.go
Yesphet: Initially, the goal was to support custom requestResult metrics and keep the original rest_client_requests_total metrics. I changed the resultAdapter to the default value of `RegisterOpts.RequestResult` in 639312c0e24c606365d41c98d95d163913a209fc

PR #2942, opened 2024-09-13
📖  Update incorrect method name on doc.go

buraksekili:
This PR addresses a discrepancy in method naming when constructing a fake client.

The documentation at https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake#pkg-overview currently references a `WithObj` method. However, it appears this may be a typo, and the correct method name should be `WithObjects` since there is no WithObj method

This PR fixes this.


sbueringer:
Thx!

/lgtm
/approve
/ok-to-test

PR #2943, opened 2024-09-14
set gvk while getting object using typed-client

HiranmoyChowdhury:

While getting or creating or patching or updating object using typed-client which is a dynamic client in controller-runtime repository, gvk found empty for that object because of the codec convention. And this is the minimal coding change and might be the optimal solution I found.

#### Which issue(s) this PR fixes:

Note: the PR need to get merged first: https://github.com/kubernetes/kubernetes/pull/127361
Then I will be able to update the deps.


alvaroaleman (changes requested):
This is by design. The GVK fields can never be relied upon to be set in code, if you do for example an `pod := &myPod{}` it won't be set. Use `apiutil.GVKForObject` if you need to find the gvk for an object.

HiranmoyChowdhury:
@alvaroaleman yeah, we can do so. But, when we are calling an object using dynamic client we expect that it set this object with gvk or typemeta. Someone already fixed that issue for unstructured client and metadata client.But, this is currently missing in object when we are calling using typed-client.

alvaroaleman:
In the unstructured and objectmeta case it has to be set because there is no other way to derive it and all client operations on those will fail if it's unset. In the typed case it can be derived from the go types, which is why we deliberately make sure it is unset, so no one comes to the conclusion they can rely on it being set.

HiranmoyChowdhury:
@alvaroaleman Thanks for describing. And this is true that we can derive GVK from structured object. But, I guess I have a little confusion on this topic. As unstructured client and metadata client and typed client is not exported so, we can't use this from outside and we are using only one client for these. So, when we try to get unstructured object, we get the GVK. But, for structured object we missed GVK with the same client(from users perspective), this is a little confusion for newcomers like me. I had to find this by a lot of debugging. Also, there is a option to determine what the GVK is: obj.GetObjectKind().GroupVersionKind(). As there is a option I assumed this is fixed and set and we don't need to worry about this. I guess this should be set as expected.
Also, while using structured object, if I use "apiutil.GVKForObject" this increase coding cost I guess and if you see this PR solution which takes a lesser amount of cost and also a time saving effort. 

I am new in this field and have a lack amount of design sense and knowledge. I have to learn a lot to do so. I will be pleased if you can explain those things to me.

alvaroaleman:
> Also, there is a option to determine what the GVK is: obj.GetObjectKind().GroupVersionKind()

That might or might not work. It will not work for example if you did a `pod := &corev1.MyPod{}`. Always use `apiutil.GVKForObject` if you need to get the gvk for an object, regardless of unstructured/objectmeta/typed.

This behavior is by design, we did changes in the past to avoid setting it where it can be avoided: https://github.com/kubernetes-sigs/controller-runtime/pull/2633

HiranmoyChowdhury:
got this now. thanks for your time

PR #2946, opened 2024-09-16
Set GVK in an object while calling typed-client

HiranmoyChowdhury:

While getting or creating or patching or updating object using typed-client which is a dynamic client in controller-runtime repository, gvk/TypeMeta found empty for that object because of the codec convention. 
As unstructured client and metadata client and typed client is not exported. So, we can't call these from outside and we are using only one client for these. So, when we try to get unstructured object, we get the GVK. But, for structured object we missed GVK with the same client(from users perspective), this is a little confusion for newcomers like me. I had to find this by a lot of debugging. Also, there is an option to determine what the GVK is: obj.GetObjectKind().GroupVersionKind(). As there is a option I assumed this is fixed and set and we don't need to worry about this. I guess this should be set as expected.
To create owner reference or many more we won't need to pass scheme or gvk externally.

issue: 

### DIFF
--- a/pkg/client/typed_client.go
+++ b/pkg/client/typed_client.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	"context"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	"k8s.io/apimachinery/pkg/runtime"
 )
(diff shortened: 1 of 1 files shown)


troy0820:
/hold

You have another [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2943) that addresses this same issues with feedback from @alvaroaleman that is put on hold.  

I believe the intent of the issue was said to be by design.  @sbueringer  thoughts?

alvaroaleman:
Lets please keep this discussion in one place at the first PR, https://github.com/kubernetes-sigs/controller-runtime/pull/2943

PR #2953, opened 2024-09-18
🌱 Do not deduplicate warnings by default

dlipovetsky:
Controllers are long-running processes, and deduplication, as implemented, increases memory use.

With this change, duplicate warnings will appear in the log by default. However, this is safe, because Kubernetes rotates container logs by default.

If a specific controller sees many duplicate warnings, it can configure the handler to deduplicate them.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -113,11 +113,11 @@ func newClient(config *rest.Config, options Options) (*client, error) {
 	}
 
 	if config.WarningHandler == nil {
-		// By default, we de-duplicate and surface warnings.
+		// By default, we surface warnings.
 		config.WarningHandler = log.NewKubeAPIWarningLogger(
 			log.Log.WithName("KubeAPIWarningLogger"),
 			log.KubeAPIWarningLoggerOptions{
-				Deduplicate: true,
+				Deduplicate: false,
 			},
 		)
 	}


alvaroaleman:
> The current default is a memory leak

It takes some memory but how is it a leak? The number of unique warnings should be pretty finite or not?

sbueringer:
> > The current default is a memory leak
> 
> It takes some memory but how is it a leak? The number of unique warnings should be pretty finite or not?

Good point. Is the number of unique warnings finite? (i.e. is there a guarantee that they never contain variable parts like object names?)

alvaroaleman:
> I mean I wonder in general who notices these warnings in controller logs, but only logging them once doesn't make it better :)

Yeah I am a bit thorn on this. On the one hand if this was net new I would keep them on because as you said, its easy to miss them if they are only logged once. On the other you can make the argument that this is a silently breaking change and ppl might complain about now getting too many warnings that they don't care about logged.

sbueringer:
Yup. We can also consider defaulting to no warning handler. Then people can add it if they want

dlipovetsky:
This came out a discussion in a cluster-api PR: https://github.com/kubernetes-sigs/cluster-api/pull/11179#discussion_r1763420636

I think deduplication can make it harder to spot warnings. Thisi what I said there:

> If the client makes two identical API calls hours apart, and both return the same warning, I would expect to find both in the log.
>
> Even worse is that, with the current implementation, if the warning is logged, and that part of the log is overwritten, then evidence of the warning would disappear, because the warning would never be logged again, for as long as the process runs.

dlipovetsky:
> Good point. Is the number of unique warnings finite? (i.e. is there a guarantee that they never contain variable parts like object names?)

The "message" is the deduplication key here. And the message does include the object name, at least in the warnings I have seen. For example: https://github.com/kubernetes/kubernetes/blob/432e8937cfb1965f863b8f841af61a8cad7db965/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L89

alvaroaleman:
> https://github.com/kubernetes/kubernetes/blob/432e8937cfb1965f863b8f841af61a8cad7db965/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L89

From what I can see there it only includes the fieldpath and the value but not object name or am I missing it?

dlipovetsky:
> > https://github.com/kubernetes/kubernetes/blob/432e8937cfb1965f863b8f841af61a8cad7db965/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L89
> 
> From what I can see there it only includes the fieldpath and the value but not object name or am I missing it?

My mistake! You're right, the object name is not part of the message.

dlipovetsky:
To recap, a warning looks like this:
```
[KubeAPIWarningLogger] metadata.finalizers: "cluster.cluster.x-k8s.io": prefer a domain-qualified finalizer name to avoid accidental conflicts with other finalizer writers
```
In this warning, the field-path is `metadata.finalizers`, and `"cluster.cluster.x-k8s.io"` is the group-kind (no version).

Currently, warnings are de-duplicated by default, so only  onewarning per field-path/group-kind combination shows up in the log for the lifetime of the manager process.

Reasons to stop de-duplicating by default:
- Easy to miss warnings if they are logged only once.

Reasons to continue de-duplication by default:
- Fewer warnings in the log, i.e. a less "noisy" log.

If there's no clear winner, then we can keep the existing behavior, and I'll close the PR.

dlipovetsky:
Sounds good.

We can try to decrease the impact of the change. Some ideas:

1. Write a user-friendly release note that explains how to enable de-duplication when creating the client.
2. Announce the change in the notes of the next minor release, but only merge the change afterward, so that the default changes in two minor releases from now.

sbueringer:
I would be fine with only 1. I think it's not a problematic change to disable the deduplication. But let's see what @vincepri and @alvaroaleman think

alvaroaleman:
> But let's see what @vincepri and @alvaroaleman think.

Yeah, I am ok with changing it, too. Lets maybe mark this change as breaking since it changes behavior so that it shows up more prominently on the release notes, but other than that, I am fine with this.

alvaroaleman (approved):

PR #2958, opened 2024-09-23
🐛 Clean up and namespace verification while removing owner ref and controller ref

HiranmoyChowdhury:

As we know for a resource there only one name can be allocated for every individual namespace. So, when we are removing ownerRef or controllerRef from an controlled object by passing the owner object, we are only checking the name+group+kind but not the namespace. To verify this I added namespace verification and also did some cleanup.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -82,7 +82,6 @@ func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Sch
 	}
 
 	ref := metav1.NewControllerRef(owner, gvk)
-
 	for _, opt := range opts {
 		opt(ref)
 	}
@@ -138,7 +137,6 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
 	if !ok {
 		return fmt.Errorf("%T is not a runtime.Object, cannot call RemoveOwnerReference", owner)
 	}
-
 	if err := validateOwnerWithNS(owner, object); err != nil {
 		return err
 	}
@@ -189,7 +187,7 @@ func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Sche
 
 	if controller := metav1.GetControllerOfNoCopy(object); controller == nil {
 		return fmt.Errorf("%T does not have a owner reference with controller equals true", object)
-	} else if !referSameObject(*controller, *NewOwnerRef(owner, gvk)) {
+	} else if !referSameObject(*controller, *metav1.NewControllerRef(owner, gvk)) {
 		return fmt.Errorf("%T owner is not the controller reference for %T", owner, object)
 	}
 
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -19,6 +19,7 @@ package controllerutil
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/ptr"
 	"reflect"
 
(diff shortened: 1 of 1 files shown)


HiranmoyChowdhury:
As there can be only one Controller Reference, then we should have something which only removes the controllerReference. Here the function named "RemoveControllerReference" removes controller reference if that matched with the given object. This way should only applicable for OwnerReference I guess. Should I change this too or make a new function?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -266,16 +238,7 @@ func validateOwner(owner, object metav1.Object) error {
...
-	if err != nil {
-		return false
-	}
-
-	bGV, err := schema.ParseGroupVersion(b.APIVersion)
-	if err != nil {
-		return false
-	}
-	return aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name
+	return a.UID == b.UID
```
vincepri, line 241: /hold

No, this isn't correct as UIDs can change across backups and clusters. References should be compared like above

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -188,20 +165,17 @@ func HasOwnerReference(ownerRefs []metav1.OwnerReference, obj client.Object, sch
...
+	idx := indexOwnerRef(ownerRefs, *NewOwnerRef(obj, gvk))
 	return idx != -1, nil
 }
 
 // RemoveControllerReference removes an owner reference where the controller
 // equals true
 func RemoveControllerReference(owner, object metav1.Object, scheme *runtime.Scheme) error {
-	if ok := HasControllerReference(object); !ok {
-		return fmt.Errorf("%T does not have a owner reference with controller equals true", object)
+	if !metav1.IsControlledBy(object, owner) {
```
troy0820, line 175: This error is now misleading because if the object doesn’t have a controller reference the returned error will only indicate that the owner being passed is not the controller reference for the object. 

The error before allowed you to recognize if the object you are passing in had one at all before trying to remove it with the owner passed in.

troy0820: https://github.com/kubernetes-sigs/controller-runtime/pull/2595#discussion_r1404608398 Discussion for why the logic exists was from several issues with how to get around overwriting and errors that exist with controller reference/owner reference.

pkg/controller/controllerutil/controllerutil.go
HiranmoyChowdhury: Got it. I am gonna change this and I guess rest are okay

pkg/controller/controllerutil/controllerutil.go
HiranmoyChowdhury: okay. gonna change it soon

pkg/controller/controllerutil/controllerutil.go
HiranmoyChowdhury: And going to ensure if they are from the same namespace.

pkg/controller/controllerutil/controllerutil.go
HiranmoyChowdhury: @troy0820 WDYT about this comment: https://github.com/kubernetes-sigs/controller-runtime/pull/2958#issuecomment-2368877244

pkg/controller/controllerutil/controllerutil.go
HiranmoyChowdhury: Now this will ensure owner by validating namespace while removing owner ref or controller ref

PR #2950, opened 2024-09-17
:bug: upgrade golangci-lint to v1.61.0

sivchari:
Upgade the golangci-lint version to v1.61.0.

PR #2971, opened 2024-10-09
✨ Bump to k8s.io/* v0.32.0-alpha.1

sbueringer:


sbueringer:
I'll rebase once https://github.com/kubernetes-sigs/controller-runtime/pull/2950 is merged

Otherwise the current PR should be ready

hack/check-everything.sh
```diff
@@ -24,6 +24,8 @@ source ${hack_dir}/common.sh
 tmp_root=/tmp
 kb_root_dir=$tmp_root/kubebuilder
 
+export GOTOOLCHAIN="go$(make go-version)"
```
sbueringer, line 27: @alvaroaleman @vincepri WDYT about this one? Seems like this could make us independent of the go version built into the CI image

alvaroaleman: TIL about this. What I don't fully understand is how this interacts with the go version in the go.mod file? It would be good if we didn't override anything specified in the go.mod file, otherwise our CI may not be reflective of what actual end users see.

sbueringer: Hm yeah this whole go toolchain feature is a bit complicated: https://go.dev/doc/toolchain

As far as I'm aware it doesn't overwrite what is in the go.mod file.

If I don't set this here and without changing the CI image we end up with this: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/2971/pull-controller-runtime-apidiff/1844073887495622656

> /home/prow/go/src/sigs.k8s.io/controller-runtime/hack/tools/bin/go-apidiff-v0.8.2 4cae9dfbf174fa5d49ebca007bd6f3784cf1ffb3 --print-compatible
> failed to get packages from new commit "HEAD" (79c1f765939e53ddc94401070922c37d797bfef6): err: exit status 1: stderr: go: go.mod requires go >= 1.23.0 (running go 1.22.8; GOTOOLCHAIN=local)

But I can also just keep it simple and bump the CI image to one that uses Go 1.23

alvaroaleman (approved):

Issue #2960, opened 2024-09-25
The fake client does not support concurrent patching of ConfigMap resources


akutz:
I was working on the PR https://github.com/vmware-tanzu/vm-operator/pull/673 when a [test](https://github.com/vmware-tanzu/vm-operator/pull/673/files#diff-fa7c32108046a841970ed96b286206b0692409eb534d8683836b77b83c1050a3R127-R198) (`there are concurrent attempts to update the ConfigMap` in `pkg/util/kube/storage_test.go`)  that I _thought_ should have worked, unexpectedly failed with a 409 (conflict):

```shell
$ ginkgo --json-report ./ginkgo.report -focus "EncryptedStorageClass when there are concurrent attempts to update the ConfigMap"   -r
[1727280290] Kube Util Test Suite - 1/21 specs SSSS
------------------------------
• [FAILED] [0.003 seconds]
EncryptedStorageClass when there are concurrent attempts to update the ConfigMap [It] more than one patch should succeed
/Users/akutz/Projects/vmop/vmop/pkg/util/kube/storage_test.go:130

  [FAILED] StorageClass.name=0
  Expected success, but got an error:
      <*errors.StatusError | 0x140003c0320>: 
      Operation cannot be fulfilled on configmaps "encrypted-storage-class-names": object was modified
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Operation cannot be fulfilled on configmaps \"encrypted-storage-class-names\": object was modified",
              Reason: "Conflict",
              Details: {
                  Name: "encrypted-storage-class-names",
                  Group: "",
                  Kind: "configmaps",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  In [It] at: /Users/akutz/Projects/vmop/vmop/pkg/util/kube/storage_test.go:151 @ 09/25/24 11:04:52.802
------------------------------
SSSSSSSSSSSSSSSS

Summarizing 1 Failure:
  [FAIL] EncryptedStorageClass when there are concurrent attempts to update the ConfigMap [It] more than one patch should succeed
  /Users/akutz/Projects/vmop/vmop/pkg/util/kube/storage_test.go:151

[...]

PR #2979, opened 2024-10-12
:bug: Fakeclient: Allow concurrent patching if RV is unset or null

alvaroaleman:
Currently, the fake client always errors if patches are done concurrently rather than only if the patch contains a ResourceVersion.

This is because we have a number of checks including the one related to resourceVersion implemented in a `versionedTracker`. The `versionedTracker` receives the patched object and assumes that the patchedObject only contains a RV if the patch had one. That turns out not to be true, it almost always has one.

The reason the object we receive in the `versionedTracker` almost always has a RV is that we use a client-go reactor to apply the patch to an object. The way this works is that the reactor takes the patch and a tracker, fetches the object from the tracker and applies the patch to it. This means that the resulting object always has a resourceVersion unless the patch explicitly set it to `null`. This `null` case apparently is special cased in the Kubernetes apiserver to be acceptable, so we do the same here.

Fix the original issue by checking in the fakeclient if the patch modifies the RV and if not, retry conflicts.

The goal of this patch is to address https://github.com/kubernetes-sigs/controller-runtime/issues/2960 without introducing any other change in behavior. It does explicitly not address https://github.com/kubernetes-sigs/controller-runtime/issues/2977 as that might break some peoples test.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -468,6 +468,11 @@ func (t versionedTracker) updateObject(gvr schema.GroupVersionResource, obj runt
 		switch {
 		case allowsUnconditionalUpdate(gvk):
 			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+			// This is needed because if the patch explicitly sets the RV to null, the client-go reaction we use
+			// to apply it and whose output we process here will have it unset. It is not clear why the Kubernetes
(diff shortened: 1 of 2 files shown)


alvaroaleman:
Closing this, I think https://github.com/kubernetes-sigs/controller-runtime/pull/2980 is the better approach. This change fixes the issue but introduces another: We allow dirty writes of the ResoruceVersion. I think just backporting #2980 is better.

PR #2980, opened 2024-10-12
:bug: Fakeclient: Fix TOCTOU races

alvaroaleman:
The fake client currently has a number of time of check time of use races, where it fetches an object to determine what to do in a mutating operation. The problem is that the object might change in between fetching it and doing the mutating operation. Most notably, this happens when:
* Patching is done in parallel. Only one of the patches will succeed, the other ones will fail with a conflict
* Updates of objects that allow unconditional updates: All updates will succeed, but not all of them will increment the resource version (i.E dirty writes for the RV)
* An update for an object that allows createOnUpdate races with a create or delete
* A DeleteAllOf call races with Delete calls
* A scale update races with a normal update

This change:
* Adds tests for all of these cases
* Fixes them by adding a lock around the write operations, including their read part, if any

Note: This patch is a more complete alternate to https://github.com/kubernetes-sigs/controller-runtime/pull/2979. Initially I thought that this change might be too breaking to backport, because tests tend to converge extremely closely to any behavior of the fakeclient regardless of it being correct or not, but I _think_ it should be fine. The fact that it took years for someone to notice a very basic issue where concurrent patching doesn't work correctly is also indicative of the fact that the fake client seems to be rarely used concurrently and only then will this change result in any change of behavior.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2977
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2960

/cc @akutz

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -69,16 +69,21 @@ type versionedTracker struct {
 }
 
 type fakeClient struct {
-	tracker               versionedTracker
-	scheme                *runtime.Scheme
+	// trackerWriteLock must be acquired before writing to
+	// the tracker or performing reads that affect a following
+	// write.
+	trackerWriteLock sync.Mutex
+	tracker          versionedTracker
+
+	schemeWriteLock sync.Mutex
+	scheme          *runtime.Scheme
+
 	restMapper            meta.RESTMapper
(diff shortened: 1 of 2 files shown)


sbueringer:
Thank you very much!!

/lgtm
/approve

PR #2981, opened 2024-10-14
✨  Fakeclient: Add apply support

### LINKED ISSUE #2341: Support server-side apply (client.Apply) in fake Client
Recently, we've been writing extensive unit test coverage of our controllers using fake.Client but a few of our controllers use client.Apply, which is not supported by fake.Client.

~We can't migrate these controllers from client.Apply to client.Merge because with upgrades, any CRs managed by the previous version will have owned fields and the API server will reject the request (my understanding, is this true?)~ based on [the SSA docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/#apply-and-update) it seems the server would only reject requests based on ownership when using SSA, so using update would work.

Envtest takes ~5-15s to set up for each test case, meaning a test suite which would take 0.05s with fake.Client, takes 5 minutes with envtest. In many cases, we can isolate tests to namespaces and reuse the same client, but not always. We prefer our unit tests to be as isolated as possible.

Would it be possible to support server-side apply / client.Apply / SSA in the fake client?

alvaroaleman:
This change adds apply support into the fake client.

This relies on the upstream support for this which is implemented in a new [FieldManagedObjectTracker][0]. In order to support many types, a custom `multiTypeConverter` is added.

The `FieldManagedObjectTracker` results in `ManagedFields` being set after any operations. As that would be a very breaking change, the fake client will by default unset them and allows to optionally leave them.

Based on all existing tests still passing, I believe this change is fully backwards-compatible (But depends on breaking changes such as a very new client-go and apimachinery and https://github.com/kubernetes-sigs/controller-runtime/pull/3228).

Fixes #2341 

[0]: https://github.com/kubernetes/kubernetes/blob/4dc7a48ac6fb631a84e1974772bf7b8fd0bb9c59/staging/src/k8s.io/client-go/testing/fixture.go#L643

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -56,13 +56,16 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/managedfields"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/watch"
+	clientgoapplyconfigurations "k8s.io/client-go/applyconfigurations"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/testing"
 	"k8s.io/utils/ptr"
@@ -98,6 +101,8 @@ type fakeClient struct {
 	indexes map[schema.GroupVersionKind]map[string]client.IndexerFunc
 	// indexesLock must be held when accessing indexes.
 	indexesLock sync.RWMutex
+
+	returnManagedFields bool
 }
 
 var _ client.WithWatch = &fakeClient{}
@@ -131,6 +136,8 @@ type ClientBuilder struct {
 	withStatusSubresource []client.Object
 	objectTracker         testing.ObjectTracker
 	interceptorFuncs      *interceptor.Funcs
+	typeConverters        []managedfields.TypeConverter
+	returnManagedFields   bool
 
 	// indexes maps each GroupVersionKind (GVK) to the indexes registered for that GVK.
 	// The inner map maps from index name to IndexerFunc.
@@ -172,6 +179,8 @@ func (f *ClientBuilder) WithRuntimeObjects(initRuntimeObjs ...runtime.Object) *C
 }
 
 // WithObjectTracker can be optionally used to initialize this fake client with testing.ObjectTracker.
+// Setting this is incompatible with setting WithTypeConverters, as they are a setting on the
+// tracker.
 func (f *ClientBuilder) WithObjectTracker(ot testing.ObjectTracker) *ClientBuilder {
 	f.objectTracker = ot
 	return f
@@ -228,6 +237,31 @@ func (f *ClientBuilder) WithInterceptorFuncs(interceptorFuncs interceptor.Funcs)
 	return f
 }
 
+// WithTypeConverters sets the type converters for the fake client. The list is ordered and the first
+// non-erroring converter is used. A type converter must be provided for all types the client is used
+// for, otherwise it will error.
+//
+// This setting is incompatible with WithObjectTracker, as the type converters are a setting on the tracker.
+//
+// If unset, this defaults to:
+// * clientgoapplyconfigurations.NewTypeConverter(scheme.Scheme),
+// * managedfields.NewDeducedTypeConverter(),
+//
+// Be aware that the behavior of the `NewDeducedTypeConverter` might not match the behavior of the
+// Kubernetes APIServer, it is recommended to provide a type converter for your types. TypeConverters
+// are generated along with ApplyConfigurations.
+func (f *ClientBuilder) WithTypeConverters(typeConverters ...managedfields.TypeConverter) *ClientBuilder {
+	f.typeConverters = append(f.typeConverters, typeConverters...)
+	return f
+}
+
+// WithReturnManagedFields configures the fake client to return managedFields
+// on objects.
+func (f *ClientBuilder) WithReturnManagedFields() *ClientBuilder {
+	f.returnManagedFields = true
+	return f
+}
+
 // Build builds and returns a new fake client.
 func (f *ClientBuilder) Build() client.WithWatch {
 	if f.scheme == nil {
@@ -236,8 +270,6 @@ func (f *ClientBuilder) Build() client.WithWatch {
 	if f.restMapper == nil {
 		f.restMapper = meta.NewDefaultRESTMapper([]schema.GroupVersion{})
 	}
-
-	var tracker versionedTracker
 
 	withStatusSubResource := sets.New(inTreeResourcesWithStatus()...)
 	for _, o := range f.withStatusSubresource {
@@ -248,10 +280,33 @@ func (f *ClientBuilder) Build() client.WithWatch {
 		withStatusSubResource.Insert(gvk)
 	}
 
+	if f.objectTracker != nil && len(f.typeConverters) > 0 {
+		panic(errors.New("WithObjectTracker and WithTypeConverters are incompatible"))
+	}
+
 	if f.objectTracker == nil {
-		tracker = versionedTracker{ObjectTracker: testing.NewObjectTracker(f.scheme, scheme.Codecs.UniversalDecoder()), scheme: f.scheme, withStatusSubresource: withStatusSubResource}
-	} else {
-		tracker = versionedTracker{ObjectTracker: f.objectTracker, scheme: f.scheme, withStatusSubresource: withStatusSubResource}
+		if len(f.typeConverters) == 0 {
+			// Use corresponding scheme to ensure the converter error
+			// for types it can't handle.
+			clientGoScheme := runtime.NewScheme()
+			if err := scheme.AddToScheme(clientGoScheme); err != nil {
+				panic(fmt.Sprintf("failed to construct client-go scheme: %v", err))
+			}
+			f.typeConverters = []managedfields.TypeConverter{
+				clientgoapplyconfigurations.NewTypeConverter(clientGoScheme),
+				managedfields.NewDeducedTypeConverter(),
+			}
+		}
+		f.objectTracker = testing.NewFieldManagedObjectTracker(
+			f.scheme,
+			serializer.NewCodecFactory(f.scheme).UniversalDecoder(),
+			multiTypeConverter{upstream: f.typeConverters},
+		)
+	}
+	tracker := versionedTracker{
+		ObjectTracker:         f.objectTracker,
+		scheme:                f.scheme,
+		withStatusSubresource: withStatusSubResource,
 	}
 
 	for _, obj := range f.initObject {
@@ -276,6 +331,7 @@ func (f *ClientBuilder) Build() client.WithWatch {
 		restMapper:            f.restMapper,
 		indexes:               f.indexes,
 		withStatusSubresource: withStatusSubResource,
+		returnManagedFields:   f.returnManagedFields,
 	}
 
 	if f.interceptorFuncs != nil {
@@ -373,6 +429,9 @@ func convertFromUnstructuredIfNecessary(s *runtime.Scheme, o runtime.Object) (ru
 	if err != nil {
 		return nil, fmt.Errorf("scheme recognizes %s but failed to produce an object for it: %w", gvk, err)
 	}
+	if _, isTypedUnstructured := typed.(runtime.Unstructured); isTypedUnstructured {
+		return o, nil
+	}
 
 	unstructuredSerialized, err := json.Marshal(u)
 	if err != nil {
@@ -395,12 +454,20 @@ func (t versionedTracker) Update(gvr schema.GroupVersionResource, obj runtime.Ob
 }
 
 func (t versionedTracker) update(gvr schema.GroupVersionResource, obj runtime.Object, ns string, isStatus, deleting bool, opts metav1.UpdateOptions) error {
-	obj, err := t.updateObject(gvr, obj, ns, isStatus, deleting, opts.DryRun)
+	gvk, err := apiutil.GVKForObject(obj, t.scheme)
+	if err != nil {
+		return err
+	}
+	obj, err = t.updateObject(gvr, obj, ns, isStatus, deleting, opts.DryRun)
 	if err != nil {
 		return err
 	}
 	if obj == nil {
 		return nil
+	}
+
+	if u, unstructured := obj.(*unstructured.Unstructured); unstructured {
+		u.SetGroupVersionKind(gvk)
 	}
 
 	return t.ObjectTracker.Update(gvr, obj, ns, opts)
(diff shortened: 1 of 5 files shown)


alvaroaleman:
/lifecycle frozen
This is a requested feature so I'd like to keep this open to demonstrate the challenges. I personally will not be working further on it though.

pkg/client/fake/client.go
```diff
@@ -160,6 +165,8 @@ func (f *ClientBuilder) WithRuntimeObjects(initRuntimeObjs ...runtime.Object) *C
 }
 
 // WithObjectTracker can be optionally used to initialize this fake client with testing.ObjectTracker.
+// Setting this is incompatible with setting WithTypeConverters, as they are a setting on the
```
jpbetz, line 168: Enforce this in code with an assertion?

pkg/client/fake/client.go
```diff
@@ -216,6 +223,18 @@ func (f *ClientBuilder) WithInterceptorFuncs(interceptorFuncs interceptor.Funcs)
 	return f
 }
 
+// WithTypeConverters sets the type converters for the fake client. The list is ordered and the first
+// non-erroring converter is used.
+// This setting is incompatible with WithObjectTracker, as the type converters are a setting on the tracker.
```
jpbetz, line 228: Enforce this in code with an assertion?

pkg/client/fake/typeconverter.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerror "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/managedfields"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+type multiTypeConverter struct {
+	upstream []managedfields.TypeConverter
+}
```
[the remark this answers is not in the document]
alvaroaleman: Yep makes sense, will do that

tomasaschan:
@alvaroaleman I ran into this myself but was able to verify that this PR solves it well enough for my own needs at least.

Is there anything I can do to help land this in a release?

alvaroaleman:
@tomasaschan this can not be merged in its current form, we need support in the upstream mapper for reloading the scheme and find a way to avoid breaking everyone by adding the managed fields.

pkg/client/fake/client.go
```diff
@@ -119,6 +123,7 @@ type ClientBuilder struct {
 	withStatusSubresource []client.Object
 	objectTracker         testing.ObjectTracker
 	interceptorFuncs      *interceptor.Funcs
+	typeConverters        []managedfields.TypeConverter
```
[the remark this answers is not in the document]
alvaroaleman: This is on the builder and often times it will be needed to have more than one, as both in-tree and CRs are used

tomasaschan:
@fsommar I think there's an old (probably closed PR, or intermediate step later refactored away) attempt at using this to test repo-enabler, if you're curious to test it on a "live" code base.

fsommar:
> @fsommar I think there's an old (probably closed PR, or intermediate step later refactored away) attempt at using this to test repo-enabler, if you're curious to test it on a "live" code base.

I found usages of client-go fakes in an old immutator commit 759c336c833e420e0b0ff33dc5f06229540b88b7 on PR#20 fwiw, but not controller-runtime fakes. IIRC you tried controller-runtime fakes first and decided to pivot to use client-go, which had the same issue in the end (I can't find those first attempts in a pushed commit). And now it's using envtest.

We have some upcoming work using controller-runtime likely in July. I'll try it out then, as I have a short vacation first!

tomasaschan:
Here's the corresponding client-go fake issue, for reference: https://github.com/kubernetes/kubernetes/issues/133549

pkg/client/fake/client.go
```diff
@@ -228,6 +237,29 @@ func (f *ClientBuilder) WithInterceptorFuncs(interceptorFuncs interceptor.Funcs)
...
+// non-erroring converter is used. A type converter must be provided for all types the client is used
+// for, otherwise it will error.
+// This setting is incompatible with WithObjectTracker, as the type converters are a setting on the tracker.
+//
+// If unset, this defaults to:
+// * clientgoapplyconfigurations.NewTypeConverter(scheme.Scheme),
+// * managedfields.NewDeducedTypeConverter(),
+//
+// Be aware that the behavior of the `NewDeducedTypeConverter` might not match the behavior of the
+// Kubernetes APIServer, it is recommended to provide a type converter for your types.
```
sbueringer: > it is recommended to provide a type converter for your types

Do we have an example somewhere on how to write one? (e.g. via unit tests, I don't see this func called there)

I think a unit test for this would be good and it also works as an example

pkg/client/fake/client.go
```diff
@@ -228,6 +237,29 @@ func (f *ClientBuilder) WithInterceptorFuncs(interceptorFuncs interceptor.Funcs)
 	return f
 }
 
+// WithTypeConverters sets the type converters for the fake client. The list is ordered and the first
+// non-erroring converter is used. A type converter must be provided for all types the client is used
```
sbueringer, line 241: Under which exact circumstances must these be provided?

Only if SSA with custom types is used with the fake client? (based on the unit test it looks like Unstructured also works without setting TypeConverters)

pkg/client/fake/client.go
```diff
@@ -248,11 +278,29 @@ func (f *ClientBuilder) Build() client.WithWatch {
...
+
 	if f.objectTracker == nil {
-		tracker = versionedTracker{ObjectTracker: testing.NewObjectTracker(f.scheme, scheme.Codecs.UniversalDecoder()), scheme: f.scheme, withStatusSubresource: withStatusSubResource}
-	} else {
-		tracker = versionedTracker{ObjectTracker: f.objectTracker, scheme: f.scheme, withStatusSubresource: withStatusSubResource}
+		if len(f.typeConverters) == 0 {
+			f.typeConverters = []managedfields.TypeConverter{
+				// Use corresponding scheme to ensure the converter error
+				// for types it can't handle.
+				clientgoapplyconfigurations.NewTypeConverter(scheme.Scheme),
```
sbueringer: The goal is to use a scheme that only contains clientgo types, right?

I wonder if we should use another scheme. This is a package global variable and people might have added all kinds of types to it (I've seen that in the past.. :))

pkg/client/fake/client.go
```diff
@@ -248,11 +278,29 @@ func (f *ClientBuilder) Build() client.WithWatch {
...
+		f.objectTracker = testing.NewFieldManagedObjectTracker(
+			f.scheme,
+			serializer.NewCodecFactory(f.scheme).UniversalDecoder(),
+			multiTypeConverter{upstream: f.typeConverters},
+		)
 	}
+	tracker := versionedTracker{
+		ObjectTracker:         f.objectTracker,
+		scheme:                f.scheme,
+		withStatusSubresource: withStatusSubResource}
```
sbueringer: nit: move `}` to the next line

pkg/client/fake/client.go
```diff
@@ -523,6 +583,10 @@ func (t versionedTracker) updateObject(gvr schema.GroupVersionResource, obj runt
 }
 
 func (c *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	if err := c.addToSchemeIfUnknownAndUnstructured(obj); err != nil {
```
sbueringer: Looks like also if Partial, we should rename the func somehow

Maybe `addUnstructuredAndPartialToSchemeIfUnknown`

or 

`addToSchemeIfUnknownAndUnstructuredOrPartial`

pkg/client/fake/client.go
```diff
@@ -940,6 +1068,12 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 	if err != nil {
 		return err
 	}
+
+	// otherwise the merge logic in the tracker complains
```
sbueringer: So the tracker complains, but the apiserver just ignores them right?
So we need this to end up with the same behavior as the apiserver?

And I assume it is also possible with the fake client to patch managedFields with the patch, right? (vs. the obj where we are dropping managedFields here)

Wondering if we should have test coverage to ensure we can patch managedFields directly with the fake client like with the regular client (assuming it is supported, otherwise it would be good to fail in this case)

pkg/client/fake/client.go
```diff
@@ -949,7 +1083,10 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 	defer c.trackerWriteLock.Unlock()
 	oldObj, err := c.tracker.Get(gvr, accessor.GetNamespace(), accessor.GetName())
 	if err != nil {
-		return err
+		if patch.Type() != types.ApplyPatchType {
+			return err
+		}
+		oldObj = &unstructured.Unstructured{}
```
sbueringer, line 1091: Fine to ignore all errors in this case or could/should we only ignore certain errors? (like "not found")

pkg/client/fake/client.go
```diff
@@ -1022,12 +1163,15 @@ func deletionTimestampEqual(newObj metav1.Object, obj metav1.Object) bool {
...
 // and easier than refactoring the k8s client-go method upstream.
 // Duplicate of upstream: https://github.com/kubernetes/client-go/blob/783d0d33626e59d55d52bfd7696b775851f92107/testing/fixture.go#L146-L194
-func dryPatch(action testing.PatchActionImpl, tracker testing.ObjectTracker) (runtime.Object, error) {
+func dryPatch(action testing.PatchActionImpl, tracker testing.ObjectTracker, newObj runtime.Object) (runtime.Object, error) {
 	ns := action.GetNamespace()
 	gvr := action.GetResource()
 
 	obj, err := tracker.Get(gvr, ns, action.GetName())
 	if err != nil {
+		if action.GetPatchType() == types.ApplyPatchType {
```
sbueringer: Fine to ignore all errors in this case or could/should we only ignore certain errors?

pkg/client/fake/typeconverter.go
```diff
@@ -0,0 +1,60 @@
...
+		res, err := u.ObjectToTyped(r, o...)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		return res, nil
+	}
+
+	return nil, fmt.Errorf("failed to convert Object to Typed: %w", kerrors.NewAggregate(errs))
```
sbueringer:
```suggestion
	return nil, fmt.Errorf("failed to convert Object to TypedValue: %w", kerrors.NewAggregate(errs))
```
nit: Or drop the Value suffix in the error in l.59

pkg/client/fake/client_test.go
```diff
@@ -2554,6 +2538,51 @@ var _ = Describe("Fake client", func() {
 		wg.Wait()
 	})
 
+	It("supports server-side apply of a client-go resource", func() {
```
sbueringer, line 2543: Should we have test coverage that tests that managedFields are returned? (and only if the flag is set)

pkg/client/fake/client_test.go
```diff
@@ -2554,6 +2538,51 @@ var _ = Describe("Fake client", func() {
...
+		cl := NewClientBuilder().Build()
+		obj := &unstructured.Unstructured{}
+		obj.SetAPIVersion("custom/v1")
+		obj.SetKind("FakeResource")
+		obj.SetName("foo")
+		Expect(unstructured.SetNestedField(obj.Object, map[string]any{"some": "data"}, "spec")).To(Succeed())
+
+		Expect(cl.Patch(context.Background(), obj, client.Apply)).To(Succeed())
+
+		result := obj.DeepCopy()
```
sbueringer: nit: Maybe just deepcopy in l.2569 so you don't need l.2574

sbueringer:
Nice! Thx for working on this.

My comments are mostly around improving test coverage and godoc a bit

pkg/client/fake/client.go
alvaroaleman: It is always requiered if the FieldManagedObjectTracker is used, it seems to error out on all of its methods if the typeConverter doesn't handle a given type

pkg/client/fake/client.go
alvaroaleman: They get generated alongside the applyconfigs, I have added a comment to that effect. As we don't have custom applyconfigs in this repo, I've added a test that uses the client-go type converter.

pkg/client/fake/client.go

pkg/client/fake/client.go

pkg/client/fake/client.go
alvaroaleman: So the KAS silently drops these if set excpet when using SSA, there it will error out if its set. I would expect the upstream tracker to have tests for that.

The question is what we should do in the SSA case if those are set - I am starting to think it would be better to not remove them and instead just return the error, WDYT?

pkg/client/fake/client.go
alvaroaleman: Updated to specifically only ignore NotFound

pkg/client/fake/client.go
alvaroaleman: Updated to specifically only ignore NotFound

pkg/client/fake/typeconverter.go
alvaroaleman: Done

pkg/client/fake/client_test.go
alvaroaleman: Added

pkg/client/fake/client_test.go
alvaroaleman: Done

pkg/client/fake/client.go
alvaroaleman: But then we have to deal with keeping that scheme in sync with client-go, right? I.E. update it every time a new type is added there?

alvaroaleman:
@tomasaschan @fsommar could you please find a different place for your comments that are not about feedback to the code changes in here to avoid polluting this? Thanks!

pkg/client/fake/client.go
sbueringer: I think I'm missing something.

So if I'm using the fake client today with my own CRDs, this will now break if I don't start setting type converters for them? This would break a lot of people

(FieldManagedObjectTracker is always used if objectTracker is not explicitly set, right?)

pkg/client/fake/client.go
sbueringer: > So the KAS silently drops these if set excpet when using SSA, there it will error out if its set.

Ah, I thought it also ignores them with SSA, but I don't really remember. All good if you're sure :)

> The question is what we should do in the SSA case if those are set - I am starting to think it would be better to not remove them and instead just return the error, WDYT?

I would prefer to get the same behavior as with the KAS if possible. So if I understood you right, we should return an error just like KAS would do

pkg/client/fake/client.go
alvaroaleman: > So if I'm using the fake client today with my own CRDs, this will now break if I don't start setting type converters for them? This would break a lot of people

No it will not break. SSA might not work correctly due to the inferencing type converter, but that is all.

pkg/client/fake/client.go
alvaroaleman: Updated it to not purge them prior to doing SSA and just return the error + test

pkg/client/fake/client.go
sbueringer: Could it be that we can do ~ the following  https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go#L159-L162 (with a local scheme that we create)

alvaroaleman: Oh nice, yeah that works, updated

pkg/client/fake/client.go
sbueringer: Okay, can we make this slightly clearer please? The current comment is
```go
A type converter must be provided for all types the client is used for, otherwise it will error.
```

I'm not sure if folks will infer that correctly otherwise

pkg/client/fake/client_test.go
sbueringer: Thx!

sbueringer:
@alvaroaleman Last one from my side: https://github.com/kubernetes-sigs/controller-runtime/pull/2981#discussion_r2169194034

PR #2982, opened 2024-10-14
🐛 Fix custom defaulter: avoid deleting unknown fields (zero change patch)

trasc:
When creating a patch for the custom defaulter, don't return the removals generated by object decoding with the local scheme.

Otherwise, the patch will include a remove instruction for fields that the go type doesn't contain.

This could happen when building a webhook for a resource that belongs to another project (including k8s types).

This PR stops the defaulter from pruning the fields that are not recognized in the local scheme.
When doing this we are creating two patches one before and one after the custom defaulter call and only send back the remove operations that are not present in both patches.

The user can opt-out from this new behavior by specifying the  `DefaulterRemoveUnknownFields`  option when calling ` WithCustomDefaulter`. 

(This is a rework of #2931)

### DIFF
--- a/pkg/webhook/admission/defaulter_custom.go
+++ b/pkg/webhook/admission/defaulter_custom.go
@@ -40,15 +40,18 @@ type defaulterOptions struct {
 	removeUnknownFields bool
 }
 
-type defaulterOption func(*defaulterOptions)
+// DefaulterOption defines the type of a CustomDefaulter's option
+type DefaulterOption func(*defaulterOptions)
 
-// DefaulterRemoveUnknownFields makes the defaulter prune the fields that are not recognized in the local scheme.
+// DefaulterRemoveUnknownFields makes the defaulter prune fields that are in the json object retrieved by the
+// webhook but not in the local go type. This happens for example when the CRD in the apiserver has fields that
+// our go type doesn't know about, because it's outdated.
 func DefaulterRemoveUnknownFields(o *defaulterOptions) {
 	o.removeUnknownFields = true
 }
 
 // WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
-func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter, opts ...defaulterOption) *Webhook {
+func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter, opts ...DefaulterOption) *Webhook {
 	options := &defaulterOptions{}
 	for _, o := range opts {
 		o(options)
--- a/pkg/webhook/admission/defaulter_custom.go
+++ b/pkg/webhook/admission/defaulter_custom.go
@@ -37,17 +37,17 @@ type CustomDefaulter interface {
 }
 
 type defaulterOptions struct {
-	removeUnknownFields bool
+	removeUnknownOrOmitableFields bool
 }
 
 // DefaulterOption defines the type of a CustomDefaulter's option
 type DefaulterOption func(*defaulterOptions)
 
-// DefaulterRemoveUnknownFields makes the defaulter prune fields that are in the json object retrieved by the
-// webhook but not in the local go type. This happens for example when the CRD in the apiserver has fields that
-// our go type doesn't know about, because it's outdated.
-func DefaulterRemoveUnknownFields(o *defaulterOptions) {
-	o.removeUnknownFields = true
+// DefaulterRemoveUnknownOrOmitableFields makes the defaulter prune fields that are in the json object retrieved by the
+// webhook but not in the local go type json representation. This happens for example when the CRD in the apiserver has
+// fields that our go type doesn't know about, because it's outdated, or the field has a zero value and is `omitempty`.
+func DefaulterRemoveUnknownOrOmitableFields(o *defaulterOptions) {
+	o.removeUnknownOrOmitableFields = true
 }
 
 // WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
@@ -57,15 +57,20 @@ func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter C
 		o(options)
 	}
 	return &Webhook{
-		Handler: &defaulterForType{object: obj, defaulter: defaulter, decoder: NewDecoder(scheme), removeUnknownFields: options.removeUnknownFields},
+		Handler: &defaulterForType{
+			object:                        obj,
+			defaulter:                     defaulter,
+			decoder:                       NewDecoder(scheme),
+			removeUnknownOrOmitableFields: options.removeUnknownOrOmitableFields,
+		},
 	}
 }
 
 type defaulterForType struct {
-	defaulter           CustomDefaulter
-	object              runtime.Object
-	decoder             Decoder
-	removeUnknownFields bool
+	defaulter                     CustomDefaulter
+	object                        runtime.Object
+	decoder                       Decoder
+	removeUnknownOrOmitableFields bool
 }
 
 // Handle handles admission requests.
@@ -100,7 +105,7 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
 
 	// Keep a copy of the object if needed
 	var originalObj runtime.Object
-	if !h.removeUnknownFields {
+	if !h.removeUnknownOrOmitableFields {
 		originalObj = obj.DeepCopyObject()
 	}
 
@@ -120,7 +125,7 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
 	}
 
 	handlerResponse := PatchResponseFromRaw(req.Object.Raw, marshalled)
-	if !h.removeUnknownFields {
+	if !h.removeUnknownOrOmitableFields {
 		handlerResponse = h.dropSchemeRemovals(handlerResponse, originalObj, req.Object.Raw)
 	}
 	return handlerResponse
--- a/pkg/webhook/admission/defaulter_custom.go
+++ b/pkg/webhook/admission/defaulter_custom.go
@@ -21,11 +21,14 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"slices"
 
+	"gomodules.xyz/jsonpatch/v2"
 	admissionv1 "k8s.io/api/admission/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // CustomDefaulter defines functions for setting defaults on resources.
@@ -71,14 +74,16 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
 	ctx = NewContextWithRequest(ctx, req)
 
 	// Get the object in the request
-	original := h.object.DeepCopyObject()
-	if err := h.decoder.Decode(req, original); err != nil {
+	obj := h.object.DeepCopyObject()
+	if err := h.decoder.Decode(req, obj); err != nil {
 		return Errored(http.StatusBadRequest, err)
 	}
 
+	// Keep a copy of the object
+	originalObj := obj.DeepCopyObject()
+
 	// Default the object
-	updated := original.DeepCopyObject()
-	if err := h.defaulter.Default(ctx, updated); err != nil {
+	if err := h.defaulter.Default(ctx, obj); err != nil {
 		var apiStatus apierrors.APIStatus
 		if errors.As(err, &apiStatus) {
 			return validationResponseFromStatus(false, apiStatus.Status())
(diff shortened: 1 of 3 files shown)


alculquicondor:
For context to c-r maintainers, we have been discussing what alternatives we have for a generic defaulter that doesn't drop unknown fields. Here is the thread for additional context https://github.com/kubernetes-sigs/kueue/pull/3194#issuecomment-2405174799

Taking @mimowo's words, we would be happy to first merge a solution in kueue and let it mature there before moving it here. But we would appreciate your input to the approaches.

trasc:
> @alculquicondor @trasc @mimowo We (Stefan, Vince and I) had some discussion around #2932 vs this and we landed at this here being better but we aren't 100% sure if its compatible in all cases and currently don't have time to spend on that question, which is why we would like an opt-out for what is suggested here (but on by default).
> 
> Does that sound reasonable to you all?

Sure, I'll ping you when I have something in that direction.

alculquicondor:
Thanks for considering the options!

Do you have any particular guideline on how the opt-out mechanism could look like? Should we have a different method in the factory? Or should the method `WithDefaulter` accept some options? Any other suggestion?

Other than that, I hope the overhead of the approach is not significant. Fortunately, a new encoding is coming to k/k CRDs https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4222-cbor-serializer#motivation

trasc:
I think 

>... should the method `WithDefaulter` accept some options ...

it leave room to maybe add an additional option that removes the `add` zero values.

trasc:
@alvaroaleman please have a look at my `DefaulterPreserveUnknownFields` proposal when you have some time.

Also please let me know if we can do something about the `pull-controller-runtime-apidiff` check, 

```
changed from 
func(*k8s.io/apimachinery/pkg/runtime.Scheme, k8s.io/apimachinery/pkg/runtime.Object, CustomDefaulter) *Webhook to func(*k8s.io/apimachinery/pkg/runtime.Scheme, k8s.io/apimachinery/pkg/runtime.Object, CustomDefaulter, ...defaulterOption) *Webhook
```
 is a flase positive in my opinion, as you don't need to actually do something to old code in order to use the new version.

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -21,29 +21,48 @@ import (
...
+type defaulterOption func(*defaulterOptions)
+
+// DefaulterPreserveUnknownFields stops the defaulter from pruning the fields that are not recognized in the local scheme.
+func DefaulterPreserveUnknownFields(o *defaulterOptions) {
+	o.preserveUnknownFields = true
+}
+
 // WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
-func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter) *Webhook {
+func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter, opts ...defaulterOption) *Webhook {
```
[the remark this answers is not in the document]
trasc: done

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -90,5 +115,40 @@ func (h *defaulterForType) Handle(ctx context.Context, req Request) Response {
...
+		return r
+	}
+
+	// Get the raw to original patch
+	marshalledOriginal, err := json.Marshal(original)
+	if err != nil {
+		return Errored(http.StatusInternalServerError, err)
+	}
+
+	patchOriginal := PatchResponseFromRaw(raw, marshalledOriginal).Patches
```
alculquicondor, line 143: since we don't care about the response, it makes more sense to call `jsonpatch.CreatePatch` directly

pkg/webhook/admission/defaulter_custom_test.go
```diff
@@ -28,6 +29,66 @@ import (
...
+			},
+			jsonpatch.JsonPatchOperation{
+				Operation: "remove",
+				Path:      "/totalReplicas",
+			},
+		))
+		Expect(resp.Result.Code).Should(Equal(int32(http.StatusOK)))
+	})
+
+	It("should preserve unknown fields when DefaulterPreserveUnknownFields is passed", func() {
```
alculquicondor, line 64: title is off now

pkg/webhook/admission/defaulter_custom.go
trasc: done

pkg/webhook/admission/defaulter_custom_test.go
trasc: done

pkg/webhook/admission/defaulter_custom_test.go
```diff
@@ -29,7 +29,7 @@ import (
 
 var _ = Describe("Defaulter Handler", func() {
 
-	It("should not preserve unknown fields by default", func() {
+	It("should  unknown fields when DefaulterRemoveUnknownFields is passed", func() {
```
alculquicondor, line 32: remove

trasc: done

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -21,29 +21,48 @@ import (
...
+type defaulterOption func(*defaulterOptions)
+
+// DefaulterRemoveUnknownFields makes the defaulter prune the fields that are not recognized in the local scheme.
+func DefaulterRemoveUnknownFields(o *defaulterOptions) {
+	o.removeUnknownFields = true
+}
+
 // WithCustomDefaulter creates a new Webhook for a CustomDefaulter interface.
-func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter) *Webhook {
+func WithCustomDefaulter(scheme *runtime.Scheme, obj runtime.Object, defaulter CustomDefaulter, opts ...defaulterOption) *Webhook {
```
alvaroaleman, line 51: Can we also please add a variadic arg for this into ` *WebhookBuilder) WithDefaulter`? Right now its not possible to configure this from there

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -21,29 +21,48 @@ import (
...
 	Default(ctx context.Context, obj runtime.Object) error
 }
 
+type defaulterOptions struct {
+	removeUnknownFields bool
+}
+
+type defaulterOption func(*defaulterOptions)
+
+// DefaulterRemoveUnknownFields makes the defaulter prune the fields that are not recognized in the local scheme.
```
alvaroaleman, line 45: How about:
```suggestion
// DefaulterRemoveUnknownFields makes the defaulter prune fields that are in the json object retrieved by the webhook but not in the our go type. This happens for example when the CRD in the apiserver has fields that the our go type doens't know about, because its outdated.
```

pkg/webhook/admission/defaulter_custom.go
trasc: done

pkg/webhook/admission/defaulter_custom.go
trasc: done

alvaroaleman: Looks like you copied my grammar and spelling mistakes from the comment 😬 Lgtm other than that

trasc: Yes, I'm pretty lazy that way :).
 Updated.

alvaroaleman (approved):
lgtm, thanks for this
/hold
For @sbueringer and @vincepri

sbueringer:
I'll take a look soon, just low bandwith at the moment

pkg/webhook/admission/defaulter_custom.go
```diff
@@ -21,29 +21,51 @@ import (
...
+	removeUnknownFields bool
+}
+
+// DefaulterOption defines the type of a CustomDefaulter's option
+type DefaulterOption func(*defaulterOptions)
+
+// DefaulterRemoveUnknownFields makes the defaulter prune fields that are in the json object retrieved by the
+// webhook but not in the local go type. This happens for example when the CRD in the apiserver has fields that
+// our go type doesn't know about, because it's outdated.
+func DefaulterRemoveUnknownFields(o *defaulterOptions) {
```
sbueringer, line 49: Not sure if the name of this option is correct. It will remove more than just unknown fields, right?

It will remove all fields that would be dropped by json.Unmarshal + json.Marshal.

If I see correctly this would also affect for example fields with omitempty. An example:

Type
```
   // +optional
   MinReadySeconds int32 `json:"minReadySeconds,omitempty"`
```

Incoming object:
```
   minReadySeconds: 0
```

After json.Unmarshal + json.Marshal this field would also be removed.

So this option removes more than just unknown fields, right?

(I"m aware that it just enables the previous behavior, but I'm wondering if the name is misleading)

sbueringer:
In general okay. One comment about the option name

pkg/webhook/admission/defaulter_custom.go
trasc: How about something like:

```go
// DefaulterRemoveUnknownOrOmitableFields makes the defaulter prune fields that are in the json object retrieved by the
// webhook but not in the local go type json representation. This happens for example when the CRD in the apiserver has
// fields that our go type doesn't know about, because it's outdated, the field has a zero value and is `omitempty`.
func DefaulterRemoveUnknownOrOmitableFields(o *defaulterOptions) {

```

trasc: Updated

sbueringer: Thx!

sbueringer:
/lgtm

@alvaroaleman Keeping the hold in case you want to take a final look after the last change

alvaroaleman:
> PR needs rebase.

LGTM modulo the rebase

trasc:
Rebased

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

sbueringer:
@trasc @alculquicondor Hey folks,

just an early heads up already. I bumped Cluster API to CR main branch and I'm now getting panics from the webhook: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api/11633/pull-cluster-api-e2e-blocking-main/1874860029216034816/artifacts/clusters/bootstrap/logs/capd-system/capd-controller-manager/capd-controller-manager-5b59b9b76-4kmz2/manager.log

Looks like jsonpatch.JsonPatchOperation is not always hashable
```
E0102 17:01:30.424172       1 <autogenerated>:1] "Observed a panic" logger="admission" webhookGroup="infrastructure.cluster.x-k8s.io" webhookKind="DockerClusterTemplate" DockerClusterTemplate="quick-start-zwk5pp/quick-start-cluster" namespace="quick-start-zwk5pp" name="quick-start-cluster" resource={"group":"infrastructure.cluster.x-k8s.io","version":"v1beta1","resource":"dockerclustertemplates"} user="kubernetes-admin" requestID="39810ae0-c339-4d68-a8cf-c007f0215a5b" panic="runtime error: hash of unhashable type map[string]interface {}" panicGoValue="\"hash of unhashable type map[string]interface {}\"" stacktrace=<
	goroutine 515 [running]:
	k8s.io/apimachinery/pkg/util/runtime.logPanic({0x259abe0, 0xc000822690}, {0x1ee0d80, 0xc00086cb10})
		k8s.io/apimachinery@v0.32.0/pkg/util/runtime/runtime.go:107 +0xbc
	sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).Handle.func1()
		sigs.k8s.io/controller-runtime@v0.19.3/pkg/webhook/admission/webhook.go:163 +0x103
	panic({0x1ee0d80?, 0xc00086cb10?})
		runtime/panic.go:785 +0x132
	type:.hash.gomodules.xyz/jsonpatch/v2.Operation(0xc0007ea7c8, 0x5aa77e?)
		<autogenerated>:1 +0x45
	k8s.io/apimachinery/pkg/util/sets.Set[...].Has(...)
		k8s.io/apimachinery@v0.32.0/pkg/util/sets/set.go:78
	sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*defaulterForType).dropSchemeRemovals.func3({{0x22158db, 0x3}, {0xc0003d34d0, 0x28}, {0x1ebeba0, 0xc000aa9500}})
		sigs.k8s.io/controller-runtime@v0.19.3/pkg/webhook/admission/defaulter_custom.go:158 +0x65
	slices.IndexFunc[...](...)
		slices/slices.go:109
	slices.DeleteFunc[...]({0xc000740a80?, 0x8, 0x8}, 0xc0007eab68)
		slices/slices.go:237 +0xe3
	sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*defaulterForType).dropSchemeRemovals(_, {{0xc000740a80, 0x8, 0x8}, {{0x0, 0x0}, 0x1, 0x0, {0x0, 0x0, ...}, ...}}, ...)
		sigs.k8s.io/controller-runtime@v0.19.3/pkg/webhook/admission/defaulter_custom.go:157 +0x4ac
	sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*defaulterForType).Handle(_, {_, _}, {{{0xc0003d2ab0, 0x24}, {{0xc000e0bc20, 0x1f}, {0xc000a75510, 0x7}, {0xc000c10780, ...}}, ...}})
		sigs.k8s.io/controller-runtime@v0.19.3/pkg/webhook/admission/defaulter_custom.go:129 +0x42e
```

sbueringer:
Not sure what the most efficient way is to compare jsonpatch.Operations (the straightforward option would be to call the `Json` method and compare that instead

alculquicondor:
I don't think we need to compare the whole thing, just the combination of operation and path.
I would be surprised if `jsonpatch.CreatePatch` generates more than one entry for a path, even.

Here is the type, for reference https://github.com/gomodules/jsonpatch/blob/e2d6410c0299d49b797482dc765a3b49ecc8a4b6/v3/jsonpatch.go#L17-L21

Should we revert for now?

cc @mbobrovskyi @mimowo

mbobrovskyi:
IIUC, the JSON Patch `remove` operation shouldn't include a `value`. I'm curious about the specific request needed to reproduce this scenario.

mbobrovskyi:
I think I've identified the problem. We're comparing all patches, which makes it possible to encounter a non-hashable value.

Created a fix https://github.com/kubernetes-sigs/controller-runtime/pull/3056.

sbueringer:
> Should we revert for now?

I think we're fine with the fix (#3056 should cover it)

> I think I've identified the problem. We're comparing all patches, which makes it possible to encounter a non-hashable value.

Looks perfect, thank you!

sbueringer:
Quick feedback. The panic is fixed 🎉 

I'm just hitting another issue in Cluster API. Not sure yet if that is something that we should care about in CR

API field:
```
	// imageTag allows to specify a tag for the image.
	// +optional
	ImageTag string `json:"imageTag,omitempty"`
```

Through a combination of unexpected behavior, we're trying to set this field to null via one of our CAPI controllers:
```
  "local" : {
    "imageTag" : null
  }
```

Before this change: our defaulting/mutating webhook generated this patch operation:
```
Operation: remove
Path: /spec/kubeadmConfigSpec/clusterConfiguration/dns/imageTag
```
=> The result was that the field was removed before CR validation

After this change: this patch operation is dropped by dropSchemeRemovals
=> The result is that imageTag stays null
=> Because the field is not nullable we end up with
```
failed to create KubeadmControlPlane.controlplane.cluster.x-k8s.io:
        FieldValueTypeInvalid: spec.kubeadmConfigSpec.clusterConfiguration.etcd.local.imageTag:
        Invalid value: "null": spec.kubeadmConfigSpec.clusterConfiguration.etcd.local.imageTag in body must be of type string
```

alculquicondor:
It looks like WAI, from the name of the defaulter option `removeUnknownOrOmitableFields`.

It's a bug in the CAPI controller that happened to be masked by the webhook implementation details.
I wonder why the CAPI controller sets `null`, tho. Doesn't it use the go types that are marked with `omitempty`?

sbueringer:
I agree. The behavior of the defaulter is exactly as expected. 

I'm only wondering how many folks we break that are in a similar situation as CAPI. CAPI itself is not a problem, I can deal with this there by setting the option. I think I have to set it because we can't make a change like this within an API version.

alculquicondor:
We did make the change opt-out. I think it's reasonable to make it opt-in for a few releases, have a deprecation note and eventually remove it.

sbueringer:
Talked to Alvaro. Let's keep it as opt-out. We'll document it as "breaking change" in the release notes, which should increase the chances that folks notice the change.

In general I woudl probably like to keep the option permanently. I don't want to force folks to fork the custom defaulter just because they want to keep the behavior of their API entirely stable.

PR #2986, opened 2024-10-15
✨  Support RateLimiterLatency & RequestLatency for client-go metrics (client, workqueue, reflector).

fengruotj:
Currently the controller-runtime only expose client-go metrics with requestLatency. In a production environment, we need to expose more client-go Metrics, such as requestLatency and rateLimiterLatency.

This PR adds more client-go metrics with rest_client_request_duration_seconds & rest_client_rate_limiter_duration_seconds.

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,6 +18,8 @@ package metrics
 
 import (
 	"context"
+	"net/url"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	clientmetrics "k8s.io/client-go/tools/metrics"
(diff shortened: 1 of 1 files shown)


fengruotj:
Hi @varshaprasad96 @vincepri this is the pr feature: Support RateLimiterLatency & RequestLatency for client-go metrics.

pkg/metrics/client_go_adapter.go
```diff
@@ -30,6 +32,26 @@ import (
...
+				1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60},
+		},
+		[]string{"verb", "host", "path"},
+	)
+
+	rateLimiterLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "rest_client_rate_limiter_duration_seconds",
+			Help: "Client side rate limiter latency in seconds. Broken down by verb and URL.",
+			Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
```
alvaroaleman, line 49: These have a ton of buckets and dimensions that include the path which means things like object names will be in there, this will result in a huge cardinality, we can not include something like this by default

fengruotj: My primary requirement is to register some custom metrics for client-go/tools/metrics. However, when using controller-runtime, I am unable to customize the registration of client-go metrics because registerClientMetrics() has a higher precedence. Therefore, I need to add some additional metrics. Perhaps we could reduce the number of buckets and some tags metrics in this case?

PR #2882, opened 2024-07-17
✨ Add HasOwnerReference

sivchari:
I added HasOwnerReference API. This API check if owners contains specific owner.

PR #2991, opened 2024-10-19
✨ Add public functions to add/remove OwnerReferences without the referenced object

### LINKED ISSUE #2989: Expose referSameObject function
If you are creating a new object, you may need a function to tie this new object to the lifecycle of the owners of the object being reconciled. For example, I have a controller that reconciles pods, which is writing secrets for the pods to use. In the case of a standalone pod, it's fine to just set the pod as an owner of this new secret using `controllerutil.SetOwnerReference()`. However, for pods which have an owner (ReplicaSets, DaemonSets, etc), the secret may be shared with all other related pods, as they are members of the same workload. This requires me to duplicate the logic of `controllerutil.upsertOwnerRef()` in my project, since I don't have the objects referred to by the pod's `ownerReferences`, only the references themselves and that function is not public. It would be useful in such cases to either make `controllerutil.upsertOwnerRef()` public, or expose that function through a new public function. If we add the ability to add an `OwnerReference` in this way, we should probably also add the ability to remove it as well.

keeganwitt:
This solves #2989 by providing new public functions in controllerutil for adding and removing OwnerReferences (current public functions only accept Objects). This seemed like a better approach than making the current private methods it uses public.

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -168,6 +168,36 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
 	return nil
 }
 
+// AppendOwnerReference is a helper method to make sure the given object contains a provided owner reference.
+// This allows you to declare that owner has a dependency on the object without specifying it as a controller.
+// If a reference already exists, it'll be overwritten with the newly provided version.
+func AppendOwnerReference(owner metav1.OwnerReference, object metav1.Object) {
+	upsertOwnerRef(owner, object)
+}
+
+// DropOwnerReference is a helper method to make sure the given object removes a provided owner reference.
+// This allows you to remove the owner to establish a new owner of the object in a subsequent call.
+func DropOwnerReference(owner metav1.OwnerReference, object metav1.Object) error {
+	ownerRefs := object.GetOwnerReferences()
+	index := indexOwnerRef(ownerRefs, metav1.OwnerReference{
+		APIVersion: owner.APIVersion,
+		Name:       owner.Name,
+		Kind:       owner.Kind,
+	})
+
+	if index == -1 {
+		return fmt.Errorf("%T does not have an controller reference for %T", object, owner)
+	}
+
+	if ownerRefs[index].Controller == nil || !*ownerRefs[index].Controller {
+		return fmt.Errorf("%T owner is not the controller reference for %T", owner, object)
+	}
+
+	ownerRefs = append(ownerRefs[:index], ownerRefs[index+1:]...)
+	object.SetOwnerReferences(ownerRefs)
+	return nil
+}
+
 // HasControllerReference returns true if the object
 // has an owner ref with controller equal to true
 func HasControllerReference(object metav1.Object) bool {
--- a/pkg/controller/controllerutil/controllerutil_test.go
+++ b/pkg/controller/controllerutil/controllerutil_test.go
@@ -276,6 +276,184 @@ var _ = Describe("Controllerutil", func() {
 		})
 	})
 
+	var _ = Describe("Controllerutil", func() {
+		Describe("AppendOwnerReference", func() {
+			It("should set ownerRef on an empty list", func() {
+				rs := &appsv1.ReplicaSet{}
+				dep := metav1.OwnerReference{
+					APIVersion: "v1",
+					Kind:       "Deployment",
+					Name:       "foo",
+					UID:        "foo-uid",
+				}
+				controllerutil.AppendOwnerReference(dep, rs)
+				Expect(rs.OwnerReferences).To(ConsistOf(metav1.OwnerReference{
+					Name:       "foo",
+					Kind:       "Deployment",
+					APIVersion: "extensions/v1beta1",
+					UID:        "foo-uid",
+				}))
+			})
+
+			It("should not duplicate owner references", func() {
+				rs := &appsv1.ReplicaSet{
+					ObjectMeta: metav1.ObjectMeta{
+						OwnerReferences: []metav1.OwnerReference{
+							{
+								Name:       "foo",
+								Kind:       "Deployment",
+								APIVersion: "extensions/v1beta1",
+								UID:        "foo-uid",
+							},
+						},
+					},
+				}
+				dep := metav1.OwnerReference{
+					APIVersion: "v1",
+					Kind:       "Deployment",
+					Name:       "foo",
+					UID:        "foo-uid",
+				}
+				controllerutil.AppendOwnerReference(dep, rs)
+				Expect(rs.OwnerReferences).To(ConsistOf(metav1.OwnerReference{
+					Name:       "foo",
+					Kind:       "Deployment",
+					APIVersion: "extensions/v1beta1",
+					UID:        "foo-uid",
+				}))
+			})
+
+			It("should update the reference", func() {
+				rs := &appsv1.ReplicaSet{
+					ObjectMeta: metav1.ObjectMeta{
+						OwnerReferences: []metav1.OwnerReference{
+							{
+								Name:       "foo",
+								Kind:       "Deployment",
+								APIVersion: "extensions/v1alpha1",
+								UID:        "foo-uid-1",
+							},
+						},
+					},
+				}
+				dep := metav1.OwnerReference{
+					APIVersion: "v1",
+					Kind:       "Deployment",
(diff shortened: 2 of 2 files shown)


keeganwitt:
We should think carefully about the naming convention. Ideally, I would have used `SetOwnerReference` and `RemoveOwnerReference` for these new methods, and `SetOwner` and `RemoveOwner` for the existing methods, but that would be a breaking change.

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,6 +168,36 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
...
+
+// DropOwnerReference is a helper method to make sure the given object removes a provided owner reference.
+// This allows you to remove the owner to establish a new owner of the object in a subsequent call.
+func DropOwnerReference(owner metav1.OwnerReference, object metav1.Object) error {
+	ownerRefs := object.GetOwnerReferences()
+	index := indexOwnerRef(ownerRefs, metav1.OwnerReference{
+		APIVersion: owner.APIVersion,
+		Name:       owner.Name,
+		Kind:       owner.Kind,
+	})
```
troy0820, line 186: We usually do this with GVK because relying on this may not be stable.  (The workaround exist because of issues around this).  Getting the GVK requires the scheme. Adding the scheme makes this method the `RemoveOwnerReference` which already exists. 

I am trying to understand the repeat logic around this functionality.  What are we trying to circumvent utilizing this logic?

pkg/controller/controllerutil/controllerutil.go
```diff
@@ -168,6 +168,36 @@ func RemoveOwnerReference(owner, object metav1.Object, scheme *runtime.Scheme) e
 	return nil
 }
 
+// AppendOwnerReference is a helper method to make sure the given object contains a provided owner reference.
+// This allows you to declare that owner has a dependency on the object without specifying it as a controller.
+// If a reference already exists, it'll be overwritten with the newly provided version.
```
troy0820, line 173: This comment is exactly like `SetOwnerReference`. How is this any different besides not using the scheme to get the GVK  and skipping `validateOwner`?

keeganwitt: I couldn't think of a way to use `Get` in `sigs.ki8s.io/controller-runtime/pkg/client` without know what struct I need in advance (ReplicaSet, Job, etc). If there's a way to do this, then perhaps this entire request is unneeded.

pkg/controller/controllerutil/controllerutil.go
keeganwitt: Could you elaborate on that instability? Is that documented somewhere?

troy0820: Discussed here: https://github.com/kubernetes-sigs/controller-runtime/pull/2882#discussion_r1681547839
Reference:

I am still trying to understand what you are trying to do because this looks very similar to what is already in `controllerutil` and conflating the two different functions where the logic is the same can cause confusion on what we are trying to achieve.

pkg/controller/controllerutil/controllerutil.go
troy0820: This also sidesteps the validation that the function provides when adding an owner reference.  Not sure if want to do that or allow the appending of owner references to any object outside of the scope where the owner would be setting it.  This seems to be very dangerous where the lifecycle is now beholden to an owner reference the owner object didn't explicitly set.

pkg/controller/controllerutil/controllerutil.go
keeganwitt: This function I added just so there was a pair of functions that work with OwnerReference instead of Object. I don't personally need the DropOwnerReference function.

pkg/controller/controllerutil/controllerutil.go
keeganwitt: I don't follow this second concern. In my experience, it's not at all unusual for a controller to create a resource on behalf of something, and have the resource share the lifecycle of that thing. Is your concern only about resources created by other controllers?

alvaroaleman:
What you are asking for here can be done with an `if !sices.ContainsFunc(o.OwnerRefs, ownerRef) { o.OwerRefs = append(o.OwnerRefs, ownerRef}`, respective `slices.DeleteFunc(o.OwnerRefs, func(or metav1.OwnerRef) bool {...})`.

Is it really necessary to add something like this? The main reason that I am not super convinced is that none of this is really specific to OwnerReferences, what we are adding here is basically generic slice functionality but only for owner refs.

keeganwitt:
> ContainsFunc

The `ContainsFunc` only takes a single predicate. How can you capture the logic of `aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name;` with a single predicate?

Also, even if you could do this generically with a slice, the logic mentioned above would still have to be copy/pasted because it's also not a public function.

troy0820:
I can't speak for alvaroaleman but I do know the non exposed function `upsertOwnerRef` 
1. Gets the ownerRefs
2. see's if the ownerRef is in the ownerReferences 
3. if not, it appends it to the owner reference
4. sets the owner references 

Which to me seems equivalent to what was provided in his comment.  The functionality of this is not specific to owner references as it's more of a golang slices thing.  

> The ContainsFunc only takes a single predicate. How can you capture the logic of aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name with a single predicate?

That's what I was asking around sidestepping the logic that exists to check that with the original functions.  The two issues I listed talks about why we have a workaround with GVK and not using that from the type directly.

keeganwitt:
> 2. see's if the ownerRef is in the ownerReferences

This is the part that isn't just a generic Go slices thing. The definition of whether two references are the same is logic that is defined in this package. But I'd agree that the rest of the functionality of these functions isn't specific to owner references.

> That's what I was asking around sidestepping the logic that exists to check that with the original functions. The two issues I listed talks about why we have a workaround with GVK and not using that from the type directly.

Any suggestions for an alternative? If I want to use the existing functions, I have to do something like

```go
var owner ctrlClient.Object
switch ownerReference.Kind {
case "ReplicaSet":
	owner = &appsv1.ReplicaSet{}
case "StatefulSet":
	owner = &appsv1.StatefulSet{}
case "DaemonSet":
	owner = &appsv1.DaemonSet{}
case "Job":
	owner = &batchv1.Job{}
default:
	return fmt.Errorf("unexpected owner reference kind %v", ownerReference.Kind)
}
c.client.Get(ctx, ctrlClient.ObjectKey{Namespace: c.pod.Namespace, Name: ownerReference.Name}, owner)
controllerutil.SetOwnerReference(owner, object, c.scheme)
```

And then hope I've accounted for all the kinds that might create a pod.

troy0820:
> This is the part that isn't just a generic Go slices thing. The definition of whether two references are the same is logic that is defined in this package

You skip over that with the function you provide. You provided a function that exposes  `upsertOwnerRef` which just does what @alvaroaleman suggested. 

```go
var owner ctrlClient.Object
switch ownerReference.Kind {
case "ReplicaSet":
	owner = &appsv1.ReplicaSet{}
case "StatefulSet":
	owner = &appsv1.StatefulSet{}
case "DaemonSet":
	owner = &appsv1.DaemonSet{}
case "Job":
	owner = &batchv1.Job{}
default:
	return fmt.Errorf("unexpected owner reference kind %v", ownerReference.Kind)
}
c.client.Get(ctx, ctrlClient.ObjectKey{Namespace: c.pod.Namespace, Name: ownerReference.Name}, owner)
controllerutil.SetOwnerReference(owner, object, c.scheme)
```
> And then hope I've accounted for all the kinds that might create a pod.

I wouldn't expect that the `controllerutil` package do this as it is targeting specific kinds and not generalize as the package is currently.    These things create pods but that doesn't mean I can't create a resource specific to my domain that creates a pod and be left with trying to extend this logic.

keeganwitt:
> I wouldn't expect that the controllerutil package do this as it is targeting specific kinds and not generalize as the package is currently. These things create pods but that doesn't mean I can't create a resource specific to my domain that creates a pod and be left with trying to extend this logic.

I was talking about in my controller, not controllerutil.

keeganwitt:
> You skip over that with the function you provide. You provided a function that exposes upsertOwnerRef which just does what @alvaroaleman suggested.

I was suggesting that we add a public version of `indexOwnerRef` instead of what I did in this PR.

PR #2997, opened 2024-10-26
:bug: Error when source.Start() never returns

alvaroaleman:
Contrary to everything else in controller-runtime, we expect `source.Start` to be non-blocking. If someone implements a custom source and gets this wrong, the resulting behavior is that the binary starts successfully, but no reconciliation happens which is extremely difficult to understand and debug.

This change makes us use the `CacheSyncTimeout` not only for the sources `WaitForSync` but also for its `Start`.

It is worth noting that the current design of both requiring `Start` to not block and `WaitForSync` to block is very confusing. It likely came to be because we basicaly require two distinct contexsts in `Start`, one to indicate the lifetime of the `Source` and one to indicate the `Start` timeout.

To overall simplify and improve the code, the change also parallelizes the `Start` of the sources.

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -21,9 +21,11 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-logr/logr"
+	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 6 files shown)


pkg/internal/controller/controller.go
```diff
@@ -171,41 +173,55 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 		// NB(directxman12): launch the sources *before* trying to wait for the
 		// caches to sync so that they have a chance to register their intendeded
 		// caches.
+		errGroup, _ := errgroup.WithContext(ctx)
```
troy0820, line 176: nit: If you don't need the context, can you just use:
```suggestion
		errGroup := new(errgroup.Group)
```

troy0820:
Wasn't sure this would merge so quickly or else I would've put a hold on it to allow others to review.  Do you want to cherry-pick this @alvaroaleman?

alvaroaleman:
> Wasn't sure this would merge so quickly or else I would've put a hold on it to allow others to review. Do you want to cherry-pick this @alvaroaleman?

No. No problem, it happens. I will do a follow-up regarding your comment and in case there are more comments

alvaroaleman:
https://github.com/kubernetes-sigs/controller-runtime/pull/3006

pkg/internal/controller/controller_test.go
```diff
@@ -200,27 +200,40 @@ var _ = Describe("controller", func() {
 			<-sourceSynced
 		})
 
-		It("should not error when cache sync timeout is of sufficiently high", func() {
-			ctrl.CacheSyncTimeout = 1 * time.Second
+		It("should error when Start() is blocking forever", func() {
+			ctrl.CacheSyncTimeout = 0
```
sbueringer, line 204: Should we maybe use something slightly higher? 

I think if we use 0 this test could also simply work because the timeout occurs before the go routine had a chance to run

pkg/internal/controller/controller_test.go
```diff
@@ -281,24 +296,26 @@ var _ = Describe("controller", func() {
...
 		It("should call Start on sources with the appropriate EventHandler, Queue, and Predicates", func() {
+			ctrl.CacheSyncTimeout = 10 * time.Second
 			started := false
+			ctx, cancel := context.WithCancel(context.Background())
 			src := source.Func(func(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
 				defer GinkgoRecover()
 				Expect(q).To(Equal(ctrl.Queue))
 
 				started = true
+				cancel()
```
sbueringer, line 307: Maybe let's add a short comment that we have to cancel the context here because otherwise ctrl.Start blocks / doesn't return after starting all the sources

Issue #868, opened 2020-03-20
Webhook marker allows handler path to be changed, but a deterministic handler is always registered


rajathagasthya:
Webhook marker like the one below lets us change `path` where webhook is served. This creates the appropriate `Mutating|ValidatingWebhookConfiguration` manifest referencing this `path`.
```
// +kubebuilder:webhook:verbs=create;update,path=/validate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=false,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,versions=v1,name=vcronjob.kb.io
```

But webhook builder always registers the same path based on GVK: https://github.com/kubernetes-sigs/controller-runtime/blob/29e923ac474ca94c021a6f8b0239a319d8b05221/pkg/builder/webhook.go#L158-L166

This is confusing when the user thinks they can customize the handler path, but it's actually ignored and worse, the generated webhook manifest doesn't work because path is incorrect. 

If we are letting users change the path through the marker, perhaps it's better to add an option to webhook builder to configure handler path as well? Or, if the path is not supposed to be changed in the marker, we should say that in docs.

alenkacz:
maybe this bug should rather be in kubebuilder project? Since it could be solved on multiple places and might also need update in documentation?

Anyway - I took a look at it. One possible solution would be to allow overriding path inside the builder here https://github.com/kubernetes-sigs/controller-runtime/blob/dc8357113a904bf02721efcde5d92937be39031c/pkg/builder/webhook.go#L38 and use that instead of generating one if provided. If we go that way, we would also have to update docs here https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html

WDYT?

rajathagasthya:
@alenkacz Yeah, we could provide an option in webhook builder to set path and then update docs to say "path in webhook build must match path in webhook marker" or something to that effect.

We probably also need some kind of validation if we let users set their own path.

ryanzhang-oss:
FYI,   https://github.com/kubernetes-sigs/kubebuilder/issues/1436

ryanzhang-oss:
> maybe this bug should rather be in kubebuilder project

https://github.com/kubernetes-sigs/kubebuilder/issues/1436

vincepri:
/kind design

We need to think more about possible solutions

ryanzhang-oss:
[...]

PR #2998, opened 2024-10-28
✨ Add custom path option for webhooks

damsien:
This PR adds a way to configure mutating and validating webhooks using a custom path (for the HTTP call). It answer a problem described in [this](https://github.com/kubernetes-sigs/controller-runtime/issues/868) issue.

By default, the `WebhookBuilder` will generates the path using the `generateMutatePath(gvk)`/`generateValidatePath(gvk)` function. But if the user wants to use a custom path, it can build the `WebhookBuilder` using the `WithCustomPath(customPath)` function.

Example:
```go
WebhookManagedBy(mgr).
	For(&TestDefaulter{}).
	WithDefaulter(&TestCustomDefaulter{}).
	WithCustomPath(customPath).
	Complete()
```

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -39,6 +40,7 @@ type WebhookBuilder struct {
 	apiType         runtime.Object
 	customDefaulter admission.CustomDefaulter
 	customValidator admission.CustomValidator
+	customPath      string
 	gvk             schema.GroupVersionKind
 	mgr             manager.Manager
 	config          *rest.Config
@@ -87,6 +89,12 @@ func (blder *WebhookBuilder) WithLogConstructor(logConstructor func(base logr.Lo
 // Defaults to true.
 func (blder *WebhookBuilder) RecoverPanic(recoverPanic bool) *WebhookBuilder {
 	blder.recoverPanic = &recoverPanic
+	return blder
+}
+
+// WithCustomPath overrides the webhook's default path by the customPath
+func (blder *WebhookBuilder) WithCustomPath(customPath string) *WebhookBuilder {
+	blder.customPath = customPath
 	return blder
 }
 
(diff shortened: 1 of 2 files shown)


pkg/builder/webhook.go
```diff
@@ -251,3 +270,11 @@ func generateValidatePath(gvk schema.GroupVersionKind) string {
 	return "/validate-" + strings.ReplaceAll(gvk.Group, ".", "-") + "-" +
 		gvk.Version + "-" + strings.ToLower(gvk.Kind)
 }
+
+func generateCustomPath(customPath string) string {
```
alvaroaleman, line 274: I don't think silently ignoring or changing the path that was passed in is a good idea, we should instead return an error.

Also, please compile the regex at the package level rather than inside this func.

alvaroaleman (commented):
Idea makes a lot of sense, one comment

damsien:
@alvaroaleman Done :heavy_check_mark:

pkg/builder/webhook.go
```diff
@@ -251,3 +285,14 @@ func generateValidatePath(gvk schema.GroupVersionKind) string {
...
+
+const webhookPathStringValidation = `^((/[a-zA-Z0-9-_]+)+|/)$`
+
+var validWebhookPathRegex = regexp.MustCompile(webhookPathStringValidation)
+
+func generateCustomPath(customPath string) (string, error) {
+	if !validWebhookPathRegex.MatchString(customPath) {
+		return "", errors.New("customPath \"" + customPath + "\" does not match this regex: " + webhookPathStringValidation)
+	}
+	return path.Clean(customPath), nil
```
alvaroaleman, line 297: This alters the path or not? Lets not do that, people might refernce the path in their Validating/MutatingWebhookConfiguration

damsien: @alvaroaleman if you're asking about the `path.Clean()` in the return section, it does not alter the path, it on makes it cleaners. For eg.

"//test" -> "/test"

So you want me to just return the `customPath`?

alvaroaleman: > So you want me to just return the customPath?

I think so. I had a look at this and this is actually meant for filepaths, not for URLs. We are not dealing with untrusted input here, the double slash case you mentioned is probably confusing but not harmful. I think if for any reason we think the path we were given is wrong, we should return an error, otherwise the person who authors such code might not realize there is a problem

damsien: Okay I understand! Just removed it!

alvaroaleman (approved):
Thanks!

PR #3002, opened 2024-11-04
:book: clarify comments of GenerationChangedPredicate

loveRhythm1990:
comments of GenerationChangedPredicate is confusing, we know that `generation` is not updated if its status is defined as subresource.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -173,7 +173,8 @@ func (TypedResourceVersionChangedPredicate[T]) Update(e event.TypedUpdateEvent[T
 // The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
 // This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
 //
-// For CustomResource objects the Generation is only incremented when the status subresource is not enabled.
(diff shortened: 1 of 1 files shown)


pkg/predicate/predicate.go
```diff
@@ -191,7 +191,7 @@ type GenerationChangedPredicate = TypedGenerationChangedPredicate[client.Object]
 // The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
 // This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
 //
-// For CustomResource objects the Generation is only incremented when the status subresource is enabled.
+// For CustomResource objects the Generation is only incremented when the status subresource is not enabled.
```
sbueringer, line 194: This now sounds like the generation is not incremented at all as soon as the status sub resource is enabled
(also wasn't great before)

loveRhythm1990: how about `For CustomResource objects the Generation is incremented when spec is changed, or status changed and status not modeled as subresource. subresource status update will not increase Generation.`? I put it in second commit.

sbueringer:
Thx!

/lgtm
/approve
/ok-to-test

PR #3008, opened 2024-11-10
:seedling: Fixups to error when source.Start never returns

alvaroaleman:
Addresses the comments on https://github.com/kubernetes-sigs/controller-runtime/pull/2997

/assign @sbueringer 
/hold

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -171,14 +171,15 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 		defer utilruntime.HandleCrash()
 
 		// NB(directxman12): launch the sources *before* trying to wait for the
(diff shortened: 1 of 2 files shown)


pkg/internal/controller/controller.go
```diff
@@ -220,7 +221,6 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 			return err
 		}
 
-		// Start the SharedIndexInformer factories to begin populating the SharedIndexInformer caches
```
alvaroaleman, line 223: You mentioned that the comment is in the wrong place and I ended up removing it entirely, as a source doesn't really necessitate a SharedIndexInformer at this point and I found it to be mostly just confusing

sbueringer: Absolutely makes sense!

PR #3012, opened 2024-11-13
:seedling: Switch to Go 1.23+ stdlib `maps`/`slices` packages

tklauser:
Use the `maps` and `slices` packages from the Go standard library instead of the respective `golang.org/x/exp` packages. Also see https://go.dev/doc/go1.23#iterators.


alvaroaleman (approved):
Thanks!

Issue #2374, opened 2023-06-12
Implement priority queue as an option for reconcile loops


fabriziopandini:
While discussing https://github.com/kubernetes-sigs/controller-runtime/issues/857#issuecomment-1587149384 an idea come up about having something like a priority queue, so we can give resync events a lower priority while assigning to other events a higher priority.

This could help when there are many objects of the same type and at every resync period there is a storm of events being added to the queue.

cc @alvaroaleman @sbueringer

alvaroaleman:
While resync events are one example, I think that it makes sense in general to allow specifying a priority when adding items to the workqueue.

One challenge here is the fact that as of today, there is no priority queue implementation in upstream kube.

alvaroaleman:
All,

I've written a small [design doc](https://github.com/kubernetes-sigs/controller-runtime/pull/3013) and [POC](https://github.com/kubernetes-sigs/controller-runtime/pull/3014) for this. Would appreciate everyones feedback

sbueringer:
I've updated the issue so we can use it as an umbrella issue for the feature going forward

lingdie:
Can we provide more example codes on priority queue feature?

lingdie:
Is that all? Can't believe. :)

sbueringer:
Yup. Literally just that. We try to implement this in a way that it just works

vishalbalaji-v:
Hi everyone,

Is there timeline on when this will be generally available? The doc for priority queue says it's currently in beta ([link](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controller.go#L91-L95)). 

We have a use case for it, but people are worried about putting it in production since the doc says it's in beta. I'd be happy to contribute to something if required. The [design](https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/priorityqueue.md) also doesn't seem to have been updated in a while.

alvaroaleman:
> Is there timeline on when this will be generally available? The doc for priority queue says it's currently in beta ([link](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controller.go#L91-L95)).

The main reason for that is that we don't want to call it GA until we are sure we don't want to change the interface definition. The priority queue is being used in production in places where it matters and there are no known issues with it.

sbueringer:
And PQ is enabled per default 🎉 

Thx to everyone that is contributing to this feature!

[...]

PR #3013, opened 2024-11-17
:book: Add a design for a priority queue

alvaroaleman:
This change describes the motivation and implementation details for a priority queue in controller-runtime.

Ref https://github.com/kubernetes-sigs/controller-runtime/issues/2374


designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+
+```go
+type PriorityQueue[T comparable] interface {
+    // AddWithOpts adds one or more items to the workqueue. Items
+    // in the workqueue are de-duplicated, so there will only ever
+    // be one entry for a given key.
+    // Adding an item that is already there may update its wait
+    // period to the lowest of existing and new wait period or
+    // its priority to the highest of existing and new priority.
+	AddWithOpts(o AddOpts, items ...T)
```
alvaroaleman, line 31: The main alternate to this would be to have `AddWithPriority`, `AddAfterWithPriority` `AddRatelimitedWithPriority` etc - I found this easier but not a strong opinion either way.

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+// or cache resyncs.
+func WithLowPriorityWhenUnchanged[object client.Object, request comparable](u TypedEventHandler[object, request]) TypedEventHandler[object, request]{
+}
+```
+
+The issue described in point two of the motivation section ("application-specific
+reasons to prioritize some events") will always require implementation of a custom
+handler or eventsource in order to inject the appropriate priority.
+
+## Implementation stages
```
Airblader, line 83: Will the default controller be modified to make use of this new queue (if used), or will it rely on using a custom controller implementation?

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+    // workqueue, the lowest of existing and new wait period will be
+    // used.
+    RateLimited bool
+
+    // Priority specifies the priority of the object. Objects with higher
+    // priority are returned before objects with lower priority. If the
+    // object is already in the workqueue, the priority will be updated
+    // to the highest of existing and new priority.
+    Priority    int
+}
```
Airblader, line 60: I’m wondering if we need to consider starvation scenarios whether some item is never retrieved because there’s always something of higher priority in the queue. It might be fair to say this is “user error”, but is there a way to detect / avoid / recover?

designs/priorityqueue.md
alvaroaleman: Yes it will be updated to make use of it - the only thing it needs is to re-use the priority though. Once we make it the default, I would also like to add a Priority parameter to the `reconcile.Result` - But really only once its enabled by default.

designs/priorityqueue.md
alvaroaleman: Interesting point. I don't think avoid or recover, but maybe detect. I'll think about it

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+In order to fix the issue described in point one of the motivation section,
+we have to be able to differentiate events stemming from the initial list
+during startup and from resyncs from other events. In both these cases, the
+informer emits an artifical create. The suggestion is to use a heuristic that
```
sbueringer, line 66: I think resyncs are updates (without resourceVersion changes)

Maybe we could make this better mid-term by flagging these events explicitly in client-go

designs/priorityqueue.md
sbueringer: Maybe also to the Request?

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+    // determine a wait period. If the object is already in the
+    // workqueue, the lowest of existing and new wait period will be
+    // used.
+    RateLimited bool
+
+    // Priority specifies the priority of the object. Objects with higher
+    // priority are returned before objects with lower priority. If the
+    // object is already in the workqueue, the priority will be updated
+    // to the highest of existing and new priority.
+    Priority    int
```
sbueringer, line 59: I assume valid values are from min to max int. What values is CR going to use? (for inital sync, resyncs, everything else) 
Let's define constants for these

sbueringer (commented):
Overall lgtm. Looking forward to using this 😀

designs/priorityqueue.md
sbueringer: Definitely a lot easier to extend without breaking changes

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+
+```go
+type PriorityQueue[T comparable] interface {
+    // AddWithOpts adds one or more items to the workqueue. Items
+    // in the workqueue are de-duplicated, so there will only ever
+    // be one entry for a given key.
+    // Adding an item that is already there may update its wait
+    // period to the lowest of existing and new wait period or
+    // its priority to the highest of existing and new priority.
+    AddWithOpts(o AddOpts, items ...T)
```
sbueringer, line 31: Do we have to be able to enqueue multiple items at once? Otherwise (item, opts) would be more consistent with the embedded interface (no strong opinion from my side)

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+
+    // GetWithPriority returns an item and its priority. It allows
+    // a controller to re-use the priority if it enqueues an item
+    // again.
+    GetWithPriority() (item T, priority int, shutdown bool)
```
sbueringer, line 36: What's the story for the shutdown return parameter? :)

designs/priorityqueue.md
```diff
@@ -0,0 +1,100 @@
...
+
+    // workqueue.TypedRateLimitingInterface is kept for backwards
+    // compatibility.
+    workqueue.TypedRateLimitingInterface[T]
+}
+
+type AddOpts struct {
+    // After is a duration after which the object will be available for
+    // reconciliation. If the object is already in the workqueue, the
+    // lowest of existing and new After period will be used.
```
sbueringer, line 46: I assume the "existing period" is a timestamp not a duration?

designs/priorityqueue.md
alvaroaleman: > Maybe also to the Request?

Ah, interesting thought. Is the idea because you want to be able to set the priority in a handler?

The way this currently works is that the workqueue doesn't have any understanding of the request object (and we should keep it that way IMHO). We could probably provide a thin wrapper for it that will use the priority from the request if the `AddWithOptions` call doesn't have one yet and inject that in the builder if its typed to `reconcile.Request`?

designs/priorityqueue.md
alvaroaleman: > I think resyncs are updates (without resourceVersion changes)

You are right, fascinating, TILd. Will update

> Maybe we could make this better mid-term by flagging these events explicitly in client-go

That would make sense but I think it might be difficult due to how client-go is factored (But would love to be wrong on this).

designs/priorityqueue.md
alvaroaleman: The default is the zero value/`0`, as that allows to easily define both higher and lower than default priorities. I've added go docs for that and a constant in the `handler` for the priority it uses. Do you think that is enough?

designs/priorityqueue.md
alvaroaleman: Yeah we do have this use-case and currently do it by filtering before adding items to the workqueue: https://github.com/kubernetes-sigs/controller-runtime/blob/e3347b5405cdd0da5bff527af3d406117938ba6b/pkg/handler/enqueue_mapped.go#L93-L95

designs/priorityqueue.md
alvaroaleman: I just copied it from the existing `Get`, we use it to exit the controller: https://github.com/kubernetes-sigs/controller-runtime/blob/e3347b5405cdd0da5bff527af3d406117938ba6b/pkg/internal/controller/controller.go#L263-L265

designs/priorityqueue.md
alvaroaleman: Internally in the workqueue it ends up being a timestamp yeah, but the parameter is always a duration - I thought it would be less confusing to consistently refer to the type present in the external api regardless of what is then internally done with that. WDYT?

designs/priorityqueue.md
sbueringer: I was thinking that the priority could also be useful information for the Reconcile func. But maybe it's a bad idea not sure 😀

designs/priorityqueue.md
sbueringer: Queue depth is definitely good enough in general to detect that the queue cannot be drained. Isn't there also a metric for the longest time something is in the queue?

Independent of that we could also consider increasing priority after each resync, so it eventually reaches or maybe even exceeds (not sure) the priority of regular events

designs/priorityqueue.md
sbueringer: Sounds good

designs/priorityqueue.md
sbueringer: I think that's fine.

designs/priorityqueue.md
sbueringer: Fine for me!

designs/priorityqueue.md
sbueringer: That's one use case. Another one would be to reconcile Requests of different priorities differently :) (not sure if allowing this is a good thing or not though)

designs/priorityqueue.md
sbueringer: Makes sense. Probably worth extending the godoc of this func to explain it a bit (~ the same comment as on Typed.Get)

designs/priorityqueue.md
akutz: I have a clear use case for it. I just opened https://github.com/vmware-tanzu/vm-operator/pull/1260, but ideally I would not base this on the source, but rather the thing enqueuing the request. That way the priority can be decided more discreetly.

Also, hey @vincepri!!

alvaroaleman: @akutz could you open a new issue for that, please?

PR #3016, opened 2024-11-20
🌱 (:seedling:, other) Log the syncing source when WaitForSync fails

haiyanmeng:
A controller may watch several sources. This change makes it easier to debug which source fails to be synced.


alvaroaleman (approved):
Thanks!

PR #3017, opened 2024-11-20
✨ Add EnableWatchBookmarks option to cache informers

vincepri:
AllowWatchBookmarks is generally pretty safe to enable as it has been available in Kuberentes for a long while, and the server ignores the flag if it doesn't implement it (per docs).

### DIFF
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -51,6 +51,7 @@ type InformersOpts struct {
 	Selector              Selector
 	Transform             cache.TransformFunc
 	UnsafeDisableDeepCopy bool
+	EnableWatchBookmarks  bool
 	WatchErrorHandler     cache.WatchErrorHandler
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cache/internal/informers.go
```diff
@@ -444,6 +446,9 @@ func (ip *Informers) makeListWatcher(gvk schema.GroupVersionKind, obj runtime.Ob
 			},
 			// Setup the watch function
 			WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
+				opts.Watch = true // Watch needs to be set to true separately
```
alvaroaleman, line 449: Why is this needed now?

alvaroaleman (commented):
Maybe add an opt-out option just in case?

pkg/cache/internal/informers.go
vincepri: I just added it because I noticed it was in the other one, not sure if it is actually needed?

alvaroaleman (approved):

alvaroaleman (approved):

PR #3019, opened 2024-11-22
✨ WIP: Cluster provider and cluster-aware controllers

embik:
This is a continuation of #2726 and #2207, a prototype for a multi-cluster support in controller-runtime. A design proposal is at #2746 which will need updating depending on the feedback for this PR. I've reorganized code changes into different commits than the previous attempts, but tried to keep authorship of changes faithful to the commits this work is based on.

The main change from the previous PR (#2726) is the update to typed generics support and adjustment to have a BYO request type. The implications of that is a) that you will need to bring your own typed EventHandler and b) that a wrapper function is needed to inject information (mostly the cluster name) when establishing a handler.

Apart from that, the core design of this is the cluster provider that can be plugged into a manager. This unlocks different ways to discover the Kubernetes "fleet" that is being reconciled.


pkg/builder/controller.go
```diff
@@ -232,6 +249,12 @@ func (blder *TypedBuilder[request]) WithEventFilter(p predicate.Predicate) *Type
 	return blder
 }
 
+// WithHandlerWrapper sets an optional wrapper function that is applied to the TypedEventHandler for watches set up via the Watches func.
+func (blder *TypedBuilder[request]) WithHandlerWrapper(handlerWrapper func(handler.TypedEventHandler[client.Object, request], cluster.Cluster) handler.TypedEventHandler[client.Object, request]) *TypedBuilder[request] {
+	blder.handlerWrapper = handlerWrapper
+	return blder
+}
```
embik, line 256: A note on this: I would have preferred for this to be an option that can be passed to `Watches`, but the main challenge there was the `untypedWatchesInput` interface not allowing to have a generic type for `request` (since it's untyped). I'm still very much open to making it an opt, but I haven't found a good way to adjust the existing interface without breaking existing usage.

pkg/cluster/cluster.go
```diff
@@ -37,6 +37,11 @@ import (
 
 // Cluster provides various methods to interact with a cluster.
 type Cluster interface {
+	// Name returns the name of the cluster. It identifies the cluster in the
+	// manager if that is attached to a cluster provider. The value is usually
+	// empty for the default cluster of a manager.
+	Name() string
```
sprsquish, line 43: We're internally working on a mechanism for dealing with multiple clusters, too. It's been helpful for the cluster identifier to be more flexible than a simple string. We've made it comparable. 

Something like this:
```go
type Cluster[ID comparable] interface {
  Identifier() ID
  ...
}
```

I understand this suggestion would cause a much larger change requiring type parameters in many places. Future work could be done to migrate.

Mostly, I wanted to raise this as something to consider and that calling the method `Identifier` would provide room to migrate in the future.

embik: I'm definitely open to that, I think renaming it to `Identifier()` should be easy enough! Maybe some maintainers can leave a note if they consider this a future endeavour and this would be a worthwhile change to prepare that.

sttts: We were there before, had a logical cluster abstraction, and then went back to string as it just simplifies everything.

@sprsquish Am curious why a string as identifier and some lookup table on the cluster provider side does not work?

sttts: Also, adding `Name() string` to the interface has no impact on the consumers of the `Cluster` type. But adding a type parameter forces them to pass e.g. string or whatever everywhere including the manager. I don't think that's a good idea. IMO, we should not do it.

sprsquish: @sttts 
> Am curious why a string as identifier and some lookup table on the cluster provider side does not work?

Strings work fine.. We like to work with more structure in our identifiers. For instance, working with multiple clusters across cloud providers it's nice to have a structure like this:

```go
type ClusterID struct {
  Provider string
  Region string
  Account string
  Name string
}
```

It can be marshaled into a string, of course, it's just nice not to have to.

I understand the need for a simpler approach and appreciate that this has been discussed already. I think there's a way to migrate over time that follows what's been happening with handlers by introducing `TypedCluster[ID comparable]` and an alias for the common use case `type Cluster = TypedCluster[string]` (Manager would need the same treatment). 

The only change to the current proposal would be to use the more generic `Identifier()` rather than `Name()`. That leaves the door open for future discussions regarding generic identifiers without having to do all the work up front.

sttts: If I understand your use-case correctly, this pattern should help you elegantly: https://goplay.space/#wRWyxv-GUrj

Or in other words: the type only matters in code that is `ClusterID` aware. Turning the string back into the rich format is one type conversion away.

pkg/cluster/multicluster.go
```diff
@@ -0,0 +1,55 @@
...
+// Provider defines methods to retrieve clusters by name. The provider is
+// responsible for discovering and managing the lifecycle of each cluster.
+//
+// Example: A Cluster API provider would be responsible for discovering and
+// managing clusters that are backed by Cluster API resources, which can live
+// in multiple namespaces in a single management cluster.
+type Provider interface {
+	// Get returns a cluster for the given identifying cluster name. Get
+	// returns an existing cluster if it has been created before.
+	Get(ctx context.Context, clusterName string) (Cluster, error)
```
sttts, line 54: is the default cluster a thing here already? I.e. can I always pass empty string?

pkg/cluster/multicluster.go
```diff
@@ -0,0 +1,55 @@
...
+	//               `--------'
+	Engage(context.Context, Cluster) error
+
+	// Disengage gets called when the component should stop operations for the given Cluster.
+	Disengage(context.Context, Cluster) error
+}
+
+// Provider defines methods to retrieve clusters by name. The provider is
+// responsible for discovering and managing the lifecycle of each cluster.
+//
```
sttts, line 47: "defines methods" is redundant for an interface.

```suggestion
// Provider allows to retrieve clusters by name. The provider is
// responsible for discovering and managing the lifecycle of each cluster.
//
```
It's kind of unclear what managing means in a getter interface.

pkg/cluster/multicluster.go
sttts: Should we specify what is returned for an unknown cluster?

pkg/cluster/multicluster.go
sttts: Maybe be concrete saying:
```suggestion
// Provider allows to retrieve clusters by name, e.g. to reconcilers. 
// The provider is responsible for discovering and managing the lifecycle
// of each cluster, calling `Engage` and `Disengage` on the manager
// it is run against, whenever a new cluster is discovered or a cluster 
// is unregistered.
```

pkg/cluster/multicluster.go
embik: The default cluster is not a thing here, that is in `Manager.GetCluster`. The cluster provider doesn't have a concept of a default cluster.

> Should we specify what is returned for an unknown cluster?

Probably. I can add a sentence to clarify this.

pkg/cluster/multicluster.go
embik: +1, I've added that.

pkg/cluster/multicluster.go
embik: Added a note on what should happen if the cluster is unknown.

pkg/manager/internal.go
```diff
@@ -68,8 +70,14 @@ type controllerManager struct {
...
 
-	// cluster holds a variety of methods to interact with a cluster. Required.
-	cluster cluster.Cluster
+	// defaultCluster holds a variety of methods to interact with a defaultCluster. Required.
+	defaultCluster        cluster.Cluster
+	defaultClusterOptions cluster.Option
+
+	// engagedCluster is a map of engaged clusters. The can come and go as the manager is running.
+	engagedClustersLock   sync.RWMutex
+	engagedClusters       map[string]cluster.Cluster
```
varshaprasad96, line 79: It would be super helpful if we can expose the set of engagedClusters from the manager. In [Kueue](https://github.com/kubernetes-sigs/kueue/tree/main) for a multi cluster implementation, we have a list of [active and missing clusters](https://github.com/kubernetes-sigs/kueue/blob/404a0cd659b2331b232cf5ddf93a1b67d5cc146c/pkg/controller/admissionchecks/multikueue/admissioncheck.go#L86-L87) based on available quota for admitting workloads.

pkg/manager/internal.go
```diff
@@ -614,6 +676,70 @@ func (cm *controllerManager) initLeaderElector() (*leaderelection.LeaderElector,
...
+	return nil
+}
+
+func (cm *controllerManager) Disengage(ctx context.Context, cl cluster.Cluster) error {
+	cm.Lock()
+	defer cm.Unlock()
+
+	var errs []error
+	for _, r := range cm.clusterAwareRunnables {
+		if err := r.Disengage(ctx, cl); err != nil {
```
varshaprasad96, line 731: Please correct me I'm wrong - While calling `Engage` we are starting a SyncingSource from multicluster cache (https://github.com/kubernetes-sigs/controller-runtime/blob/e93979ec5cdd656809f5fbb91b858665979cc040/pkg/builder/controller.go#L397), but in disengage we aren't stopping the context anywhere in here: https://github.com/kubernetes-sigs/controller-runtime/blob/e93979ec5cdd656809f5fbb91b858665979cc040/pkg/controller/multicluster.go#L126. Is this expected?

varshaprasad96 (commented):
In the process of reviewing the PR in detail, for now most of my comments are either nits or questions. 

This is a really helpful feature in controller-runtime, that can benefit other internal kubernetes-sigs projects. 
cc: @sbueringer @joelanford Could we have some eyes on this and the proposal please. Thank you!

pkg/builder/controller.go
```diff
@@ -279,59 +295,57 @@ func (blder *TypedBuilder[request]) Build(r reconcile.TypedReconciler[request])
...
+		obj, err := project(cl, cc.forInput.object, cc.forInput.objectProjection)
 		if err != nil {
 			return err
 		}
 
-		if reflect.TypeFor[request]() != reflect.TypeOf(reconcile.Request{}) {
-			return fmt.Errorf("For() can only be used with reconcile.Request, got %T", *new(request))
+		var hdler handler.TypedEventHandler[client.Object, request]
+		switch reflect.TypeFor[request]() {
+		case reflect.TypeOf(reconcile.Request{}):
```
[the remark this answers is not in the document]
embik: Good point, but I tried to just adapt the previous if condition. I'm not opposed to changing it, I just wanted to not fiddle with the condition checked.

pkg/builder/controller.go
```diff
@@ -279,59 +295,57 @@ func (blder *TypedBuilder[request]) Build(r reconcile.TypedReconciler[request])
...
+			reflect.ValueOf(&hdler).Elem().Set(reflect.ValueOf(handler.WithLowPriorityWhenUnchanged(&handler.EnqueueRequestForObject{})))
+		case reflect.TypeOf(reconcile.ClusterAwareRequest{}):
+			reflect.ValueOf(&hdler).Elem().Set(reflect.ValueOf(handler.WithLowPriorityWhenUnchanged(&handler.EnqueueClusterAwareRequestForObject{ClusterName: cl.Name()})))
```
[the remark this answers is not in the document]
embik: I've asked myself the same question when I stumbled over this part of the code (see below, it's how the code today sets the handler implementation as well). When I try to change it to something like:

```go
case reflect.TypeOf(reconcile.ClusterAwareRequest{}):
    hdler = handler.WithLowPriorityWhenUnchanged(&handler.EnqueueClusterAwareRequestForObject{ClusterName: cl.Name()})
```

it fails with:

```sh
../../pkg/builder/controller.go:335:12: cannot use handler.WithLowPriorityWhenUnchanged(&handler.EnqueueClusterAwareRequestForObject{…}) (value of type handler.TypedEventHandler[client.Object, reconcile.ClusterAwareRequest]) as handler.TypedEventHandler[client.Object, request] value in assignment: handler.TypedEventHandler[client.Object, reconcile.ClusterAwareRequest] does not implement handler.TypedEventHandler[client.Object, request] (wrong type for method Create)
		have Create(context.Context, event.TypedCreateEvent[client.Object], workqueue.TypedRateLimitingInterface[reconcile.ClusterAwareRequest])
		want Create(context.Context, event.TypedCreateEvent[client.Object], workqueue.TypedRateLimitingInterface[request])
```

My understanding of generics here is a bit limited, but I think this is because we expect `TypedRateLimitingInterface` to take whatever generic `request` we give into the function, but we are setting it to a concrete type. I suppose this only works with the reflect magic we currently have.

pkg/manager/internal.go
embik: For my understanding, you'd be looking for a list of the keys for that map, is that correct (i.e. the cluster identifiers, not the `cluster.Cluster` objects directly)?

pkg/builder/controller.go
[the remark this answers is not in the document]
varshaprasad96: Ah, thanks for explaining @embik!

pkg/manager/internal.go
varshaprasad96: Yes, a unique identifier for the cluster, so that we can fetch the relevant client and submit workloads in that cluster.

pkg/manager/internal.go
embik: The way this works (but I wonder if it's confusing and should be changed to be honest) is that the implementation expects the *cluster provider* to cancel the `clusterCtx` used for `Engage` (and subsequently, the `ctxBoundedSyncingSource`) when shutting down. Compare the `fleet` example here canceling the stored context when figuring out that a cluster has to be removed: https://github.com/embik/controller-runtime/blob/e93979ec5cdd656809f5fbb91b858665979cc040/examples/fleet/main.go#L280

The `ctx` passed to `Disengage` is a different context, one that is currently unused but could be used when disengaging should be cancelable.

I'm open to input here though. I was wondering if the manager should even have a `Disengage` (and instead clean up its internal cluster list when the respective context gets canceled), but then the manager itself wouldn't be a `cluster.Aware`.

pkg/builder/controller.go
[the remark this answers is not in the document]
embik: No problem! This is at least how I understand this line. It's a bit unclear to me too.

pkg/config/controller.go
```diff
@@ -59,4 +59,16 @@ type Controller struct {
...
 	UsePriorityQueue *bool
+
+	// EngageWithDefaultCluster indicates whether the controller should engage
+	// with the default cluster. This default to false if a cluster provider
+	// is configured, and to true otherwise.
+	//
+	// This is an experimental feature and is subject to change.
+	EngageWithDefaultCluster *bool
+
+	// EngageWithProvidedClusters indicates whether the controller should engage
```
sttts, line 70:
```suggestion
	// EngageWithProviderClusters indicates whether the controller should engage
```

embik:
As per https://github.com/kubernetes-sigs/controller-runtime/pull/2746#issuecomment-2661547096, we're focussing on getting the design presented in this PR up and running in https://github.com/multicluster-runtime/multicluster-runtime. Feel free to follow along on the new project if you are interested.

PR #3020, opened 2024-11-22
🐛 Refactor certificate watcher to use polling, instead of fsnotify

m-messiah:
### Summary

In the current controller-runtime/certwatcher behaviour if any of the files (cert or key) are deleted before the replacement, the code removes the watcher. It gets stuck with watching nothing, so when the replacement arrives - no goroutine will receive that event and all the future certificate updates (including this) are missed until the code is restarted.

### Details

The added test in the PR shows the example scenario of updating the certificate in two steps:
1. `rm /certs/mtls.pem /certs/mtls.key`
2. `tar xvf -C /certs new_mtls.tar.gz` (or similar).

**Expected behaviour:**
Return the previous certificate until the new files appear, reestablish the watch and reread the certs when ready.

**Current behaviour:**
The watch is removed and new files do not trigger anything.

### Possible options to consider

1. The `ensureAllFilesAreWatched` could be called in `select/default` branch, but it increases the load drastically. 1 second seems reasonable enough.
2. The ticker could be increased to 10 seconds to reduce load (or be customizable), but don't see a big issue here as it checks the filesystem and its internal locks only.
3. Instead of reestablishing watches we could stop the watcher and trigger controller code to be reinitialised from the beginning - it is a nice behaviour we use for Kube caches and watchers, but does not seem good for certificates to reinitialise the whole context after such small change.

### DIFF
--- a/pkg/certwatcher/certwatcher.go
+++ b/pkg/certwatcher/certwatcher.go
@@ -27,6 +27,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher/metrics"
 	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 )
@@ -120,8 +121,31 @@ func (cw *CertWatcher) Start(ctx context.Context) error {
 	return cw.watcher.Close()
 }
 
+func (cw *CertWatcher) ensureAllFilesAreWatched() {
+	watchList := sets.New(cw.watcher.WatchList()...)
+	difference := sets.New(cw.certPath, cw.keyPath).Difference(watchList)
+	if difference.Len() == 0 {
+		return
+	}
+
+	for _, missingWatchPath := range difference.UnsortedList() {
+		log.V(1).Info("re-adding missing watch", "path", missingWatchPath)
+		if err := cw.watcher.Add(missingWatchPath); err != nil {
+			log.Error(err, "failed to add watch", "path", missingWatchPath)
+			return
+		}
+	}
+
+	log.V(1).Info("all files are watched again", "list", cw.watcher.WatchList())
+
+	if err := cw.ReadCertificate(); err != nil {
+		log.Error(err, "error re-reading certificate")
+	}
+}
+
 // Watch reads events from the watcher's channel and reacts to changes.
 func (cw *CertWatcher) Watch() {
+	watcherHealthTimer := time.NewTicker(time.Second)
 	for {
 		select {
 		case event, ok := <-cw.watcher.Events:
@@ -139,6 +163,8 @@ func (cw *CertWatcher) Watch() {
 			}
 
 			log.Error(err, "certificate watch error")
+		case <-watcherHealthTimer.C:
+			cw.ensureAllFilesAreWatched()
 		}
 	}
 }
--- a/pkg/certwatcher/certwatcher.go
+++ b/pkg/certwatcher/certwatcher.go
@@ -59,12 +59,7 @@ func New(certPath, keyPath string) (*CertWatcher, error) {
 		interval: defaultWatchInterval,
 	}
 
-	// Initial read of certificate and key.
-	if err := cw.ReadCertificate(); err != nil {
-		return nil, err
-	}
-
-	return cw, nil
+	return cw, cw.ReadCertificate()
 }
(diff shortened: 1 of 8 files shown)


pkg/certwatcher/certwatcher.go
```diff
@@ -139,6 +163,8 @@ func (cw *CertWatcher) Watch() {
 			}
 
 			log.Error(err, "certificate watch error")
+		case <-watcherHealthTimer.C:
+			cw.ensureAllFilesAreWatched()
```
alvaroaleman, line 167: Permanently polling once per second seems pretty aggrsive. Can we not watch the dir intead if the file gets removed and re-add the filewatch once it exists again?

vincepri: sidenote: wondering at this point if we should generally avoid relying of fs events, and rather say poll every N seconds and reload the certificates if the hash changed or something?

vincepri: We could also make the polling configurable, and default it to something reasonable where both old (loaded) and new certificates are supposed to be valid

m-messiah: > Can we not watch the dir intead if the file gets removed and re-add the filewatch once it exists again?

It does not watch the dir, it just tries to reinstall the notify watch that fails if the file is not present. 
So technically, it is the same as "waiting until the file exists again" as the first step of `addWatch` is `os.Lstat(name)`. 

The other option is to watch the directory events instead of file events and parse all events for the file names but it:
1. Increases the number of events to parse and handle
2. Does not solve the case when the directory would be removed and recreated again (or have mount issues, whatever).

m-messiah: > sidenote: wondering at this point if we should generally avoid relying of fs events, and rather say poll every N seconds and reload the certificates if the hash changed or something?

It makes sense, and in some of our internal projects, we rely on this behaviour instead. The downside is that you need to find a balance between polling interval, certificate expiration time, and number of file reads in this case; so ideally this param should also be configurable across all usages of certwatcher (webhook server and metrics server at least)

m-messiah: Thank you for your suggestions.
@alvaroaleman @vincepri I updated the PR with full removal of fsnotify from certwatcher and having a simple read/cache/repeat ticker loop. Please check and leave your opinions.

alvaroaleman (approved):
thanks!
/hold

@vincepri does this look good to you?

m-messiah:
Do you know what I should do with `apidiff` check? 

The method Watch does not look to be important outside of the Start, but it was Exported. Shall I keep it there and do nothing?

```
sigs.k8s.io/controller-runtime/pkg/certwatcher
  Incompatible changes:
  - (*CertWatcher).Watch: removed
  Compatible changes:
  - (*CertWatcher).WithWatchInterval: added
```

vincepri (approved):

vincepri:
/retitle 🐛 Refactor certificate watcher to use polling, instead of fsnotify

skhalash:
@m-messiah After this change our tests showed the following problem - our operator starts, a certificate is generated on startup and saved in the cert dir, and only after 10 seconds it is actually getting picked up. Which was not the case with the old `fsnotify` approach. We can fix it by decreasing the timeout, but is still feels wrong to me.

m-messiah:
> @m-messiah After this change our tests showed the following problem - our operator starts, a certificate is generated on startup and saved in the cert dir, and only after 10 seconds it is actually getting picked up. Which was not the case with the old `fsnotify` approach. We can fix it by decreasing the timeout, but is still feels wrong to me.

The expectation of certwatcher is that files are already present, or be present in the next loop after interval. 
I understand the problem, but it looks like your operator is trying to start the watcher before the certs are generated and to solve this you should be able to fix the order of prerequisites and the actual start.

skhalash:
@m-messiah Hmm, we are not doing anything fancy there and as far as I understand the initial cert read and the timer creation happen here https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/webhook/server.go#L207, an at that point our cert is already there.

skhalash:
I found the root cause in our codebase, but I’m still questioning this change. Our operator manages its own webhook certificates (rotation, etc.) instead of using an external service like cert-manager. The issue is that we assume updating the cert file makes it immediately usable by the HTTPS server. This worked before but doesn’t anymore after this change.

sbueringer:
@skhalash Please see: https://github.com/kubernetes-sigs/controller-runtime/pull/3050

PR #3022, opened 2024-11-23
:bug: Fakeclient: Don't return items on invalid selector

alvaroaleman:
Currently, if a List call to the fake client references an invalid fieldSelector, we will return an error and put all items into the passed List. Avoid putting anything into the passed list if the selector is invalid.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -599,21 +599,22 @@ func (c *fakeClient) List(ctx context.Context, obj client.ObjectList, opts ...cl
 		return err
 	}
 	zero(obj)
-	if err := json.Unmarshal(j, obj); err != nil {
+	objCopy := obj.DeepCopyObject().(client.ObjectList)
+	if err := json.Unmarshal(j, objCopy); err != nil {
 		return err
 	}
 
-	if listOpts.LabelSelector == nil && listOpts.FieldSelector == nil {
-		return nil
-	}
-
-	// If we're here, either a label or field selector are specified (or both), so before we return
-	// the list we must filter it. If both selectors are set, they are ANDed.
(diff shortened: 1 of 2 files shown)


troy0820:
Would a returned list from a field selector produce an error? I am under the assumption that a list call returns no error and an empty list, making the call successful but resulting in the items being empty. `len(list.Items) == 0`

pkg/client/fake/client.go
```diff
@@ -598,22 +598,23 @@ func (c *fakeClient) List(ctx context.Context, obj client.ObjectList, opts ...cl
 	if err != nil {
 		return err
 	}
-	zero(obj)
-	if err := json.Unmarshal(j, obj); err != nil {
+	objCopy := obj.DeepCopyObject().(client.ObjectList)
+	zero(objCopy)
```
sbueringer, line 602: Are we missing something by not zero'ing obj anymore?

alvaroaleman:
> Would a returned list from a field selector produce an error? I am under the assumption that a list call returns no error and an empty list, making the call successful but resulting in the items being empty. len(list.Items) == 0

I am not sure I follow the question, are you asking if an error is returned if the selector is invalid? If so, then yeah it is: https://github.com/kubernetes-sigs/controller-runtime/blob/fafee8cfecc3d372a5b4eea9dbc5122b743d6e3f/pkg/client/fake/client_test.go#L1324

pkg/client/fake/client.go
alvaroaleman: Not sure why I did it this way, changed it to just keep zeroing `obj` instead, ptal

sbueringer:
Thx!

/lgtm
/approve

Issue #3027, opened 2024-11-27
 Fix leader election request timeout


serathius:
The `controller-runtime` uses the `resourcelock.New` function for configuring leader election. This function is known to have an incorrectly configured request timeout, which sets the request timeout equal to the leader election deadline. This allows a single request timeout to trigger a change in leadership.

**Source:**

https://github.com/kubernetes-sigs/controller-runtime/blob/8e44a4307f70a4d5f4c5aaf0869e2bec7167f4ab/pkg/leaderelection/leader_election.go#L101-L109

**Impact:**

This issue causes unnecessary leader changes, which can cause:
* Lower availability - new leader might require reinitialization of informers which can take tens of seconds in large clusters.
* Waste of resources - Increased API server load due to concurrent re-initializations, potentially triggering a KCP scale-up and wasting resources.

**Fix:**

Update controller-runtime to use `resourcelock.NewFromKubeconfig` for leader election. This will ensure that the request timeout is correctly configured and prevent unnecessary leadership changes due to transient network issues or API server unavailability. This change should involve approximately 10 lines of code.

**Example**:

https://github.com/kubernetes/kubernetes/pull/98059

alvaroaleman:
@serathius thanks for the report, I've opened https://github.com/kubernetes-sigs/controller-runtime/pull/3028 to fix it

Why is the `resourceLock.New` not deprecated if it has known issues?

serathius:
> Why is the resourceLock.New not deprecated if it has known issues?

No breaking change policy :P 
Issue is just `resourceLock.New` allows users to configure deadline and timeout independently, so most users are not aware that relation between those two parameters can impact reliability.

PR #3028, opened 2024-11-30
:warning: Use leader elector with client timeout

alvaroaleman:
This change makes the leader elector use a client that internally has a smaller timeout than the renew deadline, which avoids a situation where a single request timing out makes us lose the leader lease.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3027

### DIFF
--- a/pkg/leaderelection/leader_election.go
+++ b/pkg/leaderelection/leader_election.go
@@ -20,10 +20,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/uuid"
-	coordinationv1client "k8s.io/client-go/kubernetes/typed/coordination/v1"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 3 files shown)


vincepri (approved):
Should we backport this? Seems like an important fix

pkg/leaderelection/leader_election.go
```diff
@@ -49,6 +48,9 @@ type Options struct {
 	// LeaderElectionID determines the name of the resource that leader election
 	// will use for holding the leader lock.
 	LeaderElectionID string
+
+	// RewnewDeadline is the renew deadline for this leader election client
+	RewnewDeadline time.Duration
```
vincepri, line 53: This should be RenewDeadline, right? Typo @alvaroaleman

vincepri: https://github.com/kubernetes-sigs/controller-runtime/pull/3032

alvaroaleman:
@vincepri I don't think we can backport this, the leader election options where I added the paramter are exported, this change breaks anyone who uses that directly.

Given that no one noticed this for years, I think its okay to not backport and wait for the next release

PR #3033, opened 2024-12-01
✨ Nit: Correct a typo in the docstring

shafeeqes:
Nit: Correct a typo in the docstring

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -94,16 +94,16 @@ type SubResourceClientConstructor interface {
 	// - ServiceAccount token creation:
 	//     sa := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "bar"}}
 	//     token := &authenticationv1.TokenRequest{}
-	//     c.SubResourceClient("token").Create(ctx, sa, token)
+	//     c.SubResource("token").Create(ctx, sa, token)
 	//
 	// - Pod eviction creation:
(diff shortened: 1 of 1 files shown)


pkg/client/fake/client.go
```diff
@@ -110,6 +112,7 @@ func NewClientBuilder() *ClientBuilder {
 // ClientBuilder builds a fake client.
 type ClientBuilder struct {
 	scheme                *runtime.Scheme
+	clock                 clock.Clock
```
alvaroaleman, line 115: I don't like this idea. There might be more things in the future where a timestamp is needed and the clock ends up being shared by all of them.

Please use `WithInterceptor` if the hardcoded default result doesn't work for you.

alvaroaleman: /hold

shafeeqes: Thanks, `WithInterceptor` does suit my needs.
 
> There might be more things in the future where a timestamp is needed and the clock ends up being shared by all of them.

For testing purposes, shouldn't this be okay?

alvaroaleman: > For testing purposes, shouldn't this be okay?

You are adding something that is global to the fakeclient but only used in this one place. The next person find another place where they would want to use it and the third person will then want this to be different between the first and the second place and they are stuck.

IMHO we should avoid that altogether by not making this directly configurable at all and redirect people to use the interceptor if they need something more specific than the hardcoded default.

shafeeqes: Okay, I removed the last commit.

alvaroaleman (approved):

PR #3045, opened 2024-12-20
🐛 fix: return error when object limit exceeded for cached list calls

### LINKED ISSUE #3044: Cached clients return incomplete object lists for paginated list calls
Today, the cache reader does not support paginated list calls:

https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/cache/internal/cache_reader.go#L116

However, in the initial list call, the cache reader does not return any signal to its callers to indicate whether the returned object list is incomplete:

https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/cache/internal/cache_reader.go#L149

As a result, callers may have erroneous behaviors if it is assumed that the resulting list is complete. Till the technical challenge in #532 is resolved, it seems an error should be returned if there are too many objects.

Thoughts?

chhsia0:
Since the cache does not implement strongly consistent paginated list calls, if the Limit option is set and the number of items listed exceeds this limit, there is no way for the caller to get a complete list of objects, so return an error here to notify the caller.

Fixes #3044.


alvaroaleman (commented):
/hold

This is a breaking change and breaks the use-case "Check if there are any objects of a given type". I agree the current behavior is subpar, but the new behavior would also be subpar but different and a breaking change so I am not in favor of merging this.

Alternate approaches:
1. Add support for this into the upstream informer
2. Return some sort of sentinel `continue` token and error out if it is used

sbueringer:
I think we can implement option 2. Or at least I don't see any downsides.

It should reduce accidental usage of pagination

sbueringer:
/close 
in favor of https://github.com/kubernetes-sigs/controller-runtime/pull/3134

detro:
> I think we can implement option 2. Or at least I don't see any downsides.

May I say that the approach taken to solve this was, at the very least, "opinionable"?

I bumped to `v0.21.0` and NOWHERE I can see mentioned this breaking change: https://github.com/kubernetes-sigs/controller-runtime/releases. Not to mention, this is a _minor_ version bump, so ... 🤷 .

Also, the way this breaks relies on the fact that we store a crypting string in the `.Continue` field:
```golang
out.SetContinue("continue-not-supported")
```

That breaks things because above the logic relies on this:

```golang
	if listOpts.Continue != "" {
		return fmt.Errorf("continue list option is not supported by the cache")
	}
```

The world "convoluted" comes to mind.

Anyway, it would be good to improve the documentation and show how exactly is one supposed to grab a non-cached client when trying to do a `.List` inside a `.Reconcile` loop. As it is entirely invisible to me.

PR #3048, opened 2024-12-28
✨add CodecFactoryOptionsMutators for codecfactory

Chaunceyctx:
What does this do, and why do we need it?

we have a production scenario:

the CRD field declared in the v1 version of the controller includes `field A`. However, when we upgrade the v1 controller to v2, `field A` is removed from the corresponding CRD. 
So we want the controller to intercept this error during deserialization, so we need to switch the corresponding serializer to `strict` mode.


Chaunceyctx:
/ping gently @vincepri @sbueringer PTAL, thanks a lot :)

pkg/client/client.go
```diff
@@ -150,7 +153,7 @@ func newClient(config *rest.Config, options Options) (*client, error) {
 		config:     config,
 		scheme:     options.Scheme,
 		mapper:     options.Mapper,
-		codecs:     serializer.NewCodecFactory(options.Scheme),
+		codecs:     serializer.NewCodecFactory(options.Scheme, options.CodecFactoryOptionsMutators...),
```
sbueringer, line 156: What is the user-facing difference in impact between configuring strict here vs. using client.WithFieldValdation to set FieldValidation in the Options structs?

Is it that the latter only affects Create/Update/patch?

Chaunceyctx: > What is the user-facing difference in impact between configuring strict here vs. using client.WithFieldValdation to set FieldValidation in the Options structs?
> 
> Is it that the latter only affects Create/Update/patch?

@sbueringer thanks a lot for feedback :) Yes, `FieldValidation` will be used in CreateOptions/UpdateOptions/PatchOptions. And these Options will tell apiserver how to select mode of decodeSerializer(strict or not strict), like: https://github.com/kubernetes/kubernetes/blob/6746df77f2376c6bc1fd0de767d2a94e6bd6cec1/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/update.go#L109

So client also need to know how to select mode of decodeSerializer and use strict mode of decodeSerializer to avoid the aforementioned issue.

I will address failed ci job.

Chaunceyctx:
@sbueringer excuse me, could you give me some hints to address this failed ci job? I am a noob for go-apidiff tool

Chaunceyctx:
> @sbueringer excuse me, could you give me some hints to address this failed ci job? I am a noob for go-apidiff tool

fields added need to be comparable. @sbueringer PTAL again, thanks a lot :)

troy0820:
Do we need to add any tests for this @sbueringer ?

pkg/cache/cache.go
```diff
@@ -140,6 +140,10 @@ type Options struct {
 	// Scheme is the scheme to use for mapping objects to GroupVersionKinds
 	Scheme *runtime.Scheme
 
+	// CodecFactoryStrict/CodecFactoryPretty are used to indicate whether enable Strict/Pretty mode of CodecFactory
+	CodecFactoryStrict bool
+	CodecFactoryPretty bool
```
alvaroaleman, line 145: What is the point of `CodecFactoryPretty`, when would someone want this?

pkg/client/client.go
alvaroaleman: That seems like a pretty big footgun though, because you became unable to tolerate any skew between the CRD and the controller, even for backwards-compatible changes?

pkg/cache/cache.go
Chaunceyctx: "strict" and "pretty" are both options of Codecfactory, so for the sake of completeness, I added these two variables together. But I indeed didn't think of practical scenarios for "pretty".

pkg/client/client.go
Chaunceyctx: @alvaroaleman thanks for your comment :). IMO, these options can be true or false which depends on the user's use case. For my scenario described above, I have deployed a workload similar to `StatefulSet` and used `ControllerRevision` to record the spec of this workload. If the `workload.spec` changes (some fields have been dropped in the v2 workload controller), it will lead to rolling update of the workload, impacting production. Therefore, I need to explicitly use the strict deserializer to prevent it.

alvaroaleman: Yeah but this is a cache-wide setting, using errors in the cache deserialization as a marker just seems like a very brittle approach

Chaunceyctx: It makes sense. I will find other approaches to implement it like `ByObject` field:)

Chaunceyctx:
I have changed the implementation so that the codec's use of strict mode is restricted to a specific GroupVersionKind (GVK) instead of applying to the cache-level. @alvaroaleman PTAL, thank you so much :)

PR #3050, opened 2024-12-30
🌱 Add fsnotify watcher+polling

vincepri:
This PR adds fsnotify back to the cert-watcher.

This helps in cases where fsnotify works and that depend on an instant certificate reload.

For example if the webhook certificate is directly used as caBundle in MutatingWebhookConfigurations / ValidatingWebhookConfigurations / CRDs.

Please note: It is highly recommended to use a CA (that is used as caBundle in MutatingWebhookConfigurations / ValidatingWebhookConfigurations / CRDs) and a separate serving certificate using that CA. So during serving cert rotation both the old and new certificate are valid. In that case fsnotify is not required and a cert reload within 10 seconds is good enough.


sbueringer:
Thx!

/lgtm

/hold
for a final look from @vincepri (as I pushed a commit with test changes)

PR #3058, opened 2025-01-03
🐛 priority queue: Fix panic within spin

sbueringer:

I did some scale testing with priority queue and Cluster API and hit a panic within spin

```
panic: runtime error: index out of range [1] with length 1
goroutine 249 [running]:
[github.com/google/btree.(*node[...]).iterate(0x2ce6820](http://github.com/google/btree.(*node[...]).iterate(0x2ce6820), 0x1, {0x0, 0x0}, {0x0, 0x0}, 0x0, 0x1, 0x4000baabe0)
        /Users/buringerst/code/pkg/mod/github.com/google/btree@v1.1.3/btree_generic.go:522 +0x5b4
[github.com/google/btree.(*BTreeG[...]).Ascend(0x2ce6560](http://github.com/google/btree.(*BTreeG[...]).Ascend(0x2ce6560), 0x4000baabe0)
        /Users/buringerst/code/pkg/mod/github.com/google/btree@v1.1.3/btree_generic.go:779 +0xb4
[sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.(*priorityqueue[...]).spin.func1(0x40007d8200](http://sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.(*priorityqueue[...]).spin.func1(0x40007d8200), 0x2ce11a0)
        /Users/buringerst/code/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1-0.20250103111815-1ac370e560b1/pkg/controller/priorityqueue/priorityqueue.go:198 +0x218
[sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.(*priorityqueue[...]).spin(0x2ce11a0)](http://sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.(*priorityqueue[...]).spin(0x2ce11a0))
        /Users/buringerst/code/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1-0.20250103111815-1ac370e560b1/pkg/controller/priorityqueue/priorityqueue.go:227 +0x198
created by [sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.New[...]](http://sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue.New[...]) in goroutine 247
        /Users/buringerst/code/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1-0.20250103111815-1ac370e560b1/pkg/controller/priorityqueue/priorityqueue.go:76 +0x66c
```

The problem is that if we delete an item from the queue within Ascend it can happen that Ascend panics with `panic: runtime error: index out of range [1] with length 1`

I wasn't able yet to reproduce this with a unit test (it seems to occur only under specific circumstances)

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -195,35 +195,41 @@ func (w *priorityqueue[T]) spin() {
 			w.lockedLock.Lock()
 			defer w.lockedLock.Unlock()
 
-			w.queue.Ascend(func(item *item[T]) bool {
-				if w.waiters.Load() == 0 { // no waiters, return as we can not hand anything out anyways
-					return false
-				}
-
-				// No next element we can process
-				if item.readyAt != nil && item.readyAt.After(w.now()) {
-					readyAt := item.readyAt.Sub(w.now())
-					if readyAt <= 0 { // Toctou race with the above check
-						readyAt = 1
+			for continueLoop := true; continueLoop; {
+				continueLoop = false
+				w.queue.Ascend(func(item *item[T]) bool {
+					if w.waiters.Load() == 0 { // no waiters, return as we can not hand anything out anyways
+						return false
 					}
-					nextReady = w.tick(readyAt)
+
+					// No next element we can process
+					if item.readyAt != nil && item.readyAt.After(w.now()) {
+						readyAt := item.readyAt.Sub(w.now())
+						if readyAt <= 0 { // Toctou race with the above check
+							readyAt = 1
+						}
+						nextReady = w.tick(readyAt)
+						return false
+					}
+
+					// Item is locked, we can not hand it out
+					if w.locked.Has(item.key) {
+						return true
+					}
+
+					w.metrics.get(item.key)
+					w.locked.Insert(item.key)
+					w.waiters.Add(-1)
+					delete(w.items, item.key)
+					w.queue.Delete(item)
+					w.get <- *item
+
+					// Return false because continuing with Ascend after deleting an item
+					// can lead to panics within Ascend.
+					continueLoop = true
 					return false
-				}
-
-				// Item is locked, we can not hand it out
-				if w.locked.Has(item.key) {
-					return true
-				}
-
-				w.metrics.get(item.key)
-				w.locked.Insert(item.key)
-				w.waiters.Add(-1)
-				delete(w.items, item.key)
-				w.queue.Delete(item)
-				w.get <- *item
-
-				return true
-			})
+				})
+			}
 		}()
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -283,6 +284,41 @@ var _ = Describe("Controllerworkqueue", func() {
 		Expect(metrics.depth["test"]).To(Equal(0))
 		Expect(metrics.adds["test"]).To(Equal(2))
 	})
+
+	It("returns many items", func() {
```
sbueringer, line 288: Not sure how to verify that the spin goroutine didn't panic.

Locally with Intellij, I've hit a panic break point and after continuing the test was shown as successful. I'm not sure if the same happens in CI

alvaroaleman: Maybe that there was a breakpoint caused go test to not recognize it properly?

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -195,35 +195,41 @@ func (w *priorityqueue[T]) spin() {
...
+					w.locked.Insert(item.key)
+					w.waiters.Add(-1)
+					delete(w.items, item.key)
+					w.queue.Delete(item)
+					w.get <- *item
+
+					// Return false because continuing with Ascend after deleting an item
+					// can lead to panics within Ascend.
+					continueLoop = true
 					return false
```
alvaroaleman, line 230: Not a super strong opinion but how do you feel about instead appending to a `toDelete` slice in `Ascend` and then calling `Delete` after being  done with `Ascend`? It should be safe because we are holding the lock so a concurrent routine seeing the item when it is supposed to be deleted shouldn't be possible.

The reason is that even if it works this way now, I don't think manipluting the tree while iterating is an expected usage, even if it seems to work now there could be more bugs or it could stop working in a future version of the lib.

alvaroaleman: I ended up using that approach + your testcase in https://github.com/kubernetes-sigs/controller-runtime/pull/3060 as this issue made CI fail there, hope that is okay

pkg/controller/priorityqueue/priorityqueue_test.go
sbueringer: Jup maybe

pkg/controller/priorityqueue/priorityqueue.go
sbueringer: Yup. You're right, seems safer to just not delete within Ascend. I'll take a look at your PR. Thx for taking this over

sbueringer:
/close

in favor of #3060

PR #3062, opened 2025-01-08
📖 update cache options design to reflect the implementation

ranakan19:
📖 Updates the cache options design and examples to reflect the implementation.
for example DefaultNamespaces is of type DefaultNamespaces map[string]Config, and thus expects an empty struct in the examples instead of nil.


alvaroaleman (approved):
thanks!

PR #3063, opened 2025-01-08
🐛 Add DeletionTimeStamp when deleting objects with clients 

### LINKED ISSUE #3059: Object doesn't get updated after Client.Delete(object)
fakeClient.Delete(obj) doesn't update deletionTimestamp of the obj passed to the function. The bug is in fakeClient.deleteObject, since it gets a new object and updates the new object without rewriting the original object.

Reference: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.0/pkg/client/fake/client.go#L1085-L1095

troy0820:
When deleting the object with the fake client and type client, the DeletionTimeStamp is not updated on the object.

Resolves #3059 

/assign @sbueringer 
/assign @alvaroaleman

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -763,6 +763,8 @@ func (c *fakeClient) Delete(ctx context.Context, obj client.Object, opts ...clie
 	}
 	delOptions := client.DeleteOptions{}
 	delOptions.ApplyOptions(opts)
+	now := metav1.Now()
+	obj.SetDeletionTimestamp(&now)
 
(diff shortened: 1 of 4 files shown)


troy0820:
/hold 

This seems wrong, we don't want to add the timestamp on the delete request in the client, because I'm assuming the api-server does that.  We do what that behavior to be mimicked in the fake client but want to handle the behaviors we expect that the test suite already provides.

troy0820:
/close 

I am unassigning myself and allowing others to solve this issue.

PR #3066, opened 2025-01-10
:bug: Fix a bug in the priorityqueue metrics

alvaroaleman:
The priorityqueue needs to call `metrics.add` but only once an item was ready. When an item was initially added with `After`, so not ready, then without which makes it ready, we didn't do that, resulting in potentially negative values for the queue depth metric.

### DIFF
--- a/pkg/controller/priorityqueue/metrics.go
+++ b/pkg/controller/priorityqueue/metrics.go
@@ -4,7 +4,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/clock"
 )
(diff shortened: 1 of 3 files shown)


pkg/controller/priorityqueue/metrics.go
```diff
@@ -34,7 +33,6 @@ func newQueueMetrics[T comparable](mp workqueue.MetricsProvider, name string, cl
 		workDuration:            mp.NewWorkDurationMetric(name),
 		unfinishedWorkSeconds:   mp.NewUnfinishedWorkSecondsMetric(name),
 		longestRunningProcessor: mp.NewLongestRunningProcessorSecondsMetric(name),
-		added:                   sets.Set[T]{},
```
alvaroaleman, line 37: This secondary bookkeeping was added because originally the queue called `metrics.Add` when the queues Add was called before I realized this is wrong, now that `metrics.add` is only called when an item was newly-added or became ready, this isn't needed anymore

sbueringer:
Thx!

/lgtm
/approve

Issue #3057, opened 2025-01-03
concurrency errors when printing source.TypedSource


nicks:
**Repro steps:**
- Write a custom implementation of source.TypedSource
- Add a Watch on it

**Result:**
Non-deterministically, controller-runtime blows up with this error:

```
fatal error: concurrent map iteration and map write
```

(full stack trace included below)

**The Problem**

Controller does a `fmt.Printf` on the Source, which implicitly reads all the internal properties.
https://github.com/kubernetes-sigs/controller-runtime/blob/1ac370e560b19321af779df0610622c03106ec77/pkg/internal/controller/controller.go#L178

All the built-in Sources implement fmt.Stringer with a concurrency-safe `String() string` method, so do not have this problem.

Suggestion: the TypedSource interface should also implement `fmt.Stringer`, so that all implementers have to implement String()

**Additional info**

Full stack trace:

```
goroutine 194539 [running]:
reflect.mapiternext(0x18c24c9?)
	/usr/local/go/src/runtime/map.go:1537 +0x13
reflect.(*MapIter).Next(0xc005b6ad38?)
	/usr/local/go/src/reflect/value.go:1989 +0x74
internal/fmtsort.Sort({0x42cf740?, 0xc0202be2c8?, 0xc0267a4380?})
	/usr/local/go/src/internal/fmtsort/sort.go:59 +0x1a7
fmt.(*pp).printValue(0xc012c1b930, {0x42cf740?, 0xc0202be2c8?, 0x5cc9b4207?}, 0x76, 0x3)
	/usr/local/go/src/fmt/print.go:816 +0x988
fmt.(*pp).printValue(0xc012c1b930, {0x4659fe0?, 0xc0202be2a0?, 0xc01f0d7320?}, 0x76, 0x2)
	/usr/local/go/src/fmt/print.go:853 +0x11be
fmt.(*pp).printValue(0xc012c1b930, {0x44e0640?, 0xc0202be2a0?, 0x74cea8f28e70?}, 0x76, 0x1)
	/usr/local/go/src/fmt/print.go:853 +0x11be
fmt.(*pp).printValue(0xc012c1b930, {0x4569ce0?, 0xc0202be2a0?, 0x182a8dd?}, 0x76, 0x0)
	/usr/local/go/src/fmt/print.go:921 +0xae5
fmt.(*pp).badVerb(0xc012c1b930, 0x73)
	/usr/local/go/src/fmt/print.go:394 +0x47f
[...]

PR #3068, opened 2025-01-13
🐛 Check to see if custom source implements fmt.Stringer when logging

troy0820:
When using a custom source, the error when starting will fail because the `fmt.Stringer` interface is not implemented by the source.  The builtin sources have already implanted this interface allow the below line to not fail:

```go
log := c.LogConstructor(nil).WithValues("source", watch.String())
```
```
fatal error: concurrent map iteration and map write
```

Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3057

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -175,7 +175,16 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 		// caches.
 		errGroup := &errgroup.Group{}
 		for _, watch := range c.startWatches {
-			log := c.LogConstructor(nil).WithValues("source", watch.String())
+			log := c.LogConstructor(nil)
+			_, ok := watch.(interface {
+				String() string
+			})
+
+			if !ok {
+				log = log.WithValues("source", fmt.Sprintf("%T", watch))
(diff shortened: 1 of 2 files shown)


dongjiang1989 (approved):
Great. Looks very elegant

pkg/internal/controller/controller.go
```diff
@@ -175,7 +175,7 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 		// caches.
 		errGroup := &errgroup.Group{}
 		for _, watch := range c.startWatches {
-			log := c.LogConstructor(nil).WithValues("source", fmt.Sprintf("%s", watch))
+			log := c.LogConstructor(nil).WithValues("source", watch.String())
```
alvaroaleman, line 178: It is an interesting issue - How about  rather then requiring the implementation of `fmt.Stringer`, we check if it is implemented and if yes use that, otherwise use the Type (i.E. `fmt.Sprintf("%T", watch)`)? That way we avoid the breaking change

alvaroaleman (approved):
thanks!

PR #3070, opened 2025-01-14
✨feat: add GoCollector and ProcessCollector metrics

dongjiang1989:
Default add GoCollector and ProcessCollector metrics for controller-runtime metrics package

### DIFF
--- a/pkg/internal/controller/metrics/metrics.go
+++ b/pkg/internal/controller/metrics/metrics.go
@@ -17,6 +17,8 @@ limitations under the License.
 package metrics
 
 import (
+	"regexp"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -89,6 +91,13 @@ func init() {
 		// expose process metrics like CPU, Memory, file descriptor usage etc.
 		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
 		// expose Go runtime metrics like GC stats, memory stats etc.
-		collectors.NewGoCollector(),
+		collectors.NewGoCollector(
+			collectors.WithGoCollectorRuntimeMetrics(
+				collectors.MetricsGC,
+				collectors.MetricsScheduler,
+				collectors.MetricsMemory,
+				collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sync/.*`)},
+			),
+		),
 	)
 }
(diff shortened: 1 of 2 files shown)


pkg/metrics/registry.go
```diff
@@ -27,4 +32,22 @@ type RegistererGatherer interface {
...
+	r.MustRegister(
+		collectors.NewGoCollector(
+			collectors.WithGoCollectorRuntimeMetrics(
+				collectors.MetricsGC,
+				collectors.MetricsScheduler,
+				collectors.MetricsMemory,
+				collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sync/.*`)},
+			),
+		),
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
```
troy0820, line 50: https://github.com/kubernetes-sigs/controller-runtime/blob/1ec7c1b76e98efd4f177eb8cd96050596b0fa3d3/pkg/internal/controller/metrics/metrics.go#L90

We do this here, do we need to do this here as well?

dongjiang1989: Thanks @troy0820 
Move to kg/internal/controller/metrics better. 
Fixed.

pkg/internal/controller/metrics/metrics.go
```diff
@@ -89,6 +91,13 @@ func init() {
 		// expose process metrics like CPU, Memory, file descriptor usage etc.
 		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
 		// expose Go runtime metrics like GC stats, memory stats etc.
-		collectors.NewGoCollector(),
+		collectors.NewGoCollector(
```
alvaroaleman, line 94: I don't think we want this change - Any config we put in here will be wrong for some cases so we'd rather keep it as-is to avoid the breaking change. You can unregister and re-register these in the config you need if the default doesn't work for you: https://github.com/gitpod-io/gitpod/blob/58e1f558c90aa819b6061ce170ad7b0320183dec/components/ws-daemon/pkg/daemon/daemon.go#L63
/hold

sbueringer: @dongjiang1989 Was chatting with Alvaro.

In general we don't want to finetune things like this too much as it's hard to find something that works for everyone.

But in this specific case we're good. It did some research and it turns out we can simply enable all Go metrics as there are not that many. SO it's not too much overhead. And having all Go metrics should of course work for everyone

Please change this to:
```go
collectors.NewGoCollector(collectors.WithGoCollectorRuntimeMetrics(collectors.MetricsAll))
```

(btw component-base, i.e. kube-apiserver, ... are doing the same)

/hold cancel

dongjiang1989: Thanks @alvaroaleman and @sbueringer 
Changed it

sbueringer:
Thx!

/ok-to-test
/lgtm

alvaroaleman (approved):

PR #3075, opened 2025-01-19
🌱 Add debug logging for the state of the priority queue

alvaroaleman:
This debug logging prints the state of the workqueue in order to allow debugging it. It will continously check at runtime if debug logging is enabled and do nothing if not, making it very cheap if unused.

Sample output piped through `jq` for readability:
```
{
  "level": "debug",
  "ts": "2025-01-19T12:00:43-05:00",
  "msg": "workqueue_state",
  "controller": "configmap",
  "items": [
    {
      "key": {
        "Namespace": "kube-system",
        "Name": "kubeadm-config"
      },
      "addedCounter": 1,
      "priority": -100
    }
  ]
}
```

### DIFF
--- a/examples/priorityqueue/main.go
+++ b/examples/priorityqueue/main.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"time"
 
+	"go.uber.org/zap/zapcore"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -45,7 +46,9 @@ func main() {
 }
 
 func run() error {
-	log.SetLogger(zap.New())
+	log.SetLogger(zap.New(func(o *zap.Options) {
+		o.Level = zapcore.Level(-5)
+	}))
 
 	// Setup a Manager
 	mgr, err := manager.New(kubeconfig.GetConfigOrDie(), manager.Options{
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -5,6 +5,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/google/btree"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
@@ -36,6 +37,7 @@ type Opts[T comparable] struct {
 	// limiter with an initial delay of five milliseconds and a max delay of 1000 seconds.
 	RateLimiter    workqueue.TypedRateLimiter[T]
 	MetricProvider workqueue.MetricsProvider
+	Log            logr.Logger
 }
 
 // Opt allows to configure a PriorityQueue.
@@ -57,6 +59,7 @@ func New[T comparable](name string, o ...Opt[T]) PriorityQueue[T] {
 	}
 
 	pq := &priorityqueue[T]{
+		log:         opts.Log,
 		items:       map[T]*item[T]{},
 		queue:       btree.NewG(32, less[T]),
 		becameReady: sets.Set[T]{},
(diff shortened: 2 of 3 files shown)


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -326,36 +331,64 @@ func (w *priorityqueue[T]) Len() int {
...
-	return a.addedCounter < b.addedCounter
+	return a.AddedCounter < b.AddedCounter
 }
 
 type item[T comparable] struct {
-	key          T
-	addedCounter uint64
-	priority     int
-	readyAt      *time.Time
+	Key          T          `json:"key"`
```
alvaroaleman, line 379: Exporting these is required for the json marshalling, otherwise they get omitted

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -326,36 +331,64 @@ func (w *priorityqueue[T]) Len() int {
...
+		select {
+		case <-w.done:
+			return
+		case <-t:
+		}
+
+		// Log level may change at runtime, so keep the
+		// loop going even if a given level is currently
+		// not enabled.
+		if !w.log.V(1).Enabled() {
```
alvaroaleman, line 346: I know its a bit more common in the ecosystem to use `V(4)` for `debug`, but I think many of our users use `zap` + `zapr` as the underlying implementation for the `logr.Logger`, given that we include it and do that in our examples. The problem is that `zap` has named log levels. These do end up mapping to a `int8` but I personally only learned this recently and I've been using `zap` for years. The named `debug` level maps to `-1` and `zapr`  then switches the sign, resulting in `V(1)` in the `logr.Logger`.

Unless we have a strong reason why we need something higher than `V(1)`, i'd like to keep it this way so that this log shows up if someone uses a `zap` logger on `debug` level.

sbueringer: The recommended production log level with klog is 2

alvaroaleman: I've updated it to `V(5)` for now so this doesn't show up by default and opened an issue for what seems to be a mismatch for the log level mapping: https://github.com/kubernetes-sigs/controller-runtime/issues/3081

examples/priorityqueue/main.go
```diff
@@ -45,7 +46,9 @@ func main() {
 }
 
 func run() error {
-	log.SetLogger(zap.New())
+	log.SetLogger(zap.New(func(o *zap.Options) {
```
sbueringer, line 49: Not sure if you want to keep this the same way

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -326,36 +331,64 @@ func (w *priorityqueue[T]) Len() int {
...
+		case <-t:
+		}
+
+		// Log level may change at runtime, so keep the
+		// loop going even if a given level is currently
+		// not enabled.
+		if !w.log.V(5).Enabled() {
+			continue
+		}
+		items := make([]*item[T], 0, len(w.items))
```
sbueringer, line 349: I think this should be ideally one line below within the lock

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -326,36 +331,64 @@ func (w *priorityqueue[T]) Len() int {
...
+		w.lock.Lock()
+		w.queue.Ascend(func(item *item[T]) bool {
+			items = append(items, item)
+			return true
+		})
+		w.lock.Unlock()
+
+		w.log.V(5).Info("workqueue_state", "items", items)
```
sbueringer, line 357:
```suggestion
		w.log.V(5).Info("Workqueue items", "items", items)
```
Maybe this makes it look a bit more like a regular log message. No strong opinion, up to you.

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Done, great catch

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: I don't like the space in there because that makes the query more annoying with some log query systems and I think its uncommon to uppercase, so I ended up with `workqueue_items` - WDYT?

sbueringer: Sounds good

sbueringer:
Thank you!

/lgtm
/approve

PR #2907, opened 2024-08-07
🌱  Use envtest v1.30.2

sbueringer:
Signed-off-by: Stefan Büringer buringerst@vmware.com

Just realized we're still using 1.28 when looking into https://github.com/kubernetes-sigs/controller-runtime/pull/2901

PR #3095, opened 2025-01-29
🌱 Update dependencies, k8s 1.32.1

joshuacuellar1:
This pull request updates the dependencies of controller-runtime.
Kubernetes 1.32.1 was released 2 weeks ago.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
```
alvaroaleman, line 1: Please put this into its own PR

hack/check-everything.sh
```diff
@@ -30,7 +30,7 @@ export GOTOOLCHAIN="go$(make go-version)"
 ${hack_dir}/verify.sh
 
 # Envtest.
-ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION:-"1.28.0"}
+ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION:-"1.32.0"}
```
alvaroaleman, line 33: There is an existing PR for this and it mentions the reasons why we don't want that just yet: https://github.com/kubernetes-sigs/controller-runtime/pull/2907

alvaroaleman (commented):
/ok-to-test

go.mod
```diff
@@ -1,34 +1,34 @@
 module sigs.k8s.io/controller-runtime
 
-go 1.23.0
+go 1.23.4
```
sbueringer, line 4: Is it necessary to enforce this patch version?

.devcontainer/devcontainer.json
joshuacuellar1: I removed the Devcontainer from the main branch.

hack/check-everything.sh
joshuacuellar1: I reverted the envtest version back to 1.28.0.

go.mod
joshuacuellar1: I reverted it back to 1.23.0.
The latest kubernetes-sigs/client-go is using 1.23.0 ([https://github.com/kubernetes/client-go/blob/master/go.mod](https://github.com/kubernetes/client-go/blob/master/go.mod))

sbueringer:
Seems like some dependencies are now out of sync with what k8s.io/dependencies are using

See: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3095/pull-controller-runtime-test/1884792898218627072

PR #3098, opened 2025-01-30
:bug: Update object upon deletion

bigkevmcd:
If an object is deleted and the API responds with a resource, update the deleted object.

This causes the object to be updated with the deletion timestamp.


pkg/client/client_test.go
```diff
@@ -1654,6 +1656,36 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+				Expect(err).NotTo(HaveOccurred())
+
+				By("adding a finalizer we prevent the node from being deleted immediately")
+				controllerutil.AddFinalizer(node, "example.com/test")
+				node, err = clientset.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+				Expect(err).NotTo(HaveOccurred())
+
+				By("deleting the Node")
+				nodeName := node.Name
+				err = cl.Delete(context.TODO(), node)
```
alvaroaleman, line 1675: Nit: why `context.TODO` here?

pkg/client/client_test.go
```diff
@@ -1654,6 +1656,36 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(node.ObjectMeta.DeletionTimestamp).NotTo(BeNil())
```
alvaroaleman, line 1677: Can we also test this with `unstructured.Unstructured` and `metav1.MetaObject`, please?

alvaroaleman (commented):
Thanks, this is great! We also need to update the fake client to match the behavior

pkg/client/client_test.go
bigkevmcd: I replaced it with the ctx.

pkg/client/client_test.go
bigkevmcd: Unstructured works... metav1.MetaObject will require a fair bit more work.

https://github.com/kubernetes-sigs/controller-runtime/blob/c5bb1d4d8e7d0e49bb06f8276785e7f7cbcbf2b8/pkg/client/metadata_client.go#L67

Not sure if this is relevant? https://github.com/kubernetes-sigs/controller-runtime/blob/c5bb1d4d8e7d0e49bb06f8276785e7f7cbcbf2b8/pkg/client/metadata_client.go#L30

bigkevmcd:
> Thanks, this is great! We also need to update the fake client to match the behavior

My reading of this test is that the fake client does actually work?

https://github.com/kubernetes-sigs/controller-runtime/blob/c5bb1d4d8e7d0e49bb06f8276785e7f7cbcbf2b8/pkg/client/fake/client_test.go#L817

alvaroaleman:
That example is after a `Get` call, it should be set after a `Delete` call

pkg/client/client_test.go
sbueringer: I think we should also implement this for the metadata client. It's already tricky that we change a nuance like this, but I think it's a lot worse if our clients behave inconsistently

pkg/client/unstructured_client.go
```diff
@@ -111,7 +111,7 @@ func (uc *unstructuredClient) Delete(ctx context.Context, obj Object, opts ...De
 		Name(o.GetName()).
 		Body(deleteOpts.AsDeleteOptions()).
 		Do(ctx).
-		Error()
+		Into(obj)
```
sbueringer, line 114: I ran this through the debugger and this object looks like this afterwards if I'm not mistaken:

image

I.e. the delete endpoint does not return a resource, it returns a status

sbueringer: I think we also need a test case that when we delete via Unstructured without a finalizer the object keeps it's data and does not become a `Status` object

alvaroaleman: Should we maybe just check if the returned body is a status and only if its not update the original object?

ahmetb: As far as I can tell there's no way to tell apart you got the object back or a metav1.Status unless you actually read the request body.

alvaroaleman: Which I think might be okay, the main overhead will be that we have to deserialize it twice if it is not a status

sbueringer: I don't remember the details to be honest. Does anyone know if we would be able to get the deletionTimestamp in all cases? (or only if there are finalizers or not). The latter would then be pretty surprising behavior of our client if we implement it like that

bigkevmcd: No...if we get a status back, there's no deletion timestamp.

Basically, if the delete mechanisn returns a "result" value, then it gets returned, otherwise it's just a status value.

https://github.com/kubernetes/kubernetes/blob/aa35105213d031c25abf172ac1102a58f144c85e/staging/src/k8s.io/apiserver/pkg/registry/rest/rest.go#L177-L178

https://github.com/kubernetes/kubernetes/blob/aa35105213d031c25abf172ac1102a58f144c85e/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/delete.go#L193

alvaroaleman: So it is possible to get a status back, even if the object is still in the system?

I wonder if it is still worth to check if the result is not a status and update `obj` if so

sbueringer: I think it would be a bad UX if after calling client.Delete() you only sometimes see the deletionTimestamp set :/

ahmetb: In general, it appears if it's gone from storage you'll get Status object in the response body. (You can try Pod deletion with grace period.) But I don't think there's a guarantee around any of this in the API machinery.

alvaroaleman: > I think it would be a bad UX if after calling client.Delete() you only sometimes see the deletionTimestamp set :/

I mean what can we realistically put in there if the object is gone from storage? I guess we could put a non-nil fake deletiontimestamp in there but not sure if that is a good idea.

What is IMHO annoying about the current state is that you can not do a check of `Did I already delete this`

ahmetb: Since Delete() already doesn't do what we expect it to do, what if we changed its return signature to `(deletedFromStorage bool, err error)`?

sbueringer: > I mean what can we realistically put in there if the object is gone from storage? I guess we could put a non-nil fake deletiontimestamp in there but not sure if that is a good idea.

Yup agree. I mostly just meant that if deletionTimestamp is only sometimes set (aka in some cases) this would be tricky to rely on

alvaroaleman:
Ref https://github.com/kubernetes-sigs/controller-runtime/issues/3059

sbueringer:
Added it to the PR description

PR #3102, opened 2025-02-05
:bug: Fix custom path for webhooks conflict

damsien:
Following this PR #2998. I implemented the custom path for webhooks considering that it will be used by either only the `validating` or the `defaulting`. However, the user can actually use the both of them on the same webhook manager.

Until now, when the user was using `.WithDefaulter(&TestCustomDefaulter{})`, `.WithValidator(&TestCustomValidator{})` and `.WithCustomPath(customPath)` at the same time, an endless loop was launched by constantly calling the other webhook. This bug is caused because they are both register at the same path.

This is the reason why I made this PR. I split the `.WithCustomPath()` function into two functions `.WithDefaultingCustomPath()` and `.WithValidatingCustomPath()` in order to be used in this way:

```go
WebhookManagedBy(mgr).
	For(&TestDefaulter{}).
	WithDefaulter(&TestCustomDefaulter{}).
	WithValidator(&TestCustomValidator{}).
	WithDefaultingCustomPath(defaultingCustomPath).
	WithValidatingCustomPath(validatingCustomPath).
	Complete()
```

Also, I added two tests:
1. One that tests the example above.
2. And I take advantage of making this PR to add a second one that tests the usage of `WithDefaulter()` and `WithValidator()` at the same time (without the custom path implementation) because there were no test about this specific configuration before.

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -37,17 +37,19 @@ import (
 
 // WebhookBuilder builds a Webhook.
 type WebhookBuilder struct {
-	apiType             runtime.Object
-	customDefaulter     admission.CustomDefaulter
-	customDefaulterOpts []admission.DefaulterOption
-	customValidator     admission.CustomValidator
-	customPath          string
-	gvk                 schema.GroupVersionKind
-	mgr                 manager.Manager
-	config              *rest.Config
-	recoverPanic        *bool
-	logConstructor      func(base logr.Logger, req *admission.Request) logr.Logger
-	err                 error
+	apiType                   runtime.Object
+	customDefaulter           admission.CustomDefaulter
+	customDefaulterOpts       []admission.DefaulterOption
+	customValidator           admission.CustomValidator
+	customPath                string
+	customValidatorCustomPath string
+	customDefaulterCustomPath string
+	gvk                       schema.GroupVersionKind
+	mgr                       manager.Manager
+	config                    *rest.Config
+	recoverPanic              *bool
+	logConstructor            func(base logr.Logger, req *admission.Request) logr.Logger
+	err                       error
 }
 
 // WebhookManagedBy returns a new webhook builder.
@@ -96,8 +98,22 @@ func (blder *WebhookBuilder) RecoverPanic(recoverPanic bool) *WebhookBuilder {
 }
 
 // WithCustomPath overrides the webhook's default path by the customPath
+// Deprecated: WithCustomPath should not be used anymore.
+// Please use WithValidatorCustomPath or WithDefaulterCustomPath instead.
 func (blder *WebhookBuilder) WithCustomPath(customPath string) *WebhookBuilder {
 	blder.customPath = customPath
+	return blder
+}
+
+// WithValidatorCustomPath overrides the path of the Validator.
+func (blder *WebhookBuilder) WithValidatorCustomPath(customPath string) *WebhookBuilder {
+	blder.customValidatorCustomPath = customPath
+	return blder
+}
+
+// WithDefaulterCustomPath overrides the path of the Defaulter.
+func (blder *WebhookBuilder) WithDefaulterCustomPath(customPath string) *WebhookBuilder {
+	blder.customDefaulterCustomPath = customPath
 	return blder
 }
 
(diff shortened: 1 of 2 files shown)


damsien:
Hello @alvaroaleman , what do you think about this one? I know that it can potentially be a breaking change because we are renaming a function that already exists. I am not sure about the upgrade process that must be done (is it simply a :warning: breaking change warning in the release note or a more elaborated procedure?).

pkg/builder/webhook.go
```diff
@@ -95,9 +96,15 @@ func (blder *WebhookBuilder) RecoverPanic(recoverPanic bool) *WebhookBuilder {
 	return blder
 }
 
-// WithCustomPath overrides the webhook's default path by the customPath
-func (blder *WebhookBuilder) WithCustomPath(customPath string) *WebhookBuilder {
```
alvaroaleman, line 99: Renaming this makes this PR a breaking change for everyone, we should not do that. The two main alternatives I can think of are:
1) Only one of `WithCustomDefaulter`/`WithCustomValidator` may be set when using `WithCustomPath`, otherwise we error out
2) We use indexing, i.E. when both are set, there must also be two `WithCustomPath` calls and the first gets the first/the second the second

damsien: Hello Alvaro,

What about this one:

We keep the `WithCustomPath()`, `WithDefaultingCustomPath()` and `WithValidatingCustomPath()` functions. If the developer uses the `WithCustomPath()` function, then only one of the `WithCustomDefaulter()` or `WithCustomValidator()` should be set, otherwise the developer should use the specific custom path functions.

Maybe it is not a good idea because we have multiple function accomplishing the same purpose. Waiting for you thoughts about it.

damsien:
Hello @alvaroaleman @sbueringer,

Following our conversation, I add some pre-checks:
1. Enforce only one of `WithCustomDefaulter()`/`WithCustomValidator()` when using `WithCustomPath()`
2. Enforce only one of `WithDefaultingCustomPath()` or `WithCustomPath()`
3. Enforce only one of `WithValidatingCustomPath()` or `WithCustomPath()`

If at least one of these checks are not good, then we error out when calling `Complete()`.

Also, I added a `Deprecated` comment on top of the `WithCustomPath()` function.

What do you think?

pkg/builder/webhook.go
sbueringer: @damsien I meant something slightly diferently :)

Instead of having additional WithValidatingCustomPath & WithDefaultingCustomPath funcs I was suggesting to add options to the existing WithDefaulter / WithValidator functions to set the path (similar to the existing option that already exists for the WithDefaulter funct)

sbueringer: Okay nevermind. Just noticed taht DefaulterOptions are passed to WithCustomDefaulter and there the custom path wouldn't make any sense

pkg/builder/webhook.go
```diff
@@ -150,6 +170,18 @@ func (blder *WebhookBuilder) registerWebhooks() error {
 		return err
 	}
 
+	if blder.isThereCustomPathConflict() {
+		return errors.New("only one of the custom defaulter or custom validtor should be set when using WithCustomPath. Otherwise, WithDefaultingCustomPath() and WithValidatingCustomPath() should be used.")
```
sbueringer, line 174:
```suggestion
		return errors.New("only one of CustomDefaulter or CustomValidator should be set when using WithCustomPath. Otherwise, WithDefaulterCustomPath() and WithValidatorCustomPath() should be used.")
```

pkg/builder/webhook.go
```diff
@@ -150,6 +170,18 @@ func (blder *WebhookBuilder) registerWebhooks() error {
...
+	} else {
+		if blder.customPath != "" {
+			// isThereCustomPathConflict() already checks for potential conflicts.
+			// Since we are sure that only one of customDefaulter or customValidator will be used,
+			// we can set set both defaultingCustomPath and validatingCustomPath.
```
sbueringer, line 179:
```suggestion
			// we can set both defaultingCustomPath and validatingCustomPath.
```

pkg/builder/webhook.go
```diff
@@ -37,17 +37,19 @@ import (
...
-	recoverPanic        *bool
-	logConstructor      func(base logr.Logger, req *admission.Request) logr.Logger
-	err                 error
+	apiType              runtime.Object
+	customDefaulter      admission.CustomDefaulter
+	customDefaulterOpts  []admission.DefaulterOption
+	customValidator      admission.CustomValidator
+	customPath           string
+	validatingCustomPath string
+	defaultingCustomPath string
```
sbueringer, line 46:
```suggestion
	customValidatorCustomPath string
	customDefaulterCustomPath string
```

damsien:
Thanks @sbueringer !
I renamed the functions and variables (validating -> validator & defaulting -> defaulter). Did some changes to pass the lint check.

sbueringer:
Thank you!

/lgtm
/approve

PR #3103, opened 2025-02-06
🐛Fix behavior of rate limit option in priorityqueue.AddWithOpts

zach593:
https://github.com/kubernetes-sigs/controller-runtime/blob/2e8ba92873fd2a92ba473926f02e493619480771/pkg/controller/priorityqueue/priorityqueue.go#L130-L146

Since `priorityqueue.AddWithOpts()` allows adding multiple items, if `o.RateLimited` is true, the `o.After` value could get overwritten in the loop and affect subsequent items.


alvaroaleman (approved):
great catch, thanks for that and sending the fix!

PR #3104, opened 2025-02-07
⚠ Bump to k8s.io/* v0.33.0-alpha.1

sbueringer:


pkg/manager/manager_test.go
```diff
@@ -1819,7 +1819,7 @@ var _ = Describe("manger.Manager", func() {
 		<-m.Elected()
 
 		Eventually(func() *corev1.Event {
-			evts, err := clientset.CoreV1().Events("").Search(m.GetScheme(), &ns)
+			evts, err := clientset.CoreV1().Events("").SearchWithContext(ctx, m.GetScheme(), &ns)
```
sbueringer, line 1822: fyi. This was not a breaking change in client-go. Search is just deprecated in favor of SearchWithContext

pkg/controller/controllertest/util.go
```diff
@@ -109,6 +110,10 @@ func (f *FakeInformer) Run(<-chan struct{}) {
 	f.RunCount++
 }
```
sbueringer, line 112: The SharedIndexInformer interface add new methods. So I extended the FakeInformer accordingly.

This was expected and previously discussed in https://github.com/kubernetes/kubernetes/pull/126387#discussion_r1698030889

sbueringer:


(cc @pohly just fyi, I think everything as expected so far)

pkg/controller/controllertest/util.go
sbueringer: In general there should be some follow-up to see how we can leverage changes like https://github.com/kubernetes/kubernetes/pull/127709#event-16033741601 & https://github.com/kubernetes/kubernetes/pull/126387#event-15707700128

pohly: You can use the "trick" from https://github.com/kubernetes/kubernetes/issues/126379 with `'s;// *Contextual logging: ;//logcheck:context //;' in your vendor directory, then run `logcheck` to get errors about using functions which have better alternatives.

We haven't sketched out a timeline yet, but for 1.33 I am hoping to adapt source code. Then in 1.34, those free-form comments could become proper `logcheck:context` tags and thus active for users of logcheck.

sbueringer: THx!

alvaroaleman (approved):

alvaroaleman:
/hold
in case vince wants to check

pkg/controller/controllertest/util.go
sbueringer: For referencing. Replacing `// Contextual Logging:` with `// Deprecated:` also works (and doesn't require the logcheck linter)

sbueringer: Follow-up PR: https://github.com/kubernetes-sigs/controller-runtime/pull/3149

PR #3107, opened 2025-02-08
:warning: Deprecate `reconcile.Result.Requeue`

alvaroaleman:
There is no good reason to use this setting, either an error or `RequeueAfter` should be used instead. Deprecate it to avoid confusion.


fabriziopandini:
I just want to bring up some cases when requeue helped us in Cluster API

When a reconcile loop is depending on external objects that you cannot actively watch, having in controller runtime an option to requeue with an exponential backoff (without raising an error) helped CAPI to ensure a good reactivity of the system without overload it.

When we tried to use RequeueAfter for similar use cases instead we faced a few issues:
- short RequeueAfter periods makes the system more reactive, but if you keep requeing frequently for long you can get into other problems, like e.g.
   - rate limiting on external resources
   - scalability problems in your controllers if too many objects are requeing frequently at the same time
- long RequeueAfter periods makes the system less reactive

Note. In our experience the current implementation of requeue works when things in the external system happens quickly, less when the exponential backoff was getting too high.

alvaroaleman:
> Note. In our experience the current implementation of requeue works when things in the external system happens quickly, less when the exponential backoff was getting too high.

Right, but wouldn't it be a lot better to set up your ratelimiter for this and use what interval it emits in `RequeueAfter` rather than re-using one that is not meant for that and where you already noticed that the ceiling is too high?

fabriziopandini:
> Right, but wouldn't it be a lot better to set up your ratelimiter for this and use what interval it emits in RequeueAfter rather than re-using one that is not meant for that and where you already noticed that the ceiling is too high?

What exists today in controller runtime worked for use in most cases, and this was enough for us in Cluster API to avoid creating our own custom rate limiter (having to implement on will be a step back).

Looking at this from another angle, if CR runtime is going to provide as a building block a rated limiter to use in controller with requreAfter, then the impact of deprecation/removal of requeue will be less relevant.

alvaroaleman:
> Looking at this from another angle, if CR runtime is going to provide as a building block a rated limiter to use in controller with requreAfter, then the impact of deprecation/removal of requeue will be less relevant.

Not controller-runtime itself, but there are ratelimiters in the upstream workqueue package: https://pkg.go.dev/k8s.io/client-go/util/workqueue

fabriziopandini:
@alvaroaleman thanks for sharing your PoV, appreciated
As far as we have options to address more advanced use cases, I'm ok.

vincepri (approved):

PR #3109, opened 2025-02-09
✨ [golangci-lint] add local-prefixes to linters-settings of goimports

tariq1890:
This PR adds a check to ensure that all first-party/local imports are moved to their own imports section

Sorting imports into groups helps with code readability. As per Go conventions, we maintain the following import groups:

- stdlib/built-in imports
- third-party imports
- local project imports

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,6 +40,8 @@ linters:
     - whitespace
 
 linters-settings:
+  goimports:
+    local-prefixes: sigs.k8s.io/controller-runtime
   govet:
     enable-all: true
     disable:
--- a/alias.go
+++ b/alias.go
@@ -19,6 +19,7 @@ package controllerruntime
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
--- a/pkg/builder/builder_suite_test.go
+++ b/pkg/builder/builder_suite_test.go
@@ -26,12 +26,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
-	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/internal/testing/addr"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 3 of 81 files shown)


schrej (approved):

sbueringer:
Why?

tariq1890:
I've updated the PR description

sbueringer:
Can you please link the Go conventions you are referring to?

tariq1890:
The Go style guide goes into this [here](https://google.github.io/styleguide/go/decisions.html#imports). Plus, the goimports binary supports recognizing the local imports and enforces formatting accordingly. See [here](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.30.0:cmd/goimports/goimports.go;l=54)

sbueringer:
> The Go style guide goes into this [here](https://google.github.io/styleguide/go/decisions.html#imports). Plus, the goimports binary supports recognizing the local imports and enforces formatting accordingly. See [here](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.30.0:cmd/goimports/goimports.go;l=54)

To quote the google style guide you are linking
> Imports should be organized in two groups:
> * Standard library packages
> * Other (project and vendored) packages

sbueringer:
> Why do we have [three groups](https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci.yml?rgh-link-date=2025-02-11T10%3A19%3A31Z#L62) on cluster-api?

Very good question. I think because at some point someone had an opintion about it (likely me :joy:). But not necessarily because it's an established convention in Go.

schrej:
So I took a look out of curiosity and [you were the one that enabled `gci`](https://github.com/kubernetes-sigs/cluster-api/commit/ec1fa651a3791f356057517193c90d284ae920b8), but didn't enable moving the local prefix into a separate group. That was [done by stmcginnis](https://github.com/kubernetes-sigs/cluster-api/commit/5026786ee809c5def466049f5befd2a786fbcefa), who also provided some reasoning for it:

> Good practice is to have imports separated into three distinct groups:
standard library imports, third party package imports, and local
imports. The gci linter is used to enforce this convention, making it
easy to see what is local, what is standard with go, and what is being
pulled in externally.
> 
> The gci linter configuration has a "prefix" setting that is used to tell
it what to match as the starting string to identify local imports. This
was set to "false", effectively disabling the ability to differentiate
local imports from third party ones.

sbueringer:
Thx for looking it up. I was mostly curious if there are actually Go conventions about this somewhere :)

I was preferring the split in 3 groups a while back, but over the years I realized that it makes absolutely 0 difference for me. I'm simply never manually reading over the imports in a way that the grouping would help.

That being said, just my personal opinion. No strong opinion either way

In general I have a preference for trying to not enforce personal preferences in Open Source projects, if it doesn't matter much either way.

@vincepri @alvaroaleman WDYT?

alvaroaleman:
I don't really care much either way, I like to have stdlib in its own group but thats it

tariq1890:
> In general I have a preference for trying to not enforce personal preferences in Open Source projects, if it doesn't matter much either way.

I disagree with the characterisation that this is a "personal preference". This has been a common practice in Go for several years, it is exemplified by the fact `goimports` provides a `-local` flag and golangci-lint exposes a field for it in its config

sbueringer:
Golangci-lint has a huge number of fields that allow all sorts of linter configurations. That doesn't make every one of them a best practice.

The fact that this is configurable and not the default in goimports makes it pretty clear to me that there is no clear consensus that this is the one standard for how to format imports.

I'm not saying that you are the only one who prefers this. I'm just saying there is no clear standard that everyone should or has to do it this way. Which makes it a personal preference for me

Issue #3105, opened 2025-02-07
Handlers should default to LowPriorityWhenUnchanged without a wrapper


alvaroaleman:
When we implemented the logic in the handlers to lower the priority for events where the object didn't change, we did that with a wrapper as that can be re-used.

Unfortunately, this requires everyone to remember to use this wrapper, which is extremely easy to forget:
```
handler.WithLowPriorityWhenUnchanged(&handler.TypedEnqueueRequestForObject[*my.CRD]{})
```

We should instead implement this in the handlers themselves and default to `true` to avoid the need for this.

sbueringer:
/reopen

Not entirely done yet

sbueringer:
/reopen

@alvaroaleman Do we have to modify more handlers?
* EnqueueRequestForObject
* TypedEnqueueRequestForOwner[client.Object]


\+ Should we stop wrapping the handlers with WithLowPriorityWhenUnchanged in pkg/builder/controller.go? Seems redundant

alvaroaleman:
> EnqueueRequestForObject
> TypedEnqueueRequestForOwner[client.Object]

These already do: https://github.com/kubernetes-sigs/controller-runtime/blob/5dfe3216fb7fd7f5afb61d6d0f8956c7bec8df62/pkg/handler/eventhandler_test.go#L787-L800

> + Should we stop wrapping the handlers with WithLowPriorityWhenUnchanged in pkg/builder/controller.go? Seems redunda

Removing that here: https://github.com/kubernetes-sigs/controller-runtime/pull/3168

sbueringer:
> EnqueueRequestForObject

Sorry not sure how I got the idea that this one doesn't implement it

> TypedEnqueueRequestForOwner[client.Object]

EnqueueRequestForOwner has it
TypedEnqueueRequestForOwner[client.Object] does not have it (as it's implemented in the EnqueueRequestForOwner func)

sbueringer:
Okay. I think now we're done :)

+/- https://github.com/kubernetes-sigs/controller-runtime/issues/3171 but we have that separate issue for that

/close

PR #3111, opened 2025-02-10
🐛Implement priorityqueue as default on handlers if using priorityqueue interface

troy0820:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3105

Handlers: 
- [x] `TypedEnqueueRequestForObject`
- [x] `EnqueueRequestForOwner`
- [ ] `EnqueueRequestFromMapFunc`

This brings the handler `TypedEnqueueRequestForObject` , `enqueueRequestForOwner` and `enqueuRequestFromMapFunc` handlers to use priority queue if the `workqueue.TypedRateLimitingInterface[reconcile.Request]` is of type  `priorityqueue.PriorityQueue[reconcile.Request]` without having to use the wrapper `WithLowPriorityWhenUnchanged`.

**NOTE: Added tests implement the same logic to show that this maps to the functionality of the wrapped handler in the tests**

/assign @alvaroaleman @sbueringer

### DIFF
--- a/pkg/handler/enqueue.go
+++ b/pkg/handler/enqueue.go
@@ -52,25 +52,32 @@ func (e *TypedEnqueueRequestForObject[T]) Create(ctx context.Context, evt event.
 		enqueueLog.Error(nil, "CreateEvent received with no metadata", "event", evt)
 		return
 	}
-	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+
+	item := reconcile.Request{NamespacedName: types.NamespacedName{
 		Name:      evt.Object.GetName(),
 		Namespace: evt.Object.GetNamespace(),
-	}})
+	}}
+
+	addToQueueCreate(q, evt, item)
 }
 
 // Update implements EventHandler.
 func (e *TypedEnqueueRequestForObject[T]) Update(ctx context.Context, evt event.TypedUpdateEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	switch {
 	case !isNil(evt.ObjectNew):
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		item := reconcile.Request{NamespacedName: types.NamespacedName{
 			Name:      evt.ObjectNew.GetName(),
 			Namespace: evt.ObjectNew.GetNamespace(),
-		}})
+		}}
+
+		addToQueueUpdate(q, evt, item)
 	case !isNil(evt.ObjectOld):
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		item := reconcile.Request{NamespacedName: types.NamespacedName{
 			Name:      evt.ObjectOld.GetName(),
 			Namespace: evt.ObjectOld.GetNamespace(),
-		}})
+		}}
+
+		addToQueueUpdate(q, evt, item)
 	default:
 		enqueueLog.Error(nil, "UpdateEvent received with no metadata", "event", evt)
 	}
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -46,7 +46,7 @@ type TypedMapFunc[object any, request comparable] func(context.Context, object) 
 // For UpdateEvents which contain both a new and old object, the transformation function is run on both
 // objects and both sets of Requests are enqueue.
 func EnqueueRequestsFromMapFunc(fn MapFunc) EventHandler {
-	return TypedEnqueueRequestsFromMapFunc(fn)
+	return WithLowPriorityWhenUnchanged(TypedEnqueueRequestsFromMapFunc(fn))
 }
 
 // TypedEnqueueRequestsFromMapFunc enqueues Requests by running a transformation function that outputs a collection
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -46,7 +46,7 @@ type TypedMapFunc[object any, request comparable] func(context.Context, object) 
 // For UpdateEvents which contain both a new and old object, the transformation function is run on both
 // objects and both sets of Requests are enqueue.
 func EnqueueRequestsFromMapFunc(fn MapFunc) EventHandler {
-	return WithLowPriorityWhenUnchanged(TypedEnqueueRequestsFromMapFunc(fn))
+	return TypedEnqueueRequestsFromMapFunc(fn)
 }
 
 // TypedEnqueueRequestsFromMapFunc enqueues Requests by running a transformation function that outputs a collection
--- a/pkg/handler/eventhandler.go
+++ b/pkg/handler/eventhandler.go
@@ -18,8 +18,10 @@ package handler
 
 import (
 	"context"
+	"reflect"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
@@ -65,7 +67,7 @@ type EventHandler = TypedEventHandler[client.Object, reconcile.Request]
 //
 // Unless you are implementing your own TypedEventHandler, you can ignore the functions on the TypedEventHandler interface.
 // Most users shouldn't need to implement their own TypedEventHandler.
-//
+
 // TypedEventHandler is experimental and subject to future change.
 type TypedEventHandler[object any, request comparable] interface {
 	// Create is called in response to a create event - e.g. Pod Creation.
@@ -111,7 +113,19 @@ type TypedFuncs[object any, request comparable] struct {
 // Create implements EventHandler.
 func (h TypedFuncs[object, request]) Create(ctx context.Context, e event.TypedCreateEvent[object], q workqueue.TypedRateLimitingInterface[request]) {
 	if h.CreateFunc != nil {
-		h.CreateFunc(ctx, e, q)
+		if reflect.TypeFor[request]() != reflect.TypeOf(reconcile.Request{}) || !reflect.TypeFor[object]().Implements(reflect.TypeFor[client.Object]()) {
+			h.CreateFunc(ctx, e, q)
+		}
+
+		wq, ok := q.(workqueue.TypedRateLimitingInterface[reconcile.Request])
+		if ok {
+			evt := any(e).(event.TypedCreateEvent[client.Object])
+			item := reconcile.Request{NamespacedName: types.NamespacedName{
+				Name:      evt.Object.GetName(),
+				Namespace: evt.Object.GetNamespace(),
+			}}
+			addToQueueCreate(wq, evt, item)
+		}
 	}
 }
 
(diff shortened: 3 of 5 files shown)


troy0820:
I am not sure if you wanted to do this for the handler `*enqueueRequestsFromMapFunc[object, request])` but I can add this here or make another PR with the change.

alvaroaleman:
> I am not sure if you wanted to do this for the handler *enqueueRequestsFromMapFunc[object, request]) but I can add this here or make another PR with the change.

Yeah we would want this for all handlers we offer if possible. This PR is on my list but it might take me a few more days to find the time, sorry

pkg/handler/enqueue_mapped.go
```diff
@@ -81,7 +82,12 @@ func (e *enqueueRequestsFromMapFunc[object, request]) Create(
 	q workqueue.TypedRateLimitingInterface[request],
 ) {
 	reqs := map[request]empty{}
-	e.mapAndEnqueue(ctx, q, evt.Object, reqs)
+	priorityQueue, isPriorityQueue := q.(priorityqueue.PriorityQueue[request])
```
alvaroaleman, line 85: Here we only expose a constructor and not a type - Can we instead simply wrap the `enqueueRequestFromMapFunc` with `WithLowPriorityWhenUnchanged`?

pkg/handler/eventhandler.go
```diff
@@ -199,3 +199,23 @@ func (w workqueueWithCustomAddFunc[request]) Add(item request) {
 func isObjectUnchanged[object client.Object](e event.TypedCreateEvent[object]) bool {
 	return e.Object.GetCreationTimestamp().Time.Before(time.Now().Add(-time.Minute))
 }
+
+// addToPriorityQueueCreate adds the reconcile.Request to the priorityqueue in the handler
+// for Create requests if and only if the workqueue being used is of type priorityqueue.PriorityQueue[reconcile.Request]
+func addToPriorityQueueCreate[T client.Object](q priorityqueue.PriorityQueue[reconcile.Request], evt event.TypedCreateEvent[T], item reconcile.Request) {
```
alvaroaleman, line 205: Same here, lets just wrap the handler in the constructor

troy0820: These are functions to not repeat the logic in the `TypedEnqueueRequestForObject[T]`  Create/Update as we do export this type.  

Do you want this to have a constructor to wrap with `WithLowPriorityWhenUnchanged` because this will require adding more constraints to deal with the func signature with ^^ to equal `TypedEventHandler[object,request]`. Unless I am misunderstanding something.

alvaroaleman: Sorry I think I just commented on the wrong place here but I see you already updated it 👍

pkg/handler/enqueue.go
```diff
@@ -52,25 +53,47 @@ func (e *TypedEnqueueRequestForObject[T]) Create(ctx context.Context, evt event.
...
+	}
+	addToPriorityQueueCreate(priorityQueue, evt, item)
 }
 
 // Update implements EventHandler.
 func (e *TypedEnqueueRequestForObject[T]) Update(ctx context.Context, evt event.TypedUpdateEvent[T], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	switch {
 	case !isNil(evt.ObjectNew):
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		item := reconcile.Request{NamespacedName: types.NamespacedName{
```
alvaroaleman, line 74: Could we also do the same for `Funcs`? That is the only remaining handler

troy0820: Not sure if this is how you wanted this done but any feedback is appreciated.

alvaroaleman: I think in `Funcs` we have to do the same dance as in `TypedEnqueueRequestForObject` because we export the type and not a constructor, otherwise the original goal of `get this by default` won't be achieved

pkg/handler/enqueue.go
```diff
@@ -52,25 +53,47 @@ func (e *TypedEnqueueRequestForObject[T]) Create(ctx context.Context, evt event.
...
 		Name:      evt.Object.GetName(),
 		Namespace: evt.Object.GetNamespace(),
-	}})
+	}}
+
+	priorityQueue, isPriorityQueue := q.(priorityqueue.PriorityQueue[reconcile.Request])
+	if !isPriorityQueue {
+		q.Add(item)
+		return
+	}
```
sbueringer, line 66: Should we also move this into the util func? (and rename the func to addToQueueCreate)

Same for update

pkg/handler/eventhandler.go
```diff
@@ -199,3 +205,23 @@ func (w workqueueWithCustomAddFunc[request]) Add(item request) {
 func isObjectUnchanged[object client.Object](e event.TypedCreateEvent[object]) bool {
 	return e.Object.GetCreationTimestamp().Time.Before(time.Now().Add(-time.Minute))
 }
+
+// addToPriorityQueueCreate adds the reconcile.Request to the priorityqueue in the handler
```
sbueringer, line 209: Would be good if we can use these two funcs in WithLowPriorityWhenUnchanged

pkg/handler/eventhandler.go
```diff
@@ -87,6 +87,12 @@ var _ EventHandler = Funcs{}
 // Funcs implements eventhandler.
 type Funcs = TypedFuncs[client.Object, reconcile.Request]
 
+// NewFuncsWithLowPriority creates a Funcs Handler wrapped in
+// WithLowPriorityWhenUnchanged
+func NewFuncsWithLowPriority(u Funcs) EventHandler {
```
sbueringer, line 92: The naming of NewFuncsWithLowPriority vs WithLowPriorityWhenUnchanged seems inconsistent (New vs With)

pkg/handler/enqueue.go
troy0820: Makes sense, I did what we did with `TypedEnqueueRequestForObject` because I can't add methods to the `Funcs` type.  So I did the same thing we did with this.

pkg/handler/enqueue.go
troy0820: I can. I would have to move `isObjectUnchanged` as well.  Not sure where to move it if we want that exposed or in some internal directory.

troy0820: Gotcha, I misunderstood.  I can move them as well to do the check within the function. I did rename them but can move those lines inside.

sbueringer: Sounds good. Just a lot of code we can deduplicate :)

pkg/handler/eventhandler.go
```diff
@@ -82,10 +83,72 @@ type TypedEventHandler[object any, request comparable] interface {
 	Generic(context.Context, event.TypedGenericEvent[object], workqueue.TypedRateLimitingInterface[request])
 }
 
-var _ EventHandler = Funcs{}
+var _ EventHandler = &Funcs{}
```
alvaroaleman, line 86: I think this can not be a pointer receiver otherwise this is a breaking change 😬 Can we also typecheck taht `TypedFuncs` which is now a different implementaiton is a `TypedEventHandler`?

pkg/handler/eventhandler.go
```diff
@@ -83,9 +84,73 @@ type TypedEventHandler[object any, request comparable] interface {
...
 var _ EventHandler = Funcs{}
+var _ EventHandler = TypedFuncs[client.Object, reconcile.Request]{}
+var _ TypedEventHandler[client.Object, reconcile.Request] = TypedFuncs[client.Object, reconcile.Request]{}
 
 // Funcs implements eventhandler.
-type Funcs = TypedFuncs[client.Object, reconcile.Request]
+type Funcs = funcs[client.Object, reconcile.Request]
+
+// funcs implements eventhandler
+type funcs[object client.Object, request reconcile.Request] struct{}
```
alvaroaleman, line 94: Actually this is not correct, right now this is just a re-implementation of `EnqueueRequestForOwner` 😅 We need to have public properties for CreateFunc/UpdateFunc/DeleteFunc/GenericFunc and delegate to them just like we do in `TypedFuncs`

troy0820: Does `Funcs` need to be it's own type? I can't add methods to `Funcs` if it's a `TypedFuncs[client.Object, reconcile.Request]` and declaring `CreateFunc` methods  causes an overflow of calls if I try to inject it in the `h.CreateFunc(ctx, e, q)` call.  🤔 

Do you want the default behavior of the `priorityqueue`  to be based on if the `CreateFunc` is not defined?

alvaroaleman: > Do you want the default behavior of the priorityqueue to be based on if the CreateFunc is not defined?

I think if the `CreateFunc` is not defined we should just not do anything just like we do today.

Maybe a better approach is to use `Reflect` in `TypedFuncs` to figure out if `object` is assignable to `client.Object` and `request` is a `reconcile.Request` and if so, use the helpers you added for `EnqueueRequestForObject`? I did some of that in the `Builder` because `For` there can only be used if `request` is a `reconcile.Request`

troy0820:
Test logic needs updating because now that this is wrapped in the `WithLow.....` tests are showing back with the wrong length for reconcile requests? 🤔  I can debug the test cases but also want to make sure I do not ruin any existing functionality as that would be a breaking change and change that 🌱 into a ⚠️

alvaroaleman:
> Test logic needs updating because now that this is wrapped in the WithLow..... tests are showing back with the wrong length for reconcile requests? 🤔 I can debug the test cases but also want to make sure I do not ruin any existing functionality as that would be a breaking change and change that 🌱 into a ⚠️

No, this should have no effect on existing functionality. I can see a bunch of panics in CI, so something seems to not be working correctly.

alvaroaleman:
@troy0820 how about you leave out EnqueueRequestFromMapFunc from this change and I will have a look into that one, as it seems a bit more tricky?

pkg/handler/eventhandler.go
```diff
@@ -111,7 +113,22 @@ type TypedFuncs[object any, request comparable] struct {
...
-		h.CreateFunc(ctx, e, q)
+		if reflect.TypeFor[request]() != reflect.TypeOf(reconcile.Request{}) || !reflect.TypeFor[object]().Implements(reflect.TypeFor[client.Object]()) {
+			h.CreateFunc(ctx, e, q)
+		}
+
+		wq, ok := q.(workqueue.TypedRateLimitingInterface[reconcile.Request])
+		if ok {
+			evt, ok := any(e).(event.TypedCreateEvent[client.Object])
+			if ok {
+				item := reconcile.Request{NamespacedName: types.NamespacedName{
```
alvaroaleman, line 124: This isn't quite correct, because we need to run `CreateFunc` to know what `item` looks like, here we basically hardcode the same behavior as `EnqueueRequestForObjec` - Lets maybe leave the `TypedFuncs` out as well of this PR?

alvaroaleman (approved):
Thanks!

pkg/handler/eventhandler.go
```diff
@@ -65,7 +65,7 @@ type EventHandler = TypedEventHandler[client.Object, reconcile.Request]
 //
 // Unless you are implementing your own TypedEventHandler, you can ignore the functions on the TypedEventHandler interface.
 // Most users shouldn't need to implement their own TypedEventHandler.
-//
```
sbueringer, line 68: This change breaks godoc rendering of TypedEventHandler. Now only l.69 shows up

pkg/handler/eventhandler_test.go
```diff
@@ -819,6 +840,28 @@ var _ = Describe("Eventhandler", func() {
 			Expect(actualRequests).To(Equal([]reconcile.Request{{NamespacedName: types.NamespacedName{Name: "my-pod"}}}))
 		})
 
+		It("should not lower the priority of a create request for an object that was created less than one minute in the past without the WithLowPriority wrapperr", func() {
```
sbueringer, line 843: nit:
wrapperr => wrapperr

Please also do one search replace here for WithLowPriority => WithLowPriorityWhenUnchanged

(would have ignored this one, but as there's also another finding.. :))

sbueringer (commented):
@troy0820 Thx for working on this!

Just one small finding, would be nice if you can follow-up

pkg/handler/eventhandler_test.go
sbueringer: Maybe let's ignore this comment, because it should be resolved after => https://github.com/kubernetes-sigs/controller-runtime/pull/3152#discussion_r1992789734

PR #3126, opened 2025-02-20
🐛 fix: cache should list out of global cache when present and necessary

joelanford:
When the cache options are configured with DefaultNamespaces which include an entry with `cache.AllNamespaces`, listing from the cache should fallback to the global cache if there are no namespace-specific caches that match the namespace from the list options.


troy0820:
/lgtm 
/hold 

In case anyone else wants to review.

alvaroaleman (approved):
That is a subtle one, thanks for the fix!

PR #3129, opened 2025-02-21
🌱 Export envtest.ReadCRDFiles

sbueringer:

I would like to export ReadCRDFiles so I can use the functionality without being forced to use the entire InstallCRDs func.

CreateCRDs and WaitForCRDs are already exported

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -94,7 +94,7 @@ func InstallCRDs(config *rest.Config, options CRDInstallOptions) ([]*apiextensio
 	defaultCRDOptions(&options)
 
 	// Read the CRD yamls into options.CRDs
-	if err := readCRDFiles(&options); err != nil {
+	if err := ReadCRDFiles(&options); err != nil {
 		return nil, fmt.Errorf("unable to read CRD files: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/envtest/crd.go
```diff
@@ -115,8 +115,8 @@ func InstallCRDs(config *rest.Config, options CRDInstallOptions) ([]*apiextensio
 	return options.CRDs, nil
 }
 
-// readCRDFiles reads the directories of CRDs in options.Paths and adds the CRD structs to options.CRDs.
-func readCRDFiles(options *CRDInstallOptions) error {
+// ReadCRDFiles reads the directories of CRDs in options.Paths and adds the CRD structs to options.CRDs.
+func ReadCRDFiles(options *CRDInstallOptions) error {
```
sbueringer, line 119: Not sure if I should change the input paramter to only take / return exactly what is needed.

The current input parameter is a bit awkward. On the other side we have the same in WaitForCRDs

alvaroaleman: No strong opinion either way

alvaroaleman (approved):
/hold

In case Vince wants to have a look

pkg/envtest/crd.go
sbueringer: Yeah same :)

PR #3134, opened 2025-02-25
🐛 Return error if pagination is used with the cached client

sbueringer:

Implements option 2 mentioned in https://github.com/kubernetes-sigs/controller-runtime/pull/3045

This should make it a lot easier to detect usages of pagination with the cached client (which is not supported)


sbueringer:
I can mark the PR as breaking if we want. But if folks where using the usual pattern for pagination before they probably had an endless loop? :) So not sure if we're actually breaking anything here

alvaroaleman (approved):
thanks!

PR #3135, opened 2025-03-01
✨ envtest: add option to download binaries, bump envtest to v1.32.0

sbueringer:
Signed-off-by: Stefan Büringer buringerst@vmware.com

PR #3136, opened 2025-03-02
✨ Use watchList in typed and unstructured client if WatchListClient is enabled

### LINKED ISSUE #3055: Ensure APIStreaming is used if available
We should be using https://kubernetes.io/blog/2024/12/17/kube-apiserver-api-streaming/ if the server supports it. It could be that there are no changes required for that, but we should validate that assumption before the next release.

sbueringer:

Fixes #3055 3055


sbueringer:
/hold
First commit is from https://github.com/kubernetes-sigs/controller-runtime/pull/3135

Otherwise ready for review

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1759 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 		})
 	})
```
sbueringer, line 2207: I had to move the List tests out, but with a bit of creative diff'ing it should be okay'ish to review

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1759 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
-				}
-				Expect(hasDep).To(BeTrue())
-			})
+		It("should allow setting the field manager", func() {
+			po := &client.CreateOptions{}
+			client.FieldOwner("some-owner").ApplyToCreate(po)
+			Expect(po.AsCreateOptions().FieldManager).To(Equal("some-owner"))
+		})
 
-			It("should fetch unstructured collection of objects", func() {
```
sbueringer, line 2227: I moved the unstructured tests into the unstructured section

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1759 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
-			})
+		It("should allow setting Precondition", func() {
+			do := &client.DeleteOptions{}
+			pc := metav1.NewUIDPreconditions("uid")
+			client.Preconditions(*pc).ApplyToDelete(do)
+			Expect(do.AsDeleteOptions().Preconditions).To(Equal(pc))
+			Expect(do.Preconditions).To(Equal(pc))
+		})
 
-			It("should fetch unstructured collection of objects, even if scheme is empty", func() {
```
sbueringer, line 2261: I moved this one to the unstructured section

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1765 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "deployment-frontend",
-						Namespace: ns,
-						Labels:    map[string]string{"app": "frontend"},
+			By("validating no Deployments are returned")
+			Expect(deps.Items).To(BeEmpty())
+		})
+
+		// TODO(seans): get label selector test working
+		It("should filter results by label selector", func() {
```
sbueringer, line 3101: Added this test as it was not implemented for unstructured before

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1765 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
-			})
+		It("should allow setting Precondition", func() {
+			do := &client.DeleteOptions{}
+			pc := metav1.NewUIDPreconditions("uid")
+			client.Preconditions(*pc).ApplyToDelete(do)
+			Expect(do.AsDeleteOptions().Preconditions).To(Equal(pc))
+			Expect(do.Preconditions).To(Equal(pc))
+		})
 
-			It("should fetch unstructured collection of objects, even if scheme is empty", func() {
```
sbueringer, line 2261: Moved to the unstructured section

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1765 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
-				}
-				Expect(hasDep).To(BeTrue())
-			})
+		It("should allow setting the field manager", func() {
+			po := &client.CreateOptions{}
+			client.FieldOwner("some-owner").ApplyToCreate(po)
+			Expect(po.AsCreateOptions().FieldManager).To(Equal("some-owner"))
+		})
 
-			It("should fetch unstructured collection of objects", func() {
```
sbueringer, line 2227: This test was redundant with the one in the unstructured section, so I deleted it

sbueringer:
/hold cancel

Rebased on top of main, now that https://github.com/kubernetes-sigs/controller-runtime/pull/3135 is merged

pkg/client/typed_client.go
```diff
@@ -157,6 +160,21 @@ func (c *typedClient) List(ctx context.Context, obj ObjectList, opts ...ListOpti
...
+		err := r.Get().
+			NamespaceIfScoped(listOpts.Namespace, r.isNamespaced()).
+			Resource(r.resource()).
+			VersionedParams(&watchListOptions, c.paramCodec).
+			WatchList(ctx).
+			Into(obj)
+		if err == nil {
+			return nil
+		}
+		log.FromContext(ctx).Error(err, fmt.Sprintf("Warning: The watchlist request for %s ended with an error, falling back to the standard LIST semantics", r.resource()))
```
alvaroaleman, line 175: If I understand it correctly, this failing is expected if the KAS is too old, right? If yes, then this i IMHO a debug log. The other log re preparing watchlist opts however seems like an actual error log because that is never expected to happen or am I missunderstanding things?

pkg/client/client_test.go
```diff
@@ -2199,1464 +2205,1765 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+				Expect(logs).ToNot(ContainElement(MatchRegexp("Response Body.*DeploymentList")))
+			case !watchListEnabled && watchListClientEnabled:
+				// Expect failed watchList + successful fallback to list
+				Expect(logs).To(ContainElements(
+					"\"msg\"=\"Warning: The watchlist request for deployments ended with an error, " +
+						"falling back to the standard LIST semantics\" " +
+						"\"error\"=\"ListOptions.meta.k8s.io \\\"\\\" is invalid: sendInitialEvents: " +
+						"Forbidden: sendInitialEvents is forbidden for watch unless the WatchList feature gate is enabled\""))
+				// Fallback list logs response body
+				Expect(logs).To(ContainElement(MatchRegexp("Response Body.*DeploymentList")))
```
alvaroaleman, line 2597: Its safe to assume this is absent if watchlist is used because then we never get an `ObjectList` type but only event types?

pkg/client/client_test.go
alvaroaleman: Can the TODO go then?

sbueringer: Yup. Copied this from another case and forgot to look up why the TODO was there. Apparently when the TODO was added the test was commented out. Dropped the TODOs

pkg/client/typed_client.go
sbueringer: Took another look. 

> The other log re preparing watchlist opts however seems like an actual error log because that is never expected to happen

I agree. PrepareWatchListOptionsFromListOptions only returns an error if `scheme.Convert` fails.

> this failing is expected if the KAS is too old, right? If yes, then this i IMHO a debug log.

This error will happen either if KAS is <= v1.31 or if the WatchList feature is disabled in KAS.
I personally would never enable the WatchListClient client-side feature gate if I"m not 100% sure that I'm communicating with a v1.32 apiserver. The reason for that is that the fallback behavior means that every single List call will fail with a WatchList first. That's pretty disastrous if it happens in my opinion, so I thought it's good if in these cases the log message shows up (basically it surfaces a client-side configuration error)

pkg/client/client_test.go
sbueringer: A working watchList call logs no ResponseBodies at all.

WatchList: true WatchListClient: true
image

Failing watchList + successful list looks like this

WatchList: false WatchListClient: true
image

I know it's not super great to rely on these logs and the tests might break when we bump, but it's the best thing I found to figure out what if watchList and/or list are called

pkg/client/typed_client.go
alvaroaleman: > I personally would never enable the WatchListClient client-side feature gate if I"m not 100% sure that I'm communicating with a v1.32 apiserver

Right but this will be enabled by default in 1.33 client-go or not? So its not something people will be aware of.

pkg/client/client_test.go
alvaroaleman: Yeah, I am okay with this. An alternate approach would be to inject a custom http roundtripper and look at the requests

pkg/client/typed_client.go
sbueringer: > Right but this will be enabled by default in 1.33 client-go or not? So its not something people will be aware of.

Hm no idea when it will be enabled per default, it's currently not enabled per default in k/k main: https://github.com/kubernetes/kubernetes/blob/a9aab298b4738f4ea9111131cdf193a3b1ba14e5/staging/src/k8s.io/client-go/features/known_features.go#L83

But that's actually a good point. As soon as this is enabled per default, folks will hit these failure cases if they hit an apiserver which is too old which is not great (I think the log message is the smallest problem here :))

@p0lyn0mial I couldn't find a tracking issue for WatchListClient. What are the current plans regarding enabling this per default? Or is the idea to keep it of for the forseeable future to avoid that folks are implicitly hitting these error cases with old apiservers simply by bumping client-go?

pkg/client/client_test.go
sbueringer: Alrighty. I'll keep it as is for now. Let's consider your alternative if this turns out to be too brittle

pkg/client/typed_client.go
sbueringer: Found the tracking issue: https://github.com/kubernetes/enhancements/issues/3157 Still not sure about when WatchListClient would be enabled per default though :)

p0lyn0mial: The plan is to enable the feature (WatchListClient) by default on 1.34 (next release). The issue is quite outdated :)

sbueringer: Okay, I think that means a lot of folks will have to expliclity set the feature gate to false starting with client-go v1.34 to not run into problems.

sbueringer:
/hold 
Until we have consensus if and how we want to support watchlist in our live clients.

caozhuozi:
is it possible we can  have a one-shot global check for whether we have WATCH permission at  the controller-runtime start?

sbueringer:
> is it possible we can have a one-shot global check for whether we have WATCH permission at the controller-runtime start?

In general yes. But we're going to wait to see what client-go decides: https://kubernetes.slack.com/archives/C0EG7JC6T/p1741283769908819?thread_ts=1741283769.908819&cid=C0EG7JC6T

sbueringer:
Just an update. This stays on hold until there is a decision in upstream if the WatchList feature has a future.

Current document for discussion: https://docs.google.com/document/d/1x30DjXSSF5krpyoTCwManJg6vphpnGI37xfCRaiHAbs/edit?resourcekey=0-TluZ5mrM8l9HvkmzyKVfJQ&tab=t.0#heading=h.vvmf7xd5vwsr

sbueringer:
/remove-lifecycle stale

Still on track to be just closed, have to check the current state of upstream at some point

sbueringer:
Took a quick look at k/k. As expected WatchList is only used within the reflector.
/close

PR #3141, opened 2025-03-04
⚠️  NewTypedUnmanaged: Stop requiring a manager

alvaroaleman:
Currently, `NewUnmanaged` and `NewTypedUnmanaged` require a manager but the only properties they ever access from it are `GetControllerOptions` and `GetLogger`. It makes no sense to require a `Manager` for something that is unmanaged, so remove that.

Doing so naively means that we require two options parameters for these constructors, one that has the actual options and another one to default from which again is confusing. Remove the options struct to default from and instead implement a `DefaultFromConfig` that defaults the controller options `config.Controller` which also clarifies what belongs where. This defaulting gets automatically called in the managed constructors and is exported so anyone can use it.

This change is breaking only for users of `New[Typed]Unmanaged` in that:
* The signature changes
* The defaulting of the `controller.Options` from `config.Controller` doesn't happen anymore, but it is still possible to do that explicitly

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 
+	"sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
 	"sigs.k8s.io/controller-runtime/pkg/internal/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -80,13 +81,53 @@ type TypedOptions[request comparable] struct {
...
+
+	// UsePriorityQueue configures the controllers queue to use the controller-runtime provided
+	// priority queue.
+	//
+	// Note: This flag is disabled by default until a future version. It's currently in beta.
+	UsePriorityQueue *bool
+}
+
+// DefaultFromConfig defaults the config from a config.Controller
+func (options *TypedOptions[request]) DefaultFromConfig(config config.Controller) {
```
alvaroaleman, line 99: What is notably absent from this change are tests - That is because this behavior is already verified by existing tests

pkg/controller/controller.go
```diff
@@ -148,18 +190,14 @@ func NewTypedUnmanaged[request comparable](name string, mgr manager.Manager, opt
...
-
 	if options.SkipNameValidation == nil || !*options.SkipNameValidation {
 		if err := checkName(name); err != nil {
 			return nil, err
 		}
 	}
 
 	if options.LogConstructor == nil {
-		log := mgr.GetLogger().WithValues(
+		log := options.Logger.WithValues(
```
sbueringer, line 200: Should we validate at the beginning of the func that Logger.GetSink() is not nil?

At least for the cases where we use it: here and below for PriorityQueue, but maybe we should just make it mandatory so we can rely on it entirely

Although not sure, what happens if folks don't set a Logger?

sbueringer: If I see correctly all other options are optional or we already validate that they are not nil (Reconciler)

pkg/controller/controller.go
```diff
@@ -80,13 +81,53 @@ type TypedOptions[request comparable] struct {
 	// Only use a custom NewQueue if you know what you are doing.
 	NewQueue func(controllerName string, rateLimiter workqueue.TypedRateLimiter[request]) workqueue.TypedRateLimitingInterface[request]
 
+	// Logger is the logger used for this controller
```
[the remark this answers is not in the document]
alvaroaleman: Added docs for the logger, that makes sense. Will respond re logger being mandatory or not in the other comment.

pkg/controller/controller.go
alvaroaleman: > At least for the cases where we use it: here and below for PriorityQueue, but maybe we should just make it mandatory so we can rely on it entirely

Logger is not actually mandatory, this is because `logr.Logger` is not an interface, its a concrete type but unless a sink is set, it won't do anything. I don't know why this design was chosen and to me it seems an interface would've been better, but that is how it is.

The overall behavior remains unchanged: You have to set a sink in one of global log/manager/controller or you will not get any logs. The only thing we have to change is to default the log in `config.Controller` in the `manager` from its own log, which I added.

I am not in favor of making it mandatory and I think it would be very hard to check - The global log gets a delegating log sink by default so I think it will always have _a_ sink but the delegating log sink in turn delegates to a `nullogger` if nothing else was configured within thirty seconds.

sbueringer: Makes sense!

pkg/controller/controller.go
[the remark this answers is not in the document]
sbueringer: I think the comment is slightly off. 

> it will be used by the LogConstructor to construct a log per request.

It's more like if LogConstructor is not set we will create a LogConstructor using that Logger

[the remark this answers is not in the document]
alvaroaleman: Updated the wording, WDYT?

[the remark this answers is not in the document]
sbueringer: Yup, that's fine!

sbueringer:
Thank you!

/lgtm
/approve

PR #3143, opened 2025-03-06
:bug: Fakeclient: Fix dataraces when writing to the scheme

### LINKED ISSUE #3139: Fake client scheme locking is incomplete
In the fake client, we have a codepath where we write to the scheme. We have a `SchemeWriteLock` for that, however there are plenty of other paths that read from the scheme and they do not have any locking at all. As a result, hitting that scheme-writing codepath in parallel with anything else almost always produces a datarace.

We need a RW lock and to properly lock around everything that accesses the scheme.

alvaroaleman:
We have a scheme write lock but plenty of other codeptaths that read from the scheme and that don't do looking, resulting in dataraces if the two happen in parallel.

This change introduces a simple RW lock and makes the fakeclient acquire read locking for all its operations except when needing the write lock. This isn't particularly smart, but given that we only have one codepath that writes to the scheme, it seems good enough.

Fixes #3139 

/cherrypick release-0.20
/cherrypick release-0.19

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -75,8 +75,8 @@ type fakeClient struct {
 	trackerWriteLock sync.Mutex
 	tracker          versionedTracker
 
-	schemeWriteLock sync.Mutex
-	scheme          *runtime.Scheme
+	schemeLock sync.RWMutex
+	scheme     *runtime.Scheme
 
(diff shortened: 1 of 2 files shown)


sbueringer:
Thx!
/lgtm

PR #3149, opened 2025-03-08
⚠ Improve logging by adopting contextual logging

### LINKED ISSUE #2987: Define own WatchErrorHandler for toolscache.Reflector
`client-go` sets a default handler for watching errors `func DefaultWatchErrorHandler(r *Reflector, err error)`, see https://github.com/kubernetes/client-go/blob/23900f492969db045e8a36396852381fd189569b/tools/cache/reflector.go#L154. Right now, if it is included in the `controller-runtime` it will use its logger `klog` by default. This leads to inconsistent log output by default. Therefore, I propose that the `controller-runtime` should overwrite this method, i.e., use its own handler.

This handler could look something like this:

```go
func WatchErrorHandler(r *toolscache.Reflector, err error) {
	log := ctrl.Log.WithValues("name", r.Name(), "description", r.TypeDescription())
	switch {
	case apierrors.IsResourceExpired(err) || apierrors.IsGone(err):
		// Don't set LastSyncResourceVersionUnavailable - LIST call with ResourceVersion=RV already
		// has a semantic that it returns data at least as fresh as provided RV.
		// So first try to LIST with setting RV to resource version of last observed object.
		log.Error(err, "watch closed")
	case err == io.EOF:
		// watch closed normally
	case err == io.ErrUnexpectedEOF:
		log.Error(err, "watch closed with unexpected EOF")
	default:
		utilruntime.HandleError(fmt.Errorf("%s: Failed to watch %v: %v", r.Name(), r.TypeDescription(), err))
	}
}
```
As you can see, we would use a logger with two fields of the reflector now. I exposed these fields recently (see https://github.com/kubernetes/client-go/commit/ce42c297f4b4444f810bedb6fcf10423eebafa5f), so the `controller-runtime` can essentially redefine the default handler of `client-go`, as the default handler also logs these fields.

**I would like to implement this, if it gets your approval.**

sbueringer:

Follow-up to https://github.com/kubernetes-sigs/controller-runtime/pull/3104#discussion_r1946317566
Fixes #2987

### DIFF
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -25,20 +25,25 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/internal/syncs"
 )
+
+var log = logf.RuntimeLog.WithName("cache")
 
 // InformersOpts configures an InformerMap.
 type InformersOpts struct {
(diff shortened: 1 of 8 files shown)


sbueringer:
cc @pohly (just fyi)

pkg/cache/internal/informers.go
```diff
@@ -25,21 +25,26 @@ import (
...
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	logf "sigs.k8s.io/controller-runtime/pkg/internal/log"
 	"sigs.k8s.io/controller-runtime/pkg/internal/syncs"
 )
 
+var log = logf.RuntimeLog.WithName("cache")
```
alvaroaleman, line 46: Obviously nothing to do with this change but we should stop hardcoding loggers like this and make them configurable (they can still default to something like this if unset ofc)

alvaroaleman (approved):

pkg/cache/internal/informers.go
sbueringer: Yeah. I was following existing patterns but I also don't like it.

Btw this logger uses the one from ctrl.logger underneath

PR #3150, opened 2025-03-08
🐛 Fix race in logger test

sbueringer:

Saw tests failing a few times with race condition: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3138/pull-controller-runtime-test/1898308642759774208

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -1092,9 +1092,14 @@ var _ = Describe("manger.Manager", func() {
 				msg := "controller log message"
 				m.GetControllerOptions().Logger.Info(msg)
 
-				Expect(messages).To(ContainElement(
-					ContainSubstring(msg),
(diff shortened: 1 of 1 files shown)


pkg/manager/manager_test.go
```diff
@@ -1092,9 +1092,14 @@ var _ = Describe("manger.Manager", func() {
 				msg := "controller log message"
 				m.GetControllerOptions().Logger.Info(msg)
 
-				Expect(messages).To(ContainElement(
-					ContainSubstring(msg),
-				))
+				Eventually(func(g Gomega) {
```
sbueringer, line 1095: As we had a race condition here I assume it's also necessary to use Eventually

alvaroaleman (approved):

Issue #3133, opened 2025-02-25
RESTMapper randomly returns API versions


stefanprodan:
After #2901 we are observing in [Flux](https://github.com/fluxcd/kustomize-controller/issues/1373) that the RESTMapper can not be used reliably to determine the preferred API version of a custom resource. 

```go
	// RESTMapping identifies a preferred resource mapping for the provided group kind.
	RESTMapping(gk schema.GroupKind, versions ...string) (*RESTMapping, error)
``` 

When calling `RESTMapping` for a `GroupKind` without specified versions (this is what `kubernetes-sigs/cli-utils/pkg/kstatus` does [here](https://github.com/kubernetes-sigs/cli-utils/blob/1a9b04556153279603429504ed7ef75c82be6da6/pkg/kstatus/watcher/object_status_reporter.go#L324)), the controller-runtime RESTMapper will randomly return deprecated versions as the preferred resource mapping.

Given that the `RESTMapping` implementation picks the first mapping from the version array, it suggests that the controller-runtime RESTMapper using the Aggregated Discovery reorders the array or somehow it only contains a single recored even if in etcd the CRD has multiple versions.

```go
func (m *DefaultRESTMapper) RESTMapping(gk schema.GroupKind, versions ...string) (*RESTMapping, error) {
	mappings, err := m.RESTMappings(gk, versions...)
	if err != nil {
		return nil, err
	}
	if len(mappings) == 0 {
		return nil, &NoKindMatchError{GroupKind: gk, SearchedVersions: versions}
	}
	// since we rely on RESTMappings method
	// take the first match and return to the caller
	// as this was the existing behavior.
	return mappings[0], nil
}
```

Logs from Flux kustomize-controller:

```
{"level":"info","ts":"2025-02-24T13:28:10.312Z","logger":"KubeAPIWarningLogger","msg":"v2beta1 HelmRelease is deprecated, upgrade to v2"}
{"level":"info","ts":"2025-02-24T13:30:10.365Z","logger":"KubeAPIWarningLogger","msg":"v2beta2 HelmRelease is deprecated, upgrade to v2"}
{"level":"info","ts":"2025-02-24T13:32:10.423Z","logger":"KubeAPIWarningLogger","msg":"v2beta1 HelmRelease is deprecated, upgrade to v2"}
[...]

PR #3151, opened 2025-03-09
:bug: Restmapper: Respect preferred version

alvaroaleman:
When requesting a resource without version through methods such as `RESTMapping`, the mapper would previously return a random version rather than the preferred one, this change fixes that.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3133
/cherrypick release-0.20

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -246,10 +246,18 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
 		}
 
 		if !found {
-			groupResources.Group.Versions = append(groupResources.Group.Versions, metav1.GroupVersionForDiscovery{
+			gv := metav1.GroupVersionForDiscovery{
 				GroupVersion: metav1.GroupVersion{Group: groupVersion.Group, Version: version}.String(),
 				Version:      version,
-			})
+			}
+
+			// Prepend if preferred version, else append. The upstream DiscoveryRestMappper assumes
+			// the first version is the preferred one: https://github.com/kubernetes/kubernetes/blob/ef54ac803b712137871c1a1f8d635d50e69ffa6c/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L458-L461
+			if group, ok := m.apiGroups[groupVersion.Group]; ok && group.PreferredVersion.Version == version {
+				groupResources.Group.Versions = append([]metav1.GroupVersionForDiscovery{gv}, groupResources.Group.Versions...)
+			} else {
+				groupResources.Group.Versions = append(groupResources.Group.Versions, gv)
+			}
 		}
 
 		// Update data in the cache.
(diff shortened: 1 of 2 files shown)


pkg/client/apiutil/restmapper.go
```diff
@@ -246,10 +246,18 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
...
 
 		if !found {
-			groupResources.Group.Versions = append(groupResources.Group.Versions, metav1.GroupVersionForDiscovery{
+			doc := metav1.GroupVersionForDiscovery{
 				GroupVersion: metav1.GroupVersion{Group: groupVersion.Group, Version: version}.String(),
 				Version:      version,
-			})
+			}
+
+			// Preprend if preferred version, else append. The upstream DiscoveryRestMappper assumes
```
seh, line 254: _s/Preprend/Prepend/_

pkg/client/apiutil/restmapper.go
```diff
@@ -246,10 +246,18 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
 		}
 
 		if !found {
-			groupResources.Group.Versions = append(groupResources.Group.Versions, metav1.GroupVersionForDiscovery{
+			doc := metav1.GroupVersionForDiscovery{
```
erikgb, line 249: What is `doc` a short-name for?

alvaroaleman: document

stefanprodan (approved):
LGTM

I've tested this fix in Flux and the issue with preferred version is gone. Thanks @alvaroaleman 🥇

pkg/client/apiutil/restmapper_test.go
```diff
@@ -740,6 +741,31 @@ func TestLazyRestMapperProvider(t *testing.T) {
...
+
+						mapper, err := apiutil.NewDynamicRESTMapper(restCfg, httpClient)
+						g.Expect(err).NotTo(gmg.HaveOccurred())
+
+						mapping, err := mapper.RESTMapping(schema.GroupKind{
+							Group: "crew.example.com",
+							Kind:  "Driver",
+						})
+						g.Expect(err).NotTo(gmg.HaveOccurred())
+						g.Expect(mapping.GroupVersionKind.Version).To(gmg.Equal("v2"))
```
sbueringer, line 764: Q: Just curious. What makes v2 the preferred version in this test?

Does the apiServer some logic that prefers v2 over v1?

pkg/client/apiutil/restmapper.go
erikgb:
```suggestion
			gv := metav1.GroupVersionForDiscovery{
```

Short for GroupVersion. Just a suggestion. A short variable name makes sense here since the scope of the variable is short. Curious to know what doc/document has to do with this. 🤔

erikgb (commented):
Just a minor nit question/comment from me. Otherwise LGTM! Thanks a lot for looking into this!

pkg/client/apiutil/restmapper_test.go
alvaroaleman: Added a comment

erikgb (commented):
/lgtm

PR #3152, opened 2025-03-11
:seedling: Handlers: Use low priority when object is unchanged and priority queue

alvaroaleman:
This change makes the `TypedFuncs` and `enqueueRequestsFromMapFunc` set a low priority when the object is unchanged by default, as well as extend the test coverage to validate this behavior for `EnqueueRequestForOwner`.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3105

### DIFF
--- a/pkg/handler/enqueue_mapped.go
+++ b/pkg/handler/enqueue_mapped.go
@@ -18,9 +18,11 @@ package handler
 
 import (
 	"context"
+	"reflect"
 
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -81,7 +83,15 @@ func (e *enqueueRequestsFromMapFunc[object, request]) Create(
 	q workqueue.TypedRateLimitingInterface[request],
 ) {
 	reqs := map[request]empty{}
-	e.mapAndEnqueue(ctx, q, evt.Object, reqs)
+
+	var lowPriority bool
+	if reflect.TypeFor[object]().Implements(reflect.TypeFor[client.Object]()) && isPriorityQueue(q) && !isNil(evt.Object) {
+		clientObjectEvent := event.CreateEvent{Object: any(evt.Object).(client.Object)}
+		if isObjectUnchanged(clientObjectEvent) {
+			lowPriority = true
+		}
+	}
+	e.mapAndEnqueue(ctx, q, evt.Object, reqs, lowPriority)
 }
 
 // Update implements EventHandler.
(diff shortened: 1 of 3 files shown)


troy0820:
/lgtm 

Gotcha, yeah I do see what you are saying now @alvaroaleman ! 

Thanks for doing this.

pkg/handler/enqueue_mapped.go
```diff
@@ -81,7 +83,15 @@ func (e *enqueueRequestsFromMapFunc[object, request]) Create(
 	q workqueue.TypedRateLimitingInterface[request],
 ) {
 	reqs := map[request]empty{}
-	e.mapAndEnqueue(ctx, q, evt.Object, reqs)
+
+	var lowPriority bool
+	if reflect.TypeFor[object]().Implements(reflect.TypeFor[client.Object]()) && isPriorityQueue(q) && !isNil(evt.Object) {
```
sbueringer, line 88: I know this is probably a minor optimization, but what do you think about storing the result of `reflect.TypeFor[client.Object]()` in a packge-global unexported variable? This reduces the number of reflection calls quite a bit.

(Fine to keep it if you want, maybe the compiler optimizes it away anyway)

sbueringer: Hm, could it be that the entire check with reflection here doesn't change during the lifetime of enqueueRequestsFromMapFunc? 

If yes, can we just run this once in TypedEnqueueRequestsFromMapFunc and store the result in enqueueRequestsFromMapFunc?

pkg/handler/enqueue_mapped.go
```diff
@@ -112,14 +126,26 @@ func (e *enqueueRequestsFromMapFunc[object, request]) Generic(
...
+	e.mapAndEnqueue(ctx, q, evt.Object, reqs, false)
 }
 
-func (e *enqueueRequestsFromMapFunc[object, request]) mapAndEnqueue(ctx context.Context, q workqueue.TypedRateLimitingInterface[request], o object, reqs map[request]empty) {
+func (e *enqueueRequestsFromMapFunc[object, request]) mapAndEnqueue(
+	ctx context.Context,
+	q workqueue.TypedRateLimitingInterface[request],
+	o object,
+	reqs map[request]empty,
+	unchanged bool,
```
sbueringer, line 137: nit. I would also call this lowPriority (like the variables we're passing in)

pkg/handler/eventhandler_test.go
```diff
@@ -775,6 +775,216 @@ var _ = Describe("Eventhandler", func() {
...
+
+					Expect(q.Len()).To(Equal(1))
+					item, _ := q.Get()
+					Expect(item).To(Equal(reconcile.Request{NamespacedName: types.NamespacedName{Name: "my-pod"}}))
+				})
+			})
+		}
+	})
+
 	Describe("WithLowPriorityWhenUnchanged", func() {
```
sbueringer, line 988: Nice! Can we please remove the now redundant tests below?

sbueringer:
Otherwise all good

sbueringer:
Comments have been addressed via https://github.com/kubernetes-sigs/controller-runtime/pull/3160

Thx!

PR #3156, opened 2025-03-14
🌱 Add priority label to PQ depth metric

sbueringer:

### DIFF
--- a/pkg/internal/metrics/workqueue.go
+++ b/pkg/internal/metrics/workqueue.go
@@ -17,6 +17,8 @@ limitations under the License.
 package metrics
 
 import (
+	"strconv"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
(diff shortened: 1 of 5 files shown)


sbueringer:
/assign @alvaroaleman 
for an initial review

/hold

pkg/internal/metrics/workqueue.go
```diff
@@ -42,9 +44,15 @@ var (
 	depth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Subsystem: WorkQueueSubsystem,
 		Name:      DepthKey,
-		Help:      "Current depth of workqueue",
+		Help:      "Current depth of workqueue by workqueue",
 	}, []string{"name", "controller"})
 
+	depthWithPriority = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: WorkQueueSubsystem,
+		Name:      DepthKey,
```
alvaroaleman, line 52: Are you sure that prometheus allows to register the same metric twice in general and in particular with different labels?

alvaroaleman (commented):
Looks good but one comment re double registration

pkg/internal/metrics/workqueue.go
sbueringer: Very good point. This only "worked" because I forgot to register the new metric

sbueringer: Now I'm only using one metric. The result is the following

metrics endpoint returns

```
# controller with enabled PQ:
workqueue_depth{controller="cluster",name="cluster",priority="0"} 0
# controller with disabled PQ:
workqueue_depth{controller="machine",name="machine",priority=""} 0
```

Prometheus (apprently doesn't show the empty priority label)

image

alvaroaleman (approved):

PR #3160, opened 2025-03-15
:seedling: Followups to default low priority in mappers

alvaroaleman:
Addresses Stefans comments in https://github.com/kubernetes-sigs/controller-runtime/pull/3152

/hold


pkg/handler/eventhandler.go
```diff
@@ -117,7 +123,7 @@ func isPriorityQueue[request comparable](q workqueue.TypedRateLimitingInterface[
 // Create implements EventHandler.
 func (h TypedFuncs[object, request]) Create(ctx context.Context, e event.TypedCreateEvent[object], q workqueue.TypedRateLimitingInterface[request]) {
```
alvaroaleman, line 124: Doing or caching this directly in the `Funcs` is problematic:
* It doesn't have a constructor, so can't do it there (We have to remember to never ever export types and only constructors for this kind of scenario)
* It has value receivers and we need some kind of locking if we want to cache this, but using a lock with value receivers means we copy the lock which the linter forbids

sbueringer:
Thank you!

/lgtm
/approve

Issue #3164, opened 2025-03-18
Adopt Native Histograms


krisztianfekete:
Hello,

I'd like to introduce [Native Histograms](https://prometheus.io/docs/specs/native_histograms/) in the project, so histograms are are also exposed without all the limitations of traditional histograms (cardinality problems, bad bucket layout, etc.).

Adopting them provides a way to use them in an opt-in manner, and users can also[ scrape both old and new histograms](https://prometheus.io/docs/specs/native_histograms/#scraping-both-classic-and-native-histograms).

alvaroaleman:
This was done in [#3165](https://github.com/kubernetes-sigs/controller-runtime/pull/3165)

krisztianfekete:
Hey @alvaroaleman, I am interested in getting this into an upcoming release. Can you please help me out with the process around that?

alvaroaleman:
It will be in the next minor release

PR #3165, opened 2025-03-18
✨ Adopt native histograms

krisztianfekete:
Adopts Native Histograms in an opt-in and backward-compatible manner to overcome the limitations of traditional histograms.

This solves https://github.com/kubernetes-sigs/controller-runtime/issues/3164

### DIFF
--- a/pkg/internal/controller/metrics/metrics.go
+++ b/pkg/internal/controller/metrics/metrics.go
@@ -17,6 +17,8 @@ limitations under the License.
 package metrics
 
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -60,6 +62,9 @@ var (
 		Help: "Length of time per reconciliation per controller",
 		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
 			1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60},
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: 1 * time.Hour,
 	}, []string{"controller"})
 
 	// WorkerCount is a prometheus metric which holds the number of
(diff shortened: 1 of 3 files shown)


pkg/internal/controller/metrics/metrics.go
```diff
@@ -60,6 +62,9 @@ var (
 		Help: "Length of time per reconciliation per controller",
 		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
 			1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60},
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: 1 * time.Hour,
```
alvaroaleman, line 67: If we do this, can we please do it for all histograms and not just one? And what was the reasoning behind choosing that `NativeHistogramBucketFactor` and `NativeHistogramMaxBucketNumber`?

krisztianfekete: Sure, let me also add it for workqueue metrics too!
The values I've set are sensible defaults that most project starts with (us included).

Please see the following references, reasoning:

- [NativeHistogramBucketFactor](https://github.com/prometheus/client_golang/blob/331dfab0cc853dca0242a0d96a80184087a80c1d/prometheus/histogram.go#L405-L429)
- NativeHistogramMaxBucketNumber: I've seen this being set to 100 usually, unless it is desired to mimic Exponential Histograms in OTel closer, in which case this can also be 160.
- NativeHistogramMinResetDuration: "_If at least NativeHistogramMinResetDuration has passed since the last reset of the histogram (which includes the creation of the histogram), the whole histogram is reset, i.e. all buckets are deleted and the sum and count of observations as well as the zero bucket are set to zero. Prometheus handles this as a normal counter reset, which means that some observations will be lost between scrapes, so resetting should happen rarely compared to the scraping interval. Additionally, frequent counter resets might lead to less efficient storage in the TSDB (see the [TSDB section](https://prometheus.io/docs/specs/native_histograms/#tsdb) for details). A NativeHistogramMinResetDuration of one hour is a value that should work well in most situations_." source: https://prometheus.io/docs/specs/native_histograms/#limiting-the-bucket-count

alvaroaleman (approved):
/hold

Thanks, makes sense to me.

@sbueringer any concerns here?

sbueringer:
I'll take a look

krisztianfekete:
Thanks, @sbueringer, looking forward to the review!

sbueringer (commented):
Took a closer look and tested it with Prometheus / Grafana / Cluster API. Looks nice!

One smaller finding

pkg/internal/controller/metrics/metrics.go
krisztianfekete: Hah, not sure how could I miss that. Thanks, just pushed a commit!

Issue #3158, opened 2025-03-14
setup-envtest: add version option to print installed version


grosser:
I want to make sure the correct version is installed without having to re-install it every time,
currently --help does not how the version that is installed though

PR #3166, opened 2025-03-19
✨Add RELEASE_TAG to tools/setup-envtest to show binary version with setup-envtest version

troy0820:
With the setup-envtest binary, it is difficult to determine the version of the binary that is installed.  This will allow the user to use the command `setup-envtest version` to see what version of the binary is installed.

~~**Note: This is targeting the branch when built and not a specific tag.  The setup-envtest instructions for installation shows to target the release branch for specific versions**~~

**This will target the release tag defined in the GitHub action when binaries are built.**

Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3158

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,7 +174,7 @@ release-binary: $(RELEASE_DIR)
 		-v "$$(pwd):/workspace$(DOCKER_VOL_OPTS)" \
 		-w /workspace/tools/setup-envtest \
 		golang:$(GO_VERSION) \
-		go build -a -trimpath -ldflags "-extldflags '-static'" \
+		go build -a -trimpath -ldflags "-X 'sigs.k8s.io/controller-runtime/tools/setup-envtest/version.version=$(RELEASE_TAG)' -extldflags '-static'" \
 		-o ./out/$(RELEASE_BINARY) ./
 
 ## --------------------------------------
--- a/tools/setup-envtest/main.go
+++ b/tools/setup-envtest/main.go
@@ -184,6 +184,9 @@ Commands:
 		reads a .tar.gz file from stdin and expand it into the store.
 		must have a concrete version and platform.
 
+	version:
+		list the installed version of setup-envtest.
+
 Versions:
 
 	Versions take the form of a small subset of semver selectors.
@@ -256,7 +259,6 @@ Environment Variables:
 		version = flag.Arg(1)
 	}
 	env := setupEnv(globalLog, version)
-
 	// perform our main set of actions
 	switch action := flag.Arg(0); action {
 	case "use":
@@ -274,6 +276,8 @@ Environment Variables:
 			Input:       os.Stdin,
 			PrintFormat: printFormat,
 		}.Do(env)
+	case "version":
+		workflows.Version{}.Do(env)
 	default:
 		flag.Usage()
 		envp.Exit(2, "unknown action %q", action)
--- a/tools/setup-envtest/version/version.go
+++ b/tools/setup-envtest/version/version.go
@@ -1,0 +1,21 @@
+package version
+
+import "runtime/debug"
+
+// Version to be set using ldflags:
+// -ldflags "-X sigs.k8s.io/controller-tools/pkg/version.version=v1.0.0"
+// falls back to module information is unse
+var version = ""
+
+// Version returns the version of the main module
+func Version() string {
+	if version != "" {
+		return version
+	}
+	info, ok := debug.ReadBuildInfo()
+	if !ok || info == nil || info.Main.Version == "" {
+		// binary has not been built with module support or doesn't contain a version.
+		return "(unknown)"
+	}
+	return info.Main.Version
+}
--- a/tools/setup-envtest/workflows/workflows.go
+++ b/tools/setup-envtest/workflows/workflows.go
@@ -5,11 +5,13 @@ package workflows
 
 import (
 	"context"
+	"fmt"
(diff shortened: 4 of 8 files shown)


sbueringer:
I think setup-envtest show the exact version. We are also publishing setup-envtest binaries on every release and it's not ideal if that just shows the branch

(there was some similar work in controller-gen recently: https://github.com/kubernetes-sigs/controller-tools/issues/1046. In general I would expect setup-envtest and controller-gen to behave the same)

troy0820:
> I think setup-envtest should show the exact version. We are also publishing setup-envtest binaries on every release and it's not ideal if that just shows the branch
> 
> (there was some similar work in controller-gen recently: [kubernetes-sigs/controller-tools#1046](https://github.com/kubernetes-sigs/controller-tools/issues/1046). In general I would expect setup-envtest and controller-gen to behave the same)

This makes sense.  I made the assumption with how we install to provide the same installation version technique.  I can change this to provide the same as controller-gen.

tools/setup-envtest/workflows/workflows.go
```diff
@@ -85,3 +87,23 @@ func (f Sideload) Do(env *envp.Env) {
...
+	BinaryVersion string
+}
+
+// Do executes the workflow.
+func (v Version) Do(env *envp.Env) {
+	version := "(unknown)"
+	if v.BinaryVersion == "" {
+		info, ok := debug.ReadBuildInfo()
+		if (ok && info != nil) || info.Main.Version != "" {
+			// binary has not been built with module support or doesn't contain a version.
```
troy0820, line 103:
```suggestion
			// binary has been built with module support or doesn't contain a version.
```

Makefile
```diff
@@ -174,7 +174,7 @@ release-binary: $(RELEASE_DIR)
 		-v "$$(pwd):/workspace$(DOCKER_VOL_OPTS)" \
 		-w /workspace/tools/setup-envtest \
 		golang:$(GO_VERSION) \
-		go build -a -trimpath -ldflags "-extldflags '-static'" \
+		go build -a -trimpath -ldflags "-X 'main.BranchVersion=$(RELEASE_TAG)' -extldflags '-static'" \
```
sbueringer, line 177: Let's call this version instead of BranchVersion

I assume this works, do you know if we can also use something more like in controller-gen?

e.g. sigs.k8s.io/controller-runtime/tools/setup-envtest/main.version=

Maybe also move the const to a tools/setup-envtest/version package, but not sure what the best practices are there

tools/setup-envtest/main.go
```diff
@@ -51,6 +51,9 @@ var (
 		"directory to store binary assets (default: $OS_SPECIFIC_DATA_DIR/envtest-binaries)")
 
 	index = flag.String("index", remote.DefaultIndexURL, "index to discover envtest binaries")
+
+	// BranchVersion is the current setup-envtest branch that is used for the binary version of setup-envtest.
```
sbueringer, line 55: Let's add a godoc similar to controller-gen

```
// version to be set using ldflags:
// -ldflags "-X sigs.k8s.io/controller-tools/pkg/version.version=v1.0.0"
// falls back to module information is unset
```

tools/setup-envtest/main.go
```diff
@@ -274,6 +276,10 @@ Environment Variables:
 			Input:       os.Stdin,
 			PrintFormat: printFormat,
 		}.Do(env)
+	case "version":
```
sbueringer, line 279: Just to double check. This will result in a `setup-envtest version` command right?

tools/setup-envtest/workflows/workflows.go
```diff
@@ -85,3 +87,23 @@ func (f Sideload) Do(env *envp.Env) {
...
+
+// Version is the workflow that shows the current binary version
+// of setup-envtest.
+type Version struct {
+	BinaryVersion string
+}
+
+// Do executes the workflow.
+func (v Version) Do(env *envp.Env) {
+	version := "(unknown)"
```
sbueringer, line 99: Maybe also move this logic to the version pkg I suggested in another comment. Then the stamping and this logic is in the same place

Maybe we can just use literally the same version.go as in controller-gen (+/- controller-tools specific godoc)? and then call version.Version() here instead of passing the version through?

(this func here would then ~ only contain the print statement)

tools/setup-envtest/main.go
troy0820: This is correct. I need to update the commands to express this.

sbueringer:
Thx!

I'll cherry-pick so we can verify this e2e with the next patch release

/cherry-pick release-0.20

tools/setup-envtest/version/version.go
```diff
@@ -0,0 +1,21 @@
+package version
+
+import "runtime/debug"
+
+// Version to be set using ldflags:
+// -ldflags "-X sigs.k8s.io/controller-tools/pkg/version.version=v1.0.0"
```
troy0820, line 6:
```suggestion
// -ldflags "-X sigs.k8s.io/controller-runtime/tools/setup-envtest/version.version=v1.0.0"

````

troy0820: @sbueringer I can make a PR for this to fix the godoc rendering with the correct comment.

sbueringer: Ups. Yup, thx!

PR #3167, opened 2025-03-21
:sparkles: Controller: Retain the priority

alvaroaleman:
This change makes the controller retain the priority if a priority queue is used. The priority queue will still bump the priority if the item gets re-added to it with a higher priority.

/assign @sbueringer 
/hold

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/util/workqueue"
 
+	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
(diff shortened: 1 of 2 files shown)


pkg/internal/controller/controller.go
```diff
@@ -157,7 +158,12 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 	// Set the internal context.
 	c.ctx = ctx
 
-	c.Queue = c.NewQueue(c.Name, c.RateLimiter)
+	queue := c.NewQueue(c.Name, c.RateLimiter)
+	if priorityQueue, isPriorityQueue := queue.(priorityqueue.PriorityQueue[request]); isPriorityQueue {
```
alvaroaleman, line 162: I found it easier to use a wrapper if we don't get a priority queue vs asserting the queue type everywhere where we access it

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

PR #3185, opened 2025-03-31
:seedling: Clarify that controller-runtime is not a kubebuilder subproject

alvaroaleman:
This used to be true in the very beginning of the projects but isn't anymore today, there are different people working on these projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,6 @@ See [FAQ.md](FAQ.md)
 
 Learn how to engage with the Kubernetes community on the [community page](http://kubernetes.io/community/).
 
-controller-runtime is a subproject of the [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) project
-in sig apimachinery.
-
 You can reach the maintainers of this project at:
 
 - Slack channel: [#controller-runtime](https://kubernetes.slack.com/archives/C02MRBMN00Z)


README.md
```diff
@@ -68,9 +68,6 @@ See [FAQ.md](FAQ.md)
 
 Learn how to engage with the Kubernetes community on the [community page](http://kubernetes.io/community/).
 
-controller-runtime is a subproject of the [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) project
-in sig apimachinery.
```
camilamacedo86, line 72: I believe the way we maintain the projects — whether or not they share maintainers — doesn’t define whether they are sub-projects of Kubebuilder. We may have the same maintainers or not, and that’s fine.

However, changes in controller-runtime should still consider their impact on Kubebuilder, which is the tool — maintained under the Kubernetes SIGs — that builds on top of and has been promoting this library for years.

**IMO: This isn’t about OWNER aliases, but rather about the purpose and scope of the tools and projects themselves.**

While controller-runtime and controller-tools can technically be used independently of Kubebuilder, it's clear that Kubebuilder has been the primary driver of their adoption. In practice, usage outside of Kubebuilder seems quite limited.


So, I’d suggest we rephrase it to something like:

```
[Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) is built on top of the controller-runtime and controller-tools libraries, providing a higher-level scaffolding tool to help developers create Kubernetes APIs more easily.
```

camilamacedo86: > That is already in the readme: https://github.com/kubernetes-sigs/controller-runtime/blame/d8b679322d5cc466d12301b8875c42040fd29e46/README.md#L6-L8


I see — that makes sense, and that’s great to hear! 🎉  

I agree that `controller-runtime` and `controller-tools` can be used independently of Kubebuilder, so I’m totally fine with the clarification.

/lgtm  
/approve  

That said, Kubebuilder has also been actively documenting and promoting both libraries over the years. We'd really love to receive help, feedback, and suggestions from the maintainers of `controller-runtime` and `controller-tools` to improve Kubebuilder’s scaffolding, documentation, and overall user experience.

We truly want to work together to ensure the success of all the projects and provide the best experience for the community. 🙌

PR #3186, opened 2025-03-31
:sparkles: Allow panic as log level

rashmichandrashekar:
This allows logging to be disabled for all but panic log entries.

The Zap Level config parameter (https://github.com/uber-go/zap/blob/6d482535bdd97f4d97b2f9573ac308f1cf9b574e/config.go#L62) isn't available via the Option interface, so it's not possible to disable it completely via a user-added flag.

### DIFF
--- a/pkg/log/zap/flags.go
+++ b/pkg/log/zap/flags.go
@@ -32,6 +32,7 @@ var levelStrings = map[string]zapcore.Level{
 	"debug": zap.DebugLevel,
 	"info":  zap.InfoLevel,
 	"error": zap.ErrorLevel,
+	"panic": zap.PanicLevel,
 }
 
 var stackLevelStrings = map[string]zapcore.Level{
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -247,7 +247,7 @@ func NewRaw(opts ...Opts) *zap.Logger {
 //     Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn)
 //     Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
 //   - zap-encoder: Zap log encoding (one of 'json' or 'console')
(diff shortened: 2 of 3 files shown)


troy0820:
I am not sure that we want that to be configured to panic but I could be wrong.  

/hold

rashmichandrashekar:
> I am not sure that we want that to be configured to panic but I could be wrong.
> 
> /hold

Thanks @troy0820. Its an option that we can leverage while using this module. We have a dependency on an OSS component - opentelemetry-operator from which we utilize some of the features, while it provides a lot more capabilities. Since we are not opting in to all of them, it throws errors which fill up the logs and we would really like to have a way to configure this. 
Since this is not mandatory and it is actually supported in the configuration, this change is to just make it available with an option. Hope this helps.

rashmichandrashekar:
@troy0820 - Following up, gentle ping

rashmichandrashekar:
> @troy0820 - Following up, gentle ping

@troy0820 - Following up, gentle ping. Please let me know what the next steps should be.

troy0820:
/ok-to-test

I will have to consult with @sbueringer on this one but I can put ok-to-test.

rashmichandrashekar:
@troy0820 - Following up, the tests have passed, pls let me know what the next steps should be, Thanks!

rashmichandrashekar:
@troy0820 @sbueringer - Gentle ping. Please let me know if there are concerns. Tests have passed, waiting on approval. Thanks!

rashmichandrashekar:
@inteon @joelanford - Could you pls take a look at this PR? I see that you have been added as reviewers.
Thanks in advance!

sbueringer:
/assign 
(back from PTO)

sbueringer:
This seems fine to me. We're basically just exposing an additional value of the underlying zap library

/lgtm



/assign @alvaroaleman 
Just in case I'm missing something, IIRC you have more experience with zap

alvaroaleman (approved):

PR #3190, opened 2025-04-05
✨ [Warm Replicas] Extract startWatches into helper method.

godwinpang:
This PR is a small refactor of the controller code.

It's the first step to implementing https://github.com/kubernetes-sigs/controller-runtime/pull/3121 as the watches need to be started in two places, so pulling this into its own method.
1) An warmup method if the controller is a warm replica
2) When the controller wins leader election (current behavior)


alvaroaleman (approved):

sbueringer:
Thx!
/lgtm

Issue #1571, opened 2021-06-29
Timeout waiting for process kube-apiserver to stop


ysksuzuki:
#### What happened:

`envtest.Environment.Stop()` raises the error `timeout waiting for process kube-apiserver to stop` with kube-apiserver 1.21.2 that set up using `setup-envtest use -p env`.

```
  Unexpected error:
      <errors.aggregate | len:1, cap:1>: [
          <*errors.errorString | 0xc000328c00>{
              s: "timeout waiting for process kube-apiserver to stop",
          },
      ]
      timeout waiting for process kube-apiserver to stop
  occurred
```

#### How to reproduce it (as minimally and precisely as possible):

Run [this envtest](https://github.com/cybozu-go/coil/blob/main/v2/Makefile#L36) with kube-apiserver 1.21.2.

#### Environment:

- controller-runtime: 0.9.2
- kube-apiserver: 1.21.2

stijndehaes:
using 1.20.2 for example works fine

varshaprasad96:
+1, hit the same issue and using 1.20 binaries works fine.

erikgb:
+1, it would be nice if someone with the skills could take a look asap. 😉 

/priority important-soon

joelanford:
I'm not able to reproduce this issue with the envtest 1.22.0 binaries and controller-runtime v0.9.6.

To try to reproduce, I bumped to the latest k8s, envtest, and c-r versions [here](https://github.com/operator-framework/helm-operator-plugins/pull/102)

hickeyma:
Tested using the [cronjob-tutorial](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project), results as follows:

- Kubernetes v1.20.2: No issue
- Kubernetes v1.21.4: Issue as stated 
- Kubernetes v1.22.0: Issue as stated

So, looks like issue in Kubernetes v1.21, v1.22.

hickeyma:
This is what I have found  so far. From Kubernetes 1.21+, when it tries to cleanup the test environment, there is a clash if a custom controller is created during testing. It would seem that the controller is still running and `kube-apiserver` will not respond to shutdown during tear down.

I can get it to work by, changing controller goroutine:

```go
go func() {
		err = k8sManager.Start(ctrl.SetupSignalHandler())
		Expect(err).ToNot(HaveOccurred())
}()
```

to:

```go
go func() {
[...]

PR #3192, opened 2025-04-09
:warning: [Warm Replicas] Implement warm replica support for controllers.

godwinpang:
This change implements the proposal for warm replicas as proposed in #3121.
It adds a `NeedWarmUp` option for controllers to optionally start as warmed replicas.

Note for reviewers: This draft PR is feature complete with tests but the main purpose is to make sure that things are on the right track. Some of the naming / comments are a bit inconsistent, I will address them in a followup cleanup tomorrow.

Builds upon #3190.


sbueringer:
I would recommend using  the [WIP] PR title prefix instead of draft so prow runs unit tests

pkg/config/controller.go
```diff
@@ -60,6 +60,10 @@ type Controller struct {
 	// Defaults to true, which means the controller will use leader election.
 	NeedLeaderElection *bool
 
+	// NeedWarmUp indicates whether the controller needs to use warm up.
+	// Defaults to false, which means the controller will not use warm up.
+	NeedWarmUp *bool
```
alvaroaleman, line 65: I would call it something like `EnableWarmup`, not `Need` - The difference to LeaderElection is that LeaderElection is a dependency of the controller, vs warmup is a property of it

pkg/manager/internal.go
```diff
@@ -439,6 +439,11 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 		return fmt.Errorf("failed to start other runnables: %w", err)
 	}
 
+	// Start and wait for sources to start.
+	if err := cm.runnables.Warmup.Start(cm.internalCtx); err != nil {
```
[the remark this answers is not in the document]
godwinpang: Will change to coming right after the caches, but don't completely understand the significance of having it after vs. before the non-leader election runnables; mind explaining a bit?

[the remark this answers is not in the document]
godwinpang: > This needs to block until the Warmup has terminated, otherwise we may end up starting the controller before the sources are started, as the check we added only checks if we started to start the sources, not if we finished doing so

Is this a big problem? Replicas will only start the controller after they win leader election, are you saying that in non-leader election cases the behavior of warmup should be that it completely blocks controller startup?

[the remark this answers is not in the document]
alvaroaleman: > Will change to coming right after the caches, but don't completely understand the significance of having it after vs. before the non-leader election runnables; mind explaining a bit?

Just so it starts asap

> Is this a big problem? Replicas will only start the controller after they win leader election so I don't see an issue in the leader election failover case; are you saying that in non-leader election cases the behavior of warmup should be that it completely blocks controller startup?

Yes. The expectation in the controller is that the sources are fully started before the rest of the controller starts. If we don't ensure this actually finished before we start the controller, we violate that expectation.

Additionally, there may be other cases where ppl implement this interface and it seems surprising to me that we wouldn't let the warmup finish before we start.

If we wanted to be really fancy, we could do this per-runnable (i.E. not block all runnables on `Warmup` having finished, just the ones whose `Warmup` didn't finish) but that is probably more involved.

pkg/internal/controller/controller.go
```diff
@@ -144,6 +167,40 @@ func (c *Controller[request]) NeedLeaderElection() bool {
 	return *c.LeaderElected
 }
 
+// Warmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) Warmup(ctx context.Context) error {
+	if c.NeedWarmup == nil || !*c.NeedWarmup {
+		c.didEventSourcesFinishSync.Store(ptr.To(true))
```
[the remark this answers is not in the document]
godwinpang: because I had a brainfart while writing the UT, sorry!

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1014,82 @@ var _ = Describe("controller", func() {
 			})
 		})
 	})
+
+	Describe("Warmup", func() {
```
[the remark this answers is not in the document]
godwinpang: done

sbueringer:
@godwinpang  I'll review once the current changes are pushed

pkg/manager/manager.go
```diff
@@ -314,6 +314,17 @@ type LeaderElectionRunnable interface {
...
 
+// WarmupRunnable knows if a Runnable should be a warmup runnable. A warmup runnable is a runnable
+// that should be run when the manager is started, but before the leader election is acquired.
+type WarmupRunnable interface {
+	// Warmup is the implementation of the warmup runnable.
+	Warmup(context.Context) error
+
+	// WaitForWarmupComplete is a blocking function that waits for the warmup to be completed. It
+	// returns false if it could not successfully finish warmup.
+	WaitForWarmupComplete(context.Context) bool
```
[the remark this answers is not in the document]
godwinpang: That was the only way I could figure out how to get it to work with how the runnable groups are setup. My reading of the code for runnable groups is that when we [Start](https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/manager/runnable_group.go#L132) a runnable group, it starts all the runnables (calling [reconcile](https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/manager/runnable_group.go#L185)) then blocks on the runnableChecks before returning. The runnable check is the only mechanism that can block a runnableGroup from prematurely returning.

I copied this `WaitForWarmupComplete` naming + way of blocking from how the cache syncs was implemented [here](https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/manager/runnable_group.go#L63-L65), lmk if you have a better way of implementing this.

godwinpang:
> @godwinpang I'll review once the current changes are pushed (back from PTO and I have finally time now :))

Just in time :) It should be good for a pass now, I've addressed all the comments from @alvaroaleman's first review other than the outstanding one regarding implementing warmup as a blocking runnable group

pkg/manager/internal.go
```diff
@@ -439,6 +439,11 @@ func (cm *controllerManager) Start(ctx context.Context) (err error) {
 		return fmt.Errorf("failed to start other runnables: %w", err)
 	}
 
+	// Start and wait for sources to start.
```
sbueringer, line 442:
```suggestion
	// Start WarmupRunnables and wait for warmup to complete.
```

pkg/manager/internal.go
```diff
@@ -544,6 +549,10 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 		cm.runnables.LeaderElection.startOnce.Do(func() {})
 		cm.runnables.LeaderElection.StopAndWait(cm.shutdownCtx)
 
+		// Stop the warmup runnables
+		cm.logger.Info("Stopping and waiting for warmup runnables")
+		cm.runnables.Warmup.StopAndWait(cm.shutdownCtx)
```
sbueringer, line 554: Just to confirm my understanding. This will:
* cancel the context that was used to call Warmup()
* wait until all Warmup funcs returned

On the happy path where all Warmup funcs completed during controller startup this will do effectively nothing

pkg/internal/controller/controller.go
```diff
@@ -144,6 +167,38 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
+// Warmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) Warmup(ctx context.Context) error {
+	if c.NeedWarmup == nil || !*c.NeedWarmup {
+		return nil
+	}
+	return c.startEventSources(ctx)
+}
+
+// DidFinishWarmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) DidFinishWarmup(ctx context.Context) bool {
```
sbueringer, line 179: Probably dead code, but should we handle the `if c.NeedWarmup == nil || !*c.NeedWarmup` case here in some way specifically?

pkg/manager/internal.go
```diff
@@ -544,6 +549,10 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 		cm.runnables.LeaderElection.startOnce.Do(func() {})
 		cm.runnables.LeaderElection.StopAndWait(cm.shutdownCtx)
 
+		// Stop the warmup runnables
```
sbueringer, line 552: Should this be done after l.545? (and before we stop leaderelection runnables)

pkg/manager/manager.go
```diff
@@ -314,6 +314,19 @@ type LeaderElectionRunnable interface {
 	NeedLeaderElection() bool
 }
 
+// WarmupRunnable knows if a Runnable should be a warmup runnable. A warmup runnable is a runnable
```
sbueringer, line 317:
```suggestion
// WarmupRunnable knows if a Runnable requires warmup. A warmup runnable is a runnable
```

pkg/manager/manager.go
```diff
@@ -314,6 +314,19 @@ type LeaderElectionRunnable interface {
...
+// that should be run when the manager is started, but before the leader election is acquired.
+// Note: Implementing this interface is only useful when LeaderElection is enabled, as the behavior
```
sbueringer, line 319: nit: Yes and no. Often leader election can be enabled/disabled via command line flags. But at this time the interface is already implemented or not.

Maybe slightly reframe to "when LeaderElection can be enabled"?

pkg/manager/manager.go
```diff
@@ -314,6 +314,19 @@ type LeaderElectionRunnable interface {
...
+// without leaderelection is to run LeaderElectionRunnables immediately.
+type WarmupRunnable interface {
+	// Warmup is the implementation of the warmup runnable.
```
sbueringer, line 322: The comment doesn't really say much :)

pkg/manager/manager.go
```diff
@@ -314,6 +314,19 @@ type LeaderElectionRunnable interface {
 	NeedLeaderElection() bool
 }
 
+// WarmupRunnable knows if a Runnable should be a warmup runnable. A warmup runnable is a runnable
+// that should be run when the manager is started, but before the leader election is acquired.
+// Note: Implementing this interface is only useful when LeaderElection is enabled, as the behavior
+// without leaderelection is to run LeaderElectionRunnables immediately.
+type WarmupRunnable interface {
```
sbueringer, line 321: Do we want to export this interface?

If we only need it for controller we could probably keep it unexported for now (like "hasCache")

Also allows us refactorings going forward if necessary

pkg/manager/manager_test.go
```diff
@@ -1928,6 +1928,75 @@ var _ = Describe("manger.Manager", func() {
...
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		go func() {
+			defer GinkgoRecover()
+			Expect(m.Start(ctx)).NotTo(HaveOccurred())
+		}()
+
+		By("Waiting for the warmup runnable to be called")
```
sbueringer, line 1991: I think this does not actually verify the order. Just that all channels are closed.

e.g. the following also works
```go
		By("Verifying the leader election runnable is called after election")
		<-leaderElectionRunnableCalled

		By("Waiting for leader election to be won")
		<-m.Elected()

		By("Waiting for the warmup runnable to be called")
		<-warmupCalled
```

pkg/manager/runnable_group_test.go
```diff
@@ -224,3 +348,64 @@ var _ = Describe("runnableGroup", func() {
...
+func (r leaderElectionRunnableFunc) NeedLeaderElection() bool {
+	return r.NeedLeaderElectionFunc()
+}
+
+var _ WarmupRunnable = &warmupRunnableFunc{}
+
+// warmupRunnableFunc is a helper struct that implements WarmupRunnable
+// for testing purposes.
+type warmupRunnableFunc struct {
+	RunFunc    func(context.Context) error
```
sbueringer, line 374: nit: Let's call this StartFunc, also in leaderElectionRunnableFunc & combinedRunnable

pkg/manager/runnable_group_test.go
```diff
@@ -53,6 +53,130 @@ var _ = Describe("runnables", func() {
...
+			},
+			WarmupFunc: func(c context.Context) error {
+				return nil
+			},
+		}
+
+		r := newRunnables(defaultBaseContext, errCh)
+		Expect(r.Add(warmupRunnable)).To(Succeed())
+		Expect(r.Warmup.startQueue).To(HaveLen(1))
+		Expect(r.LeaderElection.startQueue).To(HaveLen(1))
```
sbueringer, line 71: Let's verify length of r.Others.startQueue in all these test cases as well

pkg/internal/controller/controller.go
sbueringer: Hm. Maybe we have to.

Controller will always be added to the Warmup runnable group.

If Warmup is a no-op, DidFinishWarmup will never return, right?

Which means the manager will get stuck when starting the warmup runnable group

pkg/manager/runnable_group_test.go
```diff
@@ -224,3 +348,64 @@ var _ = Describe("runnableGroup", func() {
...
+
+func (r warmupRunnableFunc) Warmup(ctx context.Context) error {
+	return r.WarmupFunc(ctx)
+}
+
+func (r warmupRunnableFunc) WaitForWarmupComplete(ctx context.Context) bool {
+	return true
+}
+
+// combinedRunnable implements both WarmupRunnable and LeaderElectionRunnable
```
sbueringer, line 390: I would prefer leaderElectionAndWarmupRunnable

sbueringer:
Mostly minor findings

pkg/manager/manager.go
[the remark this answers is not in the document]
alvaroaleman: > The runnable check is the only mechanism that can block a runnableGroup from prematurely returning.

Yes, because this use-case is new. It shouldn't be too difficult to also start tracking which runnables have ended and then compare that number to the total in the group.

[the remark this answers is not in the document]
sbueringer: > Yes, because this use-case is new. It shouldn't be too difficult to also start tracking which runnables have ended and then compare that number to the total in the group.

Just wanted to note that other runnables usually don't end. Start is blocking for them

pkg/internal/controller/controller.go
alvaroaleman: Yeah we need some way to short-circuit this code if the warmup is not enabled

pkg/manager/manager.go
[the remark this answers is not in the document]
godwinpang: The way I am reading this thread is that there is no easy / clean way of tracking runnable completion because warmup runnables are a special case of runnables that actually run to completion. (Arguably the runnables with hasCache do this too?)

There are two options
- Leave this as is because there is no better way of doing this _with the existing runnable group setup_, and also this mirrors  Since this is changed to be an unexported interface, it's probably a relatively easy/minor thing to address as a followup?
- Try to implement runnable completion counting mechanism that @alvaroaleman suggested within this PR itself.

My vote is for the first option, just because this change is already sizeable and the suggested change in the second option seems somewhat invasive despite being simple.

pkg/internal/controller/controller.go
```diff
@@ -144,6 +167,38 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
 
+// Warmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) Warmup(ctx context.Context) error {
+	if c.NeedWarmup == nil || !*c.NeedWarmup {
+		return nil
+	}
+	return c.startEventSources(ctx)
+}
+
+// DidFinishWarmup implements the manager.WarmupRunnable interface.
```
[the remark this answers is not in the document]
godwinpang: Fixed. Tried to make it more robust do a compile time assertion with `var _ manager.WarmupRunnable = &Controller[]{}`, but we can't do that anymore with keeping the interface unexported.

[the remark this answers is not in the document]
sbueringer: I would have also expected some unit test failure somewhere if this doesn't implement the interface correctly. Are we missing test coverage?

pkg/manager/runnable_group_test.go
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/manager/manager.go
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/manager/manager.go
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/manager/internal.go
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/internal/controller/controller.go
```diff
@@ -221,13 +276,19 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 // startEventSources launches all the sources registered with this controller and waits
 // for them to sync. It returns an error if any of the sources fail to start or sync.
 func (c *Controller[request]) startEventSources(ctx context.Context) error {
+	// CAS returns false if value is already true, so early exit since another goroutine must have
+	// called startEventSources previously
+	if !c.didStartEventSources.CompareAndSwap(false, true) {
+		c.LogConstructor(nil).Info("Skipping starting event sources since it was already started")
```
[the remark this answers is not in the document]
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/internal/controller/controller.go
```diff
@@ -83,6 +90,16 @@ type Controller[request comparable] struct {
...
 
+	// didStartEventSources is used to indicate whether the event sources have been started.
+	didStartEventSources atomic.Bool
+
+	// didEventSourcesFinishSync is used to indicate whether the event sources have finished
+	// successfully. It stores a *bool where
+	// - nil: not finished syncing
+	// - true: finished syncing without error
+	// - false: finished syncing with error
+	didEventSourcesFinishSync atomic.Value
```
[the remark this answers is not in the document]
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/controller/controller_test.go
```diff
@@ -474,5 +474,75 @@ var _ = Describe("controller.Controller", func() {
 			_, ok = q.(priorityqueue.PriorityQueue[reconcile.Request])
 			Expect(ok).To(BeFalse())
 		})
+
+		It("should set ShouldWarmupWithoutLeadership correctly", func() {
```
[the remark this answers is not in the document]
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/controller/controller.go
```diff
@@ -1,5 +1,4 @@
-/*
-Copyright 2018 The Kubernetes Authors.
+/* Copyright 2018 The Kubernetes Authors.
```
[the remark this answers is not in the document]
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/config/controller.go
```diff
@@ -60,6 +60,16 @@ type Controller struct {
 	// Defaults to true, which means the controller will use leader election.
 	NeedLeaderElection *bool
 
+	// NeedWarmup specifies whether the controller should start its sources when the manager is not
+	// the leader. This is useful for cases where sources take a long time to start, as it allows
+	// for the controller to warm up its caches even before it is elected as the leader. This
+	// improves leadership failover time, as the caches will be prepopulated before the controller
+	// transitions to be leader.
+	//
+	// When set to true, the controller will start its sources without transitioning to be leader.
```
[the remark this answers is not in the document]
godwinpang: Addressed in https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/6bb4616ba3e651c98015a5382549982074f563ac

pkg/internal/controller/controller.go
```diff
@@ -144,6 +167,38 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
+func (c *Controller[request]) Warmup(ctx context.Context) error {
+	if c.NeedWarmup == nil || !*c.NeedWarmup {
+		return nil
+	}
+	return c.startEventSources(ctx)
+}
+
+// DidFinishWarmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) DidFinishWarmup(ctx context.Context) bool {
+	err := wait.PollUntilContextCancel(ctx, syncedPollPeriod, true, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
godwinpang: https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/54f4fe313a89d88a55598d196f325dc825855337

Clarified and added a UT in the above commit. `WaitForSync` returns `error` and `WaitForWarmupComplete` returns `bool`, so not exactly 1:1 mapping.

`WaitForSync` treats a context cancellation as a success (nil error), so choosing the same approach here of returning true to indicate that the runnable has finished with no error.

pkg/internal/controller/controller.go
godwinpang: Yes, this was an oversight. I've added a fix with a UT [here](http://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/667bb0318710617e96f827b5da631c9c774fab23) that will fail without this short circuit.

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
godwinpang: Yes I agree, thanks for pointing this out. I've added a test that exercises manager + controller integration explicitly in 667bb0318710617e96f827b5da631c9c774fab23. The test fails if controller doesn't implement warmupRunnable.

The previous tests only covered
- Manager can handle runnables that implement warmup interface
- Controller's warmup methods behave as expect.

Question - I'm not sure where the best place to put this test is because it implicitly tests manager functionality by assuming startup order. It seems like putting it in controller (current place) is the lesser of two evils as controllers aren't useful without managers, but managers can be useful without controllers.

pkg/manager/manager.go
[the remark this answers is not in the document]
sbueringer: I'm slightly in favor of not refactoring runnable groups in this PR if we can avoid it (always has a risk of introducing bugs elsewhere and this code is already quite complicated), but I'll defer to @alvaroaleman

pkg/config/controller.go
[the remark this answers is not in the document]
sbueringer: Thx, btw you don't have to link the commit where it's fixed (less work for you :))

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
sbueringer: controller is fine for me

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
sbueringer: > WaitForSync returns error and WaitForWarmupComplete returns bool, so not exactly 1:1 mapping.

Makes sense. Thx!

sbueringer:
I'll review again once https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2077712360 is resolved, given the potential big impact on the implementation

pkg/manager/manager.go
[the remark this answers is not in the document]
alvaroaleman: Overall I'd prefer to keep it simple with a single method but we can punt on this for this PR given the interface is unexported

pkg/internal/controller/controller.go
```diff
@@ -144,6 +166,42 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
+	}
+	ticker := time.NewTicker(syncedPollPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return true
+		case <-ticker.C:
+			didFinishSync, ok := c.didEventSourcesFinishSyncSuccessfully.Load().(*bool)
```
alvaroaleman, line 191: Can we make `didEventSourcesFinishSyncSuccessfully` a channel which we close after having done that so we don't have to poll here?

pkg/manager/runnable_group.go
```diff
@@ -65,8 +67,25 @@ func (r *runnables) Add(fn Runnable) error {
 		})
 	case webhook.Server:
 		return r.Webhooks.Add(fn, nil)
-	case LeaderElectionRunnable:
-		if !runnable.NeedLeaderElection() {
+	case warmupRunnable, LeaderElectionRunnable:
+		if warmupRunnable, ok := fn.(warmupRunnable); ok {
```
alvaroaleman, line 71: If the runnable is a leaderelection runnable and has warmup enabled but leaderelection disabled, we will still do the warmup and then _I think_ the call to warmup might race with the call to `Start` as we start them in parallel? it looks to me like calling `Start` first and then `Warmup` will start the sources twice?

While on paper the above sounds like a nonsensical config it might happen if someone globally enables warmup for all their controllers but has one controller that does not require leader election. Ideally, things just work in that case.

pkg/config/controller.go
alvaroaleman: This was marked as resolved but not addressed I think?

pkg/manager/manager.go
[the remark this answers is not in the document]
sbueringer: Agree simple would be good. Just not sure how much we have to change in runnable groups.

pkg/manager/runnable_group.go
sbueringer: What do you think about 
```go
	// Start WarmupRunnables and wait for warmup to complete.
	if err := cm.runnables.Warmup.Start(cm.internalCtx); err != nil {
		return fmt.Errorf("failed to start warmup runnables: %w", err)
	}

	// Start the non-leaderelection Runnables after the cache has synced.
	if err := cm.runnables.Others.Start(cm.internalCtx); err != nil {
		return fmt.Errorf("failed to start other runnables: %w", err)
	}
```

This should solve it, right? As Warmup.Start waits for completion

alvaroaleman: > This should solve it, right? As Warmup.Start waits for completion

Ah you are right, perfect.

This makes me aware of the fact that now that this is not exported anymore we might not need the "wait for finish" logic if we synchronize inside the controller instead, the main reason I asked for that is that if someone outside of c-r implements the interface they would likely expect warmup to finish before being started. So we can do whichever is overall simpler to implement.

godwinpang: > we might not need the "wait for finish" logic

If you are referring to the `WaitForWarmupComplete` method, we still need it because it's how we tell the runnable group to make Warmup.Start wait for completion

pkg/manager/manager_test.go
godwinpang: Rewrote test to verify startup order

pkg/manager/internal.go
godwinpang: Actually there is something even more confusing here. There is a comment that says to stop the caches *before* the leader election runnables and that it is an important step, but the call is clearly after stopping leader election runnables?...

pkg/manager/internal.go
godwinpang: Yes. In the happy path `StopAndWait` would just block until the waitgroup tracking runnables has counter == 0 which is will because the Warmup funcs have already completed.

In the unhappy path `StopAndWait` would cancel the context and since the controller's implementation of warmup respects the calling context, it early exits without waiting for warmup to finish which functionally makes sense to be despite it only `Stopping` and not really `Waiting`

pkg/config/controller.go
```diff
@@ -60,6 +60,16 @@ type Controller struct {
...
+	// Defaults to false.
+	NeedWarmup *bool
```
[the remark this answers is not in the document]
godwinpang: Will document based on the ordering suggested here: https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2089531957

pkg/manager/runnable_group.go
godwinpang: > If I see correctly using sync.Once would just be better in general?

yes. I just didn't know sync.Once existed when I wrote the code 😄

godwinpang: Thought about this a bit more. Does ordering it to do warmup before non leader election runnables break backwards compatibility? Previously a manager with non leader election runnables start right after the cache runnables have been started, but now they block on Warmup to complete. 

For controllers this does not matter as they would block on the event source starting anyways, but this breaks in the case where a user has both controllers and some other runnable implemented in user land added to the manager, since now that other runnable has to wait for the controller's source to finish before it can start.

pkg/manager/manager_test.go
```diff
@@ -1928,6 +1928,79 @@ var _ = Describe("manger.Manager", func() {
...
+				warmupCalled.Store(true)
+				<-warmupRunnableWarmupBlockingChan
+				return nil
+			},
+			didWarmupFinish: make(chan bool),
+		}
+		Expect(m.Add(warmupRunnable)).To(Succeed())
+
+		By("Creating a runnable that requires leader election")
+		leaderElectionRunnable := leaderElectionRunnableFunc{
```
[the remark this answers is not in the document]
godwinpang: Ah got it, that is a bit counter intuitive but I see it in the code [here](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.4/pkg/manager/runnable_group.go#L73-L74) now

pkg/config/controller.go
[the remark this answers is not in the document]
godwinpang: updated, but also have a question in #3192 so leaving this open

pkg/config/controller.go
godwinpang: Screenshot 2025-05-16 at 12 53 16 AM
Thanks, I meant to clean this up at the end too. Seems like Github is bugged r.e. comment resolved

pkg/manager/manager_test.go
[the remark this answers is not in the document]
godwinpang: Removed

pkg/manager/runnable_group.go
sbueringer: Good point. Going back to the original comment

> If the runnable is a leaderelection runnable and has warmup enabled but leaderelection disabled, we will still do the warmup and then I think the call to warmup might race with the call to Start as we start them in parallel? it looks to me like calling Start first and then Warmup will start the sources twice?

If we are using sync.Once we will never start the sources twice.
As we are not exporting the interface anymore Controller will be the only implementation of warmupRunnable. So it seems safe to rely on that?

pkg/manager/manager.go
godwinpang: Changed, let me know if there's anything else you want me to add

pkg/config/controller.go
godwinpang: https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/89f5479fbb8b5c43bc98489dd0d3788d5f0a1450

pkg/manager/runnable_group.go
godwinpang: Maybe another way of looking at this is that to preserve backwards compatibility we need the following behavior
- If there is an outstanding call to `startEventSources` then all later calls to the same function should block until that outstanding call returns. I *think* sync.Once already guarantees this, but going to add a UT to verify this behavior. 
- Any function that calls `startEventSources` needs to hold `c.mu` because the critical section on L268 modifies `startWatches` outside of the startEventSources call. 

Simplest solution is probably to take the mu lock in the warmup call.

sbueringer: Just realized that I don't know if sync.Once blocks additional callers

sbueringer: > Any function that calls startEventSources needs to hold c.mu because the critical section on L268 modifies startWatches outside of the startEventSources call.

Maybe worth checking if l.268 should be moved into `startEventSources` (not sure if it changes the fact that we have to hold the lock, but it might be a good improvement in any case)

godwinpang:
A summary of open comments since they are getting a bit hard to track:
- https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2089505996 and https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2084453321 discuss runnable group startup ordering.
- https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2092655893 discusses Warmup needing to be threadsafe with Start because of access to `c.startWatches`.
- https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2077712360 discusses making warmupRunnable a single method interface; dependent on https://github.com/kubernetes-sigs/controller-runtime/issues/3221

Everything else is either resolved + addressed, or kept open because they are blocked on the runnable group startup ordering discussion

pkg/manager/runnable_group.go
godwinpang: > Just realized that I don't know if sync.Once blocks additional callers until the code within sync.Once is completed

Looks like it does, so that solves 1 problem: https://go.dev/play/p/XAFFCESpUOe

godwinpang: Pushed code with tests regarding the race + clearing out `startWatches`.

pkg/internal/controller/controller.go
```diff
@@ -383,6 +441,15 @@ func (c *Controller[request]) updateMetrics(reconcileTime time.Duration) {
 	ctrlmetrics.ReconcileTime.WithLabelValues(c.Name).Observe(reconcileTime.Seconds())
 }
 
+// ensureDidWarmupFinishChanInitialized ensures that the didWarmupFinishChan is initialized. This is needed
+// because controller can directly be created from other packages like controller.Controller, and
+// there is no way for the caller to pass in the chan.
+func (c *Controller[request]) ensureDidWarmupFinishChanInitialized() {
```
alvaroaleman, line 447: Please instead write a `New` func in pkg/internal/controller and do it there so we don't need this code and the locking around it

pkg/internal/controller/controller.go
```diff
@@ -83,6 +83,22 @@ type Controller[request comparable] struct {
...
+	ensureDidWarmupFinishChanInitializedOnce sync.Once
+
+	// didWarmupFinish is closed when startEventSources returns. It is used to
+	// signal to WaitForWarmupComplete that the event sources have finished syncing.
+	didWarmupFinishChan chan struct{}
+
+	// didWarmupFinishSuccessfully is used to indicate whether the event sources have finished
+	// successfully. If true, the event sources have finished syncing without error. If false, the
+	// event sources have finished syncing but with error.
+	didWarmupFinishSuccessfully atomic.Bool
```
alvaroaleman, line 100: Please make this a `chan error` and buffer it when initializing, so we can use it both to check if it finishes and if there was an issue and don't need the bool:
```suggestion
	didWarmupFinishChan chan error
```

pkg/internal/controller/controller.go
```diff
@@ -144,6 +166,37 @@ func (c *Controller[request]) NeedLeaderElection() bool {
 	return *c.LeaderElected
 }
 
+// Warmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) Warmup(ctx context.Context) error {
```
alvaroaleman, line 170: I guess related to the error comment, having this return an `error` is very confusing because it will then get surfaced in the manager rather than in the controller when starting it - I'd prefer it if we had no return here and instead from the `sync.Once` either close `didWarmupFinishChan` on success or put the error in it.

That will reduce this method to just the sync.Once call if warmup is enabled

pkg/internal/controller/controller.go
godwinpang: Good point, didn't realize it was in `internal` so was worried about keeping backwards compatibility

pkg/manager/internal.go
godwinpang: Ok, ignoring the confusing ordering for the caches...

I think it is ok to keep this here because the warmup runnable is special and blocks before returning, hence you cannot have a leader election runnable that depends on warmup which needs to be stopped before stopping the warmup runnable. There are only three times we can stop the warmup runnable.

1. It hasn't been started yet. In this case it is a no-op and is safe.
2. It was started but _hasn't_ completed yet. In this case because it hasn't completed, the leader election runnables that depend on the warm up cannot have started yet nor can they be started later, so it is safe to stop the warmup runnables before the leader election runnables.
3. It was started and _has_ completed. In this case stopping it won't do anything, so is safe.

Do you agree or am I missing something?

godwinpang: As an aside I think there should be a test to verify startup + shutdown order since it is kind of important -- I am happy to do that in a followup PR later since this code has existed forever without a test, so I don't want to block this PR on it.

pkg/internal/controller/controller.go
godwinpang: c9a2973

godwinpang: The reason I did not make New return some sort of interface and keep Controller private is because it would create a circular dep between `pkg/internal/controller` and `pkg/controller`.
I would have had to either define a new interface that replicates the ones in `/pkg/controller/controller.go` which adds the burden of keeping them in sync, so think this warning is good enough given that it's an internal pkg?

pkg/internal/controller/controller.go
godwinpang: > it will then get surfaced in the manager rather than in the controller when starting it

QQ: Are you saying here that if warmup failed with error, the controller's Start method should early exit out with that error? Is that the right way of getting the error to be surfaced in the controller when starting it?

godwinpang: > having this return an error is very confusing

Also, to make warmup a runnable I need to have warmup return error. Are you saying that warmup should swallow the error, always return nil, then delay the error to be surfaced until the controller is started?

alvaroaleman: > QQ: Are you saying here that if warmup failed with error, the controller's Start method should early exit out with that error? Is that the right way of getting the error to be surfaced in the controller when starting it?

Yes ideally. Call the sync.Once, block on the err chan, return err or continue.

> Also, to make warmup a runnable I need to have warmup return error. Are you saying that warmup should swallow the error, always return nil, then delay the error to be surfaced until the controller is started?

Either that or just no return and wrap it in the manager so it becomes a runnable

pkg/internal/controller/controller.go
```diff
@@ -221,60 +324,71 @@ func (c *Controller[request]) Start(ctx context.Context) error {
...
-				syncingSource, ok := watch.(source.TypedSyncingSource[request])
-				if !ok {
-					return
-				}
-				didStartSyncingSource.Store(true)
-				if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
-					err := fmt.Errorf("failed to wait for %s caches to sync %v: %w", c.Name, syncingSource, err)
-					log.Error(err, "Could not wait for Cache to sync")
-					sourceStartErrChan <- err
+	var retErr error
```
alvaroaleman, line 327: I would not return the error from `startEventSources` because then the caller has to make sure they hold the lock. Instead, just write from the sync.Once into the err chan

pkg/internal/controller/controller.go
```diff
@@ -144,6 +220,39 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
+
+	err := c.startEventSources(ctx)
+	c.warmupResultChan <- err
+
+	return err
+}
+
+// WaitForWarmupComplete returns true if warmup has completed without error, and false if there was
+// an error during warmup. If context is cancelled, it returns true.
+func (c *Controller[request]) WaitForWarmupComplete(ctx context.Context) bool {
```
alvaroaleman, line 242: THere is no need for this anymore now I think?

pkg/config/controller.go
```diff
@@ -60,6 +60,19 @@ type Controller struct {
...
 
+	// EnableWarmup specifies whether the controller should start its sources when the manager is not
+	// the leader. This is useful for cases where sources take a long time to start, as it allows
+	// for the controller to warm up its caches even before it is elected as the leader. This
+	// improves leadership failover time, as the caches will be prepopulated before the controller
+	// transitions to be leader.
+	//
+	// Setting EnableWarmup to true and NeedLeaderElection to true means the controller will start its
+	// sources without waiting to become leader.
+	// Setting EnableWarmup to true and NeedLeaderElection is false is a no-op as controllers without
```
sbueringer, line 71:
```suggestion
	// Setting EnableWarmup to true and NeedLeaderElection to false is a no-op as controllers without
```

pkg/controller/controller.go
```diff
@@ -93,6 +93,16 @@ type TypedOptions[request comparable] struct {
 	//
 	// Note: This flag is disabled by default until a future version. It's currently in beta.
 	UsePriorityQueue *bool
+
+	// EnableWarmup specifies whether the controller should start its sources when the manager is not
```
sbueringer, line 97: Maybe sync this comment with the one in pkg/config/controller.go (as this field is also exported)

pkg/internal/controller/controller.go
```diff
@@ -38,7 +38,54 @@ import (
...
+	// RecoverPanic indicates whether the panic caused by reconcile should be recovered.
+	// Defaults to true.
+	RecoverPanic *bool
+
+	// LeaderElected indicates whether the controller is leader elected or always running.
+	LeaderElected *bool
+
+	// EnableWarmup specifies whether the controller should start its sources
+	// when the manager is not the leader.
+	// Defaults to false, which means that the controller will wait for leader election to start
```
sbueringer, line 80:
```suggestion
	// Defaults to false, which means that the controller will wait for leader election
```
or ~ sync with the other fields

pkg/internal/controller/controller.go
```diff
@@ -83,6 +130,13 @@ type Controller[request comparable] struct {
 	// startWatches maintains a list of sources, handlers, and predicates to start when the controller is started.
 	startWatches []source.TypedSource[request]
 
+	// didStartEventSourcesOnce is used to ensure that the event sources are only started once.
+	didStartEventSourcesOnce sync.Once
+
+	// warmupResultChan receives the result (nil / non-nil error) of the warmup method. It is
+	// consumed by the WaitForWarmupComplete method that the warmup has finished.
```
sbueringer, line 137: I think there's something wrong or missing in this sentence

pkg/internal/controller/controller.go
sbueringer: What do you mean with "this" the whole func?

pkg/internal/controller/controller.go
```diff
@@ -144,6 +220,39 @@ func (c *Controller[request]) NeedLeaderElection() bool {
...
+		return true
+	}
+
+	warmupError, ok := <-c.warmupResultChan
+	if !ok {
+		// channel closed unexpectedly
+		return false
+	}
+
+	return warmupError == nil
```
sbueringer, line 253: So what happens if we hit the error case?

Seems like we then never signal ready on startReadyCh, which means the runnableGroup.Start func only returns if the ctx is cancelled? (which if I see correctly is equivalent to it only returns if the ctx passed into controllerManager.Start returns?)

pkg/internal/controller/controller.go
sbueringer: Would this mean that `cm.runnables.Warmup.Start` returns without error even if it fails and the error would only be surfaced after the controller became leader?

alvaroaleman: > Would this mean that cm.runnables.Warmup.Start returns without error even if it fails and the error would only be surfaced after the controller became leader?

Yes

alvaroaleman: @godwinpang @sbueringer and I discussed this and came to the conclusion that it would be preferable to fail in `Warmup` already so we don't need to acquire the leader lease to error out. So I think what we want in the end is:
* Get rid of the `WaitForWarmupComplete`, we don't need it
* Both `Warmup()` and `Start()` will end up calling `startEventSources` which then calls the sync.Once and returns an error if any - Since we have a guarantee that the sync.Once only executes once, this is threadsafe
* Get rid of the `warmupResultChan` as we don't need it

pkg/manager/internal.go
sbueringer: I'll get back to this once we finalized other parts of the code :)

pkg/internal/controller/controller.go
sbueringer: Resolving this one as the conversation moved on :)

pkg/internal/controller/controller.go
sbueringer: This one should be outdated now, given https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2098142231

pkg/internal/controller/controller.go
sbueringer: Discussed. The idea is to get rid of the entire func. Let's resolve to dedup with: https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2098142231

pkg/internal/controller/controller.go
godwinpang: removed that field altogether :)

pkg/internal/controller/controller.go
godwinpang: 79a7b95

There are a few discussions so it is confusing, but I am pretty sure that reading between the lines we want to keep the warmup runnable returning error so I left it that way in the commit

sbueringer: > but I am pretty sure that reading between the lines we want to keep the warmup runnable returning error so I left it that way in the commit

Yes! At a first glance the commit looks good. I'll do a full review tomorrow

pkg/manager/internal.go
sbueringer: I think we have to call this at the beginning of this go routine (l.542) and we have to start an additional go routine to do it.

Basically during Start we let Others / LeaderElection / Warmup runnables race against each other. 

Worst case. We are still in startEventSources and stuck and we want to cancel the contxt of the runnable group (which is done in StopAndWait).

The problem now is we don't know which of the runnables is stuck in startEventSources and we can't assume any ordering. If we are canceling the "wrong" one here we are stuck. Let me make an example based on the current StopAndWait sequence here.

* A Warmup runnable is stuck in startEventSources and will only return when we cancel the context of the Warmup group
* When we call Others.StopAndWait it will wait until all runnables are completed (via r.wg.Wait())
* The Other runnable corresponding to the stuck Warmup runnable won't complete because the Start func is blocked on the mutex that it shares with the Warmup func (as Warmup is still in progress)

I think there are a few variations of this deadlock possible. So as mentioned above I think the only way to make this safe is to do this in an additional go routine and trigger that at the beginning of this goroutine

@alvaroaleman Does this also make sense to you?

sbueringer (commented):
Only had enough time to review the prod code. Wil review tests on the next round. I hope I found all the issues :)

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1103,207 @@ var _ = Describe("controller", func() {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			By("Creating a channel to track execution order")
+			runnableExecutionOrderChan := make(chan string, 2)
+			const nonWarmupRunnableName = "nonWarmupRunnable"
+			const warmupRunnableName = "warmupRunnable"
+
+			ctrl.CacheSyncTimeout = time.Second
+			ctrl.startWatches = []source.TypedSource[reconcile.Request]{
```
sbueringer, line 1188: What makes this a warmupRunnable?

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1103,207 @@ var _ = Describe("controller", func() {
...
+
+			By("Starting the manager")
+			go func() {
+				defer GinkgoRecover()
+				Expect(m.Start(ctx)).To(Succeed())
+			}()
+
+			<-m.Elected()
+			Expect(<-runnableExecutionOrderChan).To(Equal(warmupRunnableName))
+			Expect(<-runnableExecutionOrderChan).To(Equal(nonWarmupRunnableName))
```
sbueringer, line 1236: What part of the prod code guarantees this behavior?

1. We set LeaderElection on the manager to false
2. Warmup & Start funcs currently should race against each other afaik?

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1103,207 @@ var _ = Describe("controller", func() {
...
+		})
+
+		It("should not race with Start and only start sources once", func() {
```
sbueringer, line 1239: is the first part true? I mean Start and Warmup are racing (but "only start sources once" is true)

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1103,207 @@ var _ = Describe("controller", func() {
...
+			}
+
+			By("Calling Warmup when EnableWarmup is false")
+			err := ctrl.Warmup(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(isSourceStarted.Load()).To(BeFalse())
+
+			By("Calling Start when EnableWarmup is false")
+			// Now call Start
+			go func() {
```
sbueringer, line 1300: Do we need the go routine?

sbueringer (commented):
Also reviewed tests now

pkg/internal/controller/controller.go
```diff
@@ -221,60 +299,71 @@ func (c *Controller[request]) Start(ctx context.Context) error {
 // startEventSources launches all the sources registered with this controller and waits
 // for them to sync. It returns an error if any of the sources fail to start or sync.
 func (c *Controller[request]) startEventSources(ctx context.Context) error {
```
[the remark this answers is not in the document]
alvaroaleman: Yes, makes sense. Also, lets add a test for this

pkg/internal/controller/controller_test.go
alvaroaleman: I think so, because `Start` is blocking, one of that or the check needs to be in a goroutine

pkg/manager/internal.go
alvaroaleman: To add some more context to the above as it took me a moment to think through this, the reason we need the goroutine is that either `Warmup` or `Start` might be holding the mutex and be blocked on the other, hence cancelling one can't wait for the other to be cancelled, it has to happen in parallel.

pkg/internal/controller/controller_test.go
sbueringer: Ah yes, you're right

pkg/internal/controller/controller_test.go
godwinpang: The JustBeforeEach block on L1108 sets `warmup == true`

sbueringer: Maybe let's make this explicit

pkg/internal/controller/controller_test.go
godwinpang: Setting LeaderElection on the manager == false was a mistake, will fix.

pkg/internal/controller/controller_test.go
godwinpang: By `race`, I meant `should not called a race condition when called concurrently with Start`. Let me reword to make it clear

pkg/internal/controller/controller.go
```diff
@@ -144,6 +219,15 @@ func (c *Controller[request]) NeedLeaderElection() bool {
 	return *c.LeaderElected
 }
 
+// Warmup implements the manager.WarmupRunnable interface.
+func (c *Controller[request]) Warmup(ctx context.Context) error {
+	if c.EnableWarmup == nil || !*c.EnableWarmup {
+		return nil
+	}
+
+	return c.startEventSources(ctx)
```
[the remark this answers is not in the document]
godwinpang: [5df573f](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/5df573fde37b68239e45503a12e9afeac2ec7ce4)

pkg/manager/internal.go
godwinpang: [ba51d28](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/ba51d280cee0066899346eeefbbc1436b7372254)

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
godwinpang: https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/ba51d280cee0066899346eeefbbc1436b7372254 is my stab at it but there is another race that this uncovers. This is also why CI is failing.

[L336](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-3f86588ef108e53ef346d3223ca106ccaa8efe72cb1055489665d7675f3da7b4R336-R339) starts a goroutine that can potentially outlive the duration of the caller holding the lock. The errGroup blocks on `case <-sourceStartCtx.Done()`, meaning that when the context is cancelled, startEventSources() doesn't wait for the the goroutine on [L336](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-3f86588ef108e53ef346d3223ca106ccaa8efe72cb1055489665d7675f3da7b4R336-R339) to complete before returning. This means that the mutex acquired in Warmup can be released before [L339](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-3f86588ef108e53ef346d3223ca106ccaa8efe72cb1055489665d7675f3da7b4R336-R339) is executed.

The problem variable is `c.Queue` since it can be read on [L339](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-3f86588ef108e53ef346d3223ca106ccaa8efe72cb1055489665d7675f3da7b4R336-R339), but also assigned to in Start on [L261](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-3f86588ef108e53ef346d3223ca106ccaa8efe72cb1055489665d7675f3da7b4R261). 

It seems *reasonable* to ensure that the `watch.Start` call executes before the errgroup returns in the context cancelled case, as the `watch.Start` ctx call should early exit when called with a cancelled / expired context.

I don't have a solution just yet since it is late, but rough ideas are to make didStartSyncingSource` a channel instead and block on it in the ctx cancelled case before proceeding with error handling. The one problem with that is that the error message on L365 would be dropped and be backwards incompatible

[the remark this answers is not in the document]
godwinpang: Maybe we can do something as follows instead? Keep the current error handling behavior, but add a channel blocking until watch.Start is called in the defer block
```
sourceStartErrChan := make(chan error, 1) // Buffer chan to not leak goroutine if we time out
didStartSyncingSourceChan := make(chan struct{}) // <-- ⏰ this is the change
go func() {
	defer close(sourceStartErrChan)
	log.Info("Starting EventSource")
	if err := watch.Start(ctx, c.Queue); err != nil {
		sourceStartErrChan <- err
		return
	}
        close(didStartSyncingSourceChan) // <-- ⏰ this is the change
	syncingSource, ok := watch.(source.TypedSyncingSource[request])
	if !ok {
		return
	}
	didStartSyncingSource.Store(true)
	if err := syncingSource.WaitForSync(sourceStartCtx); err != nil {
		err := fmt.Errorf("failed to wait for %s caches to sync %v: %w", c.Name, syncingSource, err)
		log.Error(err, "Could not wait for Cache to sync")
		sourceStartErrChan <- err
	}
}()

select {
defer func() {<-didStartSyncingSourceChan}() // <-- ⏰ this is the change
case err := <-sourceStartErrChan:
	return err
case <-sourceStartCtx.Done():
	if didStartSyncingSource.Load() { // We are racing with WaitForSync, wait for it to let it tell us what happened
		return <-sourceStartErrChan
	}
	if ctx.Err() != nil { // Don't return an error if the root context got cancelled
		return nil
	}
	return fmt.Errorf("timed out waiting for source %s to Start. Please ensure that its Start() method is non-blocking", watch)
}
```

IDT we can overload `sourceStartErrChan` because we would have to do something like produce a nil value to it right after `watch.Start(ctx, c.Queue)` which breaks the error capturing case in WaitForSync.
It wouldn't work to overload `didStartSyncingSource` either since we'd have to poll it until it turns true.

sbueringer:
@godwinpang Pleae keep the conversations open. It's really hard to find the just resolved/fixed conversations to check that everything was addressed

pkg/internal/controller/controller.go
```diff
@@ -218,63 +308,83 @@ func (c *Controller[request]) Start(ctx context.Context) error {
...
-		})
-	}
-	return errGroup.Wait()
+		// All the watches have been started, we can reset the local slice.
+		//
+		// We should never hold watches more than necessary, each watch source can hold a backing cache,
+		// which won't be garbage collected if we hold a reference to it.
+		c.startWatches = nil
+
+		// Mark event sources as started after resetting the startWatches slice to no-op a Watch()
```
sbueringer, line 382: Watch won't be a no-op. It will start the source immediately instead of storing it in startWatches

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
sbueringer: I think the problem is bigger. Does Warmup start the sources with a nil queue? (if it is called before Start)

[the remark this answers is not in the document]
sbueringer: I think we would have to write c.Queue already in New. This also starts the queue though.

So this would also change the observable behavior of functions like NewUnmanaged / NewTypedUnmanaged.

So I think we can't move it into New. 

The easiest might be to move it into startEventSourcesLocked (and rename that func to startEventSourcesAndQueueLocked)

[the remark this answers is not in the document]
alvaroaleman: Can we make sure that our tests validate that the queue we pass to the sources is non-nil? This should've been caught by tests and not by a human ideally

sbueringer:
@godwinpang Heyho, discussed this with @alvaroaleman.

I think we're making really good progress on this PR and getting very close to merging it. We would just like to take a bit more time to ensure we don't break anything (I"ll also try to test it with Cluster API).

So we'll go ahead and cut the CR v0.21 release now already (a few folks are waiting for that :)). Once this PR merges we'll cherry-pick and cut a v0.21.1 patch release.

(I also modified the PR title as this PR is not breaking)

pkg/internal/controller/controller.go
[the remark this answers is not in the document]
godwinpang: I added a UT to verify this in [12e938c](https://github.com/kubernetes-sigs/controller-runtime/pull/3192/commits/12e938c905595368c7ebfca5462fce673be73fb6) and also changed the existing integration tests at both the manager and controller levels to exercise controllers with warmup enabled so there is some e2e coverage.

Let me also go through the existing tests around `Start`  in [controller_test.go](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller_test.go) and replicate them with warmup to make sure that we have coverage for existing tested behavior too.

godwinpang:
We've been running a forked version of c-r with `EnableWarmup` set to true for the past 2 days or so, and everything is working fine. The `workqueue_depth` metric shows that the non leader-elected pod's workqueues are being populated as expected.

I'll take a stab at
1) Adding a few more tests around Warmup behavior to make sure it has parity with Start esp around context timeout/cancellation.
2) Updating the PR description + design with what the latest implementation is.

Anything else we are looking to do to gain more confidence around this feature and get it merged? 😄

sbueringer:
Sounds fine to me

alvaroaleman:
@godwinpang any plans on wrapping this up?

godwinpang:
> @godwinpang any plans on wrapping this up?

Sorry, I missed this Github notification. I've updated the PR description and also duplicated the tests between `Start` and `Warmup` as appropriate.

alvaroaleman (approved):
thanks!
/hold

in case Stefan has anything to add

pkg/internal/controller/controller.go
```diff
@@ -218,63 +297,94 @@ func (c *Controller[request]) Start(ctx context.Context) error {
...
+						log.Error(err, "Could not wait for Cache to sync")
+						sourceStartErrChan <- err
+					}
+				}()
+
+				select {
+				case err := <-sourceStartErrChan:
+					return err
+				case <-sourceStartCtx.Done():
+					defer func() { <-hasAccessedQueueChan }() // Ensure that watch.Start has been called to avoid prematurely releasing lock before accessing c.Queue
```
sbueringer: Do we still need this (i.e. hasAccessedQueueChan) since we are starting the queue inside the sync.Once? (xref previous discussion: https://github.com/kubernetes-sigs/controller-runtime/pull/3192#discussion_r2102283405)

We write c.Queue only once and directly above, I'm not sure if there is still a race condition around this field

If we actually need it, would it be a simpler way to achieve the same by passing in c.Queue as a parameter into the func in l.357?

nit for the godoc: this does not ensure that watch.Start was called only that `q  := c.Queue` is done

pkg/manager/manager_test.go
```diff
@@ -1928,6 +1928,64 @@ var _ = Describe("manger.Manager", func() {
...
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		go func() {
+			defer GinkgoRecover()
+			Expect(m.Start(ctx)).To(Succeed())
+		}()
+
+		<-m.Elected()
+		Expect(<-runnableExecutionOrderChan).To(Equal(warmupRunnableName))
```
sbueringer, line 1995: Sorry we might have discussed this before, how does this proof that warmupRunnable was run before leader election was won?
(This checks the order but not before/after leader election was won, correct?)

pkg/controller/controller_integration_test.go
```diff
@@ -48,30 +53,46 @@ var _ = Describe("controller", func() {
 	Describe("controller", func() {
 		// TODO(directxman12): write a whole suite of controller-client interaction tests
 
-		It("should reconcile", func() {
+		// Since all tests are run in parallel and share the same testenv, we namespace the objects
```
sbueringer: Are tests actually run in parallel? I thought they are run sequentially only tests of different packages are run in parallel (but each package has its own testenv)

pkg/internal/controller/controller_test.go
```diff
@@ -494,10 +511,103 @@ var _ = Describe("controller", func() {
...
+
+			src := source.Func(func(ctx context.Context, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
+				return nil
+			})
+
+			ctrl.startWatches = []source.TypedSource[reconcile.Request]{src}
+
+			err := ctrl.startEventSourcesAndQueueLocked(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ctrl.startWatches).To(BeNil(), "startWatches should be reset to nil after returning")
```
sbueringer, line 611: While we're here, maybe let's also check startedEventSourcesAndQueue is true

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1124,516 @@ var _ = Describe("controller", func() {
...
+			By("Starting the manager")
+			waitChan := make(chan struct{})
+			go func() {
+				defer GinkgoRecover()
+				defer close(waitChan)
+				Expect(m.Start(ctx)).To(Succeed())
+			}()
+
+			<-m.Elected()
+			Expect(<-runnableExecutionOrderChan).To(Equal(warmupRunnableName))
```
sbueringer: Same question here, how does this verify warmupRunnable was run before leader election is done?

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1124,516 @@ var _ = Describe("controller", func() {
...
+			var watchStartedCount atomic.Int32
+			ctrl.startWatches = []source.TypedSource[reconcile.Request]{
+				source.Func(func(ctx context.Context, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
+					watchStartedCount.Add(1)
+					return nil
+				}),
+			}
+
+			By("calling Warmup and Start concurrently")
+			go func() {
```
sbueringer, line 1523: Should we also add a channel for start and block on it at the end of the test to ensure it was actually called?

pkg/internal/controller/controller_test.go
```diff
@@ -1014,6 +1124,516 @@ var _ = Describe("controller", func() {
...
+		})
+
+		It("should not cause a data race when called concurrently with Start and only start sources once", func() {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			ctrl.CacheSyncTimeout = time.Second
+
+			var watchStartedCount atomic.Int32
+			ctrl.startWatches = []source.TypedSource[reconcile.Request]{
```
sbueringer: WDYT about adding more watches for some easy additional coverage? (e.g. add 10 of these in a for loop)

sbueringer:
@godwinpang Thank you very much! We're almost there (only one idea for simplification in the prod code, otherwise just a few comments on tests)

pkg/internal/controller/controller.go
godwinpang: Yes I agree we don't need this synchronization anymore with the move into sync.Once, looks like this was leftover as a change before the suggestions were taken into account from the linked discussion. I'll take out the code that protects against races on `c.Queue` access

pkg/controller/controller_integration_test.go
godwinpang: They're not. I was confused because when I added the tests pods from the other namespace started showing up in the reconciled channel. I checked again and looks like the main reason is that the pod isn't deleted in the integration tests unlike the deployment / replicasets.

I tried adding a delete call to the test, but tests would start consistently failing with the `timeout waiting for process kube-apiserver to stop` error which is described in https://github.com/kubernetes-sigs/controller-runtime/issues/1571. I didn't want to add a hard coded time.Sleep described in the issue, so I would propose keeping this namespace predicate as a workaround to isolate tests. (I'll remove the running tests in parallel comment)

pkg/manager/manager_test.go
godwinpang: Right, it doesn't. It seems like there isn't a way to manually override / control leader election behavior in tests, so going to add a leader election fake and add it to verify runnable behavior.

pkg/controller/controller_integration_test.go
sbueringer: Sounds good. 

> I'll remove the running tests in parallel comment

Please add the new explanation as a replacement :)

sbueringer:
@godwinpang would be nice if we could get this PR done 😃

godwinpang:
> @godwinpang would be nice if we could get this PR done 😃

Sorry for the long wait, I've addressed the PR comments

pkg/internal/controller/controller.go
```diff
@@ -218,63 +297,91 @@ func (c *Controller[request]) Start(ctx context.Context) error {
...
+				sourceStartCtx, cancel := context.WithTimeout(ctx, c.CacheSyncTimeout)
+				defer cancel()
+
+				sourceStartErrChan := make(chan error, 1) // Buffer chan to not leak goroutine if we time out
+				go func() {
+					defer close(sourceStartErrChan)
+					log.Info("Starting EventSource")
+
+					q := c.Queue
+					if err := watch.Start(ctx, q); err != nil {
```
sbueringer:
```suggestion
					if err := watch.Start(ctx, c.Queue); err != nil {
```

pkg/manager/manager_test.go
```diff
@@ -1928,6 +1928,79 @@ var _ = Describe("manger.Manager", func() {
...
+			Expect(m.Start(ctx)).To(Succeed())
+		}()
+
+		By("Waiting for the warmup runnable to be executed without leader election being won")
+		Expect(<-runnableExecutionOrderChan).To(Equal(warmupRunnableName))
+
+		By("Unblocking leader election")
+		resourceLockWithHooks.UnblockLeaderElection()
+
+		By("Waiting for the leader election runnable to be executed without leader election being won")
```
sbueringer:
```suggestion
		By("Waiting for the leader election runnable to be executed after leader election was won")
```

pkg/leaderelection/fake/leader_election.go
```diff
@@ -30,6 +32,20 @@ import (
...
 )
 
+// ControllableResourceLockInterface is an interface that extends resourcelock.Interface to be
+// controllable.
+type ControllableResourceLockInterface interface {
+	resourcelock.Interface
+
+	// BlockLeaderElection blocks the leader election process when called. It will not be unblocked
+	// until UnblockLeaderElection is called.
+	// Not thread safe.
```
sbueringer: nit: I'm not sure I follow what makes this not thread safe (same in l.46)

Or maybe the comment is outdated?

sbueringer (commented):
Thx, just some very minor findings now.

@alvaroaleman Can you please also take another look?

pkg/leaderelection/fake/leader_election.go
godwinpang: outdated, sorry!

sbueringer:
Thank you!

/lgtm

/assign @alvaroaleman Please take a final look, thx!

Once this PR merges, let's please update the proposal PR (#3121) and let's get it merged soon as well

PR #3193, opened 2025-04-09
📖 Document pitfalls of CreateOrUpdate and CreateOrPatch

### LINKED ISSUE #3191: CreateOrUpdate is not idempotent
The [controllerutil.CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/76067279145e76148cf27c01ea4ec63fed8e463b/pkg/controller/controllerutil/controllerutil.go#L307) is not idempotent for updates.

Given the same resource multiple times, `CreateOrUpdate` will always perform an UPDATE api call, and report that it did so, even if there were no updates to the resource. This can be easily verified by making the following [modifcation](https://github.com/ian-howell/controller-runtime/commit/3782f7642c5734100221c1c3172d01c8997f641a) to the [unit test](https://github.com/kubernetes-sigs/controller-runtime/blob/76067279145e76148cf27c01ea4ec63fed8e463b/pkg/controller/controllerutil/controllerutil_test.go#L532) 

Image

This diff results in the following error:

Image

This is problematic:
1. It result in unnecessary API calls, 
2. It reduces visibility into what a controller is doing. The function suggests that it will return OperationNone if no changes were made, however controller authors still need to manually check if their resource changed. This reduces the ability to properly log and emit metrics, as well as other problems such as those offered in https://github.com/kubernetes-sigs/controller-runtime/issues/847#issuecomment-599112796

ian-howell:
This updates the documentation for the `CreateOrUpdate` and `CreateOrPatch` methods to include a section on the pitfalls of using these methods.

Closes #3191

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -278,7 +278,7 @@ func referSameObject(a, b metav1.OwnerReference) bool {
 	return aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name
 }
 
-// OperationResult is the action result of a CreateOrUpdate call.
+// OperationResult is the action result of a CreateOrUpdate or CreateOrPatch call.
 type OperationResult string
 
 const ( // They should complete the sentence "Deployment default/foo has been ..."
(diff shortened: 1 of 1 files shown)


troy0820:
/ok-to-test

I want to make sure the godoc renders like we want it to and validate the additions to the documentation as well.  

/assign @alvaroaleman 
/hold

ian-howell:
I had some free time this morning, so I went ahead and wrote up some unit tests which validate the claims in this PR.

Here's the branch: https://github.com/ian-howell/controller-runtime/tree/doc-update-verification. It includes a change to the `test-all.sh` script that causes `make test` to only run the relevant tests.

Here's the commit for convenience: https://github.com/ian-howell/controller-runtime/commit/1ac6810d8ce95ee57e388c5e8bddee6bb0f6eddb

sbueringer:
> I want to make sure the godoc renders like we want it to and validate the additions to the documentation as well.

Looks all good to me

/hold cancel
/approve

/assign @troy0820

troy0820:
This looks good to me. 

/lgtm

PR #3195, opened 2025-04-11
⚠ fix: Stop accumulating lists in multi-namespace cache implementation

### LINKED ISSUE #3194: Bug: multi_namespace_cache List() method accumulates results across subsequent calls
## Problem

When using the controller-runtime cache with DefaultNamespaces configured (via `ctrl.Options.NewCache`), the `List()` method causes results to accumulate. Each subsequent call to `List()` includes the results from previous calls, causing the result set to continuously grow.

## Steps to Reproduce

1. Create a cache with DefaultNamespaces configuration:

```go
mgr, err := ctrl.NewManager(cfg, ctrl.Options{
    NewCache: func(config *rest.Config, opts cache.Options) (cache.Cache, error) {
        opts.DefaultNamespaces = map[string]cache.Config{
            "default": {},
        }
        return cache.New(config, opts)
    },
})

```

2. Repeatedly call the List() method in a loop:

```go
pods := &corev1.PodList{}
ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
defer cancel()
for {
    select {
    case <-ctx.Done():
        t.Fatal("timeout")
    default:
        err := c.List(ctx, pods)
        // Examine pods.Items length - it grows with each call
    }
}
```

Each call to List() should return the current state of resources, not accumulate previous results. The result set should be reset and then populated with the current state of resources.

- version: `v0.20.4`

zxh326:
This PR fixes an issue in the multi-namespace cache implementation where repeated calls to the `List()` method on the same list object would accumulate results instead of resetting the list each time. This happens because the current implementation extracts existing items from the provided list object, then appends new items to it, causing the list to grow with each call.

The fix initializes a fresh empty slice for collected items instead of extracting from the existing list object.

fix #3194

### DIFF
--- a/pkg/cache/multi_namespace_cache.go
+++ b/pkg/cache/multi_namespace_cache.go
@@ -279,10 +279,7 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList, 
 		return err
 	}
 
-	allItems, err := apimeta.ExtractList(list)
-	if err != nil {
-		return err
-	}
+	allItems := []runtime.Object{}
 
 	limitSet := listOpts.Limit > 0
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -1035,13 +1035,15 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
 							Version: "v1",
 							Kind:    "PodList",
 						})
-						Expect(namespacedCache.List(context.Background(), out)).To(Succeed())
+						for range 2 {
+							Expect(namespacedCache.List(context.Background(), out)).To(Succeed())
 
-						By("verifying the returned pod is from the watched namespace")
-						Expect(out.Items).NotTo(BeEmpty())
-						Expect(out.Items).Should(HaveLen(2))
-						for _, item := range out.Items {
-							Expect(item.GetNamespace()).To(Equal(testNamespaceOne))
+							By("verifying the returned pod is from the watched namespace")
+							Expect(out.Items).NotTo(BeEmpty())
+							Expect(out.Items).Should(HaveLen(2))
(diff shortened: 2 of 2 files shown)


pkg/cache/multi_namespace_cache.go
```diff
@@ -279,10 +279,7 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 		return err
 	}
 
-	allItems, err := apimeta.ExtractList(list)
-	if err != nil {
-		return err
-	}
+	allItems := []runtime.Object{}
```
alvaroaleman, line 282: While I agree this change makes sense, its a breaking one - Can we mark it as such?

/ok-to-test

zxh326: Do you have any suggestions?

sbueringer: Sorry for the delay. I changed the PR title that's enough

sbueringer:
@zxh326 If you have time, can you please address this finding: https://github.com/kubernetes-sigs/controller-runtime/pull/3195#discussion_r2084690320

Would be great to get this into the next CR minor :)

pkg/cache/multi_namespace_cache.go
```diff
@@ -279,10 +279,7 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
 		return err
 	}
```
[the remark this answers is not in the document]
zxh326: done, I changed the code back to the old version for testing, it works.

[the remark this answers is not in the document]
sbueringer: Thx!

sbueringer:
Thank you!

/lgtm
/approve

PR #3196, opened 2025-04-11
:warning: Fakeclient: Add apply support (take 2)

### LINKED ISSUE #2341: Support server-side apply (client.Apply) in fake Client

tomasaschan:
This is a polish of #2981, which fixes #2341.

All I've done is to rebase on latest `main` to resolve merge conflicts, and address the review feedback from @jpbetz on that PR.

As I have a need for this functionality, and the previous PR was already approved (but not LGTM-ed), I hope that this can be enough to land this functionality in `main`.

/assign jpbetz

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require sigs.k8s.io/structured-merge-diff/v4 v4.6.0
+
 require (
 	cel.dev/expr v0.19.1 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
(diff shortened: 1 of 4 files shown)


sbueringer:
I assume https://github.com/kubernetes-sigs/controller-runtime/pull/2981#issuecomment-2798064699 still applies?

PR #3209, opened 2025-05-08
🐛 fix verbose info lost when clone

caozhuozi:
The `Verbose` info will be lost when using methods like `With()`.


sbueringer:
Thank you!

/lgtm
/approve

Issue #3211, opened 2025-05-09
Reconciler error due to context cancellation on shutdown


dlmather:
Revisiting https://github.com/kubernetes-sigs/controller-runtime/issues/2760, we encountered context cancellation reconciler errors with https://github.com/kubernetes-sigs/aws-load-balancer-controller during shutdown.
```
operation error Elastic Load Balancing v2: DescribeTargetHealth, context canceled

operation error Elastic Load Balancing v2: DescribeTargetHealth, context canceled

operation error Elastic Load Balancing v2: DescribeTargetHealth, https response error StatusCode: 0, RequestID: , canceled, context canceled
```

Reviewing the code https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/internal/controller/controller.go#L202

it looks like there's a potential window where if we're mid-Reconcile, we will fail out when context is cancelled, and this check https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/internal/controller/controller.go#L284 just can't quite fully close the gap on that. Ultimately, this isn't the end of the world, after shutdown another controller will be leader elected, and will sort things out, so the errors are mostly a nuisance, but they do also slow down reconciliation unnecessarily, so it would be nice to take a shot at closing down more gracefully. I have a PR that I'm tinkering with to that effect, but just wanted to open the issue to anchor that with.

alvaroaleman:
> it looks like there's a potential window where if we're mid-Reconcile, we will fail out when context is cancelled

The context being cancelled when the binary is shutdown is deliberate. Why is that a problem for you?

dlmather:
> The context being cancelled when the binary is shutdown is deliberate. Why is that a problem for you?
As linked, and described in the issue, that cancelation can occur while an ongoing Reconcile is active which cases to fail out, those failures are captured and reported as non-zero reconcile errors via https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/internal/controller/controller.go#L342

So turning the question back to you, if the cancel is deliberate is it desirable and acceptable for it to be non-graceful and considered a possible error case?

dlmather:
[...]

PR #3212, opened 2025-05-09
attempt to fix shutdown reconcile error race

dlmather:
Attempts to fix the issues outlined in https://github.com/kubernetes-sigs/controller-runtime/issues/3211 by not passing on the cancelation in the parent context to the underlying process step, and only respecting it *after* the processing runs. This does have the unfortunate effect that if processNextWorkItem blocks indefinitely, we no longer have any deadline related cancelations that would unblock it, open to ideas on that front... or maybe it's acceptable for it to block and just rely on terminationGracePeriod to come along and sweep it away in those cases? Feedback and ideas very welcome, thanks.


sbueringer:
/hold

For discussion on the day issue

PR #3213, opened 2025-05-09
:sparkles Allow a user to provide a custom metric provider for firing metrics

jonathan-innis:


alvaroaleman (commented):
I think rather than introducing the complexity of making this configurable its preferred to instead add the controller-runtime metrics registry to your own, he is an example of doing it with otel: https://github.com/kubernetes-sigs/controller-runtime/issues/305#issuecomment-2767851801

/hold

jonathan-innis:
Oh interesting! I didn't realize this was possible. Completely agree that what I have is pretty complex and annoying so I'm on board with only supporting it if it's 100% necessary. I'll give it a try and see if it fulfills what we need!

jonathan-innis:
> I think rather than introducing the complexity of making this configurable its preferred to instead add the controller-runtime metrics registry to your own

Off the bat, I will say that this is a little different than what's implemented here. From what I can tell, any implementation of this is going to be on a pull and is going to lose information compared to using EMF directly since we will be using buckets to wire the values through.

We were previously just piping things like Observations through directly so we would just directly print out the float value that we would see -- looks like now we are going to lose some information with the buckets.

jonathan-innis:
@alvaroaleman Do you have any other thoughts here? I'd like to keep the accuracy of the raw metric values if we can -- is there a WG meeting where we can discuss this a bit more?

alvaroaleman:
> We were previously just piping things like Observations through directly so we would just directly print out the float value that we would see -- looks like now we are going to lose some information with the buckets.

This is the first time someone mentions this is a problem for them. IMHO, this does not warrant the added complexity over using the prometheus -> Otel bridge.

Issue #3219, opened 2025-05-15
MultiValidatingHandler omits warnings


dlipovetsky:
Validating webhooks can return warnings in an "allowed" response. The MultiValidatingHandler [does not pass along](https://github.com/kubernetes-sigs/controller-runtime/blob/927c09e54313d6476b5780b7f029f381d777c149/pkg/webhook/admission/multi.go#L73-L88) the warnings from the various handlers it calls, so warnings are returned to the API server, and therefore not shown to the client.

/kind bug

PR #3223, opened 2025-05-16
🐛Return warnings on webhook response

troy0820:
When creating an allowed response, we do not return the warnings.  This PR will allow the warnings to be returned for validating and mutating webhooks. 

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3219

/assign @sbueringer 
/assign @alvaroaleman 

cc: @dlipovetsky


dlipovetsky:
Thanks!

/lgtm

Let's consider backporting.

alvaroaleman (approved):
Thanks!

PR #3227, opened 2025-05-20
:sparkles: Add UnsafeDisableDeepCopy to GetOptions

### LINKED ISSUE #3226: Support UnsafeDisableDeepCopy for GetOption
Right now, `client.UnsafeDisableDeepCopy` is only an option for List and directly on the cache.Cache for an object. I'd like to be able to set this on `Get` as well ad-hoc when calling through the client for objects that are particularly large (like CRDs or Pods)

jonathan-innis:
Fixes #3226

This change adds the same `UnsafeDisableDeepCopy` to `GetOptions` in the same way that we add `UnsafeDisableDeepCopy` to `ListOptions`


pkg/cache/internal/cache_reader.go
```diff
@@ -81,7 +84,7 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out client.Ob
 		return fmt.Errorf("cache contained %T, which is not an Object", obj)
 	}
 
-	if c.disableDeepCopy {
+	if c.disableDeepCopy || (getOpts.UnsafeDisableDeepCopy != nil && *getOpts.UnsafeDisableDeepCopy) {
```
alvaroaleman, line 87: Please add tests for this

jonathan-innis: Done -- doesn't look like there were test in place for the ListOptions so I added those as well

pkg/client/client_test.go
```diff
@@ -2647,6 +2671,33 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 				Expect(deps.Items[1].Name).To(Equal(dep4.Name))
 			})
 
+			It("should not deep copy results when UnsafeDisableDeepCopy is set", func() {
```
alvaroaleman, line 2674: This is already tested, no need to add another: https://github.com/kubernetes-sigs/controller-runtime/blob/71f7db556ca57ce7ea6563f77d739f0d2a54233a/pkg/cache/cache_test.go#L2465

jonathan-innis: To me, it's a little odd that this is in cache_test.go when it's a List option on the client? Does this test belong there?

PR #3229, opened 2025-05-27
:warning: Fakeclient: Clear typemeta for structured 

alvaroaleman:
The fakeclient currently differs from the liveclient in that if a
structured object is created that has typemeta set, it will retain that.
In contrast to that, the liveclient always strips it.

This change makes the fakeclient strip it just like the live client.

Based on top of https://github.com/kubernetes-sigs/controller-runtime/pull/3228 which needs to merge first.

/hold


tolusha:
Hello @alvaroaleman 
Please, could you point where real client strips the TypeMeta?
Because basically I observe the difference in behavior. Only the fake client removes the TypeMeta.

alvaroaleman:
It happens in the decoder: https://github.com/kubernetes/kubernetes/issues/80609

Any code that assumes presence of type meta on objects is prone to become problematic, because if someone does `p := &corev1.Pod{}`, `p` won't have typemeta. Do not ever rely on it being present, use `apiutil.GVKForObject`

PR #3231, opened 2025-06-03
:seedling: Bump k8s.io/* deps for fakeclient SSA fixes

alvaroaleman:
To pick up https://github.com/kubernetes/kubernetes/pull/132004 and https://github.com/kubernetes/kubernetes/pull/132039 which are required for https://github.com/kubernetes-sigs/controller-runtime/pull/2981


go.mod
```diff
@@ -85,16 +85,18 @@ require (
...
+	k8s.io/component-base v0.34.0-alpha.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
 	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.7.0 // indirect
 )
+
+replace k8s.io/client-go => k8s.io/client-go v0.0.0-20250603014204-647bbbe85970
```
alvaroaleman, line 102: Trying to directly use this without the `replace` always made `go mod tidy` remove it, I am guesinng because apiextensions-apiserver or apisever have a different version requirement.

Given that this is only temporary until there is a 1.34 tag, I think this is okay

.gomodcheck.yaml
```diff
@@ -1,10 +1,11 @@
 upstreamRefs:
-  - k8s.io/api
-  - k8s.io/apiextensions-apiserver
-  - k8s.io/apimachinery
-  - k8s.io/apiserver
-  - k8s.io/client-go
-  - k8s.io/component-base
+  # Disabled until we have a 1.34 release due to https://github.com/kubernetes-sigs/controller-runtime/pull/3231
+  #  - k8s.io/api
```
sbueringer, line 3: I assume for some dependencies you need a newer version than alpha.0?

go.mod
sbueringer: There is a tag: https://github.com/kubernetes/client-go/tree/v0.34.0-alpha.0

sbueringer: Ah, I assume you mean a tag that includes your recent changes :)

sbueringer:
In general fine, once tests are green

go.mod
alvaroaleman: > Ah, I assume you mean a tag that includes your recent changes :)

Yes exactly, the second change got merged only yesterday

.gomodcheck.yaml
alvaroaleman: yep

pkg/client/client_test.go
```diff
@@ -2030,7 +2031,8 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 				Expect(actual).NotTo(BeNil())
 
 				By("validating the fetched Deployment equals the created one")
-				Expect(u).To(Equal(&actual))
+				unstructured.RemoveNestedField(actual.Object, "spec", "template", "metadata", "creationTimestamp")
```
alvaroaleman, line 2033: The test changes are required because `creationTimestamp` is now omitted when null: https://github.com/kubernetes/kubernetes/pull/130989

sbueringer: Any idea why the creationTimestamp still shows up with null in spec.template.metadata.creationTimestamp?

```json
 "spec" : {
    "replicas" : 2,
    "selector" : {
      "matchLabels" : {
        "foo" : "bar"
      }
    },
    "template" : {
      "metadata" : {
        "creationTimestamp" : null,
        "labels" : {
          "foo" : "bar"
        }
      },
...
```

pkg/client/client_test.go
```diff
@@ -2030,7 +2031,8 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+				Expect(cmp.Diff(&actual, u)).To(BeEmpty())
```
sbueringer:
```suggestion
				Expect(u).To(BeComparableTo(&actual))
```
Let's use this instead. It uses cmp.Diff internally (also in http_test.go)

pkg/client/client_test.go
alvaroaleman: I am guessing its because the apiserver used in these tests doesn't have that patch so it still outputs it and `unstructured.Unstructured` won't remove it as its schema unaware, contrary to `appsv1.Deployment`

pkg/client/client_test.go
alvaroaleman: Done, but that is a very confusing name as it sounds like it checks if they are comparable as opposed to equal.

sbueringer: Good point

pkg/client/client_test.go
```diff
@@ -2004,7 +2004,7 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 		})
 
 		Context("with unstructured objects", func() {
-			It("should fetch an existing object", func() {
+			FIt("should fetch an existing object", func() {
```
sbueringer:
```suggestion
			It("should fetch an existing object", func() {
```

pkg/client/client_test.go
sbueringer: Ah, yeah makes sense

pkg/client/client_test.go
alvaroaleman: Done

sbueringer:
Thx!

/lgtm
/approve

I'll make a note in a v0.22 issue to revert the change in .gomodcheck.yaml

Issue #3216, opened 2025-05-12
Leverage label functionality for leader election


DerekFrank:
This is an issue to mark the fact that when https://github.com/kubernetes/kubernetes/pull/131632 is merged, I will be opening a PR to add functionality to leverage the new label functions from client-go

PR #3237, opened 2025-06-17
✨ feat: add support for custom labels on leader election leases ✨

DerekFrank:
This PR makes two changes:

* Refactors leader_election.go to not use NewWithKubeConfig() and instead perform the timeout calculation itself
* Adds a new option to managers to allow users to set custom labels that will be set on the lease when it changes owner (solves #3216)

### DIFF
--- a/pkg/leaderelection/leader_election.go
+++ b/pkg/leaderelection/leader_election.go
@@ -56,6 +56,10 @@ type Options struct {
 	// Without that, a single slow response from the API server can result
 	// in losing leadership.
 	RenewDeadline time.Duration
+
+	// LeaderLabels are an optional set of labels that will be set on the lease object
+	// when this replica becomes leader
+	LeaderLabels map[string]string
 }
 
 // NewResourceLock creates a new resource lock for use in a leader election loop.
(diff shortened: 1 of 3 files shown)


pkg/leaderelection/leader_election.go
```diff
@@ -93,22 +96,21 @@ func NewResourceLock(config *rest.Config, recorderProvider recorder.Provider, op
 	}
 	id = id + "_" + string(uuid.NewUUID())
 
-	// Construct clients for leader election
-	rest.AddUserAgent(config, "leader-election")
+	// Construct config for leader election
+	config = rest.AddUserAgent(config, "leader-election")
 
+	// Timeout set for a client used to contact to Kubernetes should be lower than
```
alvaroaleman, line 102: What does this have to do with the rest of the PR?

DerekFrank: This lets controllers that have RenewDeadline set also use the label functionality, because `NewWithKubeConfig()` doesn't allow one to set labels. If we don't make this change then users can either use RenewDeadline or LeaderLabels. This change lets them use both. If you'd prefer, I can cut a PR to client-go to create a `NewWithKubeConfigAndLabels()` function so we can keep both paths, but this change seems cleaner overall

sbueringer:
Thx!

/lgtm

Issue #3239, opened 2025-06-22
Get from priority queue might get stuck when the priority queue is shut down


michaelawyu:
Hi team,

Recently I've testing with the new priority queue implementation and it occurs to me that right now getting from the priority queue when it has been shutdown might get stuck. It can be easily triggered if we:

1. create a priority queue
2. add an item to the queue
3. call the `Shutdown()` method
4. get from the priority queue
5. get again from the priority queue -> this call will be blocked

This happens as when the priority queue is shutdown, [the `spin()` method](https://github.com/kubernetes-sigs/controller-runtime/blob/b9a9ca01fd37a66225eb4793544349ee2ed1f196/pkg/controller/priorityqueue/priorityqueue.go#L198) will stop, but the `get` channel is [never closed](https://github.com/kubernetes-sigs/controller-runtime/blob/b9a9ca01fd37a66225eb4793544349ee2ed1f196/pkg/controller/priorityqueue/priorityqueue.go#L310), and [the procedure (step 5) that gets from the queue](https://github.com/kubernetes-sigs/controller-runtime/blob/b9a9ca01fd37a66225eb4793544349ee2ed1f196/pkg/controller/priorityqueue/priorityqueue.go#L276) will be stuck waiting on the `get` channel.

On the contrary, the rate limiting queue from the `workqueue` package will properly handle this situation properly by directly returning that the queue has been shut down.

The issue concerns me a bit as it will block graceful shutdown: in my project we use the `envtest` package and normally when the emulated cluster environment is stripped down, controller runtime will see that the main context has been cancelled and all reconcilers will stop upon seeing that the workqueue has been turned off; however, when priority queue is in use, the shutdown of the emulation will time out as there might be workers being stuck waiting for an item to arrive from the priority queue.

Attached also is some really simple test code that I used to trigger the behavior (with v0.21.0):

```go
import (
	"fmt"
	"testing"

	"k8s.io/client-go/util/workqueue"
	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
)

func TestBlocker(t *testing.T) {
	rateLimiter := workqueue.DefaultTypedControllerRateLimiter[any]()
	wq := workqueue.NewTypedRateLimitingQueue(rateLimiter)

	println("Writing to the workqueue\n")
	wq.Add(1)

	println("Shutting down the workqueue\n")
	wq.ShutDown()

	println("Attempting to read from the workqueue (#1)\n")
[...]

PR #3243, opened 2025-06-25
Fix: fix priority queue blocking

dongjiang1989:
Close https://github.com/kubernetes-sigs/controller-runtime/issues/3239

Fix  priority queue `GetWithPriority` blocking  when the priority queue is shut down


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -277,6 +277,13 @@ func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool)
 	w.waiters.Add(1)
 
 	w.notifyItemOrWaiterAdded()
+
+	// ref: https://github.com/kubernetes-sigs/controller-runtime/issues/3239
+	if w.shutdown.Load() == true {
```
alvaroaleman: No need to compare booleans:
```suggestion
	if w.shutdown.Load() {
```

dongjiang1989: Thanks @alvaroaleman 
Fixed
Please re-check it

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -277,6 +277,13 @@ func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool)
 	w.waiters.Add(1)
 
 	w.notifyItemOrWaiterAdded()
+
+	// ref: https://github.com/kubernetes-sigs/controller-runtime/issues/3239
```
sbueringer, line 281: @alvaroaleman Shouldn't we do this at the beginning of the func?

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -300,6 +300,26 @@ var _ = Describe("Controllerworkqueue", func() {
...
 	})
 
+	// ref: https://github.com/kubernetes-sigs/controller-runtime/issues/3239
+	It("Get from priority queue might get stuck when the priority queue is shut down", func() {
+		q, _ := newQueue()
+
+		q.Add("baz")
+		// shut down
+		q.ShutDown()
+		q.AddWithOpts(AddOpts{After: time.Second}, "foo")
```
sbueringer, line 310: Looks like Add still adds the item to the shutdown queue. Is this the expected behavior?

sbueringer: @alvaroaleman What's your take on this one?

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: https://github.com/kubernetes-sigs/controller-runtime/pull/3250

pkg/controller/priorityqueue/priorityqueue_test.go
alvaroaleman: We can't really keep anyone from calling `Add` as it has no return but we can stop acting on it: https://github.com/kubernetes-sigs/controller-runtime/pull/3250

sbueringer: Yeah that's what I was thinking

PR #3250, opened 2025-06-28
:bug: Priorityqueue: Shutdown on shutdown

alvaroaleman:
The PQ still stores new items when shutdown and triggers processing when Get is called, avoid all of that.


alvaroaleman:
cc @sbueringer

sbueringer:
Thx!

PR #3260, opened 2025-07-18
✨ envtest: search the assets index for latest of a release series

cbandy:
This lets one configure the `DownloadBinaryAssetsVersion` field of `envtest.Environment` with only the Major.Minor portions of a Kubernetes API version. When `DownloadBinaryAssets` is set, the framework will search the asset index for the latest stable version in that release series.

This matches the behavior of `setup-envtest use {major}.{minor}` without the full complexity of version selectors. The leading `v` is now optional, too.

For example:
- `"v1.29"` resolves to [v1.29.4](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L24)
- `"1.27"` resolves to [v1.27.1](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L134)
- `"1.24"` resolves to [v1.24.2](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L222)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module sigs.k8s.io/controller-runtime
 go 1.24.0
 
 require (
-	github.com/blang/semver/v4 v4.0.0
+	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/evanphx/json-patch/v5 v5.9.11
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/go-logr/logr v1.4.2
@@ -37,6 +37,7 @@ require (
 	cel.dev/expr v0.24.0 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/pkg/envtest/binaries.go
+++ b/pkg/envtest/binaries.go
@@ -124,15 +124,54 @@ func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAsse
 		}
 	}
 
+	var requestedRange semver.Range
+	if binaryAssetsVersion != "" {
+		binaryAssetsVersion = strings.TrimPrefix(binaryAssetsVersion, "v")
+		parsedVersion, errV := semver.ParseTolerant(binaryAssetsVersion)
+		parsedRange, errR := semver.ParseRange(binaryAssetsVersion)
+
+		switch {
+		// When an exact version is requested, don't look for it in the release index.
+		case errV == nil && parsedVersion.String() == binaryAssetsVersion:
+			requestedRange = nil
+
+		// When the version looks like a range, apply it to the index.
+		case errR == nil:
+			requestedRange = parsedRange
+
+		// When the version isn't exact, turn it into a range.
+		//
+		// The `setup-envtest` tool interprets a partial version to mean "latest stable with that prefix."
+		// For example, "1" and "1.2" are akin to "1.x.x" and "1.2.x" in [semver.ParseRange].
+		// [semver.ParseTolerant] fills in missing minor or patch with "0", so this replaces those with "x".
+		//
+		// That *should* produce a valid range. If it doesn't, use the original value and skip the index.
+		case errV == nil:
+			suffix := strings.TrimPrefix(parsedVersion.FinalizeVersion(), binaryAssetsVersion)
+			suffix = strings.ReplaceAll(suffix, "0", "x")
+			parsedRange, errR = semver.ParseRange(binaryAssetsVersion + suffix)
+
+			if errR == nil {
+				requestedRange = parsedRange
+			} else {
+				requestedRange = nil
+			}
+		}
+	} else {
+		// When no version is requested, look for one in the release index.
+		requestedRange = semver.MustParseRange(">0.0.0")
+	}
+
+	// When a range a versions is requested, select one from the release index.
 	var binaryAssetsIndex *index
-	if binaryAssetsVersion == "" || strings.Count(binaryAssetsVersion, ".") < 2 {
+	if requestedRange != nil {
 		var err error
 		binaryAssetsIndex, err = getIndex(ctx, binaryAssetsIndexURL)
 		if err != nil {
 			return "", "", "", err
 		}
 
-		binaryAssetsVersion, err = latestStableVersionFromIndex(binaryAssetsIndex, binaryAssetsVersion)
+		binaryAssetsVersion, err = latestStableVersionFromIndex(binaryAssetsIndex, requestedRange)
 		if err != nil {
 			return "", "", "", err
 		}
@@ -219,6 +258,8 @@ func fileExists(path string) bool {
 }
 
 func downloadBinaryAssetsArchive(ctx context.Context, index *index, version string, out io.Writer) error {
+	version = "v" + strings.TrimPrefix(version, "v")
+
 	archives, ok := index.Releases[version]
 	if !ok {
 		return fmt.Errorf("failed to find envtest binaries for version %s", version)
@@ -252,26 +293,20 @@ func downloadBinaryAssetsArchive(ctx context.Context, index *index, version stri
 	return readBody(resp, out, archiveName, archive.Hash)
 }
 
-func latestStableVersionFromIndex(index *index, prefix string) (string, error) {
+func latestStableVersionFromIndex(index *index, satisfying semver.Range) (string, error) {
 	if len(index.Releases) == 0 {
 		return "", fmt.Errorf("failed to find latest stable version from index: index is empty")
 	}
 
-	prefix = strings.TrimPrefix(prefix, "v")
 	parsedVersions := []semver.Version{}
 	for releaseVersion := range index.Releases {
-		// Filter on prefix.
-		if !strings.HasPrefix(strings.TrimPrefix(releaseVersion, "v"), prefix) {
-			continue
-		}
-
 		v, err := semver.ParseTolerant(releaseVersion)
 		if err != nil {
 			return "", fmt.Errorf("failed to parse version %q: %w", releaseVersion, err)
 		}
 
-		// Filter out pre-releases.
-		if len(v.Pre) > 0 {
+		// Filter out pre-releases and undesirable versions.
+		if len(v.Pre) > 0 || !satisfying(v) {
 			continue
 		}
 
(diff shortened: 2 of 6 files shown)


pkg/envtest/binaries.go
```diff
@@ -252,13 +252,19 @@ func downloadBinaryAssetsArchive(ctx context.Context, index *index, version stri
...
+func latestStableVersionFromIndex(index *index, prefix string) (string, error) {
 	if len(index.Releases) == 0 {
 		return "", fmt.Errorf("failed to find latest stable version from index: index is empty")
 	}
 
+	prefix = strings.TrimPrefix(prefix, "v")
 	parsedVersions := []semver.Version{}
 	for releaseVersion := range index.Releases {
+		// Filter on prefix.
+		if !strings.HasPrefix(strings.TrimPrefix(releaseVersion, "v"), prefix) {
```
cbandy: This is _too_ simple, I suppose. `1.2` resolves to v1.29.4, and in about 20 years (~70 releases) two digit numbers will match on three digit ones.

alvaroaleman: Can we use a semver lib for this to find the minor version rather than this prefix matching and dot splitting business?

sbueringer: I'm wondering if we should really use "major.minor" to indicate "latest stable major.minor" (I assume we don't want to pick up pre-releases?). I would usually assume `major.minor` stands for `major.minor.0`. But maybe this is defined somewhere, not sure :)

In Cluster API we used this format: `stable-1.29` (so `stable-major.minor`)

Also allows us to eventually add `latest-major.minor` if there is ever demand for something that picks up the latest release including pre-releases

sbueringer: If I see correctly patch is not optional for semver (https://semver.org/)

alvaroaleman: > If I see correctly patch is not optional for semver (https://semver.org/)

The semver spec may require it, but I think its possible for an actual lib to (optionally) not require it.

> I'm wondering if we should really use "major.minor" to indicate "latest stable major.minor" (I assume we don't want to pick up pre-releases?). I would usually assume major.minor stands for major.minor.0. But maybe this is defined somewhere, not sure :

My assumption would've also been that `major.minor` means `latest patch of major.minor` but good to know that different ppl have different interpretations of this.

Taking a step back, does it really make sense to ever resolve a version here dynamically for either the patch or the minor? The use-case for envtest is tests, if the k8s version it uses changes over time that means tests may break without any code change in the test or code under test, which seems bad and should probably not be natively supported here...?

sbueringer: > The use-case for envtest is tests, if the k8s version it uses changes over time that means tests may break without any code change in the test or code under test, which seems bad and should probably not be natively supported here...?

I think it really depends on what your goals are.

In Cluster API we have various tests that intentionally test Cluster API with latest patch versions of various Kubernetes release series. So that we immediately notice if a new Kubernetes patch version introduced an issue.

So I think if something like this is your goal it makes sense

(we even have some separate Jobs that are testing against latest commits of k/k main, but that has nothing to do with patch releases :))

sbueringer: > My assumption would've also been that major.minor means latest patch of major.minor but good to know that different ppl have different interpretations of this.

Hm yeah, you're probably right. E.g. if you use the golang:1.24 Docker image you also always get the latest patch version.

> The semver spec may require it, but I think its possible for an actual lib to (optionally) not require it.

Yup. I think the standard library is `github.com/blang/semver/v4`, CR already has it as a dependency.

```
// ParseTolerant allows for certain version specifications that do not strictly adhere to semver
// specs to be parsed by this library. It does so by normalizing versions before passing them to
// Parse(). It currently trims spaces, removes a "v" prefix, adds a 0 patch number to versions
// with only major and minor components specified, and removes leading 0s.
func ParseTolerant(s string) (Version, error) {
```

We have to add a bit of additional code so we can differentiate between e.g. v1.29.0 and v1.29 as 0 is the zero value of `Version.Patch`

cbandy: > Can we use a semver lib for this to find the minor version rather than this prefix matching and dot splitting business?

Yep. That's where I was going to go next.

> I would usually assume major.minor stands for major.minor.0. But maybe this is defined somewhere, not sure :)

The two-place version is how we've been using `setup-envtest`. It has an entire notion of selectors, but the relevant bit is: https://github.com/kubernetes-sigs/controller-runtime/blob/71f7db556ca57ce7ea6563f77d739f0d2a54233a/tools/setup-envtest/versions/parse.go#L26-L29

> In Cluster API we have various tests that intentionally test Cluster API with latest patch versions of various Kubernetes release series. So that we immediately notice if a new Kubernetes patch version introduced an issue.

That's how we've been using `setup-envtest` in our project. We intend to be compatible with a few minor versions of Kubernetes behind latest stable, so we have a matrix in CI like `[v1.30, v1.31, v1.32]`. We haven't needed `setup-envtest`'s selectors beyond that.

sbueringer: I'm fine if we want to go ahead with major.minor standing for latest stable patch of major.minor

cbandy: 🤔 The version package we're importing has its notion of selectors, too. [This](https://github.com/blang/semver/blob/af3461a9cbcf1f3f5889d21b83f5ef63880c33a8/v4/range.go#L321-L324) is the flavor I'm interested in:

> ```go
> // * when dealing with wildcards without
> // version operator:
> // 1.2.x       will become    >= 1.2.0 < 1.3.0
> // 1.x         will become    >= 1.0.0 < 2.0.0
> ```

cbandy: > Can we use a semver lib for this to find the minor version rather than this prefix matching and dot splitting business?

Done.

pkg/envtest/binaries.go
```diff
@@ -124,15 +124,54 @@ func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAsse
 		}
 	}
 
+	var requestedRange semver.Range
+	if binaryAssetsVersion != "" {
```
sbueringer: Do we really have to support all of this? I intentionally didn't duplicate all the messiness of setup-envtest when implementing this functionality here

To be honest, when looking at the code now I'm not sure what exactly we are supporting.

Can we limit this somehow? For example what if we only support whatever is directly parsable by semver.ParseTolerant and semvar.ParseRange?

pkg/envtest/binaries.go
```diff
@@ -124,15 +124,54 @@ func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAsse
...
+		// That *should* produce a valid range. If it doesn't, use the original value and skip the index.
+		case errV == nil:
+			suffix := strings.TrimPrefix(parsedVersion.FinalizeVersion(), binaryAssetsVersion)
+			suffix = strings.ReplaceAll(suffix, "0", "x")
+			parsedRange, errR = semver.ParseRange(binaryAssetsVersion + suffix)
+
+			if errR == nil {
+				requestedRange = parsedRange
+			} else {
+				requestedRange = nil
```
sbueringer: When are we hitting this case and what happens then?

pkg/envtest/binaries.go
sbueringer: Independent of how much we support, let's extract this logic into a separate func and table test all cases

cbandy: No problem!

cbandy:
I've rebased and made the requested changes. I switched from `blang/semver` to `Masterminds/semver` (MIT) after the tests hit edge cases in its parsing.[^1] I found the new package easier to use, and the function turned out the way you imagined:

> Can we limit this somehow? For example what if we only support whatever is directly parsable by semver.ParseTolerant and semvar.ParseRange?

I left the latest changes as a separate commit in case that helps during review. I'd like to squash this down before merge, tho.

[^1]: https://www.github.com/blang/semver/issues/80#issuecomment-2316760091

pkg/envtest/binaries.go
cbandy: This would theoretically happen when the input is a tolerated version, and/but the string manipulations produce text that is *not* a range. As I was writing tests approaching this area, I ran into other edge cases of the `blang/semver` range parsing and behavior. In the current changeset, this moment is the final `return nil, nil` of `interpretVersion`.

When this happens, there is no "search" of the index; `binaryAssetsVersion` is used directly as the key/lookup in `downloadBinaryAssetsArchive`. That function errors with:

https://github.com/kubernetes-sigs/controller-runtime/blob/682465344b9b74efad4657016668e62438000541/pkg/envtest/binaries.go#L222-L225

go.mod
```diff
@@ -3,7 +3,7 @@ module sigs.k8s.io/controller-runtime
 go 1.24.0
 
 require (
-	github.com/blang/semver/v4 v4.0.0
+	github.com/Masterminds/semver/v3 v3.4.0
```
sbueringer, line 6: The nice thing about using blang/semver was that we already had this as a dependency and we shared this dependency with k/k.

The nice thing about mostly sticking to k/k's dependencies is that they take care of all the maintenance for us (including sometimes forking abandoned dependencies if necessary).

Now we have an additional dependency that everyone that uses controller-runtime inherits

To be honest, I'm not sure if we can justify this for version parsing in envtest

sbueringer: @alvaroaleman What do you think?

alvaroaleman: I don't really mind to be honest - Its a commonly-used lib and it has zero dependencies itself: https://github.com/Masterminds/semver/blob/bf01c618459762fa583c24476ec46957a330c737/go.mod

sbueringer: Yeah. I would just try to keep the number of transitive dependencies of CR low (especially the ones we add on top of k8s.io/* transitive dependencies)

cbandy: I forgot about these packages:
- https://pkg.go.dev/k8s.io/apimachinery/pkg/version
- https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version

🤔 The fuzzy matching I'm after may just be [ParseMajorMinor](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version#ParseMajorMinor)... lol, no rush on this PR. I'll poke it again before EOW.

I've used the latter package when testing edge cases in SSA over the years:

```go
		switch {
		case serverVersion.LessThan(version.MustParseGeneric("1.25.15")):
		case serverVersion.AtLeast(version.MustParseGeneric("1.26")) && serverVersion.LessThan(version.MustParseGeneric("1.26.10")):
		case serverVersion.AtLeast(version.MustParseGeneric("1.27")) && serverVersion.LessThan(version.MustParseGeneric("1.27.7")):

			assert.Assert(t, after.GetResourceVersion() != before.GetResourceVersion(),
				"expected https://issue.k8s.io/116861")

		default:
			assert.Assert(t, after.GetResourceVersion() == before.GetResourceVersion())
		}
```

cbandy:
> I forgot about these packages:
> 
> * https://pkg.go.dev/k8s.io/apimachinery/pkg/version
> * https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version

The latter apimachinery package worked well! Rather than add a rewrite here, I'm closing this to be replaced by https://github.com/kubernetes-sigs/controller-runtime/pull/3280. Both PRs are about the same size, but the new one seems more straightforward.

PR #3261, opened 2025-07-19
 :sparkles: Priorityqueue: Optionally return items fairly

alvaroaleman:
This change makes it possible for the priorityqueue to return items fairly across one or more fairness dimensions. These fairness dimensions are extracted from items using a configurable extractor func. Fairness is only ensured within a given priority, as the purpose of priorities is to return higher-priority items before lower-priority items.

An example use-case is to use the namespace as fairness dimension, which then ensures that one very busy namespace can not starve off other namespaces.

An example use-case for more than one dimension are multi-cluster controllers, where fairness should first be achieved across clusters and then across namespaces within each cluster.

By default, no extractor func is configured so this doesn't change any behavior. We can revisit this in the future.

Naively, the whole problem looks like it could be solved by sorting the items in the queue appropriately, but that doesn't really work because:

* Most (all?) sorting algorithm assume that the order of two elements can be determined just by looking at those two elements, but for fairness we have to consider the items in front of the current elements as well as the items that we already handed out
* We can not hand out locked items and have to ignore them for the purpose of fairness. This in turn means that whenever an item gets unlocked, we'd have to redo our sorting

This change solves the problem by introducing a `fairQueue` structure which holds all items that are ready to be handed out and in the highest priority bracket that has ready items. The priorityqueue then hands out items from its internal `fairQueue` and manages its content.

The `fairQueue` works by using a slice of all the fairness dinensions it has ever observed and a map of fairness dimension to item slice. Whenever it hands out an item, it iterates over this slice and checks the corresponding map for items until it found an item to hand out.


alvaroaleman:
I am going to close this out. I am not super convinced that this is the right approach as opposed to instantiating a controller and thus workqueue per dimension, because the approach in here means a) no queue_depth metric per dimension b) if elements from one dimension dominate, they can still delay others by occupying reconciliation workers. We can re-evaluate the approach in the future should others disagree.

sbueringer:
Makes sense to me. Sorry didn't get around to take a closer look. Let me know if you want to discuss this again in the future.

PR #3262, opened 2025-07-22
⚠️ Migration to the new events API

### LINKED ISSUE #2141: Switch to use `events.EventRecorder` instead of `record.EventRecorder`
Controller runtime uses older interface, would be good to utilize new one

clebs:
This Pull Request migrates from the old events API (`k8s.io/client-go/tools/record`) to the new API (`k8s.io/client-go/tools/events`).

Closes #2141

### DIFF
--- a/pkg/internal/recorder/recorder.go
+++ b/pkg/internal/recorder/recorder.go
@@ -24,16 +24,19 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/events"
 	"k8s.io/client-go/tools/record"
 )
 
 // EventBroadcasterProducer makes an event broadcaster, returning
 // whether or not the broadcaster should be stopped with the Provider,
 // or not (e.g. if it's shared, it shouldn't be stopped with the Provider).
-type EventBroadcasterProducer func() (caster record.EventBroadcaster, stopWithProvider bool)
+// This producer currently produces both an old API and a new API broadcaster.
+type EventBroadcasterProducer func() (deprecatedCaster record.EventBroadcaster, caster events.EventBroadcaster, stopWithProvider bool)
 
 // Provider is a recorder.Provider that records events to the k8s API server
 // and to a logr Logger.
@@ -48,9 +51,13 @@ type Provider struct {
 	evtClient       corev1client.EventInterface
 	makeBroadcaster EventBroadcasterProducer
 
-	broadcasterOnce sync.Once
-	broadcaster     record.EventBroadcaster
-	stopBroadcaster bool
+	broadcasterOnce         sync.Once
+	broadcaster             events.EventBroadcaster
+	cancelSinkRecordingFunc context.CancelFunc
+	stopWatcherFunc         func()
+	// Deprecated: will be removed in a future release. Use the broadcaster above instead.
+	deprecatedBroadcaster record.EventBroadcaster
+	stopBroadcaster       bool
 }
 
 // NB(directxman12): this manually implements Stop instead of Being a runnable because we need to
@@ -71,10 +78,13 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
 		// almost certainly already been started (e.g. by leader election).  We
 		// need to invoke this to ensure that we don't inadvertently race with
 		// an invocation of getBroadcaster.
-		broadcaster := p.getBroadcaster()
+		deprecatedBroadcaster, broadcaster := p.getBroadcaster()
 		if p.stopBroadcaster {
 			p.lock.Lock()
 			broadcaster.Shutdown()
+			p.cancelSinkRecordingFunc()
+			p.stopWatcherFunc()
+			deprecatedBroadcaster.Shutdown()
 			p.stopped = true
 			p.lock.Unlock()
 		}
@@ -89,7 +99,7 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
 
 // getBroadcaster ensures that a broadcaster is started for this
 // provider, and returns it.  It's threadsafe.
-func (p *Provider) getBroadcaster() record.EventBroadcaster {
+func (p *Provider) getBroadcaster() (record.EventBroadcaster, events.EventBroadcaster) {
 	// NB(directxman12): this can technically still leak if something calls
 	// "getBroadcaster" (i.e. Emits an Event) but never calls Start, but if we
 	// create the broadcaster in start, we could race with other things that
@@ -97,17 +107,37 @@ func (p *Provider) getBroadcaster() record.EventBroadcaster {
 	// silently swallowing events and more locking, but that seems suboptimal.
 
 	p.broadcasterOnce.Do(func() {
-		broadcaster, stop := p.makeBroadcaster()
-		broadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: p.evtClient})
-		broadcaster.StartEventWatcher(
+		p.deprecatedBroadcaster, p.broadcaster, p.stopBroadcaster = p.makeBroadcaster()
+
+		// init deprecated broadcaster
+		p.deprecatedBroadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: p.evtClient})
+		p.deprecatedBroadcaster.StartEventWatcher(
 			func(e *corev1.Event) {
 				p.logger.V(1).Info(e.Message, "type", e.Type, "object", e.InvolvedObject, "reason", e.Reason)
 			})
-		p.broadcaster = broadcaster
-		p.stopBroadcaster = stop
+
+		// init new broadcaster
+		ctx, cancel := context.WithCancel(context.Background())
+		p.cancelSinkRecordingFunc = cancel
+		if err := p.broadcaster.StartRecordingToSinkWithContext(ctx); err != nil {
+			p.logger.Error(err, "error starting recording for broadcaster")
+			return
+		}
+
+		stopWatcher, err := p.broadcaster.StartEventWatcher(func(event runtime.Object) {
+			e, isEvt := event.(*eventsv1.Event)
+			if isEvt {
+				p.logger.V(1).Info(e.Note, "type", e.Type, "object", e.Related, "action", e.Action, "reason", e.Reason)
+			}
+		})
+		if err != nil {
+			p.logger.Error(err, "error starting event watcher for broadcaster")
+		}
+
+		p.stopWatcherFunc = stopWatcher
 	})
 
-	return p.broadcaster
+	return p.deprecatedBroadcaster, p.broadcaster
 }
 
 // NewProvider create a new Provider instance.
@@ -128,6 +158,15 @@ func NewProvider(config *rest.Config, httpClient *http.Client, scheme *runtime.S
 // GetEventRecorderFor returns an event recorder that broadcasts to this provider's
 // broadcaster.  All events will be associated with a component of the given name.
 func (p *Provider) GetEventRecorderFor(name string) record.EventRecorder {
+	return &deprecatedRecorder{
+		prov: p,
+		name: name,
+	}
+}
+
+// GetEventRecorder returns an event recorder that broadcasts to this provider's
+// broadcaster.  All events will be associated with a component of the given name.
+func (p *Provider) GetEventRecorder(name string) events.EventRecorder {
 	return &lazyRecorder{
 		prov: p,
 		name: name,
@@ -141,18 +180,47 @@ type lazyRecorder struct {
 	name string
 
 	recOnce sync.Once
-	rec     record.EventRecorder
+	rec     events.EventRecorder
 }
 
 // ensureRecording ensures that a concrete recorder is populated for this recorder.
 func (l *lazyRecorder) ensureRecording() {
 	l.recOnce.Do(func() {
-		broadcaster := l.prov.getBroadcaster()
-		l.rec = broadcaster.NewRecorder(l.prov.scheme, corev1.EventSource{Component: l.name})
+		_, broadcaster := l.prov.getBroadcaster()
+		l.rec = broadcaster.NewRecorder(l.prov.scheme, l.name)
 	})
 }
 
-func (l *lazyRecorder) Event(object runtime.Object, eventtype, reason, message string) {
+func (l *lazyRecorder) Eventf(regarding runtime.Object, related runtime.Object, eventtype, reason, action, note string, args ...any) {
+	l.ensureRecording()
+
+	l.prov.lock.RLock()
+	if !l.prov.stopped {
+		l.rec.Eventf(regarding, related, eventtype, reason, action, note, args...)
+	}
+	l.prov.lock.RUnlock()
+}
+
+// deprecatedRecorder implements the old events API during the tranisiton and will be removed in a future release.
+//
+// Deprecated: will be removed in a future release.
+type deprecatedRecorder struct {
+	prov *Provider
+	name string
+
+	recOnce sync.Once
+	rec     record.EventRecorder
+}
+
+// ensureRecording ensures that a concrete recorder is populated for this recorder.
+func (l *deprecatedRecorder) ensureRecording() {
+	l.recOnce.Do(func() {
+		deprecatedBroadcaster, _ := l.prov.getBroadcaster()
+		l.rec = deprecatedBroadcaster.NewRecorder(l.prov.scheme, corev1.EventSource{Component: l.name})
+	})
+}
+
+func (l *deprecatedRecorder) Event(object runtime.Object, eventtype, reason, message string) {
 	l.ensureRecording()
 
 	l.prov.lock.RLock()
@@ -161,7 +229,8 @@ func (l *lazyRecorder) Event(object runtime.Object, eventtype, reason, message s
 	}
 	l.prov.lock.RUnlock()
 }
-func (l *lazyRecorder) Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{}) {
+
+func (l *deprecatedRecorder) Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...any) {
 	l.ensureRecording()
 
 	l.prov.lock.RLock()
@@ -170,7 +239,8 @@ func (l *lazyRecorder) Eventf(object runtime.Object, eventtype, reason, messageF
 	}
 	l.prov.lock.RUnlock()
 }
-func (l *lazyRecorder) AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{}) {
+
+func (l *deprecatedRecorder) AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...any) {
 	l.ensureRecording()
 
 	l.prov.lock.RLock()
--- a/pkg/internal/recorder/recorder_integration_test.go
+++ b/pkg/internal/recorder/recorder_integration_test.go
@@ -21,7 +21,9 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/scheme"
 	ref "k8s.io/client-go/tools/reference"
@@ -36,20 +38,22 @@ import (
 )
 
 var _ = Describe("recorder", func() {
-	Describe("recorder", func() {
+	Describe("deprecated recorder", func() {
 		It("should publish events", func(ctx SpecContext) {
 			By("Creating the Manager")
 			cm, err := manager.New(cfg, manager.Options{})
 			Expect(err).NotTo(HaveOccurred())
 
 			By("Creating the Controller")
-			recorder := cm.GetEventRecorderFor("test-recorder")
+			deprecatedRecorder := cm.GetEventRecorderFor("test-deprecated-recorder") //nolint:staticcheck
+			recorder := cm.GetEventRecorder("test-recorder")
 			instance, err := controller.New("foo-controller", cm, controller.Options{
 				Reconciler: reconcile.Func(
 					func(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 						dp, err := clientset.AppsV1().Deployments(request.Namespace).Get(ctx, request.Name, metav1.GetOptions{})
 						Expect(err).NotTo(HaveOccurred())
-						recorder.Event(dp, corev1.EventTypeNormal, "test-reason", "test-msg")
+						deprecatedRecorder.Event(dp, corev1.EventTypeNormal, "deprecated-test-reason", "deprecated-test-msg")
+						recorder.Eventf(dp, nil, corev1.EventTypeNormal, "test-reason", "test-action", "test-note")
 						return reconcile.Result{}, nil
 					}),
 			})
@@ -66,7 +70,7 @@ var _ = Describe("recorder", func() {
 			}()
 
 			deployment := &appsv1.Deployment{
-				ObjectMeta: metav1.ObjectMeta{Name: "deployment-name"},
+				ObjectMeta: metav1.ObjectMeta{Name: "deprecated-deployment-name"},
 				Spec: appsv1.DeploymentSpec{
 					Selector: &metav1.LabelSelector{
 						MatchLabels: map[string]string{"foo": "bar"},
@@ -89,23 +93,42 @@ var _ = Describe("recorder", func() {
 			deployment, err = clientset.AppsV1().Deployments("default").Create(ctx, deployment, metav1.CreateOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
-			By("Validate event is published as expected")
-			evtWatcher, err := clientset.CoreV1().Events("default").Watch(ctx, metav1.ListOptions{})
+			// watch both deprecated and new events based on the reason
+			By("Validate deprecated event is published as expected")
+			deprecatedEvtWatcher, err := clientset.CoreV1().Events("default").Watch(ctx,
+				metav1.ListOptions{FieldSelector: fields.OneTermEqualSelector("reason", "deprecated-test-reason").String()})
 			Expect(err).NotTo(HaveOccurred())
 
-			resultEvent := <-evtWatcher.ResultChan()
+			resultEvent := <-deprecatedEvtWatcher.ResultChan()
 
 			Expect(resultEvent.Type).To(Equal(watch.Added))
-			evt, isEvent := resultEvent.Object.(*corev1.Event)
+			deprecatedEvt, isEvent := resultEvent.Object.(*corev1.Event)
 			Expect(isEvent).To(BeTrue())
 
 			dpRef, err := ref.GetReference(scheme.Scheme, deployment)
 			Expect(err).NotTo(HaveOccurred())
 
-			Expect(evt.InvolvedObject).To(Equal(*dpRef))
+			Expect(deprecatedEvt.InvolvedObject).To(Equal(*dpRef))
+			Expect(deprecatedEvt.Type).To(Equal(corev1.EventTypeNormal))
+			Expect(deprecatedEvt.Reason).To(Equal("deprecated-test-reason"))
+			Expect(deprecatedEvt.Message).To(Equal("deprecated-test-msg"))
+
+			By("Validate event is published as expected")
+			evtWatcher, err := clientset.EventsV1().Events("default").Watch(ctx,
+				metav1.ListOptions{FieldSelector: fields.OneTermEqualSelector("reason", "test-reason").String()})
+			Expect(err).NotTo(HaveOccurred())
+
+			resultEvent = <-evtWatcher.ResultChan()
+
+			Expect(resultEvent.Type).To(Equal(watch.Added))
+			evt, isEvent := resultEvent.Object.(*eventsv1.Event)
+			Expect(isEvent).To(BeTrue())
+
+			Expect(evt.Regarding).To(Equal(*dpRef))
 			Expect(evt.Type).To(Equal(corev1.EventTypeNormal))
 			Expect(evt.Reason).To(Equal("test-reason"))
-			Expect(evt.Message).To(Equal("test-msg"))
+			Expect(evt.Action).To(Equal("test-action"))
+			Expect(evt.Note).To(Equal("test-note"))
 		})
 	})
 })
--- a/pkg/leaderelection/leader_election.go
+++ b/pkg/leaderelection/leader_election.go
@@ -127,8 +127,10 @@ func NewResourceLock(config *rest.Config, recorderProvider recorder.Provider, op
 		corev1Client,
 		coordinationClient,
 		resourcelock.ResourceLockConfig{
-			Identity:      id,
-			EventRecorder: recorderProvider.GetEventRecorderFor(id),
+			Identity: id,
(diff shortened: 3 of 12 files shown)


clebs:
Open Items:
- `StartRecordingToSink`: manage the context/channel in the provider on `ensureRecording` and `Stop`.
- `StartWatcher`: see if this is still needed and how to adapt it if it is.
- Leader election: `resourcelock.ResourceLockConfig` is still using the old API.
  This has now been solved by adding an adapter that turns old API calls into new API calls (albeit pretty simplistic). So that way on the backend we have only the new API and places that require the old can use the adapter.
- Update unit tests.

pkg/recorder/recorder.go
```diff
@@ -21,11 +21,15 @@ limitations under the License.
...
 import (
+	"k8s.io/client-go/tools/events"
 	"k8s.io/client-go/tools/record"
 )
 
 // Provider knows how to generate new event recorders with given name.
 type Provider interface {
 	// NewRecorder returns an EventRecorder with given name.
-	GetEventRecorderFor(name string) record.EventRecorder
+	GetEventRecorder(name string) events.EventRecorder
```
alvaroaleman, line 35: Why rename this?

pkg/recorder/recorder.go
```diff
@@ -21,11 +21,15 @@ limitations under the License.
...
+	// GetOldEventRecorder returns an EventRecorder for the old events API.
+	// The old API is not 100% supported anymore, use the new one whenever possible.
+	GetOldEventRecorder(name string) record.EventRecorder
```
alvaroaleman: Rather than extending the interface here, would it be possible to build a wrapper that makes a `record.EventRecorder` out of a `events.EventRecorder`?

alvaroaleman (commented):
two questions

pkg/recorder/recorder.go
clebs: It was suggested by Stefan here https://github.com/kubernetes-sigs/controller-runtime/issues/2141#issuecomment-2263531702

pkg/recorder/recorder.go
clebs: Yes, it is already built as a wrapper under the hood in `intrec`. I added it here with the idea to support both APIs for some time. If it is not what we are going for, having the wrapper only where it is needed is better.

pkg/recorder/recorder.go
alvaroaleman: I think that suggestion was to aid in making this a non-breaking change

clebs: I interpreted it as a general suggestion, my bad. I will revert that change.

pkg/recorder/recorder.go
alvaroaleman: I think there are two main ways we could go about this:
1. We make this a non-breaking change by keeping the existing `GetEventRecorderFor` with the existing signature and add a new `GetEventRecorder` with the new `events.EventRecorder` as return but mark the `GetEventRecorderFor` as `// deprecated`. At some point in a future release we will then remove `GetEventRecorderFor`
2. We make this a breaking change by changing the signature of the existing `GetEventRecorderFor` to return an `events.EventRecorder` and we provide an adapter somewhere to go from `events.EventRecorder` to a `record.EventRecorder` for places like the leader election

My vote if this is possible would be to do 1) as this is the least disruptive for c-r users but its possible I've missed something and it isn't possible

clebs: Option 1 is doable, but I have to see if the wrapper approach is good enough. Reason being, that the wrapper calls the new API under the hood and both APIs are quite different.  The old API has `AnnotatedEventf` while the new one does not and the new one has things like `relatedObject` and `action`.

Otherwise, the other way to support both is having 2 distinct implementations in separate packages and duplicate the broadcaster, provider, interfaces and injection at the manager and cluster levels.
Finally also duplicate the tests so both versions are covered.

I will test if the wrapper is viable, present the results and then we can make an informed decision.

pkg/recorder/recorder.go
sbueringer: Ah yeah. I only meant if we introduce another func, might as well improve the name

pkg/internal/recorder/recorder.go
```diff
@@ -141,41 +143,42 @@ type lazyRecorder struct {
...
-func (l *lazyRecorder) AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{}) {
-	l.ensureRecording()
 
-	l.prov.lock.RLock()
-	if !l.prov.stopped {
-		l.rec.AnnotatedEventf(object, annotations, eventtype, reason, messageFmt, args...)
-	}
-	l.prov.lock.RUnlock()
+func (l *oldRecorder) Event(object runtime.Object, eventtype, reason, message string) {
+	l.newRecorder.Eventf(object, nil, eventtype, reason, "unsupported", message)
```
valdar: I think here a less "intimidating" wording could be used? like "no action" instead of "unsupported" maybe?

clebs: Sounds good to me, will change it.

pkg/recorder/recorder.go
clebs: I have tested out the wrapper and it has the one downside: we loose the ability to set annotations on events since there is no equivalent `AnnotatedEventf`. I also have tried to get some info on why that is the case.

Seeing that, the wrapper can not provide the same features as the old API has, so the only option to support both is probably having both implementations at the same time.

@alvaroaleman my proposal is to duplicate the internal events package and have both implementations separate from each other even if that means lots of duplication for some time. It will make it easier later on to cleanup the old API when removed. What do you think?

alvaroaleman: > @alvaroaleman my proposal is to duplicate the internal events package and have both implementations separate from each other even if that means lots of duplication for some time. It will make it easier later on to cleanup the old API when removed. What do you think?

Not great, but given its for a limited time only that seems okay to me

clebs: I am also not a great fan.

I will do the exercise and see how much I can reuse without having a tangled nightmare.

Worst case it is all duplicate.

clebs: @alvaroaleman I have gone through this and now this PR supports both the old and new events APIs with as little duplication as possible.

There is still a fair amount of duplication but it is confined mostly to internal code and should still be relatively easy to cleanup later on.

If you agree with this approach I will move to adding tests for the new API and move this out of draft.

pkg/internal/recorder/recorder.go
clebs: This change is now irrelevant since the wrapper approach to still support the old API has been discarded (see discussion above).

pkg/recorder/recorder.go
alvaroaleman: > If you agree with this approach I will move to adding tests for the new API and move this out of draft.


Sounds good to me

clebs: Tests added and moved out of draft. PTAL!
Still finishing up the recorder integration test for the new API, but the code should be ready for review.

pkg/internal/recorder/recorder.go
```diff
@@ -89,25 +97,35 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
...
+		p.deprecatedBroadcaster.StartEventWatcher(
 			func(e *corev1.Event) {
 				p.logger.V(1).Info(e.Message, "type", e.Type, "object", e.InvolvedObject, "reason", e.Reason)
 			})
-		p.broadcaster = broadcaster
-		p.stopBroadcaster = stop
+
+		// init new broadcaster
+		p.stopCh = make(chan struct{})
+		p.broadcaster.StartRecordingToSink(p.stopCh)
```
sbueringer: This seems deprecated. Let's use StartRecordingToSinkWithContext instead

pkg/internal/recorder/recorder.go
```diff
@@ -89,25 +97,35 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
...
+		_, _ = p.broadcaster.StartEventWatcher(func(event runtime.Object) {
```
sbueringer: Looks like this returns an error. I think we should handle it in some way.

At the very least let's log the error if it's not possible to propagate it up

pkg/internal/recorder/recorder.go
sbueringer: I think they forgot to add an empty line above // Deprecated in client-go and that's why the deprecation is not actually working.

Or at least I think that's how staticcheck / Intellij is detecting deprecations, not sure if this is generally applicable.

I took a look, there are a lot of cases in client-go where they have the empty line, but also a lot where they don't have it.

I've started a thread in sig-apimachinery: https://kubernetes.slack.com/archives/C0EG7JC6T/p1754999884784489

sbueringer: Please check all `// Deprecated` comments in this PR that they have the empty line. I'll open a PR for main

pkg/leaderelection/leader_election.go
```diff
@@ -128,7 +128,7 @@ func NewResourceLock(config *rest.Config, recorderProvider recorder.Provider, op
 		coordinationClient,
 		resourcelock.ResourceLockConfig{
 			Identity:      id,
-			EventRecorder: recorderProvider.GetEventRecorderFor(id),
+			EventRecorder: recorderProvider.GetEventRecorderFor(id), //nolint:staticcheck
```
sbueringer, line 133: What would be the plan to resolve this one?
Improve the upstream package so it can handle the new recorder, then use it here?

(if yes, let's make sure we have corresponding follow-up issues in CR & k/k)

pkg/manager/manager.go
```diff
@@ -337,7 +339,11 @@ func New(config *rest.Config, options Options) (Manager, error) {
 		return nil, errors.New("must specify Config")
 	}
 	// Set default values for options fields
-	options = setOptionsDefaults(options)
+	options, err := setOptionsDefaults(config, options)
+	if err != nil {
+		options.Logger.Error(err, "Failed to set defaults")
```
sbueringer: I would prefer if we only return an error and not log & return the error
I would add more context to the error though

pkg/recorder/recorder.go
```diff
@@ -21,11 +21,17 @@ limitations under the License.
...
 
 // Provider knows how to generate new event recorders with given name.
 type Provider interface {
-	// NewRecorder returns an EventRecorder with given name.
+	// GetEventRecorderFor returns an EventRecorder for the old events API.
+	//
+	// Deprecated: this uses the old events API and will be removed in a future release. Please use GetEventRecorder instead.
 	GetEventRecorderFor(name string) record.EventRecorder
+	// GetEventRecorder returns a EventRecorder with given name.
+	// The old API is not 100% supported anymore, use the new one whenever possible.
```
sbueringer: What does "not 100% supported anymore" mean?

sbueringer:
@clebs Just a heads up. If we want to get this into the next CR minor release the deadline is in 2-3 weeks. We want to release the next CR minor shortly after the Kubernetes minor which is scheduled for 27th August

pkg/leaderelection/leader_election.go
clebs: Correct, this would require `client-go` to migrate to the new API too. I will make sure to create followups once this is in.

pkg/recorder/recorder.go
clebs: This is something I forgot to remove. It made sense on the previous iteration where the old API was just a wrapper around the new one and therefore not all functionality was available.

Will delete it.

pkg/internal/recorder/recorder.go
clebs: Right, I used this one since it is more straightforward and seems more fitting in this context.
But if it is indeed going away I will change it.

clebs:
> @clebs Just a heads up. If we want to get this into the next CR minor release the deadline is in 2-3 weeks. We want to release the next CR minor shortly after the Kubernetes minor which is scheduled for 27th August

@sbueringer that should be doable. I will work in your remarks today.
Only risk I see, is if I am unable to get the newly added integration test to pass. I added a separate `testenv` to test the new and old APIs sending events in isolation from each other.
This worked locally on my machine but seems to timeout in CI.

Maybe you have a suggestion there: https://github.com/kubernetes-sigs/controller-runtime/blob/95e07d91515340c2cb81a691034ed80b5e7c9a1a/pkg/internal/recorder/recorder_integration_test.go#L118-L127

pkg/manager/manager.go
clebs: This was taken from here: https://github.com/kubernetes-sigs/controller-runtime/blob/682465344b9b74efad4657016668e62438000541/pkg/cluster/cluster.go#L161-L165

I will correct both locations.

pkg/internal/recorder/recorder.go
sbueringer: Functionality-wise it's the same (there have been similar changes across a few other components from client-go that we use)

pkg/leaderelection/leader_election.go
sbueringer: Perfect, thx!

pkg/manager/manager.go
sbueringer: Thx! I know we're fairly inconsistent with this. For new code I mostly try to stick to only returning errors.

Just not a fan of duplicate errors in logs :)

pkg/recorder/recorder.go
sbueringer: Got it, thx!

sbueringer:
> Maybe you have a suggestion there:

Very quick first look. The error seems to be in another test:
* "Manager Suite: [It] manger.Manager should not leak goroutines if the default event broadcaster is used & events are emitted" (`pkg/manager/manager_test.go:1900`)

sbueringer:
I'm wondering if the client we are using in this test case is using the clientTransport for which we are force closing idle connections here: https://github.com/kubernetes-sigs/controller-runtime/blob/bc36d51f1feacbc2967b0e626c613948d6faee5a/pkg/manager/manager_test.go#L1947

(I think it should, but maybe it's somehow not)

sbueringer:
/retest

I think one of the two failed test cases is an unrelated flake

pkg/manager/manager.go
clebs: Absolutely, keeping everything consistent is hard and fixing inconsistencies when found is the only way to achieve it long term.

pkg/internal/recorder/recorder.go
clebs: Errors are now logged here.

pkg/internal/recorder/recorder.go
clebs: Replaced with non deprecated function.

pkg/recorder/recorder.go
clebs: Removed outdated comment.

clebs:
I have worked in all the feedback, thanks!

I will try to figure out what is wrong with the tests. This new test is definitely affecting other suites somehow.
I did try many different approaches to test the new API without having a separate testenv, but none worked.

pkg/internal/recorder/recorder.go
```diff
@@ -89,25 +97,42 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
...
 				p.logger.V(1).Info(e.Message, "type", e.Type, "object", e.InvolvedObject, "reason", e.Reason)
 			})
-		p.broadcaster = broadcaster
-		p.stopBroadcaster = stop
+
+		// init new broadcaster
+		ctx, cancel := context.WithCancel(context.Background())
+		p.cancelSinkRecordingFunc = cancel
+		if err := p.broadcaster.StartRecordingToSinkWithContext(ctx); err != nil {
+			p.logger.Error(err, "error starting recording for broadcaster")
```
sbueringer, line 123: I wonder if we should return here. I guess the StartEventWatcher below does nothing if this fails here?

clebs: Had not thought about it but yes, if this fails no events will be broadcast.

pkg/internal/recorder/recorder.go
```diff
@@ -89,25 +97,43 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
...
```
sbueringer, line 123: Sorry for the back and forth on this. I thought a bit more about it. Can you please check if we can return the error here (and below)

I think otherwise a controller just silently won't produce events if this fails and that seems bad

clebs: I checked this out and we can propagate the errors by using a closure on the `once.Do` funcs.
Not sure how to handle that once propagated though. Errors will end up in the following functions that do not return errors:
- `Provider.Stop()` (used in defer statements inside goroutines where we can not handle the error properly)
- `deprecatedRecorder.Event()/Eventf()/AnnotatedEventf()` (old events API, can´t change the signature)
- `lazyRecorder.Eventf()` (new events API,  can´t change the signature).

We could probably refactor the provider to move the broadcaster initialisation to `NewProvider()`, which would allow us to handle the error and stop execution gracefully.
But it won´t be lazy loading anymore and will leak goroutines if not used. I am afraid I do not have enough historic background here to make such a call.

pkg/manager/manager_test.go
```diff
@@ -1858,6 +1852,54 @@ var _ = Describe("manger.Manager", func() {
 		Eventually(func() error { return goleak.Find(currentGRs) }).Should(Succeed())
 	})
 
+	It("should not leak goroutines if the default event broadcaster is used & events are emitted", func(specCtx SpecContext) {
```
clebs, line 1858: @sbueringer I have been debugging this test and when this suite is executed in isolation (`go test -v -race ./pkg/manager --ginkgo.fail-fast`) is succeeds, and so does the other test added to `internal/recorder/recorder_integration_test.go`.

It seems to me I am hitting some global timeout (see last Prow run) and am not sure how to best proceed here.

pkg/internal/recorder/recorder.go
sbueringer: I'll take another look next week.

pkg/manager/manager_test.go
sbueringer: > panic: test timed out after 10m0s
> FAIL	sigs.k8s.io/controller-runtime/pkg/internal/controller	600.306s
> https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3262/pull-controller-runtime-test/1956103172929884160

Looks lke the pkg/internal/controller test is timing out after 10m

If you look through the goroutine dump below `panic: test timed out after 10m0s`
It's probably possible to identify which test case it is
and if you then look at the goroutines maybe you can find the one that should not be stuck? (maybe specficially one that is stuck at a file in the pkg/internal/controller package?)

sbueringer: I took a quick look. I'm wondering if the tests are stuck here: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller_test.go#L1501

Additional hint, maybe this is somehow related to the changes we made here to the fake leader election: https://github.com/kubernetes-sigs/controller-runtime/pull/3192/files#diff-81e768ed724b57562503f8bb74d84301a8f747a68b6adf480ee3cd27c35149ef

EDIT: that PR definitely introduced that unit test. And this log is right above the panic
````
I0814 21:28:12.086587   57229 leaderelection.go:271] successfully acquired lease 54a3ce2e-f1e4-421d-a374-351ff6a586db_2cb0ca97-f953-45d1-8b6a-7d1ae306dc0d
[38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0mI0814 21:28:22.264383   57229 leaderelection.go:257] attempting to acquire leader lease default/some-leader-election-id...
E0814 21:28:22.265657   57229 leaderelection.go:448] error retrieving resource lock default/some-leader-election-id: Get "https://127.0.0.1:35179/apis/coordination.k8s.io/v1/namespaces/default/leases/some-leader-election-id?timeout=5s": context canceled
````

Comment (sbueringer) [reply pkg/manager/manager_test.go:1858]: That might also be just a flake independent of your PR to be honest

**Comment by sbueringer** [comment]
/retest

for more data

File: pkg/manager/manager_test.go (hunk 11 above)
Comment (clebs) [reply pkg/manager/manager_test.go:1858]: I have been running this many times and creating a sort of heatmap of err/fail based on a regex.

Right now my best guess is that this extra `testenv` in the recorder integration test is getting stuck until the 10 min timeout despite the test passing:
```
FAIL    sigs.k8s.io/controller-runtime/pkg/internal/recorder    600.138s
```
 Then the issue pops up randomly in places like manager leak detection or timeouts in leader election, etc...
 
 Looking at the last prow runs, this line consistently shows up.

Comment (clebs) [reply pkg/manager/manager_test.go:1858]: This manager leak detection test is now fixed; the event "search" logic was for the old API, making the test get stuck.

Now only the recorder integration test needs to be fixed and the PR should get green.

File: pkg/internal/recorder/recorder.go (hunk 10 above)
Comment (sbueringer) [reply pkg/internal/recorder/recorder.go:123]: Fun :). Apparently when designing this Events API it was not a concern that there is no way to return errors when events cannot be created

(It's not just about the broadcaster creation here, also later when we actually try to send events)

Comment (sbueringer) [reply pkg/internal/recorder/recorder.go:123]: @alvaroaleman What do you think?

Apparently event publishing is entirely best effort and apart from a few logs nobody would ever notice if it doesn't work (not specific to the current PR, in general).

I personally think that is pretty bad, but I don't see a reasonable way to solve it without building our own thing directly on top of client-go. On the client-go level we get back errors if sending Events fails, but I think there is no way with Broadcaster etc.

So I think I would keep the current implementation on this PR

For these two error cases here. Looks like with the current implementation they can only fail if we hit this error "broadcaster already stopped" in pkg/watch.Broadcaster.Watch. That shouldn't really happen I guess

File: pkg/manager/manager_test.go (hunk 11 above)
Comment (sbueringer) [reply pkg/manager/manager_test.go:1858]: > That might also be just a flake independent of your PR to be honest

I took a look at testgrid. Looks like on main we don't have the issue

Comment (sbueringer) [reply pkg/manager/manager_test.go:1858]: @clebs When looking at the goroutine dump here: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3262/pull-controller-runtime-test/1957193025012633600/build-log.txt

There are a lot of managers from various test cases still running. I wonder if SpecContext is usually canceled after a spec is finished, or only if it times out.

I wonder if we should explicitly cancel all these contexts at the end of test cases and wait for the managers to stop.

> Ginkgo will cancel the SpecContext when a node is interrupted (e.g. by the user sending an interrupt signal) or when a node has exceeded its allowed run-time.

I don't know what's going on here, but it doesn't seem good that we have so many managers running in parallel

Comment (clebs) [reply pkg/manager/manager_test.go:1858]: Yes, main is running all tests fine and so does this PR if I remove the manager and recorder tests I introduced.
Now the new manager test also runs fine and I have an idea on how to make the recorder integration test also run without an additional testenv.

After that I expect all those managers to no longer run past their spec execution.
I will double check the goroutine dumps, but first I want the remaining test to succeed.

Comment (sbueringer) [reply pkg/manager/manager_test.go:1858]: Alrighty!

Comment (clebs) [reply pkg/manager/manager_test.go:1858]: The recorder integration test is fixed too! Instead of having a separate testenv, manager, controller, etc... the same controller now fires both the deprecated and new events and they are filtered on the watcher based on their reason so both can still be checked in isolation.

Tests are now green :grin:

I also do not see any running managers on the logs of the last couple of runs

Unfortunately, the linter is wrongly failing (this flake usually gets solved by running `make clean`) :sweat_smile:

**Comment by clebs** [comment]
> /test pull-controller-runtime-test

@sbueringer not sure I am able to debug the failed pipeline both on prow and the linting GH action.
I have been running the whole thing locally several times and everything is consistently succeeding.

Maybe some caching is making things different on the pipeline? If yes, is there a way to do a clean?

File: pkg/internal/recorder/recorder.go (hunk 10 above)
Comment (alvaroaleman) [reply pkg/internal/recorder/recorder.go:123]: > Apparently event publishing is entirely best effort and apart from a few logs nobody would ever notice if it doesn't work (not specific to the current PR, in general).

Yeah I was actually aware of that. I think an error is logged somewhere in client-go, at least when this fails due to permission issues. I do think this is somewhat okay insofar as that almost everywhere the error handling for this would be to just log it

**Comment by sbueringer** [comment]
> > /test pull-controller-runtime-test
> 
> @sbueringer not sure I am able to debug the failed pipeline both on prow and the linting GH action. I have been running the whole thing locally several times and everything is consistently succeeding.
> 
> Maybe some caching is making things different on the pipeline? If yes, is there a way to do a clean?

Let's ignore the linter (we can ignore the stale cache there worst case by just merging the PR).

Let's rebase the PR on top of latest main. Maybe that changes something somehow. Prow tests the combination of this PR and main. Last time I checked the PR was based on a pretty up-to-date commit from main though. There should be no caching for the ProwJob.
Did you run the tests with race detector?

File: pkg/internal/recorder/recorder.go (hunk 10 above)
Comment (sbueringer) [reply pkg/internal/recorder/recorder.go:123]: Alrighty. Let's go with that for now

**Comment by clebs** [comment]
> > > /test pull-controller-runtime-test
> > 
> > 
> > @sbueringer not sure I am able to debug the failed pipeline both on prow and the linting GH action. I have been running the whole thing locally several times and everything is consistently succeeding.
> > Maybe some caching is making things different on the pipeline? If yes, is there a way to do a clean?
> 
> Let's ignore the linter (we can ignore the stale cache there worst case by just merging the PR).
> 
> Let's rebase the PR on top of latest main. Maybe that changes something somehow. Prow tests the combination of this PR and main. Last time I checked the PR was based on a pretty up-to-date commit from main though. There should be no caching for the ProwJob. Did you run the tests with race detector?

I have been rebasing the PR regularly to avoid drift, but seems the rebase I just pushed did indeed do something and things got green!

**Comment by sbueringer** [comment]
> I have been rebasing the PR regularly to avoid drift, but seems the rebase I just pushed did indeed do something and things got green!

That is extremely strange :). I would have at most expected that tests are now also failing locally as you can run the same locally as in CI. But I'll take it if they are consistently green :). Let's run them a few times

/test pull-controller-runtime-test

**Comment by sbueringer** [comment]
Looks like it's still occuring (at least the linter caching issue is "fixed" now)

**Comment by clebs** [comment]
@sbueringer I indeed have to correct my previous statement saying it consistently succeeds. It does succeed when running `make clean && make test`, but when running `hack/ci-check-everything.sh`, as you suggested, it does fail.

I will look into this and try to get it fixed on CI. It seems the current state makes the tests timeout in different places.

**Comment by sbueringer** [comment]
/retest

(last one was an unrelated flake)

**Comment by sbueringer** [comment]
/test pull-controller-runtime-test

(Not sure if it should be already fixed or not :))

**Comment by clebs** [comment]
I have not made any changes since my last comment where I just rebased.

I ran `ci-check-everything.sh` for a while and it is flaky; Sometimes it succeeds and other times it fails with a variety of unrelated tests failing (e.g. manager, controller, cache, certwatcher...). But I observe the same flakiness on main, at least locally. 

So I have been unable to pin this down and am not sure anymore whether it is still related to this PR.

**Comment by sbueringer** [comment]
Last time I checked I didn't see flakes like this on main in testgrid. Also didn't notice on other PRs. Maybe the changes here make the flake more likely 

Edit: https://storage.googleapis.com/k8s-triage/index.html?pr=1&text=found%20unexpected%20goroutines&job=.*controller-runtime.* (4 occurences, all on this PR) Not sure if I'm searching for the right failure message

**Comment by clebs** [comment]
> Last time I checked I didn't see flakes like this on main in testgrid. Also didn't notice on other PRs. Maybe the changes here make the flake more likely
> 
> Edit: https://storage.googleapis.com/k8s-triage/index.html?pr=1&text=found%20unexpected%20goroutines&job=.*controller-runtime.* (4 occurences, all on this PR) Not sure if I'm searching for the right failure message

Thanks for the infos @sbueringer ! It has helped me focus on this one issue and I found the leaking goroutine I think (on my potato laptop I get lots of noise from other failures when testing thus hard to confirm 100%).

TLDR; I was ignoring the stopFunc returned by `StartEventWatcher()`, now this is called when the broadcaster is shutdown.

**Comment by sbueringer** [comment]
Very nice that the test found an actual issue :)

/test pull-controller-runtime-test

(I'm going to run the test a few more times just for more data)

**Comment by sbueringer** [comment]
Damn :/

I'm starting to lean towards intentionally not merging this into the current minor release, even if the tests are fine a few times in a row eventually, because I would prefer data from testgrid over a longer period to ensure we're not introducing an issue

**Comment by clebs** [comment]
> Damn :/
> 
> I'm starting to lean towards intentionally not merging this into the current minor release, even if the tests are fine a few times in a row eventually, because I would prefer data from testgrid over a longer period to ensure we're not introducing an issue

Agreed, this needs more effort. 
The goroutine from the watcher was definitely leaking since we were never stopping it but there seems to be more to it.

I am not able to properly work on this with my current laptop. I will resume testing when I have access to modern hardware in a few days.

**Comment by clebs** [comment]
@sbueringer I have revisited this and tipped by [this comment](https://github.com/kubernetes-sigs/controller-runtime/blob/1c75cb0185ee6eb2e5571b14fd5724acee5c34d6/pkg/internal/recorder/recorder.go#L93-L97) I suspected that the event broadcaster goroutine leaks may happen due to a race condition where the broadcaster is started and shutdown in quick succession.

To avoid the race on the test, a short sleep has been added between start and shutdown. I went for this solution based on the comment linked above deliberately not attempting to fix this weakness in the current implementation.
It might be worth reconsidering once the old API is removed.

I have extensively tested this last commit and there were no leaks.

Realistically this is not an issue to usage since the broadcaster is usually long lived.

File: pkg/manager/manager_test.go [hunk 12]
```diff
@@ -1858,6 +1855,57 @@ var _ = Describe("manger.Manager", func() {
...
+
+			for i, evt := range evts.Items {
+				if evt.Reason == "BallroomBlitz" {
+					return &evts.Items[i]
+				}
+			}
+			return nil
+		}).ShouldNot(BeNil())
+
+		time.Sleep(3 * time.Second)
```
sbueringer, line 1900: Let's please add a comment with some rough hint about why we need this

pkg/internal/recorder/recorder_integration_test.go
```diff
@@ -89,23 +93,41 @@ var _ = Describe("recorder", func() {
...
+
+			resultEvent = <-evtWatcher.ResultChan()
+
+			Expect(resultEvent.Type).To(Equal(watch.Added))
+			evt, isEvent := resultEvent.Object.(*eventsv1.Event)
+			Expect(isEvent).To(BeTrue())
+
+			Expect(evt.Regarding).To(Equal(*dpRef))
 			Expect(evt.Type).To(Equal(corev1.EventTypeNormal))
 			Expect(evt.Reason).To(Equal("test-reason"))
```
sbueringer, line 129: let's also validate action

sbueringer:
/assign @alvaroaleman 

@clebs Feel free to squash if you want (merge bot can also take care of it)

alvaroaleman:
looks good to me, agree regarding the comment about explaining the sleep. Could we link in our code to an upstream issue wrt using the new api in the leader elector?

Thanks a lot for your work here.

pkg/leaderelection/leader_election.go
clebs: It seems there is an upstream k/k issue for this: https://github.com/kubernetes/kubernetes/issues/82846

I will create a followup on CR to migrate this line once it is possible.

pkg/internal/recorder/recorder_integration_test.go
clebs: Done thanks!

pkg/leaderelection/leader_election.go
clebs: #3355 to remove the old events API once this is complete and leader election has migrated as well.

clebs:
> looks good to me, agree regarding the comment about explaining the sleep. Could we link in our code to an upstream issue wrt using the new api in the leader elector?
> 
> Thanks a lot for your work here.

@alvaroaleman added comment explaining the sleep and a comment on leader election referencing https://github.com/kubernetes/kubernetes/issues/82846.

I would assume that the fact that apidiff is failing is expected and not required to merge?

alvaroaleman:
> I would assume that the fact that apidiff is failing is expected and not required to merge?

Yes, its an optional job and its purpose only is to tell us if there are breaking changes

sbueringer:
Thank you very much!!

Really appreciate the persistence. As a side note, this is not the typical experience when contributing to CR :)

/lgtm
/approve
/hold cancel

PR #3263, opened 2025-07-26
:seedling: Stop using context.Background()/TODO() in tests

alvaroaleman:
This change replaces the usage of `context.Background()` and `context.TODO()` in tests with ginkgos `SpecContext` or golangs `t.Context()` in order to ensure contexts are scoped to the test they are for and only valid for the duration of the test.

While this change is massive in terms of LOC, it should be easy to review as it is essentially just the same three changes over and over again.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,7 @@ linters:
     - errchkjson
     - errorlint
     - exhaustive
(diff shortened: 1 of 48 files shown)


sbueringer:
Thx!
lgtm

PR #3264, opened 2025-07-26
:seedling: Clarify our versioning scheme

alvaroaleman:
While its technically true that we do semver because semver allow to do anything for as long as a major version zero is used, referring to it doesn't really explain our versioning scheme. Clarify that in the readme and also mention why we won't publish a major version > 0.

Closes https://github.com/kubernetes-sigs/controller-runtime/issues/601


sbueringer:
Thx!

/lgtm
/approve

Leaving the hold to you

Issue #3267, opened 2025-08-04
fake client: SSA issues: opts propagation and dryPatch bug


ezeriver94:
I'm trying to test new SSA support for FakeClient (after [this PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2981)) and I'm facing some `conflict` errors returned by apimachinery; I found out if using FieldManager without setting the FieldOwner option, the manager name will end up being `unknown` and an error similar to `Apply failed with 1 conflict: conflict with "unknown" using {APIVERSION}: {FIELDPATH}` .
After this, I updated my call to fake client's update methods (Patch/Update) to set `client.FieldOwner("something")` but realised that it was not being propagated within Patch method (while it was fine for Update).
[pkg/client/fake/client.go](../tree/e8c5c5445a20af34453aab1194fc7d64464d2150/pkg/client/fake/client.go#L1135)

> https://github.com/kubernetes-sigs/controller-runtime/blob/e8c5c5445a20af34453aab1194fc7d64464d2150/pkg/client/fake/client.go#L1135-L1146

I believe action should be created like this:
```go
	action := testing.NewPatchActionWithOptions(gvr, accessor.GetNamespace(), accessor.GetName(), patch.Type(), data, *patchOptions.AsPatchOptions())
```

Something similar happened to CREATE:
> https://github.com/kubernetes-sigs/controller-runtime/blob/e8c5c5445a20af34453aab1194fc7d64464d2150/pkg/client/fake/client.go#L873-L877

expected:
```go
	if err := c.tracker.Create(gvr, obj, accessor.GetNamespace(), *createOptions.AsCreateOptions()); err != nil {
		// The managed fields tracker sets gvk even on errors
		_ = ensureTypeMeta(obj, gvk)
		return err
	}
```

Another thing, dryPatch function has the following logic in switch branch for SSA Apply:

> https://github.com/kubernetes-sigs/controller-runtime/blob/e8c5c5445a20af34453aab1194fc7d64464d2150/pkg/client/fake/client.go#L1264-L1278

What is the reason for the deferred function? At least for me this is not working properly, as it tries to add to the tracker the `obj`, which is the zero instance for the received type after this:
> https://github.com/kubernetes-sigs/controller-runtime/blob/e8c5c5445a20af34453aab1194fc7d64464d2150/pkg/client/fake/client.go#L1226-L1229

[...]

PR #3268, opened 2025-08-05
:bug: Fakeclient: Fix some SSA-related bugs

alvaroaleman:
* We didn't pass on the options in Create and Patch, resulting in an incorrect fieldOwner of `unknown` being set
* The `dryPatch` logic had side-effects in the case of SSA and the validation we use it for doesn't apply to SSA
* We never stripped `managedFields` from objects in a List

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3267
/hold


sbueringer:
Thank you very much!

/lgtm
/approve

PR #3269, opened 2025-08-08
:book: Add more docs to SyncPeriod

guettli:
SyncPeriod is likely to get ignored if you use a predicate.

This PR updates the docs to make this more obvious.


pkg/cache/cache.go
```diff
@@ -172,6 +170,18 @@ type Options struct {
 	// is "done" with an object, and would otherwise not requeue it, i.e., we
 	// recommend the `Reconcile` function return `reconcile.Result{RequeueAfter: t}`,
 	// instead of `reconcile.Result{}`.
+	//
+	// SyncPeriod will trigger update events with the old object being equal to the new
+	// object. Excep the cache was out of sync.
```
abdullah599: I believe there might be a typo (excep)

guettli: @abdullah599 thank you! fixed.

pkg/cache/cache.go
```diff
@@ -172,6 +170,18 @@ type Options struct {
...
+	// If you filter update events like this:
+	// Controller.Watch(
+	//     &source.Kind{Type: v1.MyCustomKind},
+	//     &handler.EnqueueRequestForObject{},
+	//     predicate.Or(predicate.GenerationChangedPredicate{},
+	//         predicate.LabelChangedPredicate{}))
+	// then the SyncPeriod will not trigger a Reconcile call, because the update event
+	// will be ignored by the predicate.
+
```
sbueringer:
```suggestion
```

pkg/cache/cache.go
```diff
@@ -42,9 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
-var (
-	defaultSyncPeriod = 10 * time.Hour
-)
+var defaultSyncPeriod = 10 * time.Hour
```
alvaroaleman: Nit: Please avoid stylistic no-op changes, they cluster the git history for no good reason IMHO

guettli: fixed

pkg/cache/cache.go
```diff
@@ -172,6 +172,17 @@ type Options struct {
 	// is "done" with an object, and would otherwise not requeue it, i.e., we
 	// recommend the `Reconcile` function return `reconcile.Result{RequeueAfter: t}`,
 	// instead of `reconcile.Result{}`.
+	//
+	// SyncPeriod will trigger update events with the old object being equal to the new
+	// object, except when the cache was out of sync.
```
sbueringer, line 177: > except when the cache was out of sync

I'm not sure that this is true. Do we have evidence that informer sometimes trigger update events with old object != new object during resync?

guettli: I wrote that, because I thought that this is what SyncPeriod should do: Check if the local cache is in-sync with the api-server. But the question of Stefan makes sense: Do we have evidence? 

@alvaroaleman Do you know more about that?

sbueringer: (Simplified version) Every syncPeriod the informer takes all objects that are stored locally and sends out update events for them.

I took a look at the implementation, but for me it's not clear if oldObj/newObj are always identical or not

xref: https://github.com/kubernetes/kubernetes/blob/5cca03792748714547c39330ac9cb8cb9c7c60ae/staging/src/k8s.io/client-go/tools/cache/controller.go#L567

guettli: I thought that this is a sync between the api-server and the local cache :-)

sbueringer: Nope :) But that's a very frequent misunderstanding

sbueringer: Re-reading the rest of the godoc above. While it contains a lot of correct information I can see how it's misleading

guettli: @sbueringer then it is time for a new PR: [:book: Docs for SyncPeriod: No sync from api-server. by guettli · Pull Request #3277 · kubernetes-sigs/controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/pull/3277)

PR #3273, opened 2025-08-11
:seedling: Certwatcher: Don't require leaderelection

alvaroaleman:
Controller-Runtime starts the certwatcher in the webhook server, which means it always runs regardless of the current replica being leader or not.

It turns out that kubebuilder adds it to the manager and as a result, it only runs on leader replicas. Make it a `LeaderElectionRunnable` and don't require leader election so that it will work correctly even if other projects use it in ways that were not originally anticipated.


sbueringer:
I didn't realize that cert watcher implements the Runnable interface. I hope it works correctly used as a Runnable, but I don't see anything obvious why it wouldn't

/lgtm
/approve

PR #3277, opened 2025-08-13
:book: Docs for SyncPeriod: No sync from api-server.

guettli:
There is a common misunderstanding about SyncPeriod. It does **not** sync between api-server and local cache.

This PR updates the docs to make this more clear.

Related: https://github.com/kubernetes-sigs/controller-runtime/pull/3269#discussion_r2272237638

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -173,16 +173,14 @@ type Options struct {
 	// recommend the `Reconcile` function return `reconcile.Result{RequeueAfter: t}`,
 	// instead of `reconcile.Result{}`.
 	//
-	// SyncPeriod will trigger update events with the old object being equal to the new
-	// object, except when the cache was out of sync.
-	// If you filter update events like this:
-	// Controller.Watch(
-	//     &source.Kind{Type: v1.MyCustomKind},
-	//     &handler.EnqueueRequestForObject{},
-	//     predicate.Or(predicate.GenerationChangedPredicate{},
-	//         predicate.LabelChangedPredicate{}))
-	// then the SyncPeriod will not trigger a Reconcile call, because the update event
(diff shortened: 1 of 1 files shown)


pkg/cache/cache.go
```diff
@@ -148,7 +148,10 @@ type Options struct {
 	Mapper meta.RESTMapper
 
 	// SyncPeriod determines the minimum frequency at which watched resources are
-	// reconciled. A lower period will correct entropy more quickly, but reduce
+	// reconciled. It causes all resources in the local cache to be re-enqueued
+	// for reconciliation, even if there are no new events. It does
+	// not sync between the api-server and the local cache.
+	// A lower period will correct entropy more quickly, but reduce
```
alvaroaleman: I am not a fan of changing this part of the paragaph, because the godoc goes from the big picture to the details/the why to the how, we are now adding parts of the how to the why.

Also it would IMHO be better to explain what does happen, rather than what does not.


WDYT about leaving this part be and changing L176 till the end of the godoc to:
```
	// SyncPeriod will locally trigger an artificial Update event with the same
	// object in both ObjectOld and ObjectNew for everything that is in the
	// cache.
	//
	// Predicates or Handlers that expect ObjectOld and ObjectNew to be different
	// (such as GenerationChangedPredicate) will filter out this event, preventing
	// it from triggering a reconciliation.
```

guettli: @alvaroaleman Thank you for your feedback. Yes, I am fine with your suggestion. For people like me (who are slow on the uptake), I added a final note stating that there is no sync between the server and the local cache. Is that okay for you now?

sbueringer:
Thank you very much!!
/lgtm

PR #3279, opened 2025-08-13
⚠️  Default selectors to Nothing if it is nil for `MatchingLabelsSelector` and `MatchingFieldsSelector`

acumino:
If the selector is not set for `MatchingLabelsSelector` and `MatchingFieldsSelector`, then the list call panics.
This PR updates `MatchingLabelsSelector` and `MatchingFieldsSelector` to default their Selector fields to labels.Nothing() and fields.Nothing() respectively if unset in ApplyToList.

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -645,7 +645,7 @@ type MatchingLabelsSelector struct {
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingLabelsSelector) ApplyToList(opts *ListOptions) {
 	if m.Selector == nil {
-		m.Selector = labels.Everything()
+		m.Selector = labels.Nothing()
 	}
 	opts.LabelSelector = m
 }
@@ -681,7 +681,7 @@ type MatchingFieldsSelector struct {
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingFieldsSelector) ApplyToList(opts *ListOptions) {
 	if m.Selector == nil {
-		m.Selector = fields.Everything()
+		m.Selector = fields.Nothing()
 	}
 	opts.FieldSelector = m
 }
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -644,6 +644,9 @@ type MatchingLabelsSelector struct {
 
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingLabelsSelector) ApplyToList(opts *ListOptions) {
+	if m.Selector == nil {
+		m.Selector = labels.Everything()
+	}
 	opts.LabelSelector = m
 }
 
@@ -677,6 +680,9 @@ type MatchingFieldsSelector struct {
 
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingFieldsSelector) ApplyToList(opts *ListOptions) {
+	if m.Selector == nil {
+		m.Selector = fields.Everything()
+	}
 	opts.FieldSelector = m
 }
 
--- a/pkg/client/options_test.go
+++ b/pkg/client/options_test.go
@@ -36,11 +36,37 @@ var _ = Describe("ListOptions", func() {
 		o.ApplyToList(newListOpts)
 		Expect(newListOpts).To(Equal(o))
 	})
+	It("Should set LabelSelector with MatchingLabelsSelector", func() {
+		labelSelector, err := labels.Parse("a=b")
+		Expect(err).NotTo(HaveOccurred())
+		newListOpts := &client.ListOptions{}
+		newListOpts.ApplyOptions([]client.ListOption{client.MatchingLabelsSelector{Selector: labelSelector}})
+		expectedListOpts := &client.ListOptions{LabelSelector: client.MatchingLabelsSelector{Selector: labelSelector}}
+		Expect(newListOpts).To(Equal(expectedListOpts))
+	})
+	It("Should set LabelSelector to everything with empty MatchingLabelsSelector", func() {
+		newListOpts := &client.ListOptions{}
+		newListOpts.ApplyOptions([]client.ListOption{client.MatchingLabelsSelector{}})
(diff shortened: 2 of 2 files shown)


pkg/client/options.go
```diff
@@ -644,6 +644,9 @@ type MatchingLabelsSelector struct {
 
 // ApplyToList applies this configuration to the given list options.
 func (m MatchingLabelsSelector) ApplyToList(opts *ListOptions) {
+	if m.Selector == nil {
```
alvaroaleman, line 647: Why would a label or field selector be set if it has no value?

acumino: Two reasons - 
- When a list call happens with empty/nil label selector, it lists everything, since `MatchingLabelsSelector` also wants to achieve a similar thing;  therefore, in case of a nil label selector, it should list all the objects.
- Currently if nil is passed for selector in `MatchingLabelsSelector` the list call panics, which should not happen

sbueringer: For what it's worth, this is the official godoc on metav1.LabelSelector
```go
// A label selector is a label query over a set of resources. The result of matchLabels and
// matchExpressions are ANDed. An *empty* label selector matches *all objects*. A *null*
// label selector matches *no objects*.
```

If we follow that semantic here as well null would mean match no objects

acumino: If I use - 
```
client.List(ctx, nodeList, &client.ListOptions{LabelSelector: nil})
```
It lists all objects. What should be the way to go here?


Also


> A *null* label selector matches *no objects*.

If we follow this, then I can modify the PR to use Nothing instead of Everything when `MatchingLabelsSelector`'s selector is not set so it doesn't panic.

alvaroaleman: Could you explain why you have code that ends up with a nil selector and why that is not considered to be a bug?

acumino: For the case where I want to write code that populates the selector based on some condition, the list call will panic cause `MatchingLabelsSelector`'s `Selector` is nil when `someCondition` is false.

```go
selectorOption := &client.MatchingLabelsSelector{}
if someCondition {
    selectorOption.Selector = labels.SelectorFromSet(labels.Set{"key": "value"})
}

err := client.List(ctx, nodeList, selectorOption)
```

If we use labelSelector in the same way, there will be no panic even if the LabelSelector is nil

```go
var labelSelector labels.Selector
if someCondition {
	labelSelector = labels.SelectorFromSet(labels.Set{"key": "value"})
}

err := client.List(ctx, nodeList, &client.ListOptions{LabelSelector: labelSelector})
```

Since `MatchingLabelsSelector` supposed to give the same behaviour as `LabelSelector`, it should not panic in case of Selector value is nil.

alvaroaleman: If you write your code as
```
opts := client.ListOptions{}
if someCondition {
   opts.LabelSelectpr = theSelector
}
if anotherCondition {
   opts.FieldSelector = theFieldSelector
}
```

or

```
var opts []client.ListOption
if someCondition {
  opts = append(opts, client.MatchingLabels{"foo": "bar"})
}
```

This issue doesn't exist.

I do not want to make `selectAll` the default because selection might be relevant for security and if you have code like:

```
func doList(ctx context.Context c client.Client, selectorKey, selectorValue string)(objectList, error){
...
}
```

and `selectorKey` and `selectorValue` are unexpectedly the empty string (bug?), the code defaults to return everything. Defaulting to `Everything()` used to be the default when constructing a selector errored and that has led to unexpected results, I don't want us to do something similar.

acumino: The code could be written as you suggested, but as I mentioned `MatchingLabelsSelector` is supposed to give the same behaviour as LabelSelector, it should not panic in case of Selector value is nil.

We are leaving panic handling on the user end, which is not right. If we don't want to use `Everything` because of security reasons, then `Nothing` should be used, or I can use an empty labelSelector which seems more Ideal. WDYT?
But the main issue here is the code should not panic.

acumino: WDYT about the below changes - 
```go
// ApplyToList applies this configuration to the given list options.
func (m MatchingLabelsSelector) ApplyToList(opts *ListOptions) {
	if m.Selector == nil {
		m.Selector = labels.SelectorFromSet(labels.Set{})
	}
	opts.LabelSelector = m
}
```

alvaroaleman: Having a nil selector in there is incorrect usage and we can not error. We do recover panics though so it seems all right to me.

Your suggestion makes  a nil selector match everything, which as mentioned earlier we do not want.

acumino: > recover panics though so it seems all right to me.

This doesn't allow us to let the panic happen at the first place.

Also, IMO we should allow list all, because as per the doc string  `MatchingLabelsSelector` is supposed to behave the same as LabelSelector, which it doesn't currently.

By my main concern for this PR was that we should not let the panic happen, we should stop it one way or another and not let the user figure it out after trial and error. I am open to suggestions for that.
If we really don't want to list anything in case of nil Selector in `MatchingLabelsSelector`, I can use `labels.Nothing` instead of `Everything`.
This will fix both of our concerns.

alvaroaleman: > Also, IMO we should allow list all, because as per the doc string MatchingLabelsSelector is supposed to behave the same as LabelSelector, which it doesn't currently.

Which could also be fixed by changing LabelSelector rather than `MatchingLabelSelector`.

I don't personally mind the panic, given our default is to recover panics and this is basically incorrect usage and we don't have a way to report that, as the listOpts can not error. I am okay changing it to `labels.Nothing()` if you want to do that, but I would veto changing it to `labels.Everything()` due to potential security implications.

acumino: My main issue was nil pointer panic, I am ok with nothing also.
Update the PR, PTAL.

alvaroaleman:
/hold

in case @sbueringer wants to check

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

PR #3280, opened 2025-08-14
✨ envtest: search the assets index for latest of a release series

cbandy:
This allows configuring the `DownloadBinaryAssetsVersion` field of `envtest.Environment` with only the Major.Minor portions of a Kubernetes API version. In that case, the framework searches the asset index for the latest stable version in that release series.

This matches the behavior of `setup-envtest use {major}.{minor}` without the full complexity of version selectors. The leading `v` is now optional, too. For example:
- `"v1.29"` resolves to [v1.29.4](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L24)
- `"1.27"` resolves to [v1.27.1](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L134)
- `"1.24"` resolves to [v1.24.2](https://github.com/kubernetes-sigs/controller-tools/blob/50b4feecf68291368219a53ef3645dc0de9a4819/envtest-releases.yaml#L222)

Setting the field to Major.Minor.Patch or Major.Minor.Patch-PreRelease behaves the same as before:
- Existing binaries are used without going to the network.
- Prereleases can only be selected by spelling them out, precisely as they are in the asset index.

This is an alternative implementation of #3260. It uses the [k8s.io/apimachinery/pkg/util/version](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version) package and basic comparisons, rather than an external package with range/constraint matchers.

### DIFF
--- a/pkg/envtest/binaries.go
+++ b/pkg/envtest/binaries.go
@@ -32,7 +32,6 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
-	"sort"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/version"
@@ -111,6 +110,25 @@ type archive struct {
 	SelfLink string `json:"selfLink"`
 }
 
+// parseKubernetesVersion returns:
+//  1. the SemVer form of s when it refers to a specific Kubernetes release, or
+//  2. the major and minor portions of s when it refers to a release series, or
+//  3. an error
+func parseKubernetesVersion(s string) (exact string, major, minor uint, err error) {
+	if v, err := version.ParseSemantic(s); err == nil {
+		return v.String(), 0, 0, nil
+	}
+
+	// See two parseable components and nothing else.
+	if v, err := version.ParseGeneric(s); err == nil && len(v.Components()) == 2 {
+		if v.String() == strings.TrimPrefix(s, "v") {
+			return "", v.Major(), v.Minor(), nil
+		}
+	}
+
+	return "", 0, 0, fmt.Errorf("could not parse %q as version", s)
+}
+
 func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAssetsVersion, binaryAssetsIndexURL string) (string, string, string, error) {
 	if binaryAssetsIndexURL == "" {
 		binaryAssetsIndexURL = DefaultBinaryAssetsIndexURL
@@ -125,14 +143,23 @@ func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAsse
 	}
 
 	var binaryAssetsIndex *index
-	if binaryAssetsVersion == "" {
-		var err error
+	switch exact, major, minor, err := parseKubernetesVersion(binaryAssetsVersion); {
+	case binaryAssetsVersion != "" && err != nil:
+		return "", "", "", err
+
+	case binaryAssetsVersion != "" && exact != "":
+		// Look for these specific binaries locally before downloading them from the release index.
+		// Use the canonical form of the version from here on.
+		binaryAssetsVersion = "v" + exact
+
+	case binaryAssetsVersion == "" || major != 0 || minor != 0:
+		// Select a stable version from the release index before continuing.
 		binaryAssetsIndex, err = getIndex(ctx, binaryAssetsIndexURL)
 		if err != nil {
 			return "", "", "", err
 		}
 
-		binaryAssetsVersion, err = latestStableVersionFromIndex(binaryAssetsIndex)
+		binaryAssetsVersion, err = latestStableVersionFromIndex(binaryAssetsIndex, major, minor)
 		if err != nil {
 			return "", "", "", err
 		}
@@ -252,12 +279,19 @@ func downloadBinaryAssetsArchive(ctx context.Context, index *index, version stri
 	return readBody(resp, out, archiveName, archive.Hash)
 }
 
-func latestStableVersionFromIndex(index *index) (string, error) {
+// latestStableVersionFromIndex returns the version with highest [precedence] in index that is not a prerelease.
+// When either major or minor are not zero, the returned version will have those major and minor versions.
+// Note that the version cannot be limited to 0.0.x this way.
+//
+// It is an error when there is no appropriate version in index.
+//
+// [precedence]: https://semver.org/spec/v2.0.0.html#spec-item-11
+func latestStableVersionFromIndex(index *index, major, minor uint) (string, error) {
 	if len(index.Releases) == 0 {
 		return "", fmt.Errorf("failed to find latest stable version from index: index is empty")
 	}
 
-	parsedVersions := []*version.Version{}
+	var found *version.Version
 	for releaseVersion := range index.Releases {
 		v, err := version.ParseSemantic(releaseVersion)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/envtest/binaries.go
```diff
@@ -111,6 +110,25 @@ type archive struct {
...
+//  1. the SemVer form of s when it refers to a specific Kubernetes release, or
+//  2. the major and minor portions of s when it refers to a release series, or
+//  3. zero values
+func interpretKubernetesVersion(s string) (exact string, major, minor uint) {
+	if v, err := version.ParseSemantic(s); err == nil {
+		return v.String(), 0, 0
+	}
+
+	// See two parseable components and nothing else.
+	if v, err := version.ParseGeneric(s); err == nil && len(v.Components()) == 2 {
```
alvaroaleman, line 123: Can we avoid calling `interpretKubernetesVersion` if `binaryAssetssVersion` is the empty string so we can return the error here and don't ~silently default to the latest version~ fall through to use `envtest--$GOOS-GOOARCH.tar.gz` if we can not parse `binaryAssetsVersion`?

pkg/envtest/binaries.go
```diff
@@ -252,34 +279,49 @@ func downloadBinaryAssetsArchive(ctx context.Context, index *index, version stri
...
+		if len(v.PreRelease()) > 0 {
 			continue
 		}
 
-		parsedVersions = append(parsedVersions, v)
+		// Filter on release series, if any.
+		if lower != nil && v.LessThan(lower) {
+			continue
+		}
+		if upper != nil && !v.LessThan(upper) {
```
alvaroaleman: Nit: `GreaterThan` might be easier to read than `!LessThan`

pkg/envtest/binaries_test.go
```diff
@@ -28,13 +28,93 @@ import (
...
+					if major != tc.major {
+						t.Errorf("expected major %v for %q, got %v", tc.major, input, major)
+					}
+					if minor != tc.minor {
+						t.Errorf("expected minor %v for %q, got %v", tc.minor, input, minor)
+					}
+					if exact != "" {
+						t.Errorf("expected no canonical version for %q, got %q", input, exact)
+					}
+				}
```
alvaroaleman: Avoid nesting conditionals if possible please:

```suggestion
				if tc.exact {
				    return
				}
				if major != tc.major {
					t.Errorf("expected major %v for %q, got %v", tc.major, input, major)
				}
				if minor != tc.minor {
					t.Errorf("expected minor %v for %q, got %v", tc.minor, input, minor)
				}
				if exact != "" {
					t.Errorf("expected no canonical version for %q, got %q", input, exact)
				}
			}
```

pkg/envtest/binaries_test.go
```diff
@@ -28,13 +28,93 @@ import (
...
 
+func TestInterpretKubernetesVersion(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name  string
+		in    []string
+		exact bool
+		major uint
+		minor uint
```
alvaroaleman: please name them such that its clear these are expectations for output, i.E. `expectExact`, `expectMajor` etc or something like that

pkg/envtest/binaries_test.go
```diff
@@ -86,6 +166,25 @@ var _ = Describe("Test download binaries", func() {
 		}
 		Expect(actualFiles).To(ConsistOf("some-file"))
 	})
+
+	It("should download binaries of latest stable version of a release series", func(ctx SpecContext) {
```
alvaroaleman, line 194: Please add a testcase for a version string that can not be parsed

pkg/envtest/binaries_test.go
```diff
@@ -28,13 +28,93 @@ import (
...
+		{
+			name: "one number is not a version",
+			in: []string{
+				"1", "v1", "v001", "1.", "v1.", "1.x",
+			},
+		},
+		{
+			name:  "two numbers are a release series",
+			major: 0, minor: 1,
+			in: []string{"0.1", "v0.1"},
```
alvaroaleman: These testcases would IMHO be less confusing to read if we ordered the properties 1. name, 2. input 3. expected outputs , ther than 1.name, 2. expected outputs, 3.input

pkg/envtest/binaries.go
cbandy: The idea was to have an inclusive lower bound and an exclusive upper bound, but upon writing the comment, I realized that effectively means `major == major && minor == minor`.

I've done away with the less/greater entirely.

pkg/envtest/binaries.go
cbandy: I understand what you're asking: When the user fills out the asset version, we should show them an error if it's not version-shaped. I've left this behavior as-is and added Ginkgo tests for error cases. Do you think those error messages are good enough?

> fall through to use `envtest--$GOOS-GOOARCH.tar.gz`

That doesn't happen, thankfully. When we cannot parse what they gave us, we see if it is in the asset index. This is the same behavior as before this PR.

pkg/envtest/binaries.go
```diff
@@ -124,15 +142,22 @@ func downloadBinaryAssets(ctx context.Context, binaryAssetsDirectory, binaryAsse
...
+		// Look for these specific binaries locally before downloading them from the release index.
+		// Use the canonical form of the version from here on.
+		binaryAssetsVersion = "v" + exact
+	} else if binaryAssetsVersion == "" || major != 0 || minor != 0 {
+		// Select a stable version from the release index before continuing.
 		var err error
 		binaryAssetsIndex, err = getIndex(ctx, binaryAssetsIndexURL)
 		if err != nil {
 			return "", "", "", err
 		}
```
sbueringer, line 160: Should we add an else branch here that returns to cover error cases?

pkg/envtest/binaries_test.go
```diff
@@ -28,13 +28,97 @@ import (
...
+
+	testCases := []struct {
+		name   string
+		inputs []string
+
+		expectExact       bool
+		expectSeriesMajor uint
+		expectSeriesMinor uint
+	}{
+		{
```
sbueringer, line 52: Let's add a test case for the empty string

pkg/envtest/binaries.go
cbandy: This line _is_ an early return, so I think you're asking a similar question as here: https://github.com/kubernetes-sigs/controller-runtime/pull/3280#discussion_r2280692113

The behavior before and after this PR is to look in the index for what the user typed when we don't understand how to interpret it. The error comes from not finding it (the nonsensical input) in the asset index.

cbandy: There is a Ginkgo test for strange input here: https://github.com/kubernetes-sigs/controller-runtime/blob/60ef409f0bc46f6bae7c5eb016aa10766d1e58c6/pkg/envtest/binaries_test.go#L197-L200

cbandy:
Failure in a timeout test. Looks unrelated:

```
  [FAILED] Expected an error, got nil
  In [It] at: /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/internal/testing/process/process_test.go:290
```

/retest

pkg/envtest/binaries.go
sbueringer: Sorry my bad. I couldn't place the comment at l.164 because of GitHub UI limitations.

I meant if we should add an else branch in l.164.

I think if someone sets `binaryAssetsVersion` to something unparsable neither the if nor the else if branch are hit.

I would prefer if we handle this error case explicitly in an else branch

cbandy: Done.

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

PR #3282, opened 2025-08-17
:bug: Fakeclient: Validate managed fields on init objects

alvaroaleman:
When the FieldManager in the FieldManagedObjectTracker encounters managed fields it can not decode, it will just silently clear them.

This breaks anyone who has tests where the fakeclient gets initialized with objects that have invalid managed fields. This worked prior to using the `FieldManagedObjectTracker`, as the default `testing.ObjectTracker` we used before has no understanding of them.

Validate the `managedFields` during fake client initialization to make this issue properly visible.

As an alternative to this patch I had initially considered to just try fix them up. That would however require that the fake client understands what constitutes valid managedFields and how they can be fixed up, which might not always be clear and would end up being pretty magic. Thus we just import the validation code and let the user know.

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -81,8 +81,9 @@ import (
 
 type versionedTracker struct {
 	testing.ObjectTracker
-	scheme                *runtime.Scheme
-	withStatusSubresource sets.Set[schema.GroupVersionKind]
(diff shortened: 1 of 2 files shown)


sbueringer:
Thank you!

/lgtm
/approve

PR #3283, opened 2025-08-17
:bug: Fakeclient: Update passed object in Apply

alvaroaleman:
Currently, the object that is passed into the fakeclients `Apply` will not get updated, this change fixes that. It also adds tests that validate that all mutating methods except Delete update the object.


sbueringer:
Thx!

/lgtm
/approve

PR #3284, opened 2025-08-18
:bug: Fix testing kube-apiserver serving certificate using wrong SANs

solidDoWant:
When configuring the kube-apiserver for a test by setting a serving address, the certificate generated for the API server does not include the serving value. In other words, if a test environment is configured like this:

```go
	ControlPlane: envtest.ControlPlane{
		APIServer: &envtest.APIServer{
			SecureServing: envtest.SecureServing{
				ListenAddr: envtest.ListenAddr{
					Address: myTestAddress,
					Port:    strconv.Itoa(6443),
				},
			},
		},
	},
}
```

Then when `testEnv.Start()` is called, the healthcheck will fail because the produced `apiserver.crt` serving certificate is only valid for localhost, and addresses that localhost resolves to.

This fixes the issue by providing the SecureServing and InsecureServing listening address to the cert generator function.

### DIFF
--- a/pkg/internal/testing/controlplane/apiserver.go
+++ b/pkg/internal/testing/controlplane/apiserver.go
@@ -374,9 +374,9 @@ func (s *APIServer) populateAPIServerCerts() error {
 		return err
 	}
 
-	servingAddresses := []string{"localhost", s.SecureServing.ListenAddr.Address}
-	if s.InsecureServing != nil {
-		servingAddresses = append(servingAddresses, s.InsecureServing.Address)
+	servingAddresses := []string{"localhost"}
+	if s.SecureServing.ListenAddr.Address != "" {
+		servingAddresses = append(servingAddresses, s.SecureServing.ListenAddr.Address)
 	}
 
 	servingCerts, err := ca.NewServingCert(servingAddresses...)
--- a/pkg/internal/testing/controlplane/apiserver_test.go
+++ b/pkg/internal/testing/controlplane/apiserver_test.go
@@ -218,6 +218,20 @@ var _ = Describe("APIServer", func() {
 			return cert
 		}
 
+		Context("when SecureServing are not set", func() {
+			It("should have localhost/127.0.0.1 in the certificate altnames", func() {
+				cert := getCertificate()
+
+				Expect(cert.Subject.CommonName).To(Equal("localhost"))
+				Expect(cert.DNSNames).To(ConsistOf("localhost"))
+				expectedIPAddresses := []net.IP{
+					net.ParseIP("127.0.0.1").To4(),
+					net.ParseIP(server.SecureServing.ListenAddr.Address).To4(),
+				}
+				Expect(cert.IPAddresses).To(ContainElements(expectedIPAddresses))
+			})
+		})
+
 		Context("when SecureServing host & port are set", func() {
 			BeforeEach(func() {
 				server.SecureServing = SecureServing{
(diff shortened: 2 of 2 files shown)


pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -374,7 +374,12 @@ func (s *APIServer) populateAPIServerCerts() error {
 		return err
 	}
 
-	servingCerts, err := ca.NewServingCert()
+	servingAddresses := []string{"localhost", s.SecureServing.ListenAddr.Address}
+	if s.InsecureServing != nil {
+		servingAddresses = append(servingAddresses, s.InsecureServing.Address)
+	}
```
sbueringer:
```suggestion
```
I would prefer if we don't handle InsecureServing here. It's already deprecated and I don't want to expand usage of this field if we can avoid it.

Also:
```go
	// Deprecated: does not work with Kubernetes versions 1.20 and above.  Use secure
	// serving instead.
```

pkg/internal/testing/controlplane/apiserver.go
```diff
@@ -374,7 +374,12 @@ func (s *APIServer) populateAPIServerCerts() error {
 		return err
 	}
 
-	servingCerts, err := ca.NewServingCert()
+	servingAddresses := []string{"localhost", s.SecureServing.ListenAddr.Address}
```
sbueringer: Are there code paths where Address might be an empty string? Should we check for that in general just to be safe? (and only add it if it's a not an empty string)

solidDoWant: Not currently, specifically because `configurePorts()` is always called first, which handles this case [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/testing/controlplane/apiserver.go#L202-L221). In addition, the cert generation function supports an empty name [by skipping it](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/testing/certs/tinyca.go#L202-L204). However, this couples the behavior of this function to inner functions and/or outer callers. Would you prefer that I add an explicit check for this here?

sbueringer: Yeah, I think it would be better to just have an explicit check

sbueringer:
Thx!

/lgtm
/assign @alvaroaleman

pkg/internal/testing/controlplane/apiserver_test.go
```diff
@@ -191,6 +196,66 @@ var _ = Describe("APIServer", func() {
...
+						Address: "1.2.3.4",
+						Port:    "5678",
+					},
+				}
+			})
+
+			It("should have the host in the certificate altnames", func() {
+				cert := getCertificate()
+
+				Expect(cert.Subject.CommonName).To(Equal("localhost"))
```
alvaroaleman, line 248: I might be missing it, but where do we test that there is actually a SAN injected? The tests here all seem to only check for `localhost` being there

solidDoWant: The following two `Expect` statements verify that the DNS SAN extension contains `localhost`, and that the IP SAN extension contains `127.0.0.1` and `1.2.3.4`.

There isn't really an easy way to verify that the DNS SAN extension will contain an additional entry if `SecureServing.Address` is a hostname instead of an IP address. This is because the current logic (that is outside the scope of this PR) for determining SAN extension values depends on the test host's DNS resolver. If I were to add say `example.com`, then the underlying business logic would perform a DNS lookup for `example.com`, who's value would be host-dependent. The `localhost` value _only_ works here because `localhost` typically resolves to `127.0.0.1` via the `hosts` file.

PR #3289, opened 2025-08-19
⚠️  Priorityqueue: Make priority opt a pointer

alvaroaleman:
Making the priority opt a pointer allows to dinstinguish "unset" from "0". While this is irrelevant for the priorityqueue itself, it is required for any wrapper that wants to inject a default priority without overriding one that was explicitly set.

### DIFF
--- a/pkg/handler/eventhandler.go
+++ b/pkg/handler/eventhandler.go
@@ -136,7 +136,7 @@ func (h TypedFuncs[object, request]) Create(ctx context.Context, e event.TypedCr
 					priority = LowPriority
 				}
 				q.(priorityqueue.PriorityQueue[request]).AddWithOpts(
-					priorityqueue.AddOpts{Priority: priority},
+					priorityqueue.AddOpts{Priority: &priority},
 					item,
 				)
 			},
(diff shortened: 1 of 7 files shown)


pkg/handler/eventhandler.go
```diff
@@ -136,7 +136,7 @@ func (h TypedFuncs[object, request]) Create(ctx context.Context, e event.TypedCr
 					priority = LowPriority
 				}
 				q.(priorityqueue.PriorityQueue[request]).AddWithOpts(
-					priorityqueue.AddOpts{Priority: priority},
+					priorityqueue.AddOpts{Priority: &priority},
```
sbueringer, line 139: Should this only be set if IsInInitialList is true?

(currently we otherwise set it to &0)

(same below)

pkg/handler/eventhandler_test.go
```diff
@@ -888,7 +888,10 @@ var _ = Describe("Eventhandler", func() {
 						IsInInitialList: false,
 					}, wq)
 
-					Expect(actualOpts).To(Equal(priorityqueue.AddOpts{}))
+					Expect(actualOpts).To(Or(
+						Equal(priorityqueue.AddOpts{}),
```
sbueringer, line 892: Wondering if &0 is correct here

(similar below)

sbueringer:
I think my findings are addressed in #3290 
So I'll merge this PR

/lgtm

Issue #3171, opened 2025-03-21
WithLowPriorityWhenUnchanged only sets priority for Add


sbueringer:
I took a closer look at workqueueWithCustomAddFunc. It currently only overwrites the Add func. Would it also make sense to overwrite AddRateLimited, AddAfter and AddWithOpts?

For AddWithOpts ideally only set priority if not already set, but atm we can't differentiate between priority intentionally set to 0 and the zero value, because Priority in AddOpts is not a pointer (probably it should be).

I know that for our builtin event handler we only have to overwrite Add. But I assume if folks use WithLowPriorityWhenUnchanged they would probably expect that it works for all Add funcs.

sbueringer:
cc @alvaroaleman

alvaroaleman:
Seems okay to me but I think its pretty uncommon to implement a full handler rather than using `handler.Funcs` or such

sbueringer:
If I'm not missing anything even if you use handler.Funcs you would implement a func like this: `func(context.Context, event.TypedCreateEvent[object], workqueue.TypedRateLimitingInterface[request])`

The `workqueue.TypedRateLimitingInterface[request]` parameter then provides all Add* funcs, but only `Add`will set the low priority

sbueringer:
Note to myself. When looking into this we might also want to make AddOpts.Priority a pointer so we can differentiate between not set and priority 0 (the zero value)

PR #3290, opened 2025-08-19
:sparkles:  WithLowPriorityWhenUnchanged: Set Priority for all add methods 

alvaroaleman:
This change makes `WithLowPriorityWhenUnchanged` set the priority for
AddAfter, AddRatelimited and AddWithOpts (if not already set) as well.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3171

Based on top of https://github.com/kubernetes-sigs/controller-runtime/pull/3289 so that needs to merge first
/hold

### DIFF
--- a/pkg/handler/eventhandler.go
+++ b/pkg/handler/eventhandler.go
@@ -19,8 +19,10 @@ package handler
 import (
 	"context"
 	"reflect"
+	"time"
 
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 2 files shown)


pkg/handler/eventhandler.go
```diff
@@ -201,13 +189,28 @@ func WithLowPriorityWhenUnchanged[object client.Object, request comparable](u Ty
 	}
 }
 
-type workqueueWithCustomAddFunc[request comparable] struct {
-	workqueue.TypedRateLimitingInterface[request]
-	addFunc func(item request, q workqueue.TypedRateLimitingInterface[request])
+type workqueueWithDefaultPriority[request comparable] struct {
+	priorityqueue.PriorityQueue[request]
+	priority int
```
sbueringer: I'm wondering if we should also make this a pointer

So that we only call AddWithOpts with a priority if it is actually set (and not &0 if unset)

I know that **currently** for **our** priority queue implementation it doesn't matter (as far as I can tell), still wondering if it would be better

pkg/handler/eventhandler.go
```diff
@@ -223,7 +226,7 @@ func addToQueueCreate[T client.Object, request comparable](q workqueue.TypedRate
 	if evt.IsInInitialList {
 		priority = LowPriority
 	}
-	priorityQueue.AddWithOpts(priorityqueue.AddOpts{Priority: priority}, item)
+	priorityQueue.AddWithOpts(priorityqueue.AddOpts{Priority: &priority}, item)
```
sbueringer: Similar here. Wondering if we should just play it safe
```go
	var priority *int
	if evt.IsInInitialList {
		priority = ptr.To(LowPriority)
	}
	priorityQueue.AddWithOpts(priorityqueue.AddOpts{Priority: priority}, item)
```

(same in l.245)

pkg/handler/eventhandler_test.go
```diff
@@ -888,7 +998,14 @@ var _ = Describe("Eventhandler", func() {
 						IsInInitialList: false,
 					}, wq)
 
-					Expect(actualOpts).To(Equal(priorityqueue.AddOpts{}))
+					if test.overridePriority != 0 {
+						Expect(actualOpts).To(Equal(priorityqueue.AddOpts{After: test.after, RateLimited: test.ratelimited, Priority: ptr.To(test.overridePriority)}))
+					} else {
+						Expect(actualOpts).To(Or(
+							Equal(priorityqueue.AddOpts{After: test.after, RateLimited: test.ratelimited}),
+							Equal(priorityqueue.AddOpts{After: test.after, RateLimited: test.ratelimited, Priority: ptr.To(0)}),
```
sbueringer: Wondering if this line should be here (if we make the changes I suggested in eventhandler.go)

(same in l.1081)

pkg/handler/eventhandler.go
alvaroaleman: Yeah, fair, also makes the tests a bit less ugly, updated it to that everywhere, ptal.

sbueringer:
Thank you very much!!

/lgtm
/approve

/hold cancel

PR #3293, opened 2025-08-21
🌱 certwatcher: add instance-specific logger with cert/key context

s-z-z:
After manually updating a certificate, it’s unclear whether the target certificate was updated. This PR adds identification of the source certificate.

### DIFF
--- a/pkg/certwatcher/certwatcher.go
+++ b/pkg/certwatcher/certwatcher.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/fsnotify/fsnotify"
+	"github.com/go-logr/logr"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


sbueringer:
Just curious. In which case was it necessary for you to distinguish which cert was "updated" at which time?

s-z-z:
Hi, @sbueringer 
I appreciate your curiosity about the changes in the PR.
My controller includes a total of three certWatchers, including metricCertWatcher.
After updating the certificate from the certificate issuer using a tool, I cannot tell from the logs whether the certificate I operated on was updated, because the logs for all certificate updates are the same.

pkg/certwatcher/certwatcher.go
```diff
@@ -208,7 +208,7 @@ func (cw *CertWatcher) ReadCertificate() error {
 		return nil
 	}
```
sbueringer, line 212: Let's create a logger for CertWatcher in ~ l.68 and then use that one everywhere instead
```go
	cw := &CertWatcher{
		certPath: certPath,
		keyPath:  keyPath,
		interval: defaultWatchInterval,
		log:      log.WithValues("cert", certPath, "key", keyPath),
	}
```

s-z-z: Good idea, this simplifies similar changes in the future. As you suggested, I've updated the code.

sbueringer:
Thx!

/ok-to-test

/lgtm
/approve

PR #3296, opened 2025-08-23
:sparkles: Add a ReconciliationTimeout option

alvaroaleman:
This change add a ReconciliatiomTimeout option to the controller options. It defaults to zero, so is backwards-compatible.

While it is easy to do the same through a Reconciler wrapper, this setting is something that pretty much any controller should have set, so add a knob for it to make that a bit simpler.


sbueringer:
Thank you!

/lgtm
/approve

PR #3301, opened 2025-08-28
🌱 cleanup: replace custom signal handling with apiserver.SetupSignalContext as they are identical

s-z-z:
I found that the `pkg/manager/signals` package is identical to the code in `k8s.io/apiserver/pkg/server/`.

The earliest commit comes from https://github.com/kubernetes/kubernetes/pull/50439.

I think we should use it directly instead of copying it, so I propose cleaning it up.

Additionally, there's a divergence trend. In k/k: `SetupSignalHandler` returns a channel, while `SetupSignalContext` returns a context.

Moreover, the current PR only implements an alias-like redirection.

Should we directly remove the `signals` package and replace all `signals.SetupSignalHandler` references with `ctrl.SetupSignalHandler`, performing the redirection only once in `alias.go`? If allowed, I can further streamline it.


pkg/manager/signals/signal.go
```diff
@@ -17,29 +17,7 @@ limitations under the License.
...
-	go func() {
-		<-c
-		cancel()
-		<-c
-		os.Exit(1) // second signal. Exit directly.
-	}()
-
-	return ctx
-}
+var SetupSignalHandler = apiserver.SetupSignalContext
```
sbueringer, line 23: I honestly prefer the copy over introducing a dependency to the apiserver code

s-z-z: Hi, thanks for your reply.
apiserver is already a dependency of cr; this PR doesn't add new packages.

sbueringer: Yes, but we would now depend on this code from the apiserver and we have 0 control over or visibility into how it changes over time

Issue #3313, opened 2025-09-09
Subsequent calls to fake.ClientBuilder.Build() causes a panic


benashz:
Hi,

While attempting to upgrade to  sigs.k8s.io/controller-runtime v0.22.0 from  sigs.k8s.io/controller-runtime v0.21.0, we encountered the following panic in previously passing tests:

```panic: WithObjectTracker and WithTypeConverters are incompatible [recovered]
	panic: WithObjectTracker and WithTypeConverters are incompatible
```

trace:
```
[...]
sigs.k8s.io/controller-runtime/pkg/client/fake.(*ClientBuilder).Build(0x14001178000)
	/Users/bash/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.0/pkg/client/fake/client.go:287 +0xbb8
[...]
```

Steps to reproduce:
- Upgrade  to sigs.k8s.io/controller-runtime v0.22.0
- Create a new client builder instance with `fake.NewFakeClientBuilder()`
- Call `Build()` twice on the same instance

Investigation:

It seems that calling `Build()` on an instance of `fake.ClientBuilder` initializes the instance's `objectTracker` and  its `typeConverters`, which are meant to be mutually exclusive.

Ref:
https://github.com/kubernetes-sigs/controller-runtime/blob/a5894805dce56522d851fdbe5b97cf20dfcdcbbd/pkg/client/fake/client.go#L291-L310

Test code:
https://github.com/hashicorp/vault-secrets-operator/blob/336f6ca2973424c1df5fec2b87701c06d8fcc193/helpers/template_test.go#L1254

Questions:
- Is the above referenced code expected to result in an invalid ClientBuilder i.e.: cause a panic on subsequent calls to `Build()`?
- Aside from adding our own `ObjectTracker` before calling `Build()` what is the best way forward here?

Thanks!

alvaroaleman:
Can you change your code to not build the same builder multiple times? That this ever worked is a coincidence, completely untested and may have unexpected sideeffect. We can add something to make that more explicit.

sbueringer:
Agree, would be probably good to have a panic that just says "Build must not be called multiple times)

benashz:
[...]

PR #3314, opened 2025-09-09
🐛Panic when trying to build more than one instance of fake.ClientBuilder

troy0820:
When using `fake.NewClientBuilder`, you recieve a `client.WithWatch`.  However, having many instances of the `fake.ClientBuilder` can cause issues when trying to call `Build()` where the behavior is inconsistent in testing.  

This will now panic when trying to use two instances of the clientBuilder and using build in both of those instances.

Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3313

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -141,6 +141,7 @@ type ClientBuilder struct {
 	interceptorFuncs      *interceptor.Funcs
 	typeConverters        []managedfields.TypeConverter
 	returnManagedFields   bool
+	isBuilt               bool
 
 	// indexes maps each GroupVersionKind (GVK) to the indexes registered for that GVK.
 	// The inner map maps from index name to IndexerFunc.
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client.go
```diff
@@ -267,6 +268,9 @@ func (f *ClientBuilder) WithReturnManagedFields() *ClientBuilder {
 
 // Build builds and returns a new fake client.
 func (f *ClientBuilder) Build() client.WithWatch {
+	if f.isBuilt {
+		panic("build must not be called multiple times when creating a ClientBuilder")
```
alvaroaleman:
```suggestion
		panic("Build() must not be called multiple times when creating a ClientBuilder")
```

troy0820: Updated with suggestion. Resolving.

alvaroaleman:
thanks!

PR #3319, opened 2025-09-14
:warning: Fakeclient: Fix a number of bugs when updating through apply

alvaroaleman:
Prior to this, updates through Apply did not call `versionedTracker.updateObject` which does a number of things such as:
* Resetting status if the object has a status resource and its not a status update
* Fail if the RV is incorrect
* Bump the RV
* Delete the object if the request removed the last finalizer

The reason this happend is that `versionedTracker` embedded an `testing.ObjectTracker` and as a result, we inherited the `Apply` of that upstream tracker rather than implementing our own.

This PR contains three commits:
1. Move `versionedTracker` into its own file
2. Make `versionedTracker` a `testing.ObjectTracker` and stop embedding its upstream
3. Call `updateObject` from the versioned trackers `Apply` and update things as needed, as well as adding tests

/assign @sbueringer 
/hold

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -660,6 +660,19 @@ var _ = Describe("Fake client", func() {
 			Expect(obj.ObjectMeta.ResourceVersion).To(Equal(trackerAddResourceVersion))
 		})
 
+		It("should reject apply with non-matching ResourceVersion", func(ctx SpecContext) {
+			cl := NewClientBuilder().WithRuntimeObjects(cm).Build()
+			applyCM := corev1applyconfigurations.ConfigMap(cm.Name, cm.Namespace).WithResourceVersion("0")
+			err := cl.Apply(ctx, applyCM, client.FieldOwner("test"))
+			Expect(apierrors.IsConflict(err)).To(BeTrue())
+
+			obj := &corev1.ConfigMap{}
+			err = cl.Get(ctx, client.ObjectKeyFromObject(cm), obj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(obj).To(Equal(cm))
+			Expect(obj.ObjectMeta.ResourceVersion).To(Equal(trackerAddResourceVersion))
+		})
+
 		It("should reject Delete with a mismatched ResourceVersion", func(ctx SpecContext) {
 			bogusRV := "bogus"
 			By("Deleting with a mismatched ResourceVersion Precondition")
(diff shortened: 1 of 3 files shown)


pkg/client/fake/versioned_tracker.go
```diff
@@ -0,0 +1,342 @@
...
+
+	oldObject, err := t.Get(gvr, ns, accessor.GetName())
+	if err != nil {
+		// If the resource is not found and the resource allows create on update, issue a
+		// create instead.
+		if apierrors.IsNotFound(err) && allowCreateOnUpdate {
+			// Pass this info to the caller rather than create, because in the SSA case it
+			// must be created by calling Apply in the upstream tracker, not Create.
+			// This is because SSA considers Apply and Non-Apply operations to be different
+			// even then they use the same fieldManager. This behavior is also observable
```
sbueringer:
```suggestion
			// even when they use the same fieldManager. This behavior is also observable
```

pkg/client/fake/versioned_tracker.go
```diff
@@ -0,0 +1,342 @@
...
+
+	if obj == nil { // Object was deleted in updateObject
+		return nil
+	}
+
+	return t.upstream.Patch(gvr, obj, ns, patchOptions)
+}
+
+// updateObject performs a number of validations and changes before the
+// related to object updates, such as checking and updating the resourceVersion.
```
sbueringer: nit: I think there is something wrong in that sentence "changes before the related to object updates"

pkg/client/fake/versioned_tracker.go
```diff
@@ -0,0 +1,342 @@
...
+func (t versionedTracker) Apply(gvr schema.GroupVersionResource, applyConfiguration runtime.Object, ns string, opts ...metav1.PatchOptions) error {
+	patchOptions, err := getSingleOrZeroOptions(opts)
+	if err != nil {
+		return err
+	}
+	gvk, err := apiutil.GVKForObject(applyConfiguration, t.scheme)
+	if err != nil {
+		return err
+	}
+	applyConfiguration, needsCreate, err := t.updateObject(gvr, gvk, applyConfiguration, ns, false, false, true, patchOptions.DryRun)
```
sbueringer, line 310: I might be wrong but I think allowCreateOnUpdate is not always true with SSA

See: https://github.com/kubernetes-sigs/cluster-api/blob/6fc41b12c0e7f77199118941e79f69461dd14ff4/internal/controllers/topology/cluster/structuredmerge/serversidepathhelper.go#L77-L84

pkg/client/fake/versioned_tracker.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return err
+	}
+
+	if needsCreate && t.withStatusSubresource.Has(gvk) {
+		if err := copyStatusFrom(&unstructured.Unstructured{}, applyConfiguration); err != nil {
```
sbueringer: Is the goal of this just to set the status on `applyConfiguration` to empty?

Maybe let's just add a short comment

sbueringer: Why are we only doing this if needsCreate is true? (I saw the unit test but still wondering why we don't have to set status to empty here)

EDIT: I looked it up. Maybe let's add a godoc comment that for update cases this is already handled in updateObject?

pkg/client/fake/client_test.go
```diff
@@ -714,6 +727,35 @@ var _ = Describe("Fake client", func() {
 			Expect(list.Items).To(ConsistOf(*dep2))
 		})
 
+		It("should handle finalizers in Apply ", func(ctx SpecContext) {
+			cl := client.WithFieldOwner(cl, "test")
+
+			By("Creating the object with a finalizer")
+			cm := corev1applyconfigurations.ConfigMap("test-cm", "delete-with-finalizers").
+				WithFinalizers("finalizers.sigs.k8s.io/test")
+			Expect(cl.Apply(ctx, cm)).NotTo(HaveOccurred())
```
sbueringer: nit: You don't have to change it but maybe going forward let's prefer the following for func calls like this
```
Expect(cl.Apply(ctx, cm)).To(Succeed())
```

pkg/client/fake/versioned_tracker.go
alvaroaleman: Fixed

pkg/client/fake/versioned_tracker.go
alvaroaleman: Fixed

pkg/client/fake/versioned_tracker.go
alvaroaleman: TILd, updated the code, added a test and linked to the relevant code in the KAS, luckily the error message was unique

pkg/client/fake/versioned_tracker.go
alvaroaleman: Added a godoc for both

sbueringer:
Thank you very much!

/lgtm
/approve
/hold cancel

PR #3321, opened 2025-09-20
⚠️ Add subresource apply support

alvaroaleman:
This change adds subresource apply support to the client. It is breaking, as that entails extending the interface.

The fake client currently only supports the `status` subresource for apply and will error out if it is used for any other subresource.

Ref https://github.com/kubernetes-sigs/controller-runtime/issues/3183

/hold

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -545,7 +545,7 @@ func (po *SubResourcePatchOptions) ApplyToSubResourcePatch(o *SubResourcePatchOp
 }
 
 // SubResourceApplyOptions are the options for a subresource
-// apply rquest.
+// apply request.
 type SubResourceApplyOptions struct {
 	ApplyOptions
 	SubResourceBody runtime.ApplyConfiguration
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -544,6 +544,30 @@ func (po *SubResourcePatchOptions) ApplyToSubResourcePatch(o *SubResourcePatchOp
 	}
 }
 
+// SubResourceApplyOptions are the options for a subresource
+// apply rquest.
+type SubResourceApplyOptions struct {
+	ApplyOptions
+	SubResourceBody runtime.ApplyConfiguration
+}
+
+// ApplyOpts applies the given options.
+func (ao *SubResourceApplyOptions) ApplyOpts(opts []SubResourceApplyOption) *SubResourceApplyOptions {
+	for _, o := range opts {
+		o.ApplyToSubResourceApply(ao)
+	}
+
+	return ao
+}
+
+// ApplyToSubResourceApply applies the configuration on the given patch options.
+func (ao *SubResourceApplyOptions) ApplyToSubResourceApply(o *SubResourceApplyOptions) {
+	ao.ApplyOptions.ApplyToApply(&o.ApplyOptions)
+	if ao.SubResourceBody != nil {
+		o.SubResourceBody = ao.SubResourceBody
+	}
+}
+
 func (sc *subResourceClient) Get(ctx context.Context, obj Object, subResource Object, opts ...SubResourceGetOption) error {
 	switch obj.(type) {
 	case runtime.Unstructured:
@@ -593,5 +617,15 @@ func (sc *subResourceClient) Patch(ctx context.Context, obj Object, patch Patch,
 		return sc.client.metadataClient.PatchSubResource(ctx, obj, sc.subResource, patch, opts...)
 	default:
 		return sc.client.typedClient.PatchSubResource(ctx, obj, sc.subResource, patch, opts...)
+	}
+}
+
+func (sc *subResourceClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts ...SubResourceApplyOption) error {
+	switch obj := obj.(type) {
+	case *unstructuredApplyConfiguration:
+		defer sc.client.resetGroupVersionKind(obj, obj.GetObjectKind().GroupVersionKind())
+		return sc.client.unstructuredClient.ApplySubResource(ctx, obj, sc.subResource, opts...)
+	default:
+		return sc.client.typedClient.ApplySubResource(ctx, obj, sc.subResource, opts...)
 	}
 }
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -1335,6 +1335,38 @@ func (sw *fakeSubResourceClient) statusPatch(body client.Object, patch client.Pa
 	return sw.client.patch(body, patch, &patchOptions.PatchOptions)
 }
 
+func (sw *fakeSubResourceClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts ...client.SubResourceApplyOption) error {
+	if sw.subResource != "status" {
+		return errors.New("fakeSubResourceClient currently only supports Apply for status subresource")
+	}
+
+	applyOpts := &client.SubResourceApplyOptions{}
+	applyOpts.ApplyOpts(opts)
+
+	data, err := json.Marshal(obj)
+	if err != nil {
+		return fmt.Errorf("failed to marshal apply configuration: %w", err)
+	}
+
+	u := &unstructured.Unstructured{}
+	if err := json.Unmarshal(data, u); err != nil {
+		return fmt.Errorf("failed to unmarshal apply configuration: %w", err)
+	}
+
+	patchOpts := &client.SubResourcePatchOptions{}
+	patchOpts.Raw = applyOpts.AsPatchOptions()
+
+	if applyOpts.SubResourceBody != nil {
+		subResourceBody := &unstructured.Unstructured{}
+		if err := json.Unmarshal(data, subResourceBody); err != nil {
+			return fmt.Errorf("failed to unmarshal subresource body: %w", err)
+		}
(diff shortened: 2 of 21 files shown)


pkg/client/fake/client.go
```diff
@@ -1335,6 +1335,38 @@ func (sw *fakeSubResourceClient) statusPatch(body client.Object, patch client.Pa
 	return sw.client.patch(body, patch, &patchOptions.PatchOptions)
 }
 
+func (sw *fakeSubResourceClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts ...client.SubResourceApplyOption) error {
+	if sw.subResource != "status" {
```
alvaroaleman, line 1339: This is different from `Patch`, including when using apply through patch which will just patch the main object rather than a subresource when used with subresources other than status.

IMHO it is better to error out rather than to silenly do the wrong thing. We could fo a follow-up to do the same in Patch for consistency (and/or implement the subresources we support through `Update` there as well)

pkg/client/client.go
```diff
@@ -544,6 +544,30 @@ func (po *SubResourcePatchOptions) ApplyToSubResourcePatch(o *SubResourcePatchOp
...
 
+// SubResourceApplyOptions are the options for a subresource
+// apply rquest.
+type SubResourceApplyOptions struct {
+	ApplyOptions
+	SubResourceBody runtime.ApplyConfiguration
+}
+
+// ApplyOpts applies the given options.
+func (ao *SubResourceApplyOptions) ApplyOpts(opts []SubResourceApplyOption) *SubResourceApplyOptions {
```
alvaroaleman, line 555: This differs from our typical naming convention of naming these `ApplyOptions` because that name is already used for the embedded field. I chose to embedd the field, because there is prior art of doing that in the `SubResourcePatchOptions` and it seems ergonomical to do it that way, because modifying options doesn't requier to know whether they are defined in `ApplyOptions` or `SubResourceApplyOptions`.

pkg/client/client.go
```diff
@@ -544,6 +544,30 @@ func (po *SubResourcePatchOptions) ApplyToSubResourcePatch(o *SubResourcePatchOp
 	}
 }
 
+// SubResourceApplyOptions are the options for a subresource
+// apply rquest.
```
sbueringer:
```suggestion
// apply request.
```
nit

pkg/client/fieldvalidation_test.go
```diff
@@ -92,6 +92,15 @@ var _ = Describe("ClientWithFieldValidation", func() {
...
+		invalidApplyConfig := client.ApplyConfigurationFromUnstructured(invalidStatusNode)
+		err = wrappedClient.Status().Apply(ctx, invalidApplyConfig, client.FieldOwner("test-owner"))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("field not declared in schema"))
+
+		err = wrappedClient.SubResource("status").Apply(ctx, invalidApplyConfig, client.FieldOwner("test-owner"))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("field not declared in schema"))
 	})
 })
```
sbueringer, line 106: Not sure if it makes much sense, but should we add some calls to TestWithStrictFieldValidation?

pkg/client/options.go
```diff
@@ -97,6 +97,12 @@ type SubResourcePatchOption interface {
 	ApplyToSubResourcePatch(*SubResourcePatchOptions)
 }
 
+// SubResourceApplyOption configures a subresoruce apply request.
```
sbueringer: nit: typo

pkg/client/patch.go
```diff
@@ -28,10 +28,7 @@ import (
 var (
 	// Apply uses server-side apply to patch the given object.
 	//
-	// This should now only be used to patch sub resources, e.g. with client.Client.Status().Patch().
-	// Use client.Client.Apply() instead of client.Client.Patch(..., client.Apply, ...)
-	// This will be deprecated once the Apply method has been added for sub resources.
-	// See the following issue for more details: https://github.com/kubernetes-sigs/controller-runtime/issues/3183
+	// Deprecated: Use client.Client.Apply() instead.
```
sbueringer: Maybe also mention subresource Apply in some way

pkg/client/typed_client.go
```diff
@@ -304,3 +304,36 @@ func (c *typedClient) PatchSubResource(ctx context.Context, obj Object, subResou
...
+	if err != nil {
+		return err
+	}
+
+	applyOpts := &SubResourceApplyOptions{}
+	applyOpts.ApplyOpts(opts)
+
+	body := obj
+	if applyOpts.SubResourceBody != nil {
+		body = applyOpts.SubResourceBody
```
sbueringer, line 319: What is the use case for letting folks overwrite the obj via the ApplyOptions?

(I don't get where that makes sense compared to just passing in the correct ApplyConfiguration as regular parameter)

pkg/client/fake/client.go
sbueringer: I think tracking a follow-up would be good

pkg/client/fake/client.go
```diff
@@ -1335,6 +1335,38 @@ func (sw *fakeSubResourceClient) statusPatch(body client.Object, patch client.Pa
...
+
+	u := &unstructured.Unstructured{}
+	if err := json.Unmarshal(data, u); err != nil {
+		return fmt.Errorf("failed to unmarshal apply configuration: %w", err)
+	}
+
+	patchOpts := &client.SubResourcePatchOptions{}
+	patchOpts.Raw = applyOpts.AsPatchOptions()
+
+	if applyOpts.SubResourceBody != nil {
```
sbueringer, line 1359: I'm really confused. Why are we overwriting SubResourceBody if it is set?

(Whatever we end up doing here, let's ensure we have test coverage for it)

pkg/client/fake/client.go
alvaroaleman: Added a point to the tracking issue

pkg/client/client.go
alvaroaleman: Fixed

pkg/client/fieldvalidation_test.go
alvaroaleman: Added

pkg/client/options.go
alvaroaleman: Fixed

pkg/client/patch.go
alvaroaleman: Extended it, WDYT?

pkg/client/typed_client.go
alvaroaleman: For example eviction (main object pod, subresource body eviction) or scale (main object deployment/statefulset/a CRD that supports scale, subresource scale).

If someone passed say an eviciton as a main object, we wouldn't know where to send it to/its restmapping, we can only infer that from the pod.

pkg/client/fake/client.go
alvaroaleman: > I'm really confused. Why are we overwriting SubResourceBody if it is set?

What do you mean by overwriting? Appy is implemented through patch, we basically convert applyOpts.SubResourceBody to patchOpts.SubResesourceBody.

That being said, the code is effectively dead as we only support status and there the subResourceBody must not be set (or to the same object as the main object, which doesn't really make sense) or the operation will fail

pkg/client/typed_client.go
sbueringer: Ahhh, thx! (I think I only ever used status subresources 😅)

pkg/client/fake/client.go
sbueringer: My bad. Overlooked that we are writing different opts ..

pkg/client/fieldvalidation_test.go
```diff
@@ -113,8 +122,9 @@ func TestWithStrictFieldValidation(t *testing.T) {
 	_ = wrappedClient.SubResource("some-subresource").Create(ctx, dummyObj, dummyObj)
 	_ = wrappedClient.SubResource("some-subresource").Update(ctx, dummyObj)
 	_ = wrappedClient.SubResource("some-subresource").Patch(ctx, dummyObj, nil)
+	_ = wrappedClient.SubResource("some-subresource").Apply(ctx, corev1applyconfigurations.Namespace(""), nil)
```
sbueringer, line 126: Let's also add the Status().Apply()

pkg/client/fake/client.go
sbueringer: Hm, but there still might be something wrong.

Looks like if `applyOpts.SubResourceBody` is set we are setting `obj` as `patchOpts.SubResourceBody`. Don't we have to use the `applyOpts.SubResourceBody`?

(I get that for status both obj and opts.SubResourceBody have to be of the same type, but I think we should still use `applyOpts.SubResourceBody` here)

alvaroaleman: You are right, good catch, fixed it

pkg/client/fieldvalidation_test.go
alvaroaleman: Done

sbueringer:
Thank you very much!!

/lgtm
/approve
/hold cancel

PR #3327, opened 2025-09-26
✨ Copy all parent context values to leader elector's context

msudheendra-cflt:
The current [implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.1/pkg/manager/internal.go#L449-L454) instantiates a fresh context and passes that along to the leaderElection.Run method which strips any loggers or other context variables. This change preserves the context variables while maintaining the same functionality of passing a fresh context to the leader elector.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3326


alvaroaleman (approved):
thanks!

PR #3329, opened 2025-09-29
🌱 [noop] Modernize finalizer utils

tbavelier:
This modernizes (https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize) the finalizer utils by using modern features of Go and its standard library. No impact for end-users, thus the 🌱.

I was running modernize of one of our projects, leading me to remove our own finalizer utils to use controller-runtime ones instead and found they were not "modernized" themselves.

A larger PR would be to modernize every single file `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./..` but is out of scope here and possibly contrary to the k8s guidelines

### DIFF
--- a/pkg/controller/controllerutil/controllerutil.go
+++ b/pkg/controller/controllerutil/controllerutil.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"slices"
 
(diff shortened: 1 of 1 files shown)


sbueringer:
Thx!

/lgtm
/approve

PR #3332, opened 2025-10-03
✨ Priorityqueue: Enable per default

sbueringer:

Part of #2374

### DIFF
--- a/pkg/config/controller.go
+++ b/pkg/config/controller.go
@@ -79,7 +79,7 @@ type Controller struct {
 	// UsePriorityQueue configures the controllers queue to use the controller-runtime provided
 	// priority queue.
 	//
-	// Note: This flag is disabled by default until a future version. This feature is currently in beta.
+	// Note: This flag is enabled by default.
 	// For more details, see: https://github.com/kubernetes-sigs/controller-runtime/issues/2374.
 	UsePriorityQueue *bool
 
--- a/examples/priorityqueue/main.go
+++ b/examples/priorityqueue/main.go
@@ -24,7 +24,6 @@ import (
 
 	"go.uber.org/zap/zapcore"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	kubeconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/config"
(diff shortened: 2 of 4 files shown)


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -290,9 +290,15 @@ func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool)
 
 	w.notifyItemOrWaiterAdded()
 
-	item := <-w.get
-
-	return item.Key, item.Priority, w.shutdown.Load()
+	select {
```
sbueringer, line 293: This was way to tricky to debug :)

Once PQ was enabled per default, this test started to fail: https://github.com/kubernetes-sigs/controller-runtime/blob/b3590eadfe1bd5521151a79b586a2bd6f376580b/pkg/controller/controller_integration_test.go#L52

* manager stop failed with "failed waiting for all runnable to end within grace period"
* because not all workers of the leader election runnable group stopped
* and they didn't stop because they were stuck in GetWithPriority()

Just tricky to figure out: 
1. which test even caused this (I just got an apiserver stop timeout, the test itself did not fail)
2. which runnable group did not stop
3. which runnable did not stop
4. why does it not stop (is reconcile blocking or something else)

I guess our logging could use some improvements ...

sbueringer: I can move this fix to a separate PR if you prefer, maybe also better for cleaner release notes?

We can also backport if we want

pkg/config/controller.go
```diff
@@ -79,7 +79,7 @@ type Controller struct {
 	// UsePriorityQueue configures the controllers queue to use the controller-runtime provided
 	// priority queue.
 	//
-	// Note: This flag is disabled by default until a future version. This feature is currently in beta.
+	// Note: This flag is enabled by default and will be removed in a future version.
```
alvaroaleman: Is there a reason to remove it? No strong opinion either way, just wondering

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Yeah, lets please do that, might also be worth to backport this. I don't fully understand it though I think, at this point we would've given them an item anyways, what difference does it make? Definitely worth go-docing that and i think it might be worth a test as well

pkg/config/controller.go
sbueringer: Good point. I thought we're mostly going to support the PQ case going forward and using the previous queue is still possible by setting `options.RateLimiter`  and `NewQueue`.

But let's keep the field for now. It's easier for users to configure this with just a bool.

We can still to decide to deprecate this field at some point if we really want to

pkg/controller/priorityqueue/priorityqueue.go
sbueringer: Moved to https://github.com/kubernetes-sigs/controller-runtime/pull/3337

I'll rebase this PR once 3337 merges

sbueringer:
@alvaroaleman Rebased, PTAL

alvaroaleman (approved):
/hold cancel
thanks!

PR #3333, opened 2025-10-03
✨ Add optional Priority field to reconcile.Result

sbueringer:

### DIFF
--- a/pkg/internal/controller/controller.go
+++ b/pkg/internal/controller/controller.go
@@ -459,6 +459,9 @@ func (c *Controller[request]) reconcileHandler(ctx context.Context, req request,
 	// resource to be synced.
 	log.V(5).Info("Reconciling")
 	result, err := c.Reconcile(ctx, req)
+	if result.Priority != nil {
+		priority = *result.Priority
+	}
 	switch {
 	case err != nil:
 		if errors.Is(err, reconcile.TerminalError(nil)) {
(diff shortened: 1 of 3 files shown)


pkg/internal/controller/controller.go
```diff
@@ -468,9 +471,6 @@ func (c *Controller[request]) reconcileHandler(ctx context.Context, req request,
 		}
 		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
 		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelError).Inc()
-		if !result.IsZero() {
```
sbueringer, line 471: If Priority is set result is not zero anymore. Not ideal, but I don't have a better idea.

Independent of that I think it's not especially bad to get rid of this warning.
I've written code specifically to set ctrl.Result{} to zero to avoid this warning which would have been absolutely not necessary otherwise

alvaroaleman: Could explicitly check for Requeue or RequeAfter being set. It is a bit awkward that we now respect a non-error return when an error is returned, that is very uncommon

pkg/reconcile/reconcile.go
```diff
@@ -44,6 +44,12 @@ type Result struct {
 	// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
 	// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
 	RequeueAfter time.Duration
+
+	// Priority is the priority with which the item gets enqueued, if Reconcile returns a non-terminal error,
```
alvaroaleman: The sentence reads a bit awkward to me. Maybe try to keep it simpler with an `Priority that will be used if the item gets re-enqueued` rather than also trying to explain when an item gets re-enqueued? And then maybe also add that that is respected in the error case, since respecting non-error returns on error is uncommon

pkg/internal/controller/controller.go
sbueringer: > Could explicitly check for Requeue or RequeAfter being set.

I'll do that

> It is a bit awkward that we now respect a non-error return when an error is returned, that is very uncommon

I think we don't "respect" it (in the sense of using it), we just don't write a warning on log level 0.

We have a bunch of controllers where the main reconcile logic returns either an error or non-zero ctrl.Result{}, but then we patch the reconciled object with defer. If that patch fails we want to return the error.

Because of this code here in controller-runtime we have to make sure that in these situations we set Result to zero to avoid this warning log: https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/internal/controllers/kubeadmconfig_controller.go#L274-L276

Not sure I see the value of this warning over CR just silently prioritizing error over ctrl.Result{} if both is returned

But this discussion is orthogonal to adding the Priority parameter. So let's continue the discussion (I'll re-add the warning for this PR)

pkg/reconcile/reconcile.go
sbueringer: Good point. PTAL

pkg/internal/controller/controller.go
alvaroaleman: > I think we don't "respect" it (in the sense of using it), we just don't write a warning on log level 0.

What I meant is the `Priority` field, we respect it even in the error case

alvaroaleman (approved):

deefreak:
Hi @alvaroaleman , This feature cannot be seen in latest controller runtime version 0.22.3 , when can we expect to use this?

sbueringer:
After CR v0.23.0 is out, which will be released soon'ish after Kubernetes v1.35.0 (i.e. December or January)

Issue #3130, opened 2025-02-21
Consider making pkg/conversion a Go sub-module


akutz:
Please consider making pkg/conversion a Go sub-module within this project. That way folks that need to implement conversion webhooks in their API modules can do so without pulling in the entire Controller-Runtime dependency graph. Thanks!

sbueringer:
I think we should find a way to implement conversion outside of API packages (just like mutating & validating webhooks today)

sbueringer:
Hm wondering if there is a pretty easy way to make this work

Currently we have the following interfaces

```go
type Convertible interface {
	runtime.Object
	ConvertTo(dst Hub) error
	ConvertFrom(src Hub) error
}

type Hub interface {
	runtime.Object
	Hub()
}
```

I think the only reason why pkg/conversion needs to be imported is because otherwise it's not possible to implement the ConvertTo / ConvertFrom functions.

If we would additionally support implementing the following interface, no import to CR would be needed at all?

```go
type Convertible interface { 
	runtime.Object
	ConvertTo(dst runtime.Object) error
	ConvertFrom(src runtime.Object) error
}
```

akutz:
> I think we should find a way to implement conversion outside of API packages (just like mutating & validating webhooks today)

Yes, 1000 times yes, but what I'm recommending is far easier and does not add any real debt. In fact, my second recommendation is even simpler.

akutz:
By the way, this is what it would look like to make this work with generics:

```diff
diff --git a/pkg/builder/webhook.go b/pkg/builder/webhook.go
index c74742d6..bd80cb55 100644
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -30,13 +30,17 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	pkgconv "sigs.k8s.io/controller-runtime/pkg/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
 )
 
 // WebhookBuilder builds a Webhook.
-type WebhookBuilder struct {
+type WebhookBuilder[
+	THub pkgconv.Hub,
+	TConvertible pkgconv.Convertible[THub],
+] struct {
 	apiType             runtime.Object
[...]

PR #3335, opened 2025-10-04
⚠ Allow implementation of conversion outside of API packages

sbueringer:
Goal is to fix the issue described in https://github.com/kubernetes-sigs/controller-runtime/issues/3130

I.e. making it possible to implement conversion without introducing a dependency to controller-runtime in API packages.

Here is an example how this can be used to move ConvertTo/ConvertFrom funcs out of the API package: https://github.com/kubernetes-sigs/cluster-api/pull/12820/files#diff-f6619b6bca7abdc3fe36434e753669ddefe9cf8c9cf85278ff815d683a9cde45

Note: As conversion functions now don't have to import controller-runtime anymore they could also be kept in the API package.

### DIFF
--- a/pkg/webhook/conversion/conversion.go
+++ b/pkg/webhook/conversion/conversion.go
@@ -43,14 +43,15 @@ var (
 	log = logf.Log.WithName("conversion-webhook")
 )
 
-func NewWebhookHandler(scheme *runtime.Scheme) http.Handler {
-	return &webhook{scheme: scheme, decoder: NewDecoder(scheme)}
+func NewWebhookHandler(scheme *runtime.Scheme, registry Registry) http.Handler {
+	return &webhook{scheme: scheme, decoder: NewDecoder(scheme), registry: registry}
 }
 
 // webhook implements a CRD conversion webhook HTTP handler.
 type webhook struct {
-	scheme  *runtime.Scheme
-	decoder *Decoder
+	scheme   *runtime.Scheme
+	decoder  *Decoder
+	registry Registry
 }
 
 // ensure Webhook implements http.Handler
@@ -119,7 +120,7 @@ func (wh *webhook) handleConvertRequest(ctx context.Context, req *apix.Conversio
 		if err != nil {
 			return nil, err
 		}
-		err = wh.convertObject(src, dst)
+		err = wh.convertObject(ctx, src, dst)
 		if err != nil {
 			return nil, err
 		}
@@ -137,7 +138,7 @@ func (wh *webhook) handleConvertRequest(ctx context.Context, req *apix.Conversio
 // convertObject will convert given a src object to dst object.
 // Note(droot): couldn't find a way to reduce the cyclomatic complexity under 10
 // without compromising readability, so disabling gocyclo linter
-func (wh *webhook) convertObject(src, dst runtime.Object) error {
+func (wh *webhook) convertObject(ctx context.Context, src, dst runtime.Object) error {
 	srcGVK := src.GetObjectKind().GroupVersionKind()
 	dstGVK := dst.GetObjectKind().GroupVersionKind()
 
@@ -147,6 +148,10 @@ func (wh *webhook) convertObject(src, dst runtime.Object) error {
 
 	if srcGVK == dstGVK {
 		return fmt.Errorf("conversion is not allowed between same type %T", src)
+	}
+
+	if converter, ok := wh.registry.GetConverter(srcGVK.GroupKind()); ok {
+		return converter.ConvertObject(ctx, src, dst)
 	}
 
 	srcIsHub, dstIsHub := isHub(src), isHub(dst)
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -45,6 +45,7 @@ type WebhookBuilder struct {
 	customPath                string
 	customValidatorCustomPath string
 	customDefaulterCustomPath string
+	converterConstructor      func(*runtime.Scheme) (conversion.Converter, error)
 	gvk                       schema.GroupVersionKind
 	mgr                       manager.Manager
 	config                    *rest.Config
@@ -83,6 +84,13 @@ func (blder *WebhookBuilder) WithDefaulter(defaulter admission.CustomDefaulter, 
 // WithValidator takes a admission.CustomValidator interface, a ValidatingWebhook will be wired for this type.
 func (blder *WebhookBuilder) WithValidator(validator admission.CustomValidator) *WebhookBuilder {
 	blder.customValidator = validator
+	return blder
+}
+
+// WithConverter takes a func that constructs a converter.Converter.
+// The Converter will then be used by the conversion endpoint for the type passed into For().
+func (blder *WebhookBuilder) WithConverter(converterConstructor func(*runtime.Scheme) (conversion.Converter, error)) *WebhookBuilder {
+	blder.converterConstructor = converterConstructor
 	return blder
 }
 
@@ -287,17 +295,30 @@ func (blder *WebhookBuilder) getValidatingWebhook() *admission.Webhook {
 }
 
 func (blder *WebhookBuilder) registerConversionWebhook() error {
-	ok, err := conversion.IsConvertible(blder.mgr.GetScheme(), blder.apiType)
-	if err != nil {
-		log.Error(err, "conversion check failed", "GVK", blder.gvk)
-		return err
-	}
-	if ok {
-		if !blder.isAlreadyHandled("/convert") {
-			blder.mgr.GetWebhookServer().Register("/convert", conversion.NewWebhookHandler(blder.mgr.GetScheme()))
+	if blder.converterConstructor != nil {
+		converter, err := blder.converterConstructor(blder.mgr.GetScheme())
+		if err != nil {
+			return err
 		}
-		log.Info("Conversion webhook enabled", "GVK", blder.gvk)
+
+		if err := blder.mgr.GetConverterRegistry().RegisterConverter(blder.gvk.GroupKind(), converter); err != nil {
+			return err
+		}
+	} else {
+		ok, err := conversion.IsConvertible(blder.mgr.GetScheme(), blder.apiType)
+		if err != nil {
+			log.Error(err, "conversion check failed", "GVK", blder.gvk)
+			return err
+		}
+		if !ok {
+			return nil
+		}
 	}
+
+	if !blder.isAlreadyHandled("/convert") {
+		blder.mgr.GetWebhookServer().Register("/convert", conversion.NewWebhookHandler(blder.mgr.GetScheme(), blder.mgr.GetConverterRegistry()))
+	}
+	log.Info("Conversion webhook enabled", "GVK", blder.gvk)
 
 	return nil
 }
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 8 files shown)


pkg/webhook/conversion/conversion.go
```diff
@@ -43,14 +45,110 @@ var (
...
+}
+
+type Converter interface {
+	GetHub() client.Object
+	GetSpoke() client.Object
+	ConvertHubToSpoke(hub, spoke runtime.Object) error
+	ConvertSpokeToHub(hub, spoke runtime.Object) error
+}
+
+func NewConverter[hubObject, spokeObject client.Object](
```
alvaroaleman: This confuses me. Shouldn't this take a slice of `ConvertToHub(spoke runtime.Object)(hub runtime.object, _ error)`? Right now, this can only be used if there are exactly to versions (and in that context, the concept of hub and spoke doesn't really make too much sense)

pkg/webhook/conversion/conversion.go
```diff
@@ -43,14 +45,110 @@ var (
...
+		if hubGVK.GroupKind() != converterSpokeGVK.GroupKind() {
+			return fmt.Errorf("converter GVK does not match builder gvk: FIXME")
+		}
+		r.convertersByHubGK[hubGVK.GroupKind()].convertersBySpokeGVK[converterSpokeGVK] = converter
+	}
+
+	return nil
+}
+
+type Converter interface {
```
alvaroaleman: While this should be how it works internally, why do we need to have the concept of hub and spoke visible in the external interface and the webhook be aware of it? All it really wants is a `convert(from, to runtime.Object) error`.

Also without really having context on our current conversion machinery, why don't we use the scheme, it allows to register and call conversion funcs (which internally can be built on a hub and spoke system, but that is nothing the scheme or anything else that wants to convert really cares about)

pkg/webhook/conversion/conversion.go
sbueringer: Some general context. Current state of conversion:
* `Hub` has to be implemented on the hup API type
* `ConvertTo` / `ConvertFrom` has to be implemented on all spoke API types (there's validation for that in CR)
* So overall the following is needed (example from CAPI):
  * v1beta2 API package:
    * `func (*Cluster) Hub()       {}`
  * v1alpha3 / v1alpha4 / v1beta1 API packages:
    * `func (src *Cluster) ConvertTo(dstRaw conversion.Hub) error`
    * `func (dst *Cluster) ConvertFrom(srcRaw conversion.Hub) error`

My main goals are:
* Be able to implement ConvertTo/ConvertFrom outside of API packages
* Of course accordingly ConvertTo/ConvertFrom can't be methods anymore
* Because they are not methods anymore I need a new way to register the funcs (but I still want to be able to verify that all necessary conversions have been provided)
* Make the conversion funcs more type-safe:
  * Today: 
    * `func (src *Cluster) ConvertTo(dstRaw conversion.Hub) error`
    * `func (dst *Cluster) ConvertFrom(srcRaw conversion.Hub) error`
  * With this PR: (func names don't matter, only the parameter)
    * `func ConvertClusterV1Beta1ToHub(src *clusterv1beta1.Cluster, dst *clusterv1.Cluster) error`
    * `func ConvertClusterHubToV1Beta1(src *clusterv1.Cluster, dst *clusterv1beta1.Cluster) error`
* Minimal migration effort for folks that already implemented ConvertTo/ConvertFrom methods and want to move to the new model

sbueringer: I think generics are a nice way to make the conversion funcs more type safe (similar to how this works with source.Kind, e.g.: https://github.com/kubernetes-sigs/cluster-api/blob/a9fbe115c8adf0de2c6a27f4f436ccdf657aa884/internal/controllers/clusterresourceset/clusterresourceset_controller.go#L100)

The webhook builder itself cannot become generic with type parameters for the hub type and an arbitrary amount of spoke types. So I thought I'll use an additioanl type that takes care of the generics and then implements an interface that allows to pass it into the builder and use it later.

A concrete example how this can be used based on Cluster API:
```go
	return ctrl.NewWebhookManagedBy(mgr).
		For(&clusterv1.Cluster{}).
		WithDefaulter(webhook).
		WithValidator(webhook).
		WithConverters(
			conversion.NewConverter(&clusterv1.Cluster{}, &clusterv1beta1.Cluster{}, ConvertClusterHubToV1Beta1, ConvertClusterV1Beta1ToHub),
			conversion.NewConverter(&clusterv1.Cluster{}, &clusterv1alpha4.Cluster{}, ConvertClusterHubToV1Alpha4, ConvertClusterV1Alpha4ToHub),
			conversion.NewConverter(&clusterv1.Cluster{}, &clusterv1alpha3.Cluster{}, ConvertClusterHubToV1Alpha3, ConvertClusterV1Alpha3ToHub),
		).
		Complete()
```

* The generic `NewConverter` func provides a conversion between the hub and one spoke type
* The `hubObject` and `spokeObject` parameter ensure:
  * `WithConverters` can validate that all necessary conversions have been provided (not fully implemented yet in the PR, but it can check hub and spoke types against the type of the webhook + all GK's registered in the scheme)
  * that the passed in conversion functions have the right type parameters

sbueringer: > Shouldn't this take a slice of ConvertToHub(spoke runtime.Object)(hub runtime.object, _ error)?

It could, but it would be less type-safe (side-note: I would prefer to keep hub an input parameter, so that CR stays responible for creating an instance of the hub type).

> Right now, this can only be used if there are exactly two versions (and in that context, the concept of hub and spoke doesn't really make too much sense)

No, see example one comment above. `NewConverter` would be just for one hub-spoke conversion. Users have to pass in all the necessary conversions (similar to how they previously had to implement ConvertTo/ConvertFrom on all spoke types)

pkg/webhook/conversion/conversion.go
```diff
@@ -43,14 +45,110 @@ var (
...
+	GetSpoke() client.Object
+	ConvertHubToSpoke(hub, spoke runtime.Object) error
+	ConvertSpokeToHub(hub, spoke runtime.Object) error
+}
+
+func NewConverter[hubObject, spokeObject client.Object](
+	hub hubObject,
+	spoke spokeObject,
+	convertHubToSpokeFunc func(src hubObject, dst spokeObject) error,
+	convertSpokeToHubFunc func(src spokeObject, dst hubObject) error,
```
sbueringer: TODO for myself: we should start passing in ctx here (e.g. for logging)

pkg/webhook/conversion/conversion.go
sbueringer: > While this should be how it works internally, why do we need to have the concept of hub and spoke visible in the external interface and the webhook be aware of it? All it really wants is a convert(from, to runtime.Object) error.

We could also say just take a `convert(from, to runtime.Object)` func, but then we have no validation on our side and users have to implement the logic in our `convertObject` func on there side. I would prefer if users still only have to implement ~ the ConvertTo/ConvertFrom funcs for all their hub-spoke combinations and we take care of the rest.

> Also without really having context on our current conversion machinery, why don't we use the scheme, it allows to register and call conversion funcs (which internally can be built on a hub and spoke system, but that is nothing the scheme or anything else that wants to convert really cares about)

Today conversions with CR work the following way:
* auto-generated funcs: https://github.com/kubernetes-sigs/cluster-api/blob/1c45137e25328d8c7ccfd70ee9d3b24d04da1c1b/api/core/v1beta1/zz_generated.conversion.go
  * These funcs are automatically registered in the scheme
* ConvertTo/ConvertFrom funcs implemented on top of auto-generated funcs https://github.com/kubernetes-sigs/cluster-api/blob/42ca147923ba1f9367419f4bd1b8748cf872c1a7/api/core/v1beta1/conversion.go

I would prefer if we could avoid mixing these two layers by putting all of these funcs into the scheme.
I also just realized that we should start passing `context.Context` into the ConvertTo/ConvertFrom funcs, that would not be possible with the scheme (it only takes `type ConversionFunc func(a, b interface{}, scope Scope) error` funcs)

If I understand correctly if we would want to delegate the conversion entirely to the scheme we would have to register conversion funcs for all combinations in the scheme, e.g.
* v1beta2 <=> v1beta1, v1beta2 <=> v1alpha4, v1beta2 <=> v1alpha3
* v1beta1 <=> v1alpha4, v1beta1 <=> v1alpha3
* v1alpha4 <=> v1alpha3

Instead of just:
* v1beta2 <=> v1beta1, v1beta2 <=> v1alpha4, v1beta2 <=> v1alpha3

I believe that's why the hub-spoke model was implemented as it is today. I would really prefer if users writing conversion code only have to implement the hub-spoke conversions and not conversion funcs for all combinations.

alvaroaleman: > I would prefer if users still only have to implement ~ the ConvertTo/ConvertFrom funcs for all their hub-spoke combinations and we take care of the rest.

I am not disagreeing with that, but IMHO there should be a separation of concerns. The webhook just takes some sort of converter interface that is not bound to the hub/spoke concept and separately, there is a `HubSpokeConverter` that fullfills that webhook interface. The hub spoke concept doesn't really seem like something the webhook as a user of conversion should be aware of and arguably if someone wanted to optimize for conversion being fast as opposed to minimizing the number of conversions to write, the hub and spoke approach would be the wrong one.

alvaroaleman: Maybe a simpler way to phrase my point: This interface and just this interface should not have any trace of the hub/spoke concept. The actual implementation of this interface absolutely can and probably should.

sbueringer: Got it. I'll take another look how this could be adjusted accordingly

sbueringer: @alvaroaleman Please take another look if that goes in the right direction. I think it does. Once it's good from your side, I'll add unit tests and some other smaller improvements (error handling, ...)

Example usage in Cluster API
```go
	return ctrl.NewWebhookManagedBy(mgr).
		For(&clusterv1.Cluster{}).
		WithDefaulter(webhook).
		WithValidator(webhook).
		WithConverter(conversion.MustNewHubSpokeConverter(mgr.GetScheme(), &clusterv1.Cluster{},
			conversion.NewSpokeConverter(&clusterv1beta1.Cluster{}, ConvertClusterHubToV1Beta1, ConvertClusterV1Beta1ToHub),
			conversion.NewSpokeConverter(&clusterv1alpha4.Cluster{}, ConvertClusterHubToV1Alpha4, ConvertClusterV1Alpha4ToHub),
			conversion.NewSpokeConverter(&clusterv1alpha3.Cluster{}, ConvertClusterHubToV1Alpha3, ConvertClusterV1Alpha3ToHub),
		)).
		Complete()
```

(I know I could avoid having to specify the spoke type in `NewSpokeConverter` and maybe even the hub type in `MustNewHubSpokeConverter` but I think it's better to be explicit like that, compared to inferring the generic types only from the conversion funcs)

alvaroaleman: Yep, I like the new interfaces :+1: To me the explicit type in the NewSpokeConverter seems somewhat redundant especially since it has to match the one from the conversion func but its not a super strong opinion.

Maybe we could avoid the `MustNewHubSpokeConverter` by taking a `func(converter.Converter, error)` in the builder? Again, no super strong opinion

sbueringer: >  To me the explicit type in the NewSpokeConverter seems somewhat redundant especially since it has to match the one from the conversion func but its not a super strong opinion.

Yup, it is redundant, but I prefer to be explicit here.

> Maybe we could avoid the MustNewHubSpokeConverter by taking a func(converter.Converter, error) in the builder? Again, no super strong opinion

Good idea, done

sbueringer:
@alvaroaleman Should be ready for a full review now

alvaroaleman (approved):
Thanks!

pkg/webhook/conversion/conversion.go
```diff
@@ -43,14 +43,15 @@ var (
 	log = logf.Log.WithName("conversion-webhook")
 )
 
-func NewWebhookHandler(scheme *runtime.Scheme) http.Handler {
-	return &webhook{scheme: scheme, decoder: NewDecoder(scheme)}
+func NewWebhookHandler(scheme *runtime.Scheme, registry Registry) http.Handler {
```
alvaroaleman, line 46: I think this requires marking the change is breaking or not?

sbueringer: Good catch, done

sbueringer:
cc @camilamacedo86 there are no best practices around how exactly this should be used (regarding package structure). But this is the preferred way to implement conversion going forward as it allows to avoid controller-runtime dependencies in API packages

PR #3337, opened 2025-10-05
🐛 Don't block on Get when queue is shutdown (2nd try)

sbueringer:

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -290,9 +290,18 @@ func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool) 
 
 	w.notifyItemOrWaiterAdded()
 
-	item := <-w.get
-
-	return item.Key, item.Priority, w.shutdown.Load()
+	select {
+	case <-w.done:
+		// Return if the queue was shutdown while we were already waiting for an item here.
+		// For example controller workers are continuously calling GetWithPriority and
+		// GetWithPriority is blocking the workers if there are no items in the queue.
+		// If the controller and accordingly the queue is then shut down, without this code
+		// branch the controller workers remain blocked here and are unable to shut down.
+		var zero T
+		return zero, 0, true
+	case item := <-w.get:
+		return item.Key, item.Priority, w.shutdown.Load()
+	}
 }
 
 func (w *priorityqueue[T]) Get() (item T, shutdown bool) {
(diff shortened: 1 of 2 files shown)


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -290,9 +290,18 @@ func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool)
 
 	w.notifyItemOrWaiterAdded()
 
-	item := <-w.get
-
-	return item.Key, item.Priority, w.shutdown.Load()
+	select {
```
sbueringer, line 293: *Moved from https://github.com/kubernetes-sigs/controller-runtime/pull/3332#discussion_r2401123826*


This was way to tricky to debug :)

Once PQ was enabled per default, this test started to fail: https://github.com/kubernetes-sigs/controller-runtime/blob/b3590eadfe1bd5521151a79b586a2bd6f376580b/pkg/controller/controller_integration_test.go#L52

* manager stop failed with "failed waiting for all runnable to end within grace period"
* because not all workers of the leader election runnable group stopped
* and they didn't stop because they were stuck in GetWithPriority()

Just tricky to figure out: 
1. which test even caused this (I just got an apiserver stop timeout, the test itself did not fail)
2. which runnable group did not stop
3. which runnable did not stop
4. why does it not stop (is reconcile blocking or something else)

I guess our logging could use some improvements ...

sbueringer: @alvaroaleman:
> Yeah, lets please do that, might also be worth to backport this. I don't fully understand it though I think, at this point we would've given them an item anyways, what difference does it make? Definitely worth go-docing that and i think it might be worth a test as well

sbueringer: > Definitely worth go-docing that and i think it might be worth a test as well

Extended the godoc and added a unit test

>  I don't fully understand it though I think, at this point we would've given them an item anyways, what difference does it make?

We could give the caller only an item if we still have items in the queue, otherwise the caller is deadlocked.

Example:

* controller is started
* controller workers are calling `GetWithPriority` and are waiting for items
* queue is empty and stays empty
* controller (and accordingly queue) is shutdown
* previous code:
  * Shutdown() closes w.done
  * workers remain blocked in l.293
  * accordingly controller and runnable group cannot shutdown
* new code:
  * Shutdown() closes w.done
  * l.301 returns

alvaroaleman: I misread the code the last time and thought this was where we send, not where we get :facepalm: Makes total sense

alvaroaleman (approved):

Issue #3334, opened 2025-10-04
Question: In the reconciliation of an object, is it possible to check if this object has been enqueued again?


matheuscscp:
Context: In Flux we have just shipped a feature to fail faster when kustomize-controller is performing health checks on the manifests it has applied to the cluster. The implementation is essentially a Goroutine fired during the Flux Kustomization reconciliation that continuously checks if the revision of the source consumed by the Flux Kustomization has changed. If it did, then we bail out earlier from the health checks to let the new enqueued reconciliation request of the object take place and apply the new source revision. We want to go beyond and let this happen for any other things we watch beyond the Flux sources. For example, we also watch ConfigMaps and Secrets that have an effect on the final manifests kustomize-controller applies to the cluster. Instead of probing all the things that are relevant for a particular Flux Kustomization object, we would like to probe the reconciliation queue instead to take advantage of the fact that our watches are already in place and will enqueue reconciliation requests. This would be the ideal solution for us.

Is there any function/method of the manager or something else that can answer queries like `IsEnqueued(reconcile.Request) bool`? Would the project consider accepting this contribution in case it's not already available? Or in the worst case would it be possible/simple for us on our side to build a thin wrapper around the default queue implementation to add this functionality? I saw in another issue that now controller-runtime supports injecting your own queue implementation.

@stefanprodan

matheuscscp:
I have started a thread in `#sig-api-machinery` to ask if a contribution for adding `.Has(item)` to the workqueue would be considered:

https://kubernetes.slack.com/archives/C0EG7JC6T/p1759596923301419

alvaroaleman:
As mentioned in the sig-apimachinery thread, I don't think a `Has` is a good solution because what you really want is to get notified and needing to poll is inefficient. And to answer the original question no, there is no direct way to do that at the moment. 

It might be nice to have this to cancel active reconciliations but that would also be a bit of a footgun, because if the reconciliation updates the object and that then causes a cancellation, the reconciliation has effectively just cancelled itself. We may be able to work around that in turn by adding a `ShouldCancel` paramter to `AddWithOpts` that for example only gets set to `true` if the generation has changed.

All of this is to say: I do see value in the feature, but there are some details we'd have to figure out. If you want to volunteer to look into that, that would be highly appreciated. Luckily, we can make it all happen downstream since we already have our own workqueue in the form of the priorityqueue.

stefanprodan:
[...]

PR #3345, opened 2025-10-08
✨ Introduce hook points for the PriorityQueue

matheuscscp:
This change introduces an interface defining hook points for the PriorityQueue. This helps enabling advanced use cases where reacting to internal events of the priority queue may be necessary, like the first hook point being introduced: OnBecameReady.

OnBecameReady is called when items become ready for consumption, i.e. can be handed to GetWithPriority calls.

Closes: #3334


alvaroaleman:
/hold
Until there is consensus on the path forward in https://github.com/kubernetes-sigs/controller-runtime/issues/3334

matheuscscp:
A simpler solution for the problem this PR was trying to solve is described here:

https://github.com/kubernetes-sigs/controller-runtime/issues/3334#issuecomment-3389633701

Issue #3344, opened 2025-10-08
Deserialization Error During Server-Side Apply (SSA) with Protobuf Content Type


filipcirtog:
Hello team! 

## Problem Statement
We're migrating from `client.Patch(Apply)` to `client.Apply().` While making this transition, we've encountered the following issue.

## Observed Behavior

When using `client.Apply` together with a configuration where `config.ContentType = "application/vnd.kubernetes.protobuf"`, an error occurs during the deserialization process.

The protobuf decoder fails to parse the configuration object and throws the following error: `object %v does not implement the protobuf marshalling interface and cannot be encoded to a *client.applyconfigurationRuntimeObject protobuf message  `

## Steps to Reproduce
To reproduce the issue, you can add `cfg.ContentType = "application/vnd.kubernetes.protobuf"` in the client_suite_test.go file and execute the tests. This triggers the error during the deserialization stage.

Please let us know if we are doing anything wrong or this is indeed a bug. Thank you!

<img width="1644" height="524" alt="Image" src="https://github.com/user-attachments/assets/c8f0f329-faa9-4900-8efd-fa53c5f7a1fc" />
<img width="1638" height="530" alt="Image" src="https://github.com/user-attachments/assets/ae9f1c22-9356-4568-be88-d9e62c1b021f" />
<img width="1658" height="458" alt="Image" src="https://github.com/user-attachments/assets/cd2a66e5-f0bf-46d8-b00d-a46b753f686a" />

alvaroaleman:
is protobuf supported for SSA? To my knowledge, only yaml and cbor are: https://github.com/kubernetes/kubernetes/blob/ddc2c5d192275207b7ebb43b066fd0b0e94fc011/staging/src/k8s.io/client-go/util/apply/apply.go#L35

Why are you setting the content-type explicitly in the config?

filipcirtog:
I explicitly set the Content-Type in the test to ensure the error can be consistently reproduced. While we don't explicitly set the Content-Type in our operator's code, we've observed situations in the controller code where the API server sometimes returns protobuf responses, especially when apply is creating a new resource.

s-urbaniak:
Thank you @alvaroaleman! Indeed we oversaw the fact that SSA doesn't support protobuf, and you're right, `NewRequest` enforces the YAML content-type so we really need to find out why we in our case a `application/vnd.kubernetes.protobuf` content-type is being constructed for the Apply, this is indeed strange.

[...]

PR #3346, opened 2025-10-10
🐛 Allow SSA after normal resource creation

filipcirtog:
Addressing issue #3344 

Issue:
- When using Server-Side Apply (SSA) users could encounter the following error: `object %v does not implement the protobuf marshalling interface and cannot be encoded to a *client.applyconfigurationRuntimeObject protobuf message.  `

Root Cause:
- By default, the content-type is set to "protobuf" when creating in-tree resources. If resources were previously created without using Server-Side Apply (SSA), attempting an SSA operation later caused the GroupVersionKind (GVK) to be identified, halting execution before a new client with the forceDisableProtoBuf option could be created. This resulted in unintended generation of Protobuf requests, which SSA does not support, causing the previous error.

Changes Made:
- The implementation now includes a cache key that combines both the actual GVK (GroupVersionKind) and the forceDisableProtoBuf flag. This ensures that we now support up to two distinct clients for each GVK: one with forceDisableProtoBuf enabled and another without it.

### DIFF
--- a/pkg/client/client_rest_resources.go
+++ b/pkg/client/client_rest_resources.go
@@ -48,11 +48,8 @@ type clientRestResources struct {
 	// codecs are used to create a REST client for a gvk
 	codecs serializer.CodecFactory
 
-	// structuredResourceByType stores structured type metadata
-	structuredResourceByType map[cacheKey]*resourceMeta
-
-	// unstructuredResourceByType stores unstructured type metadata
-	unstructuredResourceByType map[cacheKey]*resourceMeta
+	// resourceByType stores type metadata
+	resourceByType map[cacheKey]*resourceMeta
 
 	mu sync.RWMutex
 }
@@ -127,12 +124,7 @@ func (c *clientRestResources) getResource(obj any) (*resourceMeta, error) {
 
 	cacheKey := cacheKey{gvk: gvk, forceDisableProtoBuf: forceDisableProtoBuf}
 
-	resourceByType := c.structuredResourceByType
-	if isUnstructured {
-		resourceByType = c.unstructuredResourceByType
-	}
-
-	r, known := resourceByType[cacheKey]
+	r, known := c.resourceByType[cacheKey]
 
 	c.mu.RUnlock()
 
@@ -152,7 +144,7 @@ func (c *clientRestResources) getResource(obj any) (*resourceMeta, error) {
 	if err != nil {
 		return nil, err
 	}
-	resourceByType[cacheKey] = r
+	c.resourceByType[cacheKey] = r
 	return r, err
 }
 
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -962,10 +962,6 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 					"some-key": []byte("some-value"),
 				}
 				secretObject := &corev1.Secret{
-					TypeMeta: metav1.TypeMeta{
-						Kind:       "Secret",
-						APIVersion: "v1",
-					},
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "secret-one",
 						Namespace: "default",
@@ -975,8 +971,6 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 
 				secretApplyConfiguration := corev1applyconfigurations.
 					Secret("secret-two", "default").
-					WithAPIVersion("v1").
-					WithKind("Secret").
 					WithData(data)
 
 				err = cl.Create(ctx, secretObject)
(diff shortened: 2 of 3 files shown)


pkg/client/client_rest_resources.go
```diff
@@ -124,7 +124,7 @@ func (c *clientRestResources) getResource(obj any) (*resourceMeta, error) {
 	r, known := resourceByType[gvk]
 	c.mu.RUnlock()
 
-	if known {
+	if known && !forceDisableProtoBuf {
```
alvaroaleman: This means that we will create a new client for every SSA requests. Can we avoid that?

filipcirtog: I've noticed the same thing and am currently looking into it.

filipcirtog: @alvaroaleman we could also add this to the Apply req. `SetHeader("Accept", "application/json").`WDYT?

alvaroaleman: That will override any user-configured preference and is something that needs maintenace. Upstream is for example working on cbor which is more efficient than json. Can we instead just key the cached clients by both known and forceDisableProtobuf?

filipcirtog: Yes, I agree that this is the best solution. While forceDisableProtoBuf is always true for unstructuredResource, I also made the changes for consistency. Please let me know if you have any additional feedback or suggestions to improve this further.

pkg/client/client_rest_resources.go
```diff
@@ -49,10 +49,17 @@ type clientRestResources struct {
 	codecs serializer.CodecFactory
 
 	// structuredResourceByType stores structured type metadata
-	structuredResourceByType map[schema.GroupVersionKind]*resourceMeta
+	structuredResourceByType map[cacheKey]*resourceMeta
+
 	// unstructuredResourceByType stores unstructured type metadata
-	unstructuredResourceByType map[schema.GroupVersionKind]*resourceMeta
-	mu                         sync.RWMutex
+	unstructuredResourceByType map[cacheKey]*resourceMeta
```
alvaroaleman: From what I can tell the only reason for having these two was that unstructured had the same problem of never being able to use proto - Would you mind to de-duplicate them into a single `resourceByType`?

pkg/client/client_test.go
```diff
@@ -953,6 +953,58 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(cl).NotTo(BeNil())
+				data := map[string][]byte{
+					"some-key": []byte("some-value"),
+				}
+				secretObject := &corev1.Secret{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       "Secret",
+						APIVersion: "v1",
+					},
```
alvaroaleman: This isn't needed and can be removed
```suggestion
```

pkg/client/client_test.go
```diff
@@ -953,6 +953,58 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+
+				data = map[string][]byte{
+					"some-key": []byte("some-new-value"),
+				}
+				secretApplyConfiguration.Data = data
+
+				err = cl.Apply(ctx, secretApplyConfiguration, &client.ApplyOptions{FieldManager: "test-manager"})
+				Expect(err).NotTo(HaveOccurred())
+
+				cm, err = clientset.CoreV1().Secrets(ptr.Deref(secretApplyConfiguration.GetNamespace(), "")).Get(ctx, ptr.Deref(secretApplyConfiguration.GetName(), ""), metav1.GetOptions{})
```
alvaroaleman: Nit
```suggestion
				secret, err = clientset.CoreV1().Secrets(ptr.Deref(secretApplyConfiguration.GetNamespace(), "")).Get(ctx, ptr.Deref(secretApplyConfiguration.GetName(), ""), metav1.GetOptions{})
```

pkg/client/client_rest_resources.go
filipcirtog: That's right. I have addressed this. Thank you!

sbueringer:
/lgtm

> This change introduces the possibility of creating multiple clients, which might allow simultaneous requests from different entities. This could lead to concurrency considerations.

How much of a problem do we think this is? (I don't really see the problem)

filipcirtog:
@sbueringer It was this comment that triggered me to have that conclusion.
`// It's better to do creation work twice than to not let multiple people make requests at once`

However, thanks to Alvaro's feedback, I've updated the implementation to include a cache key that combines both the actual GVK (GroupVersionKind) and the forceDisableProtoBuf flag. This ensures that we now support up to two distinct clients for each GVK: one with forceDisableProtoBuf enabled and another without it. This approach is an improvement over my initial PR, which created a new client for forceDisableProtoBuf each time.

sbueringer:
Got it, thx for the quick response. Let's please update the PR description

filipcirtog:
Done. Thank you. Please let me know if there is anything more to change.

sbueringer:
All good, thank you!

PR #3350, opened 2025-10-12
:seedling: Priorityqueue tests: Use synctest

alvaroaleman:
This change moves all tests in the priorityqueue where it makes sense to use [synctest](https://go.dev/blog/synctest). Synctest provides a `Wait` func we can use to wait until all goroutines in the test are durably blocked. This allows us to block after an `Add` call until the priorityqueue has finished processing. Because we didn't have that, we previously used `time.Sleep` and `Eventually` if we wanted something to happen and `Consistently` if we wanted to assert that something does not happen. With this change, we don't need any of those three anymore.

Because [ginkgo doesn't support synctest](https://github.com/onsi/ginkgo/issues/1601), this change entails that we have to stop using it. To simplify reviewing, I've moved the tests from ginkgo to go test in the first commit and updated them to use synctest in the second commit.


alvaroaleman:
/hold
Needs a rebase once https://github.com/kubernetes-sigs/controller-runtime/pull/3349 merges to pick up the newer apidiff and go versions

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

Backporting tests won't be much fun for a bit :)

Issue #3347, opened 2025-10-10
Namespaced Client unable to List non-namespaced objects


fishmoose:
In the `List` function, the client's namespace is appended to the ListOptions regardless of the scope of the objects:
https://github.com/kubernetes-sigs/controller-runtime/blob/c8a5a35177d3efc56918023f5253ffd57dc943ad/pkg/client/namespaced_client.go#L224-L226

This makes it unusable with objects that have a cluster-wide scope, as it will never find anything.

It is also inconsistent with the `Get` implementation, where there is a check if the object is namepaced or not, before setting the namespace:
https://github.com/kubernetes-sigs/controller-runtime/blob/c8a5a35177d3efc56918023f5253ffd57dc943ad/pkg/client/namespaced_client.go#L209-L218

sbueringer:
Sounds like a bug

PR #3351, opened 2025-10-13
🐛 Update List in namespaced client to list objects that are cluster scoped

troy0820:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3347

### DIFF
--- a/pkg/client/namespaced_client_test.go
+++ b/pkg/client/namespaced_client_test.go
@@ -54,6 +54,8 @@ var _ = Describe("NamespacedClient", func() {
 
 		err := rbacv1.AddToScheme(sch)
 		Expect(err).ToNot(HaveOccurred())
+		err = corev1.AddToScheme(sch)
+		Expect(err).ToNot(HaveOccurred())
 		err = appsv1.AddToScheme(sch)
 		Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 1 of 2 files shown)


troy0820:
_I can add a test for this_

pkg/client/namespaced_client_test.go
```diff
@@ -147,6 +149,13 @@ var _ = Describe("NamespacedClient", func() {
 			Expect(result.Items[0]).To(BeEquivalentTo(*dep))
 		})
 
+		It("should successfully List objects when object is not namespaced scoped", func(ctx SpecContext) {
+			result := &corev1.NodeList{}
+			opts := &client.ListOptions{}
+			Expect(getClient().List(ctx, result, opts)).NotTo(HaveOccurred())
+			Expect(result.Items).NotTo(BeEmpty())
```
alvaroaleman, line 156: Will this work? We are not creating a node in the beforeEach or did I miss it?

troy0820: We are not creating a node but a node exist? I am making the assumption but when tested it showed that the node exists when querying the result.

troy0820: I had to change it from `len(result.Items)` to `EXPECT(result.Items)` to not be Empty when that gave me `9`.  The test passes showing that the nodelist is not empty.

sbueringer: Hm. This seems to fail on my Machine somehow (at least when run with Intellij)? Which Nodes do you get?

alvaroaleman: Is it possible the node object(s) get created by other tests running in parallel using the same client? That would both explain the flaking and the inability to use `len`.

Sorry, I missed this before approving

sbueringer: Probably. We could also simply make this case self-contained by creating a Node with some `namespaced` prefix

troy0820: The flakiness of this is unfortunate because I made the assumption that because the node is there, it is always there.  We can contain it by creating a node in the Before Spec if that helps with the testing suite.

troy0820: https://github.com/kubernetes-sigs/controller-runtime/pull/3353 <-- Follow up for updating the test.

PR #3356, opened 2025-10-13
:seedling: Deflake should execute the Warmup function when Warmup group is started

alvaroaleman:
The test is flaky as visible in both PRs and the periodic. The reason is that it calls Start() which asynchronously starts a runnable and expects said runnable to finish immediately after which may or may not work out depending on how busy the box the test is run on is.

Use `synctest` instead, as that allows us to explicitly block until all asynchronous operations that can finish are finished.


sbueringer:
Nice, thx!

/lgtm
/approve

Issue #3358, opened 2025-10-14
setup-envtest@latest requires golang >=1.25.0?


stevenhorsman:
In the latest release of `setup-envtest@latest` from this morning our pipelines (which are running on Go 1.24.7) have started reporting failures with installing `setup-envtest@latest`:

```
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20251014082336-b8f11375258f
go: downloading sigs.k8s.io/controller-runtime v0.22.3
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20251014082336-b8f11375258f requires go >= 1.25.0 (running go 1.24.7; GOTOOLCHAIN=local)
```

The docs: https://github.com/kubernetes-sigs/controller-runtime/blob/b8f11375258f85190ec46d20726d96c2e74d95a1/tools/setup-envtest/README.md?plain=1#L7-L12 still seem to suggest that the latest will work with 1.24, so is the documentation wrong and we need to pin the version we are using, or was the dependency bump unintentional?

Thanks!

sbueringer:
We should probably drop the documentation that tells folks to use setup-envtest from main. This just sounds like a very bad idea as we can't guarantee that main doesn't change.

We bumped the main branch of controller-runtime intentionally to Go 1.25 as we bumped to the first Kubernetes 1.35 alpha.

For stability I would recommend to either pin to a release branch or consume the binaries that we publish as part of controller-runtime releases.

stevenhorsman:
Thanks for the response. Which branch would you recommend? - Will `release-0.22` have the 1.25 change merged into it, or is that safe to stick with?

sbueringer:
I would use release-0.22. We don't bump k8s.io/* library and Go minor versions on release branches

stevenhorsman:
Great - thanks for the approach, feel free to close this issue, or leave as a place-holder to update the doc's guidance, whichever is most helpful to the project.

sbueringer:
I'll leave it open, it makes sense to update the doc

PR #3359, opened 2025-10-16
📖Remove latest from setupenvtest docs

troy0820:
Instead of saying use `latest` which can have a different go version in the go.mod at the time of reading the doc, we can just use the last release.  

If people want to install at latest, they will do so at their own will.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3358

/assign @sbueringer 
/assign @alvaroaleman 

_Not sure if this is sufficient but it does document this a little better than before._

### DIFF
--- a/tools/setup-envtest/README.md
+++ b/tools/setup-envtest/README.md
@@ -4,11 +4,14 @@ This is a small tool that manages binaries for envtest. It can be used to
 download new binaries, list currently installed and available ones, and
 clean up versions.
 
-To use it, just go-install it with Golang 1.24+ (it's a separate, self-contained
-module):
+To use it, download the binary from the [release page.](https://github.com/kubernetes-sigs/controller-runtime/releases)
+
+If you want to install this with Golang, you can install a release by using a release branch instead.
+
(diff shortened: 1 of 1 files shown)


tools/setup-envtest/README.md
```diff
@@ -8,7 +8,7 @@ To use it, just go-install it with Golang 1.24+ (it's a separate, self-contained
 module):
 
 ```shell
-go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.22
```
alvaroaleman, line 14: Maybe it would be better to instead tell ppl to download it from a github release to avoid keeping this up to date? We will most likely forget to do that

sbueringer: Let's point to the GitHub release first and then mention this with release-major.minor or some other placeholder

sbueringer:
/lgtm

I'm fine with mentioning a release branch. I would see this only as an example that we don't have to keep up to date

Issue #2675, opened 2024-02-07
Proposal: Generics for CustomValidator & CustomDefaulter


ahmetb:
[CustomValidator](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.0/pkg/webhook/admission#CustomValidator) and [CustomDefaulter](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.0/pkg/webhook/admissionCustomDefaulter) interfaces could use generics support to potentially shorten the handler funcs like this:

```go
func (w *KubernetesPoolWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
	oldO, ok := oldObj.(apiv1alpha1.KubernetesPool)
	if err != nil {
		return nil, errors.NewBadRequest(fmt.Sprintf("unexpected type %T", oldObj))
	}
	newO, ok := newObj.(apiv1alpha1.KubernetesPool)
	if err != nil {
		return nil,  errors.NewBadRequest(fmt.Sprintf("unexpected type %T", newObj))
	}
	return nil, validateUpdate(oldO, newO)
}
```
to

```go
func (w *KubernetesPoolWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj apiv1alpha1.KubernetesPool) (admission.Warnings, error) {
	return nil, validateUpdate(oldO, newO)
}
```

Also statements like:
```
return ctrl.NewWebhookManagedBy(mgr).
	For(&apiv1alpha1.KubernetesPool{}).
	WithDefaulter(w).
	WithValidator(w).
	Complete()
```

can probably be inferred with more type safety (right now there's nothing ensures a & b are handlers of the same type in `WithDefaulter(a).WithDefaulter(b)`?). e.g.

```
return ctrl.NewWebhookManagedBy[apiv1alpha1.KubernetesPool](mgr).
	WithDefaulter(w).
	WithValidator(w).
	Complete()
```

/kind proposal
/kind feature

alvaroaleman:
@ahmetb are you sure this is possible? My understanding is that it isn't due to https://github.com/golang/go/issues/49085, the `WebhookBuilder` would only be able to support one type. I'd love to be wrong on this though.

ahmetb:
I will give it a shot. I hit that in porting go-linq to generics as well.

alvaroaleman:
/reopen
/lifecycle frozen

[...]

PR #3360, opened 2025-10-20
:warning: Generic Validator and Defaulter

alvaroaleman:
This change adds the `Validator` and `Defaulter` interfaces which are generic versions of `CustomValidator` and `CustomDefaulter`. The existing `CusotmDefaulter` and `CustomValidator` are being deprecated.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2675
/hold

This is missing tests, I will add them once we agree on the direction in https://github.com/kubernetes-sigs/controller-runtime/pull/3360#discussion_r2446377647

### DIFF
--- a/pkg/builder/webhook.go
+++ b/pkg/builder/webhook.go
@@ -64,7 +64,7 @@ func WebhookManagedBy[T runtime.Object](m manager.Manager, object T) *WebhookBui
 
 // WithCustomDefaulter takes an admission.CustomDefaulter interface, a MutatingWebhook with the provided opts (admission.DefaulterOption)
 // will be wired for this type.
-// Deprecated: Use WithAdmissionDefaulter instead.
+// Deprecated: Use WithDefaulter instead.
 func (blder *WebhookBuilder[T]) WithCustomDefaulter(defaulter admission.CustomDefaulter, opts ...admission.DefaulterOption) *WebhookBuilder[T] {
 	blder.customDefaulter = defaulter
 	blder.customDefaulterOpts = opts
@@ -79,7 +79,7 @@ func (blder *WebhookBuilder[T]) WithDefaulter(defaulter admission.Defaulter[T], 
 }
 
 // WithCustomValidator takes a admission.CustomValidator interface, a ValidatingWebhook will be wired for this type.
-// Deprecated: Use WithAdmissionValidator instead.
+// Deprecated: Use WithValidator instead.
 func (blder *WebhookBuilder[T]) WithCustomValidator(validator admission.CustomValidator) *WebhookBuilder[T] {
 	blder.customValidator = validator
 	return blder
--- a/pkg/builder/webhook_test.go
+++ b/pkg/builder/webhook_test.go
@@ -1236,7 +1236,7 @@ func (*testValidator) ValidateUpdate(ctx context.Context, oldObj, newObj *TestVa
 		return nil, errors.New("number of replica should be greater than or equal to 0")
 	}
 	if newObj.Replica < oldObj.Replica {
-		return nil, fmt.Errorf("new replica %v should not be fewer than old replica %v", oldObj.Replica, newObj.Replica)
+		return nil, fmt.Errorf("new replica %v should not be fewer than old replica %v", newObj.Replica, oldObj.Replica)
 	}
 
 	userAgent, ok := ctx.Value(userAgentCtxKey).(string)


pkg/builder/webhook.go
```diff
@@ -56,59 +58,61 @@ type WebhookBuilder struct {
...
-// For takes a runtime.Object which should be a CR.
-// If the given object implements the admission.Defaulter interface, a MutatingWebhook will be wired for this type.
-// If the given object implements the admission.Validator interface, a ValidatingWebhook will be wired for this type.
-func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
-	if blder.apiType != nil {
-		blder.err = errors.New("For(...) should only be called once, could not assign multiple objects for webhook registration")
-	}
-	blder.apiType = apiType
-	return blder
+func WebhookManagedBy[T runtime.Object](m manager.Manager, object T) *WebhookBuilder[T] {
```
alvaroaleman, line 61: So I've spent a fair bit of time on pondering about how to best deal with this in the webhook builder:
1. Initially, I didn't have the `object T` func arg and used the existing `WithValidator` and `WithDefaulter` methods. That works but requires everyone to explicitly type `WebhookManagedBy` as the typing must be know during initial construction and can not be inferred during subsequent method calls. This has two drawbacks IMHO:
    * I anticipate that to cause confusion, because my impression has been that many go engineers are not very used to explicitly typing generics as opposed to relying on type inference
    * Anyone who has an existing `CustomValidator/Defaulter` would have to type this to `runtime.Object` which would likely cause further confusion
1. So then I added a `WebhookFor` that has the same signature as the current `WebhookMangagedBy` and made the `WebhookManagedBy` non-generic and return a `*WebhookBuilder[runtime.Object]`. This is great for existing code as it will all keep working, but once we remove this, it will be confusing to have different names for the controller and webhook builder IMHO
1. Then I finally ended up with this, make `WebhookManagedBy` generic, add an explicit type argument so type inference works and add successors to `WithValidator/Defaulter` in the form of `WithAdmissionValidator/Defaulter`. This means a breaking change for everyone that should be pretty easy to understand and fix and avoid requiring to type this to `runtime.Object` for existing validators/defaulters. The main drawback of that is that the new names aren't as nice (happy to hear suggestions for that).

All in all, the last option seemed the by far least bad one. What do you think?

sbueringer: Agree with all your points.

> The main drawback of that is that the new names aren't as nice (happy to hear suggestions for that).

I see the following options

1. Go with your current names, keep them like that going forward eventually drop WithValidator/Defaulter
2.  1 + but eventually do another round of deprecation to go from WithAdmissionValidator/Defaulter to WithValidator/Defaulter
3. Directly go to WithValidator/Defaulter, temporarily introduce deprecated WithCustomValidator/Defaulter

No absolutely strong opinions from my side, but if possible I would like to get to WithValidator/Defaulter long-term. I have a slight tendency for option 3. We already have to do a breaking change in this PR, maybe it's better to just get it over with and do slightly more breaking changes now then dragging this out over a few years. It will also give a clear hint to folks that they should just migrate to the typed versions which is super straightforward then (just start using types in Validator/Defaulter, it's not even necessary to use different methods on the builder for Validator/Defaulter). So slightly less effort to do the right migration (use types), slightly more effort to delay the migration and keep using CustomValidator/Defaulter.

Somewhat related. Do you know why `Defaulter` is spelled with `er` and `Validator` with `or`? (probably don't want to change that though because it's not worth the additional confusion)

alvaroaleman: Okay, if we say that we want to end up with `WithValidator/Defaulter` then option 3 makes most sense, otherwise its just a back and forth. I guess we can add a codesnippet before/after to the changelog to make that at least easy

> Somewhat related. Do you know why Defaulter is spelled with er and Validator with or? (probably don't want to change that though because it's not worth the additional confusion)

That is just the English language, one has an `er` suffix and the other `or`

sbueringer: > Okay, if we say that we want to end up with WithValidator/Defaulter then option 3 makes most sense, otherwise its just a back and forth. I guess we can add a codesnippet before/after to the changelog to make that at least easy

Agree, sounds good

pkg/webhook/admission/validator_custom.go
```diff
@@ -30,54 +31,77 @@ import (
...
 	}
 }
 
-type validatorForType struct {
-	validator CustomValidator
-	object    runtime.Object
+type validatorForType[T runtime.Object] struct {
+	validator Validator[T]
 	decoder   Decoder
+	new       func() runtime.Object
```
sbueringer:
```suggestion
	new       func() T
```
I think it doesn't matter, but should this return T? (just noticed the diff to defaulter)

pkg/webhook/alias.go
```diff
@@ -23,12 +23,6 @@ import (
 
 // define some aliases for common bits of the webhook functionality
 
-// CustomDefaulter defines functions for setting defaults on resources.
```
sbueringer, line 26: Should we keep these aliases and deprecate them as well?

(Just to make this slightly less breaking as it's easily possible)

pkg/builder/webhook.go
```diff
@@ -56,59 +58,61 @@ type WebhookBuilder struct {
...
-// If the given object implements the admission.Defaulter interface, a MutatingWebhook will be wired for this type.
-// If the given object implements the admission.Validator interface, a ValidatingWebhook will be wired for this type.
-func (blder *WebhookBuilder) For(apiType runtime.Object) *WebhookBuilder {
-	if blder.apiType != nil {
-		blder.err = errors.New("For(...) should only be called once, could not assign multiple objects for webhook registration")
-	}
-	blder.apiType = apiType
+// WithCustomDefaulter takes an admission.CustomDefaulter interface, a MutatingWebhook with the provided opts (admission.DefaulterOption)
+// will be wired for this type.
+// Deprecated: Use WithAdmissionDefaulter instead.
```
sbueringer:
```suggestion
// Deprecated: Use WithDefaulter instead.
```
similar in l.82

pkg/builder/webhook.go
```diff
@@ -244,20 +251,28 @@ func (blder *WebhookBuilder) registerDefaultingWebhook() error {
...
 }
 
-func (blder *WebhookBuilder) getDefaultingWebhook() *admission.Webhook {
-	if defaulter := blder.customDefaulter; defaulter != nil {
-		w := admission.WithCustomDefaulter(blder.mgr.GetScheme(), blder.apiType, defaulter, blder.customDefaulterOpts...)
-		if blder.recoverPanic != nil {
-			w = w.WithRecoverPanic(*blder.recoverPanic)
+func (blder *WebhookBuilder[T]) getDefaultingWebhook() (*admission.Webhook, error) {
+	var w *admission.Webhook
+	if defaulter := blder.defaulter; defaulter != nil {
```
sbueringer:
```suggestion
	if blder.defaulter != nil {
```
nit. Similar for validator. I would also cleanup the var in the else if. Just unnecessary noise :)

pkg/builder/webhook_test.go
```diff
@@ -1097,13 +1232,11 @@ func (*TestCustomValidator) ValidateUpdate(ctx context.Context, oldObj, newObj r
...
-	v := newObj.(*TestValidator)
-	old := oldObj.(*TestValidator)
-	if v.Replica < 0 {
+	if newObj.Replica < 0 {
 		return nil, errors.New("number of replica should be greater than or equal to 0")
 	}
-	if v.Replica < old.Replica {
-		return nil, fmt.Errorf("new replica %v should not be fewer than old replica %v", v.Replica, old.Replica)
+	if newObj.Replica < oldObj.Replica {
+		return nil, fmt.Errorf("new replica %v should not be fewer than old replica %v", oldObj.Replica, newObj.Replica)
```
sbueringer: not that it matters but old and new got flipped around here

pkg/builder/webhook_test.go
```diff
@@ -1214,3 +1346,23 @@ func (*TestCustomDefaultValidator) ValidateDelete(ctx context.Context, obj runti
 }
 
 var _ admission.CustomValidator = &TestCustomValidator{}
+
+type testValidatorDefaulter struct{}
```
sbueringer, line 1350: Do we intentionally have two structs here that both are using TestCustomDefaultValidator internally?

pkg/webhook/admission/validator_custom.go
alvaroaleman: Fixed

pkg/builder/webhook.go
alvaroaleman: Re-added them, I am not really sure what they are useful for as they are not constructors but just interfaces but there is no harm in having them here.

pkg/builder/webhook.go
alvaroaleman: Done

pkg/builder/webhook_test.go
alvaroaleman: Fixed

pkg/builder/webhook_test.go
alvaroaleman: Yeah, because one of them is a `Validator` and the other a `CustomValidator` but I wanted to avoid implementing this twice, hence one is just a shim over the other.

The tests here also define their own `runtime.Object` which I guess is from back when these where methods on that. There isn't a good reason to do that at this point vs just using any built-in type, but the diff of this PR is already quite large so I will probably clean this up in a follow-up.

pkg/builder/webhook.go
sbueringer: (wrong thread) Yeah I also don't know. Just to avoid breaking folks that use these aliases for some reason.
Doesn't matter long-term anyway as we'll drop them

pkg/builder/webhook_test.go
sbueringer: Sorry I was a bit unclear.

I meant do we need testValidatorDefaulter and testDefaultValidatorValidator
(First one implements only Default, second one Validate*)

The name of the second one is also a bit strange


In any case, fine to clean up in a follow-up

sbueringer:
Thank you very much!!!

/lgtm
/approve
/hold cancel

PR #3367, opened 2025-10-28
🐛 fix(testing/addr): prevent possible leak by avoiding defer in loop

s-z-z:
1. prevent possible leak by removing defer in loop
2. update description as `portReserveTime` constant


sbueringer:
Thx!

/ok-to-test
/lgtm
/approve

Issue #3363, opened 2025-10-23
Priority queue: priority not always respected


fossedihelm:
I suspect that there may be conditions under which the item returned by `GetWithPriority()` is not always the correct/expected element.
Both `GetWithPriority()`  and  `AddWithOpts` trigger a `spin()`: https://github.com/kubernetes-sigs/controller-runtime/blob/fb2beabd029b3d083b453ded8446cd43366c0db9/pkg/controller/priorityqueue/priorityqueue.go#L206

In the `GetWithPriority()`, the `w.waiters` are raised `w.waiters.Add(1)`, and in the `spin` items are skipped if none is waiting:
```golang
				if w.waiters.Load() == 0 {
					// Have to keep iterating here to ensure we update metrics
					// for further items that became ready and set nextReady.
					return true
				}
```

If a `GetWithPriority()` is executed right after an `AddWithOpts`, there could be the possibility that the `Ascend` of the Btree is not completed yet, causing not all the items of the BTree to evaluate the same `w.waiters.Load()` value.

This will result in the non-obtaining of the higher-priority element.

In particular, you can see that this test https://github.com/kubevirt/kubevirt/blob/66f09757e8707ae4a09c660119ec1ab3658b0cdb/pkg/virt-controller/watch/migration/migration_test.go#L2342 is flaky (a failure example: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/15887/pull-kubevirt-unit-test/1981424912933326848).

A possible solution could be adding a lock before increasing the waiters:
https://github.com/kubernetes-sigs/controller-runtime/blob/fb2beabd029b3d083b453ded8446cd43366c0db9/pkg/controller/priorityqueue/priorityqueue.go#L313-L319

might be:
```golang
 func (w *priorityqueue[T]) GetWithPriority() (_ T, priority int, shutdown bool) { 
 	if w.shutdown.Load() { 
 		var zero T 
 		return zero, 0, true 
 	} 
    w.lock.Lock()
 	w.waiters.Add(1) 
    w.lock.Unlock()
```

What do you think?
Thank you!

Edit: Checked the above solution. The flaky disappeared, but IDK the implications

alvaroaleman:
Acquiring the lock in `GetWithPriority` is fine I think as it is blocking anyways. No need to use an atomic then anymore. It would be good to have a test for this, as this is the kind of thing that is easy to regress in the future

PR #3368, opened 2025-10-28
:bug: priority queue: properly sync the `waiter` manipulation

fossedihelm:
As described in https://github.com/kubernetes-sigs/controller-runtime/issues/3363, there are some circumstances under which `GetWithPriority` does not return the correct/expected element.

This can happen when a `GetWithPriority` is executed and the `Ascend` of the queue is not completed yet, causing not all the items of the BTree to evaluate the same `w.waiters.Load()` value.

Adding a lock to manipulate the waiters will solve the issue. Since the lock is required, there is no need to use an atomic.Int64 anymore.

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/3363

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -124,8 +124,8 @@ type priorityqueue[T comparable] struct {
 	get chan item[T]
 
 	// waiters is the number of routines blocked in Get, we use it to determine
-	// if we can push items.
-	waiters atomic.Int64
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/ok-to-test

Please sign the CLA so we can accept your contribution

fossedihelm:
@alvaroaleman Signed :) Thanks

alvaroaleman (approved):
thanks!

sbueringer:
Thx!

/lgtm

PR #3371, opened 2025-10-29
[release-0.22] :bug: priority queue: properly sync the waiter manipulation 

alvaroaleman:
As described in https://github.com/kubernetes-sigs/controller-runtime/issues/3363,
there are some circumstances under which `GetWithPriority`
is not returning the correct/expected element.

This can happen when a `GetWithPriority` is executed
and the `Ascend` of the queue is not completed yet,
causing not all the items of the BTree to evaluate
the same w.waiters.Load() value.

Adding a lock to manipulate the waiters will solve the issue.
Since the lock is required, there is no need to use an
atomic.Int64 anymore.

Cherry-pick of https://github.com/kubernetes-sigs/controller-runtime/pull/3368

This cherry-pick does not include the corresponding test, because it uses synctest which is only available in go 1.25 and we use go 1.24 here. Given that all changes have to pass CI and be merged into `main` first, I think this is acceptable.

/assign @sbueringer

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -124,8 +124,8 @@ type priorityqueue[T comparable] struct {
 	get chan item[T]
 
 	// waiters is the number of routines blocked in Get, we use it to determine
-	// if we can push items.
-	waiters atomic.Int64
+	// if we can push items. Every manipulation has to be protected with the lock.
+	waiters int64
 
 	// Configurable for testing
 	now  func() time.Time
@@ -269,7 +269,7 @@ func (w *priorityqueue[T]) spin() {
 						}
 					}
 
-					if w.waiters.Load() == 0 {
+					if w.waiters == 0 {
 						// Have to keep iterating here to ensure we update metrics
 						// for further items that became ready and set nextReady.
 						return true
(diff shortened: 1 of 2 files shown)


fossedihelm:
@alvaroaleman Can we backport this up to 0.20? And maybe ask for a 20.z release? Thank you!
Our k8s dependencies are still old :)

alvaroaleman:
> @alvaroaleman Can we backport this up to 0.20? And maybe ask for a 20.z release? Thank you!
Our k8s dependencies are still old :)

This patch doesn't cleanly apply there and there are a number of fixes that are not in that version like for example https://github.com/kubernetes-sigs/controller-runtime/pull/3338 or https://github.com/kubernetes-sigs/controller-runtime/pull/3340. What keeps you from upgrading your k8s deps version?

sbueringer:
I'm also hesitant to backport it that far at this point. You might be able to copy the current PQ implementation, create a PQ instance yourself and then pass it into the manager.

fossedihelm:
> > @alvaroaleman Can we backport this up to 0.20? And maybe ask for a 20.z release? Thank you!
> > Our k8s dependencies are still old :)
> 
> This patch doesn't cleanly apply there and there are there fixes that are not in that version like for example #3338 or #3340. What keeps you from upgrading your k8s deps version?

@alvaroaleman Actually is a mix or k8s dependencies / minimum go version and stable branches.
Our release-1.6 is using go 1.23.
Of course, I can provide the manual cherry pick for both 0.21/0.20.
Thanks in advance

PR #3372, opened 2025-10-30
✨ envtest: respect pre-configured binary paths in ControlPlane

mzhaom:
This commit fixes an issue where Environment.Start() would ignore pre-configured binary paths (APIServer.Path, Etcd.Path, KubectlPath) set in ControlPlane when DownloadBinaryAssets is false.

Changes:
- Extract path configuration logic into configureBinaryPaths() method for better testability and separation of concerns
- Only auto-configure binary paths when they are empty (not pre-set)
- When DownloadBinaryAssets is true, downloaded paths are still used (preserving existing behavior)
- Update ControlPlane field documentation to clarify path behavior
- Add tests in envtest_test.go to verify path handling logic

This fixed  issue #3373

### DIFF
--- a/pkg/envtest/envtest_test.go
+++ b/pkg/envtest/envtest_test.go
@@ -963,4 +963,56 @@ var _ = Describe("Test", func() {
 			Expect(env.WebhookInstallOptions.LocalServingCertDir).ShouldNot(BeADirectory())
 		})
 	})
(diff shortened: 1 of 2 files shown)


sbueringer:
Thx!

/lgtm
/approve

PR #3375, opened 2025-11-01
✨ Add CreateOrPatch function in alias.go

tisonkun:
This is the same as alias to CreateOrUpdate in alias.go.

... for the sake of symmetry.


sbueringer:
Makes sense given we already have it for CreateOrUpdate. In general, I'm not sure if all this aliasing is really a good idea.

/ok-to-test
/lgtm
/approve

PR #3385, opened 2025-11-13
✨ Make webhook setup conditional based on user provided config

pmalek:
Make webhook configuration apply only if it's specified by the user through:

- `.WebhookInstallOptions.ValidatingWebhooks`
- `.WebhookInstallOptions.MutatingWebhooks`
- either of provided CRDs having `.Spec.Conversion.Webhook` set

Without this change, environments with CRDs that have multiple versions receive the following errors:

```
Create error: conversion webhook for apigroup.com/v1beta1, Kind=MyCustomKind failed: Post "https://127.0.0.1:59762/convert?timeout=30s": dial tcp 127.0.0.1:59762: connect: connection refused
```

x-ref: https://kubernetes.slack.com/archives/C02MRBMN00Z/p1763049500384929

### DIFF
--- a/pkg/envtest/server.go
+++ b/pkg/envtest/server.go
@@ -98,13 +98,11 @@ type (
 	Arg = process.Arg
 )
 
-var (
-	// EmptyArguments constructs a new set of flags with nothing set.
-	//
-	// This is mostly useful for testing helper methods -- you'll want to call
-	// Configure on the APIServer (or etcd) to configure their arguments.
-	EmptyArguments = process.EmptyArguments
-)
+// EmptyArguments constructs a new set of flags with nothing set.
+//
+// This is mostly useful for testing helper methods -- you'll want to call
+// Configure on the APIServer (or etcd) to configure their arguments.
+var EmptyArguments = process.EmptyArguments
 
 // Environment creates a Kubernetes test environment that will start / stop the Kubernetes control plane and
 // install extension APIs.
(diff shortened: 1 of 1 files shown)


sbueringer:
Am I understanding correctly that the use case is to have CRDs with multiple apiVersions without conversion? How does this work?

pmalek:
> Am I understanding correctly that the use case is to have CRDs with multiple apiVersions without conversion? How does this work?

For my particular use case I don't need the conversion as I'm only testing the CRD CEL expressions and not controller's behavior.

After looking at the test code in controller-runtime I came up with a stub conversion webhook server:

```
	ws := webhook.NewServer(webhook.Options{
		Port:    testEnv.WebhookInstallOptions.LocalServingPort,
		Host:    testEnv.WebhookInstallOptions.LocalServingHost,
		CertDir: testEnv.WebhookInstallOptions.LocalServingCertDir,
	})
	ws.Register("/convert", conversion.NewWebhookHandler(scheme))
	go func() {
		require.NoError(t, ws.Start(ctx))
	}()
```

which fulfilled my needs.

I think we can close this one unless there's something that I missed which would make it reasonable to skip webhook altogether for use cases like mine.

sbueringer:
I'm mostly wondering why it just doesn't work out of the box.

If the schema is configured correctly and conversion is implemented correctly envtest should just create a fully functioning conversion endpoint

pmalek:
> I'm mostly wondering why it just doesn't work out of the box.
> 
> If the schema is configured correctly and conversion is implemented correctly envtest should just create a fully functioning conversion endpoint

For historical reference (in case someone want to debug this):

- This is the code for envtest setup: https://github.com/Kong/kong-operator/blob/5bacab4eb36d952a015ea7b6ac95507f8ca5c2f6/test/envtest/setup.go#L48-L62
- This is the scheme constructor function which is used in every test: https://github.com/Kong/kong-operator/blob/378bd036120af98ac5f5bab84ed96dcccee0b19b/modules/manager/scheme/scheme.go#L22-L43
- This is the CRD in question: https://github.com/Kong/kong-operator/blob/4ce3b776313be72315b98109d8163be5ab1d9bbb/config/crd/kong-operator/gateway-operator.konghq.com_controlplanes.yaml

PR #3389, opened 2025-11-19
✨ Add FieldOwner field to client.Options

aerfio:
Introduce FieldOwner field to client.Options, to allow users to set it once, without doing it like that:
```go
cl, err := client.New(restCfg, client.Options{})
if err != nil {
  panic(err)
}
cl = client.WithFieldOwner(cl, "field-owner")
```

The main reason however is the ability to set it in manager creation, so that the client we get from it using `mgr.GetClient()` is already correctly configured. Otherwise I have to plumb the client manually or use following code:
```go
mgr, err := ctrl.NewManager(restCfg, ctrl.Options{
        ...
		NewClient: func(config *rest.Config, options client.Options) (client.Client, error) {
			cli, err := client.New(config, options)
			if err != nil {
				return nil, err
			}
			return client.WithFieldOwner(cli, "test-owner"), nil
		},
        ...
```

But the documentation of `NewClient` states that it's a low level primitive and should not be used unless you know what you're doing, so I'd rather not overwrite default client creation fn

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -53,7 +53,13 @@ type Options struct {
 	// DryRun instructs the client to only perform dry run requests.
 	DryRun *bool
 
-	// FieldOwner, if provided, instructs the client to be wrapped with WithFieldOwner function
+	// FieldOwner, if provided, sets the default field manager for all write operations
+	// (Create, Update, Patch, Apply) performed by this client. The field manager is used by
+	// the server for Server-Side Apply to track field ownership.
+	// For more details, see: https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management
+	//
+	// This default can be overridden for a specific call by passing a [FieldOwner] option
+	// to the method.
 	FieldOwner string
 }
 


aerfio:
@alvaroaleman My plan is to actually create the same PR but for the `FieldValidation`. I can also prepare other ones for any other field you've got in mind, but right now I can only see 2 `With.*` functions in client package

pkg/client/client.go
```diff
@@ -52,6 +52,9 @@ type Options struct {
 
 	// DryRun instructs the client to only perform dry run requests.
 	DryRun *bool
+
+	// FieldOwner, if provided, instructs the client to be wrapped with WithFieldOwner function
```
sbueringer: I'm fine with the idea and the implementation. Let's just please not leak through the implementation details to the option. Let's instead document what behavior folks are getting if the option is set (aka document the API, not the implementation).

I think we might get to a point where we want to implement this differently (e.g. if a bunch of these options are set do we really want to have a client that is wrapped n times?)

sbueringer:
> seems okay to me. @sbueringer WDYT? Odds are we will end up adding most/all of the wrappers in here.
> 
> /hold

All good from my side, just a minor point on the godoc: https://github.com/kubernetes-sigs/controller-runtime/pull/3389#discussion_r2546590986

pkg/client/client.go
aerfio: @sbueringer Updated! Please have a look

sbueringer:
Thank you very much!

/lgtm
/approve
/hold cancel

PR #3393, opened 2025-11-23
✨ Add FieldValidation setting to client.Options

aerfio:
This PR adds FieldValidation to client.Options:
```go
cl, err := client.New(restCfg, client.Options{
  FieldValidation: metav1.FieldValidationStrict,
})
if err != nil {
  panic(err)
}
```

Documentation explicitly lists only `Create`, `Patch` and `Update`, because Apply methods do not accept FieldValidation argument in their Options, see https://pkg.go.dev/k8s.io/apimachinery@v0.34.2/pkg/apis/meta/v1#ApplyOptions. This is also reflected in current implementation.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -64,7 +64,7 @@ type Options struct {
 
 	// FieldValidation sets the field validation strategy for all mutating operations performed by this client
 	// and subresource clients created from it.
-	// The exception are apply requests which are always strict, regardless of the FieldValidation setting
+	// The exception are apply requests which are always strict, regardless of the FieldValidation setting.
 	// Available values for this option can be found in "k8s.io/apimachinery/pkg/apis/meta/v1" package and are:
 	//  - FieldValidationIgnore
 	//  - FieldValidationWarn


pkg/client/client.go
```diff
@@ -61,6 +61,15 @@ type Options struct {
 	// This default can be overridden for a specific call by passing a [FieldOwner] option
 	// to the method.
 	FieldOwner string
+
+	// FieldValidation, if provided, sets the field validation strategy for all Patch, Update, and Create operations performed by this client and subresource clients created from it.
```
alvaroaleman:
```suggestion
	// FieldValidation sets the field validation strategy for all mutating operations performed by this client
	// and subresource clients created from it.
	// The exception are apply requests which are always strict, regardless of the FieldValidation setting
```

aerfio: @alvaroaleman I removed one extra line below your suggestion

pkg/client/client.go
```diff
@@ -61,6 +61,16 @@ type Options struct {
 	// This default can be overridden for a specific call by passing a [FieldOwner] option
 	// to the method.
 	FieldOwner string
+
+	// FieldValidation sets the field validation strategy for all mutating operations performed by this client
+	// and subresource clients created from it.
+	// The exception are apply requests which are always strict, regardless of the FieldValidation setting
```
alvaroaleman:
```suggestion
	// The exception are apply requests which are always strict, regardless of the FieldValidation setting.
```

PR #3404, opened 2025-12-21
🌱 Fake client should set CreationTimestamp

### LINKED ISSUE #3402: Fake client should set CreationTimestamp when creating the resource
This could help in making fake client better

fabriziopandini:
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3402

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -933,6 +933,14 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 			// Overwrite it unconditionally, this matches the apiserver behavior
 			// which allows to set it on create, but will then ignore it.
 			obj.SetResourceVersion("1")
+
+			now, err := time.Parse(time.RFC3339, time.Now().Format(time.RFC3339))
+			if err != nil {
+				return apierrors.NewInternalError(err)
+			}
+			obj.SetCreationTimestamp(metav1.Time{
+				Time: now,
+			})
 		} else {
 			// SSA deletionTimestamp updates are silently ignored
 			obj.SetDeletionTimestamp(oldAccessor.GetDeletionTimestamp())
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -95,7 +95,8 @@ type fakeClient struct {
 	// indexesLock must be held when accessing indexes.
 	indexesLock sync.RWMutex
 
-	returnManagedFields bool
+	returnManagedFields  bool
+	setCreationTimestamp bool
 }
 
 var _ client.WithWatch = &fakeClient{}
@@ -131,6 +132,7 @@ type ClientBuilder struct {
 	interceptorFuncs      *interceptor.Funcs
 	typeConverters        []managedfields.TypeConverter
 	returnManagedFields   bool
+	setCreationTimestamp  bool
 	isBuilt               bool
 
 	// indexes maps each GroupVersionKind (GVK) to the indexes registered for that GVK.
@@ -256,6 +258,13 @@ func (f *ClientBuilder) WithReturnManagedFields() *ClientBuilder {
 	return f
 }
 
+// WithSetCreationTimestamp configures the fake client to set metadata.creationTimestamp on Create and first Apply.
+// on objects.
+func (f *ClientBuilder) WithSetCreationTimestamp() *ClientBuilder {
+	f.setCreationTimestamp = true
+	return f
+}
+
 // Build builds and returns a new fake client.
 func (f *ClientBuilder) Build() client.WithWatch {
 	if f.isBuilt {
@@ -307,6 +316,7 @@ func (f *ClientBuilder) Build() client.WithWatch {
 		scheme:                        f.scheme,
 		withStatusSubresource:         withStatusSubResource,
 		usesFieldManagedObjectTracker: usesFieldManagedObjectTracker,
+		setCreationTimestamp:          f.setCreationTimestamp,
 	}
 
 	for _, obj := range f.initObject {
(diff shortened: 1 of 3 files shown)


alvaroaleman:
This is be a very breaking change for any unit test that compares content of the fake client with a reference. I am also not sure if using time.Now is a good idea as opposed to some hardcoded fixed tomrstamp to make comparison easier

fabriziopandini:
@alvaroaleman that's a fair comment, this is why I asked in https://github.com/kubernetes-sigs/controller-runtime/issues/3402 before proposing a PR

What do you suggest as a next step? use a const instead of time.Now or do you prefer dropping the change entirely?

alvaroaleman:
I think the ask will come up sooner or later anyways, so doing this seems fine to me in principle, we just have to mark it as breaking change. Using a static const as creation timestamp allows the comparison to still work and _I think_ doesn't really have much drawback other than being different from what the real KAS does - Any opinion on this @sbueringer ?

sbueringer:
I think the only problem with using a constant is that it won't work in cases where either prod or test code uses the creationTimestamp to check if it's before/after other timestamps

(I don't have a use case for this though)

sbueringer:
Just remembered a case. In CAPI we have a controller that rolls out new Machines if they have not been created after a certain timestamp.

Example
* MachineDeployment: rolloutAfter 2025.12.2022 13:00
* controller creates MachineSets/Machines (if there is no MachineSet with creationTimestamp after the timestamp above)

If now the timestamps would be always in the past the controller would never stop creating new MachineSets/Machines.

I imagine other folks might have similar use cases.

sbueringer:
IIRC we already have the comparison problem with managedFields (only if enabled?)

Maybe also with deletionTimestamp, although not that often with deletionTimestamp of course.

I'm leaning towards moving the fake client closer to reality (i.e. using time.Now or something similar)

We could have an option similar to `WithReturnManagedFields` to enable returning a specific timestamp (potentially a timestamp passed into that option)

alvaroaleman:
Hm yeah, fair point regarding the comparison problem. I was going to point out that creationTimestamp isn't very useful for validating one thing happens after another since it only has second granularity but if there is logic around "happened N time in the past" this still causes issues.

It is already possible to set creationTimestamp through the interceptor (Which I think this change breaks?) - Maybe that is enough, as it already provides a configurable opt-in?

fabriziopandini:
I have pushed a new version that makes setting creationTimestamp an opt-in behaviour, without the need to write interceptors.

In the future, if there will be ever agreement on this change, it will be easy to make the switch from opt-in to opt-out.

pkg/client/fake/client.go
```diff
@@ -933,6 +944,16 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 			// Overwrite it unconditionally, this matches the apiserver behavior
 			// which allows to set it on create, but will then ignore it.
 			obj.SetResourceVersion("1")
+
+			if c.setCreationTimestamp {
+				now, err := time.Parse(time.RFC3339, time.Now().Format(time.RFC3339))
+				if err != nil {
+					return apierrors.NewInternalError(err)
+				}
+				obj.SetCreationTimestamp(metav1.Time{
```
alvaroaleman, line 953: Is this the behavior of the KAS to just override it if its set?

pkg/client/fake/client.go
```diff
@@ -256,6 +258,13 @@ func (f *ClientBuilder) WithReturnManagedFields() *ClientBuilder {
 	return f
 }
 
+// WithSetCreationTimestamp configures the fake client to set metadata.creationTimestamp on Create and first Apply.
+// on objects.
```
alvaroaleman, line 262: Remainder?

pkg/client/fake/client_test.go
```diff
@@ -1473,6 +1473,48 @@ var _ = Describe("Fake client", func() {
...
+		It("should be able to Create and set metadata.CreationTimestamp", func(ctx SpecContext) {
+			By("Creating a new configmap")
+			newcm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-cm-with-creation-timestamp",
+					Namespace: "ns2",
+				},
+			}
+			err := cl.Create(ctx, newcm)
+			Expect(err).ToNot(HaveOccurred())
```
alvaroaleman, line 1492: Maybe also check here that its unset?

pkg/client/fake/client_test.go
```diff
@@ -1473,6 +1473,48 @@ var _ = Describe("Fake client", func() {
...
+			}
+			obj := &corev1.ConfigMap{}
+			err = cl.Get(ctx, namespacedName, obj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(obj).To(Equal(newcm))
+			Expect(obj.ObjectMeta.ResourceVersion).To(Equal("1"))
+			Expect(obj.ObjectMeta.CreationTimestamp.IsZero()).ToNot(BeTrue())
+		})
+
+		It("sets creatioTimestamp on SSA create when required to do so", func(ctx SpecContext) {
```
alvaroaleman, line 1507: typo and that its required is implicit from the `Context` I think:

```suggestion
		It("sets creationTimestamp on SSA create", func(ctx SpecContext) {
```

pkg/client/fake/client.go
```diff
@@ -933,6 +944,16 @@ func (c *fakeClient) patch(obj client.Object, patch client.Patch, opts ...client
 			// Overwrite it unconditionally, this matches the apiserver behavior
 			// which allows to set it on create, but will then ignore it.
 			obj.SetResourceVersion("1")
+
+			if c.setCreationTimestamp {
+				now, err := time.Parse(time.RFC3339, time.Now().Format(time.RFC3339))
+				if err != nil {
+					return apierrors.NewInternalError(err)
```
alvaroaleman, line 951: Maybe mention what failed so its easier to understand?

fabriziopandini:
/close

Sorry I do not have bandwidth to continue working on this idea

PR #3416, opened 2026-01-11
:seedling: Priorityqueue: Use separate b-trees for ready and non-ready items

alvaroaleman:
This change refactors the priorityqueue to use distinct b-trees for ready and non-ready items. This simplifies the code, because the two of them effectively have different sorting requirements. We made this work using one b-tree sorted by priority first and readyAt second and then iterating through all priorities high to low until we find a ready item. While correct, this is difficult to reason about.

The other issue is that currently, there is no explicit transition step from non-ready to ready. As a result, we always had to iterate over all items in case one got ready to then update metrics. With this change, we instead have an explicit event when an item gets ready and we can then update metrics for just this item and do not have to iterate over everything.

This change also makes it possible to optimize some of the details further, for example the fifo arguably should be based on when items became ready for items that were added with a non-nil after rather than when they were added. Implementing this using the one tree would be extremely difficult at best.

cc @zach593 I think this should also address the perf issues you mention in https://github.com/karmada-io/karmada/issues/7072

/assign @sbueringer

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -1,7 +1,6 @@
 package priorityqueue
 
 import (
-	"math"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -71,25 +70,27 @@ func New[T comparable](name string, o ...Opt[T]) PriorityQueue[T] {
 	}
 
 	pq := &priorityqueue[T]{
-		log:         opts.Log,
-		items:       map[T]*item[T]{},
-		queue:       btree.NewG(32, less[T]),
-		becameReady: sets.Set[T]{},
-		metrics:     newQueueMetrics[T](opts.MetricProvider, name, clock.RealClock{}),
-		// itemOrWaiterAdded indicates that an item or
+		log:     opts.Log,
+		items:   map[T]*item[T]{},
+		ready:   btree.NewG(32, lessReady[T]),
+		waiting: btree.NewG(32, lessWaiting[T]),
+		metrics: newQueueMetrics[T](opts.MetricProvider, name, clock.RealClock{}),
+		// readyItemOrWaiterAdded indicates that a ready item or
 		// waiter was added. It must be buffered, because
 		// if we currently process items we can't tell
 		// if that included the new item/waiter.
-		itemOrWaiterAdded: make(chan struct{}, 1),
-		rateLimiter:       opts.RateLimiter,
-		locked:            sets.Set[T]{},
-		done:              make(chan struct{}),
-		get:               make(chan item[T]),
-		now:               time.Now,
-		tick:              time.Tick,
+		readyItemOrWaiterAdded:    make(chan struct{}, 1),
+		waitingItemAddedOrUpdated: make(chan struct{}, 1),
+		rateLimiter:               opts.RateLimiter,
+		locked:                    sets.Set[T]{},
+		done:                      make(chan struct{}),
+		get:                       make(chan item[T]),
+		now:                       time.Now,
+		tick:                      time.Tick,
 	}
 
-	go pq.spin()
+	go pq.handleReadyItems()
+	go pq.handleWaitingItems()
 	go pq.logState()
 	if _, ok := pq.metrics.(noMetrics[T]); !ok {
 		go pq.updateUnfinishedWorkLoop()
@@ -100,30 +101,28 @@ func New[T comparable](name string, o ...Opt[T]) PriorityQueue[T] {
 
 type priorityqueue[T comparable] struct {
 	log logr.Logger
-	// lock has to be acquired for any access any of items, queue, addedCounter
-	// or becameReady
-	lock  sync.Mutex
-	items map[T]*item[T]
-	queue bTree[*item[T]]
+	// lock has to be acquired for any access to any of items, ready, waiting,
+	// addedCounter or waiters.
+	lock    sync.Mutex
+	items   map[T]*item[T]
+	ready   bTree[*item[T]]
+	waiting bTree[*item[T]]
 
 	// addedCounter is a counter of elements added, we need it
-	// because unixNano is not guaranteed to be unique.
+	// to provide FIFO semantics.
 	addedCounter uint64
 
-	// becameReady holds items that are in the queue, were added
-	// with non-zero after and became ready. We need it to call the
-	// metrics add exactly once for them.
-	becameReady sets.Set[T]
-	metrics     queueMetrics[T]
+	metrics queueMetrics[T]
 
-	itemOrWaiterAdded chan struct{}
+	readyItemOrWaiterAdded    chan struct{}
+	waitingItemAddedOrUpdated chan struct{}
 
 	rateLimiter workqueue.TypedRateLimiter[T]
 
 	// locked contains the keys we handed out through Get() and that haven't
 	// yet been returned through Done().
 	locked     sets.Set[T]
-	lockedLock sync.RWMutex
+	lockedLock sync.Mutex
 
 	shutdown atomic.Bool
 	done     chan struct{}
@@ -143,6 +142,9 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
 	if w.shutdown.Load() {
 		return
 	}
+
+	var readyItemAdded bool
+	var waitingItemAddedOrUpdated bool
 
 	w.lock.Lock()
 	defer w.lock.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -100,30 +101,28 @@ func New[T comparable](name string, o ...Opt[T]) PriorityQueue[T] {
 
 type priorityqueue[T comparable] struct {
 	log logr.Logger
-	// lock has to be acquired for any access any of items, queue, addedCounter
-	// or becameReady
-	lock  sync.Mutex
-	items map[T]*item[T]
-	queue bTree[*item[T]]
+	// lock has to be acquired for any access to any of items, ready, waiting
```
sbueringer: nit: based on the godoc comment of waiters "waiters" should also be in the list

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -427,23 +457,24 @@ func (w *priorityqueue[T]) logState() {
 	}
 }
 
-func less[T comparable](a, b *item[T]) bool {
+func lessWaiting[T comparable](a, b *item[T]) bool {
+	if !a.ReadyAt.Equal(*b.ReadyAt) {
+		return a.ReadyAt.Before(*b.ReadyAt)
+	}
 	if a.Priority != b.Priority {
```
sbueringer: nit: Would it make sense to dedup and reuse lessReady here? (I guess the idea is that we use the same ordering if ReadyAt is identical as a minor optimization)

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -170,138 +172,169 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
...
+			}
+
+			if len(toMove) > 0 {
+				w.notifyReadyItemOrWaiterAdded()
+			}
+		}()
+	}
+}
+
+func (w *priorityqueue[T]) spin() {
```
sbueringer: nit: maybe rename spin similar to `handleWaitingItems` (entirely up to you)

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -389,12 +422,9 @@ func (w *priorityqueue[T]) Len() int {
...
-			result++
-			return true
-		}
-		return false
+	w.ready.Ascend(func(item *item[T]) bool {
+		result++
+		return true
 	})
 
 	return result
```
sbueringer: nit: Should logState also log the waiting queue?

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -389,12 +422,9 @@ func (w *priorityqueue[T]) Len() int {
 	defer w.lock.Unlock()
 
 	var result int
-	w.queue.Ascend(func(item *item[T]) bool {
-		if item.ReadyAt == nil || item.ReadyAt.Compare(w.now()) <= 0 {
-			result++
-			return true
-		}
-		return false
+	w.ready.Ascend(func(item *item[T]) bool {
```
sbueringer: nit: let's switch to w.ready.Len() (I assume it's more efficient)

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -170,138 +172,169 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
...
+				}
+
+				nextReady = w.tick(readyIn)
+				return false
+			})
+
+			// Don't manipulate the tree from within Ascend
+			for _, toMove := range toMove {
+				w.waiting.Delete(toMove)
+				toMove.ReadyAt = nil
```
sbueringer, line 282: Might be a stupid question but just to be sure. Does this also update ReadyAt in w.items[key]?

(I assume yes because we use *item, but just to double check)

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -170,138 +172,169 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
...
+		} else if readyAt != nil && w.items[key].ReadyAt.Before(*readyAt) {
+			readyAt = w.items[key].ReadyAt
+		}
+
+		priority := w.items[key].Priority
+		addedCounter := w.items[key].AddedCounter
+		if newPriority := ptr.Deref(o.Priority, 0); newPriority > w.items[key].Priority {
+			// Update depth metric only if the item was already ready
+			if w.items[key].ReadyAt == nil {
+				w.metrics.updateDepthWithPriorityMetric(w.items[key].Priority, newPriority)
```
sbueringer, line 197: To double check. This is the only case where we have to call `updateDepthWithPriorityMetric` because this is the only case where we change the priority while an item is in ready.

Apart from that we only increase (add) / decrease (get) when an item is in ready

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -170,138 +172,169 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
...
+		tree := w.ready
+		previousTree := w.ready
+		if readyAt != nil {
+			tree = w.waiting
+			waitingItemAddedOrUpdated = true
+		}
+		if w.items[key].ReadyAt != nil {
+			previousTree = w.waiting
+			if readyAt == nil {
+				w.metrics.add(key, priority)
```
sbueringer: Do we have to set readyItemAdded to true here?

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -170,138 +172,169 @@ func (w *priorityqueue[T]) AddWithOpts(o AddOpts, items ...T) {
...
-			item.AddedCounter = w.addedCounter
+			priority = newPriority
+			addedCounter = w.addedCounter
 			w.addedCounter++
 		}
 
-		if item.ReadyAt != nil && (readyAt == nil || readyAt.Before(*item.ReadyAt)) {
-			if readyAt == nil && !w.becameReady.Has(key) {
-				w.metrics.add(key, item.Priority)
+		tree := w.ready
```
sbueringer: I have the feeling this would be more straightforward with a switch / case that simply lists all 4 cases and then sets tree / previousTree / waitingItemAddedOrUpdated / readyItemAdded / w.metrics.add accordingly

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Yes exactly, the depth metric is only about ready items

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Very good catch. Fixed and added `TestWaitingItemIsReturnedRightAfterReadWithoutAfter` for this which fails before and succeeds after the fix

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Its actually only three cases because we don't have the case of `readyAt != nil && w.items[key] == nil` but yeah, I've updated it, lmk what you think

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Yeah exactly, they are both pointers pointing to the same underlying data

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Oh ofc, updated

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Yes, updated it to do that

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Good idea, done

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Makes sense, renamed to `handleReadyItems`

pkg/controller/priorityqueue/priorityqueue.go
alvaroaleman: Done

pkg/controller/priorityqueue/priorityqueue.go
sbueringer: Yup, agree. Very nice!

sbueringer:
Thank you very much!!

/lgtm
/approve
/hold cancel

Issue #3396, opened 2025-12-01
priority queue metrics cause unbounded memory growth and scrape timeouts


chenghu2:
### Summary

When using controller-runtime's priority queue (`UsePriorityQueue: true`), the `workqueue_depth` metric can cause unbounded memory growth and extremely slow metrics serialization times, leading to Prometheus scrape timeouts.

### Problem

The `workqueue_depth` metric includes a `priority` label that creates a new metric time series for each unique priority value:

```
workqueue_depth{name="my-controller", controller="my-controller", priority="0"} 0
workqueue_depth{name="my-controller", controller="my-controller", priority="1"} 0
workqueue_depth{name="my-controller", controller="my-controller", priority="12345"} 0
```

The Prometheus client library never automatically cleans up these metric entries, even after items are dequeued. If an application uses incrementing priority values (e.g., for LIFO ordering), each enqueue creates a persistent metric entry.

### Impact

In a real-world scenario with ~70 controllers using incrementing priorities:
- **810K+ unique metric entries** accumulated over time
- **15+ second metrics serialization time** (exceeds typical 10s scrape timeout)
- Prometheus scrape failures with "broken pipe" errors

### Root Cause

The metric is defined in `pkg/internal/metrics/workqueue.go`:

```go
var (
    depth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
        Subsystem: WorkQueueSubsystem,
        Name:      DepthKey,
        Help:      "Current depth of workqueue by workqueue and priority",
    }, []string{"name", "controller", "priority"})  // <-- priority label
```

And used in `depthWithPriorityMetric`:

```go
func (g *depthWithPriorityMetric) Inc(priority int) {
    depth.WithLabelValues(append(g.lvs, strconv.Itoa(priority))...).Inc()
}
```

Each unique priority integer becomes a distinct label value, creating a new persistent metric entry.

### Workaround

Applications using custom priorities should bound their priority values to a small range (e.g., 0-100) to limit metric cardinality:

```go
const maxPriority = 100
var priorityCounter atomic.Int32

func enqueue(item T) {
    priority := priorityCounter.Add(1)
    if priority >= maxPriority {
        priorityCounter.Store(0)
    }
    queue.AddWithOpts(priorityqueue.AddOpts{Priority: ptr.To(int(priority))}, item)
}
```

### Potential Solutions

[...]

PR #3419, opened 2026-01-17
:bug: Limit depthWithPriorityMetric cardinality to 25

alvaroaleman:
If a very large number of priorities is used, the cardinality of the priorityQueues `depthWithPriority` metric can grow up to the size of int, which is unreasonable high.

As the functionality is useful and the above is an outlier use-case, make the metric "smart" by collecting the priorities up to a cardinality of 25 unique values and then start using a placeholder that is hopefully self-explanatory.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3396

### DIFF
--- a/pkg/internal/metrics/workqueue.go
+++ b/pkg/internal/metrics/workqueue.go
@@ -18,9 +18,11 @@ package metrics
 
 import (
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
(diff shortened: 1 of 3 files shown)


pkg/internal/metrics/workqueue.go
```diff
@@ -154,17 +156,55 @@ type DepthMetricWithPriority interface {
 var _ MetricsProviderWithPriority = WorkqueueMetricsProvider{}
 
 func (WorkqueueMetricsProvider) NewDepthMetricWithPriority(name string) DepthMetricWithPriority {
-	return &depthWithPriorityMetric{lvs: []string{name, name}}
+	return &depthWithPriorityMetric{depth: depth, lvs: []string{name, name}, observedPriorities: sets.Set[int]{}}
 }
 
+type promehteusGaugeVec interface {
```
sbueringer: Typo

pkg/internal/metrics/workqueue.go
```diff
@@ -154,17 +156,55 @@ type DepthMetricWithPriority interface {
...
+	WithLabelValues(lvs ...string) prometheus.Gauge
+}
+
+const (
+	priorityCardinalityExceededPlaceholder = "__exceeded_cardinality_limit__"
```
sbueringer: Do you know if there are any conventions / prior art about naming / usage of label values like this?

alvaroaleman: None that I could find at least, apparently opentelemetry adds a new label instead to indicate this `otel.metric.overflow=true`: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#cardinality-limits but we use prometheus

pkg/internal/metrics/workqueue.go
alvaroaleman: Fixed

pkg/internal/metrics/workqueue.go
sbueringer: Alright, thx. Good enough

sbueringer: Just realized this
> Label names beginning with __ (two underscores) MUST be reserved for internal Prometheus use.
> https://prometheus.io/docs/concepts/data_model/

sbueringer:
Thx!

/lgtm
/approve

Issue #3424, opened 2026-01-21
ReaderFailOnMissingInformer causes the sync check on the cache to be short-circuited


alvaroaleman:
Basically title, if this option is set the cache may not be in sync, resulting in empty LIST results for example.

The problem is that the check if its synced happens in `ic.GetInformers` but we never call that if this option is set: https://github.com/kubernetes-sigs/controller-runtime/blob/aebc15d7c68925a659ee8ae4a747802b7f87594f/pkg/cache/informer_cache.go#L182-L188

The `started` boolean we check on only checks if the cache was started, but not if it was actually synced.

PR #3425, opened 2026-01-21
 🐛 Cache reader: Wait for cache sync when ReaderFailOnMissingInformer is true

troy0820:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3424

/assign @alvaroaleman

### DIFF
--- a/pkg/cache/informer_cache.go
+++ b/pkg/cache/informer_cache.go
@@ -51,16 +51,7 @@ var _ error = (*ErrCacheNotStarted)(nil)
 // ErrResourceNotCached indicates that the resource type
 // the client asked the cache for is not cached, i.e. the
 // corresponding informer does not exist yet.
-type ErrResourceNotCached struct {
-	GVK schema.GroupVersionKind
-}
-
-// Error returns the error
-func (r ErrResourceNotCached) Error() string {
-	return fmt.Sprintf("%s is not cached", r.GVK.String())
-}
-
-var _ error = (*ErrResourceNotCached)(nil)
+type ErrResourceNotCached = internal.ErrResourceNotCached
 
 // informerCache is a Kubernetes Object cache populated from internal.Informers.
 // informerCache wraps internal.Informers.
@@ -157,7 +148,7 @@ func (ic *informerCache) GetInformerForKind(ctx context.Context, gvk schema.Grou
 		return nil, err
 	}
 
-	_, i, err := ic.Informers.Get(ctx, gvk, obj, applyGetOptions(opts...))
+	_, i, err := ic.Informers.Get(ctx, gvk, obj, false, applyGetOptions(opts...))
 	if err != nil {
 		return nil, err
 	}
@@ -171,7 +162,7 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object, opt
 		return nil, err
 	}
 
-	_, i, err := ic.Informers.Get(ctx, gvk, obj, applyGetOptions(opts...))
+	_, i, err := ic.Informers.Get(ctx, gvk, obj, false, applyGetOptions(opts...))
 	if err != nil {
 		return nil, err
 	}
@@ -179,15 +170,11 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object, opt
 }
 
 func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
-	if ic.readerFailOnMissingInformer {
-		cache, started, ok := ic.Informers.Peek(gvk, obj)
-		if !ok {
-			return false, nil, &ErrResourceNotCached{GVK: gvk}
-		}
-		return started, cache, nil
+	started, cache, err := ic.Informers.Get(ctx, gvk, obj, ic.readerFailOnMissingInformer, &internal.GetOptions{})
+	if err != nil {
+		return false, nil, err
 	}
-
-	return ic.Informers.Get(ctx, gvk, obj, &internal.GetOptions{})
+	return started, cache, nil
 }
 
 // RemoveInformer deactivates and removes the informer from the cache.
--- a/pkg/cache/internal/informers.go
+++ b/pkg/cache/internal/informers.go
@@ -45,6 +45,20 @@ import (
 
 var log = logf.RuntimeLog.WithName("cache")
 
+// ErrResourceNotCached indicates that the resource type
+// the client asked the cache for is not cached, i.e. the
+// corresponding informer does not exist yet.
+type ErrResourceNotCached struct {
+	GVK schema.GroupVersionKind
+}
+
+// Error returns the error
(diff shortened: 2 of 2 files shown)


pkg/cache/informer_cache.go
```diff
@@ -180,8 +181,8 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object, opt
 
 func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
 	if ic.readerFailOnMissingInformer {
-		cache, started, ok := ic.Informers.Peek(gvk, obj)
-		if !ok {
+		started, cache, err := ic.Informers.Get(ctx, gvk, obj, &internal.GetOptions{BlockUntilSynced: ptr.To(true)})
```
troy0820: Can't use this, this starts the cache if it wasn't started.

troy0820: Using returned cache, checked to see if `Informer.HasSynced()`

pkg/cache/informer_cache.go
```diff
@@ -184,6 +184,9 @@ func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.Grou
 		if !ok {
 			return false, nil, &ErrResourceNotCached{GVK: gvk}
 		}
+		if !cache.Informer.HasSynced() {
```
sbueringer: This might still differ too much from the behavior in Informers.Get

Maybe ic.readerFailOnMissingInformer should be passed into Informers.Get and handled there to make it easier to keep both branches in sync.

@alvaroaleman WDYT?

alvaroaleman: Yeah, we should use the same code to determine and do the syncing regardless of if `readerFailOnMissingInformer` is set or not, that it was added outside of that was basically the main mistake back then.

troy0820: >Yeah, we should use the same code to determine and do the syncing regardless of if readerFailOnMissingInformer is set or not, that it was added outside of that was basically the main mistake back then.

Wouldn't that change the test that is with it? 
```go
		Describe("as a Reader", func() {
			Context("with structured objects", func() {
				It("should not be able to list objects that haven't been watched previously", func(ctx SpecContext) {
					By("listing all services in the cluster")
					listObj := &corev1.ServiceList{}
					Expect(errors.As(informerCache.List(ctx, listObj), &errNotCached)).To(BeTrue())
				})

```

I am trying to understand the behavior we want because if we don't care whether this is set or not:
1. Should we fail if this is set? 
1. Should we add to the informerMap and then wait for sync, then fail?

alvaroaleman: Assuming the above test is for a cache that has `ReaderFailOnMissingInformer` it needs to keep passing, yes. The only difference in `Get` between that option being set and not should be that we error out if the informer is not in the map already

troy0820: Gotcha, that's how I understood it after I re-read the issue and the setting.  I am dealing with a context cancelled test failure at the moment, that I am hunting down.

pkg/cache/internal/informers.go
```diff
@@ -294,10 +294,13 @@ func (ip *Informers) Peek(gvk schema.GroupVersionKind, obj runtime.Object) (res
 
 // Get will create a new Informer and add it to the map of specificInformersMap if none exists. Returns
 // the Informer from the map.
-func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, opts *GetOptions) (bool, *Cache, error) {
+func (ip *Informers) Get(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object, readerFailOnMissingInformer bool, opts *GetOptions) (bool, *Cache, error) {
 	// Return the informer if it is found
 	i, started, ok := ip.Peek(gvk, obj)
 	if !ok {
+		if readerFailOnMissingInformer {
+			return false, nil, fmt.Errorf("failed to get informer for %v: informer not found", gvk)
```
alvaroaleman: This is wrong, this here needs to return `ErrResourceNotCached`. That in turn requires to move that to `internal`, something like this will do that without breaking the export:

```diff
+++ b/pkg/cache/informer_cache.go
@@ -51,16 +51,7 @@ var _ error = (*ErrCacheNotStarted)(nil)
 // ErrResourceNotCached indicates that the resource type
 // the client asked the cache for is not cached, i.e. the
 // corresponding informer does not exist yet.
-type ErrResourceNotCached struct {
-       GVK schema.GroupVersionKind
-}
-
-// Error returns the error
-func (r ErrResourceNotCached) Error() string {
-       return fmt.Sprintf("%s is not cached", r.GVK.String())
-}
-
-var _ error = (*ErrResourceNotCached)(nil)
+type ErrResourceNotCached = internal.ErrResourceNotCached
```

pkg/cache/informer_cache.go
```diff
@@ -171,23 +171,19 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object, opt
...
 func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
-	if ic.readerFailOnMissingInformer {
-		cache, started, ok := ic.Informers.Peek(gvk, obj)
-		if !ok {
-			return false, nil, &ErrResourceNotCached{GVK: gvk}
-		}
-		return started, cache, nil
+	started, cache, err := ic.Informers.Get(ctx, gvk, obj, ic.readerFailOnMissingInformer, &internal.GetOptions{})
+	if err != nil {
+		return false, nil, &ErrResourceNotCached{GVK: gvk}
```
alvaroaleman: This here needs to just return the error and not overwrite it with `ErrResourceNotCached`, as there is a range of other errors that `Informers.Get` may return

pkg/cache/informer_cache.go
```diff
@@ -171,23 +162,19 @@ func (ic *informerCache) GetInformer(ctx context.Context, obj client.Object, opt
...
 
-	_, i, err := ic.Informers.Get(ctx, gvk, obj, applyGetOptions(opts...))
+	_, i, err := ic.Informers.Get(ctx, gvk, obj, ic.readerFailOnMissingInformer, applyGetOptions(opts...))
 	if err != nil {
 		return nil, err
 	}
 	return i.Informer, nil
 }
 
 func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, obj runtime.Object) (bool, *internal.Cache, error) {
```
alvaroaleman, line 172: Its not very obvious from the code but this is the only place where we need to pass `ic.readerFailOnMissingInformer`, in the two others we need to pass `false`.
This is because this here is only used by the cache reader whereas `GetInformer` and `GetInformerForKind` are methods on the cache. If those error out when `readerFailOnMissingInformer` is true, it becomes impossible to construct an informer at all

Issue #3242, opened 2025-06-24
Question: FieldSelector != is not supported by FakeClient


marvinbstackit:
I am currently writing unit tests in my go application and I list events. These events can be filtered with field selectors to exclude some events. In the code everything is fine, but in the fake client the field slector will be evaluated to only support '=' and '=='. Why is this so and what does it need to support also '!='?

[This](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/field/selector/utils.go#L25) is the code piece, but in the actual client is never called because the parameter is used in the api call [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/typed_client.go#L158).

alvaroaleman:
This predates my involvement in the project, but I believe the reason is that in most cases, ppl read from the cache and not from the api. The cache allows to set up indexes and these are used through fieldSelectors. The problem now is that there is no good way to find all elements that are not in a given index, which is why both the cache reader and the fake client error out when trying to use such a fieldSelector.

PR #3427, opened 2026-01-23
:bug: add not equal as a supported operator for field selectors

zak905:
Currently, field selectors support the "not equal" operator. This affects the fake client and also the `CacheReader`. 
Here is the reference from Kubernetes documentation: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-operators

> You can use the =, ==, and != operators with field selectors (= and == mean the same thing).

### DIFF
--- a/pkg/internal/field/selector/utils.go
+++ b/pkg/internal/field/selector/utils.go
@@ -29,7 +29,7 @@ func RequiresExactMatch(sel fields.Selector) bool {
 	}
 
 	for _, req := range reqs {
-		if req.Operator != selection.Equals && req.Operator != selection.DoubleEquals {
+		if req.Operator != selection.Equals && req.Operator != selection.DoubleEquals && req.Operator != selection.NotEquals {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/internal/field/selector/utils.go
```diff
@@ -29,7 +29,7 @@ func RequiresExactMatch(sel fields.Selector) bool {
 	}
 
 	for _, req := range reqs {
-		if req.Operator != selection.Equals && req.Operator != selection.DoubleEquals {
+		if req.Operator != selection.Equals && req.Operator != selection.DoubleEquals && req.Operator != selection.NotEquals {
```
alvaroaleman, line 32: This will not actually make anything work in either the cache or the fake client, it only breaks the validation. It also doesn't make sense to allow not equal in a fucntion named `RequiesExactMatch`

alvaroaleman: Relevant context: https://github.com/kubernetes-sigs/controller-runtime/issues/3242#issuecomment-3000829251

I'd want to see a test on the cache reader that tests that not equal actual works correctly before we merge anything here.
/hold

zak905: Hi @alvaroaleman, I was not aware that this such small change has such implications. Ok let me see what I can do to for the testing.

sbueringer: I think it's not just testing. This behavior has to be explicitly implemented and in an efficient way. We have that today with indices and exact match

zak905: I thought so. ok let me see what I can do.

sbueringer:
/close

No activity in the last two months

PR #3429, opened 2026-01-23
✨ Add pre-start hook support to manager

leontappe:
I have a use case similar to the one described in [a previous PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2044) and [this issue](https://github.com/kubernetes-sigs/controller-runtime/issues/607).

My controller needs to consume job status files on start, that would normally be worked on by a custom runnable during runtime. With vanilla controller-runtime this is currently not possible as any code that is run before reconciliation, has no access to a running manager. 

This PR provides a way to register runnables as hooks that you can supply to your manager.

This PR is based on a fork of https://github.com/terinjokes/controller-runtime/tree/prestart-hook which is in turn a fork of this repo. I rebased the current master onto the latest development state of the original PR and implemented some of the proposals made by @JoelSpeed


pkg/manager/manager.go
```diff
@@ -281,6 +284,10 @@ type Options struct {
 	// +optional
 	Controller config.Controller
 
+	// HookTimeout is the duration given to each hook to return successfully.
+	// To use hooks without timeout, set to a negative duration, e.g. time.Duration(-1)
+	HookTimeout *time.Duration
```
maxihafer, line 289: When @leontappe and I evaluated the [proposed change](https://github.com/kubernetes-sigs/controller-runtime/pull/2044/files#r2321619387) in the old PR we decided to keep the `-1 -> no-timeout` in order to conform with previously decided behavior for [GracefulShutdownTimeout](https://github.com/kubernetes-sigs/controller-runtime/blob/aebc15d7c68925a659ee8ae4a747802b7f87594f/pkg/manager/manager.go#L277).
Other projects also use `-1` as a value to the denote absence of any timeout/ratelimiting ([rest.Config.QPS](https://github.com/kubernetes/client-go/blob/c5e14be2544b0035e15b704271440213cf82c6b2/rest/config.go#L122))

Given that a `0` value is not logically valid, we could reset to the default here, this is also done in GracefulShotdownTimeout

pkg/manager/manager.go
```diff
@@ -82,6 +82,9 @@ type Manager interface {
 	// AddReadyzCheck allows you to add Readyz checker
 	AddReadyzCheck(name string, check healthz.Checker) error
 
+	// Hook allows to add Runnables as hooks to modify the behavior.
+	Hook(hook HookType, runnable Runnable) error
```
alvaroaleman, line 86: I know @vincepri had suggested this name in the other PR, but I am not a huge fan because IMHO:
* It is different from `Add` in that its name is what is being added rather than whats being done/the operation
* Having a `HookType` argument when there is only one `HookType` seems needlessly complicated to me and requires us to spread out godocs explaining what this does to this method and the `HookType` rather than having one place where we can explain it

I would suggest `AddPrestartHook` since that follows the existing naming pattern and is easy to discover. If we ever add another hook, we can then name that `Add${OtherType}Hook`.

But lets wait for a second opinion from @sbueringer before you end up changing this back and forth

pkg/manager/manager.go
```diff
@@ -281,6 +284,10 @@ type Options struct {
 	// +optional
 	Controller config.Controller
 
+	// HookTimeout is the duration given to each hook to return successfully.
```
alvaroaleman, line 287: WDYT about the idea of having a per-hook timeout rather than a global one for all hooks? Different hook may need different timeouts

pkg/manager/internal.go
```diff
@@ -648,6 +674,27 @@ func (cm *controllerManager) initLeaderElector() (*leaderelection.LeaderElector,
 }
 
 func (cm *controllerManager) startLeaderElectionRunnables() error {
+	cm.logger.Info("Running prestart hooks")
+	for _, hook := range cm.prestartHooks {
```
alvaroaleman, line 678: Other than that being what you found in the original PR, what is the reason for starting the hooks sequentially and not in parallel? Couldn't the sequential starting be achieved if the user instead submitted one hook that does everything sequentlally?

pkg/manager/internal.go
```diff
@@ -648,6 +674,27 @@ func (cm *controllerManager) initLeaderElector() (*leaderelection.LeaderElector,
...
+
+		defer cancel()
+
+		if err := hook.Start(ctx); err != nil {
+			return err
+		}
+	}
+
+	// All the prestart hooks have ben run, clear the slice to free the underlying resources.
+	cm.prestartHooks = nil
```
alvaroaleman, line 696: Maybe add a log here `finished running hooks` or such so that if this hangs, there is at least some chance to figure out it is because of the hooks?

pkg/manager/manager.go
alvaroaleman: I think the comparisons are not great, `GracefulShutdownTimeout` has an additional state of `disable this` which is indicated by `0` and `restConfig.QPS` is not a pointer.

I would probably build this as `a value <= 0 means no timeout`

alvaroaleman (commented):
/cc @JoelSpeed

pkg/manager/internal.go
JoelSpeed: Probably need to think about how we would collect errors from pre-start hooks on startup and use that to bail out if there are any errors, but that sounds like it would be fairly straightforward with a waitgroup

pkg/manager/manager.go
JoelSpeed: I'd prefer an explicit value meaning no timeout that we document personally, so I'd say if this is 0, it's no timeout

* nil: no opinion, use something default
* explicitly 0: no timeout
* explicitly >0: use this as the timeout
* explicitly <0: error

pkg/manager/internal.go
```diff
@@ -55,6 +55,7 @@ const (
 	defaultRenewDeadline          = 10 * time.Second
 	defaultRetryPeriod            = 2 * time.Second
 	defaultGracefulShutdownPeriod = 30 * time.Second
+	defaultHookPeriod             = 15 * time.Second
```
sbueringer, line 58:
```suggestion
	defaultHookTimeoutPeriod             = 15 * time.Second
```
nit

pkg/manager/internal.go
```diff
@@ -240,6 +248,24 @@ func (cm *controllerManager) GetHTTPClient() *http.Client {
...
+func (cm *controllerManager) Hook(hook HookType, runnable Runnable) error {
+	cm.Lock()
+	defer cm.Unlock()
+
+	if cm.started {
+		return fmt.Errorf("unable to add new hook because the manager has already been started")
+	}
+
+	switch hook {
+	case HookPrestartType:
```
sbueringer, line 261: Should this be HookTypePrestart?

pkg/manager/internal.go
sbueringer: Agree. I think we should start them in parallel

pkg/manager/internal.go
```diff
@@ -648,6 +674,27 @@ func (cm *controllerManager) initLeaderElector() (*leaderelection.LeaderElector,
 }
 
 func (cm *controllerManager) startLeaderElectionRunnables() error {
```
sbueringer, line 676: Let's rename the func to startPreStartHooksAndLeaderElectionRunnables

pkg/manager/manager.go
sbueringer: Agree, AddPrestartHook sounds good.

Is there some prior art from other components (kube-controller-manager, other controller frameworks) that might give us an idea how likely it is that we'll have other types of hooks in the future and what these might be?

(but if we don't have more data let's go with AddPrestartHook)

Hm. Is "Prestart" precise enough to describe "before leader election runnables"? No strong opinion, but it's pretty generic.

pkg/manager/manager.go
sbueringer: What other timeout fields do we have in CR as prior art? Or do we only have GracefulShutdownTimeout? (just to end up with something somewhat consistent, it makes sense to make different decisions if the cases are different but we should minimize variations if possible)

pkg/manager/manager.go
sbueringer: \+ "to modify the behavior" sounds strange. Let's try to explain this a bit better

pkg/manager/manager.go
```diff
@@ -577,6 +594,11 @@ func setOptionsDefaults(config *rest.Config, options Options) (Options, error) {
 		options.GracefulShutdownTimeout = &gracefulShutdownTimeout
 	}
 
+	if options.HookTimeout == nil {
+		hookTimeout := defaultHookPeriod
+		options.HookTimeout = &hookTimeout
```
sbueringer, line 599: nit: if we keep this let's use ptr.To()

alvaroaleman:
@leontappe any plans to address the feedback? Otherwise this will likely not make it into the next minor.

sbueringer:
@leontappe *friendly reminder* :)

Issue #3423, opened 2026-01-21
Fake client on version 0.23.0 fails to SSA Apply status after a finalizer is set


josvazg:
We might have found an issue with version 0.23.0's **fake client** implementation around the new SSA Apply support.

We noticed that if an object is patched with a finaliser and then the status is SSA applied, the apply operation fails with:

**metadata.managedFields must be nil**

The same apply works fine if there was no finaliser patch before.

It appears the apply finds the managed fields due to the previous unrelated finaliser patch.

Steps to reproduce:
1. Set up a **fake client** for unit testing with some object, e.g. a Pod.
2. Patch a finalizer in the pod.
3. Attempt a SSA Apply on the Pod, maybe set the ready status.

Expected result: SSA Apply should work (as it does if there is no prior finalizer patch)

Actual result: SSA Apply fails with **metadata.managedFields must be nil**

We attach a minimal reproducing unit test:
(For brevity this is the structured or strongly typed version, but our unstructured variant also reproduced the same behaviour)
```
package test

import (
	"context"
	"encoding/json"
	"testing"

	"github.com/stretchr/testify/require"
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/types"
	corev1apply "k8s.io/client-go/applyconfigurations/core/v1"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/client/fake"
	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
)

var patchFinalizerTyped = true // with patching the bug appears

func TestControllerRuntimeSSABugTyped(t *testing.T) {
	ctx := context.Background()

	// Step 1: Create a default Pod object
	pod := &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "test-pod-typed",
			Namespace: "default",
		},
		Spec: corev1.PodSpec{
			Containers: []corev1.Container{
				{
					Name:  "test-container",
					Image: "nginx:latest",
				},
			},
		},
		Status: corev1.PodStatus{
			Phase: corev1.PodPending,
		},
	}

	// Create fake client with the default Pod object
	scheme := runtime.NewScheme()
	require.NoError(t, corev1.AddToScheme(scheme))
[...]

PR #3430, opened 2026-01-24
:bug: Fakeclient: Fix status apply if existing object has managedFields set

alvaroaleman:
The tracker we use in the fakeclient doesn't support status operations, so we implemented them by copying everything but status and ResourceVersion from the existing object.

This meant that we copied the existing objects managedFields and that in turn made the `managedFieldsObjectTracker` error out on Apply with an `metadata.managedFields must be nil`.

Stop copying the managedFields in status updates.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3423


sbueringer:
Thx!

/lgtm
/approve

Issue #3418, opened 2026-01-16
Fake client panics when used with a CRD that embedds a struct pointer


zoroyihan:
After upgrading controller-runtime version from v0.21.0 to v0.22.4.
A previously working test that used fakeclient is now panicking.

panic log: 

panic: reflect: call of reflect.Value.Field on ptr Value [recovered]
        panic: reflect: call of reflect.Value.Field on ptr Value

goroutine 89 [running]:
testing.tRunner.func1.2({0x47a9be0, 0xc001197de8})
        /usr/local/go/src/testing/testing.go:1734 +0x22a
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1737 +0x428
panic({0x47a9be0?, 0xc001197de8?})
        /usr/local/go/src/runtime/panic.go:792 +0x136
reflect.Value.Field({0x4838240?, 0xc000dda170?, 0x4838240?}, 0x0)
        /usr/local/go/src/reflect/value.go:1272 +0x197
reflect.Value.FieldByIndex({0x4838240?, 0xc000dda170?, 0x45e3c00?}, {0x51c7d20?, 0x2199b65?, 0x45e3c00?})
        /usr/local/go/src/reflect/value.go:1305 +0x71
sigs.k8s.io/structured-merge-diff/v6/value.(*FieldCacheEntry).GetFrom(0x0?, {0x4cd4980?, 0xc000dda110?, 0x14?})
        ${GOPATH}/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.0/value/reflectcache.go:87 +0x6c
sigs.k8s.io/structured-merge-diff/v6/value.eachStructField({0x4cd4980?, 0xc000dda110?, 0xc0010e5a58?}, 0xc0010e5a40)
        ${GOPATH}/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.0/value/structreflect.go:118 +0xeb
sigs.k8s.io/structured-merge-diff/v6/value.structReflect.IterateUsing({{0x0, 0x0, {0x4cd4980, 0xc000dda110, 0x199}, 0x1}}, {0x5172610, 0xc001b90a40}, 0xc001b8dee0)
        ${GOPATH}/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.0/value/structreflect.go:111 +0xfc
[...]

PR #3431, opened 2026-01-25
:bug: Fix panic when using CRs with embedded pointer structs

alvaroaleman:
This change bumps structured-merge-diff to pick up a fix for it panicing when encountering objects with embedded pointer structs that are nil: https://github.com/kubernetes-sigs/structured-merge-diff/pull/309

This makes us use a newer version of structured-merge-diff than upstream, but for one release that seems okay, especially given that the only usage is in the fake client.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3418
/assign @sbueringer 
/cherrypick release-0.23

### DIFF
--- a/.gomodcheck.yaml
+++ b/.gomodcheck.yaml
@@ -9,6 +9,10 @@ upstreamRefs:
   # k8s.io/utils -> conflicts with k/k deps
 
 excludedModules:
+  # Needs a newer version to fix https://github.com/kubernetes-sigs/controller-runtime/issues/3418
+  # This should not be needed by the time we update to 1.36
+  - sigs.k8s.io/structured-merge-diff/v6
+
   # --- test dependencies:
   - github.com/onsi/ginkgo/v2
   - github.com/onsi/gomega
--- a/examples/scratch-env/go.mod
+++ b/examples/scratch-env/go.mod
@@ -61,7 +61,7 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.2-0.20260122202528-d9cc6641c482 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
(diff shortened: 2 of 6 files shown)


sbueringer:
Is this only for main or is the idea to backport it?

(For a backport I wonder if this might break some folks somehow, I think they could bump directly by themselves if they want)

alvaroaleman:
> Is this only for main or is the idea to backport it?

(For a backport I wonder if this might break some folks somehow, I think they could bump directly by themselves if they want)

Backport to 0.23. How do you think this could break anyone?

sbueringer:
> Backport to 0.23. How do you think this could break anyone?

I don't know if it breaks someone but downstream consumers will inherit sigs.k8s.io/structured-merge-diff/v6 v6.3.2-0.20260122202528-d9cc6641c482 instead of sigs.k8s.io/structured-merge-diff/v6 v6.3.0.

That means they would get all the changes that happened between those two commits. Given that this library might also be used by downstream consumers outside of the usage in controller-runtime I wonder if they end up with different behavior

(This seems to be the diff: https://github.com/kubernetes-sigs/structured-merge-diff/compare/v6.3.0...d9cc6641c482)

Just not sure if it's worth the risk in a CR patch release considering we are only fixing the fake client. And consumers of CR can also use a replace to bump this dependency if they want.

alvaroaleman:
> Just not sure if it's worth the risk in a CR patch release considering we are only fixing the fake client. And consumers of CR can also use a replace to bump this dependency if they want.

I guess my assumption here is that the odds that this breaks something else are very low and then downstream consumers can still overwrite it. But if you feel strongly about not wanting to backport, I won't insist

sbueringer:
No strong opinion, your call. Just wanted to bring it up at least

(Folks can also use replace to pin back to v6.3.0 if they run into issues)

/lgtm
/approve
(definitely fine for main)

alvaroaleman:
I do feel its worth it because the fake client becomes unusable for someone with CRDs that have this in both 0.22 and 0.23

sbueringer:
Sounds good. Strange that we didn't see any issues with CR v0.22 or v0.23 in Cluster API. We don't have a lot of struct pointer fields, but still some. Or does it only happen with a top-level pointer?

alvaroaleman:
> Sounds good. Strange that we didn't see any issues with CR v0.22 or v0.23 in Cluster API. We don't have a lot of struct pointer fields, but still some. Or does it only happen with a top-level pointer?

Its not just a struct pointers, its an embedded struct pointer and must have the `json",inline"` marker. Do you have tests where such an object is passed to the fakeclient? If yes and that didn't cause the panic we should have a look.

sbueringer:
> Its not just a struct pointers, its an embedded struct pointer and must have the `json",inline"` marker. Do you have tests where such an object is passed to the fakeclient? If yes and that didn't cause the panic we should have a look.

Ah thx. That's it. I somehow missed the "embedded" part. We have no cases of embedded structs (apart from TypeMeta and ObjectMeta but both are not pointers or don't use inline)

PR #3434, opened 2026-01-25
:seedling: Test cache reader waits for cache sync

alvaroaleman:
We don't seem to have any test coverage of this at all, change that.

/hold

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -24,6 +24,8 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"testing"
+	"testing/synctest"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


pkg/cache/cache_test.go
```diff
@@ -2588,3 +2590,58 @@ func cancelledCtx(ctx context.Context) context.Context {
...
+				go func() {
+					g.Expect(c.Start(ctx)).To(Succeed())
+					close(cacheDone)
+				}()
+				synctest.Wait() // Let the cache finish starting
+				_, err = c.GetInformer(ctx, &corev1.Service{}, cache.BlockUntilSynced(false))
+				g.Expect(err).ToNot(HaveOccurred())
+
+				services := &corev1.ServiceList{}
+				listCtx, listCtxcancel := context.WithTimeout(ctx, time.Second)
```
alvaroaleman: The sync check internally works through polling (We've inherited that from upstream). I both want to keep the timeout high enough to be sure it polls a few times even if maybe in the future the poll interval changes but also not have the test actually run for poll interval duration, hence using synctest.

pkg/cache/cache_test.go
```diff
@@ -2588,3 +2590,58 @@ func cancelledCtx(ctx context.Context) context.Context {
...
+				g.Expect(err).NotTo(HaveOccurred())
+
+				ctx, cancel := context.WithCancel(t.Context())
+				defer cancel()
+				cacheDone := make(chan struct{})
+				go func() {
+					g.Expect(c.Start(ctx)).To(Succeed())
+					close(cacheDone)
+				}()
+				synctest.Wait() // Let the cache finish starting
```
alvaroaleman, line 2627: Coordinating "cache has finished starting up" with the rest would require us to `sleep` without synctest, as `Start` is blocking

sbueringer:
Thx!

/lgtm
/approve

Feel free to remove hold

PR #3440, opened 2026-01-27
 🐛 manager: fix infinite CPU spin in runnableGroup.Start on context cancellation

atharrva01:
## Summary

This PR fixes a control-flow bug in the manager startup lifecycle where `runnableGroup.Start()` could enter an infinite loop when the context is cancelled before all runnables signal readiness.
The issue could cause manager startup or shutdown to hang indefinitely while consuming CPU.

---

## Impact

Before this change, the manager could:

* Spin indefinitely at **100% CPU** after context cancellation
* Hang forever during startup if **any runnable fails its readiness check**
* Hang during **SIGTERM / shutdown** while waiting for runnables
* Leak goroutines stuck inside `Start()`

These scenarios are common in real deployments (RBAC misconfigurations, cache sync failures, API server unavailability), making the issue both user-visible and hard to diagnose.

---

### Scenario 1: Startup failure

1. Start a manager with a runnable whose cache sync fails (e.g. missing RBAC)
2. The runnable never signals readiness
3. `Start()` waits forever and the manager never becomes ready

### Scenario 2: Shutdown during startup

1. Start the manager
2. Send SIGTERM before all runnables become ready
3. Context is cancelled, but `Start()` keeps looping and burns CPU

---

## Root Cause

The startup loop correctly listens for `ctx.Done()`, but **does not exit the loop** when the context is cancelled.

```go
case <-ctx.Done():
    if err := ctx.Err(); !errors.Is(err, context.Canceled) {
        retErr = err
    }
    // loop continues → ctx.Done() fires forever
```

Since `ctx.Done()` is a closed channel after cancellation, the `select` immediately re-selects it on every iteration, causing a tight busy loop.

---

## Fix

Exit the startup loop immediately when the context is cancelled:

```go
case <-ctx.Done():
    if err := ctx.Err(); !errors.Is(err, context.Canceled) {
        retErr = err
    }
    return
```

This ensures:

* The startup loop terminates correctly
* Errors are propagated to the caller
* Shutdown and startup behave predictably

---

## Behavior After Fix

* Manager startup fails fast instead of hanging
* Shutdown completes cleanly without CPU spin
* No goroutine leaks from stuck `Start()` calls
* Already-started runnables continue running on their own context and are cleaned up via `StopAndWait()`

---

## Verification

* ✅ Code compiles successfully
* ✅ `go vet` passes
* ✅ Existing unit tests pass
* ⚠️ Integration tests require `envtest` binaries (not available in this environment)

---

### DIFF
--- a/pkg/manager/runnable_group.go
+++ b/pkg/manager/runnable_group.go
@@ -167,12 +167,6 @@ func (r *runnableGroup) Start(ctx context.Context) error {
 	var retErr error
 
 	r.startOnce.Do(func() {
-		// NOTE: We intentionally do NOT close startReadyCh here.
-		// Closing a channel that has multiple senders (the readiness check
-		// goroutines in reconcile()) violates Go's channel ownership rules
-		// and causes "send on closed channel" panics. Instead, senders check
-		// ctx.Done() before sending, and the channel is left for GC.
-
 		// Start the internal reconciler.
 		go r.reconcile()
 
@@ -212,8 +206,10 @@ func (r *runnableGroup) Start(ctx context.Context) error {
 						break
 					}
 				}
-				// We're done waiting if the queue is empty, return.
+				// We're done waiting if the queue is empty.
+				// All senders have already sent, so it is safe to close.
 				if len(r.startQueue) == 0 {
+					close(r.startReadyCh)
 					return
 				}
 			}
--- a/pkg/manager/runnable_group_test.go
+++ b/pkg/manager/runnable_group_test.go
@@ -399,7 +399,7 @@ func TestStartReturnsWhenContextCancelledWithPendingReadinessCheck(t *testing.T)
 	// Cancel the context
 	cancel()
 
-	// Start() should return promptly (within 100ms), not hang or spin
+	// Start() should return promptly after context cancellation, not hang or spin
 	select {
 	case <-startReturned:
 		// Success: Start() returned after context cancellation


atharrva01:
hi @inteon @vincepri This fixes a startup/shutdown hang where runnableGroup.Start() could spin indefinitely after context cancellation by returning immediately on ctx.Done().

pkg/manager/runnable_group.go
```diff
@@ -194,6 +194,7 @@ func (r *runnableGroup) Start(ctx context.Context) error {
 				if err := ctx.Err(); !errors.Is(err, context.Canceled) {
 					retErr = err
 				}
+				return
```
sbueringer, line 195: Would be good if we extend test coverage to cover this

sbueringer:
Nice catch, thx!

PTAL at the CLA: https://github.com/kubernetes-sigs/controller-runtime/pull/3440#issuecomment-3804524776

/ok-to-test

atharrva01:
hi @sbueringer I’ve added a focused regression test covering the context-cancellation path in runnableGroup.Start(). I ran the unit test locally; integration tests require envtest binaries, so CI should cover those. Thanks for the suggestion!

image

sbueringer:
@atharrva01 The 🐛 prefix in the PR title I added was intentional. We use this to generate release notes and without it the "PR title verifier" action will block merging this PR

atharrva01:
hi @sbueringer Thanks for the clarification! 👍
I’ve kept that  manager: prefix as suggested.

sbueringer:
Feel free to ping me for review once the CLA is done

atharrva01:
Hi @sbueringer, the Easy CLA is now completed ✅
Whenever you have time, I’ve pinged you for review as suggested. Thanks!

pkg/manager/runnable_group_test.go
```diff
@@ -363,6 +363,57 @@ func (r leaderElectionAndWarmupRunnable) NeedLeaderElection() bool {
...
+			return nil
+		}),
+		func(ctx context.Context) bool {
+			// Block forever - never signal ready
+			<-ctx.Done()
+			return false
+		},
+	)).To(Succeed())
+
+	ctx, cancel := context.WithCancel(context.Background())
```
sbueringer: Let's use t.Context() instead of context.Background(), same below (see linter findings)

pkg/manager/runnable_group_test.go
```diff
@@ -363,6 +363,57 @@ func (r leaderElectionAndWarmupRunnable) NeedLeaderElection() bool {
...
+	time.Sleep(10 * time.Millisecond)
+
+	// Cancel the context
+	cancel()
+
+	// Start() should return promptly (within 100ms), not hang or spin
+	select {
+	case <-startReturned:
+		// Success: Start() returned after context cancellation
+	case <-time.After(100 * time.Millisecond):
```
sbueringer: Maybe let's use 5s or so just to make sure we don't get a flake in CI. 100ms should be enough, but not sure how slow CI might be

atharrva01: hi @sbueringer I’ve updated the test to use t.Context() and relaxed the timeout to avoid potential CI flakes, as suggested. Changes are pushed, retriggering CI now.

/retest

sbueringer: @atharrva01 One more linter finding (please see the lint/test results below :))

image

atharrva01: hi @sbueringer Thanks for catching that , there was one remaining context.Background() in the test.
I’ve replaced it with t.Context() to satisfy lint. Changes are pushed.

pkg/manager/runnable_group_test.go
```diff
@@ -363,6 +363,54 @@ func (r leaderElectionAndWarmupRunnable) NeedLeaderElection() bool {
 	return r.needLeaderElection
 }
 
+// TestStartReturnsWhenContextCancelledWithPendingReadinessCheck verifies that
+// runnableGroup.Start() returns promptly when the context is cancelled, even if
```
sbueringer, line 367: Looks like there's an issue, not sure if in prod or test code

```
WARNING: DATA RACE
Write at 0x00c0006184e0 by goroutine 191:
  runtime.closechan()
      /home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/chan.go:414 +0x0
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).Start.func1.deferwrap1()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:170 +0x33
  runtime.deferreturn()
      /home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/panic.go:589 +0x5d
  sync.(*Once).doSlow()
      /home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/sync/once.go:78 +0xd1
  sync.(*Once).Do()
      /home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/sync/once.go:69 +0x44
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).Start()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:169 +0xb5
  sigs.k8s.io/controller-runtime/pkg/manager.init.func5.3.1()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group_test.go:181 +0xca

Previous read at 0x00c0006184e0 by goroutine 238:
  runtime.chansend()
      /home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/runtime/chan.go:176 +0x0
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1.1()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:249 +0xc4

Goroutine 191 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/manager.init.func5.3()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group_test.go:178 +0xf9
  github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
      /home/prow/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942 +0x6ed

Goroutine 238 (running) created at:
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:246 +0xfc
  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.gowrap1()
      /home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:284 +0x41
==================
panic: send on closed channel

goroutine 281 [running]:
sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1.1()
	/home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:249 +0xc5
created by sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1 in goroutine 279
	/home/prow/go/src/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:246 +0xfd
FAIL	sigs.k8s.io/controller-runtime/pkg/manager	5.433s
```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3440/pull-controller-runtime-test/2018359601396715520 (see https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-runtime/3440/pull-controller-runtime-test/2018359601396715520/build-log.txt)

atharrva01: hi @sbueringer Thanks for the clarification request, this turned out to be a real concurrency issue in runnableGroup, not just a test artifact.

Start() was closing startReadyCh while readiness goroutines could still send on it, which caused a data race and send on closed channel panic depending on timing. Tests just made this easier to reproduce.

Removing the channel close exposed a second shutdown issue: Start() could block waiting for readiness when StopAndWait() cancelled the group context and all senders had already exited.

The fix now handles both sides explicitly: senders stop on r.ctx.Done(), and Start() exits readiness waiting on either ctx.Done() or r.ctx.Done(). This avoids races and deadlocks while preserving normal startup behavior.

Unit tests and go test -race pass locally; envtest-based integration tests should be covered by CI.

pkg/manager/runnable_group.go
sbueringer, line 211: I got why we can't close the channel in the other cases above, but I think we can still close it here

pkg/manager/runnable_group.go
```diff
@@ -245,7 +255,14 @@ func (r *runnableGroup) reconcile() {
...
 				if rn.Check(r.ctx) {
 					if rn.signalReady {
-						r.startReadyCh <- rn
+						// Use select to avoid sending after Start() has returned.
+						// When ctx is cancelled, Start() exits and is no longer
+						// receiving from startReadyCh. Without this select, we'd
+						// either block forever or panic if the channel were closed.
+						select {
+						case r.startReadyCh <- rn:
+						case <-r.ctx.Done():
```
sbueringer, line 259: I think this has to be the first case otherwise this will still send always on startReadyCh

pkg/manager/runnable_group_test.go
```diff
@@ -363,6 +363,54 @@ func (r leaderElectionAndWarmupRunnable) NeedLeaderElection() bool {
...
+	startReturned := make(chan struct{})
+	go func() {
+		defer close(startReturned)
+		_ = rg.Start(ctx)
+	}()
+
+	// Cancel the context
+	cancel()
+
+	// Start() should return promptly (within 100ms), not hang or spin
```
sbueringer: I think the 100ms are outdated

atharrva01: Just pushed the updates (select reorder + channel discussion).
PTAL when you get a chance, thanks!

sbueringer: Could it be that you overlooked this one: https://github.com/kubernetes-sigs/controller-runtime/pull/3440#discussion_r2781842889?

atharrva01: Hi @sbueringer 👋

Thanks for pointing this out.

I’ve now updated the implementation so that `startReadyCh` is explicitly closed in the success path (i.e., once all runnables have reported readiness and `len(r.startQueue) == 0`).

The two cancellation paths (`ctx.Done()` and `r.ctx.Done()`) still intentionally do not close the channel, since ownership remains with the success path and closing it there would risk violating the single-closer rule.

Additionally, I removed the stale timing reference in the test comment as suggested.

Please let me know if this addresses your concern.

pkoutsovasilis:
Hi 👋 I noticed this was closed by the @atharrva01 without a closing comment. AFAICT the missing return in the `ctx.Done()` branch still appears to be present on main. Was the issue addressed elsewhere, or would it make sense to reopen this PR and/or file a dedicated issue? Thanks 🙇

PR #3443, opened 2026-01-28
🐛 Fix fake client's SSA status patch resource version check

### LINKED ISSUE #3442: Fake client on version 0.23.1 fails to SSA Apply status due to resourceVersion checks
We might have found an issue with version 0.23.1's fake client implementation around the new SSA Apply support.

We noticed that doing an SSA on the status when the resource version does not match returns an error such as:

`"Operation cannot be fulfilled on ... object was modified"`

This is unexpected as the point of SSA, like pathing, is that it is a focused update that is applied as is. 

Steps to reproduce:

- Set up a fake client for unit testing with some object, e.g. a Pod.
- Ensure the resource version differs from the one about to be used.
- Attempt a SSA Apply on the Pod status.
Expected result: SSA Apply should work 

Actual result: SSA Apply fails with "object was modified" as if it was a regular update.

We will provide a PR with a test case and a draft fix for evaluation.

josvazg:
This change includes a test showing a sample SSA on a Pod status which failed before the fix with the "object was modified".

It also includes the fix in the fake client, which is copy the resource version from the old accessor as it was already done for non status patch and apply operations.

Advice on simplifying the reproducer is welcome.

Fixes #3442

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -1929,7 +1929,7 @@ var _ = Describe("Fake client", func() {
 
 		resourceAC := client.ApplyConfigurationFromUnstructured(resourceForApply)
 
-		// This is expected to fail with the wring rv value passed in in the applied config
+		// This is expected to fail with the wrong rv value passed in in the applied config
 		err := cl.Status().Apply(ctx, resourceAC, client.FieldOwner("test-owner"), client.ForceOwnership)
 		Expect(err).To(HaveOccurred(), "SSA apply on status should not succeed when resourceVersion is wrongly set")
 		Expect(apierrors.IsConflict(err)).To(BeTrue())


pkg/client/fake/client_test.go
```diff
@@ -1847,6 +1847,60 @@ var _ = Describe("Fake client", func() {
...
+		pod.Spec.Containers[0].Image = "updated-image"
+		Expect(cl.Update(ctx, pod)).To(Succeed())
+
+		// Get the updated pod to see the new resourceVersion
+		updatedPod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "test-pod", Namespace: "default"}}
+		Expect(cl.Get(ctx, client.ObjectKeyFromObject(updatedPod), updatedPod)).To(Succeed())
+		Expect(updatedPod.ResourceVersion).NotTo(Equal("1"))
+		originalRV := updatedPod.ResourceVersion
+
+		// Now try to do an SSA apply on the status with an explicitly mismatched resourceVersion.
```
alvaroaleman: I can not confirm the behavior described here against a real kube apiserver: If you pass an incorrect RV the request will fail (you can test this with `kubectl apply --server-side`). Only if you pass no RV (which I think is the common case for SSA) will it succeed regardless of server RV.

josvazg: I see, makes sense. Still the code from which we hit this was failing even after we were wiping the resource version clean right before the SSA, when in version 0.22 it was working fine without changes.

Let me double check.

josvazg: I need to dig deeper.

When testing this again removing the fix and rv value I get it to pass. But our original test still fails.

Comparing  `if accessor.GetResourceVersion() != oldAccessor.GetResourceVersion()` from my controller-runtime checkout (using this test modified) the comparison is `rv = "1" == old rv = "1"` so it passes. But from our code I saw `rv="" == old rv = "1000"`

Not sure what is the difference between both tests. Also not sure how the passing test is comparing "1 against 1" even though we wiped the rv before calling SSA.

josvazg: We think we found it:

```go
	if accessor.GetResourceVersion() == "" {
		switch {
		case allowsUnconditionalUpdate(gvk):
			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
			// This is needed because if the patch explicitly sets the RV to null, the client-go reaction we use
			// to apply it and whose output we process here will have it unset. It is not clear why the Kubernetes
			// apiserver accepts such a patch, but it does so we just copy that behavior.
			// Kubernetes apiserver behavior can be checked like this:
			// `kubectl patch configmap foo --patch '{"metadata":{"annotations":{"foo":"bar"},"resourceVersion":null}}' -v=9`
		case bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeClient).Patch")):
			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Patch" func.
			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
		case bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeClient).Apply")):
			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Apply" func.
			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
		}
	}
```

This POD test cannot reproduce the issue we hit because it is a core resource and follows the first case `case allowsUnconditionalUpdate(gvk)` which fixed the RV to match

Our resource is a CRD, and we are setting a subresource on it. So we are missing this case:
```go
case bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Apply")):
			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Patch" func.
			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
```

I shall revamp this PR:
1) Test must reproduce the issue using a CRD, not a core resource.
2) The old code fix goes away.
3) The new fix is just the new entry case in that are missing.

fedepaol:
@josvazg @alvaroaleman just got to the same conclusion :sweat_smile: with https://github.com/kubernetes-sigs/controller-runtime/pull/3444 

Closing it now as this was here before, let me know otherwise.

pkg/client/fake/client_test.go
alvaroaleman: Not all in-tree resources allow unconditional update, so you might find one there you can use for testing

josvazg: In the end I used an adhoc custom object. It took longer to verify because if the test registers an unstructured custom object, the issue is not reproduced. In that case, the `oldAccessor` is set with an empty **resourceVersion** which matches the one in the `accessor`.

Only when the custom object registered with a proper type, like in our case, you would  hit the issue. In that case, the `oldAccessor` is set with a non empty **resourceVersion** which does **not** match the one in the `accessor`.

This was tricky to figure out, I had to debug our failing test and the reproducer until I spotted the difference between both.

This PR is split into 2 commits, one with the test and one with the fix. If you run without the fix commit you can see the error we hit in our case.

pkg/client/fake/client_test.go
```diff
@@ -1847,6 +1847,57 @@ var _ = Describe("Fake client", func() {
 		Expect(cl.Status().Apply(ctx, node, client.FieldOwner("test-owner"))).To(Succeed())
 	})
 
+	It("should allow SSA apply on status without object has changed issues", func(ctx SpecContext) {
```
alvaroaleman, line 1850: Can we also add a test that validates that status apply requests with an invalid RV fail?

josvazg: @alvaroaleman I am confused. Maybe I misunderstood some details here.
I would assume that the RV value sent on an SSA will be ignored. In fact, would not the applied config that is what we pass the SSA operation have that already removed?

alvaroaleman: No, if you send an SSA request with an RV in it and it doesn't match, the apiserver fails the request: https://github.com/kubernetes-sigs/controller-runtime/pull/3443#discussion_r2737604405

josvazg: ah! OK makes sense.

Test added now

sbueringer:
/assign

(I'll review once Alvaro is fine with it)

pkg/client/fake/versioned_tracker.go
```diff
@@ -275,6 +275,14 @@ func (t versionedTracker) updateObject(
 			// We apply patches using a client-go reaction that ends up calling the trackers Update. As we can't change
 			// that reaction, we use the callstack to figure out if this originated from the "fakeClient.Apply" func.
 			accessor.SetResourceVersion(oldAccessor.GetResourceVersion())
+		case bytes.Contains(debug.Stack(), []byte("sigs.k8s.io/controller-runtime/pkg/client/fake.(*fakeSubResourceClient).Patch")):
```
alvaroaleman: Would you mind using a single case statement for Apply, subresource.Patch and subresource.Apply that ORs them? We don't really need the same comment three times

pkg/client/fake/client_test.go
```diff
@@ -1847,6 +1847,93 @@ var _ = Describe("Fake client", func() {
...
+		resourceForApply.SetAPIVersion("chaosapps.metamagical.io/v1")
+		resourceForApply.SetKind("ChaosPod")
+		resourceForApply.SetResourceVersion("bad-resource-version") // forces the error
+		Expect(unstructured.SetNestedField(resourceForApply.Object, map[string]any{"phase": "Ready"}, "status")).To(Succeed())
+
+		resourceAC := client.ApplyConfigurationFromUnstructured(resourceForApply)
+
+		// This is expected to fail with the wring rv value passed in in the applied config
+		err := cl.Status().Apply(ctx, resourceAC, client.FieldOwner("test-owner"), client.ForceOwnership)
+		Expect(err).To(HaveOccurred(), "SSA apply on status should not succeed when resourceVersion is wrongly set")
```
alvaroaleman, line 1934: Please check that we got the correct error here as well and not just any error:
```
Expect(apierrors.IsConflict(err)).To(BeTrue())
```

pkg/client/fake/versioned_tracker.go
josvazg: yeah it was getting too verbose now

josvazg: done

pkg/client/fake/client_test.go
josvazg: added

pkg/client/fake/client_test.go
```diff
@@ -1847,6 +1847,94 @@ var _ = Describe("Fake client", func() {
...
+				Name:      "test-chaospod",
+				Namespace: "default",
+			},
+			Spec: ChaosPodSpec{
+				Image: "test-image",
+			},
+		}
+		cl := NewClientBuilder().WithScheme(testScheme).WithStatusSubresource(customResource).WithObjects(customResource).Build()
+
+		// Create an unstructured apply configuration with status but no resourceVersion
```
sbueringer:
```suggestion
		// Create an unstructured apply configuration with status but wrong non empty resourceVersion
```

pkg/client/fake/client_test.go
```diff
@@ -1847,6 +1847,94 @@ var _ = Describe("Fake client", func() {
...
+		resourceForApply.SetName("test-chaospod")
+		resourceForApply.SetNamespace("default")
+		resourceForApply.SetAPIVersion("chaosapps.metamagical.io/v1")
+		resourceForApply.SetKind("ChaosPod")
+		resourceForApply.SetResourceVersion("bad-resource-version") // forces the error
+		Expect(unstructured.SetNestedField(resourceForApply.Object, map[string]any{"phase": "Ready"}, "status")).To(Succeed())
+
+		resourceAC := client.ApplyConfigurationFromUnstructured(resourceForApply)
+
+		// This is expected to fail with the wring rv value passed in in the applied config
```
sbueringer:
```suggestion
		// This is expected to fail with the wrong rv value passed in in the applied config
```
nit

sbueringer (commented):
Thank you very much. Makes sense, just two nits

sbueringer:
Thx again!

/lgtm
/approve
/hold cancel

PR #3445, opened 2026-01-30
:bug: Fix IndexField blocking until informer is synced

### LINKED ISSUE #3422: `IndexField()` unintentionally blocks until informer is synced
Hi, I'm writing a controller reconciled based on a CRD. I'd like to setup an `IndexField()` based on the Pod annotation. I found this function will block until the informer is synced, which delay the startup of the controller.

I dug into the cause, find that in `IndexField()`, it calls `GetInformer()` with default option `blockUntilSync=true`.

https://github.com/kubernetes-sigs/controller-runtime/blob/129853d4ae0548600432fb76cd4d965cfcb854a6/pkg/cache/informer_cache.go#L215-L221

https://github.com/kubernetes-sigs/controller-runtime/blob/129853d4ae0548600432fb76cd4d965cfcb854a6/pkg/cache/internal/informers.go#L123-L128

I'd like to know
1. Is this an expected behavior?
2. Where should I put the `IndexField()`, taking [controller-runtime/examples/crd/main.go](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.0/examples/crd/main.go) as an example?

It's NOT an urgent issue, as the delayed startup won't cause too much trouble. Thanks!

gzb1128:
## What

`IndexField()` was blocking until the informer cache synced, causing unnecessary delays during controller startup.

## Why

The issue occurs because `GetInformer()` defaults to `blockUntilSync=true`. However, adding an index doesn't require the informer to be synced - the underlying `client-go threadSafeStore.AddIndexers` doesn't depend on sync state.

Only `List` operations with field selectors need the cache to be synced. Controllers already wait for their source informers to sync during startup via `WaitForCacheSync`.

## Fixes

Fixes #3422

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -2657,3 +2657,43 @@ func TestReaderWaitsForCacheSync(t *testing.T) {
 		})
 	}
 }
+
+func TestIndexFieldDoesNotBlock(t *testing.T) {
+	t.Parallel()
+	synctest.Test(t, func(t *testing.T) {
+		g := NewWithT(t)
+
+		fakeInformer := &controllertest.FakeInformer{Synced: false}
+		c, err := cache.New(&rest.Config{}, cache.Options{
+			Mapper: &fakeRESTMapper{},
+			NewInformer: func(kcache.ListerWatcher, runtime.Object, time.Duration, kcache.Indexers) kcache.SharedIndexInformer {
+				return fakeInformer
+			},
+		})
+		g.Expect(err).NotTo(HaveOccurred())
+
+		ctx, cancel := context.WithCancel(t.Context())
+		defer cancel()
+		cacheDone := make(chan struct{})
+		go func() {
+			g.Expect(c.Start(ctx)).To(Succeed())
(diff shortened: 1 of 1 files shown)


alvaroaleman:
/ok-to-test

Please sign the CLA and please add a test

alvaroaleman:
> Please sign the CLA and please add a test

Sorry, you already have, the issue is that the PR title must be updated

gzb1128:
Hi @alvaroaleman, I've added a unit test to verify that IndexField returns immediately when called before the informer cache is synced.

The test:
- Uses FakeInformer with Synced=false to simulate an unsynced informer
- Uses a 100ms timeout context - if IndexField blocks waiting for sync, it will timeout
- Follows the same pattern as TestReaderWaitsForCacheSync
- Can be run locally without etcd

I've verified that removing the fix (using BlockUntilSynced(false)) causes the test to fail with a clear timeout error, confirming the test correctly validates the non-blocking behavior.

sbueringer:
Thx!

/lgtm

PR #3447, opened 2026-02-02
:bug: ensure envtest stops the whole process group

dnwe:
Previous PR #2560 attempted to prevent envtest from leaking orphaned etcd / kube-apiserver etc. processes by adding a SysProcAttr of `Setpgid: true`. However, it made a few mistakes:

- It incorrectly described the change as causing the process to "inherit the parent's process group id", which is the opposite of what it does. In fact it causes a brand new process group to be created for the child process, which is the actual behaviour we want, so correct the comment.
- Whilst it added a new process group, it didn't update the signalling to send SIGTERM (and eventually SIGKILL) to the process group (by using a `-` prefix) and was still just sending the signal to the process itself [1].
- It didn't add a unittest to assert on the desired behaviour before and after the changes.

While with this PR the whole process group is stopped when stopping envtest, this doesn't really result in behavioral changes because the processsgroup consists of one process in all circumstances (either kas or etcd).

[1] https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773

### DIFF
--- a/pkg/internal/testing/process/process.go
+++ b/pkg/internal/testing/process/process.go
@@ -193,7 +193,7 @@ func (ps *State) Start(stdout, stderr io.Writer) (err error) {
 		close(pollerStopCh)
 		if ps.Cmd != nil {
 			// intentionally ignore this -- we might've crashed, failed to start, etc
-			ps.Cmd.Process.Signal(syscall.SIGTERM) //nolint:errcheck
+			_ = signalProcess(ps.Cmd.Process, syscall.SIGTERM)
 		}
 		return fmt.Errorf("timeout waiting for process %s to start", path.Base(ps.Path))
 	}
@@ -256,7 +256,7 @@ func (ps *State) Stop() error {
 	if done, _ := ps.Exited(); done {
 		return nil
 	}
-	if err := ps.Cmd.Process.Signal(syscall.SIGTERM); err != nil {
+	if err := signalProcess(ps.Cmd.Process, syscall.SIGTERM); err != nil {
 		return fmt.Errorf("unable to signal for process %s to stop: %w", ps.Path, err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/internal/testing/process/procattr_unix.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // GetSysProcAttr returns the SysProcAttr to use for the process,
 // for unix systems this returns a SysProcAttr with Setpgid set to true,
-// which inherits the parent's process group id.
+// which creates a new process group for the child process.
```
alvaroaleman, line 27: While your updated description is correctly describing what already happened, I am not sure I understand why that actually matters. Unless I am missing something, we directly start the apiserver and etcd binary without an intermediate like a shell so it shouldn't make a difference if we send a signal to them directly or to the processgroup?

I've seen the stale etcd and apiserver processes too, but my theory would be that happens because the test dies and `Stop()` doesn't run, not because the cleanup is buggy. But maybe I am missing something here?

dnwe: Ah yes sadly you're right I think that this won't help the etcd/apiserver case. I hadn't actually checked how they were being started, I'd just spotted the leaking processes and then found #2560 which looked a bit wrong.

As you say, it seems like in pkg/internal/testing/controlplane/plane.go we just start them one after the other as singleton processes. So I think the PR is still right, in that it corrects the behaviour when a group of processes is spawned, but it's more of a housekeeping thing rather than actually improving the enduser experience when running envtest 😔

sbueringer: Would it make sense to slightly change the PR description (and PR title) to surface that this won't change anything for envtest?

PR #3449, opened 2026-02-06
:seedling: Use forked version of btree

alvaroaleman:
Following the upstream change in https://github.com/kubernetes/kubernetes/issues/136415


alvaroaleman:
This has to wait until we bump the k8s.io/* deps to a version that has the k8s.io/utils change

skitt:
> This has to wait until we bump the k8s.io/* deps to a version that has the k8s.io/utils change

Couldn’t the k8s.io/utils dependency be bumped separately? The changes there are done carefully enough that there shouldn’t be any risk for controller-runtime or its dependants.

alvaroaleman:
@skitt we try to keep the versions of the k8s.io/ deps in sync with what is in k/k. We can setup exceptions for that if we have a good reason, but in this case I think it's fine to just wait for the next release

sbueringer:
lgtm/approve pending rebase (alpha.1 bump PR is now merged)

PR #3451, opened 2026-02-11
🌱 Switch to `StartLogging` for event debug logs

### LINKED ISSUE #3407: Use contextual logging for events
Errors of the event recorder are shown in the wrong format. Example:
```
E0811 00:00:00.000000       1 event.go:359] "Server rejected event (will not retry!)" err="Unauthorized" event="&Event{ObjectMeta:{...
```
The event recorder was extended in client-go to support contextual logging (change: https://github.com/kubernetes/client-go/commit/3595e5242a2c3e8d7b114b32361b2226f173f40e#diff-401f00d69d10cd2fce62a75c1ce524c04c978d646e13f532f0a64b8149efcfe1)

To implement structured logging in controller-runtime, a context with the logger must be passed to the broadcaster via `StartRecordingToSinkWithContext`. At the moment, `context.Background()` is used: https://github.com/kubernetes-sigs/controller-runtime/blob/adb64659bb1fd4cf84e59f24773eb964fd1d00e1/pkg/internal/recorder/recorder.go#L120-L125

To inject the logger, `log.IntoContext()` can be used:
https://github.com/kubernetes-sigs/controller-runtime/blob/adb64659bb1fd4cf84e59f24773eb964fd1d00e1/pkg/log/log.go#L103

In addition, it could be an option to use `StartLogging` (doc: https://pkg.go.dev/k8s.io/client-go@v0.35.0/tools/events#EventBroadcaster, implementation: https://github.com/kubernetes/client-go/blob/9bcb69436287b966d0c5c195efef00aed921fb1b/tools/events/event_broadcaster.go#L330-L340) instead of `StartEventWatcher` to align the log messages for events with the implementation in client-go: https://github.com/kubernetes-sigs/controller-runtime/blob/7a1b16d039d204e7e4c5c8df57124d4e060a55f0/pkg/internal/recorder/recorder.go#L102-L105

This issue is related to https://github.com/kubernetes-sigs/controller-runtime/issues/2656, but #2656 mainly discusses the adoption of structured logging for the leader election package.

If you agree to the proposed changs, I can create a PR.

timuthy:
Switches from an event watcher to `StartLogging` to log events, as proposed by #3407.

Fixes #3407

### DIFF
--- a/pkg/internal/recorder/recorder.go
+++ b/pkg/internal/recorder/recorder.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
@@ -54,7 +53,7 @@ type Provider struct {
 	broadcasterOnce         sync.Once
 	broadcaster             events.EventBroadcaster
 	cancelSinkRecordingFunc context.CancelFunc
-	stopWatcherFunc         func()
+	stopLoggingFunc         func()
 	// Deprecated: will be removed in a future release. Use the broadcaster above instead.
 	deprecatedBroadcaster record.EventBroadcaster
 	stopBroadcaster       bool
@@ -83,7 +82,7 @@ func (p *Provider) Stop(shutdownCtx context.Context) {
 			p.lock.Lock()
 			broadcaster.Shutdown()
 			p.cancelSinkRecordingFunc()
-			p.stopWatcherFunc()
+			p.stopLoggingFunc()
 			deprecatedBroadcaster.Shutdown()
 			p.stopped = true
 			p.lock.Unlock()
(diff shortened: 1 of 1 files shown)


pkg/internal/recorder/recorder.go
```diff
@@ -127,7 +127,7 @@ func (p *Provider) getBroadcaster() (record.EventBroadcaster, events.EventBroadc
 		stopWatcher, err := p.broadcaster.StartEventWatcher(func(event runtime.Object) {
 			e, isEvt := event.(*eventsv1.Event)
 			if isEvt {
-				p.logger.V(1).Info(e.Note, "type", e.Type, "object", e.Related, "action", e.Action, "reason", e.Reason)
+				p.logger.V(1).Info(e.Note, "type", e.Type, "object", e.Regarding, "action", e.Action, "reason", e.Reason)
```
sbueringer: Maybe we should do what was suggested in this issue https://github.com/kubernetes-sigs/controller-runtime/issues/3407 instead?

timuthy: Nice, I wasn't aware of #3407 - thanks for the feedback.

OTOH, what do you think about merging this small change until a proper PR for #3407 is available? It shouldn't cause any harm.

sbueringer: While logs are not really a stable API I would like to avoid changing the log format twice for folks that rely on the format.

I think we can simply implement half of #3407 here without much effort:
```go
		stopWatcher, err := p.broadcaster.StartLogging(p.logger)
		if err != nil {
			p.logger.Error(err, "error starting event watcher for broadcaster")
		}

		p.stopWatcherFunc = stopWatcher
```

(Let's also rename stopWatcher and p.stopWatcherFunc to stopLogging and p.stopLoggingFunc)

sbueringer:
Thank you!

/lgtm

/assign @alvaroaleman

PR #3452, opened 2026-02-12
:bug: Wire up base context into http servers

alvaroaleman:
The manager already has a configurable base context, but it doesn't pass that on onto http servers. As a result, configuring it does nothing for example webhooks.


sbueringer:
Thx!

/lgtm
/approve

PR #3457, opened 2026-02-15
:seedling: Deflake certwatcher tests

alvaroaleman:
In the certwatcher tests, we use `start_watcher` which starts the certwatcher in a goroutine. Once it is started, it sets up an fsnotify and reacts to events.

The metric test rely both on the gorouine starting fast enough and the fsnotify events being processed fast enough through eventually. In CI, the four second timeout that was used isn't always sufficient, resulting in flakes.

Given that there are dedicated tests for inotify event handling, this change changes the metric tests to explicitly call `ReadCertificate` and then validating the metrics.

### DIFF
--- a/pkg/certwatcher/certwatcher_test.go
+++ b/pkg/certwatcher/certwatcher_test.go
@@ -24,7 +24,6 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
-	"fmt"
 	"math/big"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/certwatcher/certwatcher_test.go
```diff
@@ -201,54 +200,18 @@ var _ = Describe("CertWatcher", func() {
...
-					readCertificateTotalAfter := testutil.ToFloat64(metrics.ReadCertificateTotal)
-					if readCertificateTotalAfter < readCertificateTotalBefore+1.0 {
-						return fmt.Errorf("metric read certificate total expected at least: %v and got: %v", readCertificateTotalBefore+1.0, readCertificateTotalAfter)
-					}
-					return nil
-				}, "4s").Should(Succeed())
-
-				ctxCancel()
-				Eventually(doneCh, "4s").Should(BeClosed())
+				Expect(testutil.ToFloat64(metrics.ReadCertificateTotal)).To(BeNumerically(">=", readCertificateTotalBefore+1))
```
sbueringer: Should this still be >= or equals now?

(same below)

alvaroaleman: Yes of course, done

sbueringer:
Thx!

/lgtm
/approve

PR #3459, opened 2026-02-17
:sparkles: only log stacktraces >= DPanic when using the Zap adapter

xrstf:
Ever since at least #86, the default was to create stacktraces even for _error_ messages.

To me this makes absolutely no sense. Errors can arise from many situations, like a NotFound error when reconciling, or invalid objects that cannot be processed. None of these situations in my opinion warrant a stacktrace being printed, since none of them indicate a bug in the software.

On the other hand, scrolling through an app's logs and seeing stacktraces makes my spider sense tingle a bit, only to then realize "oh, yeah, it's just a NotFound error, false alarm everyone!".

I suggest the default for printing stacktraces should be much higher, hence this PR to raise the default. Personally I also do not really see a value in differentiating between development and production modes. I cannot imagine any real usecases where a developer would expect and want to see a stacktrace when they print a warning message.

### DIFF
--- a/pkg/log/zap/zap.go
+++ b/pkg/log/zap/zap.go
@@ -155,8 +155,7 @@ type Options struct {
 	// See https://pkg.go.dev/github.com/go-logr/zapr for how zap level relates to logr verbosity.
 	Level zapcore.LevelEnabler
 	// StacktraceLevel is the level at and above which stacktraces will
-	// be recorded for all messages. Defaults to Warn when Development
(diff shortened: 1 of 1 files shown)


xrstf:
I am struggeling to reproduce this error locally.

/retest

[EDIT: Seems to be a flake... weird. I cannot really imagine how changing this default level could influence the test suite.]

xrstf:
Fair enough 😁

PR #3461, opened 2026-02-25
:seedling: Deprecate the scheme builder

alvaroaleman:
Controller-Runtime should never be imported into api packages and the scheme builder is only useful in those, so deprecate it.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -49,7 +49,36 @@ type Manager = manager.Manager
 type Options = manager.Options
 
 // SchemeBuilder builds a new Scheme for mapping go types to Kubernetes GroupVersionKinds.
-type SchemeBuilder = scheme.Builder
+//
+// Deprecated: This helper is only useful in api packages, but api packages should be
+// easy to import and hence have minimal dependencies. Typically, these dependencies
+// include only the standard library, k8s.io/apimachinery and other api packages.
+//
+// Use the apimachinery builder instead:
+//
+//	import (
+//		metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+//		"k8s.io/apimachinery/pkg/runtime"
+//		"k8s.io/apimachinery/pkg/runtime/schema"
+//	)
+//
+//	const GroupName = ""
+//
+//	var (
+//		SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
+//		SchemeBuilder 		 = runtime.NewSchemeBuilder(addKnownTypes)
+//		AddToScheme   		 = SchemeBuilder.AddToScheme
+//	)
+//
+//	func addKnownTypes(scheme *runtime.Scheme) error {
+//		scheme.AddKnownTypes(SchemeGroupVersion,
+//			&Pod{},
+//		)
+//
+//		metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
+//		return nil
+//	}
+type SchemeBuilder = scheme.Builder //nolint:staticcheck // this is the deprecation alias
 
 // GroupVersion contains the "group" and the "version", which uniquely identifies the API.
 type GroupVersion = schema.GroupVersion
--- a/examples/crd/pkg/groupversion_info.go
+++ b/examples/crd/pkg/groupversion_info.go
@@ -20,8 +20,9 @@ limitations under the License.
 package pkg
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
 var (
@@ -29,8 +30,17 @@ var (
 	SchemeGroupVersion = schema.GroupVersion{Group: "chaosapps.metamagical.io", Version: "v1"}
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
-	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
 
 	// AddToScheme is required by pkg/client/...
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(SchemeGroupVersion,
+		&ChaosPod{},
+		&ChaosPodList{},
+	)
+	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
+	return nil
+}
--- a/pkg/scheme/scheme.go
+++ b/pkg/scheme/scheme.go
@@ -60,6 +60,35 @@ import (
 )
 
 // Builder builds a new Scheme for mapping go types to Kubernetes GroupVersionKinds.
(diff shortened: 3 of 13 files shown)


alias.go
```diff
@@ -49,7 +49,34 @@ type Manager = manager.Manager
 type Options = manager.Options
 
 // SchemeBuilder builds a new Scheme for mapping go types to Kubernetes GroupVersionKinds.
-type SchemeBuilder = scheme.Builder
+//
+// Deprecated: Controller-Runtime should not be imported in api packages and this
```
alvaroaleman: We could also word this stronger and mention that api packages should not import anything other than other api packages and apimachinery

troy0820:
/lgtm 

I know this isn't the removal yet, but we may want to allow @camilamacedo86  to be aware of the changes if they wish to get in front of this with kubebuilder.  

/hold

pkg/scheme/scheme.go
```diff
@@ -60,6 +60,33 @@ import (
...
+//
+//	import (
+//		metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+//		"k8s.io/apimachinery/pkg/runtime"
+//		"k8s.io/apimachinery/pkg/runtime/schema"
+//	)
+//
+//	const GroupName = ""
+//
+//	var SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
```
sbueringer: nit: combine with the var block below (same in alias.go)

alias.go
sbueringer: Fine for me. Or we phrase it more generically and write that folks should be extremely careful with any dependency they add to their API (packages).

E.g. having dependencies on corev1.ObjectReference is an extremely bad idea. Similarly embeddeding any external API types into an API needs careful consideration as it couples APIs (somewhat fine with k/k v1 APIs, but even then there are non-trivial trade-offs in my opinion like inheriting all changes that are made to these APIs elsewhere)


@JoelSpeed Do you have some input for how we should phrase this?
I think this is deprecation warning is a good place to share some guidance. I wouldn't go into too much detail but some high-level points could be good here

examples/crd/pkg/groupversion_info.go
```diff
@@ -20,17 +20,27 @@ limitations under the License.
...
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
-	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
 
 	// AddToScheme is required by pkg/client/...
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+func addKnownTypes(scheme *runtime.Scheme) error {
```
sbueringer, line 39: Not sure if this is the ideal pattern.

In CAPI we have an array here
```
	objectTypes = []runtime.Object{}
```

init in the files for the corresponding API types
```
func init() {
	objectTypes = append(objectTypes, &MachineDeployment{}, &MachineDeploymentList{})
}
```

And addKnownTypes looks like this
```

func addKnownTypes(scheme *runtime.Scheme) error {
	scheme.AddKnownTypes(GroupVersion, objectTypes...)
	metav1.AddToGroupVersion(scheme, GroupVersion)
	return nil
}
```

I think it's better because it allows to add the types in the files where the API types are defined. So it works better with copy&paste & generation workflows

(Secondary, but it's also the smaller change for migrations to the new pattern)

JoelSpeed: I think this is a good pattern for this, so I'm in favour

alias.go
JoelSpeed: > Or we phrase it more generically and write that folks should be extremely careful with any dependency they add to their API (packages).

This sounds like a good way to frame it 

I would target this message to APIs should import minimal dependencies. Maybe something along the lines of:

```
// API Packages should import minimal dependencies, typically including only k8s.io/apimachinery sub-packages and standard library packages.
// Importing additional packages affects consumers of your API and adds complexity to their dependency management.
```

examples/crd/pkg/groupversion_info.go
alvaroaleman: I did it this way because the apis in k/k do it the same way.

> I think it's better because it allows to add the types in the files where the API types are defined. So it works better with copy&paste & generation workflows

I think whoever adds an api has to remember that they have to add something, I don't think at that point it matters much in what file they have to do it and the copy paste example is only helped if there is a file per api definition? I don't really think one is clearly better than the other here so I tend to want to default to what is done in the k/k apis

alias.go
alvaroaleman: I've reworded it mostly based on Joels suggestion with the addition that importing other API packages is also fine, let me know what you think

pkg/scheme/scheme.go
alvaroaleman: Done

examples/crd/pkg/groupversion_info.go
sbueringer: No strong opinion from my side. I personally prefer to keep the "schema registration" close to the type

alias.go
sbueringer: The way you phrased it sounds good to me.

I had my share of problems with importing API packages in API packages so now I'm extremely careful with those.

Problems there can be:
* Importing types that don't actually fit the use case, example metav1.ObjectReference (that one even has a godoc comment that it should not be used)
* Inheriting changes of the imported types including schema changes
* Making it necessary to cut a new apiVersion because the referenced apiVersion was bumped to avoid breaking changes (e.g. if a v1beta1 type is embedded into another API and then that type is bumped to v1beta2) 

So at that point I only consider it safe to embed external API types if:
* the lifecycles align (e.g. some k/k v1 type or something that is bumped in lock step with the API the type is embedded in)
* the use case perfectly fits (that means the fields, godoc and schema validation perfectly fits any one of those might diverge already or might diverge over time)

All that being said, I'm fine with the current wording. We don't have to be too detailed here :) (just sharing some context)

alvaroaleman: I think that comment really just tries to explain what is okay to have as a dependency. If it actually makes sense to depend on other apis is yet another question that IMHO we shouldn't try to answer here.

JoelSpeed: From an API review perspective, we only generally allow importing other API packages if you are "passing through" the data. E.g. you have a type with an embedded pod spec and you then just copy that directly onto another type

Stefan's write up is detailed enough that I don't need to add anymore, but I am fine with the current wording

pkg/scheme/scheme.go
```diff
@@ -60,6 +60,35 @@ import (
 )
 
 // Builder builds a new Scheme for mapping go types to Kubernetes GroupVersionKinds.
+//
+// Deprecated: This helper is only useful in api packages, but api packages should be
+// easy to import and hence have minimal dependencies. Typically, these dependencies
+// include only the standard library, k8s.io/apimachinery and other api packages.
+//
+// Use the apimachinery builder instead:
```
sbueringer, line 68: this line and the next one is missing in alias.go. Maybe let's add it there as well

alvaroaleman: Good catch, added it

sbueringer:
Thx!

/lgtm
/approve

Keeping the hold so @JoelSpeed can take another look

sbueringer:
I think we're fine, let's move on

/hold cancel

PR #3462, opened 2026-02-27
:warning: Bump to k8s.io/* v0.36.0-alpha.2 & implement HasSyncedChecker()

sbueringer:


sbueringer:
/hold

I have to take a closer look and was running out of time today. I thought it's just a quick bump and  a small change to the fakeInformer.

sbueringer:
/hold cancel

/assign @alvaroaleman 

Should be ready now

pkg/cache/cache_test.go
```diff
@@ -2020,7 +2039,17 @@ func CacheTest(createCacheFunc func(config *rest.Config, opts cache.Options) (ca
...
+					handlerRegistration, err := sii.AddEventHandler(kcache.ResourceEventHandlerFuncs{AddFunc: addFunc})
+					Expect(err).ToNot(HaveOccurred())
+					// Drain the channel because there are many objects that all have to be processed for the handler to be synced.
+					// Using Eventually with an interval here to give the events some time to arrive
+					Eventually(func(g Gomega) {
+						select {
+						case <-out:
+						default:
+						}
+						g.Expect(handlerRegistration.HasSyncedChecker().Done()).Should(BeClosed())
```
alvaroaleman: Its not possible to wait for a channel close or timeout without polling in gomega?

pkg/cache/multi_namespace_cache.go
```diff
@@ -325,17 +322,47 @@ func (c *multiNamespaceCache) List(ctx context.Context, list client.ObjectList,
...
+	return mni
+}
+
 // multiNamespaceInformer knows how to handle interacting with the underlying informer across multiple namespaces.
 type multiNamespaceInformer struct {
+	synced              chan struct{}
 	namespaceToInformer map[string]Informer
 }
 
+func newHandlerRegistration(handles map[string]toolscache.ResourceEventHandlerRegistration) toolscache.ResourceEventHandlerRegistration {
```
alvaroaleman: Nit: Maybe rename to clarify that this is specifically only meant for the multi-namespace cache, since everything is in the same package?

pkg/controller/controllertest/util.go
```diff
@@ -30,8 +30,9 @@ var _ cache.SharedIndexInformer = &FakeInformer{}
 // FakeInformer provides fake Informer functionality for testing.
 type FakeInformer struct {
 	// Synced is returned by the HasSynced functions to implement the Informer interface
-	Synced     bool
-	SyncedLock sync.Mutex
+	Synced        bool
+	SyncedChecker cache.DoneChecker
```
alvaroaleman: Should we really make this entire thing configurable as opposed to implemented it based on `Synced`? This package is public so changing this later on would be a breaking change

sbueringer: The question is how I can react on someone directly setting the Synced field to true (like we do today)

I see two options:
* unexporting synced and introducing a setter (setter then also closes the synced channel)
* Introducing a go routine that keeps checking Synced that closes the channel once Synced is true 

I'm fine with both. First option sounds better to me, it's a breaking change but I'm fine with it

pkg/cache/cache_test.go
sbueringer: I have to continue emptying the channel while checking for closed.
(A simple for / select with default doesn't work here to empty the channel)

Not all events are arriving immediately in the channel. Not sure how else I can do that

sbueringer: Maybe I have another idea, I'll give it a try (for / select with read from channel, check for closed, timeout, with no default)

sbueringer: Works

pkg/controller/controllertest/util.go
alvaroaleman: Good point and yes, lets go with the setter 👍

sbueringer: :/ This is even worse

We don't have a constructor, so no good place to make the channel.

Should I create a constructor and unexport the type?
Not sure if there is a good alternative to that.

I think it's not a good option to ask folks to initialize the struct with a channel.
I think it's also bad if I create the channel on demand in the setter of synced (what if someone calls the HasSyncedChecker method before that)

*Maybe* it could be okay to make the channel as part of Run? I'm not sure, feels brittle

alvaroaleman: > Should I create a constructor and unexport the type?
> Not sure if there is a good alternative to that.

How would we make the setter available if the type is unexported?

> Maybe it could be okay to make the channel as part of Run? I'm not sure, feels brittle

Makes sense to me insofar as that checking on the synced state before calling run would be wrong anyways, as an informer will never sync if it wasn't started

sbueringer: > How would we make the setter available if the type is unexported?

I guess exporting an interface, or is there another way to do this? (my goal is just to avoid incorrect instantiations of the type, apart from that I don't care about if the type is exported or not)

This would roughly align to the sharedIndexInformer.

alvaroaleman: > I guess exporting an interface, or is there another way to do this? (my goal is just to avoid incorrect instantiations of the type, apart from that I don't care about if the type is exported or not)

IMHO its okay to leave the type exported but say it must be constructructed through a constructor

sbueringer: Alright, then let's do that

sbueringer: Should be fine now

pkg/controller/controllertest/util.go
```diff
@@ -28,28 +27,49 @@ import (
...
+	// This field is used in various HasSynced and HasSyncedChecker methods.
+	synced chan struct{}
 
 	// RunCount is incremented each time RunInformersAndControllers is called
 	RunCount int
 
 	handlers []cache.ResourceEventHandler
 }
 
+func NewFakeInformer(synced bool) *FakeInformer {
```
alvaroaleman: Can we maybe use an options struct or variadic args so that we don't have to break the signature if we change the parameters?

sbueringer: Done

PR #3463, opened 2026-02-28
✨ Reduce memory usage of default webhooks

sbueringer:
Before this PR we were always going through the JSON patch calculation which is pretty expensive.
This PR optimizes this by short-circuting if the objects are equal.

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/controller-runtime/pkg/webhook/admission/bench
cpu: Apple M2 Pro
                          │   old.txt   │              new.txt               │
                          │   sec/op    │   sec/op     vs base               │
Defaulter/noop-12           95.52µ ± 7%   43.50µ ± 2%  -54.46% (p=0.002 n=6)
Defaulter/with_changes-12   89.98µ ± 2%   96.58µ ± 6%   +7.33% (p=0.002 n=6)
geomean                     92.71µ        64.82µ       -30.09%

                          │   old.txt    │               new.txt               │
                          │     B/op     │     B/op      vs base               │
Defaulter/noop-12           52.66Ki ± 0%   22.19Ki ± 0%  -57.86% (p=0.002 n=6)
Defaulter/with_changes-12   52.26Ki ± 0%   55.00Ki ± 0%   +5.25% (p=0.002 n=6)
geomean                     52.46Ki        34.94Ki       -33.40%

                          │  old.txt   │              new.txt              │
                          │ allocs/op  │ allocs/op   vs base               │
Defaulter/noop-12           950.0 ± 0%   217.0 ± 0%  -77.16% (p=0.002 n=6)
Defaulter/with_changes-12   950.0 ± 0%   982.0 ± 0%   +3.37% (p=0.002 n=6)
geomean                     950.0        461.6       -51.41%
```

### DIFF
--- a/pkg/webhook/admission/defaulter_custom_test.go
+++ b/pkg/webhook/admission/defaulter_custom_test.go
@@ -15,17 +15,24 @@ limitations under the License.
 package admission
 
 import (
+	"bytes"
 	"context"
 	"maps"
 	"net/http"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"gomodules.xyz/jsonpatch/v2"
-
 	admissionv1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 var _ = Describe("Defaulter Handler", func() {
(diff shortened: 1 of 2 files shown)


pkg/webhook/admission/defaulter_custom_test.go
```diff
@@ -167,3 +174,167 @@ func (d *TestCustomDefaulter) Default(ctx context.Context, obj runtime.Object) e
...
+		nil,
+		nil,
+		json.SerializerOptions{Yaml: false, Pretty: false, Strict: false},
+	)
+
+	// 2. Use a Buffer for efficiency
+	buf := new(bytes.Buffer)
+	err := s.Encode(obj, buf)
+	if err != nil {
+		return nil, err
```
alvaroaleman: Given that we just ignore the error at the callsite, can we remove the err return and just panic here (and maybe name `mustEncodeLikeAPIServer`)?

pkg/webhook/admission/defaulter_custom_test.go
```diff
@@ -167,3 +174,167 @@ func (d *TestCustomDefaulter) Default(ctx context.Context, obj runtime.Object) e
...
+		for i := 0; i < b.N; i++ {
+			_ = handler.Handle(b.Context(), req)
+		}
+	})
+}
+
+type PodDefaulter struct{}
+
+func (p PodDefaulter) Default(_ context.Context, pod *corev1.Pod) error {
+	pod.Spec.Hostname = "test-hostname"
```
alvaroaleman: Nit: Maybe use a const for `test-hostname` to make it more obvious this has to be the same here as in the no-op test?

pkg/webhook/admission/defaulter_custom_test.go
sbueringer: Sounds good

pkg/webhook/admission/defaulter_custom_test.go
sbueringer: I made it configurable instead and moved it very close to where we specify the hostname for the Pod

Issue #3320, opened 2025-09-15
Idea: Provide read your own write for the cache-backed client


alvaroaleman:
Upstream is discussing to mark the RV offficially as comparable: https://github.com/kubernetes/enhancements/pull/5505
If/When that happens, we could extend the cache-backed client to provide read your own write semantics, for example by:

* Blocking in mutating operations until the RV in the cache is >= response RV
* Blocking all reading operations in the cache until it it got to >= response RV

sbueringer:
In general makes sense to me.

I think we should definitely have an option to turn it off though for better performance.

> Blocking in mutating operations until the RV in the cache is >= response RV

I think this one absolutely makes sense

> Blocking all reading operations in the cache until it it got to >= response RV

I"m not so sure about this one, I assume this would essentially block all reads for a kind if there is a single write for that kind.
I think this could slow down controllers considerably. Especially if there are a lot of reconcilers that are using the same objects.

alvaroaleman:
Delete will be tricky due to https://github.com/kubernetes-sigs/controller-runtime/issues/3059

sbueringer:
Not sure I follow. Can't we wait until the object is gone from the cache?

(just have to be careful for the case where another object with the same name is created right after deletion)

EDIT: Ah, do you mean the case where a finalizer is set or not? I would hope we can handle both cases
1. => wait until object is gone or
2. => wait until resourceVersion is high enough

alvaroaleman:
Yeah, because of finalizers. The problem is that since we don't deserialize the object back if it isn't gone from storage, we don't know the post-delete RV, so 2) in your list doesn't have an obvious solution

sbueringer:
Got it. If we can't come up with a good user-facing API for Delete we could overwrite Delete in the cached client and handle the response differently there?

serathius:
Hey everyone, I look at KEP-5505 in a little different perspective, would love to hear your opinion on this. My background is in etcd and API server cache performance/correctness, so I'm newer to the intricacies of controller-runtime.

[...]

PR #3472, opened 2026-03-08
:sparkles: Add a read-your-own-write client

alvaroaleman:
Related to https://github.com/kubernetes-sigs/controller-runtime/issues/3320
/hold


pkg/client/consistency.go
```diff
@@ -0,0 +1,382 @@
...
+	return c.upstream.Get(ctx, key, obj, opts...)
+}
+
+func (c *consistentClient) List(ctx context.Context, list ObjectList, opts ...ListOption) error {
+	gvk, err := apiutil.GVKForObject(list, c.upstream.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to get GVK for list %T: %w", list, err)
+	}
+
+	keys := c.lockedKeysByGVK.getOrCreate(gvk).allValues()
```
halfcrazy: gvk.Kind = strings.TrimSuffix(gvk.Kind, "List")

pkg/cache/multi_namespace_cache.go
```diff
@@ -223,6 +223,42 @@ func (c *multiNamespaceCache) IndexField(ctx context.Context, obj client.Object,
...
+		global.SetMinimumRVForGVKAndKey(gvk, key, rv)
+	}
+}
+
+func (c *multiNamespaceCache) AddRequiredDeleteForObject(obj client.Object) error {
+	if ns := obj.GetNamespace(); ns == "" && c.clusterCache != nil {
+		return c.clusterCache.AddRequiredDeleteForObject(obj)
+	} else if cache, ok := c.namespaceToCache[ns]; ok {
+		return cache.AddRequiredDeleteForObject(obj)
+	}
```
halfcrazy: Looks like not handle global cache `namespaceToCache[metav1.NamespaceAll]`

pkg/cache/internal/readerconsistency/readerconsistency.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	allDeleted := make(chan struct{})
+	go func() {
+		h.pendingDeletesCond.L.Lock()
+		for !h.allDeletedLocked(uids) {
+			if ctx.Err() != nil {
+				break
+			}
+			h.pendingDeletesCond.Wait()
```
halfcrazy: Since `Wait()` only wakes on `Broadcast()`, and `Broadcast()` only happens when new informer events arrive, the goroutine will leak

pkg/cache/internal/readerconsistency/readerconsistency.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (h *ConsistencyHandler) waitForRV(ctx context.Context, rv int64) error {
+	observed := make(chan struct{})
+	go func() {
+		h.rvCond.L.Lock()
+		for h.observedRV < rv {
+			if ctx.Err() != nil {
+				break
+			}
+			h.rvCond.Wait()
```
halfcrazy: Since `Wait()` only wakes on `Broadcast()`, and `Broadcast()` only happens when new informer events arrive, the goroutine will leak

pkg/cache/internal/readerconsistency/readerconsistency.go
alvaroaleman: Yes I am aware, but I think that's acceptable - If someone started waiting, the assumption is that an event will come so this won't be blocked for long

pkg/client/consistency_envtest_test.go
```diff
@@ -0,0 +1,257 @@
+package client_test
```
sbueringer, line 17: Please double check (eventually, I know this is just a POC for now) that we have coverage that read-your-own write also works correctly with the multi namespace cache

pkg/client/consistency.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+var _ Client = (*consistentClient)(nil)
+
+type consistentClient struct {
+	upstream *client
+	cache    cache
+
+	// lockedKeysByGVK maps gvk -> key -> keyLocker
+	lockedKeysByGVK threadSafeMap[schema.GroupVersionKind, *threadSafeMap[types.NamespacedName, *keyLocker]]
```
sbueringer: If I saw correctly we only add but never remove from this map. While it doesn't take a lot of memory I wonder if this adds up

(didn't thoroughly review the entire PR, so I might have missed the cleanup)

pkg/cache/internal/readerconsistency/readerconsistency.go
alvaroaleman: Fixed this with a custom broadcaster

pkg/cache/multi_namespace_cache.go
alvaroaleman: I don't think there is a metav1.NamespaceAll, that is the clusterCache?

pkg/client/fake/client.go
```diff
@@ -179,6 +181,15 @@ func (f *ClientBuilder) WithObjectTracker(ot testing.ObjectTracker) *ClientBuild
 	return f
 }
 
+// WithGlobalResourceVersionCounter makes the client use a global counter for resourceVersions
+// rather than tracking them per object, mimicking what the apiserver does.
+//
+// Disabled by default.
+func (f *ClientBuilder) WithGlobalResourceVersionCounter() *ClientBuilder {
```
alvaroaleman, line 188: I needed this for the tests, the only way to avoid making this part of the fake clients public interface is to write a custom tracker, because the fakeclient will override any RVs in the incoming object (or fail the request). So this seemed to be the easiest way and may also be useful for others. If you are okay with the approach, I will create a standalone PR for this

pkg/client/client.go
```diff
@@ -319,11 +331,17 @@ func (c *client) Update(ctx context.Context, obj Object, opts ...UpdateOption) e
...
+// delete issues a delete call and returns the response or an error. The response
+// gets deserialized into an unstructured and is either a metav1.Status if the object
+// is gone from storage or the object if it remains, for example because of finalizers.
+func (c *client) delete(ctx context.Context, obj Object, opts ...DeleteOption) (*unstructured.Unstructured, error) {
 	switch obj.(type) {
 	case runtime.Unstructured:
 		return c.unstructuredClient.Delete(ctx, obj, opts...)
-	case *metav1.PartialObjectMetadata:
-		return c.metadataClient.Delete(ctx, obj, opts...)
 	default:
```
alvaroaleman, line 386: This happens to work because the typed delete can delete PartialObjectMeta

pkg/client/consistency.go
alvaroaleman: Fixed

pkg/client/consistency_envtest_test.go
alvaroaleman: Tests for that are in consitency_fake_test.go

pkg/client/consistency_envtest_test.go
```diff
@@ -0,0 +1,275 @@
+/*
```
alvaroaleman, line 1: The bulk of the testing now happens outside of envtest, I think its still nice to have a few envtests in case the wiring in the fake-based tests is buggy, WDYT?

pkg/client/export_test.go
```diff
@@ -0,0 +1,60 @@
+/*
```
alvaroaleman, line 1: The problem this solves it to export a bunch of stuff we need for testing. The tests can not be in package client, because they need to use pkg/cache which imports pkg/client, so they are now in pkg client_test.

alvaroaleman (commented):
I think this is ready for review, it currently contains
* https://github.com/kubernetes-sigs/controller-runtime/pull/3571
* https://github.com/kubernetes-sigs/controller-runtime/pull/3570
* https://github.com/kubernetes-sigs/controller-runtime/pull/3569


I will rebase this once those are merged

pkg/client/consistency.go
alvaroaleman: Missed this comment and yes, this needs to be cleaned up - Will add something tomorrow or so

pkg/client/consistency_envtest_test.go
sbueringer: >  I think its still nice to have a few envtests in case the wiring in the fake-based tests is buggy, WDYT?

Definitely

pkg/cache/cache.go
```diff
@@ -70,6 +70,31 @@ type Cache interface {
...
+type internalCache interface {
+	Cache
+	// SetMinimumRVForGVKAndKey causes subsequent Get requests for the given
+	// GVK and key to block until the informer for that GVK has observed a
+	// resource version >= rv (or the context times out). For List requests
+	// on the given GVK, it blocks until the highest minimum RV across all
+	// keys for that GVK has been observed.
+	SetMinimumRVForGVKAndKey(gvk schema.GroupVersionKind, key client.ObjectKey, rv int64)
+
+	// AddRequiredDeleteForGVKKeyAndUID causes subsequent Get requests for the
```
sbueringer:
```suggestion
	// AddRequiredDeleteForObject causes subsequent Get requests for the
```

pkg/client/client.go
```diff
@@ -85,6 +86,14 @@ type CacheOptions struct {
...
 	// If false, unstructured objects will always result in a live lookup.
 	Unstructured bool
+
+	// ReadYourOwnWriteConsistencyEnabled controls if read requests against the cache will
+	// block until the cache observed all write requests that started before the read
+	// request. Lists will wait for all pending write requests to the gvk the List is for.
+	//
+	// This is an experimental feature, a form of this will be kept but both the details of
+	// how exactly it works and how exactly it is configured may change.
+	ReadYourOwnWriteConsistencyEnabled bool
```
sbueringer: Let's make this a pointer to differentiate unset from false

pkg/client/watch.go
```diff
@@ -28,11 +28,11 @@ import (
 
 // NewWithWatch returns a new WithWatch.
 func NewWithWatch(config *rest.Config, options Options) (WithWatch, error) {
-	client, err := newClient(config, options)
+	c, err := newClient(config, options)
 	if err != nil {
 		return nil, err
 	}
-	return &watchingClient{client: client}, nil
+	return &watchingClient{client: c.(*client)}, nil // TODO: This will panic if consistency is enabled
```
sbueringer: Let's return an error instead of panic'ing?

pkg/client/export_test.go
sbueringer: Let's add a godoc explaining that.

NewKeyLocker seems unused. Let's either use or delete it

pkg/client/consistency.go
```diff
@@ -0,0 +1,416 @@
...
+
+func newConsistentClient(upstream consistentClientUpstream, c cache, newKeyLock func() keyLock) *consistentClient {
+	if newKeyLock == nil {
+		newKeyLock = func() keyLock { return &keyLocker{} }
+	}
+
+	return &consistentClient{
+		upstream: upstream,
+		cache:    c,
+		lockedKeysByGVK: newThreadSafeMap[schema.GroupVersionKind](func() *threadSafeMap[types.NamespacedName, keyLock] {
```
sbueringer: (up to you) Maybe also explicitly declare the second generic type parameter. Probably saves some time when folks read this for the first time

pkg/cache/internal/cache_reader.go
```diff
@@ -52,10 +50,12 @@ type CacheReader struct {
 	// Be very careful with this, when enabled you must DeepCopy any object before mutating it,
 	// otherwise you will mutate the object in the cache.
 	disableDeepCopy bool
+
+	*readerconsistency.ConsistencyHandler
```
sbueringer:
```suggestion
	ConsistencyHandler *readerconsistency.ConsistencyHandler
```
If you don't mind. Makes it easier to jump to reads/writes of this field with Intellij at least

pkg/cache/internal/cache_reader.go
```diff
@@ -64,6 +64,10 @@ func (c *CacheReader) Get(_ context.Context, key client.ObjectKey, out client.Ob
 	}
 	storeKey := objectKeyToStoreKey(key)
 
+	if err := c.ConsistencyHandler.WaitForGet(ctx, key, minRV); err != nil {
```
sbueringer: If I see correctly this also gets executed if ReadYourOwnWriteConsistencyEnabled is false, just with minRV set to 0?

If I got it right inside of WaitForGet/WaitForList we have to wait for the next broadcast even if minRV is 0, could we short-circuit this a bit if minRV is 0?

pkg/cache/internal/informers.go
```diff
@@ -205,6 +208,12 @@ type Informers struct {
 	// watchErrorHandler to be set by overriding the options
 	// or to use the default watchErrorHandler
 	watchErrorHandler cache.WatchErrorHandlerWithContext
+
+	// MinimumRVs stores the minimum RVs we must have seen before returning reads. Due to RV
+	// being just a version, we can store this before we have an informer. This is
+	// different from deletes, where we can only observe deletes once we do have an informer,
+	// so we only allow storing required deletes as part of the informer.
+	MinimumRVs *minimumRVStore
```
sbueringer: Context: one of the non-goals was "Consistency between multiple clients constructed using the same informercache"

If I understand correctly the current implementation:
* locks Get/List in the client until concurrent writes are completed
* implements the "wait for RV/..." inside the cache (specifically the MinimumRVs are stored in the cache)

That gives us a partial consistency if multiple clients are using the same cache, right?

pkg/client/consistency.go
```diff
@@ -0,0 +1,416 @@
...
+
+type consistentClient struct {
+	upstream consistentClientUpstream
+	cache    cache
+
+	// lockedKeysByGVK maps gvk -> key -> keyLock
+	lockedKeysByGVK *threadSafeMap[schema.GroupVersionKind, *threadSafeMap[types.NamespacedName, keyLock]]
+}
+
+func (c *consistentClient) Get(ctx context.Context, key ObjectKey, obj Object, opts ...GetOption) error {
```
sbueringer, line 122: *nit* I assume out of scope of the current PR? (which would be fine, let's just double check the proposal after this PR and write up an umbrella issue)

> The wrapper also has an internal wait time out that defaults to five seconds but can be adjusted through the ReadYourOwnWrite options. Both the passed contexts timeout and the internal timeout will be respected and if either is hit, a sentinel error that users can check for is returned.

pkg/client/consistency.go
```diff
@@ -0,0 +1,416 @@
...
+	gvk, err := apiutil.GVKForObject(obj, c.upstream.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to get GVK for object %T: %w", obj, err)
+	}
+
+	keyLock := c.lockedKeysByGVK.getOrCreate(gvk).getOrCreate(key)
+	if err := keyLock.Wait(ctx); err != nil {
```
sbueringer: Let's assume the following situation:
1. one write call in progress that holds the Lock
2. another concurrent write call on the same objects is in Lock and waiting for the lock
3. a Get call is blocked in Wait()

The Get call would be unblocked as soon as the write call from 1. completes and would not wait for the write call from 2., correct?

pkg/cache/cache.go
```diff
@@ -133,6 +158,11 @@ type Informer interface {
 
 	// IsStopped returns true if the informer has been stopped.
 	IsStopped() bool
+
+	// LastSyncResourceVersion is the resource version observed when last synced with the underlying
+	// store. The value returned is not synchronized with access to the underlying store and is not
+	// thread-safe.
+	LastSyncResourceVersion() string
```
sbueringer: Seems to be unused

pkg/cache/informer_cache.go
```diff
@@ -177,6 +178,50 @@ func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.Grou
...
+	if !started {
+		return &ErrCacheNotStarted{}
+	}
+	if !cache.Informer.HasSynced() {
+		return fmt.Errorf("informer for GVK %v is not synced", gvk)
+	}
+
+	cache.Reader.ConsistencyHandler.AddPendingDelete(
+		client.ObjectKey{Namespace: obj.GetNamespace(), Name: obj.GetName()},
+		obj.GetUID(),
```
halfcrazy: `c.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "bar"}})`
If deleted by namespaced key, not obj, there is no uid. Will cause a block read. Better to check if this obj exists and fill its uid.

pkg/cache/informer_cache.go
```diff
@@ -177,6 +178,50 @@ func (ic *informerCache) getInformerForKind(ctx context.Context, gvk schema.Grou
...
+func (ic *informerCache) SetMinimumRVForGVKAndKey(gvk schema.GroupVersionKind, key client.ObjectKey, rv int64) {
+	ic.MinimumRVs.Set(gvk, key, rv)
+}
+
+func (ic *informerCache) AddRequiredDeleteForObject(obj client.Object) error {
+	gvk, err := apiutil.GVKForObject(obj, ic.scheme)
+	if err != nil {
+		return err
+	}
+	cache, started, ok := ic.Peek(gvk, obj)
```
halfcrazy: Deleting a not-cached object will fail. Eg DisableCacheFor

pkg/client/consistency.go
```diff
@@ -0,0 +1,416 @@
...
+	lockedKeysByGVK *threadSafeMap[schema.GroupVersionKind, *threadSafeMap[types.NamespacedName, keyLock]]
+}
+
+func (c *consistentClient) Get(ctx context.Context, key ObjectKey, obj Object, opts ...GetOption) error {
+	gvk, err := apiutil.GVKForObject(obj, c.upstream.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to get GVK for object %T: %w", obj, err)
+	}
+
+	keyLock := c.lockedKeysByGVK.getOrCreate(gvk).getOrCreate(key)
```
halfcrazy: Not seen where to threadSafeMap delete the key.

alvaroaleman:
Three new commits with the primary goal of moving everything in pkg/client to be able to have multiple consistent clients that share a cache but are not consistent with each other (state before the changes is the first commit):

1. Merge minimumRVStore into ConsistencyHandler: I moved the minimumRVStore into the consistencyHandler. The reason it wasn't there is that it could exist before the consistency handler because it is valid to store an RV before having an informer. I ended up always creating an informer in the write path in 3 below, so this doesn't really matter

2. Move cache.Informer and Informers into pkg/cache/cacheapi: The reason why having the whole thing in the client is tricky is the fact that we need to add an EventHandler but the cache already imports the client, so the client can not just import the cache. I dealt with this by moving the `Informer` and `Informers` interface from the cache as well as `client.Object` and `client.ExtractFunc` which are referenced by the former into a new pkg/cache/cacheapi and leaving aliases for compat in the old places. I am not a huge fan of this, because my editor forces me to follow the alias to be able to see the definition, but I don't see a way around this if we want to do this in the client. I could've also moved the client identifiers the cache uses into a new api package instad, but I think that would be overall worse since it references the client.Reader so it would need to have all the get and list opts in there (but if you think that may be better, happy to change)

3. Move cache waiting entirely into the client wrapper: Moves everything in the client, no changes in the cache anymore besides the new pkg from 2). Much, much cleaner IMHO. Results in creating an informer in writes if not exists but I think thats fine? If this is enabled for a resource, its reasonable to expect there to be a cache. Also updated tests to not test pkg/cache anymore, since that basically only means we test upstream code and IMHO we should assume that that is correct

sbueringer:
1. Sounds good
2. Agree, sounds like the right direction. nit: maybe pkg/cache/api is better than pkg/cache/cacheapi
3. 🎉  Agree

One question about 3, should the informer be created before we do the write (e.g. in writeAndRecordRV). Because if the write goes through and then the informer start (including sync) times out we don't store the RV from that write.

Maybe the current implementation is fine because if we start the informer after the write it should (?) also contain the write anyway as soon as the informer is synced?

But maybe another reason to get the informer first is that it would be more intuitive that if the Write methods on our Client fail the Write to the apiserver actually didn't complete (which would not be the case if we fail creating the informer after the write succeeded)

pkg/client/export_test.go
alvaroaleman: Done

pkg/cache/cache.go
alvaroaleman: This doesn't exist anymore

pkg/client/client.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/watch.go
alvaroaleman: Done

pkg/cache/internal/informers.go
alvaroaleman: This is resolved now I think

pkg/client/consistency.go
alvaroaleman: Yeah, lets make it a follow-up

pkg/client/consistency.go
alvaroaleman: I replaced the keyLock with a writeBarrier which is a tracker of all current in-flight writes that can be sealed to wait for all of them

pkg/cache/cache.go
alvaroaleman: Removed

pkg/cache/informer_cache.go
alvaroaleman: Fixed, it will take the UID 1) from the precondition if set 2) from the obj if set 3) by retrieving the object from the cache

pkg/cache/informer_cache.go
alvaroaleman: Yes, a non-goal for the moment is to deal with configs where objects that are written to are not in the cache

pkg/client/consistency.go
alvaroaleman: Fixed

alvaroaleman:
> One question about 3, should the informer be created before we do the write (e.g. in writeAndRecordRV). Because if the write goes through and then the informer start (including sync) times out we don't store the RV from that write.

An argument can be made either way I think, but I changed this just so we fail as soon as we can if there is an issue with getting or syncing an informer

pkg/client/fake/client_test.go
```diff
@@ -374,6 +374,15 @@ var _ = Describe("Fake client", func() {
 			Expect(partialObjMeta.APIVersion).To(Equal("v1"))
 		})
 
+		It("should be able to Create an unregistered type using PartialObjectMetadata", func(ctx SpecContext) {
```
sbueringer: *nit* Do we intentionally want to add these tests in this PR? I wonder if they are only here because they were removed from https://github.com/kubernetes-sigs/controller-runtime/pull/3571 (IIRC)

(same for "requires a SubResourceBody to Apply scale subresources for resource")

pkg/client/typed_client.go
```diff
@@ -73,22 +73,34 @@ func (c *typedClient) Update(ctx context.Context, obj Object, opts ...UpdateOpti
...
 		Resource(o.resource()).
 		Name(o.name).
 		Body(deleteOpts.AsDeleteOptions()).
 		Do(ctx).
-		Error()
+		Get()
+	if err != nil {
+		return nil, err
+	}
+	data, err := json.Marshal(runtimeObj)
```
sbueringer, line 96: Could we implement this the same way as in the unstructured client?

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	informers cacheapi.Informers
+
+	// writeBarriersByGVK maps gvk -> key -> writeBarrier
+	writeBarriersByGVK *threadSafeMap[schema.GroupVersionKind, *writeBarriers]
+
+	consistencyHandlers *threadSafeMap[schema.GroupVersionKind, *consistencyhandler.ConsistencyHandler]
+}
+
+func (c *consistentClient) getConsistencyHandler(ctx context.Context, gvk schema.GroupVersionKind, obj cacheapi.Object) (*consistencyhandler.ConsistencyHandler, error) {
+	h := c.consistencyHandlers.getOrCreate(gvk)
```
sbueringer: Is it enough to do this by GVK or do we need GVK + Type?

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	rvRaw, err := getResourceVersion()
+	if err != nil {
+		return fmt.Errorf("failed to get resource version from %T: %w", obj, err)
+	}
+	rv, err := strconv.ParseInt(rvRaw, 10, 64)
+	if err != nil {
+		return fmt.Errorf("failed to parse resource version %s: %w", rvRaw, err)
+	}
+
+	h, err := c.getConsistencyHandler(ctx, gvk, cacheObj)
```
sbueringer: Are we intentionally not using the handler returned from getConsistencyHandler above? If yes, let's add a comment explaining why

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+}
+
+func (c *consistentClient) Get(ctx context.Context, key ObjectKey, obj Object, opts ...GetOption) error {
+	gvk, err := apiutil.GVKForObject(obj, c.upstream.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to get GVK for object %T: %w", obj, err)
+	}
+
+	select {
+	case <-c.writeBarriersByGVK.getOrCreate(gvk).seal(key):
```
sbueringer: Played a bit around with the unit test.

Following scenario:
* Write starts
  * writeBarriers.Begin
* Get starts
  * writeBarriers.Seal
    * keyWriteBarrier.Seal: sets b.current = nil
* List starts
  * writeBarriers.Seal
    * keyWriteBarrier.Seal: b.current == nil => return closedChannel
  * List returns 
* Write returns
* Get returns

(i.e. the second Get is not waiting for the Write)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache/cacheapi"
+)
+
+func NewHandler(log logr.Logger) *ConsistencyHandler {
+	return &ConsistencyHandler{
+		rvBroadCaster:             &broadcaster{},
+		pendingDeletesBroadcaster: &broadcaster{},
+		pendingDeletesLock:        sync.RWMutex{},
+		pendingDeletes:            make(map[types.NamespacedName]sets.Set[types.UID]),
+		minimumRVsLock:            sync.Mutex{},
```
sbueringer:
```suggestion
		pendingDeletes:            make(map[types.NamespacedName]sets.Set[types.UID]),
```
nit: not needed

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	rvBroadCaster *broadcaster
+	observedRV    atomic.Int64
+
+	pendingDeletesBroadcaster *broadcaster
+	pendingDeletesLock        sync.RWMutex
+	// pendingDeletes holds pending deletes. Must only be accessed when holding pendingDeletesLock
+	pendingDeletes map[types.NamespacedName]sets.Set[types.UID]
+
+	minimumRVsLock sync.Mutex
+	// minimumRVs stores the minimum RVs we must have seen before returning reads.
```
sbueringer: Maybe also mention the lock (cf. with pendingDeletes)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	if h.registration != nil {
+		return h.waitForHandlerSyncLocked(ctx, *h.registration)
+	}
+
+	registration, err := informer.AddEventHandler(h)
+	if err != nil {
+		return fmt.Errorf("failed to add consistency handler to informer: %w", err)
+	}
```
sbueringer, line 93: I assume the idea was to write h.registration here so that we don't call AddEventHandler twice even if waitForHandlerSyncLocked times out? (but atm h.registration is never written)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+
+func (h *ConsistencyHandler) SetMinimumRV(key types.NamespacedName, rv int64) {
+	h.minimumRVsLock.Lock()
+	defer h.minimumRVsLock.Unlock()
+	if rv > h.minimumRVs[key] {
+		h.minimumRVs[key] = rv
+	}
+}
+
+func (h *ConsistencyHandler) getMinimumRVForKey(key types.NamespacedName) int64 {
```
sbueringer, line 117: *nit* I think this could be a read lock (same for getMinimumRVForGVK)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	defer h.pendingDeletesLock.Unlock()
+
+	if h.pendingDeletes[key] == nil {
+		h.pendingDeletes[key] = sets.New(uid)
+		return
+	}
+	h.pendingDeletes[key].Insert(uid)
+}
+
+func (h *ConsistencyHandler) RemovePendingDelete(key types.NamespacedName, uid types.UID) {
```
sbueringer, line 156: Should we reuse this func in observeDeletion? (or have one func that is used in both places)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	if h.pendingDeletes[key] != nil {
+		h.pendingDeletes[key].Delete(uid)
+		if len(h.pendingDeletes[key]) == 0 {
+			delete(h.pendingDeletes, key)
+		}
+		h.pendingDeletesBroadcaster.broadcast()
+	}
+}
+
+func (h *ConsistencyHandler) WaitForList(ctx context.Context) error {
```
sbueringer, line 169: I wonder if it would be better if we "freeze" both the RV and pendingDeletes at the beginning of this func and then wait for both.

Atm we only "freeze" the deletes inside of waitAllDeletes after we waited for the RV

I think the easiest way to achieve this would be to call h.getMinimumRVForGVK() at the beginning and then call waitAllDeletes before waitForRV.

But maybe it's better if we do this more explicitly:
* call h.GetMinimiumRVForGVK
* call an equivalent new func for pending deletes (or just inline the locking + maps.Clone)
* then call waitforRV and waitDeletes
* + add a godoc why we are doing it this way

(similar for WaitForGet)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	if len(uids) == 0 {
+		return nil
+	}
+
+	for {
+		// must store the chan before checking the deletes to guarantee that even if the deletes
+		// get  updated after our check and before the select, we still get an event.
+		updatedChan := h.pendingDeletesBroadcaster.wait()
+		h.pendingDeletesLock.RLock()
+		done := h.allDeletedLocked(uids)
```
sbueringer: nit: maybe move the RLock/RUnlock inside the func, just a bit nicer to read

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+
+func (h *ConsistencyHandler) OnDelete(raw any) {
+	var obj cacheapi.Object
+	switch t := raw.(type) {
+	case cacheapi.Object:
+		obj = t
+	case cache.DeletedFinalStateUnknown:
+		obj = t.Obj.(cacheapi.Object)
+	default:
+		h.log.Error(nil, "OnDelete received object that is not a cacheapi.Object or DeletedFinalStateUnknown", "object", raw)
```
sbueringer:
```suggestion
		h.log.Error(nil, "OnDelete received object that is not a cacheapi.Object or toolscache.DeletedFinalStateUnknown", "object", raw)
```
nit: maybe

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+			return consistencyhandler.NewHandler(log)
+		}),
+	}
+}
+
+type consistentClient struct {
+	upstream  consistentClientUpstream
+	informers cacheapi.Informers
+
+	// writeBarriersByGVK maps gvk -> key -> writeBarrier
```
sbueringer: nit: I think this one is outdated there is one more level in between now (writeBarrierWithRefCounter)

Once we finalized how the barriers work would be good to have a godoc that explains it somewhere

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	h, err := c.getConsistencyHandler(ctx, gvk, obj)
+	if err != nil {
+		return err
+	}
+	if err := h.WaitForGet(ctx, key); err != nil {
+		return err
```
sbueringer: nit: Not sure how the error looks like that we get out of this func but I suspect we can improve it (probably inside of WaitForGet/WaitForList)

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+
+	for _, s := range c.writeBarriersByGVK.getOrCreate(gvk).sealAll() {
+		select {
+		case <-s:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+
+	listObj, err := c.upstream.Scheme().New(gvk)
```
sbueringer:
```suggestion
	obj, err := c.upstream.Scheme().New(gvk)
```
I think

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	gvk.Kind = strings.TrimSuffix(gvk.Kind, "List")
+
+	for _, s := range c.writeBarriersByGVK.getOrCreate(gvk).sealAll() {
+		select {
+		case <-s:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+
```
sbueringer, line 169: I think Scheme().New does not work for unstructured and PartialObjectMeta (If I'm correct I assume we are missing some test coverage)

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+		if err != nil {
+			return schema.GroupVersionKind{}, types.NamespacedName{}, nil, nil, fmt.Errorf("failed to get GVK for apply configuration %T: %w", obj, err)
+		}
+		cacheObj, err := scheme.New(gvk)
+		if err != nil {
+			return schema.GroupVersionKind{}, types.NamespacedName{}, nil, nil, fmt.Errorf("failed to create cache object for GVK %s: %w", gvk, err)
+		}
+		clientObj, ok := cacheObj.(Object)
+		if !ok {
+			return schema.GroupVersionKind{}, types.NamespacedName{}, nil, nil, fmt.Errorf("cache object of type %T for GVK %s does not implement client.Object", cacheObj, gvk)
```
sbueringer: nit: in other places we use cacheapi.Object in these error messages. I think maybe consistencly using client.Object could be the best option (and less confusing for users of the client)

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+		}
+		clientObj.SetName(ptr.Deref(t.GetName(), ""))
+		clientObj.SetNamespace(ptr.Deref(t.GetNamespace(), ""))
+		return gvk,
+			types.NamespacedName{Namespace: ptr.Deref(t.GetNamespace(), ""), Name: ptr.Deref(t.GetName(), "")},
```
sbueringer:
```suggestion
			ObjectKeyFromObject(clientObj),
```
nit (probably also in other cases around here)

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	rv, err := strconv.ParseInt(rvRaw, 10, 64)
+	if err != nil {
+		return fmt.Errorf("failed to parse resource version %s: %w", rvRaw, err)
+	}
+
+	h, err := c.getConsistencyHandler(ctx, gvk, cacheObj)
+	if err != nil {
+		return err
+	}
+	h.SetMinimumRV(ObjectKey{Namespace: cacheObj.GetNamespace(), Name: cacheObj.GetName()}, rv)
```
sbueringer:
```suggestion
	h.SetMinimumRV(ObjectKeyFromObject(cacheObj), rv)
```
nit

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+
+	return nil
+}
+
+func resourceVersionFromApplyConfiguration(obj applyConfiguration) (string, error) {
```
sbueringer, line 314: Wondering if there's any reason why they don't generate a GetResourceVersion func similar to GetName/GetNamespace. Would it make sense to open an issue for that? (as a follow-up)

But I think even if it gets added now to the generator we could not rely on it, but we could maybe short-circuit if the obj has that method

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+
+	uid, err := c.uidForDelete(ctx, gvk, namespacedName, obj, opts...)
+	if err != nil {
+		return err
+	}
+
+	release := c.writeBarriersByGVK.getOrCreate(gvk).Begin(namespacedName)
+	defer release()
+
+	h, err := c.getConsistencyHandler(ctx, gvk, obj)
```
sbueringer: Should we do this before we begin the barrier like in writeAndRecordRV?

(maybe even before uidForDelete, the Get in uidForDelete might also trigger informer creation so maybe it's cleaner to call getConsistencyHandler before uidForDelete)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,349 @@
...
+	case <-ctx.Done():
+		return fmt.Errorf("failed waiting for consistency handler to sync: %w", ctx.Err())
+	}
+}
+
+func (h *ConsistencyHandler) SetMinimumRV(key types.NamespacedName, rv int64) {
+	h.minimumRVsLock.Lock()
+	defer h.minimumRVsLock.Unlock()
+	if rv > h.minimumRVs[key] {
+		h.minimumRVs[key] = rv
```
sbueringer, line 113: Should we maybe skip adding the rv if it is < observedRV?

I'm wondering if there might be cases where we call SetMinimumRV when the cache already has observed the RV and then we have to wait for a new broadcast before we can unblock waitforRV.


Related. Should we do one check in waitForRV & waitDeletes before we wait for a broadcast?

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	h.AddPendingDelete(namespacedName, uid)
+
+	response, err := c.upstream.delete(ctx, obj, opts...)
+	if err != nil {
+		h.RemovePendingDelete(namespacedName, uid)
+		return err
+	}
+
+	if rvRaw := response.GetResourceVersion(); rvRaw != "" {
+		h.RemovePendingDelete(namespacedName, uid)
```
sbueringer: I assume the idea here is that if we could get an RV it's enough to wait for the RV and we don't have to wait for the delete explicitly. (let's add a comment)

Maybe let's only remove the pending delete after we parsed the RV successfully

pkg/client/consistency.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+}
+
+type consistentSubResourceClient struct {
+	writeAndRecordRV func(context.Context, any, func() error) error
+	upstream         SubResourceClient
+}
+
+func (c *consistentSubResourceClient) Get(ctx context.Context, obj, subResource Object, opts ...SubResourceGetOption) error {
+	return c.upstream.Get(ctx, obj, subResource, opts...)
```
sbueringer, line 443: So this one is not consistent, right?

pkg/client/fake/client.go
sbueringer: Approach is fine for me. Also okay for me to keep it in this PR, up to you (but yes for release notes a separate PR would be better)

pkg/client/client.go
sbueringer: Maybe let's also add a comment inline. Just so we preserve this info and nobody gets confused by this (compared to other funcs)

pkg/client/consistency.go
sbueringer: Thx! Opened a new conversation for the barrier

pkg/cache/informer_cache.go
sbueringer: Nice one. I know a few cases where that definitely would have lead to issues

pkg/client/consistency.go
alvaroaleman: That doesn't work, because applyconfigs don't implement metav1.Object, their GetNamespace/GetName return a *string, not a string

pkg/client/fake/client.go
alvaroaleman: https://github.com/kubernetes-sigs/controller-runtime/pull/3581

pkg/client/fake/client_test.go
alvaroaleman: Removed

pkg/client/typed_client.go
alvaroaleman: No, because the deserializer fails that, I've added a comment

pkg/client/consistency.go
alvaroaleman: I've changed the code to key this and the write locks by gvk + representation in `Key consistency handlers and write locks by representation`

pkg/client/consistency.go
alvaroaleman: Changed it to use the handlers returned from getConsistencyHandler

pkg/client/consistency.go
alvaroaleman: Good catch, fixed in `Writebarrier: Wait for previous batches to finish`. Because the code related to that got confusing to me, I moved all of it into a new internal package first in `Move the writebarrier-related code into pkg/client/internal/writebarrier`

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Good catch, fixed that by writing h.registration

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done in both

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Good idea, done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Added snapshotting of the toBeDeleted objects in the beginning of both + godoc

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Removed this altogether because its just a simple map. There are comments in pkg/client/internal/writebarrier/  now which hopefully explain it all enough

pkg/client/consistency.go
alvaroaleman: This ends up always being caused by a context that is done, I've added wrapping at the bottom layer saying what RV or deletion keys we were waiting for

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Yes, good point, moved it

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: yeah good point, not very unlikely to happen but would have quite an effect if it does.

> Related. Should we do one check in waitForRV & waitDeletes before we wait for a broadcast?

We already do or not?

Related the question of deletes. Initially I thought that could be an issue if a concurrent binary deletes or the UID is otherwise wrong, but I think thats fine because then our request would error out which makes us remove the delete again? The only exception to this would be if the cache filters out the object, but we explicitly said we will not handle that.

pkg/client/consistency.go
alvaroaleman: Added and done

pkg/client/consistency.go
alvaroaleman: It is, because this doesn't read from the cache but from the API. I think it only makes sense to use this for subresources that are of a different type like scale

pkg/client/client.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Fixed and also extended the envtest-based tests to test `unstructured` in `Also test unstructured in envtest`. I _think_ back when I still tested with unstructured/objectmeta because it changed the cache it coincidentally passed because those tests used the fakeclient which adds everything to the scheme.

pkg/client/consistency.go
sbueringer: It should work because `clientObj` implements Object (but yes `t` does not)

(but of course this is just a nit)

pkg/client/consistency.go
sbueringer: Ups, of course. Thx for clarifying

pkg/client/internal/consistencyhandler/consistencyhandler.go
sbueringer: > We already do or not?

Oh yup, you are right. Of course `.wait()` does not wait, it just gets the channel :) 

> Related the question of deletes. Initially I thought that could be an issue if a concurrent binary deletes or the UID is otherwise wrong, but I think thats fine because then our request would error out which makes us remove the delete again? The only exception to this would be if the cache filters out the object, but we explicitly said we will not handle that.

Agree

(feel free to resolve the conversation)

pkg/client/consistency.go
sbueringer: Maybe let's already create the umbrella issue, add this as a follow-up task there and then resolve here?

(If you think it's realistic that they might add the GetResourceVersion method, if not, let's just resolve)

alvaroaleman: https://github.com/kubernetes/kubernetes/pull/141688

pkg/client/consistency.go
alvaroaleman: ah ofc, fixed

pkg/client/consistency.go
sbueringer: Thx!

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,335 @@
...
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache/cacheapi"
+)
+
+func NewHandler(log logr.Logger) *ConsistencyHandler {
+	return &ConsistencyHandler{
+		rvBroadCaster:             &broadcaster{},
+		pendingDeletesBroadcaster: &broadcaster{},
+		pendingDeletes:            make(map[types.NamespacedName]sets.Set[types.UID]),
+		minimumRVs:                make(map[types.NamespacedName]int64),
```
sbueringer, line 39: Maybe something like pendingRVs would be slightly better (up to you, feel free to keep it as is)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,335 @@
...
+	pendingDeletesLock        sync.RWMutex
+	// pendingDeletes holds pending deletes. Must only be accessed when holding pendingDeletesLock
+	pendingDeletes map[types.NamespacedName]sets.Set[types.UID]
+
+	minimumRVsLock sync.RWMutex
+	// minimumRVs stores the minimum RVs we must have seen before returning reads. Must only be
+	// accessed when holding minimumRVsLock.
+	minimumRVs map[types.NamespacedName]int64
+
+	registerLock sync.Mutex
```
sbueringer: Not sure if this makes a relevant difference, but maybe make this also a RWMutex

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,335 @@
...
+	// snapshot deletes before waiting for the RV so we don't end up including
+	// deletes that came in while we waited.
+	h.pendingDeletesLock.RLock()
+	pendingDeletes := maps.Clone(h.pendingDeletes[key])
+	h.pendingDeletesLock.RUnlock()
+
+	rv := h.getMinimumRVForKey(key)
+	if err := h.waitForRV(ctx, rv); err != nil {
+		return err
+	}
```
sbueringer:
```suggestion
	if err := h.waitForRV(ctx, h.getMinimumRVForKey(key)); err != nil {
		return err
	}
```
Maybe slightly cleaner, but up to you

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,335 @@
...
+		// gets updated after our check and before the select, we still get an event.
+		updatedChan := h.rvBroadCaster.wait()
+		if h.observedRV.Load() >= rv {
+			return nil
+		}
+		select {
+		case <-updatedChan:
+			continue
+		case <-ctx.Done():
+			return fmt.Errorf("failed waiting for resource version %d to be observed: %w", rv, ctx.Err())
```
sbueringer:
```suggestion
			return fmt.Errorf("failed waiting to observe resource version %d: %w", rv, ctx.Err())
```
Not really important but for consistency with the error in waitDeletes

pkg/client/consistency.go
```diff
@@ -0,0 +1,474 @@
...
+
+type representationID int8
+
+const (
+	representationIDUnstructured representationID = iota
+	representationIDPartialObjectMetadata
+	representationIDTyped
+)
+
+func (c *consistentClient) getConsistencyHandler(
```
sbueringer, line 101: Maybe combine the gvk and representation parameter to one that has type gvkAndRepresentation.

And then e.g. in Get l.128 just do this `gvkr := gvkAndRepresentation{gvk: gvk, representation: representationIDForObj(obj)}` and pass it in all functions that need it

If I see correctly this should be a bit cleaner and easier to read

pkg/client/consistency.go
```diff
@@ -0,0 +1,474 @@
...
+	return c.writeAndRecordRV(ctx, obj, func() error {
+		return c.upstream.Apply(ctx, obj, opts...)
+	})
+}
+
+func writeTargetFor(obj any, scheme *runtime.Scheme) (schema.GroupVersionKind, types.NamespacedName, Object, func() (string, error), error) {
+	switch t := obj.(type) {
+	case *unstructuredApplyConfiguration:
+		return t.Unstructured.GroupVersionKind(),
+			types.NamespacedName{Namespace: t.Unstructured.GetNamespace(), Name: t.Unstructured.GetName()},
```
sbueringer:
```suggestion
			ObjectKeyFromObject(t.Unstructured),
```
nit

`ObjectKeyFromObject(t),` in l.254

similar in l.331

pkg/client/consistency.go
```diff
@@ -0,0 +1,474 @@
...
+			t,
+			func() (string, error) { return t.GetResourceVersion(), nil },
+			nil
+	default:
+		return schema.GroupVersionKind{}, types.NamespacedName{}, nil, nil, fmt.Errorf("unsupported type %T, must be either %T, %T or %T", obj, Object(nil), &unstructuredApplyConfiguration{}, applyConfiguration(nil))
+	}
+}
+
+func (c *consistentClient) writeAndRecordRV(ctx context.Context, obj any, write func() error) error {
+	gvk, namespacedName, cacheObj, getResourceVersion, err := writeTargetFor(obj, c.upstream.Scheme())
```
sbueringer: nit: maybe already return gvkAndRepresentation instead of gvk from writeTargetFor

pkg/client/internal/writebarrier/writebarrier.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Seal seals the currently-active set of locks to key and returns a channel that
+	// closes when they are done.
+	Seal(key types.NamespacedName) <-chan struct{}
+
+	// SealAll is Seal for all keys.
+	SealAll() []<-chan struct{}
+}
+
+// NewWriteBarriers construct WriteBarrierrs. newBarrier is configurable for testing purposes only.
```
sbueringer:
```suggestion
// NewWriteBarriers construct WriteBarriers. newBarrier is configurable for testing purposes only.
```

pkg/client/internal/writebarrier/writebarrier.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+var closedChannel chan struct{}
+
+func init() {
+	closedChannel = make(chan struct{})
+	close(closedChannel)
+}
+
+type writeBatch struct {
```
sbueringer: Let's move this struct and the corresponding release method to the end of the file. Makes it a bit easier to read/understand this file top-down

pkg/client/internal/writebarrier/writebarrier.go
```diff
@@ -0,0 +1,193 @@
...
+
+// NewWriteBarrier creates a new WriteBarrier
+func NewWriteBarrier() WriteBarrier {
+	return &keyWriteBarrier{previous: closedChannel}
+}
+
+// keyWriteBarrier allows to wait for a set of in-flight writes to finish.
+type keyWriteBarrier struct {
+	// mutex must be held to access current
+	mutex sync.Mutex
```
sbueringer: Maybe also call this lock instead of mutex

If this is only for current and not also for previous, maybe let's call it currentLock

pkg/client/internal/writebarrier/writebarrier.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (b *keyWriteBarrier) Seal() <-chan struct{} {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+
+	if b.current == nil {
+		return b.previous
+	}
+
+	done := make(chan struct{})
```
sbueringer, line 159: Maybe a comment here could be useful. Just something about that if there is a current we create a new done channel (and store it in previous) that is closed once b.current and b.previous are closed

pkg/client/internal/writebarrier/writebarrier_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"slices"
+	"testing"
+	"testing/synctest"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestKeyWriteBarrierSealWaitsForOverlappingBatches(t *testing.T) {
```
sbueringer: Maybe we should have more tests with additional sequences of Begin / release / Seal / SealAll

pkg/client/consistency_envtest_test.go
```diff
@@ -0,0 +1,280 @@
...
+	type objectRepresentation struct {
+		name        string
+		newObject   func() client.Object
+		newList     func() client.ObjectList
+		applyConfig func(namespace, name string, annotations map[string]string) (runtime.ApplyConfiguration, func() string)
+	}
+
+	configMapGVK := corev1.SchemeGroupVersion.WithKind("ConfigMap")
+
+	representations := []objectRepresentation{{
```
sbueringer, line 58: Would it be possible to also add PartialObjectMeta?

pkg/client/consistency_fake_test.go
```diff
@@ -0,0 +1,573 @@
...
+		fc,
+		func() writebarrier.WriteBarrier { return barrier },
+	)
+}
+
+type fakeConsistentClientUpstream struct {
+	client.Client
+}
+
+func (u *fakeConsistentClientUpstream) DeleteWithResult(ctx context.Context, obj client.Object, opts ...client.DeleteOption) (*unstructured.Unstructured, error) {
```
sbueringer: Is this only necessary because the fake client does not behave like the real client? If yes, do you think it makes sense to have a follow-up to improve the fake client?

pkg/client/consistency_fake_test.go
```diff
@@ -0,0 +1,573 @@
...
+	writebarrier.WriteBarrier
+	beginCallback func()
+}
+
+func (k *keyWriteBarrierWithBeginCallback) Begin() func() {
+	release := k.WriteBarrier.Begin()
+	k.beginCallback()
+	return release
+}
+
```
sbueringer: Up to you but I usually prefer having the test helpers below the tests

pkg/client/client.go
```diff
@@ -85,6 +93,16 @@ type CacheOptions struct {
 	// read unstructured objects or lists from the cache.
 	// If false, unstructured objects will always result in a live lookup.
 	Unstructured bool
+
+	// ReadYourOwnWriteConsistencyEnabled controls if read requests against the cache will
+	// block until the cache observed all write requests that started before the read
+	// request. Lists will wait for all pending write requests to the gvk the List is for.
```
sbueringer, line 99: Maybe let's make it clear here that the consistent client works per gvk+representation not just gvk

pkg/client/client.go
```diff
@@ -319,11 +364,18 @@ func (c *client) Update(ctx context.Context, obj Object, opts ...UpdateOption) e
...
+	return err
+}
+
+// delete issues a delete call and returns the response or an error. The response
+// gets deserialized into an unstructured and is either a metav1.Status if the object
+// is gone from storage or the object if it remains, for example because of finalizers.
+func (c *client) delete(ctx context.Context, obj Object, opts ...DeleteOption) (*unstructured.Unstructured, error) {
 	switch obj.(type) {
 	case runtime.Unstructured:
 		return c.unstructuredClient.Delete(ctx, obj, opts...)
```
sbueringer, line 384: metadataClient.Delete is now unused, let's delete it

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+
+		if h.observedRV.CompareAndSwap(current, parsed) {
+			break
+		}
+	}
+
+	h.rvBroadCaster.broadcast()
+
+	go h.cleanupMinimumRVs(parsed)
```
sbueringer: Do we have to create a new goroutine for that or can we avoid that by just calling cleanupMinimumRVs directly?

Claude:
```
 Minor goroutine churn: consistencyhandler.go spawns a goroutine per watch event for observeResourceVersion, which itself spawns another  
     goroutine for cleanupMinimumRVs. On high-throughput GVKs this doubles goroutine creation unnecessarily — the inner call could just be    
     synchronous.          
```

pkg/client/consistency_fake_test.go
```diff
@@ -0,0 +1,573 @@
...
+		d := deployment()
+		err := c.Get(ctx, client.ObjectKeyFromObject(d), d)
+		g.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "expected a NotFound error, got %v", err)
+	}
+	listDeleted := func(ctx context.Context, c client.Client, g *WithT, _ <-chan int64) {
+		result := &appsv1.DeploymentList{}
+		g.Expect(c.List(ctx, result)).To(Succeed())
+		g.Expect(result.Items).To(BeEmpty())
+	}
+	getTerminating := func(ctx context.Context, c client.Client, g *WithT, _ <-chan int64) {
```
sbueringer: Maybe also verify the rv here and in listTerminating

pkg/client/consistency_fake_test.go
```diff
@@ -0,0 +1,573 @@
...
+		result := &appsv1.DeploymentList{}
+		g.Expect(c.List(ctx, result)).To(Succeed())
+		g.Expect(result.Items).To(HaveLen(1))
+		g.Expect(result.Items[0].GetDeletionTimestamp()).ToNot(BeNil(), "expected the deletionTimestamp to be set")
+		g.Expect(result.Items[0].GetFinalizers()).To(ConsistOf("test.io/finalizer"))
+	}
+
+	testCases := []struct {
+		name            string
+		maybeInitObject func() *appsv1.Deployment
```
sbueringer, line 188: I would drop the maybe. If it is set we use it, right? (but up to you, just a nit)

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: This was a good finding, combined the two in the argument

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/consistency.go
alvaroaleman: Done

pkg/client/internal/writebarrier/writebarrier.go
alvaroaleman: Done

pkg/client/internal/writebarrier/writebarrier.go
alvaroaleman: Done

pkg/client/internal/writebarrier/writebarrier.go
alvaroaleman: Done and updated comment, it is required for both

pkg/client/internal/writebarrier/writebarrier.go
alvaroaleman: Done

pkg/client/internal/writebarrier/writebarrier_test.go
alvaroaleman: Added a few

pkg/client/consistency_envtest_test.go
alvaroaleman: That would require a bunch of specialcasing in the test, because we don't support Create or Update for them (despite it being valid when used with an object that is valid with metadata-only)

pkg/client/consistency_fake_test.go
alvaroaleman: Well, we deliberately don't export the result of the delete call in the real client either, since it will likely lead to confusion so IMHO doing this for tests is fine

pkg/client/consistency_fake_test.go
alvaroaleman: Done

pkg/client/client.go
alvaroaleman: Added a sentence about that, let me know what you think

pkg/client/client.go
alvaroaleman: I didn't realize we wrap it, TIL

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: The reason I did that was because I didn't want to block the handler. I implemented something here to avoid the goroutine and then realized that we already start this in a goroutine in the handler 🤦 I left the improved version in, just because I already did it and it improves lock contention, not sure if super relevant

pkg/client/consistency_fake_test.go
alvaroaleman: Done

pkg/client/consistency_fake_test.go
alvaroaleman: The maybe is to indicate that the constructor may be nil, not that the result of the constructor may be nil

alvaroaleman:
Addressed all the comments. I also added a commit `Add a DisableReadWriteConsistency option` because I think in certain scenarios like writing something that is not cached, this functionality may be unusable otherwise

pkg/client/internal/consistencyhandler/consistencyhandler.go
sbueringer: I think this one is not done (or I'm not seeing the change)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,344 @@
...
+
+		if h.observedRV.CompareAndSwap(current, parsed) {
+			break
+		}
+	}
+
+	h.rvBroadCaster.broadcast()
+
+	select {
+	case h.triggerCleanup <- struct{}{}:
```
sbueringer, line 280: Don't we have to read from the channel somewhere? Right now this means cleanupMinimumRVs will be only executed once per ConsistencyHandler, right?

(I think it could also be okay to simply call `h.cleanupMinimumRVs()` directly)

pkg/client/internal/writebarrier/writebarrier_test.go
sbueringer: Nice, thx!!

pkg/client/client.go
```diff
@@ -85,6 +93,23 @@ type CacheOptions struct {
 	// read unstructured objects or lists from the cache.
 	// If false, unstructured objects will always result in a live lookup.
 	Unstructured bool
+
+	// ReadYourOwnWriteConsistencyEnabled controls if read requests against the cache will
+	// block until the cache observed all write requests that started before the read
+	// request. Lists will wait for all pending write requests to the gvk the List is for.
+	//
+	// The `DisableReadWriteConsistency` option can be used to disable this functionality
```
sbueringer: Should we make the option and the field here more consistent: `ReadYourOwnWriteConsistencyEnabled` vs `DisableReadWriteConsistency`. 

Some existing fields to compare with:
* `UsePriorityQueue`
* `EnableWarmup`
* `UnsafeDisableDeepCopy`

I would suggest:
* here: `EnableReadWriteConsistency`
* option: `DisableReadWriteConsistency`

(no strong opinion, but more consistent (😂) naming would be good)

pkg/client/options.go
```diff
@@ -538,6 +628,11 @@ type ListOptions struct {
 	// +optional
 	UnsafeDisableDeepCopy *bool
 
+	// DisableReadWriteConsistency indicates that the request will not wait for the
+	// cache to have observed previous writes. It has no effect if the client was not
+	// configured with CacheOptions.ReadYourOwnWriteConsistencyEnabled.
+	DisableReadWriteConsistency bool
```
sbueringer: Should this be also set in ApplyToList?

pkg/client/client.go
alvaroaleman: To get the ~worst~ best of both options, I went with `[Enable|Disable]ReadYourWritesConsistency`

pkg/client/options.go
alvaroaleman: Good catch, fixed

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: lol yeah it was late when I wrote this code :D Added that. I think its actually good to do this, just because cleaning up all the time may cause lock contention during initial list for example

pkg/client/options.go
```diff
@@ -200,6 +200,72 @@ func (f FieldOwner) ApplyToSubResourceApply(opts *SubResourceApplyOptions) {
 	opts.FieldManager = string(f)
 }
 
+// DisableReadYourWritesConsistency disables read-your-own-write consistency for the
+// request it is passed to. It has no effect if the client was not configured with
+// CacheOptions.ReadYourWritesConsistencyEnabled.
```
sbueringer: nit: field name is off (same a few more times below)

pkg/client/internal/consistencyhandler/consistencyhandler.go
```diff
@@ -0,0 +1,345 @@
...
+		h.cleanupMinimumRVs()
+		<-h.triggerCleanup
+	default:
+	}
+}
+
+func (h *ConsistencyHandler) OnAdd(raw any, _ bool) {
```
sbueringer, line 287: (I think at most this should be a follow-up)

I'm a bit confused by this statement in the Kubernetes v1.37 blog:
> Stale controller mitigation has been Beta since v1.36, enabled by default per controller behind a StaleControllerConsistency<Controller> feature gate; Kubernetes v1.37 extends it to the HorizontalPodAutoscaler controller and adds the circuit-breaking variant and extra metrics described below. The core mechanism is a read your writes guarantee: client-go's ResourceEventHandlerFuncs gets a new BookmarkFunc callback so a controller can reliably track the resource version of objects it cares about, even through edge cases the existing add/update/delete callbacks miss.
> https://kubernetes.io/blog/2026/08/26/kubernetes-v1-37-release/#stale-controller-mitigation

Specifically this part: "client-go's ResourceEventHandlerFuncs gets a new BookmarkFunc callback so a controller".

I grepped for BookmarkFunc on k/k main and it basically doesn't exist

xref: https://cs.k8s.io/?q=BookmarkFunc&i=nope&literal=nope&files=&excludeFiles=&repos=

Also mentioned in the proposal: https://github.com/kubernetes/enhancements/blob/e849163ac4a0a5241ba626bd9a99820bf1dcd279/keps/sig-api-machinery/5647-stale-controller-handling/README.md

I think if it would exist, it would be a nice improvement to use it

pkg/client/internal/consistencyhandler/consistencyhandler.go
sbueringer: In general I'm fine having some code here to optimize this. I just want to make sure it behaves correctly.

Does the current code guarantee that there will be a cleanupMinimumRVs executed with the latest observedRV?

I think if observeResourceVersion gets called twice concurrently, the second one does not trigger cleanupMinimumRVs and there's no guarantee that the first one executed based on observedRV from the second observeResourceVersion

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: I saw that too and also searched for it and also couldn't find it. Asking about that here: https://kubernetes.slack.com/archives/C0EG7JC6T/p1788396097711009

pkg/client/internal/consistencyhandler/consistencyhandler.go
alvaroaleman: > Does the current code guarantee that there will be a cleanupMinimumRVs executed with the latest observedRV?

It does not, but the cleanup is not required for correctness, because we always check the rv once before we start waiting for changes: https://github.com/kubernetes-sigs/controller-runtime/blob/a7c5f406bcac8c9375877bebb8d338860ef218a8/pkg/client/internal/consistencyhandler/consistencyhandler.go#L247-L249

(I've renamed `wait` to `getWaitChan` because as you pointed out earlier, a `wait` that doesn't wait is a bit confusing)

pkg/client/options.go
alvaroaleman: Fixed

pkg/client/internal/consistencyhandler/consistencyhandler.go
sbueringer: Thx, looks like we're good!

pkg/client/internal/consistencyhandler/consistencyhandler.go
sbueringer: Entirely missed that the cleanup is basically just best effort and not needed for the wait :) Thx for the clarification

sbueringer:
Thank you very much!

Really really great work! :)

/lgtm
/approve
/hold cancel

🎉

PR #3475, opened 2026-03-08
:bug: Update typed Applyconfigurations with server response

### LINKED ISSUE #3474: Apply using typed apply configurations doesn't deserialize the response into the passed object
The client-go result requires a runtime.Object to decode into: https://github.com/kubernetes/kubernetes/blob/dfbe79674a33c47c1f0fd7b75882314f6db5e136/staging/src/k8s.io/client-go/rest/request.go#L1449

Because of this, the controller-runtime typed apply wraps the passed applyconfiguration with something that implements runtime.Object and embedds an interface for the applyconfiguration: https://github.com/kubernetes-sigs/controller-runtime/blob/105baa6284da3d3ca5dc23b3fe1dd30df7ea3df7/pkg/client/typed_client.go#L159 https://github.com/kubernetes-sigs/controller-runtime/blob/105baa6284da3d3ca5dc23b3fe1dd30df7ea3df7/pkg/client/applyconfigurations.go#L60-L62

The json unmarshaller however doesn't inline interfaces, only embedded structs. As a result, only hypothetical fields under an `ApplyConfiguration` key would be deserialized into the applyconfiguration and hence we never update the passed applyconfiguration. 

Deserialization of the result works fine in the case of unstructured.

/kind bug
/assign

alvaroaleman:
Prior to this, we would never update a typed applyconfiguration with the response from the server, this change fixes that.

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3474

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -943,6 +943,7 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 
 				err = cl.Apply(ctx, client.ApplyConfigurationFromUnstructured(obj), &client.ApplyOptions{FieldManager: "test-manager"})
 				Expect(err).NotTo(HaveOccurred())
+				Expect(obj.GetResourceVersion()).NotTo(BeEmpty())
 
 				cm, err := clientset.CoreV1().ConfigMaps(obj.GetNamespace()).Get(ctx, obj.GetName(), metav1.GetOptions{})
 				Expect(err).NotTo(HaveOccurred())
@@ -1015,6 +1016,7 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 
 				err = cl.Apply(ctx, obj, &client.ApplyOptions{FieldManager: "test-manager"})
 				Expect(err).NotTo(HaveOccurred())
+				Expect(obj.ResourceVersion).NotTo(BeNil())
 
 				cm, err := clientset.CoreV1().ConfigMaps(ptr.Deref(obj.GetNamespace(), "")).Get(ctx, ptr.Deref(obj.GetName(), ""), metav1.GetOptions{})
 				Expect(err).NotTo(HaveOccurred())
@@ -1289,6 +1291,7 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 				By("Applying the scale subresurce")
 				deploymentAC, err := appsv1applyconfigurations.ExtractDeployment(dep, "foo")
 				Expect(err).NotTo(HaveOccurred())
+				initialRV := deploymentAC.ResourceVersion
 				scale := autoscaling1applyconfigurations.Scale().
 					WithSpec(autoscaling1applyconfigurations.ScaleSpec().WithReplicas(replicaCount))
 				err = cl.SubResource("scale").Apply(ctx, deploymentAC,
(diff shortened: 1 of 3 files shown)


alvaroaleman:
One question is if we should backport this, the problem with that is that it may break existing code, since the response contains a resource version, so when re-using an object to do multiple requests that would have previously worked but won't anymore after this

dongjiang1989 (approved):
LGTM

sbueringer:
/hold

We should discuss this a bit

pkg/client/typed_client.go
```diff
@@ -146,17 +147,18 @@ func (c *typedClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration,
...
-		// that is not implemented by the ApplyConfigurations. The generated clients
-		// don't have this problem because they deserialize into the api type, not the
-		// apply configuration: https://github.com/kubernetes/kubernetes/blob/22f5e01a37c0bc6a5f494dec14dd4e3688ee1d55/staging/src/k8s.io/client-go/gentype/type.go#L296-L317
-		Into(runtimeObjectFromApplyConfiguration(obj))
+		Raw()
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(body, obj)
```
sbueringer, line 167: Do we have to do some sort of zeroing on the obj?

pkg/client/client_test.go
```diff
@@ -1015,6 +1016,7 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 
 				err = cl.Apply(ctx, obj, &client.ApplyOptions{FieldManager: "test-manager"})
 				Expect(err).NotTo(HaveOccurred())
+				Expect(obj.ResourceVersion).NotTo(BeNil())
```
sbueringer, line 1019: Can we add something similar for subresource Apply?

sbueringer:
> We should discuss this a bit

Added my take on the issue

pkg/client/client_test.go
alvaroaleman: Done

pkg/client/typed_client.go
alvaroaleman: Arguably yes but we don't do that for other requests either, so I'd prefer if we either do it consistently for all of them or for none

matheuscscp:
Hey @sbueringer, sorry for the ping, looks like this is the last unresolved item for v0.24.0 according to the tracking issue. Can you please review? :pray:

pkg/client/typed_client.go
```diff
@@ -146,17 +147,24 @@ func (c *typedClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration,
...
-		// don't have this problem because they deserialize into the api type, not the
-		// apply configuration: https://github.com/kubernetes/kubernetes/blob/22f5e01a37c0bc6a5f494dec14dd4e3688ee1d55/staging/src/k8s.io/client-go/gentype/type.go#L296-L317
-		Into(runtimeObjectFromApplyConfiguration(obj))
+		ContentType(&contentType).
+		Raw()
+	if err != nil {
+		return err
+	}
+
+	if contentType != "application/json" {
```
sbueringer, line 163: Do we also want this safeguard in ApplySubResource?

sbueringer (commented):
One nit

pkg/client/typed_client.go
alvaroaleman: Added it there

sbueringer:
Thank you very much!

/lgtm
/approve
/hold cancel

Issue #2744, opened 2024-03-29
Incompatibility Issue with controller-runtime Version Pinning and envtest Setup - It is requiring go 1.22 when the latest release still using go 1.21 as old project versions


camilamacedo86:
## Problem Description

We've encountered an issue where our environment cannot pin a specific version of `controller-runtime`, leading to compatibility issues with earlier versions of kubebuilder when setting up `envtest`. This incompatibility prevents our projects from maintaining a consistent development environment, particularly affecting our CI/CD pipelines.

## Error Output

While attempting to set up `envtest` (go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest), we're met with the following error message due to the version incompatibility besides the LATEST v0.17.2 does not support go 1.22 and old projects not be using it yet as well.

```
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240327193027-21368602d84b
go: downloading sigs.k8s.io/controller-runtime v0.17.2
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240327193027-21368602d84b requires go >= 1.22.0 (running go 1.21.8; GOTOOLCHAIN=local)
```

This error highlights the requirement for Go version >= 1.22.0, while we are currently using Go 1.21.8. 

## Ideal Solution (controller-runtime provide a envtest setup solution that allows us pin the tag releases)

Ideally, we need a way to setup `envtest` that allows version pinning per tag release. This would enable our community to manage dependencies more effectively and ensure compatibility across different versions of kubebuilder and `controller-runtime`. Note that in many scenarios we might need to do this setup in shell scripts. 

## Alternative Solution (how we could fix it and which seems the best available approach so far)

Use the release branch, example : `go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-17`

camilamacedo86:
@joelanford @varshaprasad96 @vincepri 

ping you folks to see if we can find a solution to provide the setup per release
because as you can see it is broken now.

lukas016:
proper versioning of setup-env is solving here https://github.com/kubernetes-sigs/controller-runtime/issues/2646
but maintainers don't react.

sbueringer:
> Hey folks,
we are currently experimenting a bit on setup-envtest because the envtest binaries are today hosted on the Google-owned kubebuilder repository.
> 
[...]

PR #3476, opened 2026-03-10
🌱 Auto-create git tags for the `tools/setup-envtest` submodule

porridge:
See [slack discussion](https://kubernetes.slack.com/archives/C02MRBMN00Z/p1770788038710919).

Using the GitHub CLI (gh api) built into GitHub's runners to create the tag, rather than using standard `git tag` and `git push` commands because it:
- does not require configuring a dummy git user/email, and
- works flawlessly in a shallow checkout.

Tested a [copy of this workflow](https://github.com/porridge/controller-runtime/blob/master/.github/workflows/tag-setup-envtest.yaml) in [this run](https://github.com/porridge/controller-runtime/actions/runs/22895347503/job/66427918078) in a clone repo.

### DIFF
--- a/.github/workflows/tag-setup-envtest.yaml
+++ b/.github/workflows/tag-setup-envtest.yaml
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+        with:
+          fetch-tags: true
 
       - name: Validate and Create Tag
         id: copy-tag
--- a/.github/workflows/tag-setup-envtest.yaml
+++ b/.github/workflows/tag-setup-envtest.yaml
@@ -13,7 +13,7 @@ jobs:
       
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
 
       - name: Validate and Create Tag
         id: copy-tag
--- a/.github/workflows/tag-setup-envtest.yaml
+++ b/.github/workflows/tag-setup-envtest.yaml
@@ -1,0 +1,26 @@
+name: Tag setup-envtest
+
+on:
+  push:
+    tags:
+      - '*' # Triggers on all tags, but the action skips non-semver ones.
+
+jobs:
+  create-mangled-tag:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # Required to create a new tag via API
+      
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Create mapped tag
+        id: map-tag
+        uses: stackrox/clone-tag@v1 
+        with:
+          prefix: 'tools/setup-envtest/'
+          
+      - name: Report result
+        if: steps.map-tag.outputs.skipped == 'false'
+        run: echo "Successfully generated tag ${{ steps.map-tag.outputs.new-tag }}"
--- a/.github/workflows/tag-setup-envtest.yaml
+++ b/.github/workflows/tag-setup-envtest.yaml
@@ -3,7 +3,7 @@ name: Tag setup-envtest
 on:
   push:
     tags:
-      - '*' # Triggers on all tags, but the action skips non-semver ones.
+      - 'v*' # Code below does further filtering.
 
 jobs:
   create-mangled-tag:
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-setup-envtest.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Validate and Create Tag
+        id: copy-tag
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
```
alvaroaleman: Please move the contents of this into a file under ./hack so its easier to run for testing

.github/workflows/tag-setup-envtest.yaml
```diff
@@ -0,0 +1,55 @@
+name: Tag setup-envtest
+
+on:
+  push:
+    tags:
+      - 'v*' # Code below does further filtering.
```
alvaroaleman: Any reason to not filter here already?

.github/workflows/tag-setup-envtest.yaml
```diff
@@ -0,0 +1,55 @@
...
+          
+          ORIGINAL_TAG="${GITHUB_REF#refs/tags/}"
+          REGEX="^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$"
+          
+          if [[ ! "$ORIGINAL_TAG" =~ $REGEX ]]; then
+            echo "Original tag '$ORIGINAL_TAG' does not match expression '$REGEX', skipping."
+            exit 0
+          fi
+    
+          echo "skipped=false" >> $GITHUB_OUTPUT
```
alvaroaleman: This seems to be unused

.github/workflows/tag-setup-envtest.yaml
```diff
@@ -0,0 +1,55 @@
...
+          NEW_TAG="${PREFIX}${ORIGINAL_TAG}"
+    
+          echo "Original tag '$ORIGINAL_TAG' matches expression '$REGEX'."
+          echo "Creating new tag: '$NEW_TAG' at SHA: ${{ github.sha }}"
```
alvaroaleman: Can we first check if the tag already exists?

.github/workflows/tag-setup-envtest.yaml
porridge: I think only basic globbing is available here, not terribly expressive...

.github/workflows/tag-setup-envtest.yaml
porridge: Sure. If it does, should we exit silently right away, or further check whether it's the same commit and vary the exit code based on that?

I'm leaning towards the former, in order to respect any pre-existing tags created out of band.

alvaroaleman: Yeah, just exit silently

.github/workflows/tag-setup-envtest.yaml
alvaroaleman: It appears to be an odd subset of regex that doesn't support match groups, but should be possible to validate for major/minor/patch here? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

porridge:
Sorry but I could not find the time for this recently, I will do my best to address the feedback next week.

porridge:
Feedback addressed, PTAL @alvaroaleman 

Tested here:
- [first run](https://github.com/porridge/controller-runtime/actions/runs/23533907154/job/68504153713)
- [subsequent re-run on the same tag](https://github.com/porridge/controller-runtime/actions/runs/23533907154/job/68504247910)

.github/workflows/tag-setup-envtest.yaml
porridge: I wasn't able to come up with an expression that accepts `v1.2.3-rc1` but rejects `v.1.2.3invalid` but perhaps it's not a big issue. Let me know if you'd like me to reinstate the regex match in bash.

alvaroaleman (approved):
thanks!

vitorfloriano:
Hi @porridge @alvaroaleman 

Question: would this also fix https://github.com/kubernetes-sigs/controller-runtime/issues/2744? 

We're currently pinning the download to release branches in our repo, but that raises security alerts in vulnerability scans. Pinning to the `controller-runtime` tag commit hash also works, but the `go` toolchain creates a pseudo version for that, which is not great for reproducibility.

Anyway, this may save a lot of developers from having to find a workaround for the missing nested submodule tagging. This also improves the use of `go tool` directives in `go.mod` and to get dependabots updates for that as well! 

Is this going to be released in a patch any soon? 

Great work!

PR #3477, opened 2026-03-10
:book: Improve help text for retries counter in workqueue

viveksb007:
Updated help text for the retries counter to provide more clarity on its current scope.

related slack thread in #sig-api-machinery https://kubernetes.slack.com/archives/C0EG7JC6T/p1772834493390239

this retry metric is getting incremented for cases where its not a true retry but queue additions with some delay. since changing metric name is a breaking change. thus improved the `Help` to inform future reader about the scope of this metric.

ex usage 
- https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/daemon/daemon_controller.go#L669

### DIFF
--- a/pkg/internal/metrics/workqueue.go
+++ b/pkg/internal/metrics/workqueue.go
@@ -95,7 +95,7 @@ var (
 	retries = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Subsystem: WorkQueueSubsystem,
 		Name:      RetriesKey,
-		Help:      "Total number of retries handled by workqueue",
+		Help:      "Total number of items added to the workqueue with a non-zero delay (rate-limited requeues, explicit RequeueAfter or AddAfter calls)",
 	}, []string{"name", "controller"})
 )
 


pkg/internal/metrics/workqueue.go
```diff
@@ -95,7 +95,7 @@ var (
 	retries = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Subsystem: WorkQueueSubsystem,
 		Name:      RetriesKey,
-		Help:      "Total number of retries handled by workqueue",
+		Help:      "Total number of items added to the workqueue with a non-zero delay (rate-limited requeues, explicit RequeueAfter or AddAfter calls)",
```
sbueringer, line 98: For me this is a bit of a confusing mixture between Reconciler and Queue wording.

Is this help text accurate both for PriorityQueue and the default queue that is used when PQ is not enabled?

For PriorityQueue the metric is incremented when AddAfter, AddRateLimitied or AddWithOpts (depends on opts) is called

PR #3481, opened 2026-03-17
✨ Disable the webhook server by setting the port to -1

### LINKED ISSUE #3456: Provide an easy way to turn off the webhook server
There is no easy way today to disable the webhook server. I think we should improve this.

I think there are use cases where no webhook server (and accordingly no conversion & admission webhooks) is needed.

In my current case I just wanted to run the reconcilers of a controller for debugging purposes.

The current situation is that if no WebhookServer is provided via Options.WebhookServer the manager will instantiate a default server.

The default server itself also has no way to disable it (e.g. via an explicit flag or setting the port to -1). For comparison the metrics server is disabled if BindAddress is set to 0.

The only way I found to disable the metrics server is to configure a "no-op" webhook server via options. Would be nice if that would be easier.

troy0820:
Resolves https://github.com/kubernetes-sigs/controller-runtime/issues/3456

Currently there is no way to disable the webhook server.  If we don't set the option, we get one automatically.  If we set the port to -1 it would set it to the default port.  

This change ~~adds a DisabledServer that Satisfies the webhook.Server interface and~~ allows the port to be set to -1 to signal the disabling of the webhook server.  

/assign @sbueringer

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -279,7 +279,7 @@ func (cm *controllerManager) GetAPIReader() client.Reader {
 func (cm *controllerManager) GetWebhookServer() webhook.Server {
 	cm.webhookServerOnce.Do(func() {
 		if cm.webhookServer == nil {
-			panic("webhook should not be nil")
+			return
 		}
 		if err := cm.Add(cm.webhookServer); err != nil {
 			panic(fmt.Sprintf("unable to add webhook server to the controller manager: %s", err))
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -279,7 +279,7 @@ func (cm *controllerManager) GetAPIReader() client.Reader {
 func (cm *controllerManager) GetWebhookServer() webhook.Server {
 	cm.webhookServerOnce.Do(func() {
 		if cm.webhookServer == nil {
-			return
+			panic("webhook should not be nil")
 		}
 		if err := cm.Add(cm.webhookServer); err != nil {
 			panic(fmt.Sprintf("unable to add webhook server to the controller manager: %s", err))
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -443,6 +443,9 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	errChan := make(chan error, 1)
 	runnables := newRunnables(options.BaseContext, errChan).withLogger(options.Logger)
+	if _, ok := options.WebhookServer.(*webhook.DisabledServer); ok {
+		options.WebhookServer = nil
+	}
 	return &controllerManager{
 		stopProcedureEngaged:          ptr.To(int64(0)),
 		cluster:                       cluster,
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -443,9 +443,6 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	errChan := make(chan error, 1)
 	runnables := newRunnables(options.BaseContext, errChan).withLogger(options.Logger)
-	if _, ok := options.WebhookServer.(*webhook.DisabledServer); ok {
-		options.WebhookServer = nil
-	}
 	return &controllerManager{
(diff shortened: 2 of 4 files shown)


troy0820:
Would we rather set it to something different and leave -1 the default port? Disabling the webhooks should be an option if the user wants to disable.

sbueringer:
Sorry for the delay. I'll take a look soon

pkg/webhook/server.go
```diff
@@ -75,6 +75,8 @@ type Options struct {
 
 	// Port is the port number that the server will serve.
 	// It will be defaulted to 9443 if unspecified.
+	//
+	// To disable the webhook server set Port to -1.
```
sbueringer, line 79: Given that 0 is the zero value and we previously used the DefaultPort in that case I guess -1 is our only option.

Unfortunately it's inconsistent with the metrics server (that uses 0 for disabled), but I don't have a better idea

@alvaroaleman is that fine for you as well or do you see an alternative?

pkg/webhook/server.go
```diff
@@ -105,6 +107,9 @@ type Options struct {
 
 // NewServer constructs a new webhook.Server from the provided options.
 func NewServer(o Options) Server {
+	if o.Port == -1 {
```
sbueringer: I'm wondering if instead of using a different server we should do the following:
* Stop defaulting the Port for -1
* Have branches that return early for Port -1 in Register, Start & StartedChecker

pkg/manager/internal.go
```diff
@@ -279,7 +279,7 @@ func (cm *controllerManager) GetAPIReader() client.Reader {
 func (cm *controllerManager) GetWebhookServer() webhook.Server {
 	cm.webhookServerOnce.Do(func() {
 		if cm.webhookServer == nil {
-			panic("webhook should not be nil")
+			return
```
sbueringer: Won't this panic now in a bunch of places where the caller assumed that it never gets nil?

(e.g. webhook builder)

I suspect we should keep having the webhook server, it just shouldn't do anything on Register, Start, ..

pkg/manager/manager.go
```diff
@@ -443,6 +443,9 @@ func New(config *rest.Config, options Options) (Manager, error) {
 
 	errChan := make(chan error, 1)
 	runnables := newRunnables(options.BaseContext, errChan).withLogger(options.Logger)
+	if _, ok := options.WebhookServer.(*webhook.DisabledServer); ok {
+		options.WebhookServer = nil
```
sbueringer: Related to my concern above in internal.go

pkg/webhook/server.go
troy0820: That makes sense,  it will stop the panicking as well.  I can change this but leave the port to -1 and have early returns

pkg/manager/manager_test.go
```diff
@@ -159,6 +159,16 @@ var _ = Describe("manger.Manager", func() {
...
+		It("should disable the webhook", func() {
+			By("setting the port to -1", func() {
+				m, err := New(cfg, Options{WebhookServer: webhook.NewServer(webhook.Options{Port: -1})})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(m).NotTo(BeNil())
+
+				svr := m.GetWebhookServer()
+				Expect(svr).To(BeNil())
+			})
+		})
```
troy0820: Removing because a nil server is not what we want.

pkg/webhook/server.go
troy0820: I am unaware of this history of why we would default to 9443 if the user set anything < 0, which to me seems like the user wants to not serve on 9443 if they are going through the trouble of setting it.  The 0 value I agree if they don't set it should be 9443, but anything less than that it seems the user is intentionally trying to set it to serve on a port that wouldn't work.

sbueringer: Agree!

sbueringer:
Thank you!

/lgtm
/assign @alvaroaleman

alvaroaleman (approved):

PR #3485, opened 2026-03-25
:bug: fakeclient: Allow updating managedFields through Update

### LINKED ISSUE #3484: Fake client: ManagedFields modified using csaupgrade.UpgradeManagedFields are not persisted correctly
The [`csaupgrade.UpgradeManagedFields`](https://github.com/kubernetes/client-go/blob/df2d882697f9f437d53ae16b8091786250ce0812/util/csaupgrade/upgrade.go#L80) utility is meant to upgrade existing resources from using client-side apply to server-side apply by modifying the managed fields so that "Update" entries of the manager that did CSA are updated to "Apply" using a new manager.

This transformation is done only in memory and is meant to be persisted to the server by the client. Using neither `client.Update(ctx, upgradedObj)` nor `client.Patch(ctx, origObj, client.MergeFrom(upgradedObj))` works correctly and the object ends up with an "Update" record in the managed fields instead of the expected "Apply".

I put together a simple reproducer for the bug:

https://github.com/metlos/fake-client-ssa-reproducer

metlos:
The managed fields could not be explicitly modified in Update or merge Patch in the fake client.

This is explicitly allowed and sometimes even necessary (e.g. upgrade from CSA to SSA) and so the fake client should support this.

Also, I think the handling of managed fields on subresources in general (apart from status) is not well supported yet, so I took the liberty of adding a pending test for that (and am happy to remove it if you don't want it as part of this PR).

This fixes #3484.

### DIFF
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -59,6 +59,7 @@ import (
 const (
 	machineIDFromStatusUpdate = "machine-id-from-status-update"
 	cidrFromStatusUpdate      = "cidr-from-status-update"
+	testOwner                 = "test-owner"
 )
 
 var _ = Describe("Fake client", func() {
@@ -1823,7 +1824,7 @@ var _ = Describe("Fake client", func() {
 			WithSpec(corev1applyconfigurations.NodeSpec().WithPodCIDR(initial.Spec.PodCIDR + "-updated")).
 			WithStatus(corev1applyconfigurations.NodeStatus().WithPhase(corev1.NodeRunning))
 
-		Expect(cl.Status().Apply(ctx, ac, client.FieldOwner("test-owner"))).To(Succeed())
+		Expect(cl.Status().Apply(ctx, ac, client.FieldOwner(testOwner))).To(Succeed())
 
 		actual := &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: initial.Name}}
 		Expect(cl.Get(ctx, client.ObjectKeyFromObject(actual), actual)).To(Succeed())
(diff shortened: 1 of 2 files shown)


pkg/client/fake/client_test.go
```diff
@@ -1952,6 +1953,62 @@ var _ = Describe("Fake client", func() {
 		Expect(node.ManagedFields[1].Manager).To(Equal("status-owner"))
 	})
 
+	// this is not working properly and can't without a larger change to the codebase
+	PIt("should not be able to manually update the managed fields through a subresource create,update", func(ctx SpecContext) {
```
alvaroaleman, line 1957: So this is a regression?

metlos: No, not a regression. I think this never really worked and I'm not 100% sure about the full extent of changes that would make this work (e.g. the special handling of the scale subresource in update, ...). That's why I asked in the issue description if you even want this test in this PR.

pkg/client/fake/client_test.go
```diff
@@ -3168,6 +3225,65 @@ var _ = Describe("Fake client", func() {
 		}
 	})
 
+	// GH-3484
+	It("respects the ManagedFields during create, update, merge patch", func(ctx SpecContext) {
```
alvaroaleman, line 3229: Please write a testcase per operation

pkg/client/fake/client_test.go
alvaroaleman, line 1957: This test is extremely confusing and tests two distinct things. It doesn't actually test that scale requests can not manipulate the objects managed fields, it tests that objects managed fields account correctly for scale, which is a different thing.

Please write a test instead that validates that setting managed fields on a scale request does not result in a change of the objects managed fields.

pkg/client/fake/client_test.go
```diff
@@ -59,6 +59,7 @@ import (
 const (
```
alvaroaleman, line 59: There is a compile failure in here

alvaroaleman:
/retitle :bug: fakeclient: Allow updating managedFields through Update

alvaroaleman:
@metlos any plans to wrap this up?

metlos:
> @metlos any plans to wrap this up?

Sorry, for the wait. I intend to finish this. I've just been short on time lately.

sbueringer:
@metlos *friendly reminder* :)

alvaroaleman:
Author doesn't seem to be working on this anymore

PR #3492, opened 2026-04-03
:bug: Add missing space in zap-log-level flag description

erikgb:
This adds a missing space to the `zap-log-level` flag description to avoid bad wording in operators inheriting the controller-runtime flags. This is how this flag output prints now:

````
-zap-log-level level                Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', 'panic'or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
````


alvaroaleman (approved):
thanks!

PR #3498, opened 2026-04-10
🐛 fakeclient: retry GenerateName on AlreadyExists collisions

### LINKED ISSUE #3497: Fake client should retry on GenerateName name collisions
## What happened?

The fake client's `Create` method generates a single random name when
`GenerateName` is set and does not retry on collision. If the generated
name matches an existing object in the tracker (including previously
deleted ones that the tracker still retains), it returns `AlreadyExists`
to the caller.

This causes flaky tests when many objects are created and deleted with
the same `GenerateName` prefix. With only a 5-character random suffix
(27^5 ≈ 14.3M possibilities), the birthday paradox makes collisions
likely at scale.

## What did you expect to happen?

The fake client should match the real Kubernetes API server behavior.
Since Kubernetes 1.32, the `RetryGenerateName` feature gate is GA and
the API server retries up to 7 times on `GenerateName` collisions:
https://github.com/kubernetes/kubernetes/issues/115489

The fake client should do the same — retry with a new random suffix
on `AlreadyExists` when `GenerateName` was used.

## Relevant code

The current single-attempt logic is in `pkg/client/fake/client.go`
in the `Create` method:

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/fake/client.go#L877-L882

The real API server's retry implementation is in:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go
(search for `generateName` / `retryCount`)

## Proposed fix

Wrap the name generation + `tracker.Create` call in a retry loop
(up to 7 attempts, matching the real API server). On `AlreadyExists`,
regenerate the random suffix and retry. Only retry when `GenerateName`
is set — explicit `Name` collisions should still return an error
immediately.

I'm happy to submit a PR for this.

vieux:
The real API server retries GenerateName collisions up to 7 times (RetryGenerateName feature gate, GA in K8s 1.32). The fake client now does the same, preventing flaky tests caused by birthday-paradox collisions when many objects share the same GenerateName prefix.

Closes https://github.com/kubernetes-sigs/controller-runtime/issues/3497

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -633,12 +633,12 @@ func (c *fakeClient) Create(ctx context.Context, obj client.Object, opts ...clie
 		return err
 	}
 
+	var generateNameBase string
 	if accessor.GetName() == "" && accessor.GetGenerateName() != "" {
-		base := accessor.GetGenerateName()
-		if len(base) > maxGeneratedNameLength {
-			base = base[:maxGeneratedNameLength]
+		generateNameBase = accessor.GetGenerateName()
+		if len(generateNameBase) > maxGeneratedNameLength {
+			generateNameBase = generateNameBase[:maxGeneratedNameLength]
 		}
-		accessor.SetName(fmt.Sprintf("%s%s", base, utilrand.String(randomLength)))
 	}
 	// Ignore attempts to set deletion timestamp
 	if !accessor.GetDeletionTimestamp().IsZero() {
@@ -653,10 +653,21 @@ func (c *fakeClient) Create(ctx context.Context, obj client.Object, opts ...clie
 	c.trackerWriteLock.Lock()
 	defer c.trackerWriteLock.Unlock()
 
-	if err := c.tracker.Create(gvr, obj, accessor.GetNamespace(), *createOptions.AsCreateOptions()); err != nil {
+	const maxRetries = 7
+	var createErr error
+	for range maxRetries {
+		if generateNameBase != "" {
+			accessor.SetName(fmt.Sprintf("%s%s", generateNameBase, utilrand.String(randomLength)))
+		}
+		createErr = c.tracker.Create(gvr, obj, accessor.GetNamespace(), *createOptions.AsCreateOptions())
+		if createErr == nil || generateNameBase == "" || !apierrors.IsAlreadyExists(createErr) {
+			break
+		}
+	}
+	if createErr != nil {
 		// The managed fields tracker sets gvk even on errors
 		_ = ensureTypeMeta(obj, gvk)
-		return err
+		return createErr
 	}
 
 	if !c.returnManagedFields {
--- a/pkg/client/fake/client_test.go
+++ b/pkg/client/fake/client_test.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
 	clientgoapplyconfigurations "k8s.io/client-go/applyconfigurations"
 	corev1applyconfigurations "k8s.io/client-go/applyconfigurations/core/v1"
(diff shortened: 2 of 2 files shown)


alvaroaleman:
@vieux please sign the CLA so we can accept the contribution

vieux:
@alvaroaleman 👍 I'm on it, waiting for my employer to approve

vieux:
@alvaroaleman cla is signed 👍

pkg/client/fake/client.go
```diff
@@ -653,10 +654,26 @@ func (c *fakeClient) Create(ctx context.Context, obj client.Object, opts ...clie
 	c.trackerWriteLock.Lock()
 	defer c.trackerWriteLock.Unlock()
 
-	if err := c.tracker.Create(gvr, obj, accessor.GetNamespace(), *createOptions.AsCreateOptions()); err != nil {
+	const maxRetries = 7
+	var createErr error
+	for i := 0; i < maxRetries; i++ {
+		if i > 0 {
+			// Regenerate name on retry
```
alvaroaleman: Please dedup the code that sets generate name. Also, the `base` here already has the suffix set from above, so isn't actually a base anymore

pkg/client/fake/client_test.go
```diff
@@ -485,6 +485,46 @@ var _ = Describe("Fake client", func() {
 			Expect(list.Items[0].Name).NotTo(BeEmpty())
 		})
 
+		It("should retry GenerateName on name collision and succeed", func(ctx SpecContext) {
+			// Create many objects with the same short GenerateName prefix so that
+			// birthday-paradox collisions are highly likely. The retry loop
+			// (max 7 attempts per object) must absorb every collision.
+			const n = 500
+			names := make(map[string]struct{}, n)
```
alvaroaleman: please use the sets package from apimachinery. Please make the test test that the name isn't longer than max base name length + length of the random suffix

alvaroaleman (commented):
/ok-to-test

pkg/client/fake/client.go
```diff
@@ -653,10 +647,25 @@ func (c *fakeClient) Create(ctx context.Context, obj client.Object, opts ...clie
 	c.trackerWriteLock.Lock()
 	defer c.trackerWriteLock.Unlock()
 
-	if err := c.tracker.Create(gvr, obj, accessor.GetNamespace(), *createOptions.AsCreateOptions()); err != nil {
+	const maxRetries = 7
+	var createErr error
+	for range maxRetries {
+		if usingGenerateName {
+			base := accessor.GetGenerateName()
```
alvaroaleman: This is still wrong, it will include the suffix on the second attempt. You need to store it outside the loop

pkg/client/fake/client_test.go
```diff
@@ -485,6 +486,47 @@ var _ = Describe("Fake client", func() {
...
+			for i := range n {
+				cm := &corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						GenerateName: "x",
+						Namespace:    "ns2",
+					},
+				}
+				Expect(cl.Create(ctx, cm)).To(Succeed(), "create #%d failed", i)
+				Expect(cm.Name).NotTo(BeEmpty())
+				Expect(len(cm.Name)).To(BeNumerically("<=", maxGeneratedNameLength+randomLength))
```
alvaroaleman: Please use a longer generate name so this test tests anything

pkg/client/fake/client.go
vieux: the suffix is added below on the set name and base is recreated, so I don't think it would include the suffix on the second attempt, but I agree, no need to get the full value + trim each time since it never changes.

alvaroaleman:
/retitle 🐛 fakeclient: retry GenerateName on AlreadyExists collisions

pkg/client/fake/client.go
alvaroaleman: Please leave it up to the author of a comment to decide when to resolve it, as resolving comments hides them

alvaroaleman (approved):
thanks!

alvaroaleman:
xref https://github.com/kubernetes/kubernetes/pull/122887

pkg/client/fake/client.go
vieux: @alvaroaleman noted, will do next time 👍

PR #3500, opened 2026-04-13
📖 webhook: clarify CertDir default

### LINKED ISSUE #900: ManagerOptions#CertDir default is confusing
# Problem Description

The default for the `CertDir` configuration option is presently nonsensical. I expect because it is out of date with other changes in the expected workflow for a developer to configure TLS. A prior PR #300 removed the cert provisioners that would create local credentials at `{TempDir}/k8s-webhook-server/serving-certs/tls.key` and `{TempDir}/k8s-webhook-server/serving-certs/tls.crt` This default was likely missed in the refactor due to some unexpected tight coupling to the prior implementation. However, it creates some confusion for a new developer when trying to run [the examples](https://github.com/kubernetes-sigs/controller-runtime/tree/master/examples).

[CertDir has the comment](https://github.com/kubernetes-sigs/controller-runtime/blob/a7c8a93c1cf395974911ef0ece977a7d27b2bf4b/pkg/manager/manager.go#L175)
```
// CertDir is the directory that contains the server key and certificate.
	// if not set, webhook server would look up the server key and certificate in
	// {TempDir}/k8s-webhook-server/serving-certs. The server key and certificate
	// must be named tls.key and tls.crt, respectively.
	CertDir string
```

In the logs running the example you will get:
```
unable to run manager	{"error": "open /var/folders/kc/7wscczc57v15s84cy1nx0d3h0000gn/T/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"}
```

# Developer Experience

A new developer is likely to run through the examples in the repository. This is what I was doing. The `CertDir` default sort of served as a red herring, masking the problem for awhile. The default value is so specific it seemed like something else was broken and the examples should work as written. However, in the current implementation it seems they require additional configuration.

# Possible Solution

Would love some feedback on the proper updates, but my inclination is to:
1. Remove the outdated default for `CertDir`
2. Update the comment on the type definition
3. Add a helpful logging message on receiving a `nil` value for `CertDir`
4. Update the examples to setup and reference local certs using `mkcert`
5. (maybe?) Change the signature for `manager.New` to reflect that option is not optional

# Additional Context

```
$ kc version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-14T04:24:29Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:07:57Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

```
$ go version
go version go1.13.8 darwin/amd64
```

# Related Prior Contributions
@mengqiy authored PR #300 and it was reviewed by @droot and @DirectXMan12 and all of them may likely have superior context to the past, present, and future state. Would love to have y'alls input and thank you for your contributions ❤️

alliasgher:
## Summary

Addresses the two small, agreed-upon asks from #900:

1. **Better docs.** The \`CertDir\` default — \`<temp-dir>/k8s-webhook-server/serving-certs\` — depends on \`os.TempDir()\`, which is \`/tmp\` on most Linux installs but is \`TMPDIR\`-dependent elsewhere. Maintainer asked to keep the default stable but update the comment to explain this and to nudge operators toward setting CertDir explicitly (usually to a mounted Secret path). Done.
2. **Helpful error.** When \`certwatcher.New\` fails because the certs weren't where \`CertDir\` was pointed, users got a bare \`open ...: no such file or directory\`. Wrap the error so it names the paths that were attempted, prints the effective \`CertDir\`/\`CertName\`/\`KeyName\`, and points at \`webhook.Server.Options.CertDir\` as the thing to fix. The underlying error is preserved via \`%w\`.

### Before

\`\`\`
open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory
\`\`\`

### After

\`\`\`
failed to load serving cert from \"/tmp/k8s-webhook-server/serving-certs/tls.crt\" / \"/tmp/k8s-webhook-server/serving-certs/tls.key\" — did you mount the certificate files at webhook.Server.Options.CertDir? (CertDir=\"/tmp/k8s-webhook-server/serving-certs\", CertName=\"tls.crt\", KeyName=\"tls.key\"): open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory
\`\`\`

Fixes #900

## Tests

\`go build ./pkg/webhook/...\` and \`go vet ./pkg/webhook/...\` pass. No behavior change except the added context on the error path.


sbueringer:
Looks like this needs some test fixes, PTAL

alliasgher:
Thanks @sbueringer. Fixed in 0e057fb1 — the builder webhook tests gate on `os.IsNotExist(err)` to tolerate the missing serving cert, but `os.IsNotExist` doesn't unwrap, so the new `fmt.Errorf(...: %w)` wrapper no longer matched. Switched those checks to `errors.Is(err, fs.ErrNotExist)`, which unwraps and still matches the underlying `os.PathError`. `TestBuilder` passes locally with envtest.

sbueringer:
> Thanks @sbueringer. Fixed in [0e057fb](https://github.com/kubernetes-sigs/controller-runtime/commit/0e057fb1b251cabcef1c89271287b2230aa35a4e) — the builder webhook tests gate on `os.IsNotExist(err)` to tolerate the missing serving cert, but `os.IsNotExist` doesn't unwrap, so the new `fmt.Errorf(...: %w)` wrapper no longer matched. Switched those checks to `errors.Is(err, fs.ErrNotExist)`, which unwraps and still matches the underlying `os.PathError`. `TestBuilder` passes locally with envtest.

So this PR would break folks that are using os.IsNotExist.

I'm not really sure if that change in error message is worth it. I don't think a lot of folks are using os.IsNotExist, but I also don't think that the current error reporting is a real problem that folks have.

@alvaroaleman WDYT?

alvaroaleman:
> @alvaroaleman WDYT?

Yeah I wouldn't change it

sbueringer:
@alliasgher Based on discussion above, if you can revert the change to the error so the PR only contains the improvement to the godoc we can go ahead with this PR

alliasgher:
Reverted the error change in 6e12450; the PR is now just the CertDir godoc clarification.

sbueringer:
@alliasgher Please change or drop the icon in the commit message, this is not a bugfix

alliasgher:
Changed it to 📖 since this is a godoc clarification, not a bugfix. Done in b28c1cf.

alliasgher:
@sbueringer I closed this by mistake, sorry about that. Could you reopen it, or should I open a new one? The icon is fixed (📖).

sbueringer:
@alliasgher The reopen button is greyed out for me, let's just create a new PR

alliasgher:
Opened #3551 to replace this.

PR #3509, opened 2026-04-27
:warning: Add AnnotatedEventRecorderProvider interface

adri1197:
Adds support for creating event recorders that can attach custom annotations to events,
mirroring the `AnnotatedEventf` capability recently added to `client-go/tools/events`
(see kubernetes/kubernetes#138103).

This introduces a new `AnnotatedEventRecorderProvider` interface in `pkg/recorder` rather
than adding a method to the existing `Provider` interface, so this is a backwards-compatible
change. Consumers can type-assert the manager/cluster to access the new functionality:

```go
if p, ok := mgr.(recorder.AnnotatedEventRecorderProvider); ok {
    rec := p.GetAnnotatedEventRecorder("my-controller")
    rec.AnnotatedEventf(obj, nil, map[string]string{"key": "value"}, corev1.EventTypeNormal, "Reason", "Action", "Note")
}
```

Changes:
- Added AnnotatedEventRecorderProvider interface in pkg/recorder
- Implemented GetAnnotatedEventRecorder on the internal recorder Provider
- Implemented AnnotatedEventf on lazyRecorder
- Added GetAnnotatedEventRecorder on cluster and controllerManager (concrete types only, no interface changes)
- Added unit and integration tests

### DIFF
--- a/pkg/cluster/internal.go
+++ b/pkg/cluster/internal.go
@@ -92,6 +92,10 @@ func (c *cluster) GetEventRecorder(name string) events.EventRecorder {
 	return c.recorderProvider.GetEventRecorder(name)
 }
 
+func (c *cluster) GetAnnotatedEventRecorder(name string) events.AnnotatedEventRecorder {
+	return c.recorderProvider.GetAnnotatedEventRecorder(name)
+}
+
 func (c *cluster) GetRESTMapper() meta.RESTMapper {
 	return c.mapper
 }
--- a/pkg/cluster/internal.go
+++ b/pkg/cluster/internal.go
@@ -24,12 +24,12 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/events"
 	"k8s.io/client-go/tools/record"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	intrec "sigs.k8s.io/controller-runtime/pkg/internal/recorder"
+	"sigs.k8s.io/controller-runtime/pkg/recorder"
 )
 
 type cluster struct {
@@ -88,12 +88,8 @@ func (c *cluster) GetEventRecorderFor(name string) record.EventRecorder {
 	return c.recorderProvider.GetEventRecorderFor(name)
 }
 
-func (c *cluster) GetEventRecorder(name string) events.EventRecorder {
+func (c *cluster) GetEventRecorder(name string) recorder.EventRecorder {
 	return c.recorderProvider.GetEventRecorder(name)
-}
-
-func (c *cluster) GetAnnotatedEventRecorder(name string) events.AnnotatedEventRecorder {
-	return c.recorderProvider.GetAnnotatedEventRecorder(name)
 }
 
 func (c *cluster) GetRESTMapper() meta.RESTMapper {
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -156,6 +156,12 @@ var _ = Describe("cluster.Cluster", func() {
 		Expect(c.GetFieldIndexer()).To(Equal(cluster.cache))
 	})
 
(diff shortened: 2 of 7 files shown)


matheuscscp (approved):
LGTM 🙌

sbueringer:
@adri1197 We just bumped to v1.37 alpha, so I think this PR can be rebased and moved out of draft

adri1197:
> @adri1197 We just bumped to v1.37 alpha, so I think this PR can be rebased and moved out of draft

@sbueringer since you mentioned this can be rebased and moved out of draft now that we're on v1.37 alpha — would it be acceptable to add the method directly to the Provider interface as a breaking change for the next minor (given controller-runtime's v0.x semver policy allows breaking changes between minors)? Or should we stick with the separate? interface approach from the PR description to keep it non-breaking?

pkg/cluster/internal.go
```diff
@@ -92,6 +92,10 @@ func (c *cluster) GetEventRecorder(name string) events.EventRecorder {
 	return c.recorderProvider.GetEventRecorder(name)
 }
 
+func (c *cluster) GetAnnotatedEventRecorder(name string) events.AnnotatedEventRecorder {
```
sbueringer: I'm wondering if we should really have an additional Get*EventRecorder method instead of just combining both "new" event recorder in an interface in controller-runtime.

I think this is an acceptable breaking change either way

I assume it's easier to use if we just have one method / one recorder. Because folks then don't have to get and pass around two different recorders if they want to use both methods.

@alvaroaleman Do you have an opinion on this?

Context: The old upstream EventRecorder had Event / Eventf / AnnotatedEventf methods. The new API has one interface each for Eventf / AnnotatedEventf

sbueringer:
> > @adri1197 We just bumped to v1.37 alpha, so I think this PR can be rebased and moved out of draft
> 
> @sbueringer since you mentioned this can be rebased and moved out of draft now that we're on v1.37 alpha — would it be acceptable to add the method directly to the Provider interface as a breaking change for the next minor (given controller-runtime's v0.x semver policy allows breaking changes between minors)? Or should we stick with the separate? interface approach from the PR description to keep it non-breaking?

Sorry just saw this message now. This is the same point I'm bringing up https://github.com/kubernetes-sigs/controller-runtime/pull/3509#discussion_r3434046037, right?

EDIT: Ah I think it's a slightly different point

pkg/cluster/internal.go
matheuscscp: Agreed!

adri1197:
> > > @adri1197 We just bumped to v1.37 alpha, so I think this PR can be rebased and moved out of draft
> > 
> > 
> > @sbueringer since you mentioned this can be rebased and moved out of draft now that we're on v1.37 alpha — would it be acceptable to add the method directly to the Provider interface as a breaking change for the next minor (given controller-runtime's v0.x semver policy allows breaking changes between minors)? Or should we stick with the separate? interface approach from the PR description to keep it non-breaking?
> 
> Sorry just saw this message now. This is the same point I'm bringing up [#3509 (comment)](https://github.com/kubernetes-sigs/controller-runtime/pull/3509#discussion_r3434046037), right?
> 
> EDIT: Ah I think it's a slightly different point

No really, it's exactly the answer that I wanted to get 😃. I had some doubts about having different Recorders (well, Providers) per interface.

I've reworked this to combine both into a single `recorder.EventRecorder` interface that embeds `events.EventRecorder` + `events.AnnotatedEventRecorder`. So now `GetEventRecorder` returns one recorder that does both - no separate method needed. Will push shortly 😄.

pkg/cluster/internal.go
adri1197: Friendly ping @sbueringer @alvaroaleman 👋 — the rework combining both into a single `recorder.EventRecorder` interface (per @sbueringer's suggestion) has been pushed. 

Would you have a chance to take another look?

Thanks! 😃

sbueringer: Thx! I'll try to take a look soon, probably tomorrow

sbueringer:
Thank you very much!!

/lgtm

/assign @alvaroaleman

alvaroaleman:
Marked the change as breaking since it changes the interface, but otherwise lgtm, thank you
/hold cancel

PR #3510, opened 2026-04-27
✨Allow opt-in for client-go REST client metrics

richabanker:
Introduce opt-in mechanism for client-go REST client metrics
These metrics were previously not supported by default due to high cardinality risks (ref: https://github.com/kubernetes-sigs/controller-runtime/pull/2298)

This PR introduces an opt-in mechanism allowing downstream users to selectively enable these metrics if needed, while still preventing `sync.Once` conflicts by registering the adapters early.

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -18,6 +18,10 @@ package metrics
 
 import (
 	"context"
+	"net/url"
+	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	clientmetrics "k8s.io/client-go/tools/metrics"
@@ -27,6 +31,11 @@ import (
 // that client-go registers metrics.  We copy the names and formats
 // from Kubernetes so that we match the core controllers.
 
+const (
+	hostLabel = "host"
+	verbLabel = "verb"
+)
+
 var (
 	// client metrics.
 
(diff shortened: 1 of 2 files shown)


richabanker:
cc @alvaroaleman

pkg/metrics/client_go_adapter.go
```diff
@@ -37,8 +41,68 @@ var (
 		},
 		[]string{"code", "method", "host"},
 	)
+
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb and host.",
+			Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
```
alvaroaleman: Can we support native histograms from the get-go?

pkg/metrics/client_go_adapter.go
```diff
@@ -37,8 +41,68 @@ var (
...
+			Help: "Number of request retries, partitioned by status code, verb, and host.",
+		},
+		[]string{"code", "verb", "host"},
+	)
 )
 
+// restMetricsEnabled gates the opt-in adapters below, they discard observations
+// until RegisterDefaultRESTClientMetrics flips it. requestResult stays on
+// unconditionally to preserve existing behavior.
+var restMetricsEnabled atomic.Bool
```
alvaroaleman: Would it help readability if we used an atomic pointer that switches the entire adapterset rather than having a check in each method? Is there a no-op from client-go we can reuse?

pkg/metrics/client_go_adapter_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+package metrics
+
+import (
+	"context"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
```
sbueringer: I would prefer if we can avoid introducing testify just for two checks. Let's align to our existing tests please

pkg/metrics/client_go_adapter.go
sbueringer: (see the native histogram config for controller_runtime_reconcile_time_seconds)

pkg/metrics/client_go_adapter.go
```diff
@@ -50,7 +114,30 @@ func registerClientMetrics() {
...
+		ResponseSize:       &sizeAdapter{metric: responseSize},
+		RateLimiterLatency: &latencyAdapter{metric: rateLimiterLatency},
+		RequestRetry:       &retryAdapter{metric: requestRetry},
+	})
+}
+
+var registerDefaultsOnce sync.Once
+
+// RegisterDefaultRESTClientMetrics enables the client metrics.
+func RegisterDefaultRESTClientMetrics() {
```
sbueringer: Could we make this more flexible so it's possible to specify exactly which metrics to enable?

In Cluster API for example we probably cannot enable all of these metrics because at scale we're communicating with 1k+ different apiservers. So it would be nice to be able to enable only specific metrics.

(not sure if there's prior art somewhere for on API that allows to enable a set of metrics)

(cc @alvaroaleman)

pkg/metrics/client_go_adapter.go
richabanker: done

pkg/metrics/client_go_adapter.go
richabanker: no client-go doesnt have any public no-op types that we can use here... 
I dont quite understand how an atomic pointer for the whole adapterset helps with readability.. clientmetrics.Register is sync.Once guarded so we wont be able to swap the values effectively?

pkg/metrics/client_go_adapter_test.go
richabanker: done

pkg/metrics/client_go_adapter.go
richabanker: Thats reasonable.. added a `func RegisterRESTClientMetrics(metrics ...RESTClientMetric)`  where RESTClientMetric is an enum supporting existing metric types. wdyt of this?

pkg/metrics/client_go_adapter.go
```diff
@@ -35,10 +44,115 @@ var (
...
 
+// RESTClientMetric identifies an opt-in client-go REST client metric.
+// Pass values to RegisterRESTClientMetrics to enable a subset.
+type RESTClientMetric int
+
+const (
+	// MetricRequestLatency enables the rest_client_request_duration_seconds metric.
+	MetricRequestLatency = iota + 1
+	// MetricResolverLatency enables the rest_client_dns_resolution_duration_seconds metric.
+	MetricResolverLatency
```
sbueringer:
```suggestion
	MetricDNSResolutionLatency
```
Maybe?

pkg/metrics/client_go_adapter.go
sbueringer: Looks good, thx!

sbueringer (commented):
Thx! Last nit from my side

pkg/metrics/client_go_adapter.go
richabanker: Done!

richabanker:
@sbueringer @alvaroaleman I think all comments are addressed now, please lmk if there are any other outstanding concerns. 

Thank you for the review!

pkg/metrics/client_go_adapter.go
sbueringer: Thx!

sbueringer:
Thank you very much!!

/lgtm
/approve

/hold
@alvaroaleman in case you want to take another look

pkg/metrics/client_go_adapter.go
```diff
@@ -50,10 +164,61 @@ func registerClientMetrics() {
...
+		RequestSize:        &sizeAdapter{metric: requestSize, enabled: &requestSizeEnabled},
+		ResponseSize:       &sizeAdapter{metric: responseSize, enabled: &responseSizeEnabled},
+		RateLimiterLatency: &latencyAdapter{metric: rateLimiterLatency, enabled: &rateLimiterLatencyEnabled},
+		RequestRetry:       &retryAdapter{metric: requestRetry, enabled: &requestRetryEnabled},
 	})
 }
 
+// RegisterRESTClientMetrics enables the client metrics.
+func RegisterRESTClientMetrics(metrics ...RESTClientMetric) {
+	seen := map[RESTClientMetric]bool{}
```
alvaroaleman: What is the point of this map given we use a sync.Once?

richabanker: oh right, thats unnecessary given the sync.Once will prevent duplicate registrations. Removed now, thanks!

alvaroaleman (approved):
/hold cancel
thanks!

PR #3512, opened 2026-05-03
🐛 pkg/envtest: validate config when handling CRDs

orangecms:
The error message is adapted from pkg/client.
Otherwise, if nil is passed, it crashes with a longer stack trace.

See https://mastodon.social/@CyReVolt/116507340236420167

### DIFF
--- a/pkg/envtest/crd.go
+++ b/pkg/envtest/crd.go
@@ -90,6 +90,10 @@ const (
 
 // InstallCRDs installs a collection of CRDs into a cluster by reading the crd yaml files from a directory.
 func InstallCRDs(config *rest.Config, options CRDInstallOptions) ([]*apiextensionsv1.CustomResourceDefinition, error) {
+	if config == nil {
+		return nil, fmt.Errorf("must provide non-nil rest.Config to InstallCRDs")
+	}
+
 	defaultCRDOptions(&options)
 
 	// Read the CRD yamls into options.CRDs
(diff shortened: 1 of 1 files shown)


pkg/envtest/crd.go
```diff
@@ -90,6 +90,10 @@ const (
 
 // InstallCRDs installs a collection of CRDs into a cluster by reading the crd yaml files from a directory.
 func InstallCRDs(config *rest.Config, options CRDInstallOptions) ([]*apiextensionsv1.CustomResourceDefinition, error) {
+	if config == nil {
```
alvaroaleman, line 93: hi @orangecms, thanks for trying to help out but I am not in favor of merging this. Assuming that nil is a valid value for pointer arguments to a function or method when the arg is not explicitly documented as such is generally wrong and a programming mistake. Panicing is how the function signals this so the author can correct that mistake.

If you look at the go stdlib, you will find that nilchecking parameters is not commonly done.

/hold

orangecms: Hi, can you help me understand?

pkg/client does have that check, so it would be inconsistent to sometimes have the check, sometimes not.

Wouldn't the error returned be what allows for graceful handling and obtaining an understanding of the error? What would otherwise be the reason to have nil checks, such as in pkg/client?

alvaroaleman: Yep, seems to be an oversight, the rest of the codebase doesn't have such checks and we don't plan on changing that.

PR #3515, opened 2026-05-08
:bug: Fix regression in Apply typed error handling

erikgb:
When attempting to upgrade K8s dependencies in multiple operators, I noticed what seems to be a regression in the error handling for the Apply method in the controller-runtime client. Sadly, only one of the operators I've been having issues with is open source, but the problem can be observed in the failing CI of https://github.com/statnett/image-scanner-operator/pull/1729. This operator contains compensating logic for the `RequestEntityTooLarge` error (HTTP status 413), which is broken after upgrading K8s deps (including controller-runtime).

In one of our closed-source operators, I observe a similar issue, which is likely to be related. There we have compensating logic if trying to modify an immutable field. The tests are failing for the upgrade PR with the following logged inside the operator:

````json
{"level":"error","ts":"2026-05-04T08:50:50Z","msg":"Reconciler error","controller":"application","controllerGroup":"stas.statnett.no","controllerKind":"Application","Application":{"name":"v1beta2-immutable-field","namespace":"application-operator"},"namespace":"application-operator","name":"v1beta2-immutable-field","reconcileID":"45487740-546a-427e-a4d4-f826bf4042eb","error":"the server rejected our request due to an error in our request (patch deployments.apps v1beta2-immutable-field)","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.0/pkg/internal/controller/controller.go:494\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.0/pkg/internal/controller/controller.go:437\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.0/pkg/internal/controller/controller.go:312"}
````

I believe the regression was introduced in https://github.com/kubernetes-sigs/controller-runtime/pull/3475, which represents a good improvement/feature. But I think the error handling was dramatically changed by the PR.

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -38,6 +38,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -1100,6 +1101,26 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 
 				Expect(secret.Data).To(BeComparableTo(data))
 				Expect(secret.Data).To(BeComparableTo(secretApplyConfiguration.Data))
+			})
+
+			It("should propagate a typed error", func(ctx SpecContext) {
+				cl, err := client.New(cfg, client.Options{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(cl).NotTo(BeNil())
+
+				obj := corev1applyconfigurations.
+					Secret("typed-secret", "default").
+					WithType(corev1.SecretTypeDockercfg).
+					WithStringData(map[string]string{".dockercfg": "{}"})
+
+				err = cl.Apply(ctx, obj, &client.ApplyOptions{FieldManager: "test-manager"})
+				Expect(err).ToNot(HaveOccurred())
+
+				obj = corev1applyconfigurations.
+					Secret(*obj.Name, *obj.Namespace).
+					WithType(corev1.SecretTypeOpaque)
+				err = cl.Apply(ctx, obj, &client.ApplyOptions{FieldManager: "test-manager"})
+				Expect(isImmutableError(err)).To(BeTrue())
 			})
 		})
 	})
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/client/client_test.go [hunk 1]
```diff
@@ -1101,6 +1102,25 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
...
+					b[i] = byte(rand.Int())
+				}
+				largeString := hex.EncodeToString(b)
+
+				obj := corev1applyconfigurations.
+					ConfigMap("large-configmap", "default").
+					WithData(map[string]string{"payload": largeString})
+
+				err = cl.Apply(ctx, obj, &client.ApplyOptions{FieldManager: "test-manager"})
+				Expect(err).To(MatchError(apierrors.IsRequestEntityTooLargeError, "IsRequestEntityTooLargeError"))
```
alvaroaleman:
```suggestion
				Expect(apierrors.IsRequestEntityTooLargeError(err)).To(BeTrue()
```

erikgb: Good feedback! No need to use the fancier Gomega features. 😅

pkg/client/typed_client.go
```diff
@@ -147,19 +147,21 @@ func (c *typedClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration,
...
-		ContentType(&contentType).
-		Raw()
-	if err != nil {
+		Do(ctx)
+	if err := resp.Error(); err != nil {
 		return err
 	}
 
+	var contentType string
+	body, _ := resp.
```
sbueringer: I would prefer to check the error here and return if it's not nil

I know with the current implementation of this methods we should not get an error here but I don't want to rely on that

erikgb: Good point, will fix! Relying on looking into the current method is not a good idea.

erikgb: Fixed.

pkg/client/client_test.go
```diff
@@ -4433,3 +4454,8 @@ func toUnstructured(o client.Object) (*unstructured.Unstructured, error) {
 	u := &unstructured.Unstructured{}
 	return u, json.Unmarshal(serialized, u)
 }
+
+func IsImmutableError(err error) bool {
```
alvaroaleman: Can we unexport this? Other than that, lgtm

erikgb: Done! Good catch!

alvaroaleman (approved):
thanks!

PR #3520, opened 2026-05-15
✨ Add TypedInformer source

cezarsa:
This PR mimics the existing pattern of having a typed version of each source. This small change would help users avoid having to write their own version of `TypedInformer`.

For an example of a use-case, see: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4678/changes#diff-aeaa3ed265c13311667100ad572f071fc3928d07227e8320322116bfee2a0e90

### DIFF
--- a/pkg/internal/source/event_handler.go
+++ b/pkg/internal/source/event_handler.go
@@ -35,7 +35,7 @@ var log = logf.RuntimeLog.WithName("source").WithName("EventHandler")
 var _ cache.ResourceEventHandler = &EventHandler[client.Object, any]{}
 
 // NewEventHandler creates a new EventHandler.
-func NewEventHandler[object client.Object, request comparable](
+func NewEventHandler[object any, request comparable](
 	ctx context.Context,
 	queue workqueue.TypedRateLimitingInterface[request],
 	handler handler.TypedEventHandler[object, request],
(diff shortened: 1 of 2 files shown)


pkg/internal/source/event_handler.go
```diff
@@ -49,7 +49,7 @@ func NewEventHandler[object client.Object, request comparable](
 }
 
 // EventHandler adapts a handler.EventHandler interface to a cache.ResourceEventHandler interface.
-type EventHandler[object client.Object, request comparable] struct {
+type EventHandler[object any, request comparable] struct {
```
alvaroaleman, line 52: Why is this being changed?

cezarsa: Because the definition of `handler.TypedEventHandler` doesn't constrain the object and allows `any`, neither does any of the source generic types. Without that change, the `TypedInformer` definition would also have to be constrained to `client.Object` which wouldn't be desirable, and it would limit its use.

alvaroaleman: Why would it limit its use? What use-case exists where TypedInformer is typed to something that does not satisfy client.Object?

cezarsa: I have one here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/4678/changes#diff-8ce64d4b8a4655d407ba70bf654823f4f4d4e7a7036da24f85d1b1fe99bf6aebR281

It's possible to use `cache.SharedIndexInformer{}.SetTransform()` to transform the original object into an instance that doesn't satisfy `client.Object`.

alvaroaleman (approved):
thanks!

PR #3521, opened 2026-05-21
📖docs: improve MaxConcurrentReconciles documentation

### LINKED ISSUE #1416: Improve doc over MaxConcurrentReconciles
Would be nice be very nice we are able to provide further information for MaxConcurrentReconciles [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L37) for we have a better go doc [here](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller#Options).  

People have been asking when/how using this option. We might add that `It is not possible to have two or more reconcile loops that handle the same object at the same time` and some other further explanations. See that people have also been providing further docs about the topic and would be nice to have something official. e.g: https://openkruise.io/en-us/blog/blog2.html

mattsu2020:
## Summary

This improves the GoDoc for `MaxConcurrentReconciles` to address issue #1416, which requested better documentation explaining when and how to use this option.

The added documentation explains:
- The option controls the number of worker goroutines processing items from the controller's work queue
- Different queue items may be reconciled concurrently, but the same item is never processed by multiple workers at the same time
- If the same item is added again while being processed, it is marked dirty and requeued after the current reconciliation finishes
- For the default `reconcile.Request` type, the item key is the object's namespace/name
- Increasing concurrency can improve throughput but may increase load on the API server and external systems; QPS/Burst settings should be considered
- Manager-level (`config.Controller.MaxConcurrentReconciles`) and GroupKind-level (`config.Controller.GroupKindConcurrency`) configuration are also available; per-controller values take precedence

## Changes

| File | Change |
|------|--------|
| `pkg/controller/controller.go` | Detailed doc on `TypedOptions.MaxConcurrentReconciles` (primary documentation) |
| `pkg/config/controller.go` | Doc on manager-level default with reference to primary doc |
| `pkg/internal/controller/controller.go` | Internal `Options` and `Controller` structs reference primary doc |

Closes #1416

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -54,7 +54,7 @@ type TypedOptions[request comparable] struct {
 	// many objects or when reconciliation involves slow operations such as external
 	// API calls.
 	//
-	// The work queue ensures that the same item is not processed by multiple
+	// The workqueue ensures that the same item is not processed by multiple
 	// workers at the same time. If the same item is added again while it is being
 	// processed, it is marked dirty and requeued after the current reconciliation
 	// finishes. For the default reconcile.Request type, the item key is the
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -49,7 +49,7 @@ type TypedOptions[request comparable] struct {
 	// that can be run. Defaults to 1.
 	//
 	// This value controls the number of worker goroutines that process items from
-	// the controller's work queue. Increasing it allows different queue items to be
+	// the controller's workqueue. Increasing it allows different queue items to be
 	// reconciled in parallel, which can improve throughput when a controller manages
 	// many objects or when reconciliation involves slow operations such as external
 	// API calls.
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -45,7 +45,31 @@ type TypedOptions[request comparable] struct {
 	// Defaults to false if Controller.SkipNameValidation setting from the Manager is also unset.
 	SkipNameValidation *bool
 
-	// MaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run. Defaults to 1.
+	// MaxConcurrentReconciles is the maximum number of concurrent reconciliations
+	// that can be run. Defaults to 1.
+	//
+	// This value controls the number of worker goroutines that process items from
+	// the controller's work queue. Increasing it allows different queue items to be
+	// reconciled in parallel, which can improve throughput when a controller manages
+	// many objects or when reconciliation involves slow operations such as external
+	// API calls.
+	//
+	// The work queue ensures that the same item is not processed by multiple
+	// workers at the same time. If the same item is added again while it is being
+	// processed, it is marked dirty and requeued after the current reconciliation
+	// finishes. For the default reconcile.Request type, the item key is the
+	// object's namespace/name.
+	//
+	// Higher concurrency can increase load on the Kubernetes API server and any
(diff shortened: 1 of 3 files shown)


mattsu2020:
Hi @joelanford @varshaprasad96,

Could you please take a look when you have a chance?

This is a docs-only PR for #1416 that improves the GoDoc around `MaxConcurrentReconciles`. 
Since this is my first PR to controller-runtime, it still needs `/ok-to-test` before CI can run.

Thank you!

pkg/controller/controller.go
```diff
@@ -45,7 +45,31 @@ type TypedOptions[request comparable] struct {
...
+	// MaxConcurrentReconciles is the maximum number of concurrent reconciliations
+	// that can be run. Defaults to 1.
+	//
+	// This value controls the number of worker goroutines that process items from
+	// the controller's work queue. Increasing it allows different queue items to be
+	// reconciled in parallel, which can improve throughput when a controller manages
+	// many objects or when reconciliation involves slow operations such as external
+	// API calls.
+	//
+	// The work queue ensures that the same item is not processed by multiple
```
alvaroaleman:
```suggestion
	// The workqueue ensures that the same item is not processed by multiple
```

pkg/controller/controller.go
```diff
@@ -45,7 +45,31 @@ type TypedOptions[request comparable] struct {
...
+	// workers at the same time. If the same item is added again while it is being
+	// processed, it is marked dirty and requeued after the current reconciliation
```
alvaroaleman: Remove the comment about "being marked dirty", that is a) an implemention detail and b) not actually the case when using the priorityqueue, its just called `locked` there

pkg/controller/controller.go
```diff
@@ -45,7 +45,31 @@ type TypedOptions[request comparable] struct {
...
+	// finishes. For the default reconcile.Request type, the item key is the
+	// object's namespace/name.
+	//
+	// Higher concurrency can increase load on the Kubernetes API server and any
```
alvaroaleman: I don't find this a useful section, because it depends a lot on what exactly the reconciler does. The bit about QPS/Burst is also not useful, because we default to disable the clientside ratelimiter and rely on APF

pkg/controller/controller.go
mattsu2020: Thank you
Edit comment

pkg/controller/controller.go
mattsu2020: Thank you.
I've deleted the relevant part of the comment.

pkg/controller/controller.go
```diff
@@ -45,7 +45,26 @@ type TypedOptions[request comparable] struct {
 	// Defaults to false if Controller.SkipNameValidation setting from the Manager is also unset.
 	SkipNameValidation *bool
 
-	// MaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run. Defaults to 1.
+	// MaxConcurrentReconciles is the maximum number of concurrent reconciliations
+	// that can be run. Defaults to 1.
+	//
+	// This value controls the number of worker goroutines that process items from
+	// the controller's work queue. Increasing it allows different queue items to be
```
alvaroaleman:
```suggestion
	// the controller's workqueue. Increasing it allows different queue items to be
```

alvaroaleman (approved):
/hold

/assign @sbueringer

sbueringer:
Thx!

/lgtm
/approve

mattsu2020:
thanks

PR #3523, opened 2026-05-21
🐛Fix process Stop timeout error handling

### LINKED ISSUE #3522: Stop() timeout error swallowed by SIGKILL EPERM on Darwin
## What happened

On Darwin (macOS), `State.Stop()` in `pkg/internal/testing/process` may return `"unable to kill process ..."` instead of the expected `"timeout waiting for process ... to stop"` error when `StopTimeout` expires.

This causes test failures in `process_test.go` where the test asserts the error contains `"timeout"`:

```go
Expect(processState.Stop()).To(MatchError(ContainSubstring("timeout")))
```

## What you expected to happen

`Stop()` should always return a timeout error when `StopTimeout` expires, regardless of whether the subsequent `SIGKILL` succeeds or fails.

## How to reproduce it

On macOS, run:

```bash
go test ./pkg/internal/testing/process/ -run "Stop.*cannot be stopped" -count=100
```

The test is flaky — it passes when `SIGKILL` succeeds, but fails when the process has already exited between the timeout and the `SIGKILL` attempt.

## Root cause

In `process.go`, the `Stop()` method's timeout case:

```go
case <-timedOut:
    if err := signalProcess(ps.Cmd.Process, syscall.SIGKILL); err != nil {
        return fmt.Errorf("unable to kill process %s: %w", ps.Path, err)
    }
    return fmt.Errorf("timeout waiting for process %s to stop", path.Base(ps.Path))
```

When the timeout fires:
1. `SIGTERM` was already sent, and the process is shutting down
2. `SIGKILL` is sent via `syscall.Kill(-process.Pid, syscall.SIGKILL)` (in `signal_unix.go`)
3. On Darwin, if the process is already in its termination sequence, `syscall.Kill` returns `EPERM`
4. The `"unable to kill process"` error is returned instead of the `"timeout waiting for process to stop"` error

The `SIGKILL` after timeout is a best-effort escalation — its failure should not override the timeout error that is the primary signal.

## Environment

- OS: macOS (Darwin)
- controller-runtime: current main branch
- Related: #3518 (similar signal handling issue, but Windows-specific)

## Suggested fix

The timeout error should always be returned when `StopTimeout` expires. The `SIGKILL` error can be included as supplementary information but should not replace the timeout error.

mattsu2020:
## What changed

- Preserve the Stop timeout error when SIGKILL fails after StopTimeout expires.
- Include the SIGKILL failure as supplementary error context with `errors.Join`.
- Add a deterministic regression test for the timeout-plus-kill-failure path.

## Why

On Darwin, `SIGKILL` can fail with `EPERM` if the process is already terminating after the timeout. That failure previously replaced the primary timeout error, making callers and tests see `unable to kill process ...` instead of `timeout waiting for process ... to stop`.

Fixes #3522.

## Validation

```sh
go test ./pkg/internal/testing/process/
go test ./pkg/internal/testing/process/ -run TestStopReturnsTimeoutWhenKillAfterTimeoutFails -count=100
```

### DIFF
--- a/pkg/internal/testing/process/process_internal_test.go
+++ b/pkg/internal/testing/process/process_internal_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package process
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestStopReturnsTimeoutWhenKillAfterTimeoutFails(t *testing.T) {
+	process, err := os.FindProcess(os.Getpid())
+	if err != nil {
+		t.Fatalf("failed to find current process: %v", err)
+	}
+
+	originalSignalProcess := signalProcess
+	defer func() {
+		signalProcess = originalSignalProcess
+	}()
+
+	signalCount := 0
+	signalProcess = func(_ *os.Process, sig syscall.Signal) error {
+		signalCount++
+		if sig == syscall.SIGKILL {
+			return errors.New("kill failed")
+		}
+		return nil
+	}
+
+	ps := &State{
+		Cmd:         &exec.Cmd{Process: process},
+		Path:        "/path/to/test-process",
+		StopTimeout: time.Nanosecond,
+		waitDone:    make(chan struct{}),
+	}
+
+	err = ps.Stop()
+	if err == nil {
+		t.Fatal("expected Stop to return an error")
+	}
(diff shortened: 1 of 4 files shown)


mattsu2020:
Hi @joelanford @troy0820 @vincepri,

This PR is ready for review. The PR title verifier is now passing, but golangci-lint is still waiting for maintainer approval because the PR has the needs-ok-to-test label.

Could one of you please take a look and run /ok-to-test if the patch looks reasonable?

Thanks!

sbueringer:
Needs rebase after https://github.com/kubernetes-sigs/controller-runtime/pull/3519 is merged

mattsu2020:
> Needs rebase after #3519 is merged

I have addressed it.

pkg/internal/testing/process/process_internal_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func TestStopReturnsTimeoutWhenKillAfterTimeoutFails(t *testing.T) {
+	process, err := os.FindProcess(os.Getpid())
+	if err != nil {
+		t.Fatalf("failed to find current process: %v", err)
+	}
+
+	originalSignalProcess := signalProcess
+	defer func() {
```
alvaroaleman: Defers are not reliable in tests because the goroutine gets terminated on Fatal(), please use t.Cleanup instead

mattsu2020: Thanks. fix it

alvaroaleman (approved):

mattsu2020:
thanks

PR #3531, opened 2026-06-18
📖 Document event recorder name length constraint

### LINKED ISSUE #3511: Indirect Length Limitation of EventRecorder Name
The migration to controller-runtime 0.23.1 introduced stricter validation
for event recorder names using the events.k8s.io API group.

The new API enforces that reportingController names must be valid Kubernetes qualified
names matching the regex: ([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9].

So far was good, but it seems that once the event is created, the event field
`reportingInstance` is populated from `reportingController`+reconciliationId.
And the reportingInstance has a limitation of 128 chars and hence we now have a limitation of 
reportingController.
As per events [doc](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/):

```
reportingController (string)
reportingController is the name of the controller that emitted this Event, e.g. kubernetes.io/kubelet. This field cannot be empty for new Events.
reportingInstance (string)
reportingInstance is the ID of the controller instance, e.g. kubelet-xyzf. This field cannot be empty for new Events and it can have at most 128 characters.
```

Here our error log message :
```
E0501 20:33:03.500254       1 event_broadcaster.go:270] "Server rejected event (will not retry!)" err="Event \"windows-laav-a2.18ab8b3faaa83129\" is invalid: reportingInstance: Invalid value: \"\": can have at most 128 characters" event="&Event{ObjectMeta:{windows-laav-a2.18ab8b3faaa83129  vmsvc-e2e-so7rzi    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []},EventTime:2026-05-01 20:33:03.490136698 +0000 UTC m=+27.204947134,Series:nil,ReportingController:vmware-system-vmop.vmware-system-vmop-controller-manager-7f885ffff5-mlnhc/virtualmachine-controller,ReportingInstance:vmware-system-vmop.vmware-system-vmop-controller-manager-7f885ffff5-mlnhc/virtualmachine-controller-423b393286e26d68e9777202c42da01a,Action:UpdateSuccess,Reason:UpdateSuccess,Regarding:{VirtualMachine vmsvc-e2e-so7rzi windows-laav-a2 a228f522-3a3b-43e9-8b5e-36eaa1b92e12 vmoperator.vmware.com/v1alpha6 5978637 },Related:nil,Note:Update success,Type:Normal,DeprecatedSource:{ },DeprecatedFirstTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedLastTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedCount:0,}"
```

mattsu2020:
## What changed

Documented the effective length constraint for `GetEventRecorder` names when using the `events.k8s.io/v1` recorder.

## Why

`client-go` uses the recorder name as `reportingController` and derives `reportingInstance` by appending `-` plus the current hostname. Kubernetes limits `reportingInstance` to 128 characters, so long recorder names can be rejected by the API server even when the recorder name itself is otherwise valid.

## User impact

Users now have explicit guidance to keep `len(name) + 1 + len(hostname) <= 128` when calling `GetEventRecorder`.

Fixes #3511

## Validation

- `go test ./pkg/recorder ./pkg/internal/recorder ./pkg/cluster ./pkg/manager -run '^$'`\n- `git diff --check`\n\nFull envtest-backed package tests were not runnable locally because `/usr/local/kubebuilder/bin/etcd` is not installed in this environment.

### DIFF
--- a/pkg/internal/recorder/recorder.go
+++ b/pkg/internal/recorder/recorder.go
@@ -161,7 +161,9 @@ func (p *Provider) GetEventRecorderFor(name string) record.EventRecorder {
 }
 
 // GetEventRecorder returns an event recorder that broadcasts to this provider's
-// broadcaster.  All events will be associated with a component of the given name.
+// broadcaster. All events will be associated with the given reportingController
+// name. client-go appends "-" and the hostname to derive reportingInstance,
+// which must fit within the events.k8s.io/v1 128-character limit.
 func (p *Provider) GetEventRecorder(name string) events.EventRecorder {
 	return &lazyRecorder{
 		prov: p,
--- a/pkg/recorder/recorder.go
+++ b/pkg/recorder/recorder.go
@@ -31,6 +31,13 @@ type Provider interface {
 	//
 	// Deprecated: this uses the old events API and will be removed in a future release. Please use GetEventRecorder instead.
 	GetEventRecorderFor(name string) record.EventRecorder
-	// GetEventRecorder returns a EventRecorder with given name.
+	// GetEventRecorder returns an EventRecorder with given name.
+	//
+	// The name is used as the reportingController of events.k8s.io/v1 Events
+	// and must be a valid Kubernetes qualified name. client-go derives the
+	// reportingInstance by appending "-" and the current hostname to
+	// reportingController. reportingInstance must be no more than 128
+	// characters, so callers should ensure that len(name) + 1 + len(hostname)
+	// is at most 128.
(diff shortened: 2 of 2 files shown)


alvaroaleman (approved):

sbueringer:
Interesting that this still has the needs-ok-to-test label :)

Issue #2972, opened 2024-10-09
webhook: Admission metrics are not reported by status code


ahmetb:
_This is several bugs in one bug since it appears multiple things are broken. Apologies if it gets too complicated._

**Problem 1:** It appears that controller-runtime webhooks always respond with HTTP status code `200 OK` even if the request is completely malformed.

Example (sending a malformed request to a webhook endpoint):

```
curl -k -v https://localhost:9443/validate-ship-example-org-v1beta1-frigate
< HTTP/2 200
< ... other response headers ...
<
{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=, expected application/json","code":400}}}
```
-----

**Problem 2:** controller-runtime webhooks do not use the `metav1.Status.Code` and always respond with HTTP status code `200 OK`. 

Example code I used to register my custom handler at `/test`:
```go
webhookServer.Register("/test", &webhook.Admission{Handler: &myHandler{}})
...

type myHandler struct{}
func (*myHandler) Handle(context.Context, admission.Request) admission.Response {
	return admission.Response{
		AdmissionResponse: admissionv1.AdmissionResponse{
			Allowed: false,
			Result: &metav1.Status{
				Code:    403,
				Message: "things suck",
				Status:  "Failure",
				Reason:  metav1.StatusReasonForbidden}}}
}
```
And here's the query I am making:
```
curl -XPOST -k -v -H "Content-Type: application/json" https://localhost:9443/test

< HTTP/2 200
...
<
{"kind":"AdmissionReview","apiVersion":"admission.k8s.io/v1","response":{"uid":"","allowed":false,"status":{"metadata":{},"status":"Failure","message":"things suck","reason":"Forbidden","code":403}}}
```

**Problem 3:** controller-runtime doesn't expose _any metrics_ about rejected requests and their `.status.code`s. This is likely because
1. we always report code=200 no matter the response (Problem 1 & 2)
2. it appears we're capping `controller_runtime_webhook_requests_total{code=...}` dimension to only `200` and `500` here: https://github.com/kubernetes-sigs/controller-runtime/blob/4cae9dfbf174fa5d49ebca007bd6f3784cf1ffb3/pkg/webhook/internal/metrics/metrics.go#L70-L76

so even if I make a faulty request, only the code=200 metric goes up:

```
controller_runtime_webhook_requests_total{code="200",webhook="/test"} 4
```

[...]

PR #3534, opened 2026-06-22
✨ Add admission response status code metrics for webhooks

mattsu2020:
## What does this do, and why do we need it?

This PR adds an admission-response-level metric:

```text
controller_runtime_webhook_admission_responses_total{webhook,allowed,admission_code}
```

This PR intentionally does not change the HTTP response status for valid AdmissionReview responses. Kubernetes admission webhooks are expected to return HTTP 200 with an AdmissionReview body, including when an admission decision is denied.

The existing `controller_runtime_webhook_requests_total{code=...}` metric continues to represent the HTTP transport status code observed by promhttp. The new metric separately exposes `AdmissionResponse.status.code`, making admission decisions and errors such as 400, 403, 429, and 500 observable without changing the existing HTTP request metric semantics.

The common response series are initialized when an Admission webhook is registered directly with the webhook server or configured through `StandaloneWebhook`. Responses with an unset status code follow the existing `Response.Complete` behavior and are recorded as 200.

## Validation

- `go test ./pkg/webhook/...`
- `go vet ./pkg/webhook/...`

Addresses the metrics aspect of #2972.

### DIFF
--- a/pkg/webhook/admission/http.go
+++ b/pkg/webhook/admission/http.go
@@ -162,8 +162,8 @@ func (wh *Webhook) writeAdmissionResponse(ctx context.Context, w io.Writer, ar v
 		}
 	} else {
 		res := ar.Response
-		code := int32(0)
-		if res.Result != nil {
+		code := int32(http.StatusOK)
+		if res.Result != nil && res.Result.Code != 0 {
 			code = res.Result.Code
 		}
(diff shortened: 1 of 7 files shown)


mattsu2020:
Hi @vincepri @troy0820,
Could you please take a look at this PR when you have a chance?

Thanks!

PR #3536, opened 2026-06-25
🐛 manager: suppress expected leader-election-lost error on graceful shutdown

### LINKED ISSUE #3535: level=ERROR msg="error received after stop sequence was engaged" err="leader election lost"
When running a controller manager with `LeaderElection` enabled, shutting down the manager will obviously release the leader lock. However, this seems to still trigger a “leader election lost” error:

```
level=ERROR msg="error received after stop sequence was engaged" err="leader election lost"
```

On its own, this is just a nitpick, but the problem is in combination with using the [t.Output() writer](https://pkg.go.dev/testing#T.Output) in your test functions. Tests are not allowed to keep writing to it after the test has finished. You will either get a panic, or [might even get a data race](https://github.com/golang/go/issues/67701) detected if you run the test with the race detector.

The problem here is that the manager keeps doing something even after its `Start` method has already returned. The log is a symptom of that, but IMO the manager should be completely gone once `Start` returns.

liketosweep:
Fixes #3535
 
## What the Issue Was
 
On graceful shutdown with `LeaderElection` enabled, cancelling the
leader election context triggers `OnStoppedLeading`, which pushes
`errLeaderElectionLost` onto `errChan`. The drain goroutine in
`engageStopProcedure` only filtered `context.Canceled`, so this
expected error was logged as unexpected:
 
```
level=ERROR msg="error received after stop sequence was engaged" err="leader election lost"
```
 
When combined with `t.Output()` in tests, this also causes panics or
data races as the manager writes to the test log after the test exits.
 
## Solution
 
- Added a package-level `errLeaderElectionLost` sentinel for reliable
  `errors.Is` matching.
- Used the sentinel in `OnStoppedLeading` instead of an anonymous
  `errors.New`.
- Added the sentinel to the `engageStopProcedure` drain filter
  alongside `context.Canceled`.
Genuine mid-run leader election loss (before stop is engaged) still
propagates to `Start`'s caller unchanged.
 
## Changes
 
- `pkg/manager/internal.go`: sentinel + filter fix.
- `pkg/manager/manager_test.go`: regression test for graceful shutdown.
## Testing
 
```
$ go test sigs.k8s.io/controller-runtime/pkg/manager -v \
  -run "should not log an error when leader election is lost during graceful shutdown"
--- PASS
PASS
```

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -51,14 +51,19 @@ import (
 
 const (
 	// Values taken from: https://github.com/kubernetes/component-base/blob/master/config/v1alpha1/defaults.go
-	defaultLeaseDuration          = 15 * time.Second
-	defaultRenewDeadline          = 10 * time.Second
-	defaultRetryPeriod            = 2 * time.Second
+	defaultLeaseDuration = 15 * time.Second
+	defaultRenewDeadline = 10 * time.Second
+	defaultRetryPeriod   = 2 * time.Second
+
 	defaultGracefulShutdownPeriod = 30 * time.Second
 
 	defaultReadinessEndpoint = "/readyz"
 	defaultLivenessEndpoint  = "/healthz"
 )
+
+// errLeaderElectionLost is returned by OnStoppedLeading when leader election is lost.
+// It is used to suppress this expected error during graceful shutdown.
+var errLeaderElectionLost = errors.New("leader election lost")
 
 var _ Runnable = &controllerManager{}
 
@@ -529,7 +534,7 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 			})
 			select {
 			case err := <-cm.errChan:
-				if !errors.Is(err, context.Canceled) {
+				if !errors.Is(err, context.Canceled) && !errors.Is(err, errLeaderElectionLost) {
 					cm.logger.Error(err, "error received after stop sequence was engaged")
 				}
 			case <-stopComplete:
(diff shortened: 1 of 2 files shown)


pkg/manager/internal.go
```diff
@@ -529,7 +534,7 @@ func (cm *controllerManager) engageStopProcedure(stopComplete <-chan struct{}) e
 			})
 			select {
 			case err := <-cm.errChan:
-				if !errors.Is(err, context.Canceled) {
+				if !errors.Is(err, context.Canceled) && !errors.Is(err, errLeaderElectionLost) {
```
alvaroaleman, line 544: This means we will not emit an error anymore even if we lost leader election for a reason other than having stopped the binary which would be wrong?

liketosweep: The two `errChan` readers are sequential, not concurrent. `engageStopProcedure` is called via `defer` in `Start()` (line 386), so this drain goroutine only starts after `Start()`'s select loop has already exited  consuming and returning any mid-run `errLeaderElectionLost` to the caller. Any instance seen here was caused by our own `internalCancel()`. I've added a comment in the latest commit to make this explicit.

sbueringer: I think it was a valid point on the issue that this Go routine should be completed before we return from `engageStopProcedure`.

Otherwise if we get any other error here and then log to cm.logger we would have the same problem again.

sbueringer:
/close

Feel free to reopen if there's still interest in pursuing this PR

PR #3540, opened 2026-07-01
🐛 Fix PriorityQueue deadlock on shutdown

### LINKED ISSUE #3538: Deadlock in PriorityQueue on shutdown
Due to a race condition, calls to PriorityQueue's `GetWithPriority()` can block forever after `ShutDown()` has been called. On my machine, this example reliably reproduces the deadlock within a few iterations:
```
package main

import (
	"fmt"
	"k8s.io/utils/ptr"
	"math/rand"
	"sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue"
	"sync"
	"time"
)

func main() {
	for i := 0; ; i++ {
		fmt.Printf("Starting iteration %d ... ", i)

		pq := priorityqueue.New[int]("example-queue")
		wg := sync.WaitGroup{}
		for j := 0; j < 20; j++ {
			wg.Go(func() {
				for {
					pq.AddWithOpts(priorityqueue.AddOpts{Priority: ptr.To(rand.Int())}, rand.Int())
					_, _, shutdown := pq.GetWithPriority()
					if shutdown {
						break
					}
				}
			})
		}

		time.Sleep(10 * time.Millisecond)
		pq.ShutDown()
		wg.Wait()
		fmt.Println("done")
	}
}
```
```
❯ go run main.go
Starting iteration 0 ... done
Starting iteration 1 ... done
Starting iteration 2 ... done
Starting iteration 3 ... done
Starting iteration 4 ... done
Starting iteration 5 ... done
Starting iteration 6 ... <blocks forever>
```

When `ShutDown()` is called, any goroutines waiting in `GetWithPriority()` will return immediately, without decrementing `w.waiters` and without reading from `w.get`. If the `handleReadyItems()` goroutine is already awake, it will acquire `w.lock`, pass the `w.waiters == 0` check, and, if there are ready items, attempt to send one to the `w.get` channel. The send will block forever, while still holding `w.lock`.

Any additional goroutines that have called `GetWithPriority()` and have already passed the initial check of `w.shutdown.Load()` will block forever trying to acquire `w.lock`.

I think the best way to fix this would be to replace the blocking send in `handleReadyItems()` with a select that also checks `w.done`, allowing `handleReadyItems` to wake up and drop the lock in this scenario. I've implemented this here, and confirmed that it fixes the example provided above: https://github.com/john-gallagher-dd/controller-runtime/commit/4fde22f444d42283d6305efbad2700e34fa5496d If this looks reasonable, I'm happy to open a PR.

arpitjain099:
Calls to `GetWithPriority` can hang forever after `ShutDown`, and the whole PriorityQueue deadlocks. This fixes #3538.

The problem is in `handleReadyItems`: it hands a ready item to a waiting consumer with an unbuffered send on `w.get` while holding `w.lock`. If `ShutDown` closes `w.done` first, the consumer that was parked in `GetWithPriority` returns through `<-w.done` instead of receiving, so nobody is left to take the item. The send then blocks forever with `w.lock` still held, and every later `GetWithPriority` and `handleAddBuffer` stalls on that lock. I wrapped the send in a select that also watches `w.done`, so it gets cancelled on shutdown and the lock is released. This matches what the reporter suggested and confirmed fixes it.

I also added a regression test that runs the reporter's add/get loop across several rounds and fails with a bounded wait if any worker stays blocked after `ShutDown`. It reliably deadlocks on the current code (usually within the first handful of rounds) and passes with the fix, including under `-race`.

### DIFF
--- a/pkg/controller/priorityqueue/priorityqueue.go
+++ b/pkg/controller/priorityqueue/priorityqueue.go
@@ -397,7 +397,16 @@ func (w *priorityqueue[T]) handleReadyItems() {
 				w.waiters--
 				delete(w.items, item.Key)
 				toDelete = append(toDelete, item)
-				w.get <- *item
+				// w.get is unbuffered, so this send blocks until a GetWithPriority
+				// consumer receives. A consumer that is parked in GetWithPriority can
+				// instead return via <-w.done once ShutDown closes it, leaving no one
+				// to receive here. Also watch w.done so the send does not block forever
+				// and deadlock the queue (the whole queue stalls because w.lock is held).
+				select {
+				case w.get <- *item:
+				case <-w.done:
+					return false
+				}
 
 				return w.waiters > 0
 			})
--- a/pkg/controller/priorityqueue/priorityqueue_test.go
+++ b/pkg/controller/priorityqueue/priorityqueue_test.go
@@ -1007,3 +1007,50 @@ func TestWhenAddingMultipleItemsWithRatelimitTrueTheyDontAffectEachOther(t *test
 		g.Expect(metrics.retries["test"]).To(Equal(2))
 	})
 }
+
+// TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems reproduces
+// https://github.com/kubernetes-sigs/controller-runtime/issues/3538: if
+// handleReadyItems is handing an item to a consumer over the unbuffered w.get
+// channel while ShutDown runs, the parked GetWithPriority consumer can return
+// through <-w.done instead of receiving from w.get. handleReadyItems is then
+// stuck on the send while still holding w.lock, so every later GetWithPriority
+// (and handleAddBuffer) blocks and the whole queue deadlocks. This is a timing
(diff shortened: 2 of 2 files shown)


alvaroaleman (approved):
thanks!

sbueringer:
/assign
/hold
(I"ll also take a look tomorrow)

Please check the unit tests / linter findings

arpitjain099:
Thanks for taking a look, @sbueringer. The failing checks were all modernize linter findings on the new test, nothing in the fix itself. I switched the counted loops to range-over-int, used WaitGroup.Go for the worker goroutines, and swapped ptr.To for the new(x) builtin. Pushed that up now. The priorityqueue package tests still pass locally with -race.

sbueringer:
Thx!

/approve
/hold cancel

/assign @alvaroaleman 
(In case you want to take another look)

alvaroaleman (approved):

troy0820:
Does this need to get backported into release-0.24 and release-0.23

/cherry-pick release-0.24
/cherry-pick release-0.23

PR #3543, opened 2026-07-06
🐛 Fix restmapper cache invalidation when version is not provided

### LINKED ISSUE #3542: RestMapper cache is never invalidated when a client requests a group without specifying a version
If a client requests a mapping for an API group without specifying the version, the current code will always rely on `findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked` to retrieve group resources using aggregated discovery.

But since this function looks at the apiGroups cache first, once an apiGroup has been discovered, its resources will never be updated. If some CRDs that declare new resources for an existing group are installed after this initial lookup, subsequent calls to mapper.RESTMapping("newKind") will continuously fail.

This has been [observed in sylva project](https://gitlab.com/sylva-projects/sylva-core/-/work_items/4272) that uses flux kustomize controller. It sometimes fails to perform healthChecks on some resources and continuously returns a `NoMatchError`, whereas the CRD is installed for a while in cluster (see [flux issue](https://github.com/fluxcd/kustomize-controller/issues/1680)).  Kustomize controller relies on kstatus library to perform healthCheck on resources, and that library [does not provide any version](https://github.com/fluxcd/cli-utils/blob/f812f045c69b75ed2843e0ed0782ec7efca82999/pkg/kstatus/polling/clusterreader/caching_reader.go#L225) while calling `mapper.RESTMapping`.

zioc:
restmapper's `findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked` method can't rely on the apiGroups cache, otherwise new versions of a group will never be discovered when the client does not provide a group version while requesting a RESTMapping.

Fixes: https://github.com/kubernetes-sigs/controller-runtime/issues/3542

### DIFF
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -275,12 +275,6 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
 // findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked tries to find the passed apiGroup.
 // If the server supports aggregated discovery, it will always perform that.
 func (m *mapper) findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked(groupName string) (_ *metav1.APIGroup, didAggregatedDiscovery bool, _ error) {
-	// Looking in the cache first
-	group, ok := m.apiGroups[groupName]
-	if ok {
-		return group, false, nil
-	}
-
 	// Update the cache if nothing was found.
 	apiGroups, maybeResources, _, err := m.client.GroupsAndMaybeResources()
 	if err != nil {
--- a/pkg/client/apiutil/restmapper.go
+++ b/pkg/client/apiutil/restmapper.go
@@ -275,13 +275,6 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
 // findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked tries to find the passed apiGroup.
 // If the server supports aggregated discovery, it will always perform that.
 func (m *mapper) findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked(groupName string) (_ *metav1.APIGroup, didAggregatedDiscovery bool, _ error) {
-	// Looking in the cache first
-	group, ok := m.apiGroups[groupName]
-	if ok {
-		return group, false, nil
-	}
-
-	// Update the cache if nothing was found.
 	apiGroups, maybeResources, _, err := m.client.GroupsAndMaybeResources()
 	if err != nil {
 		return nil, false, fmt.Errorf("failed to get server groups: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/client/apiutil/restmapper.go
```diff
@@ -275,12 +275,6 @@ func (m *mapper) addGroupVersionResourcesToCacheAndReloadLocked(gvr map[schema.G
 // findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked tries to find the passed apiGroup.
 // If the server supports aggregated discovery, it will always perform that.
 func (m *mapper) findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked(groupName string) (_ *metav1.APIGroup, didAggregatedDiscovery bool, _ error) {
-	// Looking in the cache first
-	group, ok := m.apiGroups[groupName]
```
alvaroaleman, line 279: A patch like this essentially disables all caching, thats a pretty big regression as it results in running discovery on every single request. I don't think we are willing to accept that.

zioc: I may misunderstood, but I was thinking that this function was only called when a cache miss previously occurred in interface methods like `RESTMapping`:

```
	res, err := m.getMapper().RESTMapping(gk, versions...)                       // Initial cache lookup
	if meta.IsNoMatchError(err) {                                                // Cache miss
		if err := m.addKnownGroupAndReload(gk.Group, versions...); err != nil {  // Fetch API
			return nil, err
		}
		res, err = m.getMapper().RESTMapping(gk, versions...)
	}
```

And all the code paths leading to that function are following the same pattern: 
`client request -> IsNoMatchError (cache miss) -> addKnownGroupAndReload -> findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked (when len(versions) == 0)`

So I used to believe that it is was acceptable to remove this cache lookup in  `findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked`, since this code was only reached when mapping was not found in cache during first lookup. At some point, if the resource was not found in cache, and version is not provided by client, all group resources need to be fetched from the API, otherwise it will never be discovered. This is highlighted in the provided test that fails with current implementation whereas it simulates a real use-case (CRD installed after initial call to `RESTMapping`)

I'm open to any suggestion to implement a better fix, and I am also willing to close this PR and pass the relay if some maintainer wants to tackle the issue, but I feel that it deserves to be fixed.

zioc: Any feedback on my previous message @alvaroaleman ?

pkg/client/apiutil/restmapper.go
```diff
@@ -294,12 +287,32 @@ func (m *mapper) findAPIGroupByNameAndMaybeAggregatedDiscoveryLocked(groupName s
...
+				}
+			}
+			filtered := []metav1.GroupVersionForDiscovery{}
+			for _, v := range knownGroup.Group.Versions {
+				if serverVersions[v.Version] {
+					filtered = append(filtered, v)
+				}
+			}
+			knownGroup.Group.Versions = filtered
+		}
```
zioc, line 310: I had to add this cache invalidation logic to preserve the preserve current behavior tested in "LazyRESTMapper should invalidate the group cache if a version is not found" test case. 

However I wonder if it is really relevant to attempt to preserve that specific behavior, since stale apiGroups or versions won't be pruned from cache in most cases, the only way to prune a staled version is to request an undefined kind, which is not a very clean/documented API...

Maybe I should drop that part and adapt the test case instead?

pkg/client/apiutil/restmapper_test.go
```diff
@@ -598,25 +597,40 @@ func TestLazyRestMapperProvider(t *testing.T) {
 				c, err := client.New(restCfg, client.Options{Scheme: s})
 				g.Expect(err).NotTo(gmg.HaveOccurred())
 
-				// Register another CRD in runtime - "riders.crew.example.com".
-				createNewCRD(t.Context(), g, c, "crew.example.com", "Rider", "riders")
+				// Register another CRD in runtime - "riders.crew.example.com" with differents versions (v1alpha1 and v1alpha2)
+				newCRD := newCRD(t.Context(), g, c, "crew.example.com", "Rider", "riders")
+				newCRD.Spec.Versions[0].Name = "v1alpha1"
+				newCRD.Spec.Versions[1].Name = "v1alpha2"
+				g.Expect(c.Create(t.Context(), newCRD)).To(gmg.Succeed())
```
zioc, line 604: Instead of adding a dedicated test in restmapper_wb_test.go as [proposed initially](https://github.com/kubernetes-sigs/controller-runtime/commit/446014becb543ced87c21c47dc232ec31f349d99), I'm adapting this existing test to reproduce the observed issue by adding a CRD in a group with a version that is not yet in cache.

pkg/client/apiutil/restmapper.go
zioc: As [this test](https://github.com/zioc/controller-runtime/blob/fd8c98cb23011f42e5f9c462be2e74b74410b0f6/pkg/client/apiutil/restmapper_test.go#L730-L733) is passing, we can see that current caching behavior is not altered.

PR #3553, opened 2026-07-17
cache/informer: add type-safe informer support

sttts:
## What

This ports controller-runtime informer/source wiring to the type-safe informer constructs from kubernetes/kubernetes#139821.

- Adds `cache.TypedInformer`, `cache.NewTypedInformer`, and `cache.GetTypedInformer`.
- Routes `source.Kind` and `source.TypedInformer` through `TypedResourceEventHandler`.
- Uses typed deleted-object tombstones instead of local untyped tombstone decoding in controller-runtime.
- Updates source tests to use generated typed informers.
- Adds adapter coverage for typed handlers, typed indexers, and deleted final state handling.

## Impact

This intentionally allows breaking code changes:

- `source.Kind` / `source.TypedKind` objects now need to satisfy `cache.TypedObject`.
- `source.TypedInformer.Informer` now expects `cache.TypedInformer[T]`.

The Kubernetes module replaces point at the current head used for kubernetes/kubernetes#139821 until those APIs are available from upstream Kubernetes modules.

## Validation

- `make test`
- `git diff --check`


pkg/cache/cache.go
```diff
@@ -96,6 +96,21 @@ type Informers interface {
 	client.FieldIndexer
 }
 
+// TypedObject is a client.Object that can be used with the type-safe informer API.
+type TypedObject interface {
+	client.Object
+	toolscache.Object
```
alvaroaleman: @sttts Could you tell your LLM to update this to just extend client.Object to be comparable to avoid all the noise?

sttts: Done in 06b411f7: `TypedObject` now embeds `client.Object` and `comparable` directly instead of `toolscache.Object`.

pkg/source/source.go
```diff
@@ -266,9 +266,9 @@ func (cs *channel[object, request]) syncLoop(ctx context.Context) {
 
 // TypedInformer is used to provide a source of events originating inside the cluster from Watches using generic
 // event handlers and predicates.
-type TypedInformer[object any, request comparable] struct {
+type TypedInformer[object toolscache.Object, request comparable] struct {
```
sbueringer, line 269: Looks like this reverts: https://github.com/kubernetes-sigs/controller-runtime/pull/3520

pkg/cache/cache.go
sbueringer: Wondering if toolscache.Object absolutely has to be comparable. Because if it would not be this problem wouldn't exist

PR #3556, opened 2026-07-20
🐛 Reject namespace filters for cluster-scoped resources

### LINKED ISSUE #988: The filtering behavior of List() and DeleteAllOf() are different when they are used with the "InNamespace" predicate.
Hi, I'm using operator-sdk. When I'm using the k8s client provided by controller-manager constructed by operator-sdk, I found the following surprising things:

I am accidentally trying to delete CRs under a certain namespace, which is a mistake because the CR was defined in cluster scope. The strange thing is that DeleteAllOf() removes all CRs in the cluster, but List() didn't return any result.

The relevant code probably is:

cache_reader:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L109

writer:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/unstructured_client.go#L130

dlanov:
Fixes #988.

`InNamespace` was silently ignored when `List` or `DeleteAllOf` targeted a cluster-scoped resource. This could turn a namespace-restricted list into an unfiltered cluster-wide list and, more seriously, turn a namespace-restricted `DeleteAllOf` into a cluster-wide delete-collection request.

This change:

* validates non-empty namespace options before dispatching to the typed, unstructured, metadata, or cache-backed client paths;
* returns an error when a namespace is supplied for a cluster-scoped resource;
* rejects the operation before an HTTP request is sent;
* preserves cluster-wide operations when no namespace is specified;
* normalizes functional options so each caller-supplied option is applied exactly once.

Tests cover typed, unstructured, and metadata representations for both `List` and `DeleteAllOf`, verify that rejected operations send no request, and confirm custom options are invoked only once.

A compatibility consideration is that callers passing `InNamespace` for cluster-scoped resources will now receive an error instead of silently performing an unscoped operation.

Validation:

```
go test ./pkg/client -count=1
go test ./pkg/client/fake -count=1
go test ./pkg/client/... -count=1
go test -race ./pkg/client -run '^TestClient$' -ginkgo.focus='issue #988|should invoke each caller-supplied' -count=1
go vet ./pkg/client/...
make lint
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -331,6 +331,24 @@ func (c *client) Delete(ctx context.Context, obj Object, opts ...DeleteOption) e
 
 // DeleteAllOf implements client.Client.
 func (c *client) DeleteAllOf(ctx context.Context, obj Object, opts ...DeleteAllOfOption) error {
+	// Parse opts into a normalized DeleteAllOfOptions exactly once, then pass only
+	// that normalized value downstream. *DeleteAllOfOptions implements
+	// DeleteAllOfOption, so downstream layers can copy its already-resolved fields
+	// without re-invoking any caller-supplied option a second time.
+	deleteAllOfOpts := DeleteAllOfOptions{}
+	deleteAllOfOpts.ApplyOptions(opts)
+	opts = []DeleteAllOfOption{&deleteAllOfOpts}
+
+	if deleteAllOfOpts.Namespace != "" {
+		gvk, err := c.GroupVersionKindFor(obj)
+		if err != nil {
+			return err
+		}
+		if err := c.rejectNamespaceForClusterScoped("DeleteAllOf", gvk, deleteAllOfOpts.Namespace); err != nil {
+			return err
+		}
+	}
+
 	switch obj.(type) {
 	case runtime.Unstructured:
 		return c.unstructuredClient.DeleteAllOf(ctx, obj, opts...)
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -341,6 +359,19 @@ func (c *client) DeleteAllOf(ctx context.Context, obj Object, opts ...DeleteAllO
...
+// namespace is non-empty; it is a no-op for namespace-scoped resources.
+func (c *client) rejectNamespaceForClusterScoped(op string, gvk schema.GroupVersionKind, namespace string) error {
+	namespaced, err := apiutil.IsGVKNamespaced(gvk, c.mapper)
+	if err != nil {
+		return fmt.Errorf("failed to determine if %s is namespace-scoped: %w", gvk, err)
+	}
+	if namespaced {
+		return nil
+	}
+	return fmt.Errorf("failed to %s: %s is cluster-scoped and does not support namespace %q", op, gvk, namespace)
```
sbueringer:
```suggestion
	return fmt.Errorf("failed to %s: %s is cluster-scoped and does not support to filter by namespace %q", op, gvk, namespace)
```

pkg/client/client_test.go
```diff
@@ -2089,6 +2156,43 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 				_, err = clientset.AppsV1().Deployments(ns).Get(ctx, dep2Name, metav1.GetOptions{})
 				Expect(err).To(HaveOccurred())
 			})
+
+			It("should error when InNamespace targets a cluster-scoped object, without sending a request (issue #988)", func(ctx SpecContext) {
+				cl, rt := newRecordingClient()
+
+				err := cl.DeleteAllOf(ctx, node, client.InNamespace("some-namespace"))
+				expectClusterScopedNamespaceError(err, "DeleteAllOf", "some-namespace")
```
sbueringer: Can we simply check that the error has occurred and match directly agains the full error string? Seems much easier to read 

(same below)

pkg/client/client_test.go
```diff
@@ -2089,6 +2156,43 @@ U5wwSivyi7vmegHKmblOzNVKA5qPO8zWzqBC
 				_, err = clientset.AppsV1().Deployments(ns).Get(ctx, dep2Name, metav1.GetOptions{})
 				Expect(err).To(HaveOccurred())
 			})
+
+			It("should error when InNamespace targets a cluster-scoped object, without sending a request (issue #988)", func(ctx SpecContext) {
+				cl, rt := newRecordingClient()
```
sbueringer: I don't think we need the recording client. It should be enough to simply check for error occurred (or not) and 

(same for other test cases)

pkg/client/client_test.go
```diff
@@ -142,6 +143,72 @@ func metaOnlyFromObj(obj interface {
...
+
+func (o *countingListOption) ApplyToList(opts *client.ListOptions) {
+	o.applied++
+	o.inner.ApplyToList(opts)
+}
+
+// countingDeleteAllOfOption wraps a DeleteAllOfOption and counts how many
+// times ApplyToDeleteAllOf is invoked, to prove options are applied exactly
+// once.
+type countingDeleteAllOfOption struct {
```
sbueringer: I don't think we need to verify this. Just additional noise in our tests

PR #3562, opened 2026-07-24
📖predicate: clarify predicates don't reduce cache memory

### LINKED ISSUE #3552: Predicates don't reduce cache memory — informer caches full objects for every watched resource
## Body

Watching a Kind with a predicate that rejects everything still caches every object of that Kind in memory. Predicates filter the workqueue, not the informer cache, and this isn't documented anywhere.

For operators watching Secrets, this is a real problem: a namespace full of large Helm release secrets (100KB+ each) gets fully cached even though the operator never touches them, leading to OOMKills. See cloudnative-pg/cloudnative-pg#10924.

### Reproduction

Bare controller, reject-all predicate, no-op reconcile:

```go
c, _ := controller.New("secret-watcher", mgr, controller.Options{Reconciler: &noopReconciler{}})
c.Watch(source.Kind(mgr.GetCache(), &corev1.Secret{},
    &handler.TypedEnqueueRequestForObject[*corev1.Secret]{},
    rejectAllPredicate{}, // Create/Update/Delete/Generic all return false
))
```

Create ~200 secrets of 100KB in the watched namespace, run it, then:

```
$ go tool pprof -top http://localhost:6060/debug/pprof/heap
 1659.12kB 13.45% 65.96%  5053.68kB 40.96%  k8s.io/api/core/v1.(*Secret).Unmarshal
 3394.55kB 27.51% 27.51%  3394.55kB 27.51%  ...(*ObjectMeta).Unmarshal
```

Reconcile is never called, yet all secrets sit in the cache. Memory scales linearly with secret count/size.

### Versions

controller-runtime v0.24.1, k8s.io/api v0.36.2, Go 1.22.

### Suggestion

Document that predicates don't limit what the informer caches, and provide a supported way to keep memory bounded for large-payload Kinds **without giving up the cache**. Dropping the cache entirely (e.g. `Client.Cache.DisableFor`) isn't viable here — it still rely on the cache for watches and metadata lookups.

I'm aware a `cache.ByObject` `Transform` can strip `.Data`/`.StringData` before objects enter the cache, and it works. But it's a hand-rolled workaround that's undocumented for this purpose, and I'm not proposing it as the intended fix; other options may be a better fit.

GellertBabel:
Clarifies in the `Predicate` godoc that predicates only filter which events are enqueued for reconciliation — they do **not** reduce what the informer cache stores. Every watched object is still held in memory regardless of whether a predicate rejects its events.

Adds a short pointer to the mechanisms that actually bound cache memory for a Kind: a label/field selector (`cache.Options` / `cache.ByObject`), a `cache.ByObject` Transform, or `client.Options.Cache.DisableFor`.

This is a source of confusion (and OOMKills) for controllers watching Kinds with many or large objects, e.g. Secrets in a namespace that also holds large, unrelated objects.

Fixes #3552

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -30,6 +30,9 @@ import (
 var log = logf.RuntimeLog.WithName("predicate").WithName("eventFilters")
 
 // Predicate filters events before enqueuing the keys.
+//
+// NOTE: This does not affect what the cache stores. That is configured via
+// cache.Options, which are global to all users of the cache.
 type Predicate = TypedPredicate[client.Object]
 
 // TypedPredicate filters events before enqueuing the keys.


pkg/predicate/predicate.go
```diff
@@ -30,6 +30,12 @@ import (
 var log = logf.RuntimeLog.WithName("predicate").WithName("eventFilters")
 
 // Predicate filters events before enqueuing the keys.
+//
+// NOTE: Predicates only filter which events are enqueued for reconciliation.
```
alvaroaleman: This first sentence just repeats the info two lines above. Please just say "This setting doesn't impact what the cache stores, use cache.Options to configure that. Keep in mind that cache.Options are global to all users of the cache." or some such.

Please do not mention individual cache option names here, they could change and are already documented in cache.Options.  Suggesting to use cache.Disable for is also pretty nonsensical, you never want to disable the cache for an object you have a source configured for.

alvaroaleman (approved):

PR #3565, opened 2026-08-03
🐛 Fix goroutine leaks in cache Start() methods

dongjiang1989:
## What does this PR do?

This PR fixes goroutine leaks in `multiNamespaceCache.Start()` and `delegatingByGVKCache.Start()` methods in the `pkg/cache` package.

### Problem

Both `Start()` methods used unbuffered error channels (`errs := make(chan error)`) to collect errors from goroutines. This caused goroutine leaks in two scenarios:

1. **Multiple caches return errors simultaneously**: The `select` statement only receives the first error and returns, leaving other goroutines blocked forever on `errs <- err` with no receiver.

2. **Context cancelled while goroutines try to send errors**: When `ctx.Done()` triggers first, the function returns immediately. Any subsequent error sends from goroutines will block forever.

### Example

```go
// Original problematic code
errs := make(chan error)  // unbuffered
for _, cache := range caches {
    go func() {
        if err := cache.Start(ctx); err != nil {
            errs <- err  // blocks forever if no receiver!
        }
    }()
}
select {
case err := <-errs:
    return err  // only receives ONE error, others leak
case <-ctx.Done():
    return nil  // returns immediately, senders leak
}
```

### Fix

- Use buffered channels sized to the number of caches
- Use `sync.WaitGroup` to track all goroutines
- Wait for all goroutines to complete before returning

```go
// Fixed code
errs := make(chan error, len(caches))  // buffered
var wg sync.WaitGroup
for _, cache := range caches {
    wg.Go(func() {
        if err := cache.Start(ctx); err != nil {
            errs <- err  // never blocks
        }
    })
}
done := make(chan struct{})
go func() { wg.Wait(); close(done) }()
// ... wait for all goroutines to complete
```

### DIFF
--- a/pkg/cache/delegating_by_gvk_cache.go
+++ b/pkg/cache/delegating_by_gvk_cache.go
@@ -77,8 +77,10 @@ func (dbt *delegatingByGVKCache) Start(ctx context.Context) error {
 	allCaches := slices.Collect(maps.Values(dbt.caches))
 	allCaches = append(allCaches, dbt.defaultCache)
 
+	// Use a buffered channel to prevent goroutine leaks when multiple caches
+	// return errors simultaneously.
+	errs := make(chan error, len(allCaches))
 	wg := &sync.WaitGroup{}
-	errs := make(chan error)
 	for idx := range allCaches {
 		cache := allCaches[idx]
 		wg.Go(func() {
@@ -88,12 +90,36 @@ func (dbt *delegatingByGVKCache) Start(ctx context.Context) error {
 		})
 	}
 
-	select {
-	case err := <-errs:
-		return err
-	case <-ctx.Done():
+	// Wait for all goroutines to complete in a separate goroutine
+	done := make(chan struct{})
+	go func() {
 		wg.Wait()
-		return nil
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// All caches completed, check if any returned an error
+		select {
+		case err := <-errs:
+			return err
+		default:
+			return nil
+		}
+	case <-ctx.Done():
+		// Context cancelled, wait for goroutines to finish
+		<-done
+		// Return any error that was captured
+		select {
+		case err := <-errs:
+			return err
+		default:
+			return nil
+		}
+	case err := <-errs:
+		// First error returned, wait for other goroutines to finish
+		<-done
+		return err
 	}
 }
 
--- a/pkg/cache/delegating_by_gvk_cache.go
+++ b/pkg/cache/delegating_by_gvk_cache.go
@@ -21,8 +21,8 @@ import (
 	"maps"
 	"slices"
 	"strings"
-	"sync"
 
+	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -77,24 +77,15 @@ func (dbt *delegatingByGVKCache) Start(ctx context.Context) error {
 	allCaches := slices.Collect(maps.Values(dbt.caches))
 	allCaches = append(allCaches, dbt.defaultCache)
 
-	wg := &sync.WaitGroup{}
-	errs := make(chan error)
+	group, childCtx := errgroup.WithContext(ctx)
 	for idx := range allCaches {
 		cache := allCaches[idx]
-		wg.Go(func() {
-			if err := cache.Start(ctx); err != nil {
-				errs <- err
-			}
+		group.Go(func() error {
+			return cache.Start(childCtx)
 		})
 	}
 
-	select {
-	case err := <-errs:
-		return err
-	case <-ctx.Done():
-		wg.Wait()
-		return nil
-	}
+	return ignoreContextCanceled(group.Wait())
 }
 
 func (dbt *delegatingByGVKCache) WaitForCacheSync(ctx context.Context) bool {
--- a/pkg/cache/goroutine_leak_test.go
+++ b/pkg/cache/goroutine_leak_test.go
@@ -1,0 +1,227 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"go.uber.org/goleak"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// mockCache is a mock implementation of Cache for testing
+type mockCache struct {
+	startFunc func(ctx context.Context) error
+}
+
+func (m *mockCache) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	return nil
+}
+func (m *mockCache) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+	return nil
+}
+func (m *mockCache) GetInformer(ctx context.Context, obj client.Object, opts ...InformerGetOption) (Informer, error) {
+	return nil, nil
+}
+func (m *mockCache) GetInformerForKind(ctx context.Context, gvk schema.GroupVersionKind, opts ...InformerGetOption) (Informer, error) {
+	return nil, nil
+}
+func (m *mockCache) RemoveInformer(ctx context.Context, obj client.Object) error {
+	return nil
+}
+func (m *mockCache) Start(ctx context.Context) error {
+	if m.startFunc != nil {
+		return m.startFunc(ctx)
+	}
+	<-ctx.Done()
+	return nil
+}
+func (m *mockCache) WaitForCacheSync(ctx context.Context) bool {
+	return true
+}
+func (m *mockCache) IndexField(ctx context.Context, obj client.Object, field string, extractValue client.IndexerFunc) error {
+	return nil
+}
+
+// TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors tests that goroutines
+// don't leak when multiple caches return errors simultaneously
+func TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors(t *testing.T) {
+	defer goleak.VerifyNone(t)
+
+	// Create a multiNamespaceCache with multiple caches that all return errors
+	errTest := errors.New("test error")
+	namespaceToCache := map[string]Cache{
+		"ns1": &mockCache{startFunc: func(ctx context.Context) error { return errTest }},
+		"ns2": &mockCache{startFunc: func(ctx context.Context) error { return errTest }},
+		"ns3": &mockCache{startFunc: func(ctx context.Context) error { return errTest }},
+	}
+
+	c := &multiNamespaceCache{
+		namespaceToCache: namespaceToCache,
+	}
+
+	ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
+	defer cancel()
+
+	// Start should return an error
+	err := c.Start(ctx)
+	if err == nil {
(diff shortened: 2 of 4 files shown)


thc1006 (commented):
TLDR: the leak is real and worth fixing, but waiting on `done` after the first error hangs `Start` whenever any sibling cache is healthy, which is the ordinary case. I ran it against this branch and against `main`, results below. The existing test only passes because its parent context expires on its own after 50ms.

A healthy `Cache.Start` runs until its context is cancelled. In the `case err := <-errs` branch this waits for every other `Start` goroutine to return, and nothing cancels their context, since the children are handed the same `ctx` the parent is still using. So `done` never closes and `Start` never returns. Both files have this shape.

One child returning a sentinel error, one child on `<-ctx.Done()`, parent context left alive:

| | `Start` returns | healthy child sees cancellation |
| --- | --- | --- |
| `main` | promptly, with the error | no, it keeps running, which is the leak this PR is about |
| this branch | never, 5s timeout in my test | n/a |
| `errgroup.WithContext` | promptly, with the error | yes |

So the leak is traded for a hang, and only the third row has both properties.

On the tests. `TestMultiNamespaceCacheStart_GoroutineLeak_ContextCancelWithError` is nearly the right shape already, a slow erroring cache alongside two on `<-ctx.Done()`, but it builds the parent with `context.WithTimeout(t.Context(), 50*time.Millisecond)`. The timeout fires, the healthy children return, and `done` closes. Swapping that one line for `context.WithCancel` and asserting `Start` returns is enough to turn it red:

```
NOTIMEOUT Start never returned once the parent context stopped expiring on its own
--- FAIL: TestMultiNamespaceCacheStart_GoroutineLeak_ContextCancelWithError (3.00s)
```

That matters because a manager's context lives for the process rather than 50ms, so production looks like the `WithCancel` version.

What worked for me in both files:

```go
group, childCtx := errgroup.WithContext(ctx)
for idx := range allCaches {
	cache := allCaches[idx]
	group.Go(func() error {
		return cache.Start(childCtx)
	})
}
return group.Wait()
```

The first error is kept, `childCtx` is cancelled, the healthy children return from their own `<-ctx.Done()`, and `Wait` gives back the first error. The buffered channel and the `done` goroutine both go away with it, so the leak stays fixed. One detail worth deciding: on the plain `ctx.Done()` path the group returns `context.Canceled`, where the current code returns `nil`, so you may want to keep that mapping.

Worth adding the shape that is missing from the suite either way: one cache returning an error immediately, another blocking on `<-ctx.Done()`, a parent context nobody cancels, asserting the parent returns the sentinel promptly and the blocking child observes cancellation. That case fails on this branch and on `main`, for different reasons, and passes with the group.

This review was written in part with the assistance of generative AI. Bcuz my English might be sucks.

pkg/cache/delegating_by_gvk_cache.go
```diff
@@ -88,12 +90,36 @@ func (dbt *delegatingByGVKCache) Start(ctx context.Context) error {
...
+		<-done
+		// Return any error that was captured
+		select {
+		case err := <-errs:
+			return err
+		default:
+			return nil
+		}
 	case err := <-errs:
+		// First error returned, wait for other goroutines to finish
```
thc1006: This is the branch that hangs.

The children were started with the parent's `ctx` a few lines up, so a healthy `Cache.Start` sits in its own `<-ctx.Done()` until something cancels that context. Nothing here does, so `wg.Wait()` never finishes, `done` never closes, and this never returns. I confirmed it on this branch: one child returning a sentinel error and one on `<-ctx.Done()` leaves `Start` blocked past a 5s timeout, where `main` returns immediately.

Giving the children their own cancellable context makes the wait terminate:

```go
group, childCtx := errgroup.WithContext(ctx)
for idx := range allCaches {
	cache := allCaches[idx]
	group.Go(func() error {
		return cache.Start(childCtx)
	})
}
return group.Wait()
```

That also removes the need for the buffered channel and the `done` goroutine, so the original leak stays closed.

pkg/cache/multi_namespace_cache.go
```diff
@@ -160,29 +161,63 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
...
+		<-done
+		// Return any error that was captured
+		select {
+		case err := <-errs:
+			return err
+		default:
+			return nil
+		}
 	case err := <-errs:
+		// First error returned, wait for other goroutines to finish
```
thc1006: Same shape here, and the same fix applies. The children at lines 177 and 186 are started with the parent `ctx`, so this wait cannot finish while any of them is healthy.

This is the one the existing `ContextCancelWithError` test exercises, which is why it is worth pointing out that the test only gets past it because its parent context is a 50ms `WithTimeout`. With `WithCancel` instead, `Start` does not return.

dongjiang1989:
@thc1006 Thanks for the review!

Updated to use `errgroup.WithContext()` as suggested. Added tests with `context.WithCancel` to verify healthy children are cancelled and `Start()` returns promptly.

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	<-ctx.Done()
+	return nil
+}
+func (m *mockCache) WaitForCacheSync(ctx context.Context) bool {
+	return true
+}
+func (m *mockCache) IndexField(ctx context.Context, obj client.Object, field string, extractValue client.IndexerFunc) error {
+	return nil
+}
```
alvaroaleman: Lets just embedd the interface we want to satisfy here to avoid the need for defining all its methods:

```suggestion
type mockCache struct {
	startFunc func(ctx context.Context) error
	Cache
}

func (m *mockCache) Start(ctx context.Context) error {
	if m.startFunc != nil {
		return m.startFunc(ctx)
	}
	<-ctx.Done()
	return nil
}
```

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+// TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors tests that goroutines
+// don't leak when multiple caches return errors simultaneously
+func TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors(t *testing.T) {
+	defer goleak.VerifyNone(t)
```
alvaroaleman: Lets please just use synctest here and in all other tests that use goleak or sleep. That avoids the need for any sleeps and the need to use `goleak`, synctest panics if any goroutines are left after the bubble exits:

```suggestion
func TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors(t *testing.T) {
	synctest.Test(t, func(t *testing.T){
```

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+	c := &multiNamespaceCache{
+		namespaceToCache: namespaceToCache,
+	}
+
+	ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
+	defer cancel()
+
+	// Start should return an error
+	err := c.Start(ctx)
```
alvaroaleman: Lets use gomega please

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+	c := &multiNamespaceCache{
+		namespaceToCache: namespaceToCache,
+	}
+
+	// Use a short timeout to trigger context cancellation
+	ctx, cancel := context.WithTimeout(t.Context(), 50*time.Millisecond)
+	defer cancel()
+
+	wg.Go(func() {
+		_ = c.Start(ctx)
```
alvaroaleman: Lets check the error

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+	// Wait for Start to return
+	wg.Wait()
+
+	// Give time for goroutines to leak if they will
+	time.Sleep(50 * time.Millisecond)
+}
+
+// TestMultiNamespaceCacheStart_ErrorWithHealthyChildren tests that when one cache
+// returns an error, the healthy children are cancelled and Start returns promptly.
+// This is the critical test case that verifies errgroup behavior.
```
alvaroaleman: I think we can trust errgroup to be correct/Don't need this test

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,375 @@
...
+func (m *mockCache) WaitForCacheSync(ctx context.Context) bool {
+	return true
+}
+func (m *mockCache) IndexField(ctx context.Context, obj client.Object, field string, extractValue client.IndexerFunc) error {
+	return nil
+}
+
+// TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors tests that goroutines
+// don't leak when multiple caches return errors simultaneously
+func TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors(t *testing.T) {
```
alvaroaleman: Lets please parallelize all of these:

```suggestion
func TestMultiNamespaceCacheStart_GoroutineLeak_MultipleErrors(t *testing.T) {
t.Parallel()
```

dongjiang1989: Thanks. added

pkg/cache/goroutine_leak_test.go
dongjiang1989: removed

pkg/cache/goroutine_leak_test.go
dongjiang1989: Added

dongjiang1989:
Thanks @alvaroaleman 
Please re-check it. thanks

sbueringer:
/assign
(also want to take a look before merge)

pkg/cache/multi_namespace_cache.go
```diff
@@ -160,31 +161,35 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
...
-		return err
+
+	return ignoreContextCanceled(ctx, group.Wait())
+}
+
+// ignoreContextCanceled returns nil if the parent context was cancelled,
+// otherwise returns the error from group.Wait().
+// This preserves the original semantics where context cancellation returns nil.
+func ignoreContextCanceled(ctx context.Context, err error) error {
+	if ctx.Err() != nil {
```
alvaroaleman: @sbueringer this I think is the only part that may be debatable, it mimicks the current behavior however arguably we should ignore all errors that are context cancelled errors, not ignore all errors if the context is cancelled...? The difference would only matter if we hit an error at approximately the same time as the context gets cancelled

alvaroaleman (commented):
Looks good, just the point about ignoreContextCanceled, lets wait for Stefans take there

pkg/cache/multi_namespace_cache.go
sbueringer: > we should ignore all errors that are context cancelled errors

Agree, let's do that instead

pkg/cache/multi_namespace_cache.go
```diff
@@ -160,31 +161,35 @@ func (c *multiNamespaceCache) GetInformerForKind(ctx context.Context, gvk schema
...
-		return nil
-	case err := <-errs:
-		return err
+
+	return ignoreContextCanceled(ctx, group.Wait())
+}
+
+// ignoreContextCanceled returns nil if the parent context was cancelled,
+// otherwise returns the error from group.Wait().
+// This preserves the original semantics where context cancellation returns nil.
```
sbueringer: I would drop the comments that are mentioning that the behavior is the same as some older version of the code

pkg/cache/goroutine_leak_test.go
```diff
@@ -0,0 +1,159 @@
...
+	if m.startFunc != nil {
+		return m.startFunc(ctx)
+	}
+	<-ctx.Done()
+	return nil
+}
+
+// TestMultiNamespaceCacheStart_MultipleErrors tests that Start() returns an error
+// when multiple caches return errors simultaneously
+func TestMultiNamespaceCacheStart_MultipleErrors(t *testing.T) {
```
sbueringer: I would prefer if we could integrate these test cases in our existing test coverage vs having an additional file just for these test cases

pkg/cache/multi_namespace_cache.go
dongjiang1989: Thanks @alvaroaleman  Changed

pkg/cache/multi_namespace_cache.go
dongjiang1989: Removed

pkg/cache/goroutine_leak_test.go
dongjiang1989: Move test case to existing file

alvaroaleman (approved):
lgtm, leaving on hold so stefan can have a look
/hold

sbueringer:
Thx!

/lgtm
/approve
/hold cancel

PR #3566, opened 2026-08-05
:book: Stop suggesting to use an api reader

alvaroaleman:
It's bad suggestion that is terrible for performance and still can not fully guarantee correctness, because a parallel request may be mid commit in etcd, so whatever you see in the apireader may be outdated by the time you see it.

/assign @sbueringer


sbueringer:
Thx!

/lgtm
/approve

PR #3571, opened 2026-08-16
:bug: Fakeclient: Fix PartialObjectMeta handling

alvaroaleman:
Prior to this, it would:
* Panic when getting a PartialObjectMeta in the builder
* Panic when listing a type known to the scheme where we added objects as PartialObjectMeta

/assign @sbueringer

### DIFF
--- a/pkg/client/fake/client.go
+++ b/pkg/client/fake/client.go
@@ -343,14 +343,15 @@ func (f *ClientBuilder) Build() client.WithWatch {
 
 const trackerAddResourceVersion = "999"
 
-// convertFromUnstructuredIfNecessary will convert runtime.Unstructured for a GVK that is recognized
-// by the schema into the whatever the schema produces with New() for said GVK.
+// convertFromUnstructuredOrPartialObjectMetaIfNecessary will convert runtime.Unstructured or metav1.PartialObjectMetadata
+// for a GVK that is recognized by the schema into the whatever the schema produces with New() for said GVK.
 // This is required because the tracker unconditionally saves on manipulations, but its List() implementation
 // tries to assign whatever it finds into a ListType it gets from schema.New() - Thus we have to ensure
 // we save as the very same type, otherwise subsequent List requests will fail.
-func convertFromUnstructuredIfNecessary(s *runtime.Scheme, o runtime.Object) (runtime.Object, error) {
-	u, isUnstructured := o.(runtime.Unstructured)
-	if !isUnstructured {
+func convertFromUnstructuredOrPartialObjectMetaIfNecessary(s *runtime.Scheme, o runtime.Object) (runtime.Object, error) {
+	_, isUnstructured := o.(runtime.Unstructured)
+	_, isPartial := o.(*metav1.PartialObjectMetadata)
+	if !isUnstructured && !isPartial {
 		return o, nil
 	}
 	gvk := o.GetObjectKind().GroupVersionKind()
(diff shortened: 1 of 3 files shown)


pkg/client/fake/client.go
```diff
@@ -343,14 +343,15 @@ func (f *ClientBuilder) Build() client.WithWatch {
...
 const trackerAddResourceVersion = "999"
 
-// convertFromUnstructuredIfNecessary will convert runtime.Unstructured for a GVK that is recognized
-// by the schema into the whatever the schema produces with New() for said GVK.
+// convertFromUnstructuredIfNecessary will convert runtime.Unstructured or metav1.PartialObjectMetadata
+// for a GVK that is recognized by the schema into the whatever the schema produces with New() for said GVK.
 // This is required because the tracker unconditionally saves on manipulations, but its List() implementation
 // tries to assign whatever it finds into a ListType it gets from schema.New() - Thus we have to ensure
 // we save as the very same type, otherwise subsequent List requests will fail.
 func convertFromUnstructuredIfNecessary(s *runtime.Scheme, o runtime.Object) (runtime.Object, error) {
```
sbueringer: Let's rename the func

pkg/client/fake/client_test.go
```diff
@@ -372,6 +372,146 @@ var _ = Describe("Fake client", func() {
 			Expect(partialObjMeta.APIVersion).To(Equal("v1"))
 		})
 
+		It("should be able to Create an unregistered type using PartialObjectMetadata", func(ctx SpecContext) {
```
sbueringer: Just to double check, it is actually possible to create an object with PartialObjectMeta? (with the real client)

pkg/client/fake/versioned_tracker.go
```diff
@@ -87,6 +87,18 @@ func (t versionedTracker) Add(obj runtime.Object) error {
 				return fmt.Errorf("invalid managedFields on %T: %w", obj, err)
 			}
 		}
+		// The trackers add panics when using PartialObjectMetadata without registering it to
+		// sche scheme.
```
sbueringer:
```suggestion
		// the scheme.
```

pkg/client/fake/versioned_tracker.go
```diff
@@ -87,6 +87,18 @@ func (t versionedTracker) Add(obj runtime.Object) error {
...
+		if _, isPartial := obj.(*metav1.PartialObjectMetadata); isPartial {
+			gvk, err := apiutil.GVKForObject(obj, t.scheme)
+			if err != nil {
+				return fmt.Errorf("failed to get gvk for %T: %w", obj, err)
+			}
+			if !t.scheme.Recognizes(gvk) {
+				t.scheme.AddKnownTypeWithName(gvk, obj)
```
sbueringer, line 98: Q: Should this really add obj directly to the scheme or maybe a PartialObjectMeta with (at most) GVK set

pkg/client/fake/versioned_tracker.go
alvaroaleman: Fixed

pkg/client/fake/client_test.go
alvaroaleman: No, but only because our clients reject it, not because it isn't possible (for types whose validation allows creation using only metadata). We also reject update in our client despite it also being possible to do that if the object can be metadata only. I've removed the two tests, the main thing I care about is really being able to initialize a fake client with a PartialObjectMeta

pkg/client/fake/versioned_tracker.go
alvaroaleman: Sorry, I don't think I understand the question. `obj` is a PartialObjectMeta at this point. At least one type must be registered for gvk here, otherwise the trackers `Add` fails, so we register gvk as PartialObjectMeta

pkg/client/fake/client.go

pkg/client/fake/versioned_tracker.go
sbueringer: I was thinking that the object passed into this func might contain name/namespace/labels/annotations/managedfields etc.. So I was wondering if we should add a clean minimal PartialObjectMeta in the scheme instead. But maybe my assumption is wrong

alvaroaleman: Does that matter? My understanding is the scheme is fundamentally a map gvk to a list of go types, so it only cares about the type of the variable you give it when registering something, not the actual content

sbueringer: You are right!

PR #3573, opened 2026-08-18
:sparkles: Allow overriding REST client latency histogram buckets

### LINKED ISSUE #3559: Opt-in REST client metrics: allow overriding histogram buckets (sub-5ms resolution for fast controllers)
### Context

We've been using the opt-in REST client metrics from #3510 (thanks for that work — also for the cherry-pick to release-0.24 in #3525) during a controller latency investigation in kubernetes-sigs/agent-sandbox, where they were instrumental in finding client-side bottlenecks (e.g. HTTP/2 stream-limit queuing that is invisible to server-side metrics).

### Problem

The classic buckets for `rest_client_request_duration_seconds` and `rest_client_rate_limiter_duration_seconds` start at 5ms:

```go
Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
```

For controllers whose hot-path requests are fast — in our case warm-path PATCH/PUT calls with real p50s in the 1–15ms range — everything below 5ms collapses into the first bucket and the 5–25ms span into the second, so p50/p90 estimates over the exact range being optimized are mostly quantization artifact.

We're aware `NativeHistogramBucketFactor` is set, and native histograms fully solve this where the scrape pipeline supports them — but many pipelines still consume only classic buckets.

### Ask

Since the file documents that the default buckets intentionally mirror core Kubernetes, changing the defaults seems off the table. Instead: would you accept making buckets overridable at registration, e.g. an options variant of `RegisterRESTClientMetrics` (or a `WithBuckets`-style option) that applies caller-supplied buckets before the collectors are registered? Callers who need low-end resolution could then prepend e.g. `0.001, 0.0025` without any divergence in the defaults.

Happy to send a PR if maintainers are open to it.

wieghx:
## What

Make classic histogram buckets for the opt-in REST client duration metrics overridable at registration time.

`RegisterRESTClientMetrics` is unchanged and still uses the Kubernetes-default buckets (starting at 5ms). Callers that scrape only classic buckets and need sub-5ms resolution can do:

```go
metrics.RegisterRESTClientMetricsWithOptions(metrics.RESTClientMetricsOptions{
    DurationBuckets: []float64{0.001, 0.0025, 0.005, 0.025, 0.1, 0.25, 0.5, 1, 2, 4, 8, 15, 30, 60},
}, metrics.MetricRequestLatency, metrics.MetricRateLimiterLatency)
```

This applies to:

- `rest_client_request_duration_seconds`
- `rest_client_dns_resolution_duration_seconds`
- `rest_client_rate_limiter_duration_seconds`

Native histogram settings (`NativeHistogramBucketFactor` etc.) are not changed. Size histograms are not changed. The first `RegisterRESTClientMetrics*` call wins so already-registered collectors are not replaced.

## Why

Follow-up to #3510 / #3559. Fast controller hot-path PATCH/PUT latencies often sit in the 1–15ms range; with the default 5ms / 25ms buckets, p50/p90 on classic histograms is mostly quantization artifact. Changing the defaults would diverge from Kubernetes core, so this is opt-in only.

## How to verify

```bash
go test ./pkg/metrics/ -count=1
```

This PR was written in part with the assistance of generative AI.

Fixes #3559

### DIFF
--- a/pkg/metrics/client_go_adapter.go
+++ b/pkg/metrics/client_go_adapter.go
@@ -36,6 +36,25 @@ const (
 	verbLabel = "verb"
 )
 
+// defaultRESTClientDurationBuckets matches Kubernetes core controller REST client metrics.
+// They start at 5ms; override via RESTClientMetricsOptions.DurationBuckets if a scrape
+// pipeline still consumes classic buckets and needs sub-5ms resolution.
+var defaultRESTClientDurationBuckets = []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0}
+
+func restClientDurationHistogram(name, help string, labels []string, buckets []float64) *prometheus.HistogramVec {
+	if len(buckets) == 0 {
+		buckets = defaultRESTClientDurationBuckets
+	}
+	return prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:                            name,
+		Help:                            help,
+		Buckets:                         buckets,
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: 1 * time.Hour,
+	}, labels)
+}
+
 var (
 	// client metrics.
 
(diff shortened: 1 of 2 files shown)


wieghx:
CLA is signed (`cncf-cla: yes`). Happy to address review comments.

Could an org member please add `/ok-to-test` when you get a chance?

pkg/metrics/client_go_adapter.go
```diff
@@ -164,18 +168,78 @@ func registerClientMetrics() {
...
+
+var applyDurationBucketsOnce sync.Once
+
+func applyDurationBuckets(buckets []float64) {
+	applyDurationBucketsOnce.Do(func() {
+		if len(buckets) == 0 {
+			return
+		}
+		requestLatency = restClientDurationHistogram(
+			"rest_client_request_duration_seconds",
```
alvaroaleman: Why are we duplicating the name, description and verbs between here and the package level variables? Please come up with a way to avoid that, like a per-metric constructor that is called in both cases

pkg/metrics/client_go_adapter.go
```diff
@@ -231,27 +295,27 @@ func (r *resultAdapter) Increment(_ context.Context, code, method, host string)
 }
 
 type latencyAdapter struct {
-	metric  *prometheus.HistogramVec
+	metric  func() *prometheus.HistogramVec
```
alvaroaleman: This is a datarace and having package-level variables that we then indirect seems like a needlessly complicated design. Can we clean this up such that:
* The various adapters become pkg-level variables rather than the metric itself
* The adapters use an `atomic.Pointer[*prometheus.HistogramVec]` that is nil by default
* We get rid of the `enabled *atomic.Bool`
* The adapters load the atomic.Pointer and do nothing if its nil in their `Observe`
* The only place where we set the adapters atomic.Pointer is in `RegisterRESTClientMetricsWithOptions`, potentially with custom buckets

pkg/metrics/client_go_adapter.go
wieghx: Name/help/labels now live only in the per-metric constructors (`newRequestLatency`, `newResolverLatency`, `newRateLimiterLatency`). Those are the only call sites; `RegisterRESTClientMetricsWithOptions` uses them when it first stores the collector on the adapter.

pkg/metrics/client_go_adapter.go
wieghx: The opt-in adapters are now the package-level variables. Each holds an `atomic.Pointer[prometheus.HistogramVec]` (`atomic.Pointer[T]` already stores `*T`) that stays nil until register. `Observe` loads and returns if nil, so the `enabled` flags are gone.

The pointer is stored only in `RegisterRESTClientMetricsWithOptions`, which is also where custom duration buckets are applied. `go test -race ./pkg/metrics/` is clean.

alvaroaleman (approved):
thanks

PR #3577, opened 2026-08-23
:seedling: Consolidate client option wrappers behind a shared internal backend

### LINKED ISSUE #2888: Find one scalable pattern to wrap clients, and deprecate others

yindia:
**What does this do, and why do we need it?**

Fixes #2888.

#2888 asks for one scalable pattern to wrap clients instead of adding a new wrapper for every option. Today `NewDryRunClient`, `WithFieldOwner`, and `WithFieldValidation` each ship their own near-identical `Client` and `SubResourceClient` implementation: embed a reader, re-implement every write method to push a single option into the per-call slice, forward the rest of the interface. Three copies of the same plumbing that all need updating whenever the `Client` interface grows a verb.

This PR moves all three onto one internal backend and deletes the duplicated wrapper types.

`pkg/client/defaultoptions.go` adds `withInjectedOptions(c Client, defaults, forced writeOptions) Client`, backed by `clientWithOptions` and `subResourceClientWithOptions`. `writeOptions` holds the option slices per write verb, including the subresource verbs.

Options apply in the order `defaults -> per-call -> forced`, so a per-call option overrides a default and a forced option wins over whatever the caller passed. That keeps each wrapper's existing behavior:

- `WithFieldOwner` and `WithFieldValidation` register their option as a default, so callers can still override it per call.
- `NewDryRunClient` registers `DryRunAll` as a forced option, so it cannot be overridden.

Each write method is implemented explicitly rather than promoted through embedding. If a new write verb is added to the `Client` interface, this file stops compiling until the verb is handled, instead of silently passing calls through without the configured options.

No public API or behavior changes. `fieldowner.go`, `fieldvalidation.go`, and `dryrun.go` now just call `withInjectedOptions`. New `pkg/client/defaultoptions_test.go` covers the default/forced precedence and subresource propagation; `go test ./pkg/client/...` passes.


yindia:
@camilamacedo86 now it's ready

PR #3580, opened 2026-08-25
🌱 Bump to golangci-lint v2.13.1

dongjiang1989:
## What does this do?

ref: https://github.com/kubernetes-sigs/controller-tools/pull/1469

- Upgrade golangci-lint from v2.12.2 to v2.13.1
- Use `strings.Cut` instead of `strings.SplitN` in gomodcheck
- Remove `fmtappendf` from modernize disable list (removed in v2.13.1)
- Fix SA1019 staticcheck deprecation warnings:
  - Replace deprecated `Requeue: true` with `RequeueAfter` in examples
  - Use `metav1.NewFieldsV1()` instead of deprecated `Raw` field in tests
  - Add `//nolint:staticcheck` for intentional deprecated `Requeue` usage in controller tests

## Why?

Keep CI tooling current and resolve new deprecation warnings introduced by the linter upgrade.


dongjiang1989:
cc @sbueringer PTAL, thanks

examples/crd/main.go
```diff
@@ -73,7 +73,8 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{}, err
 		}
 
-		return ctrl.Result{Requeue: true}, nil
+		// Use short backoff instead of deprecated Requeue:true to resolve SA1019 staticcheck warning
+		return ctrl.Result{RequeueAfter: 1 * time.Millisecond}, nil
```
alvaroaleman: This shouldn't use either of the two, can you just remove it?

dongjiang1989: Thanks. Removed

alvaroaleman: This is not resolved, now its using Requeue: true. Please just make it `return ctrl.Result{}, nil`

dongjiang1989: @alvaroaleman   I miss something？
ref ： https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L507-L516

`return ctrl.Result{}, nil` != `ctrl.Result{Requeue: true}, nil`

dongjiang1989: cc @sbueringer wdyt?

sbueringer: The Delete should enqueue another event/Reconcile and because of that an explicit Requeue is not needed. So `return ctrl.Result{}, nil` is the ideal option here

dongjiang1989: Thanks @alvaroaleman @sbueringer  I got it.
Fixed

dongjiang1989:
@sbueringer @alvaroaleman  Changed. PTAL, thanks

sbueringer:
Thank you!

/lgtm
/assign @alvaroaleman

alvaroaleman (approved):
thanks!

PR #3585, opened 2026-09-01
🐛 fakeclient: Allow updating managedFields through Update

### LINKED ISSUE #3484: Fake client: ManagedFields modified using csaupgrade.UpgradeManagedFields are not persisted correctly

locker95:
Fixes #3484

Picks up where #3485 left off (thanks @metlos for the original work). The fake client's update path unconditionally discarded the managed fields of the incoming object and re-read them from the tracker, so explicitly modifying managedFields through Update — which the API server allows, and which e.g. client-go's csaupgrade package relies on to migrate objects from client-side to server-side apply — was silently ignored.

The update path now only retains the stored managed fields when the incoming object doesn't set any (mirroring how the API server treats requests without managed fields), and on status updates, since managed fields can only be modified through the main resource. Compared to #3485 this keeps the retention for main-resource updates that don't set managedFields, so custom object trackers keep their current behavior.

Following the review feedback on #3485, there is one testcase per operation: update (fails without the fix), merge patch (already worked, pinned as a regression test), and a scale-subresource test validating that managed fields set on the scale body do not end up on the object. A create testcase was intentionally left out: with the field-managed tracker, managed fields provided on create are computed by the upstream tracker (the manager comes back as "unknown"), which is outside what #3484 reports and can't be changed from the fake client side.

This PR was written in part with the assistance of generative AI.


pkg/client/fake/client.go
```diff
@@ -828,12 +828,19 @@ func (c *fakeClient) update(obj client.Object, isStatus bool, opts ...client.Upd
...
+	// fields, and always on status updates, as managed fields can only be
+	// modified through the main resource: this is for example used by the
+	// csaupgrade package in client-go to migrate objects from client-side to
+	// server-side apply.
 	// We can ignore all errors here since update will fail if we encounter an error.
-	obj.SetManagedFields(nil)
-	current, _ := c.tracker.Get(gvr, accessor.GetNamespace(), accessor.GetName())
-	if currentMetaObj, ok := current.(metav1.Object); ok {
-		obj.SetManagedFields(currentMetaObj.GetManagedFields())
+	if isStatus || len(obj.GetManagedFields()) == 0 {
```
alvaroaleman: I think this has to be:

```suggestion
	if isStatus || obj.GetManagedFields()) == nil {
```

ref https://github.com/kubernetes/kubernetes/blob/241138b4469f65f908c50b54094a0e776115d750/staging/src/k8s.io/apimachinery/pkg/util/managedfields/internal/fieldmanager.go#L170-L183 which doesn't retain on non-nil but empty ManagedFields slice

locker95:
yeah, empty slice shouldn't be treated like unset. switched the check to `== nil` so a non-nil empty list is kept (matches fieldmanager). extra paren in the suggestion didn't compile so I didn't apply it as-is.

locker95:
lint was SA1019 on FieldsV1.Raw in the tests — switched those to NewFieldsV1.

alvaroaleman (approved):
I think this is correct now, cc @sbueringer 
/hold

So Stefan can take a look

sbueringer:
> I think this is correct now

Yup, agree

/lgtm
/approve
/hold cancel

PR #3586, opened 2026-09-04
🐛 Fix signalProcess redeclaration on Windows

jandubois:
`signal_windows.go` declares `signalProcess` as a function, while `process.go` declares it as a var holding `signalProcessImpl`. On Windows the two collide and `signalProcessImpl` is undefined, so anything importing `pkg/envtest` fails to compile. #3519 added the Windows file; #3523 then added the indirection and renamed the unix and other implementations but not this one.

v0.24.1 cross-builds for Windows, so this is a regression in v0.25.0 and needs a cherry-pick to `release-0.25`.

Verified with `GOOS=windows go build ./...`, plus `go test ./pkg/internal/testing/process/...` on darwin. `GOOS=windows go vet ./pkg/...` still fails: `process_test.go` calls `syscall.Kill` and `syscall.Getpgid`, so the package's tests do not build on Windows. That is separate from this compile break and I left it alone.


pkg/internal/testing/process/signal_windows.go
```diff
@@ -23,6 +23,6 @@ import (
 	"syscall"
 )
 
-func signalProcess(process *os.Process, _ syscall.Signal) error {
+func signalProcessImpl(process *os.Process, _ syscall.Signal) error {
```
alvaroaleman, line 26: I don't think this is very useful unless we also enable testing that c-r builds on windows

jandubois: I totally agree that it would be great to run CI on Windows as well. But I don't understand why that would block fixing a regression.

#3519 split off `signal_windows.go` from `signal_others.go`.

#3523 changed the declaration for `signalProcess` in `signal_unix.go` and `signal_others.go`, but didn't know about `signal_windows.go`.

This happened because #3523 was created on May 21st while #3519 wasn't merged until June 13th. So `signal_windows.go` did not exist yet on `main` by the time the second PR was written.

It is the combination of these 2 PRs that is breaking compilation on Windows, and the fix is a trivial 1 line change that is identical to what #3523 did for `signal_unix.go` and `signal_others.go`.

So while this change currently won't be verified by CI because it doesn't run on Windows, I don't think restoring compatibility is useless. But a manual comparison with `signal_unix.go` and `signal_others.go` should show that this is the right change.

jandubois: @sbueringer Since you approved #3519, which created `signal_windows.go`, could you have a look, if you have the time?

alvaroaleman: He is on pto. Fixing a regression without testing just means it's likely to break again.

sbueringer: Is there a way we could enable some testing on Windows?

PR #3588, opened 2026-09-08
🐛 priorityqueue: fix data race on []*item in logState

lanfix:
### What this does

`logState` collected the queue's items into a `[]*item[T]` under `w.lock`,
released the lock, and *then* passed those pointers to the logger for
serialization. The pointed-to `item[T]` values stay live in `w.items` /
`w.waiting` / `w.ready` and keep being mutated by the other queue goroutines
while the logger walks them, so the snapshot is a data race.

This PR takes a by-value deep copy of every item (including the `*time.Time`
behind `ReadyAt`) under `w.lock` and hands that copy to the logger. The lock is
held only for the cheap copy, not for serialization, and the logged payload is
fully detached from the live queue state. Log output is unchanged (`[]item[T]`
and `[]*item[T]` marshal identically; `readyAt,omitempty` still works).

### Why we need it

With a logger that has `V(5)` enabled, `logState` runs every 10s. The race is
observable via `go test -race`, and in production it intermittently crashes the
controller process:


pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -537,20 +537,36 @@ func (w *priorityqueue[T]) logState() {
...
-			return true
-		})
-		w.lock.Unlock()
 
-		w.log.V(5).Info("workqueue_items", "items", items)
+		w.log.V(5).Info("workqueue_items", "items", w.cloneItems())
+	}
+}
+
+// cloneItems returns a by-value deep copy of every queued item that is safe to
```
alvaroaleman: Please reduce this comment to three lines at most rather than letting your llm spit out a large block of text just to say we are deepcopying something to avoid a data race

pkg/controller/priorityqueue/priorityqueue.go
```diff
@@ -537,20 +537,36 @@ func (w *priorityqueue[T]) logState() {
...
+// use after w.lock is released. logState serializes the result outside the lock
+// while other goroutines keep mutating the live items (handleWaitingItems sets
+// ReadyAt to nil as an item becomes ready, lockedAddWithOpts reassigns it on the
+// update path), so handing the *item[T] pointers straight to the logger is a
+// data race. The race can crash the process from within encoding/json when a
+// ReadyAt pointer is observed non-nil and then dereferenced after it went nil:
+// "value method time.Time.MarshalJSON called using nil *Time pointer".
+func (w *priorityqueue[T]) cloneItems() []item[T] {
```
alvaroaleman, line 547: Please add a test for this that validates it copies all fields now and in the future, maybe using randfill or such

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1069,62 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
 		}
 	}
 }
+
+// TestPriorityQueueLogStateDoesNotRaceOnReadyAt is a regression test for a data
```
alvaroaleman: Same comment re comment

alvaroaleman (commented):
/ok-to-test

pkg/controller/priorityqueue/priorityqueue.go
lanfix: Done in [2ba4b6f](https://github.com/kubernetes-sigs/controller-runtime/pull/3588/commits/2ba4b6fdfbacd0132b6063ab0a836ca09fa03c4f)

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done in [2ba4b6f](https://github.com/kubernetes-sigs/controller-runtime/pull/3588/commits/2ba4b6fdfbacd0132b6063ab0a836ca09fa03c4f)

pkg/controller/priorityqueue/priorityqueue.go
lanfix: Done in [2ba4b6f](https://github.com/kubernetes-sigs/controller-runtime/pull/3588/commits/2ba4b6fdfbacd0132b6063ab0a836ca09fa03c4f)

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1074,107 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
...
+				}
+				if _, err := json.Marshal(q.cloneItems()); err != nil {
+					t.Errorf("marshaling queue snapshot: %v", err)
+					return
+				}
+			}
+		})
+	}
+
+	time.Sleep(500 * time.Millisecond)
```
alvaroaleman: Use synctest instead of sleeps, please

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1074,107 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
...
+	close(stop)
+	wg.Wait()
+}
+
+// TestPriorityQueueCloneItemsCopiesAllFields checks that cloneItems returns items that are
+// equal to the queued ones and share no memory with them. Filling the items with random
+// values keeps both checks valid for fields that are added to item in the future.
+func TestPriorityQueueCloneItemsCopiesAllFields(t *testing.T) {
```
alvaroaleman: Please factor the code such that this can test only the cloning, not a bunch of unrelated things. Also, the testname is self-explanatory and doesn't require a comment

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1074,107 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
...
+				}
+				q.AddWithOpts(AddOpts{
+					After:    time.Duration(rand.IntN(3)) * time.Millisecond,
+					Priority: new(rand.IntN(5)),
+				}, strconv.Itoa(rand.IntN(keySpace)))
+			}
+		})
+	}
+
+	// Readers: do what logState does - snapshot the queue and serialize it.
```
alvaroaleman: Can we simply pass in a logger at v5 instead to test what actually does happen rather than imitate it? The race detector will go off if a race could happen, it doesn't require the race to actually happen

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1074,107 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
...
+	defer q.ShutDown()
+
+	const keySpace = 20
+	stop := make(chan struct{})
+	var wg sync.WaitGroup
+
+	// Writers: churn items through waiting -> ready (handleWaitingItems clears
+	// ReadyAt) and through the update path (lockedAddWithOpts reassigns ReadyAt).
+	for range 8 {
+		wg.Go(func() {
```
alvaroaleman: I don't think using randomized keys adds anything here, the race happens if the after for the same key gets changed which only happens if its lower, so a simpler version of this would be to just add the same key in a loop with a lowering after, maybe by substracting the loop index from a start value or some such

pkg/controller/priorityqueue/priorityqueue_test.go
```diff
@@ -1068,3 +1074,107 @@ func TestPriorityQueueDoesNotDeadlockOnShutdownWhileHandingOutItems(t *testing.T
...
+		q.ready.ReplaceOrInsert(readyItem)
+
+		// cloneItems returns the waiting items first, then the ready ones.
+		expected := []item[string]{*waitingItem, *readyItem}
+		clones := q.cloneItems()
+		if diff := cmp.Diff(expected, clones); diff != "" {
+			t.Fatalf("cloned items don't match the queued ones: %s", diff)
+		}
+
+		// Equal fields are not enough, a clone must not reference the item it was cloned from.
```
alvaroaleman: If anything, make this its own test please

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done. The test now runs inside synctest.Test and advances fake time with `time.Sleep(10 * time.Second)` so that logState's 10s tick fires.

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done. Extracted the per-item copy into `item.clone()` and cloneItems just calls it, so the tests only exercise `clone()` on a randfill-ed item and don't build a queue anymore. Dropped the comment. The randfill setup lives in newItemFiller, as randfill can't fill time.Time on its own (all its fields are unexported).

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done. `newQueue` now accepts Opts and the test passes a funcr logger with Verbosity: 5, so the real `logState` does the reading. I verified that the race detector reports the old `logState` implementation with this test (read in `logState` vs. write of `item.ReadyAt` in `lockedAddWithOpts`), and that it is clean with the current one.

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done. The test now re-adds the same key in a loop with a lowering `After` (`time.Hour` minus i minutes) and sleeps between the adds so that logState logs the item in between the `ReadyAt` updates.

pkg/controller/priorityqueue/priorityqueue_test.go
lanfix: Done, split into `TestItemCloneCopiesAllFields` (`cmp.Diff` against the original) and `TestItemCloneDoesNotShareMemoryWithOriginal` (no pointer/map/slice field of the clone may point into the original).

Note that copying the `time.Time` behind `ReadyAt` isn't needed for the race itself (the pointee is never mutated in place), it only guards against a future in-place update.

alvaroaleman (approved):

sbueringer:
Thx! Nice fix!

/cherry-pick release-0.24

PR #3594, opened 2026-09-13
:sparkles: Add metrics handler options

alvaroaleman:
Allows to do thing like enable open metrics handling or custom error handling, while preserving the custom defaults.

Ref https://github.com/kubernetes-sigs/controller-runtime/issues/3592

/hold 
/assign @sbueringer

PR #3595, opened 2026-09-13
⚠️ Enable OpenMetrics by default 

### LINKED ISSUE #3592: metrics: no way to enable OpenMetrics on the manager's /metrics endpoint
Ran into this trying to expose Prometheus exemplars from a project that uses the manager's metrics server, and I can't find a supported way to do it.

Exemplars only serialise over OpenMetrics. The classic text format has no representation for them, so client_golang drops them at encode time - which means the format the endpoint serves decides whether exemplars work at all, regardless of what the project records into its collectors.

The manager's handler is built with only error handling set, so `EnableOpenMetrics` stays false:

```go
handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{
    ErrorHandling: promhttp.HTTPErrorOnError,
})
```

(v0.24.1, `pkg/metrics/server/server.go`)

As far as I can tell there's no way to change that from outside:

- nothing on `metricsserver.Options` for handler options, and `EnableOpenMetrics` doesn't appear anywhere in the package
- `ExtraHandlers` refuses `/metrics` - "overriding builtin /metrics endpoint is not allowed" - enforced in both `Options.Validate` and `AddExtraHandler`
- `FilterProvider` is the one thing left. the filter gets handed the built handler and can just ignore it, returning its own `promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{EnableOpenMetrics: true})` over the same registry

That last one is what we ended up doing in llm-d (llm-d/llm-d-router#2774). It works, but `FilterProvider` exists for authn/authz and we're using it to pick a wire format. The part that bothers me most is that a project running with metrics auth disabled now has to carry a filter purely to change the encoding.

Also noticed the `TODO(JoelSpeed)` on the line that mounts this handler about moving to existing Kubernetes machinery for serving metrics, so this area may already be up for a rework.

What I'd like:

Either an `EnableOpenMetrics bool` on `metricsserver.Options` passed through to the internal `HandlerOpts`, or a `HandlerOpts *promhttp.HandlerOpts` field so this doesn't need a new field each time promhttp grows one. No strong preference between them.

Shouldn't break anyone either way - promhttp negotiates per request off `Accept`, so a scraper that doesn't ask for `application/openmetrics-text` keeps getting the classic format from the same endpoint.

Happy to send the PR if one of those shapes is acceptable.

/kind feature

alvaroaleman:
This change enables the open metrics format by default on the metrics
endpoint. This may be breaking in certain circumstances, because `le`
labels that were previously formatted as integer will get an added `.0`.
Ref https://github.com/prometheus/client_golang/blob/d6087ee482e06716ee21dc03819432d5d40f72db/prometheus/promhttp/http.go#L601

This behavior can be reverted using the metrics servers `HandlerOpts`.

based on top of https://github.com/kubernetes-sigs/controller-runtime/pull/3594
/assign @sbueringer 
/hold

Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/3592


sbueringer:
lgtm pending rebase

PR #3599, opened 2026-09-19
docs(webhook): clarify CertDir default behavior

AruneshDwivedi:
## Problem

The default  for the webhook server uses  which returns a different path on each invocation. This can cause confusion when users expect a consistent certificate directory location.

**Issue**: #900

## Fix

Added a clarifying comment noting that  behavior may cause unexpected path changes, suggesting users set  explicitly for predictable behavior.

Signed-off-by: Arunesh Dwivedi <arunesh.devops@gmail.com>

---

**Release Notes**:
```release-note
NONE (documentation only)
```


sbueringer:
I think https://github.com/kubernetes-sigs/controller-runtime/pull/3551/changes already sufficiently improved the godoc. We just forgot to close #900.

(I also think an additional comment in this place won't make it clearer for folks setting the CertDir field)

